PR #114, opened 2014-12-22
Check media resources and handle them

leonuh:
Videomcu doesn't work if you have only audio input

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1178,26 +1178,26 @@ function Janus(gatewayCallbacks) {
 			// If we got here, we're not screensharing
 			if(media === null || media === undefined || media.video !== 'screen') {
 
-                //Check media sources
-                MediaStreamTrack.getSources(function(sources) {
-                    var audioExist = sources.some(function(source) {
-                            return source.kind === 'audio';
-                        }),
-                        videoExist = sources.some(function(source) {
-                            return source.kind === 'video';
-                        });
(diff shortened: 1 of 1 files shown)


leonuh:
I found something regarding to this issue  - https://github.com/GoogleChrome/webrtc/commit/c6b0c13c08c2de503fd1457d4f31e5f3bc9c4240 - Maybe the adapter.js needs an update too.

lminiero:
Thanks for the heads-up. That adapter also removes some checks that were there before, though, assuming some versions are not around anymore. Not sure how certain that can be. I'll see if I can prepare a new adapter.js that is more comprehensive.

lminiero:
It seems to be working fine with all browsers this way, but apparently there are a few issues with a couple of plugins right now and the way they handle the resulting SDP, which causes Firefox Nightly to rightly complain. I'll fix those in the next few days and then merge your contribution.

lminiero:
Merging, a followup commut to fix things up will follow.

amki:
This leads to a bug with Firefox:

TypeError: MediaStreamTrack.getSources is not a function janus.js:1378:5

MediaStreamTreack.getSources is not (yet?) implemented in Firefox.

lminiero:
> MediaStreamTreack.getSources is not (yet?) implemented in Firefox.

You're probably not using the latest adapter.js we have in the repo.

PR #125, opened 2014-12-30
Janus build fixes for OSX

yultide:


lminiero:
Sorry for this abysmally late reply... I just tried the patch, but on my system it fails immediately when invoking ```autogen.sh``` because it cannot find the m4 folder, probably due to the new line in Makefile.am:

      aclocal: error: couldn't open directory 'm4': No such file or directory

Creating the folder and running ```autogen.sh``` again seems to do the trick, and I can configure, compile, install and use Janus normally.

Not sure what the best way to solve this is: add an empty m4 folder to the project (ugly) or have it automatically created by ```autogen.sh``` if it doesn't exist?

lminiero:
Thinking about it, it makes sense to have autogen create it. Of course this means we'll have to add the m4 folder to the gitignore list as well. I'll make this changes myself after the merge.

PR #139, opened 2015-01-21
Use bower for front end dependency management

scottmas:
Hooked up janus gateway with the Bower package manger system. This enables two thing:

1. Cleans up the project source code by not having to commit external libraries. Now, when we want to build the HTML website, we simply need to have bower installed via node, then do the shell command "bower install" to download and extract dependencies to the "html/bower_components" folder.

2. Potentially exposes the Janus javascript library as a bower package, so that it's easier for developers to integrate the Janus front end library with their code base. If we register with bower, then all that a developer would need to do in order to get janus.js (and it's dependencies like "adapter.js", etc), will be to type "bower install --save janus". If this pull goes through, then we would need to register janus with Bower, by typing "bower register janus git://github.com/meetecho/janus-gateway.git"

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,39 @@
+{
+  "name": "janus-gateway",
+  "version": "0.0.7",
+  "homepage": "https://github.com/meetecho/janus-gateway",
+  "authors": [
+    "Lorenzo Miniero <lorenzo@meetecho.com>",
+    "Philip Withnall <philip@tecnocode.co.uk>",
+    "Jack Leigh",
+    "Pierce Lopez <pierce.lopez@gmail.com>",
+    "Benjamin Trent <ben.w.trent@gmail.com>",
+    "Dustin Oprea <dustin@randomingenuity.com>",
+    "Maurizio Porrato",
+    "Giacomo Vacca",
+    "mrauhu",
+    "Min Wang",
+    "leonuh",
+    "Nicholas Wylie",
+    "Graeme Yeates <yeatesgraeme@gmail.com>",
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for this! I'm not familiar with bower and not particularly keen on node.js in particular, but I'm sure it will be useful to a lot of web developers out there.

That said, I personally don't like the strong dependency on the bower environment in the demos (see, paths). I'm not going to use it, as many others probably won't, and as such I'd rather keep it separated from the main purpose and target of the gateway. Can this be done in a separate folder and/or project instead?

scottmas:
I can totally see your reasoning Lorenzo. Using tools like Bower probably makes more sense on a actual web application like project rather than demo app you have here. a A simpler approach would be only focus on advantage (2) I talked about above, and get rid of "advantage" (1). 

Let me make the changes and let you know.

ploxiln:
latest version looks a lot better than the first, to me :)

scottmas:
And to clarify a bit, Bower is simply an online registry that points pretty names, like "janus" to a specific github and specific file within that github. So after registering janus, front end developers could type "bower install janus" to download janus.js and adapter.js (which is a dependency). The package registry is quite large and becoming pretty mainstream among front end developers.

lminiero:
Hi Scott, sorry for the lack of feedback so far but I didn't have time to look into this these last few days. I'll definitely get back to you tomorrow. Considering there's simply a single file to add I don't think there will be any issue in adding this to the package.

lminiero:
Finally merged, thanks again :)

PR #142, opened 2015-01-25
Moved some includes for easier plugin building

nowylie:
Moved headers around to make it easier to build plugins outside of the Janus source tree.

Plugins can then be built with `gcc -o <plugin>.so <plugin>.c -fpic -shared -l<libs>` without having to recompile the entire source tree.

For example, the audiobridge demo can be built with ``gcc -o janus_audiobridge.so janus_audiobridge.c -fpic -shared `pkg-config --cflags glib-2.0 sofia-sip-ua` -lopus``

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -25,6 +25,7 @@
 #include "cmdline.h"
 #include "config.h"
 #include "apierror.h"
+#include "debug.h"
 #include "rtcp.h"
 #include "sdp.h"
 #include "utils.h"
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -41,6 +41,7 @@ record_file =	/path/to/recording.wav (where to save the recording)
 #include <opus/opus.h>
 #include <sys/time.h>
 
+#include "../debug.h"
 #include "../apierror.h"
 #include "../config.h"
 #include "../mutex.h"
--- a/plugins/janus_echotest.c
+++ b/plugins/janus_echotest.c
@@ -26,6 +26,7 @@
 
 #include <jansson.h>
 
+#include "../debug.h"
 #include "../apierror.h"
 #include "../config.h"
 #include "../mutex.h"
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -44,6 +44,7 @@
 #include <sys/time.h>
 #include <jansson.h>
 
+#include "../debug.h"
 #include "../apierror.h"
 #include "../config.h"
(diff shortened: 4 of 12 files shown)


ploxiln:
ah, this doesn't actually move any files, it just includes `debug.h` and `apierror.h` explicitly instead of implicitly through `plugin.h`. I'm surprised that makes a difference for this case though. In fact, without this change, I'm able to compile plugins one at a time, exactly as you describe.

nowylie:
Sorry, I wasn't very clear. The difference it makes is enabling you to build plugins outside of the main source tree.

If we could then install the `plugin.h` header into an include directory, there could be a pkg-config for building plugins.

ploxiln:
right, that makes sense

ploxiln:
actually, I do regularly build a plugin outside the source tree, with `-I/path/to/src/janus-gateway` and

```
#include "plugins/plugin.h"
#include "apierror.h"
#include "config.h"
```

at the top of the plugin.c file, and that's worked for a while.

(the awkward bit is that I also need to build objects for apierror.c, config.c, utils.c, rtcp.c from the janus source tree and link them into my plugin because otherwise it can't link, because the functions it uses from those source files aren't in a shared library it can access at build time)

nowylie:
You shouldn't need to build the object files. Janus builds with `--export-dynamic`, which should expose those to any plugin loaded.

I just tried copying the audiobridge demo out of the source tree, and building it with ``gcc -o janus_audiobridge.so janus_audiobridge.c -fpic -shared `pkg-config --cflags glib-2.0 sofia-sip-ua` -I../janus-gateway -lopus`` and it worked fine.

ploxiln:
You're right again, I got that to work for my out-of-tree plugin (though I didn't have to modify any in-tree includes).

nowylie:
Shifting the includes from the `plugin.h` file simple removes any explicit expectation on the location of the `plugin.h` file relative to the rest of the source tree. Otherwise it must always be stored somewhere one directory lower then the rest of the headers. Plus, you might not always want to use some of the helper functions in the rest of the source tree. `plugin.h` is the only include *required* to build a plugin.

ploxiln:
All that's true. This change doesn't seem to be necessary to build out-of-tree plugins, but I never said I didn't think it was a good change for code-style anyway :)

I don't make the decisions around here, but :+1:

lminiero:
Looks like a reasonable change. I'll give it a test later in the afternoon before merging it. Thanks for your contribution!

PR #146, opened 2015-01-28
Modify build to output header files

nowylie:
This causes `make install` to put the header files in `$(prefix)/include/janus` so that plugin code can `#include <janus/plugin.h>` (as long as` $(prefix)/include` is in the `C_INCLUDE_PATH`)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -26,6 +26,11 @@ CLEANFILES = $(NULL)
 
 bin_PROGRAMS = janus
 
+headerdir = $(includedir)/janus
+header_HEADERS = \
+	plugins/plugin.h apierror.h config.h debug.h dtls.h ice.h mutex.h record.h \
+	rtcp.h rtp.h sctp.h sdp.h utils.h
+
 confdir = $(sysconfdir)/janus
 conf_DATA = conf/janus.cfg.sample
 
--- a/configure.ac
+++ b/configure.ac
@@ -86,8 +86,11 @@ AC_CHECK_LIB([srtp],
 AC_CHECK_LIB([websock],
              [libwebsock_init],
              [
-               AC_DEFINE(HAVE_WEBSOCKETS)
-               JANUS_MANUAL_LIBS+=" -lwebsock"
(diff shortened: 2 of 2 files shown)


nowylie:
Second commit fixes #147

lminiero:
Good catch, thanks for this!
About the headers, I have some doubts about that. Just providing the headers for building a plugin may not work or be enough, as all of those have the actual implementations in the related .c files that are part of the core itself (including plugin.c). Should there be something like a shared library plugins can link to for the features they may be interested in?

lminiero:
Anyway, that's probably another issue we can discuss separately. Your changes make already a lot of sense so I'll merge them right away. Thanks again!

nowylie:
Linking to a shared library isn't required. As discussed in the last pull request, because Janus builds with `--export-dynamic`, it exposes symbols to any objects loaded with `dlopen`. As long as you build a plugin with `-shared`, it will have access to the symbols at runtime.

The only objects a plugin needs to link against are those not exposed by Janus. For instance, when building the `janus_audiobridge.c` plugin, we need to link with opus (`-lopus`) as it's not a part of Janus.

lminiero:
Ah thanks for the clarification, I was not aware of that. The ```-export-dynamic``` may have gone away in one of the latest pull requests (see #125), though, in case we'll have to add that again.

lminiero:
More specifically, the comment by the user who provided the patch about the removal of those flags was this:

```The flags were confusing libtool to try to link .so flags on a mac. If I removed them, libtool automatically figured out how to create the .so files on linux and .dylib on mac.```

@yultide does libtool automatically add the ```--export-dynamic``` as needed?

nowylie:
The `Makefile.am` appears to have `AM_CFLAGS += ... -rdynamic`, which should pass `--export-dynamic` to the linker. I'll try and build a plugin in this way with the latest merge.

nowylie:
Everything still works fine :)

lminiero:
Great, so I was worried about nothing! Thanks for your help and insight on this.

nowylie:
You're most welcome. I'm enjoying contributing!

PR #152, opened 2015-02-12
janus.js: make "lowres" "hires" hints affect capture resolution on firefox 35

ploxiln:
I was trying to get 16/9 capture on firefox, and found this technique of controlling the capture resolution to work with current firefox. I got the tip from https://groups.google.com/forum/#!topic/mozilla.dev.media/OxgXMRStD8Q

It's arguable whether you want to constrain min and max so strictly, and maybe cause a video capture to fail. I think that, generally, the browser never treats the constraints strictly, and picks the closest resolution from a limited list of supported resolutions. see https://webrtchacks.com/video-constraints-2/

Another "to taste" part of this is, which resolutions to use. For my application, I used 16/9 resolutions of 320/180, 640/360, 1280/720. (I also added a "stdres" name for the middle resolution.) For this PR I figured I would keep it close to the existing behavior, and just show how to get firefox to work with it.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1016,24 +1016,39 @@ function Janus(gatewayCallbacks) {
 			pluginHandle.consentDialog(true);
 			var videoSupport = isVideoSendEnabled(media);
 			if(videoSupport === true && media != undefined && media != null) {
-				if(media.video === 'lowres') {
-					// Add a video constraint (320x240)
-					if(!navigator.mozGetUserMedia) {
-						videoSupport = {"mandatory": {"maxHeight": "240", "maxWidth": "320"}, "optional": []};
-						Janus.log("Adding media constraint (low-res video)");
-						Janus.log(videoSupport);
+				if(media.video && media.video != 'screen') {
+					var width = 0;
+					var height = 0;
+					if       (media.video === 'lowres') {
+						height = 240;
+						width  = 320;
+					} else if(media.video === 'hires' ) {
+						height = 720;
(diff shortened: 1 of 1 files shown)


lminiero:
The changes look reasonable, and the 16/9 attempted resolution is not a bad idea. I guess that in the future we might add, in the library, another flag where we also specify the ratio in order to get, e.g., 4/3. Thanks!

lminiero:
Just fixed a typo (janus.log --> Janus.log) that caused the JS code to block on an undefined.

ploxiln:
sorry about that, was a copy/paste job, didn't actually test :P

lminiero:
No prob!

lminiero:
Looks like Firefox doesn't digest the 16:9 resolutions...

ploxiln:
Well, 640x360 did make a difference for me. It looked like 4:3 squashed/stretched to 16:9, but the video in the capture preview and on the playback side was 16:9. With 640x480 the preview/playback was 4:3, even if the video element was explicitly sized to 16:9 (the 4:3 image was centered horizontally in the video element).

So, that's correct, firefox doesn't actually capture in 16:9, but it does seem to, at least in one case, stretch and deliver in 16:9. For this point in the development of my project, it was enough :)

lminiero:
Ok, just thought I'd mention it in case it could be useful. I modified janus.js to also allow 16:9 requests for video, which fallback to 4:3 when it's Firefox though for the aforementioned reasons.

PR #155, opened 2015-02-16
recordplay: avoid stopping if already stopped

davibe:
If the client has already requested a ‘stop’ of 
the recording before, the ‘hangup’ procedure 
would try to stop it again resulting in duplicate 
‘stopped’ event sent to the client.


lminiero:
Makes sense, thanks.

PR #157, opened 2015-02-19
recordplay: allow client to specify filename (optional)

davibe:
I found this useful for my integration. See if you care to merge it.


lminiero:
Thanks for this! It's definitely helpful, as we have similar features in the streaming and videoroom plugins.

I'm not convinced by the ```g_stpcpy```, though, especially considering the buffer is limited and this can lead to overflows and crashes. Could you modify the code to use g_snprintf instead as the rest of the code does?

davibe:
sure

davibe:
You forgot about this :)

lminiero:
Apologies, github didn't notify me about the new revision... merging now, thanks!

PR #158, opened 2015-02-19
Fixed GLib-CRITICAL after session timeout

uxmaster:
Fixed GLib-CRITICAL **: g_hash_table_iter_next: assertion


lminiero:
Thanks for that, this was part of the recent cleanup but I guess this wasn't always working as expected. Merging immediately.

PR #163, opened 2015-02-26
Fix Fedora package name for pkgconfig

EvanDotPro:
Checked on F20 and F21 just to be sure.

```
$ sudo yum provides "*/bin/pkg-config"

1:pkgconfig-0.28-3.fc20.x86_64 : A tool for determining compilation options
Repo        : fedora
Matched from:
Filename    : /usr/bin/pkg-config
```


lminiero:
Good catch, thanks! And I should have known as I'm a Fedora user... :-)

PR #164, opened 2015-02-27
Set SIP From header when sending INVITE requests

saghul:


lminiero:
Good catch, it looks like we forgot to add that... thanks!

saghul:
Wow, that was quick!  Thanks! Also, I'm loving Janus so far, it's awesome!

PR #168, opened 2015-03-03
Some fixes for the SIP gateway plugin

saghul:
I have some more in the works, I thought I'd send them in small PRs.

### DIFF
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -350,7 +350,7 @@ function registerUsername() {
 	$('#register').attr('disabled', true).unbind('click');
 	$('#guest').attr('disabled', true);
 	var sipserver = $('#server').val();
-	if(sipserver === "" || sipserver.indexOf("sip:") != 0) {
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks for working on this!

The string dupliation for tags in particular is something I was indeed interested in, as I do recall looking for info in the Sofia-SIP documentation about whether or not this was needed and I couldn't find anything. Since at the time there were some segfaults in the SIP plugin, I assumed that in part they could be blamed on strings that were not duped before passing them to the stack. Can you confirm this is unneeded?

saghul:
I also struggled to find documentation about it, but when a constant is passed nobody tries to free it or it would explode. Also some testing and digging in the FreeSWITCH code strongly suggest is not really necessary :-)

lminiero:
Thanks for the clarification, merging then!

saghul:
Fantastic, thanks!

PR #169, opened 2015-03-04
SIP: some more fixes and extra settings

saghul:


lminiero:
Thanks, this is definitely helpful!

PR #170, opened 2015-03-04
More fixes for the SIP gateway :-)

saghul:


lminiero:
Thanks again :-)

Just as a heads-up for future pull requests, could you keep the same indentation approach we use throughout the rest of the code? I noticed that in some patches you seem to rely on spaces for this, while we make use of tabs which makes the code representation incoherent in some editors.

saghul:
Oh sorry about that! I'll tweak my vim config.

Thanks for merging!

PR #171, opened 2015-03-05
fix memory leak when constructing admin response with handle info

ploxiln:
... was allocating json strings in boolean expression


lminiero:
Good catch, thanks for spotting this!

PR #172, opened 2015-03-06
Fix videoremote id for spinner

LetsVape:
The spinner wasn't working because the getElementById selecter included #

I also included anoteher change in my own UI which you might want to utilize or maybe there's a better way of getting a callback to disable this spinner?

At the moment the spinner is stopped whenever the video element is created, but when it is it could still take a while for the connection to be established and until that time the video element will be black.

To keep the spinner going until the video actually shows video I changed the code which adds the bitrate label to the following;
```javascript
					if(webrtcDetectedBrowser == "chrome") {
						spinstopTimer[remoteFeed.rfindex] = setInterval(function() {
						if(remoteFeed.webrtcStuff==undefined){
							//if the remotefeed was cleared before it reached bitr > 0 clear it so it doesn't overlay another cam
							remoteFeed.spinner.stop();
							spinStopped[remoteFeed.rfindex]=true;
							clearInterval(spinstopTimer[remoteFeed.rfindex]);
							spinstopTimer[remoteFeed.rfindex] = null;
						} else {
						var bitrate = remoteFeed.getBitrate();
						var bitr = parseInt(bitrate);
							if(spinStopped[remoteFeed.rfindex]==false){
								if(bitr>0){
									//Stop the spinner when the bitrate gets over 0
									if(remoteFeed.spinner !== undefined && remoteFeed.spinner !== null){
										remoteFeed.spinner.stop();
										spinStopped[remoteFeed.rfindex]=true;
										clearInterval(spinstopTimer[remoteFeed.rfindex]);
										spinstopTimer[remoteFeed.rfindex] = null;
									}
								}
							}
						}
						}, 250);

						$('#curbitrate'+remoteFeed.rfindex).removeClass('hide').show();
							bitrateTimer[remoteFeed.rfindex] = setInterval(function() {
							// Display updated bitrate, if supported
							var bitrate = remoteFeed.getBitrate();
							$('#curbitrate'+remoteFeed.rfindex).text(bitrate);
						}, 1000);
					} else {
					//Stop the spinner in non-chrome browsers which don't support getBitrate();
					remoteFeed.spinner.stop();
					}
```

And I commented out 
```javascript
				if(remoteFeed.spinner !== undefined && remoteFeed.spinner !== null)
					remoteFeed.spinner.stop();
```

Now as this is using a counter to check the bitrate label and only works in Chrome it's not the most efficient, is there a callback that is received when the video actually starts?

Thanks

### DIFF
--- a/html/videomcutest.js
+++ b/html/videomcutest.js
@@ -396,7 +396,7 @@ function newRemoteFeed(id, display) {
 						remoteFeed.rfid = msg["id"];
 						remoteFeed.rfdisplay = msg["display"];
 						if(remoteFeed.spinner === undefined || remoteFeed.spinner === null) {
-							var target = document.getElementById('#videoremote'+remoteFeed.rfindex);
+							var target = document.getElementById('videoremote'+remoteFeed.rfindex);
 							remoteFeed.spinner = new Spinner({top:100}).spin(target);
 						} else {
 							remoteFeed.spinner.spin();


lminiero:
Thanks for spotting this! We rarely use getElementById directly, so I guess that explains the jQuery kind of selector that broke things :-) Merging now.

About the other fix, this is a known issue but I don't think relying on a browser-specific issue is the right approach for our library. The right thing to do is probably, within the janus.js code, check when the ICE process has successfully resulted in a connection, as after that it's just a matter of doing the DTLS handshake and then receive the media. The period of no-video is due to the time it may take to establish connectivity via ICE after a successful setLocalSDP/setRemoteSDP, which is what triggers the remote stream availability in our code now. I still didn't have any time to work on this though.

LetsVape:
No worries, glad I could contribute a little bit in return at least, hope to get more familiar with the server-side code to get more involved in fixing bugs.

Thanks for your explanation on how to programatically check for the video state without relying on the bitrate counter, I'll spend some time on that to see if I can get it working for the non-chrome users too as this indeed is just a workaround not a proper fix.

gatecrasher777:
The <video> element has an onplaying event which works in both Firefox and Chrome, and should probably work in all HTML5 enabled browsers.  In my app, I use it as a hook to obtain accurate video dimensions.

lminiero:
That's a good hint, especially considering that the ```loadedmetadata``` stuff we have now has never worked consistently on all browsers. This also means that this would need to be done within the web application itself, though, and couldn't be done in the janus.js library which has no view of the DOM elements. That's probably not an issue, though, as it makes sense to have this part of the UI code.

lminiero:
I've played a bit with the ```playing``` event and it does seem to do the job. Nevertheless, getting the video size doesn't seem to work in Firefox 35/36, while it does in Nightly (39). Apparently for Firefox 35/36 you need a setTimeout as we had for loadedmetadata before the size can be effectively retrieved.

That said, I'm happy enough with the event, so I'll update the demos to use that instead.

gatecrasher777:
Yes, you are right, sometimes the dimensions are not ready at that point in Firefox. When the onplaying event fires, I have a test for video width and height, and if they are zero it will repeat the test every 500ms until they become non-zero. Typically, though, it is within a second or two.

lminiero:
Yes, I've left the setTimeout with 2000ms for Firefox stable.

LetsVape:
Great stuff, thanks for the hint gatecrasher, and thanks for adding a commit with it Lorenzo

PR #174, opened 2015-03-06
Recording quality

davibe:
Some steps towards better recording quality

### DIFF
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -351,7 +351,7 @@ function startRecording() {
 		// bitrate can be set at any time: before, after, during recording
 		recordplay.send({"message": {
 			request: 'video-bitrate-max',
-			value: 256
+			value: 256000
 		}});
 		
 		recordplay.createOffer(
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -347,6 +347,13 @@ function startRecording() {
 		$('#list').unbind('click').attr('disabled', true);
 		$('#recset').attr('disabled', true);
 		$('#recslist').attr('disabled', true);
+
+		// bitrate can be set at any time: before, after, during recording
+		recordplay.send({"message": {
+			request: 'video-bitrate-max',
+			value: 256
+		}});
+		
 		recordplay.createOffer(
 			{
 				// By default, it's sendrecv for audio and video...
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -350,15 +350,13 @@ function startRecording() {
 
 		// bitrate and keyframe interval can be set at any time: 
 		// before, after, during recording
-		recordplay.send({"message": {
-			request: 'video-bitrate-max',
-			value: 256000
-		}});
-
-		recordplay.send({"message": {
-			request: 'video-keyframe-interval',
-			value: 15000 // 15 seconds
-		}});
+		recordplay.send({
+			'message': {
+				'request': 'configure',
+				'video-bitrate-max': 256000, // a quarter megabit
+				'video-keyframe-interval': 15000 // 15 seconds
+			}
+		});
 		
 		recordplay.createOffer(
 			{
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -348,10 +348,16 @@ function startRecording() {
 		$('#recset').attr('disabled', true);
 		$('#recslist').attr('disabled', true);
 
-		// bitrate can be set at any time: before, after, during recording
+		// bitrate and keyframe interval can be set at any time: 
+		// before, after, during recording
 		recordplay.send({"message": {
 			request: 'video-bitrate-max',
 			value: 256000
+		}});
+
+		recordplay.send({"message": {
+			request: 'video-keyframe-interval',
+			value: 15000 // 15 seconds
 		}});
 		
 		recordplay.createOffer(
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -52,6 +52,7 @@ var janus = null;
 var recordplay = null;
 var started = false;
 var spinner = null;
+var bandwidth = 1024 * 1024;
 
 var myname = null;
 var recording = false;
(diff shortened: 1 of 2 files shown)


Will-I4M:
could it be extended on other plugins ? (video MCU) ?

davibe:
I would say yes but there are more constraints there and it depends on what you want to achieve exactly. My current priority is simple single face recording. Next thing for me is to send proper RTCP Receiver Reports to the browser to signal the amount of lost packets. First i would like someone to validate my current work. Thank you.

lminiero:
Thanks for working on this.

That said, I don't think this can be merged yet. As far as I've understood, this is sending RTCP feedback for every single RTP packet that comes in, which is way overkill.

Apart from that, could you have the request pass the exact target bitrate, instead of the value/1000? This would make it consistent with the usage we have in other plugins (echotest, videocall, videoroom) where we specify the precise value (e.g., 128000 instead of 128).

davibe:
@lminiero It sends one rtcp every second, not every rtp packet, which is very gentle.
I changed the "bitrate * 1000" thing already.

lminiero:
Right, sorry about that! I thought the per-second check was only being applied to REMB, and not to the RR feedback as well.

That said, you probably don't need the lost count part, as that is already handled by the NACK management in the code: whenever lost packets are detected, a NACK is sent, which has them retransmitted by the browser and as a consequence passed to the plugin, even if a bit late. Since the post processor re-orders the packets before saving them to opus/webm, a late packet doesn't affect the quality of the recording.

davibe:
@lminiero 
I will take a look at how NACKs are generated but i can say it's not working in my case. I tried to set a bitrate higher than what my adsl can manage. Looking at the resulting video recording i see that after the first packet loss the video is totally messed up and it does not recover till the end.

If i send the fir/pli instead, i see it recovering very quickly. 

I think when bandwidth gets saturated NACKs are not enough. Re-transmitting itself causes more traffic and the nacks queue overflows. 
So... Maybe, as an optimization, fir-pli generation should be used when "something" detects NACKs were not successful ? Is there already a way to detect this from the plugin ?

Another simpler option could be to just request a keyframe every N seconds where N can be set by the client.

lminiero:
Yes, in that case you're right: RFCs 4585 and 4588 clarify that NACKs can actually make things worse when retransmissions are involved on a link that can't satisfy the bandwidth requirements, and as such bitrate should be reduced as well after a retransmission. To detect such issues in the plugin you can check the ```slow_link``` event which currently, though, only addresses too many NACKs received and not those sent. I guess that we might change the core to also trigger the callback for the latter, although the callbacks should be extended to clarify in what direction the link is slow.

The "key frame every N seconds" is what we do in the videoroom plugin in a configurable way. That said, in your current code you might want to also force a reduction of the bitrate the user asked through the API, besides sending a FIR/PLI, as a FIR/PLI results in a new key frame being sent which can also kill the bandwitdh since key frames are much more bandwidth "hungry".

Will-I4M:
@Lorenzo I think you're right for the bandwitdh but in some cases the video
is "broken" (if you replay the video, the video stream stop but the sound
is ok for me). I guess there is something to do with keyframes here ?
(maybe if the NACK is not handled correcty, or lost ? ).

lminiero:
That's probably related to what Davide anticipated, NACKs on a too slow connection can make things worse, which is why the bitrate that is set should be lowered for the user who is getting those. Anyway, the videoroom plugin sends a FIR/PLI every X seconds, if configured, which should help.

davibe:
@lminiero: I moved things around so that fir-pli are sent based on an interval set with plugin apis. I think results in better quality recordings already. Please let me know what else I can do to make this merge-able.

On a different branch i am trying to modify ice.c to detect nack retransmission failures which would trigger "slow_link_downstream" event on the plugins. Plugins then can handle that event and lower the bandwidth cap.  
"nack retransmission failure" = a nack has been sent to requested retransmission of a rtp packet but that packet, identified by its seq, has not been received within a certain amount of time (timeout, which could be 1 or 2 seconds probably).
However this is being a bit difficult and time consuming I hope i can get to another pull request.

lminiero:
No need to work on the ice.c stuff in the core, as I'm experimenting with that myself. I'll probably have the same slow_link callback invoked for "publishing" issues too, although in the feature this may be done in a separate callback (or the same callback with a different signature).

I'm working on this to have a slow_link kind of feedback on the receiving side too, which would allow plugins to understand when there are problems in a media "publisher" and act accordingly, e.g., by decreasing the bitrate cap that was set before in a programmatic way. This is related to the work @gatecrasher777 is doing in #175 as a separate effort. As of now I'm doing some tests with the EchoTest plugin, which I'll then extend to VideoCall and VideoRoom in order to involve more loosely couple participants.

As such, although your patch looks merge-able aready, I'd just ask you to wait until I have stabilized and evaluated that part of the code a bit, before merging this patch, as there might be things that could be useful in your modifications too. This will probably happen early next week, as I'm out of the office now.

davibe:
Ok then i can wait :+1:  Oh, let me just suggest that slow_link could also give enough information to fill-in receiver reports (i.e. number of packets lost) so that even the browser side could take actions to decrease bandwidth.. or just collect meaningful stats.

Thank you for your work @lminiero

lminiero:
@davibe I just committed the slow_link changes I mentioned. Specifically, now the callback has been modified by adding a new parameter, which means you should update the patch to take that into account.

Could you also modify the way the bitrate and FIR frequency are modified through the API? In all plugins we have a single ```configure``` request that can then modify one or more things at the same time, so it would be cool to have the Record&Play plugin behave pretty much the same way, rather than having separate requests for the purpose. I know I'm being picky right now, but I'd like to keep the API consistent across plugins when this can help developers.

Anyway, this is something I can do myself after a merge if you're busy, so no problem there! Let me know when the slow_link signature has been updated in the patch and I'll merge it.

Thanks!

davibe:
@lminiero: Today I'm busy. Tomorrow I will have something for you.

lminiero:
No hurry, don't worry, thanks again for your contributions to the project and for "adopting" the plugin to make it better :)

davibe:
@lminiero: I added "configure" api and "slow_link" handling. The bandwidth adapting is on the client-side (application).

lminiero:
Thanks! I'll merge right away.
I'll also turn the json_boolean in a json_string(true/false), though, as dated versions of jansson (e.g., on ubuntu 12.04) don't have json_boolean and would fail with an undefined reference.

lminiero:
About the rampup, we introduced it exactly to address Chrome for monodirectional streams, and in particolar for the videoroom: without it, a REMB message sent at the beginning of a communication would be ignored.

davibe:
According to my tests (using current stable chrome and recordplay) the ramp-up had 0 effects. I left it there because you never know :)
About true/false you could consider using an integer. I think it could be better because 0 is 'falsy' in js while "false" would be true. It looks more "forward-compatible".

lminiero:
That's probably a good point, but it would mean changing too many of the APIs in the other plugins that make use of it as of now. I'll change it to an integer in your plugn, though, as it's a new feature.

PR #175, opened 2015-03-06
Add slow link event to videoroom plugin.

gatecrasher777:
Add slow link event to videoroom plugin


lminiero:
I've been working on a similar approach yesterday, but I still haven't committed anything. Just like you, the idea was not to act directly on the bitrate, as different applications might prefer to apply different policies according to the feedback they receive.

That said, at the moment your approach wouldn't work as the ```slow_link``` callback currently only applies to NACK being received by Janus, and not to those sent. This means you'll always get such a callback only for listeners and not for publishers in the current revision. This is something I modified in a local branch I have, but haven't committed yet. As such, I'll keep the patch "on ice" if it's ok for you for a couple of days, until I stabilize che core features that are required for this to work.

gatecrasher777:
Oh!!!  So session->participant_type == janus_videoroom_p_type_publisher will never be true?  Yes, the main corrective action desired here would be to reduce the publisher's bitrate in response to too many NACKS.  I'm less concerned about the listener's experience, since there is not much corrective action that can be taken.

lminiero:
@gatecrasher777 I've completed the work I anticipated. As you can see, I already integrated the new slow_link callback (which now has a way to discriminate uplink/downlink issues) in the videoroom plugin, which just in your patch results in an event to be notified on the handle. This event is triggered for both publishers and listeners: it's then up to the application to decide what to do with the event. For instance, for a publisher it may be reducing the bitrate (what we now try and do automatically in the echotest and videocall) or something else; for a listener it may be closing some extra viewer handles and rely on a ```switch``` command to watch all feeds.

As such, I'm closing this issue without merging it. That said, thanks anyway for your contribution and for keeping on providing useful feedback and code!

Let me know if this does indeed work as expected in your scenarios, especially for what concerns the detection of slow links.

PR #177, opened 2015-03-09
Simplify code in the SIP plugin

saghul:
* Use Sofia-SIP's facilities for parsing SIP URIs
* Delete duplicated code in register command

PS: Testing locally I did not find issues with this code, if you do please let me know and I'll fix them.


lminiero:
The code does look much cleaner and lightweight, thanks for this. I wasn't aware of the Sofia helpers for parsing URIs, that definitely make this much easier and cleaner to handle in the code. I'll merge right away and test it later myself, although I don't think we'll encounter any issue.

PR #178, opened 2015-03-09
sip: Make the SIP proxy optional

saghul:
When not specified Sofia-SIP will conduct the standard (RFC3263) lookup
mechanism.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -935,26 +935,25 @@ static void *janus_sip_handler(void *data) {
 			}
 			/* Parse address */
 			json_t *proxy = json_object_get(root, "proxy");
-			if(!proxy) {
-				JANUS_LOG(LOG_ERR, "Missing element (proxy)\n");
-				error_code = JANUS_SIP_ERROR_MISSING_ELEMENT;
-				g_snprintf(error_cause, 512, "Missing element (proxy)");
-				goto error;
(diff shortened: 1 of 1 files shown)


lminiero:
While I agree that the proxy can be made optional to allow for the scenario you envisaged, not all deployments implement this in their DNS servers. Besides, removing the check in the demo application may confuse users that are not aware of this, which would get them a non-working setup. As such, I'd either keep the check we had in the demo page, forcing users to provide a proxy (developers can always skip it in their implementations), or display a bootbox alert explaining this in order to allow users to choose (e.g., "You're not providing a proxy, this will cause/expect this, proceed?").

I can take care of this if you restore the JavaScript code to what it was before.

saghul:
@lminiero fair enough :-) I removed the JS part.

lminiero:
Thanks! I'll merge the patch right away.
At the moment I can't work on the dialog in JavaScript, but plan to do so as soon as I get back to the lab tomorrow in the morning.

saghul:
> Thanks! I'll merge the patch right away.
> At the moment I can't work on the dialog in JavaScript, but plan to do so as soon as I get back to the lab tomorrow in the morning.

That would be great!

lminiero:
Just modified the web page as we discussed, let me know if that works for you.

saghul:
Works beautifully, thanks!

PR #180, opened 2015-03-09
keepalive event payload not expected to be allocated

ploxiln:
... so it was not freed.
so don't actually allocate it (just use rodata string literal)


lminiero:
Good catch, the payload was not freed because ```allocated``` was set to 0.

PR #182, opened 2015-03-10
sip: Add register_ttl configuration option

saghul:
It can also be overridden on the 'register' command.


lminiero:
Merging, thanks ;-)

PR #183, opened 2015-03-10
streaming plugin: fix line accidentally remoted in memory leak cleanup

ploxiln:
line accidentally removed in commit 011610cd4450


lminiero:
Ops, good catch, thanks.

PR #184, opened 2015-03-11
sip: Add setting for local IP address

saghul:
This does away with the c_address setting, effectively replacing it. Now the local_ip (which is either set or autodetected) is used for SIP signaling and media.

I also simplified the local IP detection code, it now takes the IP of the default (IPv4) route, instead of the IP of the first interface. While one could think both approaches are "correct", this one is simpler, allows us to remove the autodetect_ignore setting and it's also portable.

### DIFF
--- a/conf/janus.plugin.sip.cfg.sample
+++ b/conf/janus.plugin.sip.cfg.sample
@@ -1,17 +1,19 @@
 
 [general]
 ; Specify which local IP address to use. If not set it will be automatically
-; guessed from the system, ignoring the interfaces in autodetect_ignore
+; guessed from the system
 ; local_ip = 1.2.3.4
-autodetect_ignore = vmnet
+
 ; Enable local keep-alives to keep the registration open. Keep-alives are
 ; sent in the form of OPTIONS requests, at the given interval inseconds.
 ; (0 to disable)
 keepalive_interval = 120
+
 ; Indicate if the server is behind NAT. If so, the server will use STUN
 ; to guess its own public IP address and use it in the Contat header of
 ; outgoing requests
 behind_nat = no
+
 ; User-Agent string to be used
 ; user_agent = Cool WebRTC Gateway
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks, this looks like a nice improvement, something that we might envisage in the core as well as there's a very similar method there.

This also removes the "ignore interface" option, though, which was there because someone cared about it. Even with your enhanced way of autodetecting what to use, someone may want to explicitly rule out some of the available interfaces even though they might be usable, e.g., because they're for management purpose only or other reasons. Could you extend the patch to bring that back in?

saghul:
The problem is that because of how autodetecting works now there is nothing to ignore, since we don't iterate the list.

By looking at the patch which introduced that feature (the ignore list), I suspect it was taken as-is from the ICE code. It makes sense there, because you don't want to create candidates for those interfaces, but here there is no way that interface is the default route, so IMHO it's not needed.

lminiero:
Mh, I guess it makes sense. Merging then, if needed we'll revisit this later!

saghul:
Great, thanks! I'll take a look at the core / ICE code, I think it can be simplified a bit while keeping the ignore list in that case, because it does make sense there.

ploxiln:
I can mention a case where actually including an interface which isn't used by the default route might be needed: for a VM which has a NAT interface for accessing the internet, which is emulated by the hypervisor without creating a corresponding interface on the host, as well as another interface which is on a private subnet and connected to a virtual interface on the host which is also on the subnet.

In this case the default route goes out the NAT interface, but arbitrary applications on the host can't send packets to that interface. However, they can send packets to shared private subnet interface.

saghul:
@ploxiln I don't think I fully understand the scenario, but I'd say it's a corner case. IMHO, if you're running a somewhat exotic scenario you can set the interface to the one necessary manually.

PR #186, opened 2015-03-12
Replace getifaddrs for getting local IP address in core

saghul:
Interface enumeration is just used for ICE candidates now. Also some small fixes / enhancements :-)

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1840,6 +1840,12 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 		for(ifa = ifaddr, n = 0; ifa != NULL; ifa = ifa->ifa_next, n++) {
 			if(ifa->ifa_addr == NULL)
 				continue;
+			/* Skip interfaces which are not up and running */
+			if (!((ifa->ifa_flags & IFF_UP) && (ifa->ifa_flags & IFF_RUNNING)))
+				continue;
+			/* Skip loopback interfaces */
+			if (ifa->ifa_flags & IFF_LOOPBACK)
+				continue;
 			family = ifa->ifa_addr->sa_family;
 			if(family != AF_INET && family != AF_INET6)
 				continue;
@@ -1856,9 +1862,8 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 				JANUS_LOG(LOG_ERR, "[%"SCNu64"] getnameinfo() failed: %s\n", handle->handle_id, gai_strerror(s));
 				continue;
 			}
-			/* Skip localhost and 0.0.0.0 */
-			if(!strcmp(host, "127.0.0.1") || !strcmp(host, "0.0.0.0")
-					|| !strcmp(host, "::1") || !strcmp(host, "::") || strchr(host, '%'))
+			/* Skip 0.0.0.0, :: and local scoped addresses  */
+			if(!strcmp(host, "0.0.0.0") || !strcmp(host, "::") || strchr(host, '%'))
 				continue;
 			/* Check if this IP address is in the ignore list, now */
 			if(janus_ice_is_ignored(host))
--- a/ice.c
+++ b/ice.c
@@ -1864,7 +1864,7 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 				continue;
 			}
 			/* Skip 0.0.0.0, :: and local scoped addresses  */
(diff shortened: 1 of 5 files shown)


ice.c
```diff
@@ -1856,9 +1863,8 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 				JANUS_LOG(LOG_ERR, "[%"SCNu64"] getnameinfo() failed: %s\n", handle->handle_id, gai_strerror(s));
 				continue;
 			}
-			/* Skip localhost and 0.0.0.0 */
-			if(!strcmp(host, "127.0.0.1") || !strcmp(host, "0.0.0.0")
-					|| !strcmp(host, "::1") || !strcmp(host, "::") || strchr(host, '%'))
+			/* Skip 0.0.0.0, :: and local scoped addresses  */
+			if(!strcmp(host, "0.0.0.0") || !strcmp(host, "::") || strchr(host, '%'))
```
ploxiln: That doesn't end up true for `::1`; you probably want to use `!strncmp(host, "::", 2)`

I think the `strchr(host, '%')` is attempting to match ipv6 automatic non-routable subnet addresses; it may be better to do `!strncmp(host, "fe80:", 5)` for that.

saghul: > That doesn't end up true for ::1; you probably want to use !strncmp(host, "::", 2)
> 

That should be taken cared of above, see the check for IFF_LOOPBACK.

> I think the strchr(host, '%') is attempting to match ipv6 automatic non-routable subnet addresses; it may be better to do !strncmp(host, "fe80:", 5) for that.

Both should work, yes, but the fe80 test look smore explicit, I'll add it.

janus.c
```diff
@@ -3925,60 +3955,38 @@ gint main(int argc, char *argv[])
...
+			JANUS_LOG(LOG_WARN, "Invalid local IP specified: %s, guessing the default...\n", item->value);
 		} else {
-			JANUS_LOG(LOG_WARN, "Couldn't find any address! using 127.0.0.1 as local IP... (which is NOT going to work out of your machine)\n");
-			local_ip = g_strdup("127.0.0.1");
+			/* Verify that we can actually bind to that address */
+			struct sockaddr_in addr;
+			int r;
+			int fd = socket(AF_INET, SOCK_DGRAM, 0);
+			if (fd == -1)
+				JANUS_LOG(LOG_WARN, "Error creating test socket, falling back to detecting IP address...\n");
```
ploxiln: it seems a little awkward to let the code plow on when `fd == -1` (though I guess `r < 0` will trigger)

saghul:
@ploxiln nice catches! Force pushed, can you re-check?

ploxiln:
The fixups to those bits look good, thanks.

I won't yet claim to have reviewed the entire changeset :)

saghul:
@ploxiln thanks anyway! :-)

saghul:
@lminiero this PR got a little out of hand since some commits depend on others,but it's (hopefully!) for the greater good! :-)

lminiero:
@saghul thanks as usual for the tremendous work, and thanks to @ploxiln has well for the great feedback! That said, since this affects the core I'll need some time to evaluate the changes before merging them.

Not sure what you mean by IPv6 addresses in the SDP, though, since IPv6 was already supported by providing IPv6 candidates and this works for media as it is. Do you mean as in c-lines? If so, we need to change the way we merge the SDP right now, as it's always an IN4 that we add there, and I guess the same applies to the SDP mangling in the SIP plugin as well. It doesn't matter when parsing as we ignore c-lines anyway, only candidates matter.

saghul:
> That said, since this affects the core I'll need some time to evaluate the changes before merging them.

Sure, take your time!



> Not sure what you mean by IPv6 addresses in the SDP, though, since IPv6 was already supported by providing IPv6 candidates and this works for media as it is. Do you mean as in c-lines? If so, we need to change the way we merge the SDP right now, as it's always an IN4 that we add there, and I guess the same applies to the SDP mangling in the SIP plugin as well. It doesn't matter when parsing as we ignore c-lines anyway, only candidates matter.

Yeah, now that I think of it, I should have been more precise :-S What those patches do is accept an IPv6 address as the "interface" or "local_ip" setting. More changes might be necessary, to make it work, unfortunately I currently cannot test that :-S At least the setting is now correctly parsed and stored, which is a start! :-)

lminiero:
Just went through it and it looks fine, thanks!

saghul:
Fantastic, thanks!

PR #187, opened 2015-03-13
sip: Simplify code for connecting RTP and RTCP sockets

saghul:


lminiero:
Thanks, simpler is better! :-)

PR #188, opened 2015-03-13
janus: Fix max_events query parameter

saghul:
The documentation mentions 'max_events', but the code looked for
'maxev'.


lminiero:
I think it would be better to update the documentation instead of the code, as there may be people out there using this right now and it would stop working for them...

lminiero:
e.g., maxev is what we have in janus.js as well.

lminiero:
Fixed the documentation, thanks for spotting this!

PR #189, opened 2015-03-13
sctp: Fix segmentation fault

uxmaster:
after destroying SCTP association.

### DIFF
--- a/sctp.c
+++ b/sctp.c
@@ -1305,11 +1305,11 @@ void *janus_sctp_thread(void *data) {
 					JANUS_LOG(LOG_WARN, "Couldn't relay outgoing SCTP data to the monitor...\n");
 #endif
 				/* Encapsulate this data in DTLS and send it */
-				janus_mutex_unlock(&sctp->mutex);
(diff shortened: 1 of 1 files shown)


lminiero:
Hi, not sure what moving the unlock/lock achieves? Why do you think this fixes the segfault you're experiencing?

lminiero:
Just noticed that the unlock/lock around ```janus_dtls_send_sctp_data``` is unneeded. I guess it was there because of how the feature worked in early revisions, and where it could lead to deadlock for trying to lock a mutex already locked. That's no longer the case with the current code, so I'll remove them.

uxmaster:
Actually, I'm not so sure. I have experienced segfault just after
```
[152834908] Destroying SCTP association
[152834908] Data from SCTP to DTLS stack: 20 bytes
[152834908] Data from SCTP to DTLS stack: 20 bytes
[152834908] WebRTC resources freed
```
`janus_dtls_send_sctp_data` uses `sctp->dtls`, thus without the lock `janus_sctp_association_destroy` could free it in the meantime...

Now, after your commit, it is locked, so it should be OK.

lminiero:
Ok, let me know if that's not the case.

PR #199, opened 2015-04-01
exit cleanly on SIGTERM

ploxiln:
... in addition to SIGINT which is already handled. SIGTERM is the signal typically used by docker and other service/process managers.


lminiero:
Good point, thanks, merging!

PR #200, opened 2015-04-03
update README.md

xenyou:
Update README.md to change the install path of libwebsockets.
This fixes not found error when executing janus server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ HTTP REST API, you'll have to install it manually:
 	cd libwebsockets
 	mkdir build
 	cd build
-	cmake ..
+	cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr ..
 	make && sudo make install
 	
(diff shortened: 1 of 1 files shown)


lminiero:
I wasn't aware of any issue there: libwebsockets is already installed in /usr/lib64 on my machine, so I assumed cmake would assume /usr/ as the prefix by default. If you can confirm that's not always the case, and that you had indeed experienced issues with this, I'll merge right away.

xenyou:
It's installed /usr/local/lib/ on my machines without this.

lminiero:
Ok, merging then!

ploxiln:
If you install the library manually (such as following these instructions) instead of with the OS's package manager, it probably *should* go to `/usr/local` or similar, and not to the system `/usr/lib` or similar.

You may just need to start janus with `LD_LIBRARY_PATH=/usr/local/lib` in the env.

PR #203, opened 2015-04-06
logging level changes

ploxiln:
Consider this not a request for merge, but a request for comment, since some of these changes are a matter of taste. For example, I often prefer for a cluster of log lines to be just one line. And this is cosmetic, so not urgent; I can easily maintain this as my own often-rebased branch for a while.

I think these changes make `--debug-level=4` and `--debug-level=5` more useful and less noisy.


ploxiln:
I'm going to break this up into smaller pieces, most palatable first

PR #204, opened 2015-04-08
make JANUS_LOG macro less redundant

ploxiln:
This revised version should have the exact same output as the old version. But it's shorter, and IMHO simpler.

Notice that the old version prepared the `janusstime` time string even if it didn't end up logging for that level (if debug-timestamps were enabled).

### DIFF
--- a/debug.h
+++ b/debug.h
@@ -74,38 +74,26 @@ static const char *log_prefix[] = {
  * or not according to the configuration of the gateway.
  * The format must be a string literal. */
 #define JANUS_LOG(level, format, ...) \
-{ \
-	if (!log_timestamps) { \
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks, I'll look into that ASAP to make sure it works for me too.

ploxiln:
oh, two more little notes:

the `do { } while(0)` should make it more compatible with being used anywhere

Now that the time format is explicit, it could be easily changed to something that's more concise, perhaps better for logging: `"%Y-%m-%d %T"` (e.g. `2015-04-11 14:15:07`)

lminiero:
Yes a configurable (via settings) string for the date/time format is something I had considered. I'll probably add that sometime soon.

lminiero:
Ops, forgot this! Merging now.

PR #217, opened 2015-05-03
Add RTSP support to the streaming plugin using libcurl

mpromonet:
As libcurl is already present for other purpose and provide a basic support of RTSP, I build this pull request in order to ask to an RTSP server the streams to relay.
Basically it use libcurl to get RTP information and next use the RTP streaming code of the streaming plugin.

It could be configured through the janusjanus.plugin.streaming.cfg like this : 

      [h264-rapicam-rtsp]
      type = rtsp
      id = 99
      description = RTSP Test
      audio = no
      video = yes
      url=rtsp://127.0.0.1:8554/unicast


lminiero:
Thanks for taking the time to work on this! I didn't know libcurl had an RTSP stack bundled in. Does it only support RTSP over TCP, or over UDP as well? As to the code, it looks clean to me, I'll inspect it in better detail as soon as I can. For now I'd only ask you to rename a couple of methods:
* ```janus_streaming_callback``` to something like ```janus_streaming_curl_callback``` or ```janus_streaming_rtsp_curl_callback```, to make it clear the method belongs to the RTSP context of the plugin;
* same for ```janus_parse_sdp```, e.g., ```janus_streaming_rtsp_sdp_parse```; we already have a ```janus_sdp_parse``` in the core, and the automatically generated doxygen code could confuse readers.

That said, please beware that we can't accept contributions unless our CLA is signed, so please do so if you didn't already: https://cla.conf.meetecho.com/janus/

mpromonet:
I signed the CLA and rename the method as you suggest.
libcurl just implement the RTSP exchange, not the RTSP parsing (it should be nice). 
The submitted modification parse the RTSP, choose the UDP endpoint for receiving RTP datas and send RTSP SETUP to give this port to the server.
One big advantage for janus-gateway is its lightweight that makes easy to run on a platform like a raspberry, this is different with other approach that transcode the H264 stream to VP8.

lminiero:
Thanks for signing!

Yep, I thought libcurl only implemented signalling, I was wondering whether it implemented support for RTSP itself over UDP as well but I don't think that really matters. Parsing of the SDP could be handled via Sofia, since it already is a dependency for the Janus core and so could be used in this plugin as well, but I think your approach makes things simpler, as after all we're just interested in codec-related stuff anyway. About transcoding, it's worth pointing out that, if the RTSP source is H.264, then transcoding might not be needed but the stream would only be available to Firefox right now.

I have a couple of concerns I wanted to ask you about though:

1. I'd add a limit to the format string of the ```sscanf``` calls in ```janus_streaming_rtsp_parse_sdp``` that is consistent with the sizes of the target strings; while it's really unlikely that any of them will exceed the size of the helping char arrays you pass to it, it still is a potential cause of overflow.
2. I noticed that you added a new "rtsp" type for mountpoints, which is fine; when generating the mountpoint out of the RTSP communication, though, it is created as a "live rtp" one. Again, this in general is fine, as it allowed you to re-use, for instance, the broadcasting thread and so on, but I see an issue when the mountpoint needs to be destroyed. In fact, I'd expect a mountpoint destroy in this case to result in an RTSP TEARDOWN request sent to the RTSP server, but I couldn't find any such call in the code. This way, the mountpoint would indeed be destroyed from the plugin perspective, but the RTSP server would keep on sending the media packets to the previous destination, which would be discarded. Is this actually the case, or did it happen differently in your tests? I guess that, to keep the "live rtp" re-use, this could be solved by saving the RTSP url in the related structure, so that in case of a destroy a TEARDOWN is sent if and only if the url is there (which means that what look like a "live rtp" mountpoint was actually originated by a "rtsp" create). Does this make sense?

Thanks again for working on this!

mpromonet:
You are right about sscanf and for TEARDOWN also (usually RTSP server stop to send RTP data when communication is close by the curl_easy_cleanup). I will look if Sofia can handle the SDP parsing.
The plugin code is difficult to read due to many code duplication, I will rework later.

lminiero:
Ah, I didn't notice the curl_easy_cleanup part, sorry about that! Then in that case concern 2. disappears: I guess the libcurl cleanup takes care of a TEARDOWN automatically so no need to do that.

No need to look into Sofia SDP either: for now I think your approach works fine.

As such, don't worry about fixing anything in the code apart from some refinement you might deem necessary. I'll review the code in the next few days and then I'll merge it.

lminiero:
Sorry if this took a while... merging, thanks again!

PR #218, opened 2015-05-05
Trickle error log messages lacked trailing newlines

ploxiln:
I also deleted some code comments that were redundant with the log message; IMHO you can as easily read the log message in the code there.


lminiero:
Good catch, thanks.

PR #219, opened 2015-05-05
quiet logging around webserver requests

ploxiln:
* The "Got a HTTP ? request on ?" message changed to INFO (was VERB), and only printed once per req
  * log messages that had no request-identifying information changed to DBG
  * other messages changed to HUGE
  * post request size/contents logged at HUGE, this might be controversial, let me know if I should tweak back to VERB

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -507,7 +507,7 @@ void janus_session_free(janus_session *session) {
 int janus_ws_client_connect(void *cls, const struct sockaddr *addr, socklen_t addrlen) {
 	struct sockaddr_in *sin = (struct sockaddr_in *)addr;
 	char *ip = inet_ntoa(sin->sin_addr);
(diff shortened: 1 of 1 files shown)


lminiero:
I think the "Got a HTTP ? request on ?" log lines are still too verbose: just think of trickling, for instance, that alone would get tens of lines for each PeerConnection at INFO level, which I think is too much. The same applies to polling or keepalives. The other changes look fine instead, thanks!

ploxiln:
What if I made that "Got a HTTP ..." message VERB ?

lminiero:
Yep, that's what I meant: I think VERB is more apt that INFO for that message, as it can happen quite often.

ploxiln:
updated

lminiero:
Thanks, merging!

PR #220, opened 2015-05-05
convert double trinary-operator to single trinary for event->payload

ploxiln:
see title. that's it :)


lminiero:
Makes sense, thanks!

ploxiln:
(also, uh, it's "ternary", not "trinary", I must have been tired...)

PR #224, opened 2015-05-06
quiet cleaning up session / destroying session log messages

ploxiln:
cleanup always calls destroy, so make cleanup DBG

Also, change destroy from INFO -> VERB level

By the way, all the PRs in this logging-cleanup series are subjective, I won't feel bad if you reject most of them :)


lminiero:
Makes sense, thanks!

PR #225, opened 2015-05-06
log only when starting to wait for webrtc state to change

ploxiln:
instead of every 100 milliseconds


lminiero:
Good point, merging!

PR #227, opened 2015-05-06
quiet logging: ice candidate added, looping ice thread

ploxiln:
quieter logging of final "ice candidate added" message
  * an earlier log message says it is being added
  * if there is some error adding it, that will show

The "Looping (ICE)" message is almost debug-level information, IMHO.

(Again, this is subjective, so I understand if you reject.)

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1432,7 +1432,7 @@ void *janus_ice_thread(void *data) {
 		return NULL;
 	}
 	g_usleep (100000);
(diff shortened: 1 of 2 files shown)


lminiero:
I'm ok with the debug level changes. The looping message was added there, IIRC, when I had to investigate some issues when getting rid of PeerConnections and/or handles, so it's ok to change it to DBG.

PR #228, opened 2015-05-08
log number of recent retransmits once per 5 seconds at INFO level

ploxiln:
only if there were any, of course
logging about individual retransmits is pushed to HUGE level

Let me know what you think...


lminiero:
Not sure about the INFO level, considering that when hundreds of PeerConnections are being handled this could clog the log (no pun intended!). But that also applies to other INFO lines we have at the moment, so I guess it's ok. We can always change that to VERB in case it's too much.

PR #230, opened 2015-05-08
slow_link callback refactor: count NACKs over full second

ploxiln:
... instead of only counting over a compound RTCP packet, to see if the threshold of NACKs is exceeded.

This mostly affects outgoing streams, since peers usually send NACKs to janus one-per-packet under reasonable-loss circumstances (for latency reasons). I would typically see many separate RTCP packets, with 1 to 4 NACKs each, when testing with 2% packet loss.

Janus batches up NACKs to send for incoming streams over 0.5 seconds, so was already much more likely to reach the threshold per-packet.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -971,6 +971,28 @@ void janus_ice_component_free(GHashTable *components, janus_ice_component *compo
 	//~ janus_mutex_unlock(&handle->mutex);
 }
 
+/* call plugin slow_link callback if enough NACKs within a second */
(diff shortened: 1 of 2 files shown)


lminiero:
Seems to make sense, I'll look into this deeper ASAP.

lminiero:
Looks good, merging, thanks!

ploxiln:
Thanks!

PR #232, opened 2015-05-12
only log once when "Still cleaning up from previous media session"

ploxiln:
same type of logging cleanup as in #225 

I also threw in a commit to remove a check for g_strdup() failing. Like most glib2 functions, it will abort the process if it fails.


lminiero:
Merging, thanks!

PR #234, opened 2015-05-14
consider retransmits, instead of received NACKs, for slow_link

ploxiln:
avoids considering a peer slow because of packets janus never sent it
(... which is because another peer, the original source, is slow)

(still consider *sent* NACKs for slow_link)

also don't unlock the component mutex until after updating the component stats


lminiero:
Good point! Thanks.

PR #235, opened 2015-05-15
avoid starting more requests while janus is stopping

ploxiln:
use usleep() instead of g_usleep() so the stop signal cancels the sleep
(g_usleep() calls nanosleep() in a loop) (usleep() is available on both osx 

use MHD_quiesce_daemon() to stop accepting new requests immediately,
then wait a bit for current requests to finish

The inspiration for this, was seeing a flurry of about 10 quick "long-poll" requests for a single session (from a backend service closely situated to janus) during janus shutdown.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -5043,27 +5043,31 @@ gint main(int argc, char *argv[])
 		}
 	}
 
-#ifdef HAVE_WEBSOCKETS
 	while(!g_atomic_int_get(&stop)) {
+		/* Loop until we have to stop */
+#ifdef HAVE_WEBSOCKETS
 		if(wss || swss) {
 			/* libwebsockets needs us to call the service ourselves on a regular basis */
 			if(wss != NULL)
 				libwebsocket_service(wss, 100);
 			if(swss != NULL)
 				libwebsocket_service(swss, 100);
-		} else {
-			/* Loop until we have to stop */
-			g_usleep(250000);
+		} else
+#endif
+		{
+			usleep(250000); /* a signal will cancel usleep() but not g_usleep() */
 		}
 	}
-#else
-	while(!g_atomic_int_get(&stop)) {
-		/* Loop until we have to stop */
-		g_usleep(250000);
-	}
-#endif
 
 	/* Done */
+
+	JANUS_LOG(LOG_INFO, "Closing webserver(s)...\n");
+	if(ws)        MHD_quiesce_daemon(ws);
+	if(sws)       MHD_quiesce_daemon(sws);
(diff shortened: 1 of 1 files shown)


lminiero:
Thinking about this, the usage of ```MHD_quiesce_daemon``` might also solve the occasional issue where the gateway couldn't close until the requests in the browser were closed as well, because the pthread loop in libmicrohttpd was stuck there. Thanks for the improvements, as usual!

lminiero:
I just did some more testing and it actually seems to lock my Janus shutdown systematically... commenting the quiesce lines I can shutdown normally instead. It may be the libmicrohttpd version that I have installed on my development machine (0.9.34), though, as on a different machine with a more recent one (0.9.39) the shutdown goes smoothly instead.

ploxiln:
Sorry about that. This patch has less testing than the others I've been submitting, I only tested it locally a couple of times. Perhaps it should be reverted.

lminiero:
There's probably no need to do that, as on the updated libmicrohttpd it works fine, and my development machine isn't exactly up to date (Fedora 18). I'll check what happens on some of the machines we use normally, just to check if I can find a configure tweak I can use to selectively enable/disable the quiesce stuff if needed.

ploxiln:
I'm still mostly using ubuntu-14.04 which has libmicrohttpd-0.9.33. I didn't experience the problem because the application which usually controls janus over http is not google chrome. If I test with google chrome, which keeps extra connections open (to speed up later requests), I can reproduce this problem pretty easily.

There are some interesting changelog entries about this:

```
Sun Jun 23 15:28:13 CEST 2013
Added option 'MHD_USE_PIPE_FOR_SHUTDOWN' to cleanly support
'MHD_quiesce_daemon' with thread pools and per-connection
threads (we then need a pipe for shutdown, but if
'MHD_quiesce_daemon' is not used, we do not want to
require the use of a pipe; introducing the pipe after
the threads have been started can also fail, so the
application needs to tell us early on). -CG 
```

(that went in before 0.9.28)

```
Sat Apr  4 17:48:13 CEST 2015
Fix issue in thread-pool mode where a MHD_stop_daemon()
might not reach threads that stopped listening because
we hit the maximum number of concurrent connections and
the option MHD_USE_PIPE_FOR_SHUTDOWN was also not used.
Testcase added as well. -CG 
```

(that went in before 0.9.40)

So altogether it looks like this might be a somewhat complicated issue.

It might be best to just keep the g_usleep() -> usleep() change so shutdown starts faster, but take MHD_quiesce_daemon() back out.

lminiero:
Yep, makes sense, I'll comment them right away.

PR #236, opened 2015-05-15
in_stats and out_stats: add total new nacks

ploxiln:
instead of just incrementing by one for any number of nacks in a RTCP packet

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1293,7 +1293,8 @@ void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint component_i
 							seqs = seqs->next;
 						}
 					}
-					if(nacks != NULL) {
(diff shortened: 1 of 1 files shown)


lminiero:
I'm wondering whether we should actually have two separate counters: one that explains how many times NACKs were sent, no matter how muchin a bunch (what we have now), and another that counts the actual number of packets hat were NACKed. My feeling is that they're both useful for different reasons, rather than just sticking to one. Maybe even a count of NACKs in the last second, if it makes sense at all to have? I'll work on something like this when I get back to the lab.

ploxiln:
OK, but take a quick look at #238; this nack-counting logic might make more sense with that change to NACK generation, where we won't potentially send many nack packets with the same sequence numbers, if the stream cuts off completely for a while.

PR #238, opened 2015-05-16
re-write NACK generation from missing rtp sequence numbers

ploxiln:
send nacks for missing seq nums faster: don't wait up to half a second

keep track of nacks sent per seq num, at most nack twice

This uses a custom doubly-linked-list implementation. In my first efforts I tried to add nodes in the middle of the list, which wasn't (efficiently) possible with the glib2 list types. With the current state, I could use GList, but the custom implementation requires half the allocations (because nodes/links are embedded rather than external to the data structs).

I've been running this for two weeks now, and it seems to work OK. But I don't expect this to be immediately merged. I just want to propose this strategy and show an example.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1293,7 +1293,8 @@ void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint component_i
 							seqs = seqs->next;
 						}
 					}
-					if(nacks != NULL) {
+					guint nacks_count = g_slist_length(nacks);
+					if(nacks_count) {
 						/* FIXME Generate a NACK and send it */
 						if(now-component->last_nack_time > 2*G_USEC_PER_SEC) {
 							JANUS_LOG(LOG_VERB, "[%"SCNu64"] Missed some packets, NACKing them now...\n", handle->handle_id);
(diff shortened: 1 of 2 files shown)


lminiero:
Tremendous work, thanks for this!
Small nit: s/Recieved/Received

The half-second interval for NACKs was introduced to avoid the "NACK storm" issue, considering we could get a NACK again for a packet we just retransmitted and as such retransmit again. Add state to NACKs as your patch does should indeed fix that.

As a side note, I agree that the custom linked list is lighter in terms of allocation, but does this mean that you'd see a similar approach being used in other usages we have of GList and GSList (at least for data that is not elementary) for the same reason? e.g., Asterisk for instance has macros to add links to structs and turn them to lists you can manage through other macros, so something like that could be done here as well. Just asking not only for potential optimizations but also for consistency in the code.

That said, I want to study the code in a bit more detail before merging it, but especially considering your extensive testing I don't see issues in doing that ASAP.

lminiero:
PS: does this include #236? Looks like it but wanted to be sure.

ploxiln:
This is indeed based on #236 just for my convenience.

I'll revisit the linked-list implementation. GList may be just fine, even if it is twice the allocations. Two existing macro-based implementations that I'm familiar with are openbsd/freebsd [queue.h](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/sys/queue.h?rev=1.38&content-type=text/x-cvsweb-markup) and [utlist](https://troydhanson.github.io/uthash/utlist.html) - to use either, one typically just copies the stand-alone header file.

lminiero:
There's no need to revisit the linked-list stuff, don't worry! I was just thinking out loud, and sharing with you some thoughts about whether or not it made sense to look into that for future revisions. This is something I'd carry on my shoulders anyway :-)

I've checked the code and everything looks fine to me. If you confirm that you don't have any pending change you'd like to submit, I can merge it when you're ready. I assume that merging this means we can skip #236 which is already included.

ploxiln:
I have no pending changes for this, this is what we use right now. Yes, merging this includes #236 and github should automatically show that as merged when it happens.

I am willing to work on changes which you think should be done, to make it more maintainable within the janus codebase, but if that's not necessary before merge, OK :)

lminiero:
Ok, merging now! I'll probable just rename one of the methods to add the janus_ prefix, so nothing complicated :-)

PR #239, opened 2015-05-18
Example Matrix Integration

dbkr:
Do not merge. Sharing the current Matrix integration as requested on the mailing list.

This version is the unmodified version we used for controlling and relaying video from our drone, ie. it contains drone control code in janus_streaming too which can be ignored. The Matrix integration is all in janus.c / janus.h.

The current matrix integration is also very hacky in that sessions are referenced directly rather than by ID which causes crashes when they are cleaned up, and in several places memory is allocated and never freed. It's also hard-wired to a fixed ID in the streaming plugin.

This just shows how we integrated Matrix as another source alongside HTTP, WebSockets and RabbitMQ.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,7 +12,9 @@ AM_CFLAGS = \
 	-Warray-bounds -Wold-style-definition -Wsign-compare -Wlarger-than=65537
 AM_CFLAGS += -Wno-undef  # sophiasip is not -Wundef--safe
 AM_CFLAGS += -Wno-redundant-decls  # sophiasip also contains redundant declarations
-AM_CFLAGS += -Wno-override-init  # we need this for JANUS_PLUGIN_INIT
+AM_CFLAGS += -Wno-override-init -Wno-initializer-overrides  # we need this for JANUS_PLUGIN_INIT
+# make clang stop whining:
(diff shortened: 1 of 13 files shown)


lminiero:
Hey, just a quick post to tell you I haven't forgotten about this...

I'm in the process of refactoring the way Janus handles transports, something that should help you integrate Matrix in a much easier and transparent way. Basically the idea is to separate transports from the core pretty much as the plugins do now with respect to media management: each transport would be implemented as a different module, implementing a simple API and interacting with the core this way. The core would then handle communication in a transparent way, not knowing/caring where a message came from or is going through. This should make it easy (well, easier) for other protocols to be added in a transparent way, e.g., Matrix, nanomsg, jabber or whatever else.

I only started recently after studying how this could be done for a bit, so this might take a while: I hope I'll have a first working version to share in a week or two, maybe as a pull request people can play with, so I'll keep you posted.

lminiero:
As promised, here it is:

https://github.com/meetecho/janus-gateway/pull/281

This should make a Matrix integration easier, feedback welcome!

lminiero:
Closing as outdated, the ```modular-transports``` branch can be used as a base for future developments in that direction.

PR #244, opened 2015-05-21
Hack in janus.js to fix the success callback for pluginHandle.send()

ancorgs:
See https://groups.google.com/forum/#!topic/meetecho-janus/EBvJet3Upco for details about the possible bug.


lminiero:
I just prepared a patch that should fix that issue in a cleaner way. Let me know if that still doesn't work for you.

PR #246, opened 2015-05-23
janus.js: firefox 38 supports new video constraints format

ploxiln:
This simplifies the constraints a bit, though Firefox still refuses to do anything except 640x480 or 320x240 in my experience.

You may not want to actually merge this, to keep support for older firefox. Luckily, firefox 38 is becoming the new "ESR" release (and you can already download it as such). You may just want to wait 3 months until Firefox 31 ESR is no longer supported.

see also http://stackoverflow.com/questions/28282385/webrtc-firefox-constraints/28911694#28911694

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1080,13 +1080,6 @@ function Janus(gatewayCallbacks) {
 						height = 720;
 						maxHeight = 720;
 						width = 1280;
-						if(navigator.mozGetUserMedia) {
-							// Unless this is Firefox, which doesn't support it
-							Janus.log(media.video + " unsupported, falling back to stdres (Firefox)");
-							height = 480;
-							maxHeight = 480;
-							width  = 640;
-						}
 					} else if(media.video === 'stdres') {
 						// Normal resolution, 4:3
 						height = 480;
@@ -1105,10 +1098,10 @@ function Janus(gatewayCallbacks) {
 					}
 					Janus.log("Adding media constraint " + media.video);
 					if(navigator.mozGetUserMedia) {
+						// http://stackoverflow.com/questions/28282385/webrtc-firefox-constraints/28911694#28911694
 						videoSupport = {
-						    'require': ['height', 'width'],
-						    'height': {'max': maxHeight, 'min': height},
-						    'width':  {'max': width,  'min': width}
+						    'height': {'ideal': height},
+						    'width':  {'ideal': width}
 						};
 					} else {
 						videoSupport = {


lminiero:
Thanks. I guess I'll have to also update the syntax for Chrome, since we're still using the outdated constraints there too.

ploxiln:
I don't think Chrome supports the "standard" "non-deprecated" syntax yet.

lminiero:
I just played with it a bit (both Firefox 38 and Nightly) and I couldn't get it working: if I ask for "stdres-16:9" (640x360) I still get 640x480. Does it work as expected for you instead?

I verified, though, that the limitation we had on higher resolutions has been removed for Firefox, which means getting a 1280x720 works instead, so I'll fix that check.

ploxiln:
I haven't seen 16:9 resolutions actually work in firefox. I've only checked that this new syntax will bring up 640x480 or 320x240, which is the closest thing firefox can do to the requested resolution.

The reason I switched to this is that before I was trying to force the widescreen aspect ration with "maxHeight = minHeight = 360", and in previous versions of firefox it would stretch 640x480 to 640x360, which I figured was good enough for testing purposes until firefox could actually do widescreen. But then in firefox 38 it started failing to capture video at all if it couldn't do the strictly-requested resolution.

I noticed that in janus master you allowed a range of heights for firefox, "height = 360, maxHeight = 480", so it would fall back to 640x480. That works fine. But while I was fixing up my copy of janus.js, I figured why not switch it to the latest spec, which accomplishes the same thing. Not a big deal though :)

lminiero:
Yep, the new spec is what I was playing with, but it looks like it's blatantly ignored anyway :-(
Apparently an ideal height of 360 is still turned into 480...

ploxiln:
The new spec does not offer different resolutions than the old one did. The new spec is just a possibly more convenient (and more future-proof) way to specify/request resolutions. "ideal" means that the resolution would be the best, but other resolutions are OK if the best is not possible.

So it's a way of requesting a particular resolution, but letting firefox fall back to the closest thing it can do, *anything* it can do, without specifying a min and max. So when firefox did not support 720p, "ideal" would allow it to fall back to 640x480, without needing extra logic for that in the javascript. That's all it really does for you.

ploxiln:
looks like this is incorporated in https://github.com/meetecho/janus-gateway/commit/ff257ca1b1af4124030cdd8c41c674c7b2795bbb

(and maybe when firefox < 38 is theoretically all unsupported in a couple months, some of that can be cleaned out)

lminiero:
Yep, I actually forgot I incorporated most of these changes when I committed the updated janus.js: I actually pushed it for some other fixes. Thanks for bringing the issue up though!

PR #248, opened 2015-05-28
Add ability to enable outgoing video on the SIP demo

saghul:


lminiero:
We actually hid the video box on purpose. Most SIP infrastructures don't support VP8 or the RTCP feedback that makes video valuable in WebRTC. This caused, for instance, browsers sending increasing amounts of video data because the other side wouldn't limit this by means of REMB feedback, something they wouldn't understand. As such, the video experience was pretty much broken.

Not sure if things have changed in that respect now (Chrome should now limit itself to ~300kbps if REMB are missing, not sure though), but if that's what's still happening I'd rather keep the video checkbox hidden, as users would try it, fail, and assume it's the plugin that doesn't work. If you can confirm it works as expected in your tests, I can try some more checks on my side too and in case merge your addition.

saghul:
Firefox can do H264 now, and I can confirm that works :-) I'm now testing VP8 (adding support for it in Blink), but my decoder is broken somehow, so for now I do see the video in the browser but not the other way around.

Jitsi also has VP8 support, though I got mixed results.

IMHO, the only way to get people's code fixed is to allow them to test it, so I'd leave the checkbox enabled (though not checked by default), but of course I'd say that :-)

lminiero:
Makes sense... besides, we do have a disclaimer that video may be broken in the demo page text, so users can't say they haven't been warned! :-)

saghul:
Thanks! :+1:

PR #249, opened 2015-05-29
MQTT protocol support

khejing:
The MQTT protocol(http://mqtt.org/) support with Paho MQTT C client(https://www.eclipse.org/paho/clients/c/) for Janus has been added in this pr. Because AMQP can't be used in browser, and I need integrate Janus with AcitveMQ, not just RabbitMQ. First I've tried using MQTT and AMQP with Camel, but soon I found that I need create MQTT topic dynamically, and I don't know how to do it with Camel. And now just subscribe or publish to some topic, if it doesn't exist, it will be created automatically by MQTT server.

If there are any style problem or something I shouldn't modify, just tell me, and I'll fix them. And I've already test it with my own usage context, if any bug found, I'll fix it as soon as I can.

PS: I don't know how to reset commit with git, so my commit history is crappy, just ignore them, and see the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ Makefile.in
 
 .deps
 .dirstamp
+
+.project
+.cproject
--- a/conf/janus.cfg.sample.in
+++ b/conf/janus.cfg.sample.in
@@ -64,6 +64,28 @@ to_janus = to-janus			; Name of the queue for incoming messages
 from_janus = from-janus		; Name of the queue for outgoing messages
 
 
+; Recent versions of Janus now also support MQTT based messaging as
+; an alternative "transport" for API requests, responses and notifications.
+; This is only useful when you're wrapping Janus requests in your server
+; application, and handling the communication with clients your own way.
+; MQTT topic can be created on the fly easily, just publish or subscribe it,
+; then it will be created on MQTT server. So Janus can implement multiple
+; topics to handle multiple concurrent "application servers" taking advantage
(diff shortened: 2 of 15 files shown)


lminiero:
Wow impressive feat, thanks for the effort! I'm not familiar with MQTT (actually this is the first time I've heard of it) but looking at the code I guess it's not very dissimilar from RabbitMQ, am I right?

Anyway, I'm afraid there's too much stuff here for a single pull request. In fact, there's not only the support for the new protocol (which I looked into and seems compliant in terms of code style with what we have), but also some typo/errors fixes (e.g., where we use ```stun_server_arg``` for RabbitMQ stuff), methods that are not strictly related to the new feature that have a new signature, etc. From the quick glance I gave, they mostly seem acceptable solutions, but I simply cannot integrate them all in a single pull request that is related to something else. It would be much easier and cleaner if you could separate those parts in order to produce separate pull requests (e.g., one for the fixes, one for the new signature, one for the new protocol, etc.), so that we can incorporate the "simple" parts right away, and then discuss the others in more detail. Is this something you could take care of?

That said, please beware that we need all contributors to sign our CLA, before we can accept patches: you can find it here https://cla.conf.meetecho.com/janus/

khejing:
I've already sign the CLA.

Yes, MQTT is some like AMQP, but it's lightweight, and be used in mobile device more. I don't know PR should be small before. OK, I'll separate them, and send them again. But it will take some time, and I'll do it ASAP

Issue #251, opened 2015-06-04
NSS 3.19.1 disallows RSA keys < 1023 bits


roysjosh:
The example cert and key used for DTLS is no longer valid. No one should use it outside of testing, but for newcomers it will add unnecessary confusion. See the NSS release notes at https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/NSS_3.19.1_release_notes

lminiero:
Thanks for the heads up, we'll replace the sample certificates we have (which are outrageously old anyway).

saghul:
For reference, this is what happens when using Firefox Developer Edition (40.0a2):

~~~~
[ERR] [dtls.c:janus_dtls_srtp_incoming_msg:312] [3763557713] Handshake error: error:1410242F:SSL routines:DTLS1_READ_BYTES:tlsv1 alert insufficient security
~~~~

lminiero:
@saghul I just tried with both Firefox Nightly and Developer Edition to start the Echo Test on the online demos, and I didn't get the error you experienced... were you using our sample certificates or some other 512 bit ones?

saghul:
@lminiero I was using the sample certs and the SIP gateway demo.

lminiero:
Just tried that as well (janus.conf.meetecho.com/siptest.html) and still nothing, it seems to work :S 
Anyway, if you're getting that it means we probably need to update the certificates anyway: what worries me is that since I cannot replicate the issue, I may end up providing new sample certificates that don't work for anybody else. I'll let you guys know when I push them, so that you can try those yourselves.

saghul:
Weird. At any rate, I can test the new certs once you have them if you put them in a branch, for example.

lminiero:
Makes sense, thanks.

saghul:
I guess this can be closed now :-)

lminiero:
Yep. ;-)

PR #253, opened 2015-06-04
Created 1024 bits certificate (see #251), and added small documentati…

lminiero:
@saghul can you check whether these certificate files fix it for you?


saghul:
Works for me, thanks!

lminiero:
Ok, thanks, merging then!

PR #254, opened 2015-06-05
Implemented new OpenSSL BIO filter to fix fragmentation issue in DTLS…

lminiero:
Thanks to the help and feedback from the OpenSSL and libre guys, I've implemented a new OpenSSL BIO filter that should fix the issue of fragmentation when large certificates are used.

Feedback and testing welcome... if this does indeed work as expected and doesn't break anything else in the process I'll merge it.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -77,6 +77,8 @@ janus_SOURCES = \
 	debug.h \
 	dtls.c \
 	dtls.h \
+	dtls-bio.c \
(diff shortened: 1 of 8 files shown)


saghul:
I just tried with a 4096 bit key / cert and it works brilliantly. Great work! :+1:

lminiero:
Ok, merging it then!

PR #255, opened 2015-06-08
rtp_listener feature added for videoroom plugin

Computician:
This adds a new feature to the videoroom plugin.

Publishers in a specific room can have an rtp_listen request made against them which will cause the publisher to forward the correct media give the specific port to the given host. 

```
{
   "request" : "rtp_listen",
   "publisher_id": 1234,
   "room" : 111,
   "video_port": 5001,
   "audio_port": 5000,
   "host" : "192.168.0.123"
}
```
response
```
{
   "videoroom" : "rtp_listen",
   "publisher_id" : 1234,
   "room" : 111,
   "rtp_stream" :
   {
      "stream_id" : <unique_id>,
      "audio" : 5000,
      "video": 5001,
      "host": "192.168.0.123"
    }
}
```

A stream can also be told to stop forward rtp packets through a `stop_rtp_listen` request which references the publisher_id, room number, and stream_id.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -122,6 +122,8 @@ rec_dir = <folder where recordings should be stored, when enabled>
 #include "../rtcp.h"
 #include "../record.h"
 #include "../utils.h"
+#include <sys/types.h>
+#include <sys/socket.h>
 
 
 /* Plugin information */
@@ -264,6 +266,13 @@ static GHashTable *sessions;
 static GList *old_sessions;
 static janus_mutex sessions_mutex;
 
+//a host whose ports gets streamed rtp packets of the corresponding type.
+typedef struct rtp_listener {
+	int audio_port;
+	int video_port;
+	gchar* host;
+} rtp_listener;
+
 typedef struct janus_videoroom_participant {
 	janus_videoroom_session *session;
 	janus_videoroom *room;	/* Room */
@@ -286,9 +295,12 @@ typedef struct janus_videoroom_participant {
 	janus_recorder *vrc;	/* The Janus recorder instance for this publisher's video, if enabled */
 	GSList *listeners;
 	janus_mutex listeners_mutex;
+	GHashTable *rtp_listeners;
+	janus_mutex rtp_listeners_mutex;
+    int udp_sock; /* Socket over which rtp streams can be forwarded */
 } janus_videoroom_participant;
 static void janus_videoroom_participant_free(janus_videoroom_participant *p);
-
+static void janus_rtp_listener_free_helper(gpointer data);
 typedef struct janus_videoroom_listener_context {
 	/* Needed to fix seq and ts in case of publisher switching */
 	uint32_t a_last_ssrc, a_last_ts, a_base_ts, a_base_ts_prev,
@@ -404,6 +416,18 @@ static void session_free(gpointer data) {
 		session->handle = NULL;
 		g_free(session);
 		session = NULL;
+	}
+}
+
+static void janus_rtp_listener_free_helper(gpointer data) {
+	if(data) {
+		rtp_listener* listener = (rtp_listener*)data;
+		if(listener) {
+			free(listener->host);
+			listener->host = NULL;
+			free(listener);
+			listener = NULL;
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for working on this!

A couple of quick notes form the first look I gave at the code:

* This currently requires both audio and video ports to be passed in a ```rtp_listen``` request. This doesn't seem necessary, considering there may be scenarios where the videoroom plugin is only used to route video or audio (e.g., you handle video with the videoroom, and audio with the audiobridge, or you're doing screensharing).
* It's a good idea to protect the feature behind a secret to avoid abuses, but I'm wondering if we may involve a different one than the room secret for that. Anyway, this is a broader discussion that relates to roles, so please disregard this if not as a thought I've given voice. I'll probably think about something like this myself in the future.
* The destination address for each RTP listener is computed for every packet (inet_pton + htons for the port), wouldn't it be better to compute it just once when the listener is created, and make the address part of the struct, since you cannot modify an existing RTP listener but just get rid of it when it has been created?
* I'm also wondering whether we might want to add a check for the first key frame before forwarding media: some tools don't digest media well when they start receiving stuff in the middle, and so waiting for the first key frame before forwarding media to them might help (although packet loss could cause the same issue anyway). That said, this is something I can add myself later, maybe as an optional setting s that we kee both options open.
* I've seen a few lines with weird indentation and C++ style commenting (// instead of /* */), could you fix those? This is just me being overly finicky, please indulge me :blush: 

To conclude, amazing effort, thanks! This will be very helpful also in the prospect of implementing large scale conferences or webinars, as you can just, for instance, forward a publisher to streaming mountpoints in other Janus instances and allow for many more viewers that way.

Computician:
I was thinking about changing the listener to be

```
typedef struct rtp_listener {
   int port;
   int is_video;
   int socket;
   gchar* host;
   struct sockaddr_in serv_addr;
} rtp_listener;
```


Or would it be better to still keep just one socket per participant? I am not sure about overloading the socket and inadvertently causing unneeded packet loss.

lminiero:
There's shouldn't be any issue in re-using the same socket for both media, I thought that was a clever way of handling that: that's what WebRTC does anyway when bundling (which is the default).

By the way, why not call the feature ```rtp_forwarder``` or something like this? ```rtp_listener``` may mislead the users into thinking the videoroom would listen on those ports for receiving RTP, as the streaming plugin does, while instead it's the other way around.

Computician:
Alright, I will keep the single socket and changing the name is not that big of a deal.

Computician:
Actually, the struct will end up being:
```
{
int is_video;
struct sockaddr_in serv_addr;
}
```
There is no need to store the port and host separately as the sockaddr_in structure already keeps track of them.

lminiero:
It may be worth keeping those anyway just for auditing, e.g., when using the query method to get info on a publisher (which might also return the list of RTP destinations, if useful), although an alternative is resolving them from the address when needed.

Computician:
Getting port and host information from the struct is fairly easy. I can add them back but I am not sure it is really needed.

Computician:
As for the Key frame issue, it may be beneficial to either add a functionality for an rtp_forwarder to request a keyframe or for a key frame to be sent ASAP after a new video stream forward is requested.

Computician:
I have pushed an update. Let me know how it looks.

lminiero:
Why did you replace Opus with A-law? Was this a mistake?

Computician:
Oops! HAHA! That was for my testing

Computician:
Changed it back...my bad :(

lminiero:
No problem, thought that wasn't intentional!

The update looks good, I'm just wondering whether it would be more optimized if there were two different tables for audio and video forwarders: the way it is now, you iterate on them all, and then only use the ones related to the media you just received, which can result in a lot of skipping. Anyway, don't worry about this, I'm probably just thinking ahead and I can change this along the way.

I'll give a deeper look at the code to test it ASAP, and then I'll merge it, thanks again!

lminiero:
Sorry if this took a while, just tested and it looks fine! I'll probably make some of the changes I anticipated in the previous comments in the next few days. I'll post them as pull requests so that you can revisit those as well.

Thanks!

PR #260, opened 2015-06-15
fixed double free in echotest handler

mtdxc:
echo plugin. When client detach request send twice for the same handler, janus-gateway crashed. 
It's safer to fixed it on server code.


lminiero:
Thanks for spotting this, I'll check if this is an issue in other plugins as well. I wonder if we should actually look at the value of ```destroyed``` instead...

That said, have you signed our CLA? We can't accept contributions without it:

lminiero:
I've modified all the plugins myself using the mutex-protected ```destroyed``` check I anticipated. This should fix the issue you experienced. Let me know if that's not the case. Closing the pull request as this was fixed in a different way. Thanks for your input!

mtdxc:
sorry about this,  I has worked on webrtc iOS client last mouth, and don't notice your reply. 
you are right, I will merge your code .

PR #262, opened 2015-06-16
fix a problem in wav header

khejing:
see http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html, the nAvgBytesPerSec fileld should be (sample rate * sample format / 8), here is audiobridge->sampling_rate * 2

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -2374,7 +2374,7 @@ static void *janus_audiobridge_mixer_thread(void *data) {
 				1,
 				1,
 				audiobridge->sampling_rate,
-				audiobridge->sampling_rate,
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for spotting this. Did this actually cause any issue, e.g., trouble with playout? I never had any problem in replaying the recordings of the audiobridge.

khejing:
So far as I see, this will make player software estimate the duration of the recording file twice long as the actual duration. I have tested recording a audio of 14 second, when I play it, the player say it's  28 second, and when the progress is 50% in the progress bar, the playing of the audio is over unexpectedly.

lminiero:
Ok, then your patch definitely makes sense. Merging, thanks!

PR #264, opened 2015-06-16
janus: reject incoming WS connections if sub-protocol is not set

saghul:
Currently a connection made without any sub-protocol specified is left lingering.


lminiero:
Good catch, thanks!

PR #265, opened 2015-06-19
echo: return error if unrecognizable message is received

saghul:
Right now the following message just returns a OK event:

~~~~
{"body": {"booooo": 123}, "janus": "message", "handle_id": 1818527068, "session_id": 2418339675, "transaction": "3bcad0e3e4ed44c1936b2f61445ad631"}
~~~~


lminiero:
Rather than "unrecognizable message" I'd say something like "Missing any supported attribute (audio, video, bitrate)" or similar, as if I send ```{ audio: true, foo: "bar" }``` the request succeeds despite the foo thing there. The problem happens when audio, video and bitrate are all missing. What do you think?

saghul:
Right, updated!

lminiero:
Looks good, thanks!

PR #266, opened 2015-06-20
Add multicast support to RTSP streaming plugins 

mpromonet:
Hi,
This pull request mainly brings support for multicast streams negociated using RTSP DESCRIBE.
It brings also minor modifications : 
- multiple delete when RTSP SETUP fails
- send TEARDOWN before closing RTSP connection
- reset the IP_MULTICAST_ALL if option exists in order to share same port between streams using different multicast group.

Regards,
Michel.


lminiero:
Thanks, I'll look into this next week!
Quick question, why is ```janus_streaming_create_fd``` asking for both ```listenername``` and ```medianame``` if they're always the same?

mpromonet:
It was just to keep the same log, the listenername start with a capital letter, but not the medianame.
Could be better to make a choice an use the same information in both log.

lminiero:
Sorry for this late response, I've been very busy and had no time to look into pull requests...

At its current stage, I cannot merge this patch. It currently breaks the regular RTP streaming for some reason: I tried the default demo with the default GStreamer script, and I only see video being detected. This likely means some issues in setting up the sockets to listen for RTP packets. Since your patch changes the way we implemented this to use your new helper method instead, I guess the problem is somewhere there.

If you find a way to fix this I'll gladly evaluate the path again. Thanks!

mpromonet:
Sorry, I mixed audio and video information, I will upgrade the pull request.

mpromonet:
I made tries with the script in plugins/streams/test_gstreamer_1.sh (I guess this is "the default demo with the default GStreamer script") 

     gst-launch-1.0  audiotestsrc ! audioresample ! audio/x-raw,channels=1,rate=16000 ! opusenc bitrate=20000 ! rtpopuspay ! udpsink host=127.0.0.1 port=5002 \
                      videotestsrc ! video/x-raw,width=320,height=240,framerate=15/1 ! videoscale ! videorate ! videoconvert ! timeoverlay ! vp8enc error-resilient=true ! rtpvp8pay ! udpsink host=127.0.0.1 port=5004

It seems to working fine, do you have still problems with this push request ?

lminiero:
I'll check the new changes as soon as I get back to the office, thanks!

lminiero:
Sorry if this took a while... I can confirm it is now working as expected, thanks! Merging.

Issue #268, opened 2015-06-26
[sip] Response code when declining a call


saghul:
(I thought I'd open an issue before making a patch this time around)

Currently, when declining a call, the response code used is 603, which is unfortunate, because all other devices configured with the same account will stop ringing.

@lminiero how about getting an extra parameter with the decline command specifying the code? In its absence 486 would be used, since it better communicates the reason for declining the call.

saghul:
(Once agreed I volunteer to write the patch)

lminiero:
I agree that a parameter for explicitly requiring a 603 is better, since in general we have to assume that the SIP client is unaware of other devices configured for the same account and should behave accordingly.

saghul:
Fixed in #270

PR #270, opened 2015-06-29
sip: add ability to choose the response code for 'decline'

### LINKED ISSUE #267: [sip] Even when call is declined
(I thought I'd open an issue before making a patch this time around)

The SIP plugin emits an event of type "ack" when the user uses the "decline" command to reject an incoming call: https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_sip.c#L1234

This is kind-of weird, since the initial decline command already got an "ack" response. @lminiero what do you think about emitting a "declined" event instead? It could also be a "hangup" event with reason "Call Declined".

saghul:
Also emit a 'declined' event afterwards.

Closes #267 and #268.

### DIFF
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -160,6 +160,9 @@ $(document).ready(function() {
 										} else if(event === 'calling') {
 											console.log("Waiting for the peer to answer...");
 											// TODO Any ringtone?
+											$('#call').removeAttr('disabled').html('Hangup')
+												  .removeClass("btn-success").addClass("btn-danger")
+												  .unbind('click').click(doHangup);
 										} else if(event === 'incomingcall') {
 											console.log("Incoming call from " + result["username"] + "!");
 											var doAudio = true, doVideo = true;
@@ -224,9 +227,6 @@ $(document).ready(function() {
 											if(jsep !== null && jsep !== undefined) {
 												sipcall.handleRemoteJsep({jsep: jsep, error: doHangup });
 											}
-											$('#call').removeAttr('disabled').html('Hangup')
-												.removeClass("btn-success").addClass("btn-danger")
-												.unbind('click').click(doHangup);
 										} else if(event === 'hangup') {
 											if(incoming != null) {
 												incoming.modal('hide');
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -232,8 +232,8 @@ $(document).ready(function() {
 												incoming.modal('hide');
 												incoming = null;
 											}
-											console.log("Call hung up by " + result["username"] + " (" + result["reason"] + ")!");
-											bootbox.alert(result["reason"]);
+											console.log("Call hung up (" + result["code"] + " " + result["reason"] + ")!");
+											bootbox.alert(result["code"] + " " + result["reason"]);
 											// Reset status
 											sipcall.hangup();
 											$('#dovideo').removeAttr('disabled').val('');
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -1336,6 +1336,33 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			break;
 		case nua_i_state:
 			JANUS_LOG(LOG_VERB, "[%s]: %d %s\n", nua_event_name(event), status, phrase ? phrase : "??");
+			tagi_t const *ti = tl_find(tags, nutag_callstate);
+			enum nua_callstate callstate = ti ? ti->t_value : -1;
+			/* There are several call states, but we only care about the terminated state
+			 * in order to send the 'hangup' event.
+			 * http://sofia-sip.sourceforge.net/refdocs/nua/nua__tag_8h.html#a516dc237722dc8ca4f4aa3524b2b444b
+			 */
+			if (callstate == nua_callstate_terminated) {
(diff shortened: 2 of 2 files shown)


lminiero:
Thanks! Just a couple of quick comments:

* Should we also add a upper bound for a valid response code, or considering the fact that new error codes might be added later on it's better to keep this open?
* Have you checked if the new ```declined``` event confuses the existing demo? IIRC we don't really care about that ack and still wait for a ```hangup``` whatever the cause, so if that's still there it shouldn't be an issue.

saghul:
> Should we also add a upper bound for a valid response code, or considering the fact that new error codes might be added later on it's better to keep this open?

That's what I thought, so I left it "open".

> Have you checked if the new declined event confuses the existing demo? IIRC we don't really care about that ack and still wait for a hangup whatever the cause, so if that's still there it shouldn't be an issue.

I'll double check, but IIRC there is no "hangup" event as a result of the "decline" action.

lminiero:
In the plugin code, a "hangup" event is sent everytime a ```nua_r_bye```, ```nua_i_bye```, ```nua_i_cancel``` or an error greater than or equal to ```400``` is notified by the Sofia SIP stack. So, in general, it notifies the SIP session is over, if it ever started, whatever the cause. Not sure if the 486, like the 603, triggers any of those anyway, so that's why I wanted to be sure.

saghul:
In this case none of those callbacks get's executed, just `nua_i_terminated`, I guess we should emit the hangup event from there, right now there is no hangup event:

~~~~
[2424582235] There's a message for JANUS SIP plugin
{
   "request": "decline"
}
Handling message: {
   "request": "decline"
}
Pushing event: {
   "sip": "event",
   "result": {
      "event": "declined",
      "code": 486
   }
}
[2424582235] Adding event to queue of messages...
  >> 0 (Success)
[nua_i_state]: 486 Busy Here
[nua_i_terminated]: 486 Busy Here
We have a message to serve...
	{
   "janus": "event",
   "session_id": 1408897944,
   "sender": 2424582235,
   "transaction": "xUMky90PDMQj",
   "plugindata": {
      "plugin": "janus.plugin.sip",
      "data": {
         "sip": "event",
         "result": {
            "event": "declined",
            "code": 486
         }
      }
   }
}
Got a HTTP GET request on /janus/1408897944...
Session 1408897944 found... returning up to 1 messages
Long poll time out for session 1408897944...
We have a message to serve...
	{"janus" : "keepalive"}
~~~~

I'll update the patch.

lminiero:
IIRC ```nua_i_terminated``` is more generic, though, that is it's called whenever a session has been terminated in general. As such, with the current code this might lead to multiple ```hangout``` events for the same event (e.g., you send a BYE, you get a ```hangout``` at ```nua_r_bye``` and another at ```nua_i_terminated```). The documentation seems to confirm that:

http://sofia-sip.sourceforge.net/refdocs/nua/nua_8h.html#abca36033336ce16e538a279413d2ca51a7db4a7a146b770f1636f612c07b63da0

although apparently it's been deprecated in the meantime in favour of ```nua_i_state```. Anyway, whatever the actual event, if that's indeed the case, we might think of only issuing a ```hangout``` at that specific event, as all origins would flow to that state anyway. What do you think about this?

saghul:
Yeah, that might work! I'll give it a shot.

saghul:
@lminiero Done, can you please take a look?

Note this introduces a couple of changes in the hangup event:

* there is no longer a 'username' parameter. I'm not 100% sure what it was supposed to represent, was it the origin? At any rate, I don't see how we can guess who triggered the hangup, but IMHO is not that relevant and the user could figure it out by remembering if (s)he called hangup/decline or not.
* there is a new 'code' parameter. This can be used to or example catch cancellations, since the code will be 487, or to give a better phrase, since the code it what matters, not the phrase.

lminiero:
Looks good, thanks! I'll give it a try myself later as well just to be sure there isn't any scenario that may not be covered, which I'm sure is not the case.

I'm fine with the changes you made to the hangout event: the username part wasn't actually used in the demo, as the only occurrence I see is in a console.log,

https://github.com/meetecho/janus-gateway/blob/master/html/siptest.js#L235

The only thing that is notified to the user in the alert is the reason that should always be there (by the way, with the state interception, is this actually the case?). I'll check what the codes can be in the different cases (e.g., remote cancel, response to a BYE, etc.) in order to see if I can update the demo verbosity accordingly.

saghul:
> Looks good, thanks! I'll give it a try myself later as well just to be sure there isn't any scenario that may not be covered, which I'm sure is not the case.

Thanks! There is one thing uncovered: local CANCEL, but that's currently not working anyway, and it will be covered once I add support for it, but that's for another PR.

> The only thing that is notified to the user in the alert is the reason that should always be there (by the way, with the state interception, is this actually the case?). I'll check what the codes can be in the different cases (e.g., remote cancel, response to a BYE, etc.) in order to see if I can update the demo verbosity accordingly.

I can modify the demo so it prints "%d %s", code, reason, which is what "reason" used to contain in certain cases.

saghul:
@lminiero added a new commit, the alert box will now print the code and phrase.

saghul:
Fixed cancelling outgoing calls in the SIP demo.

lminiero:
Just checked a few use cases and it looks great, thanks again!

saghul:
:tada: Thanks!

PR #271, opened 2015-07-01
sip: add ability to specify a prehashed secret (ha1) 

saghul:


lminiero:
If I look at what the SIP plugin was before you started playing with it I'm almost embarassed, you did a terrific job :smile: 

The patch looks fine to me, I just have a code style related nit: could you prefix the new enum and its values with ```janus_sip_``` instead of just ```janus_```, and place the property at the end? e.g., ```janus_sip_secret_type_unknown``` instead of ```janus_unknown_secret_type```. It makes it better adhere to the code style we have in the core and other plugins, and also simplifies any search in the Doxygen documentation.

Thanks!

saghul:
> If I look at what the SIP plugin was before you started playing with it I'm almost embarassed, you did a terrific job :smile:
> 

Thanks! Janus was exactly what I was looking for, you can't imagine how much time this has saved me :heart_eyes: 

> The patch looks fine to me, I just have a code style related nit: could you prefix the new enum and its values with janus_sip_ instead of just janus_, and place the property at the end? e.g., janus_sip_secret_type_unknown instead of janus_unknown_secret_type. It makes it better adhere to the code style we have in the core and other plugins, and also simplifies any search in the Doxygen documentation.

Fixed that!

lminiero:
Perfect, thanks again, merging!

PR #273, opened 2015-07-02
memory check and clean; if videoroom is null, create it firstly; verify room secret

nanguantong:
memory check and clean; if videoroom is null, create it firstly; verify room secret

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -583,6 +583,8 @@ int janus_videoroom_init(janus_callbacks *callback, const char *config_path) {
 			janus_videoroom *videoroom = calloc(1, sizeof(janus_videoroom));
 			if(videoroom == NULL) {
 				JANUS_LOG(LOG_FATAL, "Memory error!\n");
+				/* FIXME add by vivi */
+				cat = cat->next;
 				continue;
 			}
 			videoroom->room_id = atoi(cat->name);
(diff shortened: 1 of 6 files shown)


nanguantong:
"add videomcu custom page"   --- this commit (diff) is my own app scene, so don't need to merge this

plugins/janus_videoroom.c
```diff
@@ -593,6 +595,8 @@ int janus_videoroom_init(janus_callbacks *callback, const char *config_path) {
 				description = g_strdup(cat->name);
 			if(description == NULL) {
 				JANUS_LOG(LOG_FATAL, "Memory error!\n");
+				/* FIXME add by vivi */
+				cat = cat->next;
 				continue;
```
ploxiln: `g_strdup()` will never return NULL, it will abort the program instead. So the whole if-body can be removed instead.

https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html

  > If any call to allocate memory fails, the application is terminated. This also means that there is no need to check if the call succeeded.

nanguantong: Yes, so this diff does't affect the program exit.

lminiero:
I'm not sure I understand the purpose of this patch. Was it actually meant to be shared? What is it trying to fix? Or did you just want to highlight some parts of the code you have concerns about?

lminiero:
Closing for lack of feedback.

PR #275, opened 2015-07-02
sip-demo: use HA1 hashed passwords

saghul:
This way the user's password will never travel unhashed towards the
server.

The MD5 implementation used is this one (MIT licensed):
https://github.com/blueimp/JavaScript-MD5


lminiero:
Cool, thanks! Before merging I'll try it later with some of the servers I know of just to make sure forcing secrets this way doesn't break the behaviour for users trying the demo.

lminiero:
Can't merge it yet, sorry... I've just tried it with Asterisk and the registration failed with a "wrong password". Changing it to use ```secret``` instead of ```ha1secret``` worked instead. I guess that not all SIP servers support this way of registering, or at least not in their default configuration; or simply it's a more delicate approach when it comes to configuration (e.g. realms) and so more prone to errors. As such, the SIP demo cannot be configured to use that by default.

saghul:
I'll re-test , but just to be clear, there is no difference in what goes on the SIP packet. My guess is that the realm needs to be empty / or set to that of the challenge. I'll have another go at it.

saghul:
@lminiero do you happen to have a capture of the 401 / 407 packet from Asterisk?

lminiero:
I'll generate one for you, just give me a few minutes.

lminiero:
You can find it here:

http://conf.meetecho.com/pull275-badauth.pcapng

Hope it helps! Let me know if you need any additional info.

saghul:
Got it. I see we cannot really apply this PR to the main demo, here is the reasoning:

The HA1 secret is basically MD5(username:realm:password). In SIP the realm is usually set to the domain, since the credentials for saghul@foo.com should not automagically work on saghul@bar.com even if both are handled by the same server.

Asterisk, however, uses "asterisk" as the realm (you can see that in the traces), which means that the HA1 we'd need to generate is MD5(username:asterisk:password).

We cannot really know what realm we're going to get in the demo, since it might be connected to any server, so I'll just dump this PR.

In paradise city, where the grass is green, girls are pretty and people use user@domain SIP addresses this works beautifully :-)

lminiero:
Well the pull request looked useful anyway! Especially since you have a use for it with your servers, for instance, as I'm sure is the case for others. I'd just keep the default secret mechanism in the demo, leaving the optional usage of HA1 secret to those who want it. But you're much more experienced with SIP than I am, so I'll  leave the decision to you :smile:

saghul:
Ok, I'll reopen and make it optional with a checkbox pehaps? Then a note can explain what this is actually about.

lminiero:
What about if we turn the guestbox checkbox we have now into a select box of some sort? Something like a select with different options, namely regular secret, HA1 secret and guest. This way we keep the UI clean enough without putting too much stuff there: selecting one of the options might show a small dialog explaining what the option is going to do.

Just thinking out loud, and this is something I can take care of myself if you want, maybe starting from the guest/noguest stuff we have now, so that you can expand on that.

saghul:
> What about if we turn the guestbox checkbox we have now into a select box of some sort? Something like a select with different options, namely regular secret, HA1 secret and guest. This way we keep the UI clean enough without putting too much stuff there: selecting one of the options might show a small dialog explaining what the option is going to do.
> 

That could work!

> Just thinking out loud, and this is something I can take care of myself if you want, maybe starting from the guest/noguest stuff we have now, so that you can expand on that.

That would be great, since my UI / bootstrap skills are lacking :-) If you create the selection box I could just add the option on top and rebase this PR.

lminiero:
Ok, I'll notify you here as soon as it is ready.

lminiero:
Just pushed the changes. Integrating your new approach should be simple enough.

Add a new li line here to list the HA1 secret approach before guest:
https://github.com/meetecho/janus-gateway/blob/master/html/siptest.html#L124

Add a new case here for the info dialog when switching approach (the ID of the element created before is used to discriminate approaches):
https://github.com/meetecho/janus-gateway/blob/master/html/siptest.js#L101

Add an else here to specify the ha1secret if selected:
https://github.com/meetecho/janus-gateway/blob/master/html/siptest.js#L449

Let me know if you need any clarification on this.

saghul:
@lminiero Updated! I tested it both with normal secret and with ha1 and it works fine :-)

lminiero:
Cool, thanks! Merging then :-)

PR #276, opened 2015-07-02
sip: add ability to skip registration

### LINKED ISSUE #274: sip: Registering... but not really
Hi!

I wanted to share the ideas I have for the patch I'm working on, in case someone (I'm looking at you @lminiero :smile: ) has feedback.

What I really need is the ability to make multiple calls with the SIP plugin from the same account. Now, the plugin seems to be designed around the idea of "one plugin handle -> one SIP account -> one call", and adding multiple sessions support looks messy, so I came up with the following workaround:

* An application would have a couple (or more) plugin handles for a given account, so calls can be received over them.
* The same application would create new plugin handles for each outgoing call, but **not** send a REGISTER over them, just use the "register" command to store the credentials and then make the outgoing call.

In order to achieve this I plan to add a boolean attribute to the "register" command, in order to skip the registration process.

saghul:
Fixes #274

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -175,6 +175,7 @@ void janus_sip_message_free(janus_sip_message *msg) {
 
 
 typedef enum {
+	janus_sip_registration_status_disabled = -2,
 	janus_sip_registration_status_failed = -1,
 	janus_sip_registration_status_unregistered = 0,
 	janus_sip_registration_status_registering,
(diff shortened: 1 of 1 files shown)


lminiero:
I'll check the patch in more detail as soon as I can, but I'd already ask you to change the ```json_boolean``` you have in something's different. Apparently some older versions of Jansson don't have ```json_boolean```, which is why I had to use strings ("true", "false") or integers(1, 0) throughout the core and the other plugins to account for that.

saghul:
Oh, I didn't know this, will fix!

saghul:
Fixed.

lminiero:
Sorry about this late response, I've been awfully busy these days... merging, thanks again!

saghul:
No worries, thanks!

PR #282, opened 2015-07-14
Prevent bower to use a too recent adapter.js

ancorgs:
Janus.js uses MediaStreamTrack.getSources which is not longer provided for Firefox by adapter.js. This change ensures bower uses the last adapter.js version still defining that function which prevents janus.js from breaking. The limitation can be removed in the future if janus,js stops calling MediaStreamTrack.getSources directly.

Since I was modifying the file, I also updated the version.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "janus-gateway",
-  "version": "0.0.7",
+  "version": "0.0.9",
   "homepage": "https://github.com/meetecho/janus-gateway",
   "authors": [
     "Lorenzo Miniero <lorenzo@meetecho.com>",
@@ -33,7 +33,7 @@
     "tests"
   ],
   "dependencies": {
-    "adapter.js": "*",
+    "adapter.js": "git://github.com/webrtc/adapter.git#d197a97c198a6dbcde7aa58c7d50d454af7bb99c",
     "jquery": "*"    
   }
 }


lminiero:
Not familiar with bower as the related support file was provided by someone else as a pull request. Why not have bower get the adapter.js we ship ourselves? That's where we define getSources anyway, and is compliant with janus.js.

lminiero:
https://github.com/meetecho/janus-gateway/blob/master/html/adapter.js

saghul:
Since the Janus demos ship with both jQuery and adapter.js, I don't see why the bower file is necessary, at all. If You were to split the client library to another repo then yeah, I guess.

ancorgs:
I just wanted to be consistent with the current definition of bower.json. I can also try to completely remove the dependency and use the bundled adapter.js if you prefer so. I'll try and open a separate pull request if it works.

ancorgs:
If you want it to work with bower, you should use module imports instead of the current `addJS`. That would make possible to modern javascript package managers, like bower, to process the dependencies. Is that ok for you?

http://www.2ality.com/2014/09/es6-modules-final.html

lminiero:
To be brutally honest, I can't care less about bower ;-)
I'd rather not constrain our JS API to any approach like bower, browserify or the like, if possible, but keep it simple.

ancorgs:
And to be honest, I'm not sure how the whole import thingy will work in this case or even if it's possible at all (I'm not a javascript or bower expert).

On the other hand, it would probably be possible to place both files `janus.js` and `adapter.js` in the `main` section of the bower.js That will eventually work (I think) but would be actually incorrect (there should be only one `.js` file in the `main` section).

If you don't care, just merge the PR. It only improves the situation. :wink: for those (like me) that are kind of forced to use bower.

lminiero:
Makes sense, thanks!

PR #292, opened 2015-07-25
Introduce Janus JS logger

yan-foto:
As I was lost in the huge amount of logs, which are produced by `janus.js` when debugging is enabled, I created as simple logger with log level support which makes use of browsers `console.trace`, `console.warn`, etc. and provides fallback methods if these are not supported.

If it is something desirable, I could go ahead and replace `Janus.log` in javascript code with appropriate methods of the logger before the merge.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -28,19 +28,19 @@ Janus.logger = function(level) {
 	methods = ["trace", "debug", "info", "warn", "error"];
 	levels = {};
 
-	this.log = function(msg) {
-		(console.log === undefined) ? Janus.noop : console.log(msg);
+	this.log = function() {
+		(console.log === undefined) ? Janus.noop : console.log.apply(console, arguments);
 	};
 
 	methods.forEach(function(level, index) {
 		levels[level] = index;
 
-		this[level] = function(msg) {
-			if(this.level > levels[level]) {
+		this[level] = function() {
+			if(this.level - 1 > levels[level]) {
 				return; // No need to log at all!
 			}
 
-			(console[level] === undefined) ? this.log(msg) : console[level](msg);
+			(console[level] === undefined) ? this.log.apply(this, arguments) : console[level].apply(console, arguments);
 		};
 	}.bind(this));
 }
@@ -165,7 +165,7 @@ function Janus(gatewayCallbacks) {
 	this.destroy = function(callbacks) { destroySession(callbacks); };
 	this.attach = function(callbacks) { createHandle(callbacks); };
 
-	
+
 	// Private method to create random identifiers (e.g., transaction)
 	function randomString(len) {
 		charSet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
@@ -209,7 +209,7 @@ function Janus(gatewayCallbacks) {
 			dataType: "json"
 		});
 	}
-	
+
 	// Private event handler: this will trigger plugin callbacks, if set
 	function handleEvent(json) {
 		retries = 0;
@@ -328,7 +328,7 @@ function Janus(gatewayCallbacks) {
 			Janus.log("Unknown message '" + json["janus"] + "'");
 		}
 	}
-	
+
 	// Private helper to send keep-alive messages on WebSockets
 	function keepAlive() {
 		if(server === null || !websockets || !connected)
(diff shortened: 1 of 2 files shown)


lminiero:
That looks realy interesting! I'll look into it when I get back home (still abroad).
That said, before I can merge this, I'd ask you to sign our CLA if you haven't done that yet:

Thanks!

yan-foto:
Agreement is signed.

lminiero:
Yep, I get that :smile: 

It definitely looks like something that would improve things. Can you adapt just one of the demos (e.g., echo test or whatever you use the most) just to see how it would look like?

Thanks!

yan-foto:
I enhanced the logger and used it in `streamingtest.js`. However to have clean output It should be added to `janus.js` first and foremost!

yan-foto:
I enhanced the logger and used it in `streamingtest.js`. However to have clean output It should be added to `janus.js` first and foremost!

yan-foto:
Any updates on this?

lminiero:
Sorry, github apparently didn't notify me about the further changes you made 5 days ago and I'm only now seeing them... I'll check on Monday when I get back to work. Thanks!

lminiero:
Sorry for the abysmal delay, just manager to get to this after my vacation... just tested and this looks great, thanks!

Just wondering if we can make this slightly more "transparent" than creating a logger specifically. Namely, for now we have a generic ```Janus.log``` that either prints something or doesn't, according to what you passed as the ```debug``` value in ```Janus.init```. My feeling is that this new logging approach belongs to the ```init``` as well, as conceptually you shouldn't have a logger before you initialized the library: besides, it being outside of the ```init``` might make it less clear on what should be done within ```janus.js``` itself.

Not sure what the best way to proceed on this would be: one way could be creating wrappers called ```Janus.trace```, ```Janus.warn``` etc. pretty much as we do for ```Janus.log=console.log```, and then turning the ```debug``` value in init from boolean to something else (e.g., numeric identifier for logging level, comma separated list of stuff you want logged, etc.). What's your feeling about this? This is not something I'd ask you to do, of course: I can take care of this starting from the examples you provided.

Thanks again!

yan-foto:
> My feeling is that this new logging approach belongs to the init as well, as conceptually you shouldn't have a logger before you initialized the library

I do absolutely agree with you. I think its a good idea to initialize the logger inside init and expose it through the wrappers as you mentioned.

> What's your feeling about this?

Big Picture: to be honest as I was trying to integrate `janus.js` into a project, and I would say (thinking on #310) we need to extract the javascript framework from the main structure and let it exist as its own module. I could most definitely give a hand to conceptualize and kick start a `janus.js` module.

lminiero:
> I do absolutely agree with you. I think its a good idea to initialize the logger inside init and expose it through the wrappers as you mentioned.

Ok, I'll craft a new version of ```janus.js``` deriving from your changes. I'll probably create a new PR for that, I'll notify it here.

> Big Picture: to be honest as I was trying to integrate janus.js into a project, and I would say (thinking on #310) we need to extract the javascript framework from the main structure and let it exist as its own module. I could most definitely give a hand to conceptualize and kick start a janus.js module.

Not sure I understand, but everything that can make it easier to integrate elsewhere would be more than welcome! If you have ideas on how to do that (I'm unfortunately all but a JavaScript wizard, I'm afraid) that would be great.

yan-foto:
Sorry I have this problem of not being able to express myself!

At the moment if someone wants to use the JS module, he should try to make sense of the files included under `html` directory and copy them into their own projects. My suggestion is to have an npm package to be used in CommonJS compatible environments and a bower package for everyone who wants to have it inside the browser. It would be easier this way to build a modular project and maintain dependencies.

lminiero:
Ah in this case I'm not so sure about it... I'm not a fan of frameworks, and I don't want to constrain our JavaScript library to any of them. Different people use different things, and forcing them to rely, e.g., on bower/npm could prevent them from integrating it somewhere else.

yan-foto:
I can understand your concerns. But let me put it this way: we will have a separate repository with the following structure:

```
|-- dist
|   |-- janus.js
|   |-- stream.plugin.js
|-- src
|   |-- janus.es6
|   |-- stream.plugin.es6
|   |-- other-deps.es6
|-- gulpfile.js
|-- bower.json
|-- package.json
```

Everyone who needs the production version uses the ready-to-use file available under `dist` or let bower (see `bower.json`) download it. The others who want to use it in CommonJS environments can just use the npm package (see `package.json`). Moreover the project will be easier to maintain using a build system such as gulp (see `gulpfile.js`).

Does it make sense?

lminiero:
I eventually integrated part of your approach in janus.js, so that we now also have Janus.trace, Janus.debug, etc. The way you selectively enable and disable them is through the pre-existing ```debug``` attribute in Janus.init: passing ```true``` or ```"all"``` enables them all, ```false``` disables debugging entirely, and an array similar to the ```methods``` you had (e.g., ```debug: ["warn", "error", "log"]```) makes sure that only the ones you chose be printed.

Thanks for your feedback and help on this!

PR #295, opened 2015-07-28
sip: fix setting the correct caller for the incomingcall event

saghul:
Before the change the "username" field in the "incomingcall" event would only contain the username, not the domain. Now it contains the full SIP URI. Example:

~~~~
{
   "janus": "event",
   "session_id": 890176274,
   "sender": 874965051,
   "plugindata": {
      "plugin": "janus.plugin.sip",
      "data": {
         "sip": "event",
         "result": {
            "event": "incomingcall",
            "username": "sip:31208005163@ag-projects.com"
         }
      }
   },
   "jsep": {
      "type": "offer",
      "sdp": "...."
~~~~

Now, "username" sounds a bit weird, would you be open to changing it to "caller"? I think it makes more sense in the context of this event.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -1463,7 +1463,6 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				nua_respond(nh, 486, sip_status_phrase(486), TAG_END());
 				break;
 			}
-			const char *caller = sip->sip_from->a_url->url_user;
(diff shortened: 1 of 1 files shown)


lminiero:
Makes sense but as usual I'm a bit afraid to change things that may break existing applications using this. I'd say let's keep it named ```username``` for now, with your change.

saghul:
:+1:

PR #296, opened 2015-07-29
sip: send a 'registration_failed' event when SIP registration fails

saghul:
Make it easy for applications / libraries to know that registration
failed and with what code and reason.


lminiero:
Definitely helpful, merging, thanks!

lminiero:
Ops, wrong button, sorry :smile:

PR #301, opened 2015-08-05
doc: update usrsctp repository location

saghul:
It has moved to GitHub.


lminiero:
Accepting blindly as I'm on vacation with no access to a pc...

saghul:
Haha, enjoy Lorenzo! And stay away from computers for a while :-)

PR #303, opened 2015-08-07
build: clean doxygen generated sqlite files

saghul:
The SQLite DB is left back, thus creating problems with Debian
packaging, because the tree is not reverted back to its original form
when doing `make distclean`.


lminiero:
Blindly accepting again... Thanks!

PR #305, opened 2015-08-10
build: don't build static versions of the modules by default

saghul:


lminiero:
Looks useful, I don't know why we were building static versions anyway... I'll merge this when I get back to work on Monday, thanks!

ploxiln:
Just want to say thanks for this kind of thing... I'm good with makefiles, but put off enough by the arcane m4 of autotools to not dig in except when absolutely necessary.

PR #311, opened 2015-08-20
Check if adapter is already loaded

MathRobin:
I adapter is already loaded, it's not necesary to try to charge it again. Or if it's not available at root, trying to load it as it will cause a fatal error. It is probably most pertinent if you can let the choice to the user of the place of the script

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -56,7 +56,6 @@ Janus.init = function(options) {
 			if(src === 'adapter.js') {
 				if(window.RTCPeerConnection) {
 					// Already loaded
-					options.callback();
 					return;
 				}
 			}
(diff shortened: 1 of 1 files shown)


lminiero:
This makes sense, although I'm a bit concerned about what may happen if a different, non-compliant version of the adapter is being loaded that may screw things up (shouldn't happen). I'll merge this on Monday when I get back to work, thanks!

MathRobin:
I thing that you should precise in readme what is the compliant(s) version(s) of adapter. If people wants to use another one, they do it at their own risks.

lminiero:
Fair point, I think I'll add this to the documentation somewhere.
As a side note, have you signed our CLA?

Unfortunately we can't accept pull requests without that. Thanks!

lminiero:
Sorry, completely forgot about this in the meantime! Merging right now, thanks again.

PR #313, opened 2015-08-27
MOAR init systems docs

saghul:


lminiero:
Nice addition, thanks!

lminiero:
Updated online docs as well: http://janus.conf.meetecho.com/docs/service.html

Issue #242, opened 2015-05-20
Audio conference audio is very delayed (in audiobridge)


mattvv:
Steps to reproduce

1. Join a call using audioconf, in my case using 48000hz has the mixing.
2. Talk and observe that both parties can hear eachother without delay
3. Wait a few minutes
4. Observe delay of voice is increasing
5. Wait 10-15 minutes
6. Observe delay increase again

I've had the delay go to 30+ seconds from speaking in some of my tests.

nowylie:
I'm guessing you mean `48000hz`, which could introduce lag if the mixing thread is taking more then 20ms to mix streams. 

AFAIK this is caused by the way that the audiobridge demo plugin buffers/unbuffers packets before mixing. I also noticed the delay before I started work on my own plugin.

mattvv:
Yep you're right. I can also reproduce this on 24000hz. I'm running janus on an ec2 c4.large instance.

lminiero:
I never experienced anything like that, and I've tested it for quite a long time on my "crappy" laptop, so something much less powerful than a large ASW instance... Are you using the latest version?

lminiero:
Just did one more test for about 15 minutes in a 48khz room and it worked fine for me. Make sure you're using the latest version of Janus.

mattvv:
Upgraded to latest version of Janus and tried exact same situation with same person for 3+hour call. No delay. Thank you @lminiero again! Closing.

willbarrett:
@lminiero we're seeing this issue as of 76a4d5cce831bc77ff09cd25c778132831debf59 - what commit do you believe resolved the problem?

lminiero:
@willbarrett not sure as I could never replicate the issue, or at least it wasn't there anymore when I tried.

willbarrett:
OK. We have multiple reports of audio delays from folks using our platform. What information can we get you that would help figure out the cause of this?

lminiero:
I don't think there's any piece of info that could pin down the issue as of now.

For progressive delays, I guess the queues might be at fault here. Is this happening in rooms with a lot of participants? Or on servers that are mixing a lot? Or is it just random?

Tomorrow I'll try to add the queues length info to the admin API, so that you can poll that when it happens: a normal behaviour should be a low and constant queue size; if the queues keep on increasing instead it might mean the mixer can't keep up with the packets for some reason.

willbarrett:
[...]

PR #317, opened 2015-08-31
First attempt at getting rid of the occasional increasing delay in audiobridge

lminiero:
This pull request changes the way incoming packets are handled in the audiobridge. It's not an actual jitter buffer or something like that, but it's some sort of way to buffer and sort (using sequence numbers) incoming packets, putting a cap to how many packets are queued. To make sure there's a buffer to start from in general, when users first join a room a few packets are queued before adding them to the mix: this means a bit of additional delay for all users but nothing sensible; once the prebuffer is filled, no more buffering is done. When during the conference the threshold is reached for a user (queue too large), older packets are removed and not added to the mix, favouring more recent packets.

Feedback welcome.

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -518,12 +518,13 @@ typedef struct janus_audiobridge_participant {
 	janus_audiobridge_room *room;	/* Room */
 	guint64 user_id;		/* Unique ID in the room */
 	gchar *display;			/* Display name (just for fun) */
+	gboolean prebuffering;	/* Whether this participant needs pre-buffering of a few packets (just joined) */
 	gboolean active;		/* Whether this participant can receive media at all */
 	gboolean working;		/* Whether this participant is currently encoding/decoding */
 	gboolean muted;			/* Whether this participant is muted */
 	int opus_complexity;	/* Complexity to use in the encoder (by default, DEFAULT_COMPLEXITY) */
 	/* RTP stuff */
-	GQueue *inbuf;			/* Incoming audio from this participant */
+	GList *inbuf;			/* Incoming audio from this participant, as an ordered list of packets */
 	GAsyncQueue *outbuf;	/* Mixed audio for this participant */
 	janus_mutex qmutex;		/* Incoming queue mutex */
 	int opus_pt;			/* Opus payload type */
@@ -544,6 +545,25 @@ typedef struct janus_audiobridge_rtp_relay_packet {
 	uint32_t timestamp;
 	uint16_t seq_number;
 } janus_audiobridge_rtp_relay_packet;
+
+/* Helper to sort incoming RTP packets by sequence numbers */
+static gint janus_audiobridge_rtp_sort(gconstpointer a, gconstpointer b) {
+	janus_audiobridge_rtp_relay_packet *pkt1 = (janus_audiobridge_rtp_relay_packet *)a;
+	janus_audiobridge_rtp_relay_packet *pkt2 = (janus_audiobridge_rtp_relay_packet *)b;
+	if(pkt1->seq_number < 100 && pkt2->seq_number > 65000) {
+		/* Sequence number was probably reset, pkt2 is older */
+		return 1;
+	} else if(pkt2->seq_number < 100 && pkt1->seq_number > 65000) {
+		/* Sequence number was probably reset, pkt1 is older */
+		return -1;
+	}
+	/* Simply compare timestamps */
+	if(pkt1->seq_number < pkt2->seq_number)
+		return -1;
+	else if(pkt1->seq_number > pkt2->seq_number)
+		return 1;
+	return 0;
+}
 
 /* SDP offer/answer template */
 #define sdp_template \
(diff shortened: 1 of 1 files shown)


lminiero:
I'm talking to you, #242 :wink:

willbarrett:
@lminiero this seems to work. We haven't been able to replicate #242 since putting it in production, nor have our users most bothered by it.

lminiero:
Thanks, I'll merge then!

plugins/janus_audiobridge.c
```diff
@@ -1503,17 +1535,35 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+			/* Still pre-buffering: do we have enough packets now? */
+			if(g_list_length(participant->inbuf) == DEFAULT_PREBUFFERING) {
+				participant->prebuffering = FALSE;
+				JANUS_LOG(LOG_VERB, "Prebuffering done! Finally adding the user to the mix\n");
+			} else {
+				JANUS_LOG(LOG_VERB, "Still prebuffering (got %d packets), not adding the user to the mix yet\n", g_list_length(participant->inbuf));
+			}
+		} else {
+			/* Make sure we're not queueing too many packets: if so, get rid of the older ones */
+			if(g_list_length(participant->inbuf) >= DEFAULT_PREBUFFERING*2) {
```
tommoor: @lminiero is the `DEFAULT_PREBUFFERING*2` here very specific or just a value that should work? 

I've been hearing choppyness in calls where I think that packets are getting dropped off the queue too soon. Could it be that this queue should be larger when the quality is increased, that would increase the number of flowing packets right?

lminiero: It was just an attempt to put a cap on the buffer queue, which had no limit before. ```DEFAULT_PREBUFFERING``` being 6 packets, this means the queue contains 12 packets tops, hence about 240ms of audio (each RTP packet is 20ms). You may want to increase that if in your experience you noticed this causes trouble, and if you're ok with a potentially increased delay for affected users: e.g., ```3*6=18``` would mean up to 360ms of delay for users who get their queue filled. Users without issues, instead, will only have about 100ms of delay if you don't change the value of ```DEFAULT_PREBUFFERING``` itself.

If you come up with some values that give a good experience in most of your scenarios, just let us know and we'll make those the defaults.

lminiero: When I talk of delay, I'm talking of delay on the contributions from those users of course. Other participants would hear them a bit delayed because their queue is closer to full.

tommoor: Lorenzo - what do you think about making this value dynamic based on the bitrate of the connection? We've increased the bitrate of our calls to the maximum audio fidelity but now it appears that the connections are more susceptible to packets being dropped here...

lminiero: A dynamic value based on the connection quality would bring you back to step one. The main reason why we introduced this limited queue was to prevent the problem of users with troubled networks whose audio would increasingly become delayed over time, as we were queueing their audio indefinitely. Adapting the queue for those users and making it grow for them would cause the exact same problem over time.

Thinking about it, if you're talking about the receiving part (users getting audio) packets being dropped aren't necessarily being caused by this feature, and actually the cause may be substantially different. If users can't cope with the bitrate of the audio stream they're receiving because it's too high quality, a potential soultion would be to limit the bitrate for them only when we detect problems (e.g., encode a 16kHz stream for them although the room is 48kHz). We have an event that can be used this (plugins receive a slow_link event when too many NACKs are sent/received), so that can be used as a trigger to change the encoding parameters on the fly for the affected user. I'll have to check how easy it would be to add this to AudioBridge, but I'll only be able to look into that in a week more or less (busy in Japan ATM). The downside here is that we have no event for when things start being good again and you might increase the quality again, but it's a start.

tommoor: Yes, the drops are heard by the receiving end. To be clear we were definitely getting complains about drops in the audio before upping to 48khz however I think this might be more frequent now at the higher quality, which makes some sense.

Up until last week we were using the default quality which I believe is 16khz and this still exhibits the issue - you probably hear this on meetecho occasionally? Perhaps it could drop the quality in increments in an ideal world...

lminiero: Yeah dropping the quality for specific users was what I was talking about as a possibility to investigate, e.g., as a conseguence of slow_lnk events. We don't use the audiobridge on Meetecho usually, we have a SIP server (Asterisk) for audio conferencing and so call that via SIP plugin: that conference bridge still uses Opus and is configured at 16kHz, so the bandwidth is pretty much the same. Since we did the implementation of Opus support in Asterisk and we basically reused the same encoding/decoding code in the AudioBridge plugin as well, it's safe to assume the bandwidth consumption is pretty much the same.

Audio packets don't usually consume much bandwidth, but there are 50 of them per second (20ms each). Lost packets are typically NACKed by the peer who didn't receive them, so you may want to look into the admin API info for the affected handles, as we list the number of incoming (user lost packet Janus sent) and outgoing (Janus lost packet user sent) NACKs on there as well. If you see a lot of NACKs, something's broken on the media path between the server and the user (e.g., flaky WiFi or poor connectivity in general), and dropping the quality may or may not help.

PR #318, opened 2015-09-02
config: fix typo, 'apisecret' -> 'api_secret'

saghul:


lminiero:
Good catch! Although in the Janus API we then call it ```apisecret``` again...
https://github.com/meetecho/janus-gateway/blob/master/janus.c#L957
What a mess I am :smile:

PR #322, opened 2015-09-11
Added flags to check whether offer and/or answer have been received

lminiero:


saghul:
@lminiero I just deployed this to our servers. Will report back.

PR #325, opened 2015-09-14
README: update link to libsrtp, update build instructions

ploxiln:
I can't find any link to libsrtp 1.5.x on the sourceforge srtp site. It may help to link to the "new home" of libsrtp directly.

Also, I'd avoid suggesting to manually install libraries to `/usr` instead of `/usr/local`.

I understand both of these things are subjective. Just offering :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ To install it, you'll need to satisfy the following dependencies:
 * [Jansson](http://www.digip.org/jansson/)
 * [libnice](http://nice.freedesktop.org/wiki/)
 * [OpenSSL](http://www.openssl.org/) (at least v1.0.1e)
-* [libsrtp](http://srtp.sourceforge.net/srtp.html) (at least v1.5 suggested)
+* [libsrtp](https://github.com/cisco/libsrtp) (at least v1.5 suggested)
 * [Sofia-SIP](http://sofia-sip.sourceforge.net/)
 * [usrsctp](https://github.com/sctplab/usrsctp) (only needed if you
 are interested in Data Channels)
@@ -73,7 +73,7 @@ is quite straightforward:
 	wget https://github.com/cisco/libsrtp/archive/v1.5.0.tar.gz
 	tar xfv v1.5.0.tar.gz
 	cd libsrtp-1.5.0
-	./configure --prefix=/usr --enable-openssl
+	./configure --prefix=/usr/local --enable-openssl
 	make libsrtp.so && sudo make install
 
 * *Note:* you may need to pass --libdir=/usr/lib64 to the configure


ploxiln:
Also, fwiw, I / we have been using libsrtp-1.4.5 from ubuntu 14.04 all this time ... most stuff seems to work, but maybe we'll get fewer strange error log messages after I finish this round of updates ...

lminiero:
I'm not a fan of ```/usr/local```, personally, as I've often ended up with libraries that either couldn't be found or were in conflict with another one I had. That was exactly the case for some users with libsrtp, who despite hving installed 1.5 still encountered issues: turned out they still had the "repo" version of libstr installed in ```/usr``` and that was used instead. Since we want the library to be the base one for the system, I'd rather stick to ```/usr```.

lminiero:
libsrtp 1.4 is not broken per se: it just seems much more prone to leaks, at least in the usage we make of it. Several issues that we (and others) had with 1.4 disappeared when switching version. I guess changes were made to make it more compliant with what WebRTC requires.

saghul:
> Also, fwiw, I / we have been using libsrtp-1.4.5 from ubuntu 14.04 all this time ... most stuff seems to work, but maybe we'll get fewer strange error log messages after I finish this round of updates ...

I've been using 1.5 for quite a while and I did get fewer crashes, but warnings are still there. I also compiled it without OpenSSL support, I seemed to get fewer crashes. FWIW the Debian / Ubuntu packages on the AG Projects repository are built without OpenSSL integration.

> I'm not a fan of /usr/local, personally, as I've often ended up with libraries that either couldn't be found or were in conflict with another one I had. That was exactly the case for some users with libsrtp, who despite hving installed 1.5 still encountered issues: turned out they still had the "repo" version of libstr installed in /usr and that was used instead. Since we want the library to be the base one for the system, I'd rather stick to /usr.

Then you might override the version installed with the system, which might not even be binary compatible , doesn't that create more trouble?

lminiero:
The openssl support should make it more efficient, as part of the AES stuff is done in hardware. I wouldn't know if it makes crashes more frequent or not. Which warnings are you getting?

About /usr/local, I trust you guys about it being the better approach, but as I said it makes things harder on some systems, with libraries not being detected properly or conflicting versions. If you have any suggestion on how to make this easy for the dummy user in README, configure, etc., I'm quite willing to change that.

saghul:
> The openssl support should make it more efficient, as part of the AES stuff is done in hardware. I wouldn't know if it makes crashes more frequent or not.

I did some non-scientific benchmarks a while ago and I didn't really notice a high variance in performance.

> Which warnings are you getting?

Nothing to worry about, the usual ones at the start of the session.

lminiero:
> I did some non-scientific benchmarks a while ago and I didn't really notice a high variance in performance.

This should depend on the AES-NI availability in the OpenSSL code installed on the machine. Some time ago I found a couple of examples that showed the times OpenSSL would take with AES operations with and without that support:

     openssl speed -elapsed -evp aes-128-cbc
     OPENSSL_ia32cap="~0x200000200000000" openssl speed -elapsed -evp aes-128-cbc

which should give you different results in terms of speed. If they don't, it might mean the support is not available on your machine. That said, I haven't made specific tests in that direction myself.

ploxiln:
I install a few libraries manually to `/usr/local`; the way to get the janus build to reliably find them is to add

```
CFLAGS="-I/usr/local/include" LDFLAGS="-L/usr/local/lib -Wl,-rpath=/usr/local/lib"
```

as arguments to `./configure` for janus-gateway.

lminiero:
Fixed the path to libsrtp in both README and mainpage.dox (which was missing in this pull request). I didn't merge this directly as I'm still uncomfortable about the /usr/local stuff as a default, sorry!

ploxiln:
no problem, thanks

PR #326, opened 2015-09-14
refactor logging color output

ploxiln:
use a single log prefix table, and some bitops, instead of a ternary if

This is totally a subjective code style thing, consider it just a humble suggestion.


lminiero:
Sound like a cool addition, as it allows us to spare some operations. Would this still work should we add new debugging levels in the future?

ploxiln:
Yes, but you'd have to double the size of the array each time, and shift by one more (so `<< 4` and add 8 slots to both the no-color and with-color sections of the prefix array).

lminiero:
Got, thanks!

Issue #327, opened 2015-09-16
Timeout and Trickle error: no audio stream


vbraun:
I'm playing with the echo test plugin using my own code to call the janus rest API. It sends ice candidates and jsep offer asynchronously and as soon as the browser (crome 45) generates them.

Sometimes I get {code: 466, reason: "Trickle error: no audio stream"} back from the janus server. Generally it takes a few retries (first connect always works), and only if I have >=3 stun servers configured. It seems that the timeout for the trickle ice is too short:
~~~~
Got a HTTP OPTIONS request on /janus/3894809613/1981012534...
Got a HTTP OPTIONS request on /janus/3894809613/1981012534...
Got a HTTP OPTIONS request on /janus/3894809613/1981012534...
Got a HTTP OPTIONS request on /janus/3894809613/1981012534...
Got a HTTP OPTIONS request on /janus/3894809613/1981012534...
Got a HTTP OPTIONS request on /janus/3894809613/1981012534...
Got a HTTP POST request on /janus/3894809613/1981012534...
Handle 1981012534 supports trickle even if it didn't negotiate it...
[1981012534] Trickle candidate (audio): candidate:3013953624 1 udp 2122260223 192.168.1.100 55449 typ host generation 0
[1981012534] No stream, wait a bit in case this trickle got here before the SDP...
Got a HTTP POST request on /janus/3894809613/1981012534...
[1981012534] Trickle candidate (audio): candidate:854413036 1 udp 1686052607 91.64.xxx.xxx 55449 typ srflx raddr 192.168.1.100 rport 55449 generation 0
[1981012534] No stream, wait a bit in case this trickle got here before the SDP...
Got a HTTP POST request on /janus/3894809613/1981012534...
Got a HTTP POST request on /janus/3894809613/1981012534...
[1981012534] Trickle candidate (video): candidate:3013953624 1 udp 2122260223 192.168.1.100 50652 typ host generation 0
[1981012534] No stream, wait a bit in case this trickle got here before the SDP...
[1981012534] Trickle candidate (audio): candidate:3013953624 2 udp 2122260222 192.168.1.100 38641 typ host generation 0
[1981012534] No stream, wait a bit in case this trickle got here before the SDP...
Got a HTTP OPTIONS request on /janus/3894809613/1981012534...
Got a HTTP POST request on /janus/3894809613/1981012534...
[...]

PR #330, opened 2015-09-17
Asynchronous trickle request management

lminiero:
This pull request tries to address the issue reported in #327, where multiple trickle requests sent before  the offer could cause some sort of race condition due to the limit of connections browsers usually make of. The patch makes the ```trickle``` request asynchronous: each ```trickle``` request is acknowledged immediately, but is only handled right away if the any SDP has already been received/created; if not, the candidate is queued and acted upon receival of the remote SDP. Candidates older than 15 seconds are discarded.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -26,6 +26,7 @@
 #include "ice.h"
 #include "turnrest.h"
 #include "dtls.h"
+#include "sdp.h"
 #include "rtp.h"
 #include "rtcp.h"
 #include "apierror.h"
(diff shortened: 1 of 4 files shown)


vbraun:
Nice! This branch 100% fixes my issue; I haven't seen a single failure so far. I also didn't have to change my JS code.

lminiero:
Cool! I'll keep it on for a while, just to have more feedback on this. As anticipated, I don't want to risk breaking something that might work for others until I'm sure it's safe.

lminiero:
Looks like this isn't adding any issue, merging.

PR #332, opened 2015-09-18
Fixed a couple of memory leaks regarding sdp_parser usage. sdp_parser…

MichaelB76:
…_free() was not being called if the call to sdp_session() failed or if a SIP re-invite was received. In the latter case a significant amount of memory was being leaked when using a SIP client that sends periodic re-invites. The memory was cleaned up when the SIP session was terminated, only really being a problem for connections that stay up for a long time, such as a SIP trunk.


lminiero:
Good catch, thanks for spotting!

MichaelB76:
No problem, glad I found something useful.

PR #334, opened 2015-09-21
Raise default buffer size to 8192

saghul:
Ideally the buffer should grow according to the needed space, but that would be a pretty extensive change.

I can easily go over the default 4096 size when Janus generates an SDP with audio + video, ICE + TCP and the server has 4 IP addresses.


lminiero:
Makes sense, merging, thanks!

PR #336, opened 2015-09-21
build: fix compilation error

saghul:
janus.c: In function ‘janus_process_incoming_request’:
janus.c:1732:5: error: format not a string literal and no format arguments [-Werror=format-security]
     ret = janus_process_error(source, session_id, transaction_text, error, error_string);
     ^


saghul:
Bug was introduced by https://github.com/meetecho/janus-gateway/commit/c8421680cd4bb7436c511c03767d29c7995c4bb8

lminiero:
It should just be a warning, not an error. Anyway thanks for the fix!

saghul:
FWIW, I'm compiling it as a Debian package on Jessie, no extra flags specified manually.

PR #338, opened 2015-09-21
A couple more memory leak fixes concerning SIP re-invites. Also fixed…

MichaelB76:
… some per-session leaks where memory associated with the janus_sip_session structure was not begin freed.


lminiero:
Hi, thanks for this new patch! A couple of quick notes before I look into this:

1. please use tabs instead of whitespaces for tabulation;
2. have you already signed our CLA? we need this for all contributions, https://cla.conf.meetecho.com/janus/

I'll check the code as soon as I have some time.

MichaelB76:
Hi, I've now signed the CLA. Sorry about the formatting issues, I'll look again at the options in Netbeans.

MichaelB76:
I've signed the CLA now. Re-testing this the call to g_free(caller_text) needs to be changed to su-free(session->stack->s_home, caller_text), as url_as_string() allocates memory in the Sofia stack. Given the formatting issues as well it might be easier if you rejected this pull request and I sent you a new one, let me know what you want me to do please.

I suspect there are other issues like this regarding the Sofia memory allocations in the SIP plugin. The memory does get freed when the session is terminated but in-call memory usage keeps climbing. Valgrind is reporting blocks of 65MB in use when I leave a call open for a few days. I'll see what else I can find.

lminiero:
Thanks for signing! As to the patch, new PR or changing this one, whatever is easier for you works.

Yep, Sofia SIP memory management can get quite confusing, I remember fighting with it at the beginning too. Each session has its own home instance, which is where memory is supposed to be stored, and we do get rid of that. Have you checked closing cleanly with either Valgrind or AddressSanitizer to see if there's any hint on where the leaks may be? Unless those portions of memory are actually freed when we destroy the plugin itself.

MichaelB76:
Hopefully the formatting is OK now, let me know if there are any problems please.

I'm currently running Valgrind with the --leak-check=full and --show-leak-kinds=all option, so it doesn't hide memory that is still reachable such as that allocated to a Sofia handle. Clearing the home instance when the plugin session is closed does work, but if sessions are left up for a long time then the memory usage has been reaching tens of MB per-session, eventually killing Janus. I've found the last two problems by looking at the Janus output but I seem to be getting some spurious reports as well, possibly the Sofia memory management is confusing Valgrind. After adding the su_free() above about 70 leak reports seem to have vanished, none of which had any Janus code on the call stack.

The last two fixes seem to have reduced the memory useable considerably, here's the result of my last two overnight tests:

==38426== LEAK SUMMARY:
==38426==    definitely lost: 6,009 bytes in 165 blocks
==38426==    indirectly lost: 761 bytes in 32 blocks
==38426==      possibly lost: 7,052 bytes in 101 blocks
==38426==    still reachable: 93,656,208 bytes in 349,465 blocks
==38426==         suppressed: 0 bytes in 0 blocks


==42641== LEAK SUMMARY:
==42641==    definitely lost: 3,749 bytes in 72 blocks
==42641==    indirectly lost: 100 bytes in 10 blocks
==42641==      possibly lost: 6,412 bytes in 99 blocks
==42641==    still reachable: 26,450,861 bytes in 84,981 blocks
==42641==         suppressed: 0 bytes in 0 blocks

lminiero:
That sounds cool! I'll give it a check later in the day and in case everything's fine merge it, thanks!

lminiero:
Just tried and it crashes consistently for me, when closing the handle. Using gdb it seems to indicate it crashes here:

https://github.com/MichaelB76/janus-gateway/blob/master/plugins/janus_sip.c#L393

My guess is that it may be because you're freeing ```stack->session``` before freeing ```stack```. While we do allocate ```stack``` ourselves with a ```g_malloc0```, we don't do it for ```stack->session```, which may be something the stack handles itself. If I comment out the free of ```stack->session```, it doesn't crash, but not sure if it still causes leaks for you.

MichaelB76:
Ok thanks, I'll take a look at this.

MichaelB76:
It crashes for me too, at a guess I probably accidently ran it without rebuilding after making this change. It's a circular reference so it shouldn't need to free stack->session. I'll fix the pull request.

MichaelB76:
Sorry about that, hope that's OK now.

lminiero:
Thanks for the update, I'll look into that.

lminiero:
Looks good here, it doesn't crash for me anymore. Before merging, @saghul could I ask you to give a quick test on this branch too? Just asking as you're a heavier user of the plugin, and so it's more likely you'd encounter an issue, if any, while doing test calls.

saghul:
Will do, thanks for the heads up.

saghul:
Sorry for the delay folks. I'm building this right now and deploy it to our servers.

saghul:
No crashes with re-INVITEs. :+1:

lminiero:
Cool, thanks! Are you waiting for more tests to finish, or was this all you needed to know?

saghul:
Nope, that's all! Feel free to merge it.

lminiero:
Ok, merging it then.

MichaelB76:
Thanks @saghul and @lminiero. We're planning on running another stress test later this week. Prior to these fixes Janus died after a few days with about 12 clients connected, we'll see what happens this time.

saghul:
Fantastic, let us know!

lminiero:
@MichaelB76 please use the latest revision of Janus: I fixed some more things in the weekend and got rid of all malloc/calloc/free stuff to use the GLib versions instead (which should make it all more consistent).

MichaelB76:
@lminiero OK, will do thanks.

MichaelB76:
@lminiero , @saghul - We've been running Janus for a couple weeks now and it's behaving much better since these fixes. The CPU usage is no longer creeping up over time, probably there was an increasing overhead allocating memory in the Sofia stack when there were leaked blocks. Thanks for your help, this is a significant step towards turning Janus into something we could deploy.

lminiero:
That's great to know, thanks for the update! Your patches definitely helper fix the leaks you experienced, so you have to thank yourself too :smile:

MichaelB76:
I'll have to buy myself a pint the next time i see me. ;)

PR #339, opened 2015-09-21
ice: fix enabling ICE Lite mode

saghul:
'full-mode' is a constructor only property, setting it at runtime gives
the following warning and doesn't actually enable ICE Lite mode:

GLib-GObject-WARNING **: g_object_set_valist: construct property "full-mode" for object 'NiceAgent' can't be set after construction
libnice-DEBUG: Created NiceStream (1 created, 0 destroyed)
libnice-DEBUG: Created NiceComponent (1 created, 0 destroyed)
libnice-DEBUG: Created NiceComponent (2 created, 0 destroyed)
libnice-DEBUG: Agent 0xb2409000 : allocating stream id 1 (0xb2409e00)
libnice-DEBUG: Agent 0xb2409000 : In ICE-FULL mode, starting candidate gathering.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -2205,11 +2205,13 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 	handle->controlling = janus_ice_lite_enabled ? FALSE : !offer;
 	JANUS_LOG(LOG_INFO, "[%"SCNu64"] Creating ICE agent (ICE %s mode, %s)\n", handle->handle_id,
 		janus_ice_lite_enabled ? "Lite" : "Full", handle->controlling ? "controlling" : "controlled");
-	handle->agent = nice_agent_new(handle->icectx, NICE_COMPATIBILITY_DRAFT19);
+	handle->agent = g_object_new(NICE_TYPE_AGENT,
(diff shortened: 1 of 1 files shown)


saghul:
After the fix:

~~~~
libnice-DEBUG: Created NiceStream (1 created, 0 destroyed)
libnice-DEBUG: Created NiceComponent (1 created, 0 destroyed)
libnice-DEBUG: Created NiceComponent (2 created, 0 destroyed)
libnice-DEBUG: Agent 0xb4809000 : allocating stream id 1 (0xb4809f98)
libnice-DEBUG: Agent 0xb4809000 : In ICE-LITE mode, starting candidate gathering.
~~~~

lminiero:
Thanks for catching this. I guess this is something that changed in latter versions of libnice, as I definitely remember testing ICE Lite on my old Fedora 18 (which I think shipped 0.1.4). Have you checked whether creating the agent this way doesn't break it for older versions?

saghul:
Look slike it has been this way from the start: https://github.com/libnice/libnice/commit/355dafd011162bd6c5adceed7fe3125899ac1d51

The property was marked as constructoir only since 2007 AFAIS.

lminiero:
Thanks, I'll merge it right away then!

PR #335, opened 2015-09-21
ice: ignore non-matching host candidates when public IP is specified

saghul:
Before:

~~~~
v=0
o=- 3651822123 3651822123 IN IP4 127.0.0.1
s=Blink Pro 4.3.0 (MacOSX)
t=0 0
a=ice-lite
a=group:BUNDLE audio video
a=msid-semantic: WMS janus
m=video 1 RTP/SAVPF 97 102
c=IN IP4 81.23.228.146
a=mid:video
a=sendrecv
a=rtcp-mux
a=rtpmap:97 H264/90000
a=rtpmap:102 VP8/90000
a=fmtp:97 profile-level-id=42e01f;packetization-mode=1
a=ice-ufrag:M5F3
a=ice-pwd:7V8z14IB9er85rndvkT3MS
a=ice-options:trickle
a=fingerprint:sha-256 D2:B9:31:8F:DF:24:D8:0E:ED:D2:EF:25:9E:AF:6F:B8:34:AE:53:9C:E6:F3:8F:F2:64:15:FA:E8:7F:53:2D:38
a=setup:actpass
a=connection:new
a=zrtp-hash:1.10 e71010c50859245ef2005a8d1b59e811ea8f79daa7c3132d2fcb9965147e078c
a=ssrc:670684509 cname:janusvideo
a=ssrc:670684509 msid:janus janusv0
a=ssrc:670684509 mslabel:janus
a=ssrc:670684509 label:janusv0
a=candidate:13 1 udp 2013266431 81.23.228.146 58393 typ host
a=candidate:14 1 tcp 1019216383 81.23.228.146 0 typ host tcptype active
a=candidate:15 1 tcp 1015022079 81.23.228.146 41801 typ host tcptype passive
a=candidate:16 1 udp 2013266431 81.23.228.146 54218 typ host
a=candidate:17 1 tcp 1019216383 81.23.228.146 0 typ host tcptype active
a=candidate:18 1 tcp 1015022079 81.23.228.146 50295 typ host tcptype passive
a=candidate:19 1 udp 2013266431 81.23.228.146 59208 typ host
a=candidate:20 1 tcp 1019216383 81.23.228.146 0 typ host tcptype active
a=candidate:21 1 tcp 1015022079 81.23.228.146 34745 typ host tcptype passive
a=candidate:22 1 udp 2013266431 81.23.228.146 46982 typ host
a=candidate:23 1 tcp 1019216383 81.23.228.146 0 typ host tcptype active
[...]

Issue #329, opened 2015-09-17
regression: dtls does not connect for chrome-published streams


ploxiln:
A version of janus from June 11 (11a33f5cad1d71ca396c6f8da02f813d9654b0f8) works with both firefox and chrome both publishing and viewing.

Current master seems to work with firefox publishing, firefox viewing, and chrome viewing, but not chrome publishing. I've determined this by testing with the videoroom plugin (and turning off "publish" for one side to make it uni-directional between two browsers).

Maybe related to #328?

Further details:

  * ubuntu 14.04
  * openssl-1.0.2d
  * libsrtp-1.5.2
  * sofia-sip-1.12.11
  * direct connection possible between browsers on 172.30.1.1 and janus on 172.30.1.2
  * janus log: https://gist.github.com/ploxiln/88dc0a32dd6fdf5a486a

key log lines (these two identical lines emitted together, every 20 seconds or so)
```
[WARN] [4287276313] Still waiting for the DTLS stack for component 1 in stream 1...
[WARN] [4287276313] Still waiting for the DTLS stack for component 1 in stream 1...
```

This state persists indefinitely. Janus admin says this (sdp trimmed):
```
{
    "session_id": 3447375110,
    "handle_id": 4287276313,
    "current_time": 20794988550,
    "plugin": "janus.plugin.videoroom",
    "plugin_specific": {
        "type": "publisher",
        "room": 1234,
        "id": 1679004654,
        "display": "usr1",
        "media": {
            "audio": 1,
            "video": 1,
            "data": 0
        },
        "destroyed": 0
    },
    "flags": {
        "got-offer": 1,
        "got-answer": 1,
        "processing-offer": 0,
        "starting": 1,
        "ready": 0,
        "stopped": 0,
        "alert": 0,
        "bundle": 1,
        "rtcp-mux": 1,
        "trickle": 1,
        "all-trickles": 1,
        "trickle-synced": 0,
        "data-channels": 0,
        "has-audio": 1,
        "has-video": 1,
        "plan-b": 0,
        "cleaning": 0
    },
    "agent-created": 20531834105,
    "ice-mode": "full",
    "ice-role": "controlled",
    "sdps": {
        "local": "v=0\r\no=- 20531862462 20531862461 IN IP4 127.0.0.1\r\ns=Demo Room\r\n...",
        "remote": "v=0\r\no=- 3285441602913217416 2 IN IP4 127.0.0.1\r\ns=-\r\nt=0 0\r\n..."
[...]

PR #341, opened 2015-09-22
ICE 'enforce' list

lminiero:
Added new ICE 'enforce' list, to specify the only interfaces to use for gathering candidates, to address the point made in #335 by @saghul.

This acts as a complementary feature to the existing ICE 'ignore' list, over which it takes precedence. While the ignore list allowed you to specify partial/full IP addresses or interface names to ignore for the purpose of ICE gathering, this new enforce list allows you to tell Janus which are the only interfaces you want to be used for that. As such, it only accepts interface names and not IP addresses, partial or not: if an enforce list is provided and an interface name is not listed, the interface is NOT used for any ICE candidate. When the enforce list is active, the ignore list is (no pun intended) ignored.

This patch also removes the old (and useless) ```-e``` flag to set the public IP. Now the public IP is automatically set to either the autodetected local IP address, or whatever STUN gave us back if configured. The public IP is then only used in some parts of the SDP, but not for candidates.

Seems to be working fine here, feedback welcome!

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -134,6 +134,12 @@ void janus_ice_debugging_disable(void) {
 }
 
 
+/* NAT 1:1 stuff */
+static gboolean nat_1_1_enabled = FALSE;
+void janus_ice_enable_nat_1_1(void) {
+	nat_1_1_enabled = TRUE;
+}
+
 /* Interface/IP enforce/ignore lists */
 GList *janus_ice_enforce_list = NULL, *janus_ice_ignore_list = NULL;
 janus_mutex ice_list_mutex;
@@ -1826,6 +1832,11 @@ void janus_ice_candidates_to_sdp(janus_ice_handle *handle, char *sdp, guint stre
 	NiceAgent* agent = handle->agent;
 	/* adding a stream should cause host candidates to be generated */
 	char *host_ip = NULL;
+	if(nat_1_1_enabled) {
+		/* A 1:1 NAT mapping was specified, overwrite all the host addresses with the public IP */
+		host_ip = janus_get_public_ip();
+		JANUS_LOG(LOG_VERB, "[%"SCNu64"] Public IP specified and 1:1 NAT mapping enabled (%s), using that as host address in the candidates\n", handle->handle_id, host_ip);
+	}
 	GSList *candidates, *i;
 	candidates = nice_agent_get_local_candidates (agent, stream_id, component_id);
 	JANUS_LOG(LOG_VERB, "[%"SCNu64"] We have %d candidates for Stream #%d, Component #%d\n", handle->handle_id, g_slist_length(candidates), stream_id, component_id);
--- a/ice.c
+++ b/ice.c
@@ -138,22 +138,22 @@ void janus_ice_debugging_disable(void) {
 GList *janus_ice_enforce_list = NULL, *janus_ice_ignore_list = NULL;
 janus_mutex ice_list_mutex;
 
-void janus_ice_enforce_interface(const char *interface) {
-	if(interface == NULL)
+void janus_ice_enforce_interface(const char *ip) {
+	if(ip == NULL)
 		return;
-	/* The enforce list only handles interfaces */
+	/* Is this an IP or an interface? */
 	janus_mutex_lock(&ice_list_mutex);
-	janus_ice_enforce_list = g_list_append(janus_ice_enforce_list, (gpointer)interface);
+	janus_ice_enforce_list = g_list_append(janus_ice_enforce_list, (gpointer)ip);
 	janus_mutex_unlock(&ice_list_mutex);
 }
-gboolean janus_ice_is_enforced(const char *interface) {
-	if(interface == NULL || janus_ice_enforce_list == NULL)
+gboolean janus_ice_is_enforced(const char *ip) {
+	if(ip == NULL || janus_ice_enforce_list == NULL)
 		return false;
 	janus_mutex_lock(&ice_list_mutex);
 	GList *temp = janus_ice_enforce_list;
 	while(temp) {
 		const char *enforced = (const char *)temp->data;
-		if(enforced != NULL && strstr(interface, enforced)) {
+		if(enforced != NULL && strstr(ip, enforced)) {
 			janus_mutex_unlock(&ice_list_mutex);
 			return true;
 		}
@@ -2300,14 +2300,9 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 			/* We only add IPv6 addresses if support for them has been explicitly enabled (still WIP, mostly) */
 			if(family == AF_INET6 && !janus_ipv6_enabled)
 				continue;
-			/* Check the interface name first, we can enforce/ignore that as well: enforce list always has precedence */
-			if(janus_ice_enforce_list != NULL) {
-				if(ifa->ifa_name != NULL && !janus_ice_is_enforced(ifa->ifa_name))
-					continue;
-			} else {
-				if(ifa->ifa_name != NULL && janus_ice_is_ignored(ifa->ifa_name))
-					continue;
-			}
+			/* Check the interface name first, we can ignore that as well: enforce list would be checked later */
+			if(janus_ice_enforce_list == NULL && ifa->ifa_name != NULL && janus_ice_is_ignored(ifa->ifa_name))
+				continue;
 			s = getnameinfo(ifa->ifa_addr,
 					(family == AF_INET) ? sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6),
 					host, NI_MAXHOST, NULL, 0, NI_NUMERICHOST);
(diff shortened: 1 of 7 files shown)


ice.c
```diff
@@ -668,13 +696,17 @@ int janus_ice_set_stun_server(gchar *stun_server, uint16_t stun_port) {
 	StunMessageReturn ret = stun_message_find_xor_addr(&msg, STUN_ATTRIBUTE_XOR_MAPPED_ADDRESS, (struct sockaddr *)&address, &addrlen);
 	JANUS_LOG(LOG_VERB, "  >> XOR-MAPPED-ADDRESS: %d\n", ret);
 	if(ret == STUN_MESSAGE_RETURN_SUCCESS) {
-		JANUS_LOG(LOG_INFO, "  >> Our public address is %s\n", inet_ntoa(address.sin_addr));
+		char *public_ip = inet_ntoa(address.sin_addr);
```
saghul: maybe use inet_ntop for eventual IPv6 compatibility?

saghul:
Thanks for working on this! Sadly I need to enforce an IP address, not an interface. In IPv4 this approach should work, since you can only have an IP address per interface (so I'd enforce the alias), but not so on IPv6. Why can't this option behave like the ignore list? I can specify IP addresses there.

ice.c
lminiero: You've picked the only part of the code that was already there :smile: 
Sure, I'll fix this in future commits.

lminiero:
> Why can't this option behave like the ignore list? I can specify IP addresses there.

Because the enforce list works in a different way. If an interface is not in the enforce list, you don't even get to the IPs as of now. Allowing you to go further nonetheless would make the interface filtering useless. I'll have to check if this can be changed without messing too much with the code.

lminiero:
This new commit allows IP addresses to be set in the enforce list too.

saghul:
Fantastic, thanks! I'll give it a go tomorrow.

ploxiln:
I use the public_ip option, btw. It seems like the perfect option for ec2's 1-to-1 NAT. (Even though I also configure a useless stun server on the janus side as well, currently, to appease the ICE candidate timing demons.)

lminiero:
> I use the public_ip option, btw. It seems like the perfect option for ec2's 1-to-1 NAT. (Even though I also configure a useless stun server on the janus side as well, currently, to appease the ICE candidate timing demons.)

If you recall the public IP setting was exctly what caused Chrome to fail in your setup at the time, something that could be fixed by relying on a STUN server even in that case. Besides, my feeling is that the public_ip setting was too prone to configuration mistakes, and could lead to issues like the one saghul experienced, with tons of candidates from other interfaces being overwritten. STUN is the safe way to go, and will set the candidate address to the proper address anyway.

ploxiln:
Theoretically, no STUN is ever needed anywhere, if one side has a stable public address and open ports. Janus-Gateway is often in exactly that situation (even in EC2).

STUN adds complexity that shouldn't be there, it adds extra candidates, adds extra time to send and test candidates, adds extra log lines to obscure what's happening and what's going wrong. That's why I want to do without. What causes Chrome to fail is bugs in Chrome or bugs in libnice or bugs in janus, not lack of a stun server itself.

The stun server setting just affects the number and order of candidates, which has alchemical-like incomprehensible effects on the extremely complicated ICE implementations. In fact, I have a case in #329 where removing the stun server on the janus side *makes Chrome work* where it otherwise did not. I'm not blaming anything in particular, just showing that mixing the pot produces pseudorandom results, like a hash function. I don't like that, I prefer logic. And logic says, no STUN server is ever needed for any of my use cases.

ploxiln:
Also, just to clarify, the stun server I currently configure on the janus side when running in ec2 with the public_ip option, is to make firefox work, not chrome.

saghul:
I wonder if libnice gets confused if you end up using a candidate which was for a different IP address than the plubic one which maps 1-to-1.

Right now, as I mentioned in #335, the public_ip option is broken if you have more than interface, more so if you're actually not behind NAT.

Another option which we could turn on in libnice would be UPnP. That should be fast, but I'm not sure it works on EC2.

lminiero:
> Another option which we could turn on in libnice would be UPnP. That should be fast, but I'm not sure it works on EC2.

We explicitly disable UPnP because I remember that at the time (first months of Janus' life) this triggered some issues. Since we don't make use of it, disabling it was a no-brainer.

ploxiln:
UPnP (for port forwarding setup I presume?) is only for home routers, and is not implemented anywhere in AWS. But it's not needed there, as EC2 instances can have a public ip address (and usually do) which forwards all packets on all ports to the matching port on the instance (which are then blocked if you haven't un-blocked them in your security groups).

When running janus in ec2 I have it configured to ignore all but the one external interface. But even if there were multiple interfaces, each generating an ICE candidate with an ip address that only works for one of them, that should be OK, because one would be discovered to work. I actually do have this situation in a testing VM, and it works. It would probably still be a good idea to de-duplicate the ice candidates though.

ploxiln:
I don't mind you renaming the option. But it seems like the new implementation is a lot more complicated - two variables, getters and setters, etc.

Also consider this funny aspect of local_ip:

```
    [nat]
        stun_server: stun.getperch.com
        stun_port: 3478
        nice_debug: false
        ice_lite: false
        ice_tcp: false
        ice_ignore_list: vmnet,docker,lo,10.0.
Debug/log level is 5
Debug/log timestamps are disabled
Debug/log colors are enabled
Adding 'vmnet' to the ICE ignore list...
Adding 'docker' to the ICE ignore list...
Adding 'lo' to the ICE ignore list...
Adding '10.0.' to the ICE ignore list...
Selecting local IP address...
Autodetecting local IP...
Using 10.0.2.15 as local IP...
[WARN] Token based authentication disabled
RTP port range: 10000 -- 29999
Initializing ICE stuff (Full mode, ICE-TCP candidates disabled, IPv6 support disabled)
ICE port range: 10000-29999
STUN server to use: stun.getperch.com:3478
ICE handles watchdog started
  >> 52.2.20.148:3478
  >> Sent 20 bytes 52.2.20.148:3478, waiting for reply...
  >> Got 92 bytes...
  >> XOR-MAPPED-ADDRESS: 0
  >> Our public address is 98.7.29.101
```

specifically:

```
Adding '10.0.' to the ICE ignore list...
Selecting local IP address...
Autodetecting local IP...
Using 10.0.2.15 as local IP...
```

I guess it doesn't hurt? But what is it for?

And I like the idea of setting public_ip for my VM, because the external ip it figures out via STUN is not the correct IP to use to talk to a browser running on my host. But, just setting STUN doesn't seem to hurt, I get the correct candidates anyway:

```
                    "local-candidates": [
                        "1 1 udp 2013266431 172.30.1.2 10000 typ host\r\n",
                        "2 1 udp 1677721855 98.7.29.101 51024 typ srflx raddr 172.30.1.2 rport 10000\r\n"
                    ],
```

screen shot 2015-09-24 at 01 52 34

Everything is OK, just a bit confusing. Go ahead and rename. Though if it were me I might choose a name like "experts_only_public_ip"

lminiero:
The reason why it's two separate variables is because we now also try to autodetect the public IP now, if STUN is configured, to use it for something else than candidates. Previously if you didn't set it via the setting you wouldn't have any. Specifically, we set it in some fields of the SDP that are not strictly needed for establishing connectivity, in place of whatever local_ip we identify. The additional variable I introduced is to instruct Janus to _also_ use the public IP to rewrite all host candidates, which is what it did before.

I agree that it's weird the autodetected local_ip ends up being one of the interfaces you disabled: as anticipated, it's not really needed if not to fill some unrelated SDP fields, but it is indeed confusing, I'll fix that.

ploxiln:
It makes sense now that you explain.

I'm using janus master now, without this branch, and I just tired setting up a scenario with STUN server enabled and no public_ip set, where the default route interface is the one used to communicate with the browser, so the host candidate with that local ip would indeed be what was needed. The theory was that this replacement would break things. But it still didn't seem to replace. So to my disappointment, it worked fine...

lminiero:
> EDIT2: "but use for stuff other than candidates" ah yes so this still makes sense

Yep, like the o- and c-lines in the SDP.
Does this mean we can safely merge the PR, as it works for you too?

saghul:
> Yep, like the o- and c-lines in the SDP.

Maybe we could use something else to avoid confusion? The o= line is not used for routing, so we can use whatever IP. RFC 5245 requires the c= line to contain the IP of the default candidate (so it must actually point to some candidate) but I'm not sure if in WebRTC they chose to ignore that.

lminiero:
I don't think it's being used: Chrome sets it to ```0.0.0.0``` IIRC.

lminiero:
Firefox inserts a valid IP, instead, so that's why I wanted to do that in Janus too. The public IP, whether autodetected via STUN or passed via the new setting, definitely is one of the addresses in the provided candidates, and so should be fine.

saghul:
> Firefox inserts a valid IP, instead, so that's why I wanted to do that in Janus too. The public IP, whether autodetected via STUN or passed via the new setting, definitely is one of the addresses in the provided candidates, and so should be fine.

:+1:

ploxiln:
:+1:

lminiero:
Ok, thanks for the feedback and useful discussions guys! Merging.

mattvv:
Hey guys,

Getting a bunch of issues with

```
[ERR] [ice.c:janus_ice_cb_component_state_changed:1309] [1817927522] ICE failed for component 1 in stream 1...
```
on ec2. I'm almost certain it's due to my config. I'm on AWS and I have coturn on the same server. Can anyone paste a config that I can take a look at as an example?

ploxiln:
ICE components can fail for many reasons. If something is not working, the result is usually ICE component failed.

Co-locating the STUN / TURN server on the same server as janus is a bad idea. If you only have one server, you're small enough to go ahead and just use google's STUN server: `stun.l.google.com:19302` (default in janus.js).

You'll notice in the janus.cfg.sample, under the `[nat]` section, there's the `nat_1_1_mapping` setting, this is for environments just like ec2. It's not always needed because the STUN server can tell janus what its real external IP address is (assuming janus does not contact the STUN server via its ec2-internal address). But results will vary along with browsers, home networks, and versions of openssl and libnice.

I currently both specify the nat_1_1_mapping, and use a separate STUN server (which I make sure to contact via its external IP). Good luck.

lminiero:
Yep, and besides you should also make sure the security group on EC2 is configured so that Janus can receive packets on the ports it allocates for ICE: outgoing packets always work, incoming packets are subject to the firewalling the security group enforces.

mattvv:
Thanks @ploxiln and @lminiero,

I changed to using googles stun servers and no longer using TURN, This appears to help significantly in connectively for all users (especially those that were having trouble connecting before). I also use the nat_1_1_mapping ip. 

In terms of the EC2 Security group. Is there a list of ports that I should open by default for ICE somewhere? (I'm not using icelite or the other ice flag).

lminiero:
> In terms of the EC2 Security group. Is there a list of ports that I should open by default for ICE somewhere? (I'm not using icelite or the other ice flag).

ICE ports are dynamically chosen for each call. The best you can do is costrain them to a min-max in the settings, and open that range in EC2, otherwise you have to open them all. You need a version of libnice that supports the min/max range, though: if your version doesn't have it, you'll see a warning when launching Janus.

PR #342, opened 2015-09-23
fix read cert_pem for REST https

fpn:
When configuring https for the REST api the cert_pem config entry wasn't read. The last config value was used instead.

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -399,8 +399,11 @@ int janus_http_init(janus_transport_callbacks *callback, const char *config_path
 			}
 		}
 		/* Do we also have to provide an HTTPS one? */
-		char *server_pem = (char *)item->value;
-		char *server_key = (char *)item->value;
+		char *server_pem = NULL;
+		item = janus_config_get_item_drilldown(config, "certificates", "cert_pem");
+		if(item && item->value)
+			server_pem = (char *)item->value;
+		char *server_key = NULL;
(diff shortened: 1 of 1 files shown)


lminiero:
Yep, sorry about that, copy/paste error from the code we have in ```janus.c``` in ```master```. Anyway, your patch also sets '``server_key``` to NULL when initializing, while the idea was to willingly set it to the value of ```server_pem``` by default, and if a key value is provided, use that one instead:

	item = janus_config_get_item_drilldown(config, "certificates", "cert_pem");
	server_pem = (char *)item->value;
	server_key = (char *)item->value;
	item = janus_config_get_item_drilldown(config, "certificates", "cert_key");
	if(item && item->value)
		server_key = (char *)item->value;

The rationale for that, IIRC, was that you could also just pass the certificate as the key to libssl. I realize it doesn't make much sense, but I seem to recall this worked back then when we made the change.

I think that's a wrong behaviour, though, and that we should indeed make sure a key has been provided, and fail otherwise, as your patch does. I'll update ```janus.c``` in ```master``` as well.

Thanks!

lminiero:
As a side note, have you signed our CLA? We require this for all contributions: https://cla.conf.meetecho.com/janus/

fpn:
yes - have signed the CLA

lminiero:
Thanks! I already fixed the code in ```master```, so I'll merge your contribution to fix the ```modular-transports``` too.

Issue #136, opened 2015-01-16
DTLS1_READ_BYTES:tlsv1 alert decrypt error


LetsVape:
Hello,

I've been running the janus-gateway for 28 hours yesterday and ran into the following error:
[31m[ERR][0m [dtls.c:janus_dtls_srtp_incoming_msg:311:] Handshake error: error:1410241B:SSL routines:DTLS1_READ_BYTES:tlsv1 alert decrypt error

Last time I ran into this issue it was after 21 hours so it doesn't seem to be a consistent interval

When someone joins and this error pops up once it will keep popping up (on the server) for every new person that joins, everyone that previously joined the videoroom will be able to continue to view the feeds but new people joining will only see publisher names, no video and the gateway will throw the above error for everyone that joins, I will have to restart the gateway to allow people to join again

I've included a debug-level=5 log which includes the first error and then 2 more, I've stopped there for pastebin limits but have a longer log if you prefer, I hope this helps, please let me know if you'd like any other logs/info

http://pastebin.com/qKaAXQcD

Thanks

lminiero:
No idea what that is, honestly, especially considering it only seems to happen after a long time and not for every call (which probably excludes certificate or openssl errors, for instance). Unfortunately line 311 doesn't specify the handle ID this is happening on, so I'll have to modify the code to enrich the related debug.

What's the CPU/mem status after those ~20 hours? Just wanted to exclude memory leaks or something like that as a potential cause of failure here.

lminiero:
Can you try again with the latest push I made? It should print the handle-related info as well.

LetsVape:
Thanks a lot for your reply and adding the extra debug logging
I will rebuild from the latest commit and wait get another log of the error

I see you added some info for LOG_HUGE, should I include that in the log too, (Is that debug-level 7?)
I went with 5 as the file-size was still manageable this way, 10 mins on 7 will give a 2GB file or so

The CPU was only around 150%-200% (8 cores) and the memory usage was around %15-%20 (8GB) (Linux also cached ~5GB but that is to be expected with the I/O activity)
No other processes were consistently using over %1 CPU or %MEM during this time so resources don't seem to be an isse

Just to clarify as you said "after a long time and not for every call"
It does happen for every call after this error showed up once, I don't see any successful DTLS after this.
But it's not like anyone is persistently unable to connect and causing this error, after a refresh everyone can join


Thanks for you help

lminiero:
[...]

PR #343, opened 2015-09-24
Add optional BoringSSL support via configure

lminiero:
This pull request adds optional BoringSSL support via a new flag in the configure script, ```--enable-boringssl```. If enabled, it looks for BoringSSL in a specific folder and updates the Makefile accordingly to make use of it.

The README has also been updated with instructions on how to build BoringSSL, partly derived from info gathered on #136 and partly coming from what I needed to get it working.

Feedback welcome on whether this is indeed working for everybody.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,37 @@ is quite straightforward:
 * *Note:* you may need to pass --libdir=/usr/lib64 to the configure
 script if you're installing on a x86_64 distribution.
 
+If you want to make use of BoringSSL instead of OpenSSL for any reason
+(read [here](https://github.com/meetecho/janus-gateway/issues/136) for
+some background on this), you'll have to manually install a specific
+version of the library to a specific location. Use the following steps:
+
+	git clone https://boringssl.googlesource.com/boringssl
+	cd boringssl
+	# We need a specific revision
+	git fetch origin 12fe1b25ead258858309d22ffa9e1f9a316358d7
+	# Don't barf on errors
+	sed -i s/" -Werror"//g CMakeLists.txt
+	# Build
+	mkdir -p build
+	cd build
+	cmake -DCMAKE_CXX_FLAGS="-lrt" ..
+	make
+	cd ..
+	# Install
+	sudo mkdir -p /opt/boringssl
+	sudo cp -R include /opt/boringssl/
+	sudo mkdir -p /opt/boringssl/lib
+	sudo cp build/ssl/libssl.a /opt/boringssl/lib/
+	sudo cp build/crypto/libcrypto.a /opt/boringssl/lib/
+
+Once the library is installed, you'll have to pass an additional
+```--enable-boringssl``` flag to the configure script, as by default
+Janus will be build assuming OpenSSL will be used. If you were using
+OpenSSL and want to switch to BoringSSL, make sure you also do a
+```make clean``` in the Janus folder before compiling with the new
+BoringSSL support.
+
 For what concerns usrsctp, which is needed for Data Channels support, it
 is usually not available in repositories, so if you're interested in
 them (support is optional) you'll have to install it manually. It is a
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ version of the library to a specific location. Use the following steps:
 	git clone https://boringssl.googlesource.com/boringssl
 	cd boringssl
 	# We need a specific revision
-	git fetch origin 12fe1b25ead258858309d22ffa9e1f9a316358d7
+	git checkout 12fe1b25ead258858309d22ffa9e1f9a316358d7
 	# Don't barf on errors
 	sed -i s/" -Werror"//g CMakeLists.txt
 	# Build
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -79,6 +79,37 @@ is quite straightforward:
...
 
+If you want to make use of BoringSSL instead of OpenSSL for any reason
+(read [here](https://github.com/meetecho/janus-gateway/issues/136) for
+some background on this), you'll have to manually install a specific
+version of the library to a specific location. Use the following steps:
+
+	git clone https://boringssl.googlesource.com/boringssl
+	cd boringssl
+	# We need a specific revision
+	git fetch origin 12fe1b25ead258858309d22ffa9e1f9a316358d7
```
ploxiln: Probably better as `git checkout 12fe1b25ead258858309d22ffa9e1f9a316358d7`

lminiero: Thanks for the heads-up, I'll change that.

masahji:
Thanks @lminiero! Everything built just fine for us; however, after checking links with ldd we could still see janus linked to the system libssl. So we build boringssl with shared libs instead (`cmake -DBUILD_SHARED_LIBS=1 -DCMAKE_CXX_FLAGS="-lrt" ..`) and then added /opt/boringssl/lib as an ldconfig. 

It appears as though your instructions call for building boringssl in statically. But I think we'd have to modify autoconf to not link in libssl when building statically. 

Thanks

lminiero:
@masahji actually all my attempts to use a shared version of the library failed somehow, or at least I couldn't find a way to reliably force that upon the stock OpenSSL, so I eventually resigned to using the static build (which is that BoringSSL tries to build by default, anyway). I also noticed ldd still listing the OpenSSL shared objects, but the ```nm janus | grep DTLSv1_2``` confirms BoringSSL is used apparently.

ploxiln:
btw, if you do end up building shared libs and putting them somewhere besides /usr/lib/, the best way to make janus use them (at run time) is to add to the "rpath", the runtime library path.

As I mentioned in https://github.com/meetecho/janus-gateway/pull/325#issuecomment-140401007 I build with `LDFLAGS="-L/usr/local/lib -Wl,-rpath=/usr/local/lib"`, and you can see the result of `-Wl,-rpath=...` in the output from "readelf", and "ldd" will confirm:

```
$ readelf -d /usr/local/bin/janus

Dynamic section at offset 0x5ad00 contains 38 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libnice.so.10]
 0x0000000000000001 (NEEDED)             Shared library: [libgobject-2.0.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libglib-2.0.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libmicrohttpd.so.10]
 0x0000000000000001 (NEEDED)             Shared library: [libjansson.so.4]
 0x0000000000000001 (NEEDED)             Shared library: [libssl.so.1.0.0]
 0x0000000000000001 (NEEDED)             Shared library: [libcrypto.so.1.0.0]
...
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000f (RPATH)              Library rpath: [/usr/local/lib]
 0x000000000000000c (INIT)               0x407fb0
 0x000000000000000d (FINI)               0x44b9c4
...
$ ldd /usr/local/bin/janus
	linux-vdso.so.1 =>  (0x00007ffc4d896000)
	libnice.so.10 => /usr/lib/x86_64-linux-gnu/libnice.so.10 (0x00007f464fdbd000)
	libgobject-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007f464fb6c000)
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f464f863000)
	libmicrohttpd.so.10 => /usr/lib/x86_64-linux-gnu/libmicrohttpd.so.10 (0x00007f464f64c000)
	libjansson.so.4 => /usr/lib/x86_64-linux-gnu/libjansson.so.4 (0x00007f464f440000)
	libssl.so.1.0.0 => /usr/local/lib/libssl.so.1.0.0 (0x00007f464f1c6000)
	libcrypto.so.1.0.0 => /usr/local/lib/libcrypto.so.1.0.0 (0x00007f464ed78000)
...
```

("..." is where I trimmed the output.) You could do the same with "/opt/boringssl/lib"; you can have multiple rpath additions.

lminiero:
I think I tried playing with rpath for a shared build of BoringSSL as well, using the info you provided in that other discussion, but I still couldn't get it to work, can't remember why. I'll give it another try next week.

ploxiln:
Yeah this stuff can be fiddly and confusing. I wonder if it was due to the linker pulling in the system openssl library on behalf of another shared library before pulling in boringssl on behalf of janus proper... Shared libraries can also have an rpath of their own. So libsophia-sip / libsrtp / etc should also be built against the openssl lib you want to be used.

Now that I think about it, it's possible that janus was statically linking in boringssl stuff while libsophia-sip etc. were dynamically linking to the system openssl.

lminiero:
Merging for now as it does indeed work for those who are interested in this, we can worry about shared vs. static and other things later.

PR #344, opened 2015-09-28
sip: fixed reporting re-INVITEs as missed calls

saghul:


lminiero:
Will Sofia simply reply with the SDP it had previously?

saghul:
Yep, that's what my tests showed.

lminiero:
Ok, I guess it's safe to merge then, thanks!

saghul:
It does increase the SDP version number, in order to behave like a good citizen :-)

lminiero:
:+1:

PR #345, opened 2015-09-29
sip: manually handle re-INVITEs and reject them with 488

saghul:


saghul:
We could leave it as it was, but explicit is better than implicit!

Tested for incoming and outgoing. :+1:

lminiero:
Makes sense, thanks for this. As a side note, please use tabs for indentation and not spaces.

saghul:
damn, let me fix that real quick!

saghul:
Should be kosher now. Sorry, I have my editor configured for The One True Indenting Model (R) and I sometimes forget to change it for Janus ;-)

lminiero:
Thanks! Sorry for nit-picking :smile:

lminiero:
Hey, I really _AM_ lazy! http://www.perlmonks.org/?node_id=158886

saghul:
I'm not sure I can trust a Perl website! :sunglasses:

PR #346, opened 2015-09-29
Use editorconfig to help editors use the preferred style

saghul:


lminiero:
You really did take my crazyness seriously :smile: 
I didn't know about this file, that's interesting!

saghul:
I've been neglecting to try this stuff for a while, I guess today was a day as good as any :-) It does help, now my editor (The One True Editor, in case you were wondering) automagically configures to match the project style :-)

On a more serious note: I prefer indenting with spaces, but I prefer consistency above any of my style choices, so if the project author (you) chose tabs, so be it!

lminiero:
How many spaces? Just so that I can match you to one of the profiles from that Perl forum...

saghul:
4 spaces!

lminiero:
Spiffy!

PR #347, opened 2015-09-29
sip: fix handling subsequent incoming calls

saghul:
In 1cefe04 yours truly relied on the fact that the Sofia handle would be
reset after each call terminates. Turns out I was wrong. This patch
fixes that by null-ing the handle once the call is terminated. This is
generally a good practice since the handle is dead anyway, but more
importantly, it makes the logic work properly.


saghul:
Sorry about that!

lminiero:
Thanks for the fix!

PR #348, opened 2015-09-29
sip: fix SDP parser leak when handling reinvites

saghul:
Thanks @MichaelB76 for noticing.


lminiero:
Thanks @saghul and @MichaelB76, merging.

PR #349, opened 2015-09-29
off by one buffer overflow

fpn:
Found this off by one bug using address sanitizer.
You may want to check the callers of janus_http_random_string to make sure that the string is long enough e.g. if you want 12 chars you need to provide a 13 char array because of the null termination.

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -202,7 +202,7 @@ static void janus_http_random_string(int length, char *buffer) {
 			int key = rand() % l;
 			buffer[i] = charset[key];
 		}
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for spotting this, really a rookie mistake! I'll check if we have something like this in other parts of the code in ```master``` too, although that shouldn't be the case.

ploxiln:
btw that loop still writes a random char to that last byte in the buffer which is then overwritten with NULL. So just a teeny tiny bit wasteful.

lminiero:
Fair point, I'll clean that up in the next few days.

PR #350, opened 2015-09-29
protect access to freed janus_websockets_client with old_wss_mutex

fpn:
Hi,

Address Sanitizer implicated the client-> context accesses when I kill my load websocket client and janus-gateway cleans up.

This patch tries to address the issue by using the old_wss_mutex to protect against accessing freed instances of the websocket_client / transport.

Personally I feel that it may be better to use reference counts instead of keeping a list of deleted references around.

best regard,
Florian
ps. named this branch wrong not a double free

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -592,24 +592,30 @@ int janus_websockets_send_message(void *transport, void *request_id, gboolean ad
 		g_free(message);
 		return -1;
 	}
-	janus_websockets_client *client = (janus_websockets_client *)transport;
-	if(!client->context || !client->wsi) {
-		g_free(message);
-		return -1;
-	}
-	/* Make sure this is not related to a closed WebSocket session */
+	/* Make sure this is not related to a closed /freed WebSocket session */
 	janus_mutex_lock(&old_wss_mutex);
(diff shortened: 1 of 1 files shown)


fpn:
@lminiero do you want me to open issues for these or is just the pull requests is ok?

lminiero:
Pull requests are fine, thanks! Especially as you're raising an issue and proposing a solution at the same time :wink: 
You're right about reference counting, and there's an issue open that discussed about a need for something like this. Anyway, this would require some bit of refactoring, especially if we were to conceive a base object all the stuff we use in Janus would be built upon. Besides, everything in Janus is heavily asynchronous, which makes things more complicated to handle in that respect. Short term solutions like the one you propose are definitely fine for now.

I'll check the code later, but please (and I know I'm being nitty here, apologies for that) use tabs and not whitespaces for indentation. We have a new ```.editorconfig``` file as part of the project that can automatically help with that, if your editor supports it.

fpn:
Updated the patch - hope it is ok now - using the "editorconfig/editorconfig-vim" plugin now.

lminiero:
Thanks! I'll look at that when I get back to the office tomorrow.

lminiero:
Merging, thanks!

PR #356, opened 2015-10-08
Enhancement: also report display-name of caller when present.

tuijldert:
With this change, a client can also show the display name of who is calling (json will also contain a `displayname` of the caller when present in the `INVITE`).

Please review and commit.

TIA,
Tom.


lminiero:
Can you change this check:

    if (sip->sip_from->a_display) {

to this?

    if (sip->sip_from && sip->sip_from->a_display) {

I don't think it's ever going to happen that there's no From (IIRC it's not even permitted) but since we usually always have multiple checks anyway to try and avoid accessing invalid data, better safe than sorry...

tuijldert:
You're fast :)
Done.

lminiero:
Thanks :smile: 
Merging!

PR #359, opened 2015-10-12
Send a FIR to the new RTP forward publisher

amnonbb:

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1407,6 +1407,18 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 			video_handle = janus_rtp_forwarder_add_helper(publisher, host, video_port, 1);
 		}
 		janus_mutex_unlock(&videoroom->participants_mutex);
+		/* Send a FIR to the new RTP forward publisher */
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! I guess this was indeed needed.

Can you remove the extra space before each of those lines? It's messing up the indentation.
Please also sign our [CLA](https://cla.conf.meetecho.com/janus/), as we require that for all contributions.

amnonbb:
Done

lminiero:
Thanks for the quick fix. Just looked into the patch a bit more and I think we just need a further check, that is whether the RTP forwarder actually involves video or not: as it is, this would send a FIR/PLI even for an audio only forwarder, which is probably not what we want.

Can you move that code in the ```if(video_handle > 0) {``` that is a few lines below, so that it only kicks in when we actually created a video handle?

amnonbb:
i'll test it..

amnonbb:
Thanks for review. I tested, it's work.
Worth to know that 1 from 20-30 switches may miss keyframe.
but it's may gstreamer pipeline problem.
Anyway without FIR request in gstreamer out it's look horrible on every switch.

lminiero:
Thanks, merging!

Issue #337, opened 2015-09-21
Add support for Microsoft EDGE


simongymer-speakerbus:
Very early days, but would be good to look into supporting Microsoft EDGE once it's moved from Insider Preview to mainstream.

lminiero:
Of course we'd like to do that, but I don't use Windows and most definitely won't have access to a Windows 10. One of my colleagues might attempt to do that in the future but there's no plan on a date. It's not going to be super-easy, as we'll need some shim on the JavaScript side that translates whatever ORTC generates to SDP and viceversa, since Janus is based on SDP right now.

That said, if any of you guys happen play with this and feel like sharing a pull request, we'll be glad to look at that!

lminiero:
First attempt here, #360

lminiero:
PR merged.

PR #360, opened 2015-10-14
Edge support

lminiero:
This is a first attempt at getting Edge and Janus to talk to each other, as discussed in #337. Few changes were needed on the Janus side, just a couple of simple bug fixes that have already been added to master.

Long story short (read below for more gory details) this makes Edge mostly works with Janus, but audio only: video is not supported (Edge only supports X-H264UC, which apparently is not compliant with the H264 Firefox uses) and data channels are not available either in Edge. As such, video and data channels are automatically disabled by ```janus.js``` when Edge is detected. I still get some failed PeerConnections due to ICE failures, but most of the times I managed to get this working fine with most plugins, that is:

* echo test (audio only)
* streaming plugin (audio only mountpoints)
* audio bridge
* SIP calls (audio only)
* voicemail and record&play

I didn't test the videoroom plugin for audio only rooms and the videocall plugin for audio only calls, but I don't see why those shouldn't work.

Coming to the abovementioned gory details, it mostly is a heavy modification to the ```adapter.js```, and a couple of conditional checks in ```janus.js``` as well. It is heavily based on this code:

https://simplewebrtc.com/bundle-edge.js

that is the code that is currently used by SimpleWebRTC for the Edge-to-Chrome/Firefox audio call demo. That code already did what I wanted to do, that is prepare a shim that could translate to/from SDP, and also created the prototypes that made it WebRTC 1.0 compliant, which really helped. I only had to fix some things here and there to make it more "adapter.js" compliant, and to fix issues I was experiencing, namely:

* adding some shims that we had and were missing there (getUserMedia, getSources, attachMediaStream)
* disabling the ICE servers, as no matter what I put there Edge always barfed on them (a potential I already [reported](https://connect.microsoft.com/IE/feedbackdetail/view/1894505/rtcicegatherer-rtcicetransport-issues))
* adding candidates from the SDP, if any (the original version only used trickle)
* codec name matching uses toLowerCase (e.g., the original version failed for opus vs OPUS)
* codecs that are known not to be WebRTC compliant are skipped before passing them to the stack (some codecs Asterisk sent caused exceptions in Edge)
* a few more things I can't remember

I think that's all. Please test and provide feedback. If you play with it and improve the code, that's even more welcome, especially considering I'm doing these tests with a crappy VM and I don't have access to an actual machine with Windows 10/Edge installed.

### DIFF
--- a/html/adapter.js
+++ b/html/adapter.js
@@ -11,7 +11,14 @@
 mozRTCSessionDescription, webkitRTCPeerConnection */
 /* exported trace,requestUserMedia */
 
-'use strict';
+/*
+ *  Edge support provided thanks to the adaptation devised by SimpleWebRTC
+ *
+ * 		https://simplewebrtc.com/bundle-edge.js
+ * 
+ */
+
+//~ 'use strict';
 
 var RTCPeerConnection = null;
 var getUserMedia = null;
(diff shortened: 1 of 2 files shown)


simongymer-speakerbus:
Data channel support is essential for our application so I hope Microsoft get that into EDGE sooner rather than later. :-) 

I will keep a close eye on EDGE upgrades to see what and when they put all the features in.

lminiero:
I'm afraid you'll have to wait a while for that :wink: 
https://webrtchacks.com/ortc-edge-microsoft-qa/

MichaelB76:
Thanks for the info, I'm interested in how EDGE is coming on as well.

warrenjmcdonald:
I encountered similar ICEServers issue when playing with the Twilio adaptor but didn't have time to follow up with more tests. I am glad its not just me :) This is not such an issue for Client to Media Server, but it will kill them in any Internet peer to peer cases :(

lminiero:
The ICEServers issue is apparently a known issue that should be fixed in subsequent revisions of Edge. As it is now, the workaround we have is the only way to cope with that.

Anyway, merging this PR for now, as it mostly seems to work, and doesn't introduce any regression for other browsers.

Issue #154, opened 2015-02-13
eventfd spinning


ploxiln:
I figured out what was sometimes making threads "spin", consuming as much cpu as they can: write()s to eventfd.

First I identified a thread that was "going crazy", and then counted syscalls with strace over a few seconds:
```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 59.84    0.712945           6    110297           write
 19.47    0.231949           6     36765           clock_gettime
 10.60    0.126276           7     18383           poll
  9.95    0.118586           6     18382           read
  0.14    0.001683          12       135        32 futex
------ ----------- ----------- --------- --------- ----------------
100.00    1.191439                183962        32 total
```

that seems like more write() than makes sense for sending rtp/srtp packets. Let's see what it's writing...

```
clock_gettime(CLOCK_MONOTONIC, {823252, 946459022}) = 0
poll([{fd=15, events=POLLIN}, {fd=26, events=POLLIN}, {fd=28, events=POLLIN}, {fd=32, events=POLLIN}, {fd=34, events=POLLIN}, {fd=25, events=POLLIN}, {fd=27, events=POLLIN}, {fd=31, events=POLLIN}, {fd=33, events=POLLIN}], 9, 2553) = 4 ([{fd=15, revents=POLLIN}, {fd=27, revents=POLLIN}, {fd=31, revents=POLLIN}, {fd=33, revents=POLLIN}])
read(15, "\6\0\0\0\0\0\0\0", 16)        = 8
clock_gettime(CLOCK_MONOTONIC, {823252, 946602023}) = 0
write(15, "\1\0\0\0\0\0\0\0", 8)        = 8
write(15, "\1\0\0\0\0\0\0\0", 8)        = 8
write(15, "\1\0\0\0\0\0\0\0", 8)        = 8
write(15, "\1\0\0\0\0\0\0\0", 8)        = 8
write(15, "\1\0\0\0\0\0\0\0", 8)        = 8
write(15, "\1\0\0\0\0\0\0\0", 8)        = 8
clock_gettime(CLOCK_MONOTONIC, {823252, 946880949}) = 0
[...]

PR #362, opened 2015-10-20
Enforce trickle after offer/answer

lminiero:
Pull request to make sure trickle candidates are not passed to the stack until we have both offer and answer ready. This should help fixing issue #154, as that way we're sure we're never going to set any remote candidate for a remote candidate we won't be using (see rtcp-mux).

Published as a PR for now as I want to be sure this doesn't cause potential failures in negotiating a PeerConnection.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1268,10 +1268,14 @@ void janus_ice_cb_candidate_gathering_done(NiceAgent *agent, guint stream_id, gp
 
 void janus_ice_cb_component_state_changed(NiceAgent *agent, guint stream_id, guint component_id, guint state, gpointer ice) {
 	janus_ice_handle *handle = (janus_ice_handle *)ice;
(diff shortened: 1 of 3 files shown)


ice.c
```diff
@@ -339,6 +384,18 @@ static gboolean janus_ice_handles_check(gpointer user_data) {
 	}
 	janus_mutex_unlock(&old_handles_mutex);
 
+	if(janus_force_rtcpmux_blackhole_fd > 0) {
```
ploxiln: janus_force_rtcpmux_blackhole_fd is initialized to `-1`, so if it is `>= 0` then it was set. Of course, it would only be zero if stdin was closed and no other fds took that number and stayed open, which is an impossible corner case, just thought I'd mention the very minor inconsistency :)

lminiero: Good point, thanks, I'll fix that as soon as I can get back to the code!

lminiero: Simply changed the initialization to 0: since we only use it if it's >0, then it should be ok.

ploxiln: sounds good to me

Issue #363, opened 2015-10-21
Google Chrome Deprecating MediaStream.stop in November 2015


simongymer-speakerbus:
'MediaStream.stop()' is deprecated and will be removed in M47, around November 2015. Please use 'MediaStream.active' instead.

https://developers.google.com/web/updates/2015/07/mediastream-deprecations?hl=en

Looks like they want you to stop all the tracks within a stream rather than have a nice convenient stop method on the stream? :-(

This method is used by Janus in the cleanupWebrtc() function of janus.js.

lminiero:
Yep, we know, that's why we put the ```MediaStream.stop()``` stuff in a try/catch, to avoid it causing an exception in JavaScript. We'll have to add the ```MediaStreamTrack.stop()``` stuff too.

LetsVape:
I meant to do a pull request for this sooner when I saw it pop up in the console, this reminded me so here you go.
I've used this for a week or 2 without any errors, warnings or issues as far as I know, hope the if conditionals and formatting are fine to fit with the projects coding

lminiero:
I've just finished working on a different approach, that iterates on all the available tracks rather than just checking for the first instance of audio and video. This is to account for when we'll add multi track support in Janus (soon, I hope). Besides, I'm keeping the MediaStream.stop() where it is since there may still be Chrome/Firefox versions around that expect that. Thanks for the effort though!

PR #364, opened 2015-10-21
Use MediaStreamTrack.stop() instead of MediaStream.stop() which is being deprecated

LetsVape:
As mentioned in issue #363 MediaStream.stop() is being deprecated, I've changed janus.js to use MediaStreamTrack.stop() instead


lminiero:
Already taken care of in https://github.com/meetecho/janus-gateway/commit/389082b3a2380e5be8f51c6792f8eb84a8390a19, thanks anyway!

LetsVape:
Cool, that's a bit cleaner way of doing it indeed, I'll merge that instead. Thanks

Issue #366, opened 2015-10-27
AddressSanitizer: heap-use-after-free janus_streaming.c:2040 janus_streaming_handler


fpn:
```
janus-stream: [Tue Oct 27 13:22:01 2015] Freeing old Streaming session
janus-stream: [Tue Oct 27 13:22:01 2015] Freeing old Streaming session
    #0 0x7f2562e44b5b (/usr/lib/x86_64-linux-gnu/libasan.so.0+0xab5b)
    #1 0x7f2562723e61 (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x8ae61)
Thread T25 (websockets thre) created by T0 here:
    #0 0x7f2562e44b5b (/usr/lib/x86_64-linux-gnu/libasan.so.0+0xab5b)
    #1 0x7f2562723e61 (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x8ae61)
SUMMARY: AddressSanitizer: heap-use-after-free /home/fpn/janus-gateway/plugins/janus_streaming.c:2040 janus_streaming_handler
```

```
2040         janus_streaming_session *session = (janus_streaming_session *)msg->handle->plugin_handle;
```

We add a link to the handle -> session to the message before we put it on the queue. Under load there may be messages still in the queue which have a reference and the watchdog already has freed the session.

The associated pull request is only for the streaming plugin, but I noticed that other plugins e.g. videoroom have the same issue.

lminiero:
I'm in Japan for about 10 days for the IETF, so I might be slow at responding... I'll try to give a look at the PR as soon as I have some time to do so. Thanks!

PR #367, opened 2015-10-27
msg->handle->plugin_handle may not exist when message is handled

fpn:
https://github.com/meetecho/janus-gateway/issues/366


lminiero:
Sorry for the delay, I'll look into this in the next few hours/days: been abroad so far and couldn't check this.

lminiero:
Looks fine, thanks! I'll take care of adding this to all the other plugins as well.

PR #369, opened 2015-11-04
BUGFIX : opus fill silence packet

jing3018:


lminiero:
What issue did you notice that needed this fix? It seemed to be working fine for us.

jing3018:
1. the granulepos is not monotonous，the value in the silence packet is big the the follow valid packet. 
```
in silence packet : pos = tmp->seq-list->seq+steps*65536+i+1;
but when write packet tmp
in tmp packet  :   tmp->seq-list->seq+steps*65536
```

2. when lost `N` packet,  writen `N+1` silence packet
```
tmp->seq-tmp->prev->seq  - 0     == N + 1
tmp->seq-tmp->prev->seq - 1 -0  == N
```

lminiero:
Thanks for the clarification! I'll look into this when I get back to the office, as I'm currently in Japan for work and so have only limited time to look into code and things. Ping me if I forget to do that! :smile:

jing3018:
@lminiero ， have you review this pull request?

lminiero:
@jing3018 I've returned to work this morning, I'll look into this in the next few hours. Thanks for the reminder!

lminiero:
I was able to verify this, thanks for fixing it. I also noticed that we're doing an error writing twice the size of opus_silence (it should just be ```sizeof(opus_silence)```), I'll commit that after I merge this.

PR #371, opened 2015-11-09
INI configuration utilities changes

lminiero:
This pull request changes a few things in how INI configuration files are handled. Specifically:

1. it removes the dependency from ```libini_config```, which apparently was not that easy to satisfy in some environments (see OSX), and replaces it with a homemade implementation;
2. it changes the way you access categories and items: now both ```janus_config_get_categories``` and ```janus_config_get_items``` return a GList, which means ```janus_config_category``` and ```janus_config_item``` don't have a ```next``` anymore: to traverse them you use GList instead;
3. the patch also adds permanent saving features to Streaming, AudioBridge and VideoRoom plugins, meaning that you can now make sure a mountpoint/room created via API is also saved in the configuration file if you want; this feature was already available in the old config.c, but wasn't being used in any plugin so far.

This is a PR and not a plain commit mostly because of point 2., as there may be third-party plugins out there that may break if they're not fixed accordingly. If you're writing a plugin and using ```janus_config_get_categories``` and/or ```janus_config_get_items```, make sure you change the way they're called as I'm planning to merge this soon after this announcement.

Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,6 @@ on [github](https://github.com/meetecho/janus-gateway/issues) instead.
 To install it, you'll need to satisfy the following dependencies:
 
 * [libmicrohttpd](http://www.gnu.org/software/libmicrohttpd/)
-* [libini-config](https://fedorahosted.org/sssd/) (INI configurations)
 * [Jansson](http://www.digip.org/jansson/)
 * [libnice](http://nice.freedesktop.org/wiki/)
 * [OpenSSL](http://www.openssl.org/) (at least v1.0.1e)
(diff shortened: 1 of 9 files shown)


ploxiln:
I haven't reviewed the code closely, but I like the idea.

bhakimi:
can you guys update the wiki to remove the libini_config dependency please =) that library was tricky to get on gentoo, it was part of a dev-libs/ding-libs and took a while to find

lminiero:
I'll remove it when I merge this pull request. Right now it's still a
dependency.
Il 11/nov/2015 03:45, "bhakimi" <notifications@github.com> ha scritto:

lminiero:
Merging this PR now. If you still want to use the older Janus version without this change, I just tagged what can be considered the stable 0.0.9 version of Janus here: https://github.com/meetecho/janus-gateway/tree/v0.0.9

bhakimi:
i made a gentoo ebuild that can be used for this project, where can i submit it for others to use?

PR #373, opened 2015-11-09
feature/videoroom/camera

leonuh:
I've added a camera switcher functionality based on the API, but I noticed a bug. I tested with Ubuntu Chrome, and Android Chrome but I suppose it is a general bug.

Steps:

1. Join to the room with at least 2 client - Audio works
2. "Turn off" the camera - Audio works
3. Join with a new client - The new client can't hear the other one where the camera was turned off
4. "Turn on" the camera - Audio works again for the new client's too

### DIFF
--- a/html/videoroomtest.js
+++ b/html/videoroomtest.js
@@ -56,6 +56,7 @@ var myusername = null;
 var myid = null;
 var mystream = null;
 var muted = false;
+var camera = true;
 
 var feeds = [];
 var bitrateTimer = [];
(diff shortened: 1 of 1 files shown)


leonuh:
Thank you very much, I have checked the last commits. It solved the problem!

lminiero:
Well we didn't change anything there, so I guess there wasn't an issue in the first place :smile:

leonuh:
I assume the bug itself is still there in the MCU. But you have added some new functions, unmuteAudio, muteAudio, unmuteVideo, muteVideo (I was not aware of the possibility to deactivate the stream through the Stream API). These have similar results that I wanted to solve from the API call.

lminiero:
Ah ok, I guess you were trying to renegotiate streams, then, which is currently not supported.

PR #380, opened 2015-11-11
JS client: Properly remove WebSocket event listeners

zazabe:
When the `Janus` client is destroyed and recreated, the previous `WebSocket` is still listening to previous messages and throw `Ooops: 458 No such session 1896383161` error.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -144,7 +144,7 @@ function Janus(gatewayCallbacks) {
 	}
 	var websockets = false;
 	var ws = null;
-	var wsHanders = {};
+	var wsHandlers = {};
 
 	var servers = null, serversIndex = 0;
 	var server = gatewayCallbacks.server;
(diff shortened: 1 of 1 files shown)


lminiero:
What problem is this PR solving?

zazabe:
`Ooops: 458 No such session xxx` error due to websocket listeners not removed from a destroyed `Janus` instance... 
Actually i suppose [this FIXME comment](https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L323) is related to that too, isn't it?

lminiero:
Well that FIXME is for error handling in general, which is left to applications anyway as the payload is sent up. Thanks for looking into this though!

About the patch, I'll look at it deeper later, just a couple of quick notes:

1. s/wsHanders/wsHandlers?
2. not sure about the ```unbindWebSocket``` part, can't this now cause situations where ```gatewayCallbacks.destroyed()``` is never called and the application never notified, since it's waiting for a response/event matching the session?

zazabe:
about the `unbindWebSocket`, I added a listener on `error` to remove listeners and call the `destroyed` method anyway, but i'm still not sure to cover all cases, wdyt?

lminiero:
This looks much better, thanks! I'll give it a test in the next few hours and come back with some feedback, if needed.

Before I can merge this, though, have you signed our [CLA](https://cla.conf.meetecho.com/janus/)? We need this for every contribution to Janus.

lminiero:
Sorry, the "few hours" became days and I still haven't found some time to look into this... I promise I'll do that between today and tomorrow, thanks again!

lminiero:
Looks good, merging!

PR #382, opened 2015-11-13
RTCP Receiver Reports for videoroom publisher (first draft)

fpn:
Hi,

The current pull request implements RTCP Receiver Reports for videoroom publishers.
So now we can see round trip time, jitter and packetsLost in webrtc-internals and via the getStats call.
 
* On the technical side one of the questions is - where should the stats live? I saw the stats in ice.c and feel that maybe the whole thing should live there? What do you think?

* QoS stats output - I think there is a value to output latency, jitter, packets lost, received for a session, but probably more like once a minute and not with every RR/SR, also I personally would prefer to output it as JSON so I can feed it into some stats tool.

best regards,
Florian

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,6 +84,8 @@ PKG_CHECK_MODULES([JANUS],
                     libcrypto
                     sofia-sip-ua
                   ])
+
+JANUS_MANUAL_LIBS+=" -lm"
 AC_SUBST(JANUS_MANUAL_LIBS)
 
 AS_IF([test "x$enable_boringssl" = "xyes"],
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -290,6 +290,8 @@ typedef struct janus_videoroom_participant {
 	gboolean audio, video, data;		/* Whether audio, video and/or data is going to be sent by this publisher */
 	guint32 audio_ssrc;		/* Audio SSRC of this publisher */
 	guint32 video_ssrc;		/* Video SSRC of this publisher */
+	rtcp_context *audio_rtcp_ctx;
+	rtcp_context *video_rtcp_ctx;
 	gboolean audio_active;
 	gboolean video_active;
 	gboolean firefox;	/* We send Firefox users a different kind of FIR */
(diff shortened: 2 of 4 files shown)


lminiero:
Amazing effort, thanks for sharing! I've been meaning to work on something like this forever, and never led myself to that due to tons of stuff to do in the meantime... I haven't looked into this in detail, but from a quick glance this looks great!

As a side note, why is the explicit linking to ```-lm``` there? Isn't this usually automatically added? Although I guess the same could be said about ```-pthread``` and we've seen it failing...

As to your questions:

1. I do believe this belongs to the core, yes, rather than a single plugin, so in ice.c as you said: that would make the feature transparently available to all plugins without the need for replicating the same thing over and over. That's what we do for NACKs, for instance, and all RTCP packets are filtered there anyway.
2. We can re-use the admin API for that. If you check the info related to a handle through an admin API, you'll notice that basically everything we have on the media transmission ends there (e.g., number of bytes we've sent/received in general and in the last second, NACKs, etc.), so adding this extra pieces of information there would be exactly what you're looking for, as the admin API is JSON based and you can decide the rate for the info yourself (request/response). I've also considered, from an issue page that suggested this some time ago, to allow for admin API subscriptions and asynchronous events, and this could become part of that as well when it's ready.

Does this make sense? Let me know if I can help you in any way on this, and thanks again!

lminiero:
@fpn I can try moving the code around so that it's part of the core instead, if you like. I can prepare a new pull request so that you can check everything works as expected there too.

IfpnI:
@lminiero that would be awesome! Hopefully SR can then follow the same pattern. Thank you!

lminiero:
Ok, I'll work on a PR based on your code to make available ASAP, I'll keep you posted.

mtdxc:
@lminiero Is there any updates on this PR? WebRTC guys has a feature of adaptive FEC which needs RTCP to enable. You can search HANDLING PACKET LOSS IN WEBRTC

arjenveenhuizen:
I am also very interested in this PR.

lminiero:
I still didn't have much time to work on this, as I've mostly worked on the reference counters stuff lately. I'll try and get to it next week, as I'm a bit busy this one.

lminiero:
I've started working on this and managed to move things around. It's still not complete (I still have to figure out how to make some of this new core functionality available, in an opaque way, to plugins), but in the next few days I'll clean it up a bit and push what's ready at the moment. This will allow us to check if at least it does what it's supposed to do (especially with respect to the stats collection in the Admin API), and the discuss how to proceed from there.

lminiero:
First PR created, feedback welcome.

lminiero:
Closing as the patch is ready in #467, we can continue the discussion there.

PR #384, opened 2015-11-18
Using conditions (wait/signal) instead of sleeps for worker threads

lminiero:
Most of the worker threads we have currently make a heavy use of sleeps when they're waiting for things. That's the case, for instance, of message handling threads in all plugins, which check if there are messages in queue, and if not sleep for a few milliseconds.

This is far from optimal, of course, so after some feedback and suggestions from @ploxiln I prepared an attempt to make this all cleaner using pthread conditions. This patch adds conditions support in mutex.h (basically wrapping pthread stuff as we already did for mutexes), and uses them just in the EchoTest plugin (because I'm lazy) in place of the sleep we had before.

This seems to be working as expected in my case, but feedback is of course welcome, especially in terms of:

1. is this really needed? (I think it is, especially if we extend this to all other current sleeps)
2. are we doing it correctly or do you see room for improvements?
3. do you see problems ahead if we go down this road?

As soon as I have confirmation that we're doing the right thing, I'll proceed and modify other plugins and core functions so that we do pretty much the same thing there as well.


saghul:
I like the idea, and agree it's needed, FWIW.

A couple of general comments: 

- why not use [glib's conditions](https://developer.gnome.org/glib/stable/glib-Threads.html#GCond)? it probably takes care of cross-platform differences already, which we might run into, down the road.
- we need to watch out for spurious wakeups, that is, even if a condition returns, the condition needs to be rechecked.

:+1:

lminiero:
> why not use glib's conditions? it probably takes care of cross-platform differences already, which we might run into, down the road.

That's a good point, that probably applies to mutexes as well. The reason is that I just reused the same approach we had in mutex.h. We can change that along the road at any time, since we're redefining/wrapping the methods we really use anyway.

> we need to watch out for spurious wakeups, that is, even if a condition returns, the condition needs to be rechecked.

Yep, in fact the echotest handler thread currently checks for available messages after a wait ends, and if there aren't it restarts the loop and hence waits again.

ploxiln:
The handler checks `stopping` outside of the mutex-protected region, so it's theoretically possible (though unlikely) that
  1. ..._handler checks stopping and sees 0
  2. ..._destroy sets stopping to 1
  3. ..._destroy takes mutex, notifies condition variable, releases mutex
  4. ..._handler takes mutex and waits

ploxiln:
Now that I've looked at it, it looks like g_async_queue does something similar internally. If you use `g_async_queue_pop()` instead of `g_async_queue_try_pop()` it will sleep the thread until a message is available. So the last piece needed would be a way for the destroy function to wake the handler thread so it can check stopping - I'd just send an (empty/trivial) message to it to do that.

lminiero:
So you mean using the code we had before (whih already used GAsyncQueue) and change try_pops to pops? Hadn't thought of that, I'll have to check what GAsyncQueue does internally to wait for the packet: if it's sleeping instead of using something like conditions we're just doing the same thing in a different way. If it's using conditions instead it much less work for us!

lminiero:
Looking around apparently it uses conditions, so waits and signals: tomorrow I'll modify the code to revert the behaviour bak to what it was and use a pop.

ploxiln:
yeah when I said "sleep" I meant "block on synchronization primitives", sorry :)

lminiero:
This new commit restores the echotest plugin code to what it was before, and just uses ```g_async_queue_pop``` instead of the ```g_async_queue_try_pop``` we had before. When destroying the plugin, we send an empty (just allocated) message, so that ```g_async_queue_pop``` can return and immediately continue to the end of the loop.

This seems to be working fine as well, and since GAsyncQueue uses conditions internally for that it suits our purpose all the same.

ploxiln:
looks great to me

lminiero:
Cool, thanks for checking that! I'll update the other plugins next week and push this, then will start getting rid of the other sleeps we have around (e.g., watchdogs in the plugins themselves, where a glib source might be better suited).

ploxiln:
The watchdogs in particular are probably a trickier case to handle, and not as much of a problem because they wake up less often. For those I might just make the wakeup period longer, maybe 1 or 2 seconds instead of 500ms.

(also the name "watchdog" for those threads is a bit odd, they're more like GC, but that's not a real problem :)

lminiero:
> The watchdogs in particular are probably a trickier case to handle, and not as much of a problem because they wake up less often. For those I might just make the wakeup period longer, maybe 1 or 2 seconds instead of 500ms.

Good point, I'll find something else to improve! :smile:

> (also the name "watchdog" for those threads is a bit odd, they're more like GC, but that's not a real problem :)

You're right, you can blame me for that, I have a knack for calling things in a confusing way :wink:

saghul:
I wonder if this could be simplified even further by doing the following (pseudocode):

~~~~
static janus_foo_message exit_message;
...
// when we need to exit
g_async_queue_push(messages, &exit_message);
...
...
// worker loop
for (;;) {
    msg = g_async_queue_pop(messages);
    if (msg == NULL)
        continue;
    if (msg == &exit_message)
        break;
    ...
}
// cleanup worker
...
~~~~

lminiero:
> I wonder if this could be simplified even further by doing the following (pseudocode):

That's a good suggestion, and would allow us to avoid the allocation of a new message just for the purpose of leaving the thread. I'll modify the code so that it's done.

In the meanwhile, I've modified the main loop in janus.c to use conditions, and am working on making use of async queues in sctp.c as well, where I believe we were not doing this in a very optimized way. I'll commit those as soon as they're decent enough.

ploxiln:
I agree, static address of exit message is even better.

I might initialize the exit message though ... I know it doesn't matter because it's never read by anything, but it's really easy to initialize it to all fields 0/NULL, like so:

```
static janus_audiobridge_message exit_message = {0};
```

lminiero:
:+1:

saghul:
Being static it's guaranteed to be zeroed, the standard says so.

ploxiln:
Indeed, indeed ... call it being explicit then :)

lminiero:
I'll be the Switzerland here and do absolutely nothing :smile: 
I'll get back to that later, as it's not a pressing issue.

lminiero:
Just used the same approach for data channels as well, where I've simplified the management of messages: we had two different, not async, queues before, and frequent sleeps.

I guess this concludes this first round of worker threads fixup. I'll merge this as soon as there's enough feedback that says so, we can then work on improving different parts in a different patch. I think that, among other things, I'll work on the plugin watchdogs anyway mostly to reduce the shutdown times, which are now in the range of seconds due to those sleeps over there (we join the threads).

janus.c
```diff
@@ -203,7 +206,14 @@ void janus_handle_signal(int signum)
 			JANUS_PRINT("Ok, leaving immediately...\n");
 			break;
 	}
+
+	/* Wake the main loop */
+	janus_mutex_lock(&stop_mutex);
```
ploxiln: It's generally not a good idea to lock a mutex from a signal handler. Some pthread_mutex_lock man pages mention this. Also you can see that there's a limited list of "async-signal safe functions" in the `man 7 signal` man page.

It's not hard to understand why: the same thread may have already taken the mutex, when the signal interrupts it and jumps it to the signal handler, where if it tries to take the mutex again and sleeps waiting for it to be available, it will deadlock.

One thing you *can* do here is write a single byte to a pipe. Or just let the while/sleep loop be, maybe making the sleep longer since it will be interrupted by a signal anyway.

(this comment moved from the commit to the pull request diff)

lminiero: ```stop_mutex``` is only ever used by the main loop, so I don't see how that may be happening? or you mean the signal might be interrupting the main loop as well and hence cause that issue?

ploxiln: Indeed, the main loop (which is run by the first/main thread of the process) receives the signal and jumps to the signal handler (even though it's unlikely to come just as the main loop takes the mutex). That's why the signal interrupts the usleep() there (I think g_usleep() is in a loop that sleeps again for the remaining time if its interrupted).

lminiero: Thanks for the hint, I was thinking in terms of generic event-based notifications and didn't bother actually checking how it worked with system signals. I'll fix this tomorrow.

lminiero: I've restored what we had before. I'll think of a better way to handle that.

lminiero:
Reverted the dangerous condition in the signal handler and removed the sleeps from both HTTP and RabbitMQ transport modules. The HTTP one is worth giving a second look at, just to be sure we're not messing anything up (all my tests seemed to go fine).

The HTTP changes also fix a lingering bug there was: if the core took to long to answer (>10s), the plugin hung up the connection, but since the core still had a reference to the object as an opaque pointer, when the answer finally came the pointer was used in the plugin thus leading to a segfault. Fixed this by always keeping an hash table of currently active message transactions.

transports/janus_http.c
```diff
@@ -1186,15 +1209,18 @@ int janus_http_handler(void *cls, struct MHD_Connection *connection, const char
...
+	struct timeval now;
+	gettimeofday(&now, NULL);
+	struct timespec wakeup;
+	wakeup.tv_sec = now.tv_sec+10;	/* Wait at max 10 seconds for a response */
+	wakeup.tv_nsec = now.tv_usec*1000UL;
+	while(TRUE) {
+		pthread_mutex_lock(&msg->wait_mutex);
+		int res = pthread_cond_timedwait(&msg->wait_cond, &msg->wait_mutex, &wakeup);
+		pthread_mutex_unlock(&msg->wait_mutex);
+		if(msg->got_response || res == ETIMEDOUT)
```
ploxiln: Where you call `janus_condition_signal()` you should set `msg->got_response` while you hold the mutex, and here you should check `msg->got_response` also while holding the mutex, *before* waiting on the condition variable.

Otherwise, it's theoretically possible that the condition is signaled just before you take the mutex and wait on it.

lminiero: I thought I had done that already, but just saw it's done immediately *before* the lock, so you're right. Fixing this right away.

ploxiln:
I think you added the file `ciccio.mkv` accidentally

lminiero:
> I think you added the file ciccio.mkv accidentally

Ahaha I totally did! Good thing I wasn't naked :smile:

transports/janus_http.c
ploxiln: this may have been forgotten :)

lminiero: I did fix that: https://github.com/meetecho/janus-gateway/commit/2320562744890b0d5c7f2f77ed9e8779d612b8b7

ploxiln: ah, I see now you did in one place, but the same pattern and problem appears in this (second) place as well

lminiero: Good catch, forgot we had the same code for the admin responses too... fixed in https://github.com/meetecho/janus-gateway/commit/55889e15505d2f49890ada32e36124b0c42730a9

transports/janus_rabbitmq.c
```diff
@@ -333,12 +323,22 @@ int janus_rabbitmq_init(janus_transport_callbacks *callback, const char *config_
...
 	}
+	g_free(rmqhost);
+	if(to_janus)
+		g_free((char *)to_janus);
+	if(from_janus)
+		g_free((char *)from_janus);
+	if(to_janus_admin)
+		g_free((char *)to_janus_admin);
+	if(from_janus_admin)
+		g_free((char *)from_janus_admin);
```
ploxiln: no real need for the `if()`s

https://developer.gnome.org/glib/2.46/glib-Memory-Allocation.html#g-free

> If mem is NULL it simply returns, so there is no need to check mem against NULL before calling this function.

lminiero: We have in several places, so that's good to know. I'll change those as part of the new memory management PR effort.

ploxiln: sounds good to me

Will-I4M: Hi guys, could someone tell me the gain (CPU...) we expect by replacing
sleeps with conditions (wait/signal) ?
Just curious :)

lminiero: I don't have figures (didn't do any new test), but there should be several advantages. We had a lot of places where we only slept a sequence of tiny amount of time in a loop waiting for something to happen, as we also needed to keep on eye on when something relevant happened (e.g., session destroyed in the meanwhile, Janus hanging up, etc.). This caused the related threads to wake up continuously, and was against the idea of just going to sleep and really waiting for something to happen without needing a constant poll. This approach should help with that, waking threads only when it's really needed and also reducing context switches as a consequence.

lminiero:
Any objections on merging this? I've been using this as the base code for all my tests and it seems to be working fine.

ploxiln:
No objections.

But I've thought of one more corner case: when janus is shutdown and plugins are cleaning up and call `g_async_queue_unref()`, the exit message pointer might still be in the queue (if the queue already had messages in it and the handler function processed a different message and then failed the while loop condition), in which case it will call the message free function on the exit message, which can't be freed.

lminiero:
You're right, I fixed this in my setup while working on the reference counters and forgot to make the same changes here, I'll do that tomorrow.

lminiero:
Just taken care of that, and while I was at it I also made the ```janus_*_message_free``` methods static.

Issue #375, opened 2015-11-10
latest checkout has issues with sofia-sip stack


bhakimi:
when i register a sip client i get this error
```
nta: bind(XXX.XXX.XXX.XXX:*;transport=*): Protocol not supported
nua: initializing SIP stack failed
```

bhakimi:
here is more debug info
```

[janus.c:janus_ws_handler:701] Done getting payload, we can answer
{"janus":"message","body":{"request":"register","username":"sip:128@XXX.XXX.XXX.XXX","secret":"password"},"transaction":"oCPVNNlJSyE4"}
[1476096338] There's a message for JANUS SIP plugin
{
   "request": "register",
   "username": "sip:128@XXX.XXX.XXX.XXX",
   "secret": "password"
}
Creating plugin result...
Destroying plugin result...
[janus.c:janus_ws_request_completed:2930] Request completed, freeing data
Handling message: {
   "request": "register",
   "username": "sip:128@XXX.XXX.XXX.XXX",
   "secret": "password"
}
Registering user sip:128@XXX.XXX.XXX.XXX (secret password) @ 208.78.138.35 through (null)
Joining sofia loop thread (128)...
Setting up sofia stack (sip:128@XXX.XXX.XXX.XXX)
nta: bind(208.78.138.103:*;transport=*): Protocol not supported
nua: initializing SIP stack failed
[ERR] [plugins/janus_sip.c:janus_sip_handler:1187] Two seconds passed and still no NUA, problems with the thread?
Pushing event: {
   "sip": "event",
   "error_code": 499,
   "error": "Two seconds passed and still no NUA, problems with the thread?"
}
[1476096338] Adding event to queue of messages...
  >> 0 (Success)
We have a message to serve...
        {
   "janus": "event",
   "session_id": 533647822,
   "sender": 1476096338,
   "transaction": "oCPVNNlJSyE4",
   "plugindata": {
      "plugin": "janus.plugin.sip",
      "data": {
         "sip": "event",
         "error_code": 499,
         "error": "Two seconds passed and still no NUA, problems with the thread?"
      }
   }
}

```

saghul:
What sofia-sip version did you install?

bhakimi:
1.12.11, i even tried 1.12.10 and the same issue

lminiero:
[...]

PR #386, opened 2015-11-20
Optional SIPS when registering

lminiero:
See #375

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2467,32 +2467,18 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 		g_strlcat(outbound_options, " options-keepalive", sizeof(outbound_options));
 	if(!behind_nat)
 		g_strlcat(outbound_options, " no-natify", sizeof(outbound_options));
-	if(session->account.sips) {
-		session->stack->s_nua = nua_create(session->stack->s_root,
-					janus_sip_sofia_callback,
-					session,
-					SIPTAG_ALLOW_STR("INVITE, ACK, BYE, CANCEL, OPTIONS"),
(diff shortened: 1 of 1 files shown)


plugins/janus_sip.c
```diff
@@ -2450,18 +2467,32 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
-				SIPTAG_ALLOW_STR("INVITE, ACK, BYE, CANCEL, OPTIONS"),
-				NUTAG_M_USERNAME(session->account.username),
-				NUTAG_URL(sip_url),
-				NUTAG_SIPS_URL(sips_url),
-				SIPTAG_USER_AGENT_STR(user_agent),
-				NUTAG_KEEPALIVE(keepalive_interval * 1000),	/* Sofia expects it in milliseconds */
-				NUTAG_OUTBOUND(outbound_options),
-				SIPTAG_SUPPORTED(NULL),
-				TAG_NULL());
+	if(session->account.sips) {
```
saghul: I think you can simplify this by using TAG_IF: http://sofia-sip.sourceforge.net/refdocs/su/su__tag_8h.html#a8ccbc9cc9f07d5670065e990d07bb428

saghul:
LGTM, but have a look at TAG_IF :-)

plugins/janus_sip.c
lminiero: Oh, I wasn't aware of that, that's neat! I'll update the PR right away.

saghul:
:+1:

lminiero:
Thanks for the quick review, merging then!

bhakimi:
just a note, the taf if still included the sips even when i did not make a sips request, i also posted these findings in the merged ticket

PR #388, opened 2015-11-21
Added commandline option that allows to daemonize janus upon successful start

mspanc:


lminiero:
There already is a daemonizing command line attribute:
https://github.com/meetecho/janus-gateway/blob/master/janus.ggo#L1
```-b``` or ```--daemon```

mspanc:
Oh I am sorry, I was looking at 0.0.9.

lminiero:
No problem! What we have daemonizes Janus immediately, though, and doesn't wait for a successful start. Not sure if we should change that. I guess it depends on whether external initialization scripts can figure out something went wrong after that.

mspanc:
It think it's important to wait until successful start. If you daemonize immediately you cannot properly return non-zero exit code to the parent process so basically external scripts/monitoring tool will think that everything went well while it didn't.

lminiero:
Fair point, I'll keep that in mind. The daemon stuff is still WIP anyway, as we logging is still stdout/stderr redirection, something we'll have to change soon so that you can also log to file directly.

PR #389, opened 2015-11-21
janus buffered logging

jayridge:
\+ @ploxiln 

A stab at buffered logging in janus core. A few notes.

1. I took a look at glib async queues. The implementation was similar to this pull with extra overhead. Not implemented.
1. The init sequence is likely not technically correct. The application logs before the signal handlers are installed or glib is initialized ( N.B. not required for modern versions ).
1. There was no clear advantage to explicit buffers over glib dynamic strings. Resizing of the buffers would have followed the same model. Also, glib provides a platform agnostic vsprintf that does not reallocate if the buffer is bug enough.
1. Reclamation of buffers is prioritized over lock avoidance. I do not anticipate lock contention here - we can test with mutex metering on.

### DIFF
--- a/debug.h
+++ b/debug.h
@@ -102,10 +102,10 @@ do { \
 			         "[%s:%s:%d] ", __FILE__, __FUNCTION__, __LINE__); \
 		} \
 		JANUS_PRINT("%s%s%s" format, \
-    		        janus_log_ts, \
-    		        janus_log_prefix[level | ((int)janus_log_colors << 3)], \
-    		        janus_log_src, \
-    		        ##__VA_ARGS__); \
+			janus_log_ts, \
+			janus_log_prefix[level | ((int)janus_log_colors << 3)], \
+			janus_log_src, \
+			##__VA_ARGS__); \
 	} \
 } while (0)
 ///@}
--- a/debug.h
+++ b/debug.h
@@ -4,7 +4,7 @@
  * \brief    Logging and Debugging
  * \details  Implementation of a wrapper on printf (or g_print) to either log or debug.
  * \todo     Improve this wrappers to optionally save logs on file
- * 
+ *
  * \ingroup core
  * \ref core
  */
@@ -14,7 +14,8 @@
 
 #include <glib.h>
 #include <glib/gprintf.h>
- 
+#include "log.h"
+
 extern int janus_log_level;
 extern gboolean janus_log_timestamps;
 extern gboolean janus_log_colors;
@@ -80,7 +81,7 @@ static const char *janus_log_prefix[] = {
  */
 ///@{
 /*! \brief Simple wrapper to g_print/printf */
-#define JANUS_PRINT g_print
+#define JANUS_PRINT janus_vprintf
 /*! \brief Logger based on different levels, which can either be displayed
  * or not according to the configuration of the gateway.
  * The format must be a string literal. */
@@ -100,11 +101,11 @@ do { \
 			snprintf(janus_log_src, sizeof(janus_log_src), \
 			         "[%s:%s:%d] ", __FILE__, __FUNCTION__, __LINE__); \
 		} \
-		g_print("%s%s%s" format, \
-		        janus_log_ts, \
-		        janus_log_prefix[level | ((int)janus_log_colors << 3)], \
-		        janus_log_src, \
-		        ##__VA_ARGS__); \
+		JANUS_PRINT("%s%s%s" format, \
+    		        janus_log_ts, \
+    		        janus_log_prefix[level | ((int)janus_log_colors << 3)], \
+    		        janus_log_src, \
+    		        ##__VA_ARGS__); \
 	} \
 } while (0)
 ///@}
--- a/janus.c
+++ b/janus.c
@@ -2896,7 +2896,7 @@ gint main(int argc, char *argv[])
 	core_limits.rlim_cur = core_limits.rlim_max = RLIM_INFINITY;
 	setrlimit(RLIMIT_CORE, &core_limits);
 
-        janus_log_init();
+	janus_log_init();
 
 	struct gengetopt_args_info args_info;
 	/* Let's call our cmdline parser */
--- a/janus.c
+++ b/janus.c
@@ -28,6 +28,7 @@
 #include "cmdline.h"
 #include "config.h"
 #include "apierror.h"
+#include "log.h"
 #include "debug.h"
 #include "rtcp.h"
 #include "sdp.h"
@@ -2895,6 +2896,8 @@ gint main(int argc, char *argv[])
 	core_limits.rlim_cur = core_limits.rlim_max = RLIM_INFINITY;
 	setrlimit(RLIMIT_CORE, &core_limits);
 
+        janus_log_init();
+
 	struct gengetopt_args_info args_info;
 	/* Let's call our cmdline parser */
 	if(cmdline_parser(argc, argv, &args_info) != 0)
@@ -3704,5 +3707,6 @@ gint main(int argc, char *argv[])
 	}
 
 	JANUS_PRINT("Bye!\n");
+	janus_log_destroy();
 	exit(0);
 }
--- a/log.c
+++ b/log.c
@@ -16,14 +16,14 @@
 #define THREAD_NAME   "log"
 
 
-static gint	 initialized = 0;
-static gint	 stopping = 0;
+static gint     initialized = 0;
+static gint     stopping = 0;
 /* Maximum sleep in ms for the print thread */
-static gint	 maxdelay = 3000;
+static gint     maxdelay = 3000;
 /* Buffers over this size will be freed */
-static gsize	maxbuffer = 1024*16;
+static gsize    maxbuffer = 1024*16;
 static GMutex   lock;
-static GCond	cond;
+static GCond    cond;
 static GSList   *printqueue = NULL;
 static GQueue   *freebufs = NULL;
 static GThread  *printthread = NULL;
--- a/log.c
+++ b/log.c
@@ -34,13 +34,19 @@ static struct Buffer *printtail = NULL;
 static struct Buffer *bufferpool = NULL;
 
 
-static void freebuffer(struct Buffer *b)
+static void freebuffers(struct Buffer **list)
 {
-	if (b) {
+	struct Buffer *b, *head = *list;
+
+	while (head) {
+		b = head;
+		head = b->next;
 		g_string_free(b->s, TRUE);
 		g_free(b);
 	}
+	*list = NULL;
 }
+
 
 static struct Buffer * getbuf(void)
 {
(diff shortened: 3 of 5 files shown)


janus.c
```diff
@@ -2895,6 +2896,8 @@ gint main(int argc, char *argv[])
 	core_limits.rlim_cur = core_limits.rlim_max = RLIM_INFINITY;
 	setrlimit(RLIMIT_CORE, &core_limits);
 
+        janus_log_init();
```
ploxiln: janus code style is tab for indents

debug.h
```diff
@@ -100,11 +101,11 @@ do { \
...
-		g_print("%s%s%s" format, \
-		        janus_log_ts, \
-		        janus_log_prefix[level | ((int)janus_log_colors << 3)], \
-		        janus_log_src, \
-		        ##__VA_ARGS__); \
+		JANUS_PRINT("%s%s%s" format, \
+    		        janus_log_ts, \
+    		        janus_log_prefix[level | ((int)janus_log_colors << 3)], \
+    		        janus_log_src, \
+    		        ##__VA_ARGS__); \
```
ploxiln: you replace the first tab indent with spaces here too :)

ploxiln:
I'm not worried about the lock contention, just the cacheline-locking/bouncing potential of taking and releasing the lock that often.

Why limit the sleep time to 3 seconds? Why any limit there?

log.c
```diff
@@ -0,0 +1,129 @@
...
+{
+	GSList  *head, *p;
+	GString *s;
+	gint64  abstime;
+
+	while (!g_atomic_int_get(&stopping)) {
+		abstime = g_get_monotonic_time() + maxdelay * G_TIME_SPAN_MILLISECOND;
+		g_mutex_lock(&lock);
+		g_cond_wait_until(&cond, &lock, abstime);
+		head = printqueue;
```
ploxiln: I think you want to check `printqueue` before you `g_cond_wait...` because the wakeup signal for a message could have come before you `g_cond_wait...`

jayridge: you are correct. this was a simplification from the plugin version as i did not care about the latency.

janus.c
jayridge: not sure my brain can handle tabs

log.c
jayridge: sleep time was simply a maximum time to have log data buffered in the event the condition signal was not caught. if the lock is held AND printqueue checked this is not necessary.

janus.c
ploxiln: I noticed later you correctly used tabs everywhere but this line (and the debug.h bit) :smile:

lminiero:
Thanks for the PR, I'll give it a check tomorrow when I'm back at the office!

ploxiln:
updates look good to me

log.c
```diff
@@ -0,0 +1,131 @@
...
+{
+	va_list args;
+	GString *s = getbuf();
+
+	va_start(args, format);
+	g_string_vprintf(s, format, args);
+	va_end(args);
+
+	g_mutex_lock(&lock);
+	printqueue = g_slist_append(printqueue, s);
```
ploxiln: Appending requires traversing the list, prepending does not (https://developer.gnome.org/glib/2.42/glib-Singly-Linked-Lists.html#g-slist-append). You could prepend, and then in janus_log_thread, after you take the whole printqueue, reverse it (https://developer.gnome.org/glib/2.42/glib-Singly-Linked-Lists.html#g-slist-reverse). Or use the doubly-linked GList.

ploxiln:
My problem with glib dynamic strings isn't (necessarily) the worry about how they work internally (that can be examined and the worry dispelled). The problem is that they do not include intrusive linked-list pointers. So even though you re-cycle GStrings to avoid re-allocating them for every log message, it is still necessary to re-allocate a (small) GSList structure for every log message. And these GSList node allocations are not freed by the same thread that allocated them, so there is lock overhead returning them to their originating thread's arena free pool.

Might not be a big deal in practice for this application, I don't know. Just making my concern known :)

ploxiln:
Also, I'm not as worried about needing to resize log message buffers (and then freeing over-large ones), I'd just truncate the log message after 2000 characters or so (that's 10 lines of a 200-column terminal).

I had mentioned before (in a separate thread) about freeing extra buffers, for example if there's a pause writing to the log and hundreds of log message buffers are allocated by other threads wanting to log, which will thereafter usually not be needed but be using up memory permanently. So I was interested in capping the length of the freelist. I'm not 100% sure of how necessary this cap is though, I'm still thinking about it.

jayridge:
I agree re gslist. There is no need to allocate/deallocate just for a next pointer on a persistent struct. I will update.

As an api at the core level I think it makes sense to allow users to log large rather than truncate. Which means there should be some facility for releasing large buffers. Although releasing memory seems to rarely reduce RSS. I assume this "free" memory is only reclaimed in starved conditions - at that point you may as well let exit clean things up. I suppose same is true for many small buffers. I guess I need to think on it as well.

lminiero:
Just found some time to look into this, thanks for the amazing effort! I have a few questions, notes and nits, some of which are purely aesthetic or code-style related and so are less important:

1. First of all, I agree that large log lines are needed: just think about printing a received/merged SDP object, for instance, those can easily exceed 2k characters. Anyway, one idea could be to add a truncate option as part of the settings, for those still interested in getting more info (more debug) even if it's not complete (truncate) so that it's more optimized. That said, from what I've understood ```g_string_vprintf``` makes use of resizable strings, and so there's no such thing, for instance, as a ```g_string_vnprintf``` to print N characters at max. I have no strong opinion on this, I guess this really depends on how reliable and fast GString objects are (not familiar with them).
1. I see this is still logging to stdout, even though in a much more structured (and needed) way: do you think adapting the code so that it can optionally log to a file as well (or only there, e.g., for daemons) would require much changes? That's not something you need to do, I can take care of that, I'm just asking to figure out whether the way it's done right now easily allows for that or if some refactoring would be needed.
1. Since we're using the ```janus_FOO_``` prefix for every struct and method of certain files, could you rename those that are currently not compliant with that? e.g., ```s/Buffer/janus_log_buffer```, ```s/freebuffers/janus_log_freebuffers``` and so on. Again, not really important at this stage (I can do that after a merge).
1. Sorry for having forced [tabs](http://www.perlmonks.org/?node_id=158886) upon you! :smile:

lminiero:
One more note: log.c must be added to the sources of ```janus-pp-rec``` as well, or otherwise there are several undefined references (the post processor uses the same debug mechanism as the core). This also means adding ```janus_log_init()``` and ```janus_log_destroy()``` calls in ```janus-pp-rec```, as otherwise no output is printed anymore.

lminiero:
I worked a bit on your code to fix what was mentioned above (including code style) and allow for a logfile to be used. Specifically, it modifies ```janus_log_init``` to accept two arguments:

1. ```gboolean console``` (whether the output should be printed on stdout)
2. ```const char *logfile``` (whether the output should be saved to that log file)

and I modified the printing thread so that it selectively prints to what has been configured. It seems to be working great, and that's thanks to your approach which proved quite adaptable for the purpose! (the same string can be printed to different targets when iterating).

Since making a fork of a fork seems a bit silly, if you guys feel the current status of the patch is fine I can merge it right away and then apply my changes over it. As an alternative, I can send you the files I modified so that you can update your pull request accordingly.

ploxiln:
Nah, you can keep your long log lines, no need to add a truncate option. I patch out pretty much all log lines that could go over ~500 characters in my tweaks branch anyway :) (SDPs in particular are easily looked up in the client, or the janus admin interface).

It's conventional for a an output file of `-` to mean stdout, so you might not need a separate boolean.

ploxiln:
About your changes to this, why not offer the name of a branch that @jayridge could merge into his if he wanted to continue on it.

ploxiln:
Oh, I didn't realize jay had made further changes, it looks like he's about done now, but give me a bit of time to review.

lminiero:
Ok, keep me posted!

jayridge:
After reviewing the glib code in more detail, the printf routines are less than ideal. They degenerate into g_vasprintf which mallocs, sprintfs, ( possibly ) resizes the gstring buffer, memcpy then frees. If you believe all target platforms support a true C99 snprintf - that return needed buffer size when passing a NULL buffer and size of 0 - that is probably the way to go.

ploxiln:
I'm pretty sure all target platforms support snprintf()

(janus doesn't run on Windows, but incidentally Microsoft just added support for snprintf() in VS2015, so consider it equivalent to hell freezing over and the least likely platform adding support)

log.c
```diff
@@ -0,0 +1,158 @@
...
+					bufferpool = b;
+				}
+			}
+			g_mutex_unlock(&lock);
+			fflush(stdout);
+			freebuffers(&tofree);
+		}
+	}
+	/* print any remaining messages, stdout flushed on exit */
+	for (b = printhead; b; b = b->next) {
```
ploxiln: You might want to take the lock one last time; other threads could print some messages themselves due to noticing `stopping` and cleaning up before exiting.

ploxiln: eh, scratch that, of course this is a private stopping, and of course logging shutdown is the very last thing that happens before janus exits, so this is fine :smile:

ploxiln:
Thread locking and linked list stuff looks great, very nice and fairly simple.

I might allow more than 8 free buffers (I was worried about hundreds :)

One advantage to always using a *macro* around some standard print-related varargs function (instead of a properly implemented varargs c function yourself) is that you can get warnings from the compiler when your format string conversions don't match the arguments. I haven't actually tested this branch myself, but IIRC you need a gcc-specific annotation to get that with your own varargs function.

lminiero:
How do you guys think we should proceed at this point? Start merging what we have and use it and then update it along the road (snprintf stuff) to improve it, or fix it before that?

jayridge:
I have an update that uses snprintf to get the necessary size and reallocs as necessary. Let me do a little testing and I will submit. If you have a branch I can work on I am happy to post it there.

lminiero:
Thanks for this! Feel free to keep on using this PR, I'll apply the file-based logs and the code style when it's time, no need to hurry.

jayridge:
Latest update favours plain old malloced buffers over GStrings. This should avoid the extra malloc/free from the glib printf routines. This one definitely needs some looking over.

ploxiln:
I'm not sure how, but format string warnings still work, my concerns about that were unfounded.

I've looked this over and also tested it, looks good to me.

lminiero:
Ok, I'll give it a look as well and in case merge it to add the rest. Thanks again for the great work!

lminiero:
Looks great! Merging right away, I'll commit my changes in a few minutes as well.

lminiero:
You can find the changes I made in this commit: https://github.com/meetecho/janus-gateway/commit/ccd7cadaafcc2ad51276f0057f9077708be80233
Apart from a few code style changes (which I hope won't make @jayridge angry! :smile: ) the only changes are a fix to the postprocessor (which needed the log.c object and the init/destroy) and the possibility to redirect to a log file as well. I'll work on ways to configure this behaviour in config and/or command line, so that the existing daemonization can actually be used.

jayridge:
Thank you @lminiero. I am happy to ( finally ) contribute back to a great project.

PR #392, opened 2015-11-25
janus_process_error(): use buf on stack

### LINKED ISSUE #391: Memory leak on line 2127 in janus.c file
Hi guys,

please, fix memory leak in janus_process_error function:
https://github.com/meetecho/janus-gateway/blob/master/janus.c#L2127

notice, you allocated memory for string, but not freed one

Thanks,
Ivan

ploxiln:
avoid leaking allocated error string buf to fix #391


lminiero:
Thanks, merging!

PR #393, opened 2015-11-25
simplify log thread - printing and allocation

ploxiln:
some follow up suggestions to #389 and ccd7cadaafcc2ad51276f0057f9077708be80233
 - @jayridge @lminiero please feel free to argue against or reject, and I can easily trim this

```
logging: let janus_log_file be stdout if console or "-"
  - avoid conditionals for every logging write
```

```
logging: simplify buffer sizing

no need to try to optimize allocations by rounding up to a power of two
  - malloc implementations already do that internally if appropriate (up to a certain size)
  - malloc implementations usually include metadata just before or after the allocation,
    which this logic does not take into account, and thus likely rounds up to the *next* size

avoid double call to vsnprintf() in the very common case that the buffer is already sufficient
```

### DIFF
--- a/log.c
+++ b/log.c
@@ -26,9 +26,7 @@ struct janus_log_buffer {
 	char str[1];
 };
 
-#define INITIAL_BUFSZ		1024*2
-#define BUFFER_STRSZ(b)		(b ? b->allocated - sizeof(*b) : 0)
-#define BUFFER_ALLOCSZ(r)	(r + sizeof(janus_log_buffer))
+#define INITIAL_BUFSZ		2000
 
 static gboolean janus_log_console = TRUE;
 static char *janus_log_filepath = NULL;
@@ -39,7 +37,7 @@ static gint stopping = 0;
 static gint poolsz = 0;
 static gint maxpoolsz = 32;
 /* Buffers over this size will be freed */
-static gsize maxbuffersz = 1024*8;
+static size_t maxbuffersz = 8000;
 static GMutex lock;
 static GCond cond;
 static GThread *printthread = NULL;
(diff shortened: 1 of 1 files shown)


jayridge:
cleaner. not certain why i was concerned about vsnprintf doing the right thing with a small buffer. as for power of 2, i suppose that is a holdover from the 90s.

lminiero:
Actually the whole point of having console and logfile as different arguments was to allow people to enable either or both (e.g., still interested in console output but also want a log file or just want stdout vs. logfile).

lminiero:
FYI, I made some changes today to customize (via cmd line and/or config) the logging and daemonization: https://github.com/meetecho/janus-gateway/commit/f82a860522799f0312e905e1ba42c3971ea305e7
It still assumes stdout and logfile can be configured separately, so your new changes may need to be adapted a bit.

ploxiln:
Eh well I tried ;)  It was easy enough to just drop the logfile/console affecting commit.

lminiero:
Cool! I'll give it a look tomorrow morning and merge in case. Thanks again for the tons of help on this!

lminiero:
Ops, forgot about this, sorry! Looking into this right now.

lminiero:
Looks fine to me. I only had doubts about the double ```vsnprintf```, but I didn't know that when the buffer is too small it still returns the number of bytes that would have been written if it were large enough, so it works.

Just let me know if you're still making changes on this or if it's safe to merge, thanks!

ploxiln:
I had no further changes planned here, thanks :)

lminiero:
Merging then, thanks again!

Issue #316, opened 2015-08-28
double free or corruption (fasttop)


ancorgs:
We have been experiencing several (apparently different) stability problems and very high CPU usage issues in Janus lately. For many of them, I don't have further information, but the latest crash left this on the logs.

```
*** Error in `/usr/bin/janus': double free or corruption (fasttop): 0x00007f6ea8d37e70 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7283f)[0x7f6ebe3f883f]
/lib64/libc.so.6(+0x780ae)[0x7f6ebe3fe0ae]
/lib64/libc.so.6(+0x78db6)[0x7f6ebe3fedb6]
/usr/bin/janus(janus_sdp_parse+0x6e8)[0x453a58]
/usr/bin/janus(janus_process_incoming_request+0x3fb0)[0x43ddf0]
/usr/bin/janus(janus_wss_task+0x32e)[0x441c0e]
/usr/lib64/libglib-2.0.so.0(+0x707cc)[0x7f6ec07167cc]
/usr/lib64/libglib-2.0.so.0(+0x6fe15)[0x7f6ec0715e15]
/lib64/libpthread.so.0(+0x80a4)[0x7f6ebe7360a4]
/lib64/libc.so.6(clone+0x6d)[0x7f6ebe46b08d]
======= Memory map: ========
00400000-0046c000 r-xp 00000000 fe:01 5681                               /usr/bin/janus
0066b000-0066c000 r--p 0006b000 fe:01 5681                               /usr/bin/janus
0066c000-0066e000 rw-p 0006c000 fe:01 5681                               /usr/bin/janus
01390000-014d8000 rw-p 00000000 00:00 0                                  [heap]
7f6de66f5000-7f6de66f6000 ---p 00000000 00:00 0
7f6de66f6000-7f6de6ef6000 rw-p 00000000 00:00 0                          [stack:3923]
7f6de6ef6000-7f6de6ef7000 ---p 00000000 00:00 0
7f6de6ef7000-7f6de76f7000 rw-p 00000000 00:00 0                          [stack:3922]
7f6de76f7000-7f6de76f8000 ---p 00000000 00:00 0
7f6de76f8000-7f6de7ef8000 rw-p 00000000 00:00 0                          [stack:3921]
7f6de7ef8000-7f6de7ef9000 ---p 00000000 00:00 0
[...]

PR #394, opened 2015-11-27
Attempt to fix the infamous DTLS decrypt alert error (issues #136 and #316)

lminiero:
This is an attempt to fix the infamous issues #136 and #316, that is the "tlsv1 alert decrypt error" randomly happening when doing handshakes that force Janus to be restarted (issue affecting OpenSSL but NOT BoringSSL, apparently). The cause might be related to race conditions, and in fact [OpenSSL docs](https://www.openssl.org/docs/manmaster/crypto/threads.html) state that:

		"OpenSSL can safely be used in multi-threaded applications
		provided that at least two callback functions are set,
		locking_function and threadid_func."

The fix proposed here is heavily derived from a [discussion related to RTPEngine](http://lists.sip-router.org/pipermail/sr-dev/2015-January/026860.html), where it was mentioned the issue was fixed in [this commit](https://github.com/sipwise/rtpengine/commit/935487b66363c9932684d8085f47450d65a8c37e), which does indeed implement the callbacks the OpenSSL docs suggest.

Please test this estensively with OpenSSL (not BoringSSL) and let us know if this does fix it for you. Mentioning those who experienced the issue in the past in particular (at least those who discussed it on those issues pages), that is @ancorgs @ploxiln @LetsVape @Will-I4M @jacobhub @frk2 @amnonbb @Computician @jing3018 @tuijldert @saghul

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -110,8 +110,66 @@ void *janus_dtls_sctp_setup_thread(void *data);
 #endif
 
 
+/*
+ * FIXME DTLS locking stuff to make OpenSSL thread safe
+ *
+ * Note: this is an attempt to fix the infamous issue #316:
+ * 		https://github.com/meetecho/janus-gateway/issues/316
+ * that is the "tlsv1 alert decrypt error" randomly happening when
+ * doing handshakes that force Janus to be restarted (issue affecting
+ * OpenSSL but NOT BoringSSL, apparently). The cause might be related
+ * to race conditions, and in fact OpenSSL docs state that:
+ *
+ * 		"OpenSSL can safely be used in multi-threaded applications
+ * 		provided that at least two callback functions are set,
+ * 		locking_function and threadid_func."
+ *
+ * See here for the whole docs:
+ * 		https://www.openssl.org/docs/manmaster/crypto/threads.html
+ *
+ * The fix proposed here is heavily derived from a discussion related to
+ * RTPEngine:
+ * 		http://lists.sip-router.org/pipermail/sr-dev/2015-January/026860.html
+ * where it was mentioned the issue was fixed in this commit:
+ * 		https://github.com/sipwise/rtpengine/commit/935487b66363c9932684d8085f47450d65a8c37e
+ * which does indeed implement the callbacks the OpenSSL docs suggest.
+ *
+ */
+static janus_mutex *janus_dtls_locks;
+
+static void janus_dtls_cb_openssl_threadid(CRYPTO_THREADID *tid) {
+	/* FIXME Assuming pthread, which is fine as GLib wraps pthread and
+	 * so that's what we use anyway? */
+	pthread_t me = pthread_self();
+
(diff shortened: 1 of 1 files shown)


ploxiln:
It'll be a while before I get an appropriate testing version out and have it stewing for long enough to see for sure, but I think this is the right approach (and this specific implementation looks good to me). Much appreciated :smile:

LetsVape:
Awesome, good find on the discussion, thanks for working on this Lorenzo
I've compiled this right away and started testing with 30+ users

ancorgs:
I have just deployed it. We are planning a heavy load this week. I will let you know how it behaves.

LetsVape:
I have been running this for 110 hours and haven't seen the issue, so it's safe to say things improved, as I couldn't run it for much longer than 2-24 hours until handshakes started locking up with OpenSSL before this pull request

Connections seem to fail less in general, though that might be placebo, people had to retry camming up sometimes but that seems less often now

lminiero:
@LetsVape thanks for the quick feedback, and really excited about the results!
I'll merge the PR then, it's easy enough to get rid of for people who don't want to use it.

LetsVape:
:+1: Thanks for that! Same here, I hope we can keep this process going and beat the BoringSSL uptime :)

ancorgs:
After three days of usage (heavier usage than usual) I have noticed no slow down, I have experienced no crash, I have not needed to restart the service in order for people to connect and there is no trace of the DTLS decrypt error in the logs.

I would call this a huge success. Thanks a lot!!!

lminiero:
Cool, thanks for the update, good to have other feedback that confirms this works!

HouzuoGuo:
congratulations on the resolution!

LetsVape:
I got 29 days and 23 hours out of this, I then got the following core dump, hope it tells you something useful @lminiero , otherwise I'll try again with a newer commit later, maybe the upcoming reference counters will improve things a bit :+1: 

https://gist.github.com/LetsVape/80880c1bdc92411e5018

Thanks!

LetsVape:
No big CPU or memory jumps, dark blue is cached memory used, but that has been high throughout the uptime, the OS should free that up when needed as far as I'm aware 
screen shot 2015-12-27 at 22 44 10

lminiero:
@LetsVape looks like some internal Glib stuff, not really sure how I can debug this (won't happen before a couple of weeks anyway). If it is some kind of race condition, then yes, a polished and stable version of the reference counters branch may help on that.

~30 days without an issue seems like a good improvement though :smile:

LetsVape:
Thanks for the reply, no worries at all, just to make you aware in case this backtrace was useful, as running it in AddressSanitizer for 30 days might be a bit much.
It's indeed a great improvement already, I can deal with a minute downtime per 30 days.
I'll keep an eye on the commits and try those out later so we can try for another 30+ days :+1:

Cheers and happy new year

PR #399, opened 2015-12-02
Added option to create/destroy/check PID file

lminiero:
This patch introduces a way to make Janus create (at startup) and remove (at shutdown) a PID file. If a PID file exists and is locked (flock), or the PID file cannot be created/updated/locked, Janus failes to start. Homemade implementation as ```pidfile_*``` methods are only available using [libbsd](http://libbsd.freedesktop.org/wiki/), and I'd rather avoid adding yet another dependency for something apparently so trivial.

Can be enabled in configuration file (```pid_file``` in ```[general]```) and via command line (```--pid-file``` or ```-p```).

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -215,6 +215,8 @@ static void janus_handle_signal(int signum) {
 static void janus_termination_handler(void) {
 	/* Remove the PID file if we created it */
 	janus_pidfile_remove();
+	/* Close the logger */
+	janus_log_destroy();
 }
 
 
@@ -3762,8 +3764,6 @@ gint main(int argc, char *argv[])
 	}
 
 	JANUS_PRINT("Bye!\n");
-
-	janus_log_destroy();
 
 	exit(0);
 }
--- a/janus.c
+++ b/janus.c
@@ -2939,6 +2939,7 @@ gint main(int argc, char *argv[])
 	gboolean use_stdout = TRUE;
 	if(args_info.disable_stdout_given) {
 		use_stdout = FALSE;
+		janus_config_add_item(config, "general", "log_to_stdout", "no");
 	} else {
 		/* Check if the configuration file is saying anything about this */
 		janus_config_item *item = janus_config_get_item_drilldown(config, "general", "log_to_stdout");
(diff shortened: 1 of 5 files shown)


janus.c
```diff
@@ -3325,10 +3342,12 @@ gint main(int argc, char *argv[])
 	janus_ice_init(ice_lite, ice_tcp, ipv6, rtp_min_port, rtp_max_port);
 	if(janus_ice_set_stun_server(stun_server, stun_port) < 0) {
 		JANUS_LOG(LOG_FATAL, "Invalid STUN address %s:%u\n", stun_server, stun_port);
+		janus_pidfile_remove();
```
saghul: Maybe register a handler with atexit in order to avoid putting the cleanup code everywhere?

lminiero: That's why I love you guys: you always manage to tell me about things I had no idea existed! :smile: 
I'll update the code to do that, thanks!

utils.c
```diff
@@ -305,3 +307,59 @@ char *janus_address_to_ip(struct sockaddr *address) {
...
 }
+
+/* PID file management */
+static char *pidfile = NULL;
+static int pidfd = -1;
+static FILE *pidf = NULL;
+int janus_pidfile_create(const char *file) {
+	if(file == NULL)
+		return 0;
+	pidfile = g_strdup(file);
```
saghul: This variable is only used here and feed at the end, maybe just not copy it altogether?

lminiero: It's freed in another method, ```janus_pidfile_remove```, which is the one we call before leaving Janus. In fact we need to keep the path to the PID file to remove it (```unlink(pidfile)```) at the end.

utils.c
```diff
@@ -305,3 +307,59 @@ char *janus_address_to_ip(struct sockaddr *address) {
...
+	pidf = fdopen(pidfd, "r+");
+	if(pidf == NULL) {
+		JANUS_LOG(LOG_FATAL, "Error opening/creating PID file %s, does Janus have enough permissions?\n", pidfile);
+		close(pidfd);
+		return -1;
+	}
+	/* Try locking the PID file */
+	int pid = 0;
+	if(flock(pidfd, LOCK_EX|LOCK_NB) < 0) {
+		fscanf(pidf, "%d", &pid);
```
saghul: fscanf could fail here

utils.c
lminiero: It's strdup-ed because it might come from the configuration file that we destroy after we used it.

saghul: Ah, true that, I forgot we have to unlink it. Sorry for the noise.

utils.c
lminiero: Yep but we don't care, we just use it to fill in an informational pid integer (if it fails, e.g., empty file, it stays 0).

lminiero:
Seems to be working as expected, merging.

PR #400, opened 2015-12-02
Improve sample configuration

saghul:
- add transports_folder
- document force-bundle and force-rtcp-mux
- add [transports] section


lminiero:
I was pretty sure I had already added that stuff there. I'm wondering if this got lost in some commit, or if I simply forgot to push the updates on the configuration files? Whatever the case, thanks for addressing these, merging!

saghul:
Oh, I got the nice PR number :-)

lminiero:
Calling dibs on 500 and 1000!

Issue #406, opened 2015-12-06
PID file includes mess when debug_level = 7 


idubinets:
Hi guys,

we tried pid file feature today and when debug_level = 7, janus writes some debug traces to it. something like this:

    3911
    (process:3911): libnice-DEBUG: Agent 0x7fbb60009c10 : allocating stream id 1 (0x7fbb6000da70)
    (process:3911): libnice-DEBUG: Agent 0x7fbb60009c10: added relay server [64.71.3.142]:3478 of type 0
    (process:3911): libnice-DEBUG: Agent 0x7fbb60009c10 : In ICE-FULL mode, starting candidate gathering.
    (process:3911): libnice-DEBUG: Agent 0x7fbb60009c10 : libnice compiled without UPnP support
    (process:3911): libnice-DEBUG: Agent 0x7fbb60009c10: Trying to create host candidate on port 0
    (process:3911): libnice-DEBUG: Agent 0x7fbb60009c10: Could not set IPV6 socket ToS
    (process:3911): libnice-DEBUG: Agent 0x7fbb60009c10 : Adding new srv-rflx candidate discovery 0x7fbbb68e2010

Can you please fix that?

lminiero:
As in, Janus is writing debug traces to the PID file?

saghul:
can you try https://github.com/meetecho/janus-gateway/pull/407 ? libnice is probably just printing to stdout, but since we used to close those file descriptors, fd 1 is asigned to the pid file. The PR reopens stdin/out/err to /dev/null, in order to preserve those fds.

lminiero:
Could not replicate it in any of the following scenarios:

1. debug 7 and PID fle specified (```janus -d 7 -p /tmp/janus.pid```);
2. debug 7, PID fle specified and log file (```janus -d 7 -p /tmp/janus.pid -L /tmp/janus.log```);
3. debug 7, PID fle specified, log file and no console output (```janus -d 7 -p /tmp/janus.pid -L /tmp/janus.log -N```).

All worked as expected. I thought that maybe the "no stdout" was causing that (maybe stdout/stderr redirection was to blame and it conflicted with the file descriptor used by the PID file) but it doesn't look like it.

lminiero:
Check https://github.com/meetecho/janus-gateway/commit/365b8c95171537d656138fe21462e3af33d05e5a instead, it includes the fix @saghul made in #407.

idubinets:
Hi guys,

[...]

PR #407, opened 2015-12-07
core: replace std{in,out,err} with /dev/null when daemonizing

saghul:


saghul:
(sorry about the bonus whitespace fix, editors! I can remove that if you want)

lminiero:
s/stout/stdout

saghul:
Fixed.

lminiero:
Ok, waiting for feedback from #406 before merging. In the meanwhile, I guess I should probable disable the automatic libnice logging when debug level is 7: we have a separate option for that, and it's confusing.

saghul:
As a general rule, replacing fds 0-2 with /dev/null when daemonizing is A Good Idea, since some libraries could expect them to point to the real thing.

lminiero:
:+1:

lminiero:
I eventually integrated your fix in another commit that fixes a different issue I encountered: https://github.com/meetecho/janus-gateway/commit/365b8c95171537d656138fe21462e3af33d05e5a
So we can close this one, thanks for the heads up!

saghul:
IMHO that fix is not correct. First, it closes the descriptors anyway, which it shouldn't since fdreopen does that automagically, but more importantly, the daemonizing logic is now scattered around, which makes it harder to follow, and there is no reason to move the logic to the log initialization routine. Reopening stdout actually makes the rest of the softwre transparent, since there exists some stdout, it just happens to point to /dev/null.

lminiero:
> First, it closes the descriptors anyway

Sorry, copy/paste error, will fix this...

> there is no reason to move the logic to the log initialization routine

Only the output redirection is moved, nothing else, and for a specific reason. I was getting errors daemonizing, but nothing was printed, as the logger init uses ```g_print``` and the console output had already been closed, and so I had no idea what was wrong. Disabling that I found out I was doing something wrong with the log paths and could fix that. That's why I decided to only take care of the console output when initializing the logger, which IMHO makes sense as that's the whole point of the logger anyway.

saghul:
Honest question: why not use syslog, and have syslog divert to a file instead of reinventing that?

lminiero:
syslog has been an option I considered about a month ago, but eventually I gave it up for a few reasons, namely debugging level collisions that would have forced me to update the whole code (LOG_WARNING, LOG_INFO, etc. are used by syslog too) and the fact that it logs to somewhere you cannot choose (e.g., journald in Fedora or /var/log/messages on some other systems, rather than a log file of your choice as we wanted).

saghul:
> namely debugging level collisions that would have forced me to update the whole code (LOG_WARNING, LOG_INFO, etc. are used by syslog too)

Prefixing them with JANUS_ would solve it easily.


> and the fact that it logs to somewhere you cannot choose (e.g., journald in Fedora or /var/log/messages on some other systems, rather than a log file of your choice as we wanted).

Janus could have a configurable log local and them have the syslog daemon divert those to a file, this is common practice.

lminiero:
> Prefixing them with JANUS_ would solve it easily

See rewriting the whole code, as they're used everywhere :smile: 

> Janus could have a configurable log local and them have the syslog daemon divert those to a file, this is common practice.

From what I read around this was not so easy, mostly because of the different backend systems that different distros use (e.g., journald vs rsyslog vs whatever) and that would have needed different setups. Not sure this can be done on Fedora, for instance.

Anyway, I think Pierce and Jay did an excellent job on the logger, and I'm happy with the result.

saghul:
> From what I read around this was not so easy, mostly because of the different backend systems that different distros use (e.g., journald vs rsyslog vs whatever) and that would have needed different setups.

This is no different from using syslog-ng vs rsyslog, each syslog software has its own configuration, but configuring the log local is standard.

> Not sure this can be done on Fedora, for instance.

Sure it does. Fedora uses Systemd, so one coud launch Janus and forward its output to a given file from the unit file.

> Anyway, I think Pierce and Jay did an excellent job on the logger, and I'm happy with the result.

Oh, I never implied otherwise!

Issue #410, opened 2015-12-08
run configure failed


langhuihui:
checking for JANUS... configure: error: Package requirements (
                    glib-2.0 >= glib_version
                    nice
                    jansson
                    libssl >= ssl_version
                    libcrypto
                    sofia-sip-ua
                  ) were not met:

Requested 'glib-2.0 >= glib_version' but version of GLib is 2.47.3
Requested 'libssl >= ssl_version' but version of OpenSSL is 1.0.1e

-----------------------
I found 

glib_version=2.32
ssl_version=1.0.1

in the configure file

lminiero:
Do you have the development versions of those libraries installed? Just the runtime won't help.

langhuihui:
Name        : openssl-devel
Arch        : x86_64
Version     : 1.0.1e
Release     : 42.el6
Size        : 2.2 M
Repo        : installed
From repo   : base

-----------
I Compiled Glib from the source code.

lminiero:
.el6? If this is CentOS, that's known to give issues with Janus, especially 6.5.

langhuihui:
But why configure is not passed

langhuihui:
is this code glib-2.0 >= glib_version missing '$'?
maybe the right code like this
 glib-2.0 >= $glib_version

saghul:
Can you pastebin your configure.log?

lminiero:
configure scripts are not created manually, they're created by the autogen script.
Try launching autogen.sh again. Also check what's the result of ```pkg-config --libs --cflags glib-2.0``` as pkg-config is used to find glib.

langhuihui:
[config.txt](https://github.com/meetecho/janus-gateway/files/55115/config.txt)

lminiero:
Please use pastebin.com or gist.github.com instead

langhuihui:
`pkg-config --libs --cflags glib-2.0
-I/usr/local/include/glib-2.0 -I/usr/local/lib/glib-2.0/include  -L/usr/local/lib -lglib-2.0`

I make install the Glib to /usr/local
I set PKG_CONFIG_PATH=/usr/local/lib/pkgconfig

ploxiln:
The key part seems to be:
```
configure:12423: checking for pkg-config
configure:12441: found /usr/bin/pkg-config
configure:12453: result: /usr/bin/pkg-config
configure:12478: checking pkg-config is at least version 0.9.0
configure:12481: result: yes
[...]

PR #411, opened 2015-12-09
configure.ac: ssl_version and glib_version should be shell variables

ploxiln:
otherwise, pkgconfig compares the literal strings ssl_version and glib_version
to the available versions, and for most versions of pkgconfig
(except apparently for the version included in RHEL6) finds the
available versions are "greater than" these literal strings

found by #410 

tested by invoking pkg-config manually:

```
vagrant@dockerdev:~/src/janus-gateway$ PKG_CONFIG_PATH=/opt/janroot/lib/pkgconfig pkg-config --exists --print-errors "
>                     glib-2.0 >= glib_version
>                     nice
>                     jansson
>                     libssl >= ssl_version
>                     libcrypto
>                     sofia-sip-ua
>                   "
vagrant@dockerdev:~/src/janus-gateway$ PKG_CONFIG_PATH=/opt/janroot/lib/pkgconfig pkg-config --exists --print-errors "
                    glib-2.0 >= 2.48        
                    nice
                    jansson
                    libssl >= 1.0.3      
                    libcrypto
                    sofia-sip-ua
                  "
Requested 'glib-2.0 >= 2.48' but version of GLib is 2.40.2
Requested 'libssl >= 1.0.3' but version of OpenSSL-libssl is 1.0.1
```

Also inspected the generated configure script before and after this change.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -77,10 +77,10 @@ AC_ARG_ENABLE([turn-rest-api],
 
 PKG_CHECK_MODULES([JANUS],
                   [
-                    glib-2.0 >= glib_version
+                    glib-2.0 >= $glib_version
                     nice
                     jansson
-                    libssl >= ssl_version
+                    libssl >= $ssl_version
                     libcrypto
                     sofia-sip-ua
                   ])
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for spotting this hidden bug.

PR #413, opened 2015-12-09
Use the bundled adapter.js instead of an external dependency

ancorgs:
This is actually not 100% correct, since the "main" property of a bower.json file should not include two different javascript files, but after investing quite some time testing many alternative approaches, this seems to be our current best option.

If, at some point, Janus goes back to using "vanilla" adapter.js we can bring the dependency back.

@lminiero I know you don't care much about bower support, so if you could please just merge this (I ensure you it's better than before :wink:)...

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "janus-gateway",
-  "version": "0.0.9",
+  "version": "0.1.0",
   "homepage": "https://github.com/meetecho/janus-gateway",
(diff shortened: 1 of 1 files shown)


lminiero:
> @lminiero I know you don't care much about bower support, so if you could please just merge this (I ensure you it's better than before )...

I'm starting to become an open book for you guys! :smile: 
Merging, thanks!

Issue #414, opened 2015-12-10
make failed


langhuihui:
In file included from janus.h:33,
                 from dtls.c:15:
ice.h:413: 错误：typedef‘janus_ice_trickle’重定义
ice.h:152: 附注：‘janus_ice_trickle’的上一个声明在此
In file included from dtls.c:15:
janus.h:111: 错误：typedef‘janus_request’重定义
janus.h:42: 附注：‘janus_request’的上一个声明在此

---
config.log
[https://gist.github.com/langhuihui/cf1fc9857d183659b18c]

---
chinese words mean there are some redefine errors

ploxiln:
It seems your unusual configuration has uncovered another bug which newer build tools don't seem to complain about. I opened a pull request with a potential fix.

PR #415, opened 2015-12-10
fix structs janus_request and janus_ice_trickle being typedef'ed twice

ploxiln:
In addition to this diff, see
  * https://github.com/ploxiln/janus-gateway/blob/ad1cdba29589d024a3b31c3ca36351c5e5e56a1f/ice.h#L152
  * https://github.com/ploxiln/janus-gateway/blob/ad1cdba29589d024a3b31c3ca36351c5e5e56a1f/janus.h#L42

recent gcc doesn't care about it, but gcc-4.4 from RHEL6 does - see #414

### DIFF
--- a/ice.h
+++ b/ice.h
@@ -401,7 +401,7 @@ struct janus_ice_component {
 };
 
 /*! \brief Helper to handle pending trickle candidates (e.g., when we're still waiting for an offer) */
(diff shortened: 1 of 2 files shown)


lminiero:
:+1:

PR #417, opened 2015-12-11
config comment parsing off by one

ploxiln:
the character just before the semicolon would be stripped

I went ahead and rewrote that little part of the comment parsing while I was fixing it.

### DIFF
--- a/config.c
+++ b/config.c
@@ -105,25 +105,15 @@ janus_config *janus_config_parse(const char *config_file) {
 		/* Strip comments */
 		char *line = line_buffer, *sc = line, *c = NULL;
 		while((c = strchr(sc, ';')) != NULL) {
-			if(c == line) {
-				/* Comment starts here */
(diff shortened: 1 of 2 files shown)


config.c
```diff
@@ -105,25 +105,15 @@ janus_config *janus_config_parse(const char *config_file) {
...
-			if(c == line) {
-				/* Comment starts here */
-				*c = '\0';
-				break;
-			}
-			c--;
-			if(*c != '\\') {
+			if(c == line || *(c-1) != '\\') {
 				/* Comment starts here */
 				*c = '\0';
```
ploxiln: the real bug is that this would do `*c = '\0';` just after `c--;` so it would get the character before the semicolon

lminiero:
Good catch: I guess I never noticed this since the character before the semicolon is always a tab or a space in our samples. Merging right away, thanks!

PR #418, opened 2015-12-14
Correctly skip candidates when using bundle.

sgotti:
When using bundle with at least audio and video, video_stream is null since all
the data is flowing throught the first stream (the audio_stream).

So in janus_ice_trickle_parse the value of the video variable is always 0 while
data is incorrectly set to 1.
Having this data == 1 the candidates for the additional medias are correctly
skipped.

If instead HAVE_SCTP isn't defined also data is 0 and the candidates for the
other medias are uncorrectly added to the nice agent.

This also triggers a problem with libnice handling multiple candidates having
the same foundation in the same component (randomly choosing the wrong remote
candidate).

The right fix will be problaby be to save the media type for every
sdpMLineIndex in the ice handle (something like the one just reported in the
FIXME).

This patch fixes this by skipping all the candidates with sdpMLineIndex != 0
since the code already assumes that the bundling is done on all the medias
instead of relying on the values of video and data variables.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -538,10 +538,7 @@ gint janus_ice_trickle_parse(janus_ice_handle *handle, json_t *candidate, const 
 		data = 0;
 #endif
 		if(janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_BUNDLE)
-				&& (
-					((video || data) && handle->audio_stream != NULL) || 
-						((data) && handle->video_stream != NULL))
-					) {
+				&& sdpMLineIndex != 0) {
 			JANUS_LOG(LOG_VERB, "[%"SCNu64"] Got a %s candidate but we're bundling, ignoring...\n", handle->handle_id, json_string_value(mid));
 		} else {
 			janus_ice_stream *stream = video ? handle->video_stream : (data ? handle->data_stream : handle->audio_stream);


sgotti:
If you are interested about the related libnice problem I wrote this [email](http://lists.freedesktop.org/archives/nice/2015-December/001279.html)

lminiero:
> So in janus_ice_trickle_parse the value of the video variable is always 0 while
> data is incorrectly set to 1.

Good catch, thanks for spotting this. I guess that in order for that if to work the right thing we had to do was to AND both data and video with ```!janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_BUNDLE)```. Anyway your proposed fix is simpler and does the trick for the time being. About your other point, 

> The right fix will be problaby be to save the media type for every
> sdpMLineIndex in the ice handle (something like the one just reported in the
> FIXME).

Well, not in the ICE handle but in the stream, actually (the handle abstracts the whole agent). That FIXME comment is there more as a reminder for the fact that we need to support multiple media streams of the same type per PeerConnection (as in, two or more video streams, for instance), while we're currently limited to one. In that case, the assumptions we currenrly make on ```sdpMLineIndex``` would of course be very different. This is part of a planned refactoring we'll start working on early next year, as we want to add support for Unified Plan.

I'll give a check to your patch as soon as I can. By the way, have you signed our [CLA](https://cla.conf.meetecho.com/janus/)? We cannot merge this otherwise... Thanks again!

sgotti:
> I guess that in order for that to work the right thing we had to do was to AND both data and video with !janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_BUNDLE). Anyway your proposed fix is simpler and does the trick for the time being. About your other point,

If I got it correctly your fix seems good.

> Well, not in the ICE handle but in the stream, actually (the handle abstracts the whole agent). That FIXME comment is there more as a reminder of the fact that we need to support multiple media streams of the same type per PeerConnection (as in, two or more video streams, for instance), while we're currently limited to one. In that case, the assumptions we currently make on sdpMLineIndex would of course be very different. This is part of a planned refactoring we'll start working on early next year, as we want to add support for Unified Plan.

Ok!

> I'll give a check to your patch as soon as I can. By the way, have you signed our CLA? We cannot merge this otherwise... Thanks again!

Done!

lminiero:
Thanks! Just checked and it seems to be working fine, merging!

PR #420, opened 2015-12-15
Make DTLS alert and related events more asynchronous

lminiero:
This pull request tries to fix a problem that had been lingering in the background for some time, now.

Plugins can ask the core to tear down a PeerConnection using the ```close_pc``` callback request. This method was conceived so that it would send a DTLS alert to the peer and get rid of the PeerConnection at the core level. Of course, this would also result in a ```hangup_media``` event to the plugin itself, to confirm the PeerConnection was actually closed and trigger the related logic in the plugin.

The problem with the code we had was that the thread invoking ```hangup_media``` was the same that called ```close_pc```. As you can imagine, this could easily lead to trouble: e.g., if the part of the code calling ```close_pc``` and the ```hangup_media``` code had to work on the same mutex, this would lead to a deadlock.

This was caused by the fact that ```janus_plugin_close_pc``` (the core method ```close_pc``` redirects to) called ```janus_ice_webrtc_hangup```, which in turn invoked ```SSL_shutdown```. This OpenSSL method, apparently, also automatically results in a callback to ```janus_dtls_callback``` with SSL_CB_ALERT in the same thread context, and from there to ```hangup_media```, while I thought this would always require an incoming DTLS alert from the peer to happen.

Having all of this happen within the same thread context made little sense, considering we have an ad-hoc thread for all outgoing data (which is why, for instance, a ```relay_rtp``` never blocks: it just adds data to the outgoing queue and immediately returns), so this patch tries to make sending a DTLS alert more in line with that approach. Whenever a DTLS alert needs to be sent, a fake message to address that intention is enqueued, and it's the outgoing data thread that sends the alert. This in turn triggers all the logic that eventually leads to the PeerConnection being removed from the core and the plugin being informed about it. In case no outgoing data thread is available (e.g., ICE not completed yet), it's ```janus_ice_webrtc_hangup``` that stops the ICE loop and not ```janus_plugin_close_pc``` (which now is much "thinner").

I verified this fixed the issue for me, and apparently with no harm done to other parts of Janus. Before merging, I'd like to be sure that's indeed the case, so please test and let me know if it's safe to do so.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1031,7 +1031,7 @@ void janus_ice_webrtc_hangup(janus_ice_handle *handle) {
 		return;
 	if(handle->queued_packets != NULL)
 		g_async_queue_push(handle->queued_packets, &janus_ice_dtls_alert);
-	if(handle->icethread != NULL) {
+	if(handle->icethread == NULL) {
 		/* Get rid of the PeerConnection */
 		if(handle->iceloop) {
 			gint64 waited = 0;
--- a/ice.c
+++ b/ice.c
@@ -265,6 +265,21 @@ gboolean janus_is_rtcp(gchar *buf) {
 }
 
 
+#define JANUS_ICE_PACKET_AUDIO	0
+#define JANUS_ICE_PACKET_VIDEO	1
+#define JANUS_ICE_PACKET_DATA	2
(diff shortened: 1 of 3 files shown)


lminiero:
I see no opposition, merging then.

PR #422, opened 2015-12-16
fix for janus_videoroom_listener leak for janus_videoroom_listener_muxed

mtdxc:
i has find memleaks when i use janus_videoroom_listener_muxed 
it don't free any janus_videoroom_listener , this is the pactch i made for it
can you check if is ok and then merge with master .

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -2158,7 +2158,7 @@ void janus_videoroom_hangup_media(janus_plugin_session *handle) {
 				janus_videoroom_participant *publisher = l->feed;
 				if(publisher != NULL) {
 					janus_mutex_lock(&publisher->listeners_mutex);
-					publisher->listeners = g_slist_remove(publisher->listeners, listener);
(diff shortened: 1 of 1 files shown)


ploxiln:
This forgets to free the GSList nodes. You can just add `g_slist_free(l->listeners)` after you free all the listeners.

(Also the added log line is redundant, I'd remove the original)

lminiero:
Yep, the suggestions made by @ploxiln would definitely improve things.

That said, the muxed listeners part probably needs much more work than just this leak fix, as we never devoted much effort on making it really usable. It will probably become obsolete anyway when we start working on the Unified Plan refactoring early next year, which will make multiple streams supported at the core level itself.

lminiero:
PS: have you signed our [CLA](https://cla.conf.meetecho.com/janus/)? We need this for every contribution by new authors before we can merge.

mtdxc:
yes I have accept the  @ploxiln suggestions and signed CLA. 
please recheck it

lminiero:
Looks fine, thanks :+1:

PR #424, opened 2015-12-16
Add support for partial writes in websockets transport

### LINKED ISSUE #416: [Videoroom] "SyntaxError: JSON.parse: expected ',' or '}' after property value in object" in FF
Hi guys,

I continue playing with Janus and now I am testing video room. Today I faced issue which should be fixed for sure.

I noticed that firefox browser can't parse messages which come from janus. Especially big messages with SDP inside.

I added some log traces to the janus_websockets.c. This is result:
https://gist.github.com/idubinets/a666e02c1ca95825d3ff

notice: 
Sent 4796/4798 bytes

two bytes were not sent!

Thanks,
Ivan

lminiero:
Fixes #416, to test.

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -799,7 +799,7 @@ static int janus_websockets_callback(struct libwebsocket_context *this,
 				/* Check if we have a pending/partial write to complete first */
 				if(ws_client->buffer && ws_client->bufoffset > 0
 						&& !ws_client->destroy && !g_atomic_int_get(&stopping)) {
-					int pending = ws_client->buflen - ws_client->bufoffset;
+					int pending = ws_client->buflen - ws_client->bufoffset - LWS_SEND_BUFFER_POST_PADDING;
 					JANUS_LOG(LOG_VERB, "[WSS-%p] Completing pending WebSocket write (still need to write last %d/%d bytes)...\n",
 						wsi, pending, ws_client->buflen - LWS_SEND_BUFFER_PRE_PADDING);
 					int sent = libwebsocket_write(wsi, ws_client->buffer + ws_client->bufoffset, pending, LWS_WRITE_TEXT);
@@ -823,12 +823,12 @@ static int janus_websockets_callback(struct libwebsocket_context *this,
 					int buflen = LWS_SEND_BUFFER_PRE_PADDING + strlen(response) + LWS_SEND_BUFFER_POST_PADDING;
 					if(ws_client->buffer == NULL) {
 						/* Let's allocate a shared buffer */
-						JANUS_LOG(LOG_WARN, "[WSS-%p] Allocating %d bytes (response is %zu bytes)\n", wsi, buflen, strlen(response));
+						JANUS_LOG(LOG_VERB, "[WSS-%p] Allocating %d bytes (response is %zu bytes)\n", wsi, buflen, strlen(response));
 						ws_client->buflen = buflen;
 						ws_client->buffer = g_malloc0(buflen);
 					} else if(buflen > ws_client->buflen) {
 						/* We need a larger shared buffer */
-						JANUS_LOG(LOG_WARN, "[WSS-%p] Re-allocating to %d bytes (was %d, response is %zu bytes)\n", wsi, buflen, ws_client->buflen, strlen(response));
+						JANUS_LOG(LOG_VERB, "[WSS-%p] Re-allocating to %d bytes (was %d, response is %zu bytes)\n", wsi, buflen, ws_client->buflen, strlen(response));
 						ws_client->buflen = buflen;
 						ws_client->buffer = g_realloc(ws_client->buffer, buflen);
 					}
@@ -839,6 +839,7 @@ static int janus_websockets_callback(struct libwebsocket_context *this,
 					if(sent > -1 && sent < (int)strlen(response)) {
 						/* We couldn't send everything in a single write, we'll complete this in the next round */
 						ws_client->bufoffset = LWS_SEND_BUFFER_PRE_PADDING + sent;
+						JANUS_LOG(LOG_VERB, "[WSS-%p]   -- Couldn't write all bytes, setting offset %d\n", wsi, ws_client->bufoffset);
 					} else {
 						/* We can get rid of the message */
 						g_free(response);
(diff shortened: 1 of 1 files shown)


idubinets:
I like your project and suggesting some ideas to improve it

I want help you some way, but I don't like C development, you guys already added a lot of code and these huge files with ton of lines are driving me crazy :-)

so I will suggest stuff, but you decide what is good what is not ;-)

transports/janus_websockets.c
```diff
@@ -790,17 +794,46 @@ static int janus_websockets_callback(struct libwebsocket_context *this,
...
 			if(!ws_client->destroy && !g_atomic_int_get(&stopping)) {
 				janus_mutex_lock(&ws_client->mutex);
-				/* Shoot all the pending messages first */
+				/* Check if we have a pending/partial write to complete first */
+				if(ws_client->partial_message && ws_client->partial_offset > 0
+						&& !ws_client->destroy && !g_atomic_int_get(&stopping)) {
+					int pending = strlen(ws_client->partial_message) - ws_client->partial_offset;
+					JANUS_LOG(LOG_VERB, "[WSS-%p] Completing pending WebSocket write (still need to write last %d/%zu bytes)...\n",
+						wsi, pending, strlen(ws_client->partial_message));
+					int sent = libwebsocket_write(wsi, (unsigned char *)(ws_client->partial_message + ws_client->partial_offset), pending, LWS_WRITE_TEXT);
```
ploxiln: The partial_message doesn't include the post-padding.
The easy fix is to set `partial_message = buf` instead of `partial_message = response`, below.

But I guess that if less than the response was sent, the 4 bytes in buf after the last byte that was sent are clobbered by the libwebsocket framing, and response needs to be copied to buf again.

I also wonder if it wouldn't be reasonable to pick a maximum size for buf, and make it an array on the stack, to avoid the extra malloc/free.

lminiero: > The partial_message doesn't include the post-padding

That's done on purpose, as I followed this example [here](http://ml.libwebsockets.org/pipermail/libwebsockets/2014-November/001450.html): I guess the idea is that you're explicitly saying, this way, that this is part of a previous message you didn't complete, while padding would make it send a new frame, but not sure about this. Anyway it seemed to work as expected in my tests.

> I also wonder if it wouldn't be reasonable to pick a maximum size for buf, and make it an array on the stack, to avoid the extra malloc/free.

Since libwebsockets is single threaded and so you'd never write twice at the same time, it might be a good idea. I'm just hesitant at setting a maximum size, as in normal cases the only reason you might not send everything are TCP buffers, and so we might want to at least try pass everything we can. On the other hand, a static buffer to re-use would always have the safe padding, and would avoid the risk of writing over parts not allocated in a buffer. I'll give this a thought and a go later.

ploxiln: The example you link to uses the same buffer for the subsequent partial sends as it used for the first attempt, and presumably that buffer `pss-buf` still has the pre-padding and post padding space. In your usage, you use a different buffer for the first write (one with padding) vs for subsequent writes (one without).

You're not really explicitly saying anything by including or not including the padding - it doesn't know whether or not the buffer extends before the start or after the length, it just assumes it does if appropriate.

Now that I think about this interface more, and see that the example does not worry about the tail part of the framing (which goes in the POST_PADDING space) landing in the middle of the message if the write is partial, I think I see how the interface works - multiple write calls can continue the same websocket framed message (it is over tcp after all), the websocket client state must remember if the previous write was incomplete.

lminiero: > Now that I think about this interface more, and see that the example does not worry about the tail part of the framing (which goes in the POST_PADDING space) landing in the middle of the message if the write is partial, I think I see how the interface works - multiple write calls can continue the same websocket framed message (it is over tcp after all), the websocket client state must remember if the previous write was incomplete (and if so, omit the head part of the framing).

Yes, that's what I thought: not sure if it checks this by looking at the beginning of the new write to match it to what it couldn't write before, even though that sounds feeble. I guess I might add a shared buffer for all writes, but if we make it static we're putting a cap on the max size of packets (and possibly reserving much more than needed for the whole duration of the session), and if we make it dynamic we'll have to realloc when needed.

lminiero: Looking at the lws_write code, from what I understood it seems we do not need to have the additional padding if we're continuing a partial write: https://github.com/warmcat/libwebsockets/blob/6d64539fcbab0143903a549430445ff039a3aa52/lib/output.c#L271
There's a flag called ```inside_frame``` that is reset if all the message was sent:
https://github.com/warmcat/libwebsockets/blob/6d64539fcbab0143903a549430445ff039a3aa52/lib/output.c#L491
If it's not, it goes to a different branch of the code, which I guess makes sure that only the bytes are sent. I'll have to do a better check later to see if that's indeed the case.

lminiero:
This new patch uses a shared buffer for outgoing messages, since libwebsockets is single-threaded and so we're never going to use it from different contexts. The buffer is enlarged when needed, and handles partial writes by just making use of an offset variable.

transports/janus_websockets.c
```diff
@@ -924,17 +975,54 @@ static int janus_websockets_admin_callback(struct libwebsocket_context *this,
 			}
 			if(!ws_client->destroy && !g_atomic_int_get(&stopping)) {
 				janus_mutex_lock(&ws_client->mutex);
-				/* Shoot all the pending messages first */
+				/* Check if we have a pending/partial write to complete first */
+				if(ws_client->buffer && ws_client->bufoffset > 0
+						&& !ws_client->destroy && !g_atomic_int_get(&stopping)) {
+					int pending = ws_client->buflen - ws_client->bufoffset - LWS_SEND_BUFFER_POST_PADDING;
```
ploxiln: `ws_client->buflen` is based on the longest response sent so far, not on the length of the current response. So `pending` can be larger than it should be.

lminiero: You're right, what a rookie mistake, I need to keep the response length
somewhere and use that instead... I'll be far from home until Sunday, I'll
fix that when I'm back.
Il 24/dic/2015 03:33, "Pierce Lopez" <notifications@github.com> ha scritto:

lminiero: Pushed a new commit while I'm almost literally with one foot out of the door! Hope I got it right this time, otherwise this will definitely have to wait till Sunday...

transports/janus_websockets.c
```diff
@@ -924,17 +982,57 @@ static int janus_websockets_admin_callback(struct libwebsocket_context *this,
...
+					JANUS_LOG(LOG_VERB, "[AdminWSS-%p]   -- Sent %d/%zu bytes\n", wsi, sent, strlen(response));
+					if(sent > -1 && sent < (int)strlen(response)) {
+						/* We couldn't send everything in a single write, we'll complete this in the next round */
+						ws_client->bufpending = strlen(response) - sent;
+						ws_client->bufoffset = LWS_SEND_BUFFER_PRE_PADDING + sent;
+						JANUS_LOG(LOG_VERB, "[AdminWSS-%p]   -- Couldn't write all bytes (%d missing), setting offset %d\n",
+							wsi, ws_client->bufpending, ws_client->bufoffset);
+					} else {
+						/* We can get rid of the message */
+						g_free(response);
```
ploxiln: since response was already copied into buffer, and you'll complete partial writes with buffer, I think you should free "response" message in all cases

lminiero: :+1:

lminiero:
Merging as, dumb mistakes aside (which have been fixed), this seems to be fixing the original broken scenario.

PR #426, opened 2015-12-16
fix leak of component (timeout) source

### LINKED ISSUE #421: possible leak of janus_ice_component GSource
I was valgrind-ing again and noticed this, after "cleanly" ending all sessions and stopping janus-server:

```
==8708== 680 (520 direct, 160 indirect) bytes in 5 blocks are definitely lost in loss record 661 of 680
==8708==    at 0x4C2CC70: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==8708==    by 0x532E668: g_malloc0 (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4002.0)
==8708==    by 0x5326C36: g_source_new (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4002.0)
==8708==    by 0x53299B8: g_timeout_source_new (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4002.0)
==8708==    by 0x41BB4A: janus_ice_cb_new_selected_pair (ice.c:1437)
==8708==    by 0x720CADB: ffi_call_unix64 (in /usr/lib/x86_64-linux-gnu/libffi.so.6.0.1)
==8708==    by 0x720C40B: ffi_call (in /usr/lib/x86_64-linux-gnu/libffi.so.6.0.1)
==8708==    by 0x509FBF7: g_cclosure_marshal_generic (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.4002.0)
==8708==    by 0x509F3B7: g_closure_invoke (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.4002.0)
==8708==    by 0x50B0D3C: signal_emit_unlocked_R (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.4002.0)
==8708==    by 0x50B7BF0: g_signal_emitv (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.4002.0)
==8708==    by 0x4E435EF: agent_unlock_and_emit (in /opt/janus-gateway/lib/libnice.so.10.6.2)
```

Considering I can't find any leaks of the component itself or other members of it, it looks unlikely to me that the code forgets to free `component->source`, and more likely that `janus_ice_cb_new_selected_pair()` is called twice for a component, and re-assigns to a non-NULL `source` with `g_timeout_source_new()`.

In fact, `janus_ice_cb_new_selected_pair()` has code to free an existing `component->selected_pair`, but not for `component->source`. Does that make sense?

ploxiln:
attempt to fix #421 - seems to work for me, but I'm not familiar with stuff, and don't know if it's correct in all cases

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -760,8 +760,7 @@ gboolean janus_dtls_retry(gpointer stack) {
 		JANUS_LOG(LOG_VERB, "[%"SCNu64"]  DTLS already set up, disabling retransmission timer!\n", handle->handle_id);
 		if(component->source != NULL) {
 			g_source_destroy(component->source);
-			if(G_IS_OBJECT(component->source))
-				g_object_unref(component->source);
+			g_source_unref(component->source);
 			component->source = NULL;
 		}
 		return FALSE;
--- a/ice.c
+++ b/ice.c
@@ -1169,8 +1169,7 @@ void janus_ice_component_free(GHashTable *components, janus_ice_component *compo
 	component->stream = NULL;
 	if(component->source != NULL) {
 		g_source_destroy(component->source);
-		if(G_IS_OBJECT(component->source))
-			g_object_unref(component->source);
+		g_source_unref(component->source);
 		component->source = NULL;
 	}
 	if(component->dtls != NULL) {


lminiero:
That's most definitely the right approach, as it's what this sample does as well:
http://es.gnu.org/~aleksander/glib/test-gsource.c

Not sure why I used the ```g_object_unref``` instead of ```g_source_unref```: I think I found it in some tutorial or example and replicated the same thing, but ```g_source_unref``` is surely the method to use. Apparently, we were [not the only ones](https://sigquit.wordpress.com/2010/01/05/g_source_unref-and-g_source_destroy-are-your-friends/) to be confused about this :smile:

ploxiln:
thanks for reviewing :)

PR #427, opened 2015-12-17
skip check/warn about local ip and no stun server if nat_1_1_mapping set

ploxiln:
a subjective cosmetic change, feel free to reject

(another idea is, if nat_1_1_mapping is set, check that to see if it is a local_ip. I'd be willing to make that change in this branch)

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -3273,6 +3273,7 @@ gint main(int argc, char *argv[])
 	uint16_t stun_port = 0, turn_port = 0;
 	char *turn_type = NULL, *turn_user = NULL, *turn_pwd = NULL;
 	char *turn_rest_api = NULL, *turn_rest_api_key = NULL;
+	const char *nat_1_1_mapping = NULL;
 	uint16_t rtp_min_port = 0, rtp_max_port = 0;
 	gboolean ice_lite = FALSE, ice_tcp = FALSE, ipv6 = FALSE;
 	item = janus_config_get_item_drilldown(config, "media", "ipv6");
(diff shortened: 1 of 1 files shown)


lminiero:
The change makes sense: in that case, though, I agree that verifying whether ```nat_1_1_mapping``` is private would be a good idea, since we'd be skipping the test there. As an alternative, we might restore the ```if(stun_server == NULL && turn_server == NULL) {``` and use ```nat_1_1_mapping``` rather than che local IP when doing the "private IP" check.

lminiero:
(or do both!)

lminiero:
A simple way to do that could be to change the test we have like this:

	if(stun_server == NULL && turn_server == NULL) {
		/* No STUN and TURN server provided for Janus: make sure it isn't on a private address */
		gboolean private_address = FALSE;
		struct sockaddr_in addr;
		if(inet_pton(AF_INET, nat_1_1_mapping ? nat_1_1_mapping : local_ip, &addr) > 0) {
			unsigned short int ip[4];
			sscanf(nat_1_1_mapping ? nat_1_1_mapping : local_ip, "%hu.%hu.%hu.%hu", &ip[0], &ip[1], &ip[2], &ip[3]);
			if(ip[0] == 10) {
				/* Class A private address */
				private_address = TRUE;
			} else if(ip[0] == 172 && (ip[1] >= 16 && ip[1] <= 31)) {
				/* Class B private address */
				private_address = TRUE;
			} else if(ip[0] == 192 && ip[1] == 168) {
				/* Class C private address */
				private_address = TRUE;
			}
		}
		if(private_address) {
			JANUS_LOG(LOG_WARN, "Janus is deployed on a private address (%s) but you didn't specify any STUN server! Expect trouble if this is supposed to work over the internet and not just in a LAN...\n", nat_1_1_mapping ? nat_1_1_mapping : local_ip);
		}
	}

That is, basically replacing all occurrences of ```local_ip``` with ```nat_1_1_mapping ? nat_1_1_mapping : local_ip``` in the check, so that if ```nat_1_1_mapping``` was passed it is used for the local IP check instead.

ploxiln:
I pushed up pretty much what you described.

lminiero:
Cool, thanks! I'll merge this right away.

PR #429, opened 2015-12-22
new valgrind suppressions file

ploxiln:
The valgrind suppressions file `janus.supp` didn't catch everything for me (though it only missed a handful), and it was long and difficult to refine incrementally, so I started again from scratch. I tried to find the right balance between being too-lenient and too-specific. This version works well for me and is shorter, but I don't enable all janus plugins and features (and it's worth mentioning that I only tested with glib2-2.40.2 from ubuntu-14.04, and libnice-0.1.13).

(I intentionally left one leak un-suppressed, some thread-local-storage leak due to the glib2 thread pool for transport requests. I haven't fully given-up on fixing it yet :)

Some other ideas about this file:

If you don't use this, nor work directly with someone who does, it might be reasonable to remove from the repo and say developers who use valgrind should create their own, or refer to a gist for an example.

Or, it might make sense to rename this file to something like `valgrind.supp` or `janus-valgrind.supp` to make it clear to others what the file is for.

### DIFF
--- a/janus.supp
+++ b/janus.supp
@@ -1,242 +1,141 @@
+# annoying glib2 object type registration stuff
 {
-	gnutls-init-calloc
+	glib2-type-register-fundamental
 	Memcheck:Leak
 	fun:calloc
+	fun:g_malloc0
 	...
-	fun:gtls_gnutls_init
-}
-
-{
-	gnutls-init-realloc
-	Memcheck:Leak
-	fun:realloc
+	fun:g_type_register_fundamental
 	...
-	fun:gtls_gnutls_init
 }
-
 {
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for updating that file, it was indeed very old. I remember collecting some suppression info at the time, but that was before I came to the knowledge that you could have Valgrind itself generate suppression directives for you using ```--gen-suppressions=all```. I always meant to collect those and do a proper one, but you beat me to it, which I'm of course very happy about! In case more is needed for specific plugins I'll take care of that.

I agree that the file should be renamed to make its purpose clearer, I'll do this after merging, thanks!

PR #430, opened 2015-12-24
janus_ice_send_thread(): use g_async_queue_timeout_pop()

ploxiln:
Instead of g_usleep() for 10ms when queue is empty, wait up to 500ms for a message on the queue. 500ms timeout is needed so receive silence over 1 second period is still detected. (It might end up not checking until almost 1.5 seconds since the previous check, but I think that's OK.)

No rush to merge this, I've only done a bit of testing so far, and this will go out for much more testing on some of our servers in a few days.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -2911,11 +2911,16 @@ void *janus_ice_send_thread(void *data) {
 	janus_ice_handle *handle = (janus_ice_handle *)data;
 	JANUS_LOG(LOG_VERB, "[%"SCNu64"] ICE send thread started...\n", handle->handle_id);
 	janus_ice_queued_packet *pkt = NULL;
-	gint64 now = janus_get_monotonic_time(), before = now;
+	gint64 before = janus_get_monotonic_time();
 	while(!janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_ALERT)) {
-		now = janus_get_monotonic_time();
+		if(handle->queued_packets != NULL) {
+			pkt = g_async_queue_timeout_pop(handle->queued_packets, 500000);
+		} else {
+			g_usleep(100000);
+		}
+		/* First of all, let's see if everything's fine on the recv side */
+		gint64 now = janus_get_monotonic_time();
 		if(now-before >= G_USEC_PER_SEC) {
-			/* First of all, let's see if everything's fine on the recv side */
 			if(handle->audio_stream && handle->audio_stream->rtp_component) {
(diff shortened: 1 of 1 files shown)


ploxiln:
I've been running this for a while, and I've seen a noticeable improvement in context switching. Just for fun, here are some graphs.

janus31 is not using this patch:
interrupts-day-janus31

janus34 is using this:
interrupts-day-janus34

(As Lorenzo knows) we have a videoroom like plugin with "captures" and "viewers", and the below graph shows that there are a comparable number of clients of each type connected to these two servers over this period of time:
screen shot 2015-12-27 at 17 43 24

lminiero:
That's a very welcome improvement, thanks for working on this!
I'm away from my working pc now (and have been away from home these last few days, so sorry for this late response), I'll merge this into the main branch as soon as I get back to it.

ploxiln:
I was just posting stuff up while it's fresh. Enjoy your vacation :)

lminiero:
> 500ms timeout is needed so receive silence over 1 second period is still detected. (It might end up not checking until almost 1.5 seconds since the previous check, but I think that's OK.)

How does it go from 500ms to only wake up 1.5s later? Is ```g_async_queue_timeout_pop``` so unreliable/vague?

ploxiln:
I mean, if it wakes up 999ms after it last checked for "no data", it won't check. Then, it could be another 500ms until it checks again.

(but in the typical case it'll wake up 33ms or so later when the next packet is sent)

lminiero:
Ah, got it: is this why you added the ```g_usleep(100000);``` if no message was received after 500ms? To make it more likely the >1s check triggers sooner rather than later? Since the patch is relying on signals, isn't it better to remove the sleep entirely, and just let the check go through and maybe fail (I guess it's really not important if you no-media event is sent after 1s or 1.5s of inactivity), in case you start receiving packets immediately after that, and that you'd only handle 100ms later?

ploxiln:
The `g_usleep(100000)` is just for the case that the g_async_queue is itself NULL, which I'd think should be impossible, but it's a case that the original code handled. So it avoids busy looping in his possibly impossible situation, and it avoids me having to figure out if it's really impossible or not, for this not-directly-related change :)

lminiero:
You're right, sorry... it's in the ```if(handle->queued_packets != NULL) {``` else, not in the "packet is NULL" one... I should really stop looking at code during the holidays! :smile:

lminiero:
Finally found some time to look into this and it seems fine, thanks, merging!

PR #432, opened 2015-12-30
fixbug for videoroom message handler thread

jing3018:
fixbug for videoroom message handler thread, the pointer `msg->handle` may freed in janus core.
I have 70+ participants in the videoroom。 the videoroom message handler thread process message will delay more then 3 seconds。in that condition the pointer `msg->handle` may freed in janus core, Janus will crash.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -2193,9 +2193,7 @@ static void *janus_videoroom_handler(void *data) {
 		}
 		janus_videoroom_session *session = NULL;
 		janus_mutex_lock(&sessions_mutex);
-		if(g_hash_table_lookup(sessions, msg->handle) != NULL ) {
-			session = (janus_videoroom_session *)msg->handle->plugin_handle;
-		}
+		session = g_hash_table_lookup(sessions, msg->handle);
 		janus_mutex_unlock(&sessions_mutex);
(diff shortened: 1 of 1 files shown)


lminiero:
This is something that will be fixed when we complete work on reference counters.

Anyway, for the time being, not sure what your patch does. The check you removed is necessary, as ```sessions``` uses the handle as a key: if the session was removed in the meanwhile but the message was handled later by the thread, without that check you might end up with a crash, as you might actually USE the handle that in the first chek you only access as a opaque (harmless) pointer. This is a fix that was introduced lately for exactly that reason. As such, we can't accept that, sorry.

jing3018:
Ok,  reference counters is a perfect solution. 
A janus_plugin_session pointer validate is done in janus core. so this patch works for me.

jing3018:
@lminiero  When can you finish the reference counters. It's a excellent feature.

lminiero:
Well, it's what I've been trying to do these last weeks... it's going to take time to fix all the bugs, as it was a huge change in perspective that affected both core and plugins. If you guys can help me test what's available right now in the related pull request (which is not stable), especially with respect to leaks, race conditions and so on, it will make things faster. Anyway, I won't start working on it again before 1-2 weeks, as I'm on vacation now.

Issue #436, opened 2016-01-04
libwebsocket


interfaith:
the new api will not compile,

ploxiln:
Can you please include what commands you ran to build/install libwebsockets and build janus, and what the configure or make error message was.

(If you include a long log of the compile, please put it in a pastebin/gist and only include the link in the issue comment)

interfaith:
git clone git://git.libwebsockets.org/libwebsockets
cd libwebsockets
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_C_FLAGS="-fpic" ..
make && sudo make install

then configure for janus gives

checking for TRANSPORTS... yes
checking for MHD... yes
checking for libwebsocket_get_internal_extensions in -lwebsockets_shared... no
checking for libwebsocket_get_internal_extensions in -lwebsockets... no
configure: error: libwebsockets not found. See README.md for installation instructions or use --disable-websocket

ploxiln:
If you carefully examine the janus README.md, there's a new argument to cmake for libwebsockets: `-DLWS_WITH_OLD_API_WRAPPERS=1`

ploxiln:
huh that might not work anymore let me check

ploxiln:
yup, removed in https://github.com/warmcat/libwebsockets/commit/11c05bfa09a0d4a71588a6eb8d03aba5e1c48e59

simplest thing to do is checkout tag `v1.5-chrome47-firefox41` of libwebsockets and use that (doesn't require the new cmake define or anything)

interfaith:
one comment from libwebsocket
The name of libwebsocket_get_internal_extensions has been changed to lws_get_internal_extensions in the latest master branch. So janus would have to be changed to use the new updated API.

So you should checkout the latest libwebsockets release instead, that has the old API and build that instead.

interfaith:
download this ? 
http://git.libwebsockets.org/cgi-bin/cgit/libwebsockets/tag/?h=v1.5-chrome47-firefox41

ploxiln:
yup, downloading and using that tarball should work; another option is to add the command `git checkout v1.5-chrome47-firefox41` to the existing commands in the README, just before `cmake ...`

lminiero:
Ah, so they've removed the compatibility layer? Why add it in the first place to then just get rid of it almost immediately? I'll update the README tomorrow to refer to the tagged version.

lminiero:
[...]

Issue #434, opened 2016-01-01
why libwebsockets not found


majorye:
follow the [janus-gatway](https://github.com/meetecho/janus-gateway) readme,  when excute 

```
./configure --prefix=/opt/janus
```

see libwebsockets not found like below screenshot:
![image](https://cloud.githubusercontent.com/assets/36510/12071096/65632c1c-b0d4-11e5-9012-f91c4e5fbc71.png)

OS: Ubuntu 15.10  same as #14

lminiero:
Github is for issues, not questions. Please use the [meetecho-janus](https://groups.google.com/forum/#!forum/meetecho-janus) for those.

PR #438, opened 2016-01-09
upgrade to libwebsockets 1.6

FrozenCow:
janus-gateway did not work with v1.6 of libwebsockets nor the latest master. This PR should solves that.

The changes are as follows:

* Change prefix `libwebsockets_` to `lws_`.
* Remove `libwebsocket_context *this` argument from libwebsockets functions.
* Change 1.5 to 1.6 in readme.

Related to #436 and #434

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ HTTP REST API, you'll have to install it manually:
 
 	git clone git://git.libwebsockets.org/libwebsockets
 	cd libwebsockets
-	git checkout v1.5-chrome47-firefox41
+	git checkout v1.6.0-chrome48-firefox42
 	mkdir build
 	cd build
 	cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_C_FLAGS="-fpic" ..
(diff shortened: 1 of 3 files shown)


lminiero:
Not sure about this, as this forces everybody to upgrade their own version of libwebsockets when they update Janus: besides, apparently they gave up on the wrapper layer as there were breaking changes, which might mean that in the near future we might have to keep track of what might break the websockets module. Thanks for the effort, though, I'll keep note of this for the future.

FrozenCow:
Yes, upgrading now might not be the best timing. Keep this around if you
ever want to skip that grunt work ;)

lminiero:
@FrozenCow just noticed that one of the changes the new version lf libwebsockets introduces is that it doesn't have the context in callbacks anymore. For this reason, the context we have in ```janus_websockets_client``` is never set: it is still being checked in your PR, though, which means the checks involving it will always fail, and would result in websockets to always fail right after they connect.

Is this actually working in your setup? If not, could you update the PR so that the context is removed from ```janus_websockets_client``` and from wherever it's used in the code?

Thanks!

lminiero:
PS: I'll probably work on a way to have the ```configure``` figure out by itself which version of the libwebsockets API you have available, so that the right one is used automatically without the need to stick to one as we do now. Would you be ok in me incorporating your changes in that effort? I couldn't just merge the PR as the old API would still need to be supported, and not just replaced.

lminiero:
Closing as outdated and (due to above mentioned context bit) not working.

PR #442, opened 2016-01-20
Ability to configure virtual host, username, and password for RabbitMQ

hasbean:
Nothing huge, just a requirement I had and thought I'd push the changes.

### DIFF
--- a/conf/janus.transport.rabbitmq.cfg.sample
+++ b/conf/janus.transport.rabbitmq.cfg.sample
@@ -18,6 +18,9 @@
 enable = no					; Whether the support must be enabled
 host = localhost			; The address of the RabbitMQ server
 ;port = 5672				; The port of the RabbitMQ server (5672 by default)
+;username = guest
+;password = guest
+;vhost = /
 to_janus = to-janus			; Name of the queue for incoming messages
 from_janus = from-janus		; Name of the queue for outgoing messages
 
--- a/transports/janus_rabbitmq.c
+++ b/transports/janus_rabbitmq.c
@@ -175,7 +175,7 @@ int janus_rabbitmq_init(janus_transport_callbacks *callback, const char *config_
 		rmqport = atoi(item->value);
     
     /* Credentials and Virtual Host */
-    const char *vhost = NULL, *username = NULL, *password = NULL
+    const char *vhost = NULL, *username = NULL, *password = NULL;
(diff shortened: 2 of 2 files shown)


lminiero:
Whoops, sorry, I saw this, couldn't comment right away and then completely forgot about this! I'll look into it today and come back with comments (if any) ASAP. Thanks!

lminiero:
A couple of quick comments:

1. vhost, username and password should be freed in case something goes wrong, as we do for other parsed config values here, https://github.com/meetecho/janus-gateway/blob/master/transports/janus_rabbitmq.c#L343
2. why are host, username and password only parsed and used when logging in for the Janus API queues? the new RabbitMQ module can be used for the Admin API as well.

hasbean:
1. Whoops, didn't catch that!  I'll add them.

2. That was the plan originally, but the Janus API and Admin API use the same connection, not two separate connections.  These parameters can only be used in amqp_login which is only called once in the code.

lminiero:
> That was the plan originally, but the Janus API and Admin API use the same connection, not two separate connections. These parameters can only be used in amqp_login which is only called once in the code.

Good catch: I thought we actually allowed for different RabbitMQ servers to be used for the two of them, but that's indeed not the case, we just use different queues. I'm wondering now if it makes sense to have such an option... anyway, not relevant to this patch, so please disregard that comment then!

lminiero:
PS: partly related to the point before, though, the ```vhost```, ```username``` and ```password``` part are only parsed if ```enable=yes``` in ```[general]```, that is if the Janus API over RabbitMQ is enabled. It might happen that the Janus API usage for RabbitMQ is disabled while the Admin API is enabled, which would still lead to a login, but with NULL values for ```vhost```, ```username``` and ```password```.

As such, I'd put reading those values from the config outside of the ```enable=yes``` if (maybe before that?), so that it's actually shared for both APIs.

hasbean:
Half awake, but it should be good now.

hasbean:
I'm not really sure what's with the inconsistent indentation.  It looked fine on Xcode but not here.

lminiero:
Thanks (especially for doing this almost asleep eheh), I'll give it a quick test after lunch and in case everything's fine I'll merge it right away.

lminiero:
Merging, thanks!

Issue #408, opened 2015-12-07
Improvements for the daemonzing code


saghul:
I thought I'd first open an issue and discuss here before making a PR.

Currently the daemonizing code has the following issue: when Janus is started, if the process is daemonized, we return 0 almost immediately, even if Janus couldn't actually start.

The way to solve it involves the traditional double-fork plus a pipe over which the granchild would tell its grandparent that it succeeded to start (or not).

So, the initialization would be like:

- parse cmdline and read the config line
- if we have to fork, create a pipe and fork once
- parent wait on the read end of the pipe
- setsid
- fork again, and have the parent exit
- the granchild continues initialization
- after initializing the logger, stdin/out/err are replaced with /dev/null
- pid file is created
- ...
- when Janus is ready, the granchild writes a 0 on the pipe where the grandfather is waiting
- grandparent exits with the exit code received over the pipe

Thoughts?

lminiero:
Sounds reasonable, and would solve the issue you mentioned: I tried moving the daemon code at the end of the setup process, but of course then nothing worked because killing the father I'd kill everything it created in the first place. I didn't know about this double-fork approach which apparently solves the issue.

I have just one doubt: if the initialization causes Janus to crash so that it never writes anything on the pipe, will it freeze the grandfather waiting forever?

saghul:
> I have just one doubt: if the initialization causes Janus to crash so that it never writes anything on the pipe, will it freeze the grandfather waiting forever?

Yes. Note however, that while the grandfather waits for the grandchild to write on the pipe, it's blocked there, as in, the shell hasn't returned the prompt, so you can still ctrl-c it.

lminiero:
> Yes. Note however, that while the grandfather waits for the grandchild to write on the pipe, it's blocked there, as in, the shell hasn't returned the prompt, so you can still ctrl-c it.

I guess we might also put a timed wait there (e.g., select/poll with a timeout) so that we can leave if it looks like it's taking too long.

lminiero:
PR merged, closing.

PR #443, opened 2016-01-21
Have the parent wait for an exit code from the child during startup, when daemonizing

lminiero:
A quick'n'dirty patch to allow for feedback on whether Janus started or not, when daemonizing. A pipe is created, and the parent waits on one end for an exit code before actually exiting: the child either writes a 0 when everything was started correctly, or a 1 when it didn't (currently done in ```janus_termination_handler```).

The patch is much more basic than what @saghul proposed in #408 (no grandparents/grandchildren, for instance, but just the parent/child we already had), but seems to do the job. I'm just puzzled as to how to handle the cases where Janus crashes instead: I thought this would result in the child end of the pipe being closed, and so the poll to exit, but it didn't (the poll keeps waiting), meaning we should handle it accordingly. I guess there are some flags/controls that can be set on the pipe for that, I'll look into it.

Feedback welcome!

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -2995,6 +2995,7 @@ gint main(int argc, char *argv[])
 		}
 		if(pid > 0) {
 			/* Ok, we're the parent: let's wait for the child to tell us everything started fine */
+			close(pipefd[1]);
 			int code = -1;
 			struct pollfd pollfds;
 
@@ -3006,6 +3007,8 @@ gint main(int argc, char *argv[])
 					break;
 				if(res == 0)
 					continue;
+				if(pollfds.revents & POLLERR || pollfds.revents & POLLHUP)
+					break;
 				if(pollfds.revents & POLLIN) {
 					read(pipefd[0], &code, sizeof(int));
 					break;
--- a/janus.c
+++ b/janus.c
@@ -23,6 +23,7 @@
 #include <getopt.h>
 #include <sys/resource.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


janus.c
```diff
@@ -217,6 +223,11 @@ static void janus_termination_handler(void) {
 	janus_pidfile_remove();
 	/* Close the logger */
 	janus_log_destroy();
+	/* If we're daemonizing, we send an error code to the parent */
+	if(daemonize) {
+		int code = 1;
+		write(pipefd[1], &code, sizeof(int));
```
saghul: I'd surround the write with a do - while, which checks if the return code is == -1 and errno is EINTR.

janus.c
```diff
@@ -3690,6 +3729,12 @@ gint main(int argc, char *argv[])
 		exit(1);
 	}
 
+	/* Ok, Janus has started! Let the parent now about this if we're daemonizing */
+	if(daemonize) {
+		int code = 0;
+		write(pipefd[1], &code, sizeof(int));
+	}
+
```
saghul: Same.

janus.c
lminiero: Good catch, thanks for noticing!

lminiero:
This fix (checking POLLERR/POLLHUP) makes the parent aware also when the child crashes, and so fixes the last concern I had. This seems to be working fine for me, so merging as this has been around a while already: there will be time for improving it along the road.

Issue #439, opened 2016-01-17
Video and Audio device changing (videoroom)


leonuh:
Hi there,

I am planning to develop a feature to change the video / audio device (stream) in the videoroom. Do you have any opinion / suggestion / concern?

The basic idea:
- getting the sources through getSources
- extends createOffer with a "forced" parameter - to recreate the offer with the new devices
so prepareWebrtc function would be extended with something like:
```
if(callbacks.forced && config.sdpSent) {
    config.sdpSent = false;
    config.pc = null;
    pluginHandle.send({
        message: { request: 'detachstream' },
        success: function() {
            callbacks.forced = false;
            prepareWebrtc(handleId, callbacks);
        }
    });
    return;
}
```
The "detachstream" would prepare the feed to reattach the new stream (i tried it out and it works - but maybe other things are necessary)
```
 if(participant->sdp) g_free(participant->sdp);
 participant->sdp = NULL;
 participant->audio_active = FALSE;
 participant->video_active = FALSE;
```
- createOffer media's audio / video param would be extended to accept objects(constraint) like:
```
{
    optional: [{
        sourceId: audioDeviceId
    }]
}
```

lminiero:
I don't think this needs any change in the videoroom plugin: as far as the plugin is concerned, you're negotiating a media channel, and it really doesn't care if the video is camera1, camera2 or a screen share. When you want to change the source of the media, there's no need for a detachstream request: you just unpublish, and then publish again, this time with a different source for the media.

Any improvement on the way sources are detected and used on the JavaScript side is more than welcome instead!

lminiero:
I think we can close this?

PR #446, opened 2016-01-24
Feature/videoroom devices

leonuh:
To handle the issue #439

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1303,7 +1303,7 @@ function Janus(gatewayCallbacks) {
 						};
 					}
 					if(typeof media.video === 'object') {
-						videoSupport = media.video;
+						videoSupport = jQuery.extend({}, videoSupport, media.video);
 					}
 					Janus.debug(videoSupport);
 				} else if(media.video === 'screen') {
--- a/html/janus.js
+++ b/html/janus.js
@@ -1223,6 +1223,12 @@ function Janus(gatewayCallbacks) {
 			var constraints = { mandatory: {}, optional: []};
 			pluginHandle.consentDialog(true);
 			var videoSupport = isVideoSendEnabled(media);
+			var audioSupport = isAudioSendEnabled(media);
+			if(audioSupport === true && media != undefined && media != null) {
+				if(typeof media.audio === 'object') {
+					audioSupport = media.audio;
+				}
+			}
 			if(videoSupport === true && media != undefined && media != null) {
 				if(media.video && media.video != 'screen') {
 					var width = 0;
@@ -1296,6 +1302,9 @@ function Janus(gatewayCallbacks) {
 						    'optional': []
 						};
 					}
+					if(typeof media.video === 'object') {
+						videoSupport = media.video;
+					}
 					Janus.debug(videoSupport);
 				} else if(media.video === 'screen') {
 					// Not a webcam, but screen capture
@@ -1320,7 +1329,7 @@ function Janus(gatewayCallbacks) {
 					function getScreenMedia(constraints, gsmCallback) {
 						Janus.log("Adding media constraint (screen capture)");
 						Janus.debug(constraints);
-						getUserMedia(constraints,
+						navigator.mediaDevices.getUserMedia(constraints,
 							function(stream) {
 								gsmCallback(null, stream);
 							},
@@ -1442,13 +1451,12 @@ function Janus(gatewayCallbacks) {
 			// If we got here, we're not screensharing
 			if(media === null || media === undefined || media.video !== 'screen') {
 				// Check whether all media sources are actually available or not
-				// as per https://github.com/meetecho/janus-gateway/pull/114
-				MediaStreamTrack.getSources(function(sources) {
-					var audioExist = sources.some(function(source) {
-						return source.kind === 'audio';
+				navigator.mediaDevices.enumerateDevices().then(function(devices) {
+					var audioExist = devices.some(function(device) {
+						return device.kind === 'audioinput';
 					}),
-					videoExist = sources.some(function(source) {
-						return source.kind === 'video';
+					videoExist = devices.some(function(device) {
+						return device.kind === 'videoinput';
 					});
 
 					// FIXME Should we really give up, or just assume recvonly for both?
@@ -1458,12 +1466,17 @@ function Janus(gatewayCallbacks) {
 						return false;
 					}
 
-					getUserMedia(
-						{audio: audioExist && isAudioSendEnabled(media), video: videoExist ? videoSupport : false},
-						function(stream) { pluginHandle.consentDialog(false); streamsDone(handleId, jsep, media, callbacks, stream); },
-						function(error) { pluginHandle.consentDialog(false); callbacks.error(error); });
+					navigator.mediaDevices.getUserMedia({
+						audio: audioExist ? audioSupport : false,
+						video: videoExist ? videoSupport : false
+					})
+					.then(function(stream) { pluginHandle.consentDialog(false); streamsDone(handleId, jsep, media, callbacks, stream); })
+					.catch(function(error) { pluginHandle.consentDialog(false); callbacks.error(error); });
+				})
+				.catch(function(error) {
+					pluginHandle.consentDialog(false);
+					callbacks.error('enumerateDevices error', error);
 				});
-
 			}
 		} else {
 			// No need to do a getUserMedia, create offer/answer right away
--- a/html/videoroomtest.html
+++ b/html/videoroomtest.html
@@ -107,19 +107,6 @@
 				</div>
 			</div>
 			<div class="hide" id="videos">
-				<div id="devices hide">
-					<form class="form-inline">
-						<div class="form-group" style="margin-right: 20px;">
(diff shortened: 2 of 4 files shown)


html/janus.js
```diff
@@ -1210,6 +1210,12 @@ function Janus(gatewayCallbacks) {
 			var constraints = { mandatory: {}, optional: []};
 			pluginHandle.consentDialog(true);
 			var videoSupport = isVideoSendEnabled(media);
+			var audioSupport = isAudioSendEnabled(media);
+			if(audioSupport === true && media != undefined && media != null) {
+				if(typeof media.audio === 'object') {
+					audioSupport = media.audio;
```
lminiero: Not sure what this block is for?

html/janus.js
```diff
@@ -1283,6 +1289,9 @@ function Janus(gatewayCallbacks) {
 						    'optional': []
 						};
 					}
+					if(typeof media.video === 'object') {
+						videoSupport = media.video;
```
lminiero: This is overwriting what was assigned by janus.js...

html/janus.js
```diff
@@ -1307,7 +1316,7 @@ function Janus(gatewayCallbacks) {
 					function getScreenMedia(constraints, gsmCallback) {
 						Janus.log("Adding media constraint (screen capture)");
 						Janus.debug(constraints);
-						getUserMedia(constraints,
+						navigator.mediaDevices.getUserMedia(constraints,
```
lminiero: We define getUserMedia in adapter.js to account for browser differences, what's this other method about?

html/janus.js
```diff
@@ -1429,13 +1438,12 @@ function Janus(gatewayCallbacks) {
 			// If we got here, we're not screensharing
 			if(media === null || media === undefined || media.video !== 'screen') {
 				// Check whether all media sources are actually available or not
-				// as per https://github.com/meetecho/janus-gateway/pull/114
-				MediaStreamTrack.getSources(function(sources) {
-					var audioExist = sources.some(function(source) {
-						return source.kind === 'audio';
+				navigator.mediaDevices.enumerateDevices().then(function(devices) {
```
lminiero: See above... are you updating adapter.js? I can't see the changes. Are we risking putting browser specific stuff in here that will work, e.g., for Chrome but not Firefox?

html/janus.js
```diff
@@ -1445,12 +1453,17 @@ function Janus(gatewayCallbacks) {
...
-						{audio: audioExist && isAudioSendEnabled(media), video: videoExist ? videoSupport : false},
-						function(stream) { pluginHandle.consentDialog(false); streamsDone(handleId, jsep, media, callbacks, stream); },
-						function(error) { pluginHandle.consentDialog(false); callbacks.error(error); });
+					navigator.mediaDevices.getUserMedia({
+						audio: audioExist ? audioSupport : false,
+						video: videoExist ? videoSupport : false
+					})
+					.then(function(stream) { pluginHandle.consentDialog(false); streamsDone(handleId, jsep, media, callbacks, stream); })
+					.catch(function(error) { pluginHandle.consentDialog(false); callbacks.error(error); });
+				})
```
lminiero: Besides the afore mentioned navigator.mediaDevices point, is the promise-based getUserMedia supported by ALL browser versions around? We didn't start using it yet as there may be versions of Chrome and Firefox who don't support it that we still want to allow to use Janus.

html/videoroomtest.html
```diff
@@ -71,7 +71,7 @@
 					<button class="btn btn-default" autocomplete="off" id="start">Start</button>
 				</h1>
 			</div>
-			<div class="container" id="details">
```
lminiero: Nitpicking here, but why removing the container class from all those elements in the demo? It's bootstrap stuff and I think we need it for layout reasons.

html/janus.js
leonuh: If the user define a custom object to solve the audio device directly (not just the default from boolean):
```
{
    deviceId: {
        exact: 'deviceId'
    }
} 
```

html/videoroomtest.js
```diff
@@ -265,6 +268,45 @@ $(document).ready(function() {
...
+						deviceId: {
+							exact: $('#video-device').val()
+						}
+					});
+				}, 1000);
+			}
+		});	
+	});
+}
+
```
lminiero: Again, if the main purpose of this is to showcase how you can enumerate devices and choose one over another, I wouldn't do this in one of the existing demos. I'd make a new demo page that uses the EchoTest instead (or maybe extend the EchoTest itself, as it's much more evident there as you can see the changes). The VideoRoom demo purpose is to show how you can implement media routing simply, and these additions are distracting IMHO.

html/janus.js
leonuh: Similar as the audio custom constraint definition. If we know the deviceId - we don't need for janus to recognize it

html/janus.js
lminiero: Ah ok, got it, thanks. So this partly answers one of the questions I made below.

html/janus.js
lminiero: But then shouldn't this be BEFORE checking lowres, hires, etc.? This way, we're still autodetecting, and then overwriting.

lminiero:
PS: doing a lot of comments as janus.js is a delicate part of this all, thanks for this effort!

html/janus.js
leonuh: I overrided the adapter.js with the newest one and applied the newer API : https://github.com/webrtc/adapter/blob/master/adapter.js 
Need more test, based on what you support and how would you define the proper fallbacks. But you can find here the basic idea: https://webrtc.github.io/samples/src/content/devices/input-output/

lminiero: Mh, replacing adapter.js is a bit risky as we had some custom changes there, especially with respect to Edge support that this might currently break... I'll have to look into this.

html/videoroomtest.html
leonuh: Container in container. It is not a good practice. You can check the wrappers on the demo site - they are shifted. Inspect it.

html/janus.js
leonuh: It was a missing feature for me. It is on you what do you support and what polyfill would you require. http://caniuse.com/#search=promise

html/videoroomtest.html
lminiero: Thanks for the heads up! Proof enough I DO suck at graphics and HTML :smile:

html/janus.js
lminiero: Thanks for the pointer. This seems reasonably implemented around, although I'm aware, for instance, of older Firefox browsers still around in some Janus deployments. That said, it does make sense to try and be more "up-to-date" on the client side too, and older browsers may be configured by web developers to use an older version of janus.js instead.

leonuh: IMHO: The current frontend stack is pretty stable, but hard to include in new projects easily, besides hard to extend with new features / tests. I don't know what are the plans for the future, but it would be great to see janus frontend side in es6 with npm / webpack or similar structures.

lminiero:
Just a ping to reassure you I haven't forgot about this... just a ton of things to do! I'll get back to this ASAP, hopefully tomorrow.

lminiero:
@leonuh just found some time to play with it, really nice, thanks! I got some multiple "not published" dialogs in a row from time to time, especially after changing device 3-4 times, but I guess that's not something too harmful. I haven't tested whether Edge still works or not with the new adapter, but there will be time for that later.

That said, could you restore the VideoRoom demo to its former state? My plan, after merging, is to move the device enumeration to new method in janus.js, and then use this + the selection you conceived in the echotest instead, as it's more relevant (and immediately apparent, due to the nature of the demo) there. Interested users can then integrate the same approach in their own applications based on other plugins.

Thanks!

leonuh:
Yes, I will. Hopefully on the weekend. Apart from that, will be there an npm version of Janus in the future without jquery dependency?

lminiero:
I'm not an expert of npm, unfortunately, and besides jQuery is the easiest way I know to get effective cross-browser AJAX requests in JavaScript without having to worry about those myself (IIRC we don't use jQuery for anything else, apart from the UI demo code that is).

Should anybody decide to start working on such an effort in the future, maybe as a separate, janus.js-only related project, I'll be glad to reference it in our [resources](https://janus.conf.meetecho.com/docs/resources) documentation page.

lminiero:
@leonuh did you make those changes eventually? I'm asking as I see some commits after our comments, but they all seem to pre-date them. Thanks!

leonuh:
Yes - I pushed those changes based on your request

lminiero:
Thanks, I wasn't sure as the commit is dated 24th of January, but our latest comments were from February instead, so not sure what happened there. I'll have a look at the updated branch: I may have to try and copy/paste some parts around, as janus.js changed a bit in the meanwhile and it looks like it's not immediately mergeable at the moment.

lminiero:
Closing the PR here, as we have another one ready that includes that: let me know if that one satisfies your requirements or if any change is needed. Thanks for your contribution!

html/janus.js
chadfurman: this other method appears to be consistent with devicetest.js

PR #453, opened 2016-02-01
Send DTMF tones using SIP INFO messages

phillcz:
Hi,

this change adds the ability to send DTMF tones using SIP INFO messages as often used and described at https://tools.ietf.org/html/draft-kaplan-dispatch-info-dtmf-package-00.

The current implementation, which sends DTMF inside the audio stream, requires audio analysis on the server side. and may not be reliable.

### DIFF
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -312,12 +312,11 @@ $(document).ready(function() {
 												$('#dtmf').append('<button class="btn btn-info dtmf">*</button>');
 										}
 										$('.dtmf').click(function() {
-											// Send DTMF tone
-											// Inband
-											// sipcall.dtmf({dtmf: { tones: $(this).text()}});
+											// Send DTMF tone (inband)
+											sipcall.dtmf({dtmf: { tones: $(this).text()}});
 
-											// Use SIP INFO messages
-											sipcall.send({"message": {"request": "dtmf_info", "digit": $(this).text()}});
+											// You can also send DTMF tones using SIP INFO
+											// sipcall.send({"message": {"request": "dtmf_info", "digit": $(this).text()}});
 										});
 									}
 									// Show the peer and hide the spinner when we get a playing event
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -313,7 +313,11 @@ $(document).ready(function() {
 										}
 										$('.dtmf').click(function() {
 											// Send DTMF tone
-											sipcall.dtmf({dtmf: { tones: $(this).text()}});
+											// Inband
+											// sipcall.dtmf({dtmf: { tones: $(this).text()}});
+
+											// Use SIP INFO messages
+											sipcall.send({"message": {"request": "dtmf_info", "digit": $(this).text()}});
 										});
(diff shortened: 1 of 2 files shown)


plugins/janus_sip.c
```diff
@@ -1711,6 +1711,49 @@ static void *janus_sip_handler(void *data) {
...
+			const char *digit_text = json_string_value(digit);
+			if(strlen(digit_text) != 1) {
+				JANUS_LOG(LOG_ERR, "Invalid element (digit should be one character))\n");
+				error_code = JANUS_SIP_ERROR_INVALID_ELEMENT;
+				g_snprintf(error_cause, 512, "Invalid element (digit should be one character)");
+				goto error;
+			}
+
+			char payload[64];
+			g_snprintf(payload, sizeof(payload), "Signal=%s\r\nDuration=%d", digit_text, 160);
```
saghul: Why not make the duration also configurable, defaulting to 160 if missing?

html/siptest.js
```diff
@@ -313,7 +313,11 @@ $(document).ready(function() {
 										}
 										$('.dtmf').click(function() {
 											// Send DTMF tone
-											sipcall.dtmf({dtmf: { tones: $(this).text()}});
+											// Inband
+											// sipcall.dtmf({dtmf: { tones: $(this).text()}});
```
saghul: It's probably a good idea to do both, since not all endpoints support INFO.

phillcz:
I've never had the need to change the default value, but it won't hurt to make it configurable. Updated.

saghul:
LGTM :+1:

html/siptest.js
lminiero: Please don't modify the demo, RFC2833 is a standard, SIP INFO DTMF is not (actually, some IETF meetings ago there were discussions about INFO being actually evil! :wink:). Besides, the main purpose of that demo is to also showcase how you can generate DTMF tones via WebRTC. Since the new backend would be there, I'd add a comment after where we currently send DTMF tones (e.g., something like "You can also send DTMF tones via SIP INFO this way" followed by the example) so that users are aware of the functionality.

I'd rather not send them both at the same time, s there are systems that would process them both thus leading to duplicated input and broken applications.

saghul: Agreed :+1: FWIW: https://tools.ietf.org/html/draft-rosenberg-sip-info-harmful-00

lminiero:
Sorry, I didn't notice you had updated the patch, thanks! I'll give it a go and merge if everything looks fine.

lminiero:
Seems to be working fine, I just found a couple of nits to fix:

1. an error is not returned if ```duration``` is provided but is NOT an integer;
2. there's no ```nua_r_info```, and so everything a DTMF is sent via INFO an error is printed on the console (```[ERR] [plugins/janus_sip.c:janus_sip_sofia_callback:2129] Unknown event 38 (nua_r_info)```)

Anyway, those are nits I can take care of myself after the merge. Thanks again!

Issue #450, opened 2016-01-28
Missing the authorization header in SIP message


hoangtuansu:
Hello,

When I test the gateway with OpenCoreIMS, it fails at REGISTER step to P-CSCF because of the missing of authorization field in SIP message. I tried to add some codes in janus_sip.c (just for the sake of getting through REGISTER step) as following but it still not work. 
`...
_static char *authorization_h;_

...

gpointer janus_sip_sofia_thread(gpointer user_data) {

...

_authorization_h = g_strdup("Digest USERNAME=\"alice@open-ims.test\", REALM=\"open-ims.test\", NONCE=\"\", RESPONSE=\"\", URI=\"sip:open-ims.test\"");_

session->stack->s_nua = nua_create(session->stack->s_root,
                                janus_sip_sofia_callback,
                                session,
                                SIPTAG_ALLOW_STR("INVITE, ACK, BYE, CANCEL, OPTIONS"),
                                NUTAG_M_USERNAME(session->account.username),
                                NUTAG_URL(sip_url),
                                _SIPTAG_AUTHORIZATION_STR(authorization_h),_
                                TAG_IF(session->account.sips, NUTAG_SIPS_URL(sips_url)),
                                SIPTAG_USER_AGENT_STR(user_agent),
                                NUTAG_KEEPALIVE(keepalive_interval * 1000),     /* Sofia expects it in milliseconds */
                                NUTAG_OUTBOUND(outbound_options),
                                SIPTAG_SUPPORTED(NULL),
                                TAG_NULL());
...

`
I don't know if this is an issue of gateway or I should check further with sofia-sip (and rebuild the sofia lib ). So please correct me if I'm wrong.

Regards,

lminiero:
There shouldn't be any need to mess with the code, the authorization stuff is handled automatically by the plugin. If you need a different authorization username than the SIP username to register, check the "authuser" option.

That said, it's normal for the first REGISTER message to fail, as the server will reply with a 401 and the stack will provide the authorization string in a new REGISTER that follows.

hoangtuansu:
Hi,

The IMS system always requires the Authorization header in every REGISTER message (you can check the message flow [here](https://www.eventhelix.com/ims/registration/ims_registration.pdf). In the first message, its values may be empty in some fields (e.g. NONCE, RESPONSE) but it must exists, otherwise the P-CSCF will notify the missing of Authorization header, as following (from OpenIMSCore system):

[...]

PR #456, opened 2016-02-02
Optional Authorization header in first REGISTER

lminiero:
This patch tries to address issue #450 by adding two optional attributes in a ```register``` request:

1. ```authheader```, if ```true```, forces an Authorization header in the first REGISTER;
2. ```real``` is a string that MUST be specified if ```authheader``` is ```true```, as without feedback from the server (e.g., 401) there's no way for the plugin to figure out what the realm is going to be by itself.

An example of how to use this in the demo:

    sipcall.send({message: {request: "register", username: "sip:lminiero@example.com", secret: "password", authheader: true, realm: "example.com"}});

which would result in this header being added to the REGISTER:

    Authorization: Digest username="lminiero", realm="example.com", nonce="", response="", uri="sip:example.com"

This shouldn't break anything, especially since it's optional, but I'm making this a PR nevertheless, also to make sure this actually solves @hoangtuansu's use case in a programmatic way.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -1048,6 +1048,38 @@ static void *janus_sip_handler(void *data) {
 				}
 				send_register = json_is_true(do_register);
 			}
+			gboolean add_authheader = FALSE;
+			const char *realm_text = NULL;
+			if(send_register) {
+				/* Check if we need to add an authorization header (apparently some servers expect this) */
+				json_t *authheader = json_object_get(root, "authheader");
+				if(authheader) {
+					if(!json_is_boolean(authheader)) {
+						JANUS_LOG(LOG_ERR, "Invalid element (authheader should be boolean)\n");
+						error_code = JANUS_SIP_ERROR_INVALID_ELEMENT;
+						g_snprintf(error_cause, 512, "Invalid element (authheader should be boolean)");
+						goto error;
+					}
+					add_authheader = json_is_true(authheader);
+					if(add_authheader) {
+						/* We do, make sure we have a realm to provide */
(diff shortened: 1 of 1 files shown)


saghul:
Can't we just ignore IMS altogether? I'm not kidding here.

lminiero:
I don't care about it either (I'm surprised there's people that still do!), but if there's people interested and the only needed fix to get it working is really tiny then I guess there's no harm. I'd of course refuse any major change that tried to address IMS: that may belong to a different plugin entirely, if any.

saghul:
That's my fear, that this patch will open the can of worms :-)

lminiero:
No need to worry, our IMS tolerance ends here! :smile:

hoangtuansu:
Hello,
I'm very sorry for my false report. I've checked again and see that the error is caused due to my mistake. It's because of the value of "authuser", no matter whether or not the authorization header is present in first REGISTER message. 

In case the "authuser" is not specified, the GW uses username (i.e. bob) by default. In OpenIMSCore, this leads to the error "403 Forbidden : HSS user unknown". So I modified client's code adding an additional field (i.e. "authuser" : "bob@open-ims.test") and it works.

Iminiero, could you please discard this change? I'm very sorry for this inconvenience.

lminiero:
@hoangtuansu do you mean that the version of the plugin in master works with OpenIMSCore already, without the need for the header to be present in the first REGISTER too?

hoangtuansu:
Yes, it works already.

lminiero:
Ok, tnx, closing the pull request then.

PR #458, opened 2016-02-03
New transport module (Unix Sockets)

lminiero:
This patch adds a new transport module, specifically for controlling a Janus instance via Unix Sockets as an alternative to the already available HTTP, WebSockets and RabbitMQ transports. This would be particularly useful in contexts where Janus and its controlling application are colocated, and where a dedicated control channel would be more efficient. Credits for the suggestion about the new module and hints go to @saghul!

As of now, you can configure two named server sockets, one for the Janus API and one for the Admin API. When writing a client, you need to specify the ```SOCK_SEQPACKET``` type, as that's what both interfaces make use of. Changing this to something different (e.g., ```SOCK_DGRAM```) would be trivial and I guess could be added to the configuration as well, so if there's interest in that we can do that. There's no plan on adding support for ```SOCK_STREAM```, instead, as we need the automatic message framing that a stream does not provide.

I haven't tested this extensively (just did a few validations), so feedback and bug reports are more than welcome!

### DIFF
--- a/transports/janus_pfunix.c
+++ b/transports/janus_pfunix.c
@@ -184,7 +184,7 @@ int janus_pfunix_init(janus_transport_callbacks *callback, const char *config_pa
 		/* Handle configuration */
 		janus_config_print(config);
 
-		/* First of all Initialize the socketpair for writeable notifications */
+		/* First of all, initialize the socketpair for writeable notifications */
 		if(socketpair(PF_LOCAL, SOCK_STREAM, 0, write_fd) < 0) {
 			JANUS_LOG(LOG_FATAL, "Error creating socket pair for writeable events: %d, %s\n", errno, strerror(errno));
 			return -1;
@@ -403,7 +403,63 @@ void *janus_pfunix_thread(void *data) {
 			break;
 		}
 		for(int i=0; i<fds; i++) {
-			if((poll_fds[i].revents & POLLOUT) == POLLOUT) {
+			if(poll_fds[i].revents & (POLLERR | POLLHUP)) {
+				/* Socket error? Shall we do something? */
+				JANUS_LOG(LOG_ERR, "Error polling %d: %s...\n", poll_fds[i].fd,
+					poll_fds[i].revents & POLLERR ? "POLLERR" : "POLLHUP");
+				if(poll_fds[i].fd == write_fd[0]) {
+					/* Error in the wake-up socketpair, that sucks: try recreating it */
+					JANUS_LOG(LOG_WARN, "Error in the wake-up socketpair, recreating it...\n");
+					close(write_fd[0]);
+					close(write_fd[1]);
+					if(socketpair(PF_LOCAL, SOCK_STREAM, 0, write_fd) < 0) {
+						JANUS_LOG(LOG_FATAL, "Error creating socket pair for writeable events: %d, %s\n", errno, strerror(errno));
+						continue;
+					}
+				} else if(poll_fds[i].fd == pfd) {
+					/* Error in the Janus API socket */
+					JANUS_LOG(LOG_WARN, "Disabling Unix Sockets Janus API interface\n");
+					close(pfd);
+					pfd = -1;
+					continue;
+				} else if(poll_fds[i].fd == admin_pfd) {
+					/* Error in the Admin API socket */
+					JANUS_LOG(LOG_WARN, "Disabling Unix Sockets Admin API interface\n");
+					close(admin_pfd);
+					admin_pfd = -1;
+					continue;
+				} else {
+					/* Error in a client socket, find and remove it */
+					janus_mutex_lock(&clients_mutex);
+					janus_pfunix_client *client = g_hash_table_lookup(clients_by_fd, GINT_TO_POINTER(poll_fds[i].fd));
+					if(client == NULL) {
+						/* We're not handling this, ignore */
+						continue;
+					}
+					JANUS_LOG(LOG_INFO, "Unix Sockets client disconnected (%d)\n", poll_fds[i].fd);
+					/* Notify core */
+					gateway->transport_gone(&janus_pfunix_transport, client);
+					/* Close socket */
+					shutdown(SHUT_RDWR, poll_fds[i].fd);
+					close(poll_fds[i].fd);
+					client->fd = -1;
+					/* Destroy the client */
+					g_hash_table_remove(clients_by_fd, GINT_TO_POINTER(poll_fds[i].fd));
+					g_hash_table_remove(clients, client);
+					if(client->messages != NULL) {
+						char *response = NULL;
+						while((response = g_async_queue_try_pop(client->messages)) != NULL) {
+							g_free(response);
+						}
+						g_async_queue_unref(client->messages);
+					}
+					g_free(client);
+					janus_mutex_unlock(&clients_mutex);
+					continue;
+				}
+				continue;
+			}
+			if(poll_fds[i].revents & POLLOUT) {
 				/* Find the client from its file descriptor */
 				janus_mutex_lock(&clients_mutex);
 				janus_pfunix_client *client = g_hash_table_lookup(clients_by_fd, GINT_TO_POINTER(poll_fds[i].fd));
@@ -421,7 +477,7 @@ void *janus_pfunix_thread(void *data) {
 				}
 				janus_mutex_unlock(&clients_mutex);
 			}
-			if((poll_fds[i].revents & POLLIN) == POLLIN) {
+			if(poll_fds[i].revents & POLLIN) {
 				if(poll_fds[i].fd == write_fd[0]) {
 					/* Read and ignore: we use this to unlock the poll if there's data to write */
 					res = read(poll_fds[i].fd, buffer, BUFFER_SIZE);
--- a/transports/janus_pfunix.c
+++ b/transports/janus_pfunix.c
@@ -96,13 +96,15 @@ janus_transport *create(void) {
 static gint initialized = 0, stopping = 0;
 static janus_transport_callbacks *gateway = NULL;
 
+#define BUFFER_SIZE		8192
+
 
 /* Unix Sockets server thread */
 static GThread *pfunix_thread = NULL;
 void *janus_pfunix_thread(void *data);
 
 /* Unix Sockets servers */
-static int pfd = 0, admin_pfd = 0;
+static int pfd = -1, admin_pfd = -1;
 /* Socket pair to notify about the need for outgoing data */
 static int write_fd[2];
 
@@ -158,10 +160,10 @@ int janus_pfunix_init(janus_transport_callbacks *callback, const char *config_pa
 				pfname[108] = '\0';
 			}
 			/* Create socket */
-			pfd = socket(PF_UNIX, SOCK_SEQPACKET, 0);
+			pfd = socket(PF_UNIX, SOCK_SEQPACKET | SOCK_NONBLOCK, 0);
 			if(pfd < 0) {
 				JANUS_LOG(LOG_FATAL, "Unix Sockets %s creation for Janus API failed: %d, %s\n", pfname, errno, strerror(errno));
-				pfd = 0;
+				pfd = -1;
 			} else {
 				/* Unlink before binding */
 				unlink(pfname);
@@ -174,13 +176,13 @@ int janus_pfunix_init(janus_transport_callbacks *callback, const char *config_pa
 				if(bind(pfd, (struct sockaddr *)&address, sizeof(address)) != 0) {
 					JANUS_LOG(LOG_FATAL, "Bind for Unix Socket %s (Janus API) failed: %d, %s\n", pfname, errno, strerror(errno));
 					close(pfd);
-					pfd = 0;
+					pfd = -1;
 				} else {
 					JANUS_LOG(LOG_VERB, "Listening on Unix Socket %s... (Janus API)\n", pfname);
-					if(listen(pfd, 5) != 0) {
+					if(listen(pfd, 128) != 0) {
 						JANUS_LOG(LOG_FATAL, "Listening on Unix Socket %s (Janus API) failed: %d, %s\n", pfname, errno, strerror(errno));
 						close(pfd);
-						pfd = 0;
+						pfd = -1;
 					}
 				}
 			}
@@ -190,17 +192,17 @@ int janus_pfunix_init(janus_transport_callbacks *callback, const char *config_pa
 		if(!item || !item->value || !janus_is_true(item->value)) {
 			JANUS_LOG(LOG_WARN, "Unix Sockets server disabled (Admin API)\n");
 		} else {
-			item = janus_config_get_item_drilldown(config, "general", "path");
+			item = janus_config_get_item_drilldown(config, "admin", "admin_path");
 			char *pfname = (char *)(item && item->value ? item->value : "/tmp/ux-janusadmin");
 			if(strlen(pfname) > 108) {
 				JANUS_LOG(LOG_WARN, "The provided path name (%s) is longer than 108 characters, it will be truncated\n", pfname);
 				pfname[108] = '\0';
 			}
 			/* Create socket */
-			admin_pfd = socket(PF_UNIX, SOCK_SEQPACKET, 0);
+			admin_pfd = socket(PF_UNIX, SOCK_SEQPACKET | SOCK_NONBLOCK, 0);
 			if(admin_pfd < 0) {
 				JANUS_LOG(LOG_FATAL, "Unix Sockets %s creation for Admin API failed: %d, %s\n", pfname, errno, strerror(errno));
-				admin_pfd = 0;
+				admin_pfd = -1;
 			} else {
 				/* Unlink before binding */
 				unlink(pfname);
@@ -213,13 +215,13 @@ int janus_pfunix_init(janus_transport_callbacks *callback, const char *config_pa
 				if(bind(admin_pfd, (struct sockaddr *)&address, sizeof(address)) != 0) {
 					JANUS_LOG(LOG_FATAL, "Bind for Unix Socket %s (Admin API) failed: %d, %s\n", pfname, errno, strerror(errno));
 					close(admin_pfd);
-					admin_pfd = 0;
+					admin_pfd = -1;
 				} else {
 					JANUS_LOG(LOG_VERB, "Listening on Unix Socket %s... (Admin API)\n", pfname);
-					if(listen(admin_pfd, 5) != 0) {
+					if(listen(admin_pfd, 128) != 0) {
 						JANUS_LOG(LOG_FATAL, "Listening on Unix Socket %s (Admin API) failed: %d, %s\n", pfname, errno, strerror(errno));
 						close(admin_pfd);
-						admin_pfd = 0;
+						admin_pfd = -1;
 					}
 				}
 			}
@@ -227,7 +229,7 @@ int janus_pfunix_init(janus_transport_callbacks *callback, const char *config_pa
 	}
 	janus_config_destroy(config);
 	config = NULL;
-	if(pfd == 0 && admin_pfd == 0) {
+	if(pfd < 0 && admin_pfd < 0) {
 		JANUS_LOG(LOG_FATAL, "No Unix Sockets server started, giving up...\n");
 		return -1;	/* No point in keeping the plugin loaded */
 	}
@@ -239,7 +241,7 @@ int janus_pfunix_init(janus_transport_callbacks *callback, const char *config_pa
 
 	GError *error = NULL;
 	/* Start the Unix Sockets service thread */
-	if(pfd > 0 || admin_pfd > 0) {
+	if(pfd > -1 || admin_pfd > -1) {
 		pfunix_thread = g_thread_try_new("pfunix thread", &janus_pfunix_thread, NULL, &error);
 		if(!pfunix_thread) {
 			g_atomic_int_set(&initialized, 0);
@@ -261,7 +263,11 @@ void janus_pfunix_destroy(void) {
 
 	/* Stop the service thread */
 	char c;
-	write(write_fd[1], &c, 1);
+	int res = 0;
+	do {
+		res = write(write_fd[1], &c, 1);
+	} while(res == -1 && errno == EINTR);
+
 	if(pfunix_thread != NULL) {
 		g_thread_join(pfunix_thread);
 		pfunix_thread = NULL;
@@ -302,11 +308,11 @@ const char *janus_pfunix_get_package(void) {
 }
 
 gboolean janus_pfunix_is_janus_api_enabled(void) {
-	return pfd > 0;
+	return pfd > -1;
 }
 
 gboolean janus_pfunix_is_admin_api_enabled(void) {
-	return admin_pfd > 0;
+	return admin_pfd > -1;
 }
 
 int janus_pfunix_send_message(void *transport, void *request_id, gboolean admin, json_t *message) {
@@ -319,7 +325,7 @@ int janus_pfunix_send_message(void *transport, void *request_id, gboolean admin,
 	/* Make sure this is related to a still valid Unix Sockets session */
 	janus_pfunix_client *client = (janus_pfunix_client *)transport;
 	janus_mutex_lock(&clients_mutex);
-	if(g_hash_table_lookup(clients, client) == NULL || client->fd <= 0) {
+	if(g_hash_table_lookup(clients, client) == NULL || client->fd < 0) {
 		janus_mutex_unlock(&clients_mutex);
 		JANUS_LOG(LOG_WARN, "Outgoing message for invalid client %p\n", client);
 		g_free(message);
@@ -333,7 +339,10 @@ int janus_pfunix_send_message(void *transport, void *request_id, gboolean admin,
 	g_async_queue_push(client->messages, payload);
 	/* Notify the thread there's data to send */
 	char c;
-	write(write_fd[1], &c, 1);
+	int res = 0;
+	do {
+		res = write(write_fd[1], &c, 1);
+	} while(res == -1 && errno == EINTR);
 	return 0;
 }
 
@@ -348,7 +357,7 @@ void janus_pfunix_session_over(void *transport, guint64 session_id, gboolean tim
 	/* FIXME Should we really close the connection in case of a timeout? */
 	janus_pfunix_client *client = (janus_pfunix_client *)transport;
 	janus_mutex_lock(&clients_mutex);
-	if(g_hash_table_lookup(clients, client) != NULL && client->fd > 0) {
+	if(g_hash_table_lookup(clients, client) != NULL && client->fd > -1) {
 		/* Shutdown the client socket */
 		shutdown(client->fd, SHUT_WR);
 	}
@@ -362,7 +371,7 @@ void *janus_pfunix_thread(void *data) {
 
 	int fds = 0;
 	struct pollfd poll_fds[1024];	/* FIXME Should we allow for more clients? */
-	char buffer[4096];
+	char buffer[BUFFER_SIZE];
 
 	while(g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping)) {
 		/* Prepare poll list of file descriptors */
@@ -371,13 +380,13 @@ void *janus_pfunix_thread(void *data) {
 		poll_fds[fds].fd = write_fd[0];
 		poll_fds[fds].events = POLLIN;
 		fds++;
-		if(pfd > 0) {
+		if(pfd > -1) {
 			/* Janus API */
 			poll_fds[fds].fd = pfd;
 			poll_fds[fds].events = POLLIN;
 			fds++;
 		}
-		if(admin_pfd > 0) {
+		if(admin_pfd > -1) {
 			/* Admin API */
 			poll_fds[fds].fd = admin_pfd;
 			poll_fds[fds].events = POLLIN;
@@ -390,19 +399,16 @@ void *janus_pfunix_thread(void *data) {
 		g_hash_table_iter_init(&iter, clients_by_fd);
 		while(g_hash_table_iter_next(&iter, NULL, &value)) {
 			janus_pfunix_client *client = value;
-			if(client->fd > 0) {
+			if(client->fd > -1) {
 				poll_fds[fds].fd = client->fd;
-				poll_fds[fds].events = POLLIN;
-				if(g_async_queue_length(client->messages) > 0) {
-					poll_fds[fds].events = POLLOUT;
-				}
+				poll_fds[fds].events = g_async_queue_length(client->messages) > 0 ? POLLIN | POLLOUT : POLLIN;
 				fds++;
 			}
 		}
 		janus_mutex_unlock(&clients_mutex);
 
 		/* Start polling */
-		int res = poll(poll_fds, fds, 10000);
+		int res = poll(poll_fds, fds, -1);
 		if(res == 0)
 			continue;
 		if(res < 0) {
@@ -414,10 +420,13 @@ void *janus_pfunix_thread(void *data) {
 				/* Find the client from its file descriptor */
 				janus_mutex_lock(&clients_mutex);
 				janus_pfunix_client *client = g_hash_table_lookup(clients_by_fd, GINT_TO_POINTER(poll_fds[i].fd));
-				if(client != NULL && client->fd > 0) {
+				if(client != NULL) {
 					char *payload = NULL;
 					while((payload = g_async_queue_try_pop(client->messages)) != NULL) {
-						int res = write(client->fd, payload, strlen(payload));
+						int res = 0;
+						do {
+							res = write(client->fd, payload, strlen(payload));
+						} while(res == -1 && errno == EINTR);
 						/* FIXME Should we check if sent everything? */
 						JANUS_LOG(LOG_HUGE, "Written %d/%zu bytes on %d\n", res, strlen(payload), client->fd);
 						g_free(payload);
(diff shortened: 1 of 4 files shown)


saghul:
Fantastic! I'll review it ASAP!

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+				address.sun_family = AF_UNIX;
+				g_snprintf(address.sun_path, 108, "%s", pfname);
+				JANUS_LOG(LOG_VERB, "Binding Unix Socket %s... (Janus API)\n", pfname);
+				if(bind(pfd, (struct sockaddr *)&address, sizeof(address)) != 0) {
+					JANUS_LOG(LOG_FATAL, "Bind for Unix Socket %s (Janus API) failed: %d, %s\n", pfname, errno, strerror(errno));
+					close(pfd);
+					pfd = 0;
+				} else {
+					JANUS_LOG(LOG_VERB, "Listening on Unix Socket %s... (Janus API)\n", pfname);
+					if(listen(pfd, 5) != 0) {
```
saghul: Why 5? Not that it's going to be abused but I'd set it to 128, which is the usual default.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+			JANUS_LOG(LOG_WARN, "Unix Sockets server disabled (Janus API)\n");
+		} else {
+			item = janus_config_get_item_drilldown(config, "general", "path");
+			char *pfname = (char *)(item && item->value ? item->value : "/tmp/ux-janusapi");
+			if(strlen(pfname) > 108) {
+				JANUS_LOG(LOG_WARN, "The provided path name (%s) is longer than 108 characters, it will be truncated\n", pfname);
+				pfname[108] = '\0';
+			}
+			/* Create socket */
+			pfd = socket(PF_UNIX, SOCK_SEQPACKET, 0);
```
saghul: After checking the [man page](http://man7.org/linux/man-pages/man7/unix.7.html once more, it mentions that Unix sockets are always reliable, so I think it's better to use SOCK_STREAM, so this can be used / tested with netcat, for example, which does support Unix domain datagram sockets (nc -U -u /tmp/ux-janusapi)

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+					}
+				}
+			}
+		}
+		/* Do the same for the Admin API, if enabled */
+		item = janus_config_get_item_drilldown(config, "admin", "admin_enabled");
+		if(!item || !item->value || !janus_is_true(item->value)) {
+			JANUS_LOG(LOG_WARN, "Unix Sockets server disabled (Admin API)\n");
+		} else {
+			item = janus_config_get_item_drilldown(config, "general", "path");
```
saghul:
~~~~
item = janus_config_get_item_drilldown(config, "admin", "admin_path");
~~~~

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+			char *pfname = (char *)(item && item->value ? item->value : "/tmp/ux-janusadmin");
+			if(strlen(pfname) > 108) {
+				JANUS_LOG(LOG_WARN, "The provided path name (%s) is longer than 108 characters, it will be truncated\n", pfname);
+				pfname[108] = '\0';
+			}
+			/* Create socket */
+			admin_pfd = socket(PF_UNIX, SOCK_SEQPACKET, 0);
```
saghul: Ditto.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+						pfd = 0;
+					}
+				}
+			}
+		}
+		/* Do the same for the Admin API, if enabled */
+		item = janus_config_get_item_drilldown(config, "admin", "admin_enabled");
+		if(!item || !item->value || !janus_is_true(item->value)) {
+			JANUS_LOG(LOG_WARN, "Unix Sockets server disabled (Admin API)\n");
+		} else {
```
saghul: This block is the same as above. How about factoring it out to a function which returns the allocated fd for a given path? We'd call it twice, once for the API socket and another time for the Admin API.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+						close(admin_pfd);
+						admin_pfd = 0;
+					}
+				}
+			}
+		}
+	}
+	janus_config_destroy(config);
+	config = NULL;
+	if(pfd == 0 && admin_pfd == 0) {
```
saghul: I'd use -1 as a marker for invalid fds. If sdtdin is closed it's possible to get that fd here.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+	}
+
+	/* Create a couple of hashtables for all clients */
+	clients = g_hash_table_new(NULL, NULL);
+	clients_by_fd = g_hash_table_new(NULL, NULL);
+	janus_mutex_init(&clients_mutex);
+
+	GError *error = NULL;
+	/* Start the Unix Sockets service thread */
+	if(pfd > 0 || admin_pfd > 0) {
```
saghul: >= (see comment above)

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+}
+
+void janus_pfunix_destroy(void) {
+	if(!g_atomic_int_get(&initialized))
+		return;
+	g_atomic_int_set(&stopping, 1);
+
+	/* Stop the service thread */
+	char c;
+	write(write_fd[1], &c, 1);
```
saghul: *paranoid mode on* use the do-while EINTER loop.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+const char *janus_pfunix_get_author(void) {
+	return JANUS_PFUNIX_AUTHOR;
+}
+
+const char *janus_pfunix_get_package(void) {
+	return JANUS_PFUNIX_PACKAGE;
+}
+
+gboolean janus_pfunix_is_janus_api_enabled(void) {
+	return pfd > 0;
```
saghul: >=

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+}
+
+gboolean janus_pfunix_is_admin_api_enabled(void) {
+	return admin_pfd > 0;
```
saghul: ditto

transports/janus_pfunix.c
lminiero: Good catch, just lazyness on my part, as I copy/pasted the test I had made outside the context of the plugin and forgot to change.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+	if(message == NULL)
+		return -1;
+	if(transport == NULL) {
+		g_free(message);
+		return -1;
+	}
+	/* Make sure this is related to a still valid Unix Sockets session */
+	janus_pfunix_client *client = (janus_pfunix_client *)transport;
+	janus_mutex_lock(&clients_mutex);
+	if(g_hash_table_lookup(clients, client) == NULL || client->fd <= 0) {
```
saghul: maybe first check if the fd is < 0 to avoid looking it up on the hash table?

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+		return -1;
+	}
+	janus_mutex_unlock(&clients_mutex);
+	/* Convert to string and enqueue */
+	char *payload = json_dumps(message, JSON_INDENT(3) | JSON_PRESERVE_ORDER);
+	json_decref(message);
+	g_async_queue_push(client->messages, payload);
+	/* Notify the thread there's data to send */
+	char c;
+	write(write_fd[1], &c, 1);
```
saghul: ditto (do-while dance)

transports/janus_pfunix.c
lminiero: Do you mean SOCK_DGRAM? SOCK_STREAM would give us a stream of bytes, not datagrams/packets, which means we wouldn't get boundaries but a stream of bytes.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+}
+
+void janus_pfunix_session_over(void *transport, guint64 session_id, gboolean timeout) {
+	/* We only care if it's a timeout: if so, close the connection */
+	if(transport == NULL || !timeout)
+		return;
+	/* FIXME Should we really close the connection in case of a timeout? */
+	janus_pfunix_client *client = (janus_pfunix_client *)transport;
+	janus_mutex_lock(&clients_mutex);
+	if(g_hash_table_lookup(clients, client) != NULL && client->fd > 0) {
```
saghul: same as above, reverse the check maybe

transports/janus_pfunix.c
lminiero: Ack.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+}
+
+
+/* Thread */
+void *janus_pfunix_thread(void *data) {
+	JANUS_LOG(LOG_INFO, "Unix Sockets thread started\n");
+
+	int fds = 0;
+	struct pollfd poll_fds[1024];	/* FIXME Should we allow for more clients? */
+	char buffer[4096];
```
saghul: Make this bigger, IIRC the global size for buffers is 8K.

transports/janus_pfunix.c
lminiero: Good point, I'll work on that.

transports/janus_pfunix.c
lminiero: I always forget about that! Thanks for reminding me...

transports/janus_pfunix.c
lminiero: Will do!

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+		GHashTableIter iter;
+		gpointer value;
+		g_hash_table_iter_init(&iter, clients_by_fd);
+		while(g_hash_table_iter_next(&iter, NULL, &value)) {
+			janus_pfunix_client *client = value;
+			if(client->fd > 0) {
+				poll_fds[fds].fd = client->fd;
+				poll_fds[fds].events = POLLIN;
+				if(g_async_queue_length(client->messages) > 0) {
+					poll_fds[fds].events = POLLOUT;
```
saghul: I think in this we would want to listen for both POLLIN and POLLOUT.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+				}
+				fds++;
+			}
+		}
+		janus_mutex_unlock(&clients_mutex);
+
+		/* Start polling */
+		int res = poll(poll_fds, fds, 10000);
```
saghul: why not make it infinite? if we're shutting down we'll wakeup the loop by writing on the socketpair anyway. this would prevent unnecessary wakeups.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+		if(res < 0) {
+			JANUS_LOG(LOG_ERR, "poll() failed\n");
+			break;
+		}
+		for(int i=0; i<fds; i++) {
+			if((poll_fds[i].revents & POLLOUT) == POLLOUT) {
+				/* Find the client from its file descriptor */
+				janus_mutex_lock(&clients_mutex);
+				janus_pfunix_client *client = g_hash_table_lookup(clients_by_fd, GINT_TO_POINTER(poll_fds[i].fd));
+				if(client != NULL && client->fd > 0) {
```
saghul: how would the fd be < 0 if we found the client on the hash table?

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+					char *payload = NULL;
+					while((payload = g_async_queue_try_pop(client->messages)) != NULL) {
+						int res = write(client->fd, payload, strlen(payload));
+						/* FIXME Should we check if sent everything? */
```
saghul: not if we use SOCK_DGRAM, because it's all or nothing, so bonus nachos. But you also want to do the do-while-eintr dance here.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+						client->admin = FALSE;	/* Janus API client */
+						client->messages = g_async_queue_new();
+						client->session_timeout = 0;
+						/* Take note of this new client */
+						janus_mutex_lock(&clients_mutex);
+						g_hash_table_insert(clients_by_fd, GINT_TO_POINTER(cfd), client);
+						g_hash_table_insert(clients, client, client);
+						janus_mutex_unlock(&clients_mutex);
+					}
+				} else if(poll_fds[i].fd == admin_pfd) {
```
saghul: You can avoid some duplication here by executing the same code if the fd is admin_fd or pfd, and derive the admin boolen from that.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+						client->session_timeout = 0;
+						/* Take note of this new client */
+						janus_mutex_lock(&clients_mutex);
+						g_hash_table_insert(clients_by_fd, GINT_TO_POINTER(cfd), client);
+						g_hash_table_insert(clients, client, client);
+						janus_mutex_unlock(&clients_mutex);
+					}
+				} else {
+					/* Client data: receive */
+					res = read(poll_fds[i].fd, buffer, 4096);
```
saghul: whatever buffer size you choose, better use sizeof, simplifies the changes if you increase it later.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+						/* Close socket */
+						shutdown(SHUT_RDWR, poll_fds[i].fd);
+						close(poll_fds[i].fd);
+						continue;
+					}
+					if(res == 0) {
+						JANUS_LOG(LOG_VERB, "Unix Sockets client disconnected (%d)\n", poll_fds[i].fd);
+						/* Close socket */
+						shutdown(SHUT_RDWR, poll_fds[i].fd);
+						close(poll_fds[i].fd);
```
saghul: set the fd to -1 on the client

transports/janus_pfunix.c
saghul: Am I missing something or are you not putting these sockets in non-blocking mode? (same goes for accept).

saghul: Note that you can use the following Linux-ism to do it in one go:

~~~~
pfd = socket(PF_UNIX, SOCK_DGRAM | SOCK_NONBLOCK, 0)
~~~~

lminiero: Ok, so you DID mean SOCK_DGRAM :smile: 
Yep, forgot to set non-blocking, since again I just merged in a plugin the client/server tests I had made. Will fix this too!

saghul:
I did a first pass review by eyeball-compiling it, so there might be other issues I haven't discovered yet.

Thanks for putting this together so quickly!

transports/janus_pfunix.c
saghul: Yep, I did :-)

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+void janus_pfunix_session_over(void *transport, guint64 session_id, gboolean timeout) {
+	/* We only care if it's a timeout: if so, close the connection */
+	if(transport == NULL || !timeout)
+		return;
+	/* FIXME Should we really close the connection in case of a timeout? */
+	janus_pfunix_client *client = (janus_pfunix_client *)transport;
+	janus_mutex_lock(&clients_mutex);
+	if(g_hash_table_lookup(clients, client) != NULL && client->fd > 0) {
+		/* Shutdown the client socket */
+		shutdown(client->fd, SHUT_WR);
```
saghul: No need if we switch to SOCK_DGRAM

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+						continue;
+					}
+					/* Find the client from its file descriptor */
+					janus_mutex_lock(&clients_mutex);
+					janus_pfunix_client *client = g_hash_table_lookup(clients_by_fd, GINT_TO_POINTER(poll_fds[i].fd));
+					if(client == NULL) {
+						janus_mutex_unlock(&clients_mutex);
+						JANUS_LOG(LOG_WARN, "Got data from unknown Unix Sockets client %d, closing connection...\n", poll_fds[i].fd);
+						/* Close socket */
+						shutdown(SHUT_RDWR, poll_fds[i].fd);
```
saghul: No need if we switch to SOCK_DGRAM.

transports/janus_pfunix.c
lminiero: No, the whole point is to first check if the client is still valid, and only then try to access it. If the client is not in the hashtable, it most likely means the core was still referring to a pointer that has been freed in the meanwhile (e.g., event after a disconnection), and so a client->fd would crash Janus.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,532 @@
...
+						close(poll_fds[i].fd);
+						continue;
+					}
+					if(res == 0) {
+						JANUS_LOG(LOG_VERB, "Unix Sockets client disconnected (%d)\n", poll_fds[i].fd);
+						/* Close socket */
+						shutdown(SHUT_RDWR, poll_fds[i].fd);
```
saghul: No need if we switch to SOCK_DGRAM.

transports/janus_pfunix.c
lminiero: See above (validity check).

transports/janus_pfunix.c
lminiero: Ack.

transports/janus_pfunix.c
saghul: Oh, that sounds dangerous! Is there no way to avoid dangling references in that case?

transports/janus_pfunix.c
lminiero: Ops, you're right: this is what I meant, but I'm actually replacing it. Will fix that.

transports/janus_pfunix.c
lminiero: Ack.

transports/janus_pfunix.c
lminiero: I guess I was still reasoning in threaded terms, while I'm actually doing a single threaded management of all sockets and as such that should never happen. I'll remove the extra check.

transports/janus_pfunix.c
lminiero: :+1:

transports/janus_pfunix.c
lminiero: I replaced the 4096 with a define to 8192, so that should fix it.

transports/janus_pfunix.c
lminiero: Ack.

transports/janus_pfunix.c
lminiero: Not with the way opaque transports are handled as of now. This is something I triedto change/fix in the new reference counters branch, where the transport reference is wrapped pretty much as we do with plugins (a core struct that contains both ends of the rope), so that both the core and plugin can ref/unref it accordingly and it never gets freed while it's used. Since we don't have that now, and I can't just envisage a delayed free (due to some misshap the core might try and send a message several seconds after the transport has gone) this is the closest thing I could come up with for a sanity check.

transports/janus_pfunix.c
lminiero: Mh, but with SOCK_DGRAM will we still get a disconnection event (recv=0) when the client goes away? I started thinking about this when you mentioned we wouldn't need the shutdown in that case, and knowing when a client has left without needing timeouts and stuff would be incredibly helpful.

transports/janus_pfunix.c
saghul: Ack.

transports/janus_pfunix.c
saghul: Hum, you're right. On UDP there is no accept to be done either, you just get datagrams let me double check how that works on Unix sockets. Otherwise we'd have to use SOCK_SEQPACKET.

lminiero: Ok, I'll keep SOCK_SEQPACKET for the moment, as besides apparently you cannot use ```listen()``` with SOCK_DGRAM, and it's not clear to me how you can have the Janus end be the only "server" for all "clients" connecting to it. We can change this later on, when I've a clearer picture on all this!

saghul: Yeah, no dice. SOCK_SEQPACKET it is then!

nowylie: I've been playing around with the idea of a transport using SOCK_DGRAM over AF_UNIX and nearly have some code to show. Some thoughts:
- I don't think listen() is needed. You bind() to an address, and then get the address of the sender when receiving with recvfrom()/recvmsg(). You should be able to use this address to send replies with sendto()/sendmsg()
- SOCK_SEQPACKET over AF_UNIX isn't supported on BSD family (I've been coding on OS X) I don't think.
- sendto()/recvfrom() should be atomic operations as far as the calling code is concerned, i.e. they succeed or fail in one hit. This should reduce the amount of thread synchronization you need for sending or receiving.

lminiero:
This patch should address all the points you raised in your detailed review, let me know if that's not the case! I also noticed I was not notifying the core by means of a ```transport_gone``` when the clients disconnects or goes away, something I added now.

transports/janus_pfunix.c
lminiero: @nowylie thanks for the feedback. Yep, looking around I noticed the need for just a bind, just as with UDP. The unavailability of SOCK_SEQPACKET on other systems we're now, apart from that, compliant with (or at least should be) is of course an additional point in favour of SOCK_DGRAM.

What's puzzling me, though, besides the already made points on disconnect management etc., is how you get multiple clients to work in that case: if 3 people bind to a path, as they'd do here, and two send a message, who gets it? We do need multiple clients here, as otherwise we have to do different paths for each clients, which brings the need for some kind of "reserve a path" message and the like, which I'd rather avoid.

lminiero: As a side note, SOCK_SEQPACKET seems to be available on BSD, or at least on FreeBSD, since it appears to be listed in the docs: https://www.freebsd.org/cgi/man.cgi?query=socket&sektion=2
Not sure if this is a new thing they've added, or it's actually not available at all.

nowylie: The clients don't bind to the same path as the gateway, they bind to their own paths. The client path is then the "address" I was talking about that you use to send replies. The gateway receives messages from different clients on the same path.

saghul: You need to use recvfrom in that case, and you'll get the address of the sender, so you can reply to them.

lminiero: Ah ok, got it, thanks for the clarification.

saghul: @nowylie OSX does have SOCK_SEQPACKET: https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man2/socket.2.html

saghul: Also, I just remembered: reading a datagram of size 0 is legit, so there is no way to signal disconnection. Now, when using datagrams we wouldn't need a client structure per se anyway.

That said, I'd continue with the current solution, that is, using SOCK_SEQPACKET. Sorry for the noise!

nowylie: I don't think OSX does, other BSD's might though. From the `unix(4)` on the FreeBSD man page:

> The	UNIX-domain family supports the	SOCK_STREAM, SOCK_SEQPACKET, and SOCK_DGRAM socket types and uses file system path-names for addressing.

From the `unix(4)` on the OS X man page:

> The UNIX-domain family supports the SOCK_STREAM and SOCK_DGRAM socket types and uses filesystem pathnames for addressing.

saghul: Ah, you're right, it doesn't seem to be supported for Unix sockets. At any rate, when / if Janus works on OSX, we could leave out this transport / find another way.

lminiero: Yeah, I'll have to add a check in configure.ac for the time being.

lminiero: Just added a check in configure.ac that tries to compile a tiny application which uses SOCK_SEQPACKET. This should make sure the plugin is not built were it can't.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,526 @@
...
+static int pfd = -1, admin_pfd = -1;
+/* Socket pair to notify about the need for outgoing data */
+static int write_fd[2];
+
+/* Unix Sockets client session */
+typedef struct janus_pfunix_client {
+	int fd;					/* Client socket */
+	gboolean admin;			/* Whether this client is for the Admin or Janus API */
+	GAsyncQueue *messages;	/* Queue of outgoing messages to push */
+	gint session_timeout:1;	/* Whether a Janus session timeout occurred in the core */
```
saghul: since this is a boolean a gboolean sounds more appropriate

transports/janus_pfunix.c
```diff
@@ -0,0 +1,526 @@
...
+	}
+
+	/* Create a couple of hashtables for all clients */
+	clients = g_hash_table_new(NULL, NULL);
+	clients_by_fd = g_hash_table_new(NULL, NULL);
+	janus_mutex_init(&clients_mutex);
+
+	GError *error = NULL;
+	/* Start the Unix Sockets service thread */
+	if(pfd > -1 || admin_pfd > -1) {
```
saghul: based on the test a few lines above, this is always true

transports/janus_pfunix.c
```diff
@@ -0,0 +1,526 @@
...
+void janus_pfunix_destroy(void) {
+	if(!g_atomic_int_get(&initialized))
+		return;
+	g_atomic_int_set(&stopping, 1);
+
+	/* Stop the service thread */
+	char c;
+	int res = 0;
+	do {
+		res = write(write_fd[1], &c, 1);
```
saghul: I think you can write "x" and avoid using that char c there.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,526 @@
...
+				janus_mutex_unlock(&clients_mutex);
+			}
+			if((poll_fds[i].revents & POLLIN) == POLLIN) {
+				if(poll_fds[i].fd == write_fd[0]) {
+					/* Read and ignore: we use this to unlock the poll if there's data to write */
+					res = read(poll_fds[i].fd, buffer, BUFFER_SIZE);
+				} else if(poll_fds[i].fd == pfd || poll_fds[i].fd == admin_pfd) {
+					/* Janus/Admin API: accept the new client */
+					struct sockaddr_un address;
+					socklen_t address_length = 0;
```
saghul: Doesn't this need to be sizeof(struct sockaddr_un) ?

transports/janus_pfunix.c
```diff
@@ -0,0 +1,526 @@
...
+						client->session_timeout = 0;
+						/* Take note of this new client */
+						janus_mutex_lock(&clients_mutex);
+						g_hash_table_insert(clients_by_fd, GINT_TO_POINTER(cfd), client);
+						g_hash_table_insert(clients, client, client);
+						janus_mutex_unlock(&clients_mutex);
+					}
+				} else {
+					/* Client data: receive */
+					res = read(poll_fds[i].fd, buffer, BUFFER_SIZE);
```
saghul: catch EWOULDBLOCK and ignore it

saghul: silently, that is

transports/janus_pfunix.c
lminiero: I think I originally meant this to become an atomic int, but boolean sounds more apt as it's all basically single threaded anyway.

transports/janus_pfunix.c
lminiero: Ack.

transports/janus_pfunix.c
lminiero: Fair point, I recalled this was actually only filled by the system call, but it indeed needs to be initialized to the size of the struct itself first. Will fix.

transports/janus_pfunix.c
lminiero: You mean as we do with EINTR for writes? Ack.

saghul: No, without the loop. It's possible the fd is not really readable, so you get EAGAIN or EWOULDBLOCK, you should just ignore it.

lminiero: Yep thought so, I've added the check in res<0 in the latest commit.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,525 @@
...
+	char filename[255];
+	g_snprintf(filename, 255, "%s/%s.cfg", config_path, JANUS_PFUNIX_PACKAGE);
+	JANUS_LOG(LOG_VERB, "Configuration file: %s\n", filename);
+	janus_config *config = janus_config_parse(filename);
+	if(config != NULL) {
+		/* Handle configuration */
+		janus_config_print(config);
+
+		/* First of all Initialize the socketpair for writeable notifications */
+		if(socketpair(PF_LOCAL, SOCK_STREAM, 0, write_fd) < 0) {
```
saghul: these also need to be non-blocking, though we might be able to get away with it.  Or maybe we can use a `pipe` and set the read end in non-blocking mode. I'll check what's best.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,525 @@
...
+		}
+
+		/* Setup the Janus API Unix Sockets server(s) */
+		janus_config_item *item = janus_config_get_item_drilldown(config, "general", "enabled");
+		if(!item || !item->value || !janus_is_true(item->value)) {
+			JANUS_LOG(LOG_WARN, "Unix Sockets server disabled (Janus API)\n");
+		} else {
+			item = janus_config_get_item_drilldown(config, "general", "path");
+			char *pfname = (char *)(item && item->value ? item->value : "/tmp/ux-janusapi");
+			if(strlen(pfname) > 108) {
```
saghul: Maybe add a define? I see we use it in a few places.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,525 @@
...
+			char *pfname = (char *)(item && item->value ? item->value : "/tmp/ux-janusadmin");
+			if(strlen(pfname) > 108) {
+				JANUS_LOG(LOG_WARN, "The provided path name (%s) is longer than 108 characters, it will be truncated\n", pfname);
+				pfname[108] = '\0';
+			}
+			/* Create socket */
+			admin_pfd = socket(PF_UNIX, SOCK_SEQPACKET | SOCK_NONBLOCK, 0);
+			if(admin_pfd < 0) {
+				JANUS_LOG(LOG_FATAL, "Unix Sockets %s creation for Admin API failed: %d, %s\n", pfname, errno, strerror(errno));
+				admin_pfd = -1;
```
saghul: no need, in case of error -1 is the only possible return value

transports/janus_pfunix.c
lminiero: It should be defined in un.h (UNIX_MAX_PATH) but it isn't on my Fedora (it's in /linux/un but not in /sys/un.h) so I've used the value it's set to. I'll try to define it myself it if it's undefined.

transports/janus_pfunix.c
```diff
@@ -0,0 +1,507 @@
...
+}
+
+
+/* Thread */
+void *janus_pfunix_thread(void *data) {
+	JANUS_LOG(LOG_INFO, "Unix Sockets thread started\n");
+
+	int fds = 0;
+	struct pollfd poll_fds[1024];	/* FIXME Should we allow for more clients? */
+	char buffer[BUFFER_SIZE];
```
jayridge: Might buffer size be a candidate for config? The difficulty with SOCK_SEQPACKET here is that there is no fixed max request size.

transports/janus_pfunix.c
jayridge: Maybe `recvmsg()` so you can check flags for truncations? An error/warning here would be nice.

transports/janus_pfunix.c
lminiero: I guess it might, and considering the module wouldn't know about what core/plugins and applications might want to send it's probably a good idea.

transports/janus_pfunix.c
lminiero: I wasn't aware of this possibility, thanks! Looking at the docs, you'd get a MSG_TRUNC back, which means we might then try again, later or right away. The only issue with this is that, being single threaded, I'll have to handle the fragmentation of the message in a subsequent poll myself, which means keeping info on partial messages in the client structure (pretty much as we do in the websockets transport). I was hoping I could get away with this "for free" with Unix Sockets but better safe than sorry :-) What I haven't clear is, can I get the rest through another ```recvmsg()```, or do I need something else? Would I get the whole message by passing a larger buffer, or only what was left to read?

saghul: AFAIK you can't do anything if the message got truncated. The kernel won't give you the remaining data.

transports/janus_pfunix.c
saghul: IMHO the user should never worry about this. The same way the core buffer is not adjustable, it should just be big enough.

lminiero: Ok, let's keep it like that for now then. The main purpose of this PR is getting something usable for the time being, it doesn't have to be perfect: there will be room for improvement later on once it starts getting actually used in real deployments and requirements/constraints start appearing.

transports/janus_pfunix.c
lminiero: So it would simply trigger a warning, as @jayridge suggested: fair enough, as it could act as an indication on whether the default buffer size we have is large enough, I'll modify this.

By the way, should we use ```sendmsg``` instead ```write``` for sending stuff back too, then?

transports/janus_pfunix.c
```diff
@@ -0,0 +1,521 @@
...
+						/* Take note of this new client */
+						janus_mutex_lock(&clients_mutex);
+						g_hash_table_insert(clients_by_fd, GINT_TO_POINTER(cfd), client);
+						g_hash_table_insert(clients, client, client);
+						janus_mutex_unlock(&clients_mutex);
+					}
+				} else {
+					/* Client data: receive message */
+					iov[0].iov_len = sizeof(buffer);
+					res = recvmsg(poll_fds[i].fd, &msg, MSG_WAITALL);
```
saghul: :+1:

transports/janus_pfunix.c
```diff
@@ -0,0 +1,521 @@
...
+					if(res < 0) {
+						if(errno != EWOULDBLOCK) {
```
saghul: Also check for EAGAIN, in case is different than EWOULDBLOCK (they are the same on Linux, but one can never be too sure!)

saghul:
Look good! I'll run it for a spin next week and report back.

lminiero:
Cool, thanks!

nowylie:
Here's a link to the SOCK_DGRAM transport that I've been working on: https://github.com/nowylie/janus-ud-transport

lminiero:
@saghul @nowylie the latest commit adds support for SOCK_DGRAM as well: to enable it, just uncomment and set the "type" in the configuration file to SOCK_DGRAM. Tested briefly with a tiny test app I made and it seems to be doing what it needs, but of course feedback is more than welcome!

As a side note, the timeout mechanism seems to work but, thinking about it, will only do its job for the Janus API: in fact, for the Admin API no session is ever created and so no timer is kept there. Besides, the code currently assumes that a client with a specific path can only be either a Janus API or an Admin API client, it cannot be both. An application interested in using both should use two different sockets for the purpose.

lminiero:
Did you guys have the chance to play with this yet?

saghul:
On 02/24/2016 04:03 PM, Lorenzo Miniero wrote:
> Did you guys have the chance to play with this yet?
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/meetecho/janus-gateway/pull/458#issuecomment-188293162>.
> 

Not yet, sorry!

-- 
Saúl Ibarra Corretgé
bettercallsaghul.com

Globik:
And for windows sockets when?

saghul:
@Globik Janus doesn't work on Windows.

lminiero:
Any objection to merge this the way it is? It might get this new transport more visibility, with people wanting to give it a try: we can fine tune it along the way.

saghul:
Ok, let's go for it. I haven't been able to thoroughly test it, but we can do that along the way indeed.

lminiero:
Cool, merging then!

Globik:
@saghul  janus gateway works on Windows7 32-bit too) Successfully: )

saghul:
Without changes?!

Issue #459, opened 2016-02-06
janus_streaming.c poll revents not all handled


marchaase:
I'm getting a tight loop in a streaming thread causing high CPU.  I have a webcam on a laptop streaming video to the janus server, but I occasionally close the lid and let it go to suspend and unsuspend it starts the stream up again.  Not sure if that has anything to do with this, but I have had the high cpu usage when the stream was active as well. 

In htop I am able to identify the process with a name of one of my video streams.  When I strace the process it's tight looping on a poll call.
`poll([{fd=0, events=0}, {fd=8, events=POLLIN}], 2, 1000) = 1 ([{fd=0, revents=POLLERR|POLLHUP}])`

It's coming back with a POLLERR and POLLHUP response in revents.

Looking at the source in janus_streaming.c it seems there is a file descriptor for the audio and video streams that look at the responses to the poll @ https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_streaming.c#L3421 and https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_streaming.c#L3469.  But they only handle the POLLIN event.

I have not had much time to dig through the source code, but it looks like this condition isn't handled and causes the process to spin on the poll event.  It looks like the POLLHUP means the other side hung up, but I'm not sure if that is the case as I've seen this occur while the stream was active.

In any case it looks like there is a hole here where only POLLIN revents are handled.

lminiero:
I don't see how POLLHUP can be set since it's a UDP socket and so there's no connection to be closed in the first place. It sounds more like it's the local listening socket that "breaks down" for some reason. When I get back to the office on Monday I'll give this a look.

marchaase:
I agree it doesn't make sense for a POLLHUP to occur on a UDP socket.  But maybe it's because the POLLERR as well?  If there was an error on the socket then maybe the POLLHUP indicates that the socket is closed or has an error?  I don't think it's a connection per say breaking down, but maybe the socket itself?  Though I can't imagine how or why that could occur.

I searched the source here on github for other poll calls and the only other call I see is in janus_sip.c which I'm not using.

The error isn't occurring right now and I'm not sure what/when causes the issue to appear, but if I see it I'll try to gather more information on it.  I'll be running with debug level 7 to see if I can see anything in the logs.  And I can try to attach to the process with gdb to see if I can get some information that way, but I don't have the source compiled in debug so I'm not sure if I'm competent enough to see what's going on there.

[...]

PR #460, opened 2016-02-08
Better management of poll in streaming and SIP plugins

lminiero:
Tries to address the missing management of POLLERR/POLLHUP in the streaming plugin, as highlighted in #459. Also improves a bit the poll related code (especially in case a single medium is shared) and notifies users when a live RTP mountpoint is closed (whatever the reason).

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2213,22 +2213,22 @@ static int janus_sip_allocate_local_ports(janus_sip_session *session) {
 		return -1;
 	}
 	/* Reset status */
-	if(session->media.audio_rtp_fd >= 0) {
+	if(session->media.audio_rtp_fd != -1) {
 		close(session->media.audio_rtp_fd);
 		session->media.audio_rtp_fd = -1;
 	}
-	if(session->media.audio_rtcp_fd >= 0) {
+	if(session->media.audio_rtcp_fd != -1) {
 		close(session->media.audio_rtcp_fd);
 		session->media.audio_rtcp_fd = -1;
 	}
 	session->media.local_audio_rtp_port = 0;
 	session->media.local_audio_rtcp_port = 0;
 	session->media.audio_ssrc = 0;
-	if(session->media.video_rtp_fd >= 0) {
+	if(session->media.video_rtp_fd != -1) {
 		close(session->media.video_rtp_fd);
 		session->media.video_rtp_fd = -1;
 	}
-	if(session->media.video_rtcp_fd >= 0) {
+	if(session->media.video_rtcp_fd != -1) {
 		close(session->media.video_rtcp_fd);
 		session->media.video_rtcp_fd = -1;
 	}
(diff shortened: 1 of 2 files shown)


plugins/janus_streaming.c
```diff
@@ -3386,135 +3386,183 @@ static void *janus_streaming_relay_thread(void *data) {
...
+	int num = 0;
 	janus_streaming_rtp_relay_packet packet;
 	while(!g_atomic_int_get(&stopping) && !mountpoint->destroyed) {
+		/* Prepare poll */
+		num = 0;
+		if(audio_fd >= 0) {
+			fds[num].fd = 0;
+			fds[num].events = 0;
+			fds[num].revents = 0;
+			if(audio_fd > 0) {
```
saghul: this test is covered by the one above

lminiero: Damn copy/paste... ack :wink:

lminiero:
Changed the title as I've made the same change for the SIP plugin as well, which is also using poll for incoming data.

plugins/janus_sip.c
```diff
@@ -2213,24 +2213,24 @@ static int janus_sip_allocate_local_ports(janus_sip_session *session) {
 		return -1;
 	}
 	/* Reset status */
-	if(session->media.audio_rtp_fd > 0) {
+	if(session->media.audio_rtp_fd >= 0) {
```
saghul: != -1 maybe?

lminiero: Isn't it the same check? Or do you mean for code optimization?

saghul: yep, it's the same, IMHO it's more explicit.

lminiero: :+1: did that for the streaming plugin too, as I was at it.

plugins/janus_sip.c
```diff
@@ -2243,10 +2243,10 @@ static int janus_sip_allocate_local_ports(janus_sip_session *session) {
 		while(session->media.local_audio_rtp_port == 0 || session->media.local_audio_rtcp_port == 0) {
 			if(attempts == 0)	/* Too many failures */
 				return -1;
-			if(session->media.audio_rtp_fd == 0) {
+			if(session->media.audio_rtp_fd < 0) {
```
saghul: == -1

plugins/janus_sip.c
```diff
@@ -2432,93 +2432,97 @@ static void *janus_sip_relay_thread(void *data) {
...
-			bytes = recvfrom(session->media.video_rtcp_fd, buffer, 1500, 0, (struct sockaddr*)&remote, &addrlen);
-			//~ JANUS_LOG(LOG_VERB, "************************\nGot %d bytes on the video RTCP channel...\n", bytes);
-			//~ rtp_header_t *rtp = (rtp_header_t *)buffer;
-			//~ JANUS_LOG(LOG_VERB, " ... parsed RTP packet (ssrc=%u, pt=%u, seq=%u, ts=%u)...\n",
-				//~ ntohl(rtp->ssrc), rtp->type, ntohs(rtp->seq_number), ntohl(rtp->timestamp));
-			/* Relay to browser */
-			gateway->relay_rtcp(session->handle, 1, buffer, bytes);
-			continue;
+		for(int i=0; i<num; i++) {
+			if((fds[i].revents & POLLERR) == POLLERR || (fds[i].revents & POLLHUP) == POLLHUP) {
```
saghul: how about:

~~~~
if(fds[i].revents & (POLLERR | POLLHUP)) {
~~~~

plugins/janus_sip.c
```diff
@@ -2432,93 +2432,97 @@ static void *janus_sip_relay_thread(void *data) {
...
+				/* Socket error? */
+				JANUS_LOG(LOG_ERR, "[SIP-%s] Error polling: %s...\n", session->account.username,
+					(fds[i].revents & POLLERR) == POLLERR ? "POLLERR" : "POLLHUP");
+				JANUS_LOG(LOG_ERR, "[SIP-%s]   -- %d (%s)\n", session->account.username, errno, strerror(errno));
+				break;
+			} else if((fds[i].revents & POLLIN) == POLLIN) {
```
saghul: no need to check if it's POLLIN

saghul:
:+1: with a couple of suggestions.

lminiero:
Thanks, just pushed the changes you suggested!

saghul:
LGTM :+1:

lminiero:
I guess this is safe to merge, as it does indeed fix the poll issues. How to properly cope with such an error when it happens can be done in a separate issue/PR, if needed.

saghul:
Nice! :+1:

PR #463, opened 2016-02-10
Sequential loading of required JS scripts

Tajilso:
janus.js now loads required JS scripts in a sequential and synchronous manner before invoking the 'initialization done' callback.


lminiero:
Looks good to me!

PR #464, opened 2016-02-11
Removed jQuery dependency from janus.js

lminiero:
This PR attempts to remove all dependencies to jQuery from the ```janus.is``` library, as suggested by @leonuh in #446.

Apart from a couple of tiny things, we were mostly using jQuery only for AJAX requests, so the PR introduces a new static method called ```Janus.ajax``` that takes care of that. Since I'm inherently lazy and I didn't want to change tons of things in the code, I conceived the method to have the same signature as ```$.ajax``` (for the things I cared about, of course: most arguments are ignored). This way, I only had to rename all occurrences of ```$.ajax``` to ```Janus.ajax```. I tested it briefly and it seems to be working fine, but as anticipated I'm no true JS expert, so any review/feedback on getting this fixed and/or optimized before merging will be very helpful.

PS: jQuery is still being used in the demos themselves, and since (again) I'm lazy, neither @leonuh nor @Tajilso will get me to change THOSE too :stuck_out_tongue_winking_eye:

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -109,7 +109,7 @@ Janus.init = function(options) {
 				};
 			}
 			try {
-				XHR.send(XHR.data);
+				XHR.send(params.data);
 				if(!params.async) {
 					if(XHR.status !== 200) {
 						// Got an error?
--- a/html/janus.js
+++ b/html/janus.js
@@ -13,7 +13,7 @@ Janus.isExtensionEnabled = function() {
 			// Older versions of Chrome don't support this extension-based approach, so lie
 			return true;
 		}
-		return ($('#janus-extension-installed').length > 0);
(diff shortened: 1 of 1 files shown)


lminiero:
The "overwhelming feedback makes me think not many people care about this... I'll add the modified file as ```janus.nojquery.js``` to master, keeping the ```janus.js``` we have as the main version, and add info to the documentation to people intesested in it. Closing this PR.

kewde:
@lminiero the version without jQuery seems to have been removed.
Looking through janus.js, it seems that the no jQuery is the default now?

lminiero:
@kewde yep, they've been merged and a new version of janus.js has been refactored by a contributor to allow for custom dependencies. As you pointed out, the default is not to use jQuery anymore.

PR #466, opened 2016-02-11
Autodetect libwebsockets version and use the right API

lminiero:
This PR tries to merge the proposed libwebsockets 1.6 API change from #438 into a more "automagical" patch. It tries to autodetect the libwebsockets version when configuring (pre-1.6 or post-1.6) and, using a variable defined there, selectively use the right API signatures accordingly.

I verified this seems to be backward compatible with the pre-1.6 version of libwebsockets I have installed. I haven't tested with a more recent version instead (too many branches still depending on the version I have to mess with the environment), so not sure if this will even compile, but I don't see why it shouldn't. If anybody can test that one and make sure everything still works as expected, I'll go on and merge.

Mentioning @FrozenCow as he authored the original patch and so will probably be interested in this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,6 @@ HTTP REST API, you'll have to install it manually:
 
 	git clone git://git.libwebsockets.org/libwebsockets
 	cd libwebsockets
-	git checkout v1.5-chrome47-firefox41
 	mkdir build
 	cd build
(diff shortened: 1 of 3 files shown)


lminiero:
This has been around long enough, it seems to be working for us, and I've read no complaints.
Merging.

PR #467, opened 2016-02-12
First take at RTCP SR/RR in core

lminiero:
This PR continues the effort started by @fpn in #382. Specifically, it moves the original code from plugin to core, so that the functionality is actually automatically shared across different plugins for each PeerConnection. The collected info is then also made available in the Admin API.

It is not at all complete, as I still have to figure out the best way for getting this to work seamlessly with what plugins originate and do. To make a simple example, the VideoRoom plugin still sends empty RR just to attach a compound REMB, all while the core tries to send proper RR in the meantime every 5 seconds.

That said, it should be a good starting point to get better RTCP management here, so feedback welcome.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -139,6 +139,7 @@ void janus_ice_force_rtcpmux(gboolean forced) {
 		janus_force_rtcpmux_blackhole_port = ntohs(serveraddr.sin_port);
 		JANUS_LOG(LOG_VERB, "  -- RTCP component blackhole socket bound to port %d\n", janus_force_rtcpmux_blackhole_port);
 		janus_force_rtcpmux_blackhole_fd = blackhole;
+
 	}
 }
 gint janus_ice_get_rtcpmux_blackhole_port(void) {
@@ -1168,6 +1169,8 @@ void janus_ice_stream_free(GHashTable *streams, janus_ice_stream *stream) {
 	if(stream->video_rtcp_ctx != NULL)
 		g_free(stream->video_rtcp_ctx);
 	stream->video_rtcp_ctx = NULL;
+	stream->audio_last_ts = 0;
+	stream->video_last_ts = 0;
 	g_free(stream);
 	stream = NULL;
 }
@@ -1993,6 +1996,7 @@ void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint component_i
 					component->retransmit_recent_cnt = 0;
 					component->retransmit_log_ts = now;
 				}
+
 				janus_plugin *plugin = (janus_plugin *)handle->app;
 				if(plugin && plugin->incoming_rtcp)
 					plugin->incoming_rtcp(handle->app_handle, video, buf, buflen);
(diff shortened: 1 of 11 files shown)


mtdxc:
any news about this PR?

lminiero:
Actually I was waiting for feedback here :wink:

rtcp.c
```diff
@@ -327,6 +348,113 @@ int janus_rtcp_fix_ssrc(char *packet, int len, int fixssrc, uint32_t newssrcl, u
...
+}
+
+uint32_t janus_rtcp_context_get_lost(rtcp_context *ctx) {
+	if(ctx == NULL)
+		return 0;
+	uint32_t lost;
+	if(ctx->lost > 0x7fffff) {
+		lost = 0x7fffff;
+	} else {
+		lost = 0x00FFFFFF & ctx->lost;
```
ploxiln: you mix upper-case and lower-case hex literals here. also, the mask in the else case is redundant - there can't be any high bits set to mask off if `! (ctx->lost > 0x7FFFFF)`

ice.c
```diff
@@ -3049,14 +3129,14 @@ void *janus_ice_send_thread(void *data) {
...
-					janus_rtcp_fix_ssrc((char *)&sbuf, pkt->length, 1,
+					janus_rtcp_fix_ssrc(NULL, (char *)&sbuf, pkt->length, 1,
 						video ? stream->video_ssrc : stream->audio_ssrc,
 						video ? stream->video_ssrc_peer : stream->audio_ssrc_peer);
 				} else {
 					/* Plan B involved, we trust the plugin to set the right 'local' SSRC and we don't mess with it */
 					JANUS_LOG(LOG_HUGE, "[%"SCNu64"] Fixing peer SSRC (Plan B, peer %u)\n", handle->handle_id,
 						video ? stream->video_ssrc_peer : stream->audio_ssrc_peer);
-					janus_rtcp_fix_ssrc((char *)&sbuf, pkt->length, 1, 0,
+					janus_rtcp_fix_ssrc(NULL, (char *)&sbuf, pkt->length, 1, 0,
```
ploxiln: you don't need to cast any of these buffers to `(char *)`. just `sbuf` should do it.

(this is actually one of those interesting cases where the address value of `sbuf` is the same as the address value of `&sbuf`. the types are a bit different. but just `sbuf` will work for a `char *`)

ice.c
```diff
@@ -1162,6 +1161,12 @@ void janus_ice_stream_free(GHashTable *streams, janus_ice_stream *stream) {
 		g_free(stream->rpass);
 		stream->rpass = NULL;
 	}
+	if(stream->audio_rtcp_ctx != NULL)
+		g_free(stream->audio_rtcp_ctx);
```
ploxiln: more unnecessary NULL checks before g_free() which also does null checks :)

ice.c
lminiero: Ack!

ice.c
lminiero: Yeah, I'll never learn :wink:

rtcp.c
lminiero: This was part of the original PR from @fpn so I haven't messed with that part yet.

lminiero:
This patch should basically complete the RTCP related changes: now we terminate SR/RR and so generate our own. As such, we also filter whatever RTCP plugins try to send, in order to make sure the only outgoing ST/RR/SDES is what we send. The patch also takes care of prepending REMB feedback with RR (something plugins had to do on their own, before).

Mentioning @fpn @lfpnl as author of the original contribution and @mtdxc @arjenveenhuizen @ploxiln as you showed interest in this effort. Please test and let me know if this is working as expected for you, and whether you believe you're getting consistent information out of this.

mtdxc:
great !!!

lminiero:
Any feedback?

lminiero:
Merging as this has been around long enough. I guess that's the best way for people to finally and really test it :wink:

PR #468, opened 2016-02-15
Buffer the latest received keyframe in streaming plugin for new viewers

lminiero:
This is an experimental patch that addresses some feedback I received from a few users in the past. Specifically, it adds an option in the Streaming plugin to allow RTP mountpoints to keep a copy of the latest received keyframe: this way, when new users join and they start receiving media, the latest keyframe is sent first, and then the live video frames, so that new viewers do not have to wait for the next keyframe in case they joined in between (and so black screen for a few seconds).

To enable it, just add a ```videobufferkf=yes``` to the mountpoint configuration (it's disabled by default); to do the same programmatically, i.e., via Streaming plugin API, just add a boolean ```"videobufferkf"``` attribute to the ```"create"``` request for the new mountpoint and set it to ```true```.

It is admittedly not perfect and yet experimental (viewers do get a meaningful image right away, but it's frozen), but is a nice addon and so I'm publishing it anyway. Feel free to provide feedback and suggestions on how to improve, I plan to merge this sooner than later.

### DIFF
--- a/conf/janus.plugin.streaming.cfg.sample.in
+++ b/conf/janus.plugin.streaming.cfg.sample.in
@@ -27,6 +27,8 @@
 ; videomcast = multicast group port for receiving video frames, if any
 ; videocodec = <video RTP payload type> (e.g., 100)
 ; videortpmap = RTP map of the video codec (e.g., VP8/90000)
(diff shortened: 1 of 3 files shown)


lminiero:
Merging.

PR #469, opened 2016-02-16
Fix for occasional race conditions when shutting down the SIP stack

lminiero:
This PR fixes a problem that @bhakimi reported. Specifically, Janus would at times crash, apparently for no reason, while using the SIP plugin. Looking at traces, we found out that, when this happened, the Sofia stack had just received a shutdown event (```nua_r_shutdown```), and the exact cause was a race condition where the plugin session had already been freed, thus crashing the plugin when trying to access its content.

To fix this, I modified the plugin code to do the following:

1. only initialise the SIP stack within the ```janus_sip_sofia_thread``` Sofia loop thread: this way the stack is only freed when the loop ends, that is after the shutdown event;
2. modified the way the lazy garbage collector kicks in: immediately adding the session to the "old" list if there is no stack (i.e., plugin session never handled a successfull register request), or at the end of the above mentioned Sofia loop thread (so that the session is only collected after the stack has been shut down).

I also fixed a couple of leaks here and there, and verified this seems to be working as expected now, but since this is a serious change and I don't want it to break anything people has working correctly as of now, I decided to publish it as a PR for the moment.

If you're using SIP please test this and let me know if it's safe to merge. Mentioning some of the people I know are using this because of recent related issues/PR activity, so forgive the spam: @saghul @hoangtuansu @cwerner-bfs @s-schumann @phillcz @MichaelB76 @shrhoads

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2547,6 +2547,7 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 	session->stack->s_nh_r = NULL;
 	session->stack->s_nh_i = NULL;
 	session->stack->s_root = su_root_create(session->stack);
+	su_home_init(session->stack->s_home);
 	JANUS_LOG(LOG_VERB, "Setting up sofia stack (sip:%s@%s)\n", session->account.username, local_ip);
 	char sip_url[128];
 	char sips_url[128];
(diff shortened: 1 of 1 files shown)


bhakimi:
will test all day tomorrow and let you guys know

bhakimi:
@lminiero ran all day with absolutely no crash or errors, memory looks fine and cpu also looks great, id say merge it in. thank you so much for this fix, this was the cause of our crash we've been having for a long time.

saghul:
FWIW I won't be able to test this soon, so if you're confident just go with it and if/when we find issues we can solve them later :-)

lminiero:
Cool, thank you both for the ack, I'll merge then!

MichaelB76:
Looks OK to me but will test it when we merge in the latest code, will will likely be in another two or three weeks time.

lminiero:
:+1:

bhakimi:
still no crashes, i think you can merge it in

lminiero:
Already merged a couple of days ago, good to know it's improved things!

PR #470, opened 2016-02-18
Added optional SDES-SRTP support to SIP plugin

lminiero:
This PR adds support for SDES-SRTP to calls handled by the SIP plugin. To enable this for outgoing calls, just add a ```srtp:true``` attribute to a ```"call"``` request. It will be enabled automatically, instead, for incoming calls with a crypto line in it.

Tested in different scenarios and it seems to be working fine both for incoming and outgoing calls. The parsing of crypto lines can definitely be improved (we don't handle all the possibilities for the ```inline``` part, for instance, assuming it's always just the base64 encoded key/salt), but there will be time for that.

Feel free to play with this, especially if you have SDES endpoints you want to test this with (I've only tested this with Asterisk 13.6.0). Feedback welcome as I plan to merge this soon.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -431,13 +431,13 @@ static void janus_sip_srtp_cleanup(janus_sip_session *session) {
 	session->media.audio_srtp_out = NULL;
 	g_free(session->media.audio_local_policy.key);
 	session->media.audio_local_policy.key = NULL;
-	session->media.audio_srtp_suite_in = 0;
+	session->media.audio_srtp_suite_out = 0;
 	if(session->media.audio_srtp_in)
 		srtp_dealloc(session->media.audio_srtp_in);
 	session->media.audio_srtp_in = NULL;
 	g_free(session->media.audio_remote_policy.key);
 	session->media.audio_remote_policy.key = NULL;
-	session->media.audio_srtp_suite_out = 0;
+	session->media.audio_srtp_suite_in = 0;
 	/* Video */
 	if(session->media.video_srtp_out)
 		srtp_dealloc(session->media.video_srtp_out);
@@ -450,7 +450,7 @@ static void janus_sip_srtp_cleanup(janus_sip_session *session) {
 	session->media.video_srtp_in = NULL;
 	g_free(session->media.video_remote_policy.key);
 	session->media.video_remote_policy.key = NULL;
-	session->media.video_srtp_suite_out = 0;
+	session->media.video_srtp_suite_in = 0;
 }
 
 
@@ -2638,7 +2638,7 @@ char *janus_sip_sdp_manipulate(janus_sip_session *session, sdp_session_t *sdp) {
 		if(session->media.has_video) {
 			char *crypto = NULL;
 			session->media.video_srtp_suite_out = 80;
-			janus_sip_srtp_set_local(session, FALSE, &crypto);
+			janus_sip_srtp_set_local(session, TRUE, &crypto);
 			/* FIXME 32? 80? Both? */
 			char cryptoline[100];
 			g_snprintf(cryptoline, 100, "a=crypto:1 AES_CM_128_HMAC_SHA1_80 inline:%s", crypto);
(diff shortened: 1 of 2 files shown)


saghul:
This is fantastic, thanks Lorenzo! I'll add it to my ToTest list!

lminiero:
Cool, thanks!

saghul:
I had a quick glance and noticed you're using RTP/SAVP when this is set, which means that SRTP is "forced". Can we make that `srtp` setting take a string indicating the mode? I suggest these:

- sdes_optional: profile is RTP/AVP, with crypto lines
- sdes_mandatory: profile is RTP/SAVP, with crypto lines
- dtls: imaginary case in which we also use SRTP-DTLS on the SIP side (not to be implemented)

WDYT?

lminiero:
Seems reasonable, as it is indeed too strict as it is now. I'll do this in the next commit.

saghul:
:+1: :heart_eyes:

lminiero:
That should do it. The only problem with ```sdes_optional``` is that, when testing this with Asterisk, it wouldn't accept an INVITE with RTP/AVP and a crypto line (```Can't provide secure video requested in SDP offer```), which I guess might be the same with other endpoints out there, but that's really out of scope to the feature itself (translated: if you know it won't work with your infrastructure, stick to "sdes_mandatory" instead for outgoing calls).

saghul:
IIRC there is a configuration option for that... Using RTC/AVP and crypto lines ins not really standard, but it's how everyone has been doing it forever.

I see Asterisk had some problems with this at some point: https://issues.asterisk.org/jira/browse/ASTERISK-23989

lminiero:
I think they did something like this for the PJSIP based SIP support (I remember things like "optimistic SRTP" or something like that in the config at the time), but maybe it's not in chan_sip instead. Anyway, that's not really important, especially as you said a lot of people out there do that anyway, and we provide the means to choose what to do and how to do it.

plugins/janus_sip.c
```diff
@@ -2207,6 +2571,84 @@ void janus_sip_sdp_process(janus_sip_session *session, sdp_session_t *sdp) {
...
+			/* FIXME 32? 80? Both? */
+			char cryptoline[100];
+			g_snprintf(cryptoline, 100, "a=crypto:1 AES_CM_128_HMAC_SHA1_80 inline:%s", crypto);
+			g_free(crypto);
+			new_sdp = janus_string_replace(new_sdp, "a=crypto:audio", cryptoline);
+		}
+		if(session->media.has_video) {
+			char *crypto = NULL;
+			session->media.video_srtp_suite_out = 80;
+			janus_sip_srtp_set_local(session, FALSE, &crypto);
```
saghul: s/FALSE/TRUE/

plugins/janus_sip.c
```diff
@@ -304,12 +314,153 @@ struct ssip_s {
...
+	g_free(session->media.audio_remote_policy.key);
+	session->media.audio_remote_policy.key = NULL;
+	session->media.audio_srtp_suite_out = 0;
+	/* Video */
+	if(session->media.video_srtp_out)
+		srtp_dealloc(session->media.video_srtp_out);
+	session->media.video_srtp_out = NULL;
+	g_free(session->media.video_local_policy.key);
+	session->media.video_local_policy.key = NULL;
+	session->media.video_srtp_suite_out = 0;
```
saghul: session->media.video_srtp_suite_in

saghul:
I ran it for a spin, worked with audio but failed with video. The comments I made fix the video crashes.

Great work!

lminiero:
Great, thanks for testing! Currently abroad for a few days, will fix when I
get back.
Il 25/feb/2016 16:14, "Saúl Ibarra Corretgé" <notifications@github.com> ha
scritto:

lminiero:
Just made the changes: could you check if these were what you suggested, in particular with respect to the second comment? I'm not sure what typo you referred to there.

saghul:
Yet, that was exactly whay I meant. :+1:

lminiero:
Cool, thanks! Merging then.

PR #471, opened 2016-02-22
Added support for time reset in the rtp payload

kris-lab:
Part of https://github.com/cargomedia/janus-gateway-rtpbroadcast/issues/62

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -13,22 +13,22 @@
  * files. This utility allows you to process those files, in order to
  * get a working .webm (if the recording includes VP8 frames) or .opus
  * (if the recording includes Opus frames) file.
- * 
+ *
  * Using the utility is quite simple. Just pass, as arguments to the tool,
  * the path to the .mjr source file you want to post-process, and the
  * path to the destination file (a .webm if it's a video recording,
  * .opus otherwise), e.g.:
- * 
+ *
 \verbatim
-./janus-pp-rec /path/to/source.mjr /path/to/destination.[opus|webm] 
-\endverbatim 
- * 
+./janus-pp-rec /path/to/source.mjr /path/to/destination.[opus|webm]
+\endverbatim
+ *
  * \note This utility does not do any form of transcoding. It just
  * depacketizes the RTP frames in order to get the payload, and saves
  * the frames in a valid container. Any further post-processing (e.g.,
  * muxing audio and video belonging to the same media session in a single
  * .webm file) is up to third-party applications.
- * 
+ *
  * \ingroup postprocessing
  * \ref postprocessing
  */
@@ -75,7 +75,7 @@ int main(int argc, char *argv[])
 			janus_log_level = val;
 		JANUS_LOG(LOG_INFO, "Logging level: %d\n", janus_log_level);
 	}
-	
+
 	/* Evaluate arguments */
 	if(argc != 3) {
 		JANUS_LOG(LOG_INFO, "Usage: %s source.mjr destination.[opus|webm]\n", argv[0]);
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for working on this! Did you notice problems when post-processing media without this fix? IIRC we had code in place to handle these resets, but it's been some time since last time I've worked on the post processor so I may be wrong.

As a side note, have you signed our [CLA](https://cla.conf.meetecho.com/janus/) already?

lminiero:
Nevermind my question, just read your explaination on the referenced issue and your fix makes sense. I'll review ASAP and come with feedback, thanks!

kris-lab:
@lminiero yes, I have experienced small issue with conversion of large RTP dumps. I have noticed that if there is time reset on the RTP payload then conversion cannot be finished. I can see there is implementation to cover that case but in our scenario it looks like looping back for finding previous packet is not possible. That means for each `next` packet in the list the `while` loop always goes to the beginning.

I believe my workaround is not perfect. I would rather try to combine this implementation with  `reset` https://github.com/meetecho/janus-gateway/blob/master/postprocessing/janus-pp-rec.c#L243 which is already there. 

wdyt?

lminiero:
I do remember working on that a long time ago, as it was happening for some of our recordings that were longer than others, and it seemed to do the trick: the internal timestamp we use for ordering is a 64 bit integer that we can do our ordering on then.

Not sure what's causing it to fail in your case. Looking at the code, I thought we were also keeping track of the number of resets there had been (that's what I remembered), but apparently that's not the case, and so we're only handling a pre-reset and post-reset case, meaning that if there is a second reset along the road things may break there. I guess we might have disregarded the case as unlikely (IIRC you need more than 13 hours for a reset, starting at 0). If so, and that's what's happening to you (loop at the second reset), then just keeping track of the number of resets and using that in conjunction with the reset calculator we have may do the trick.

lminiero:
PS: do you have a mjr file to share that I can use to test this, and to check how I can fix it with my idea on the resets counter approach?

kris-lab:
I do have it but it is 36GB! Will check can I easily share it with you. The problem currently appears ~28GB so testing is very painful as you have to read this data every time to build the list of packets.

kris-lab:
And yes, I agree with your tips and approach. Will try to have a look asap.

@njam wdyt about file sharing?

kris-lab:
I am uploading this file into the google drive. Could you share you e-mail so I will send you invite. thx

lminiero:
That's huge! :fearful: If so, this seems to confirm it might indeed be a double reset that's causing the issue: anything that lasts between 13 and 26 hours (depending on what the initial timestamp was) may have a second reset in it, and so causing the issue. In that case, while processing you should actually see a first successfull timestamp reset:
https://github.com/meetecho/janus-gateway/blob/master/postprocessing/janus-pp-rec.c#L244
and then a problematic one later on.

That said, with such a huge file, sharing it is out of the question I guess... I'll try to come up with a patch to handle the resets counter shortly (probably tomorrow) so that you can test it yourself.

lminiero:
> I am uploading this file into the google drive. Could you share you e-mail so I will send you invite. thx

Nevermind, it would take forever to download, I'll work blind and try to prepare a PR ASAP :wink:

kris-lab:
Ok then! I am stopping my upload :smile: and also fyi we do have files up to 120GB with tens of hours of streaming. I guess the reset can occur for multiple time there.

Many thanks for taking care of it. I will test your patch asap.

lminiero:
@kris-lab @njam can you have a look at the new PR I created? It should address the multiple timestamp resets we discussed yesterday.

I don't have access to a recording that actually includes them, so I can't be sure it works as expected: I did a simple test by generating a fake mjr recording that contained a couple of resets, and it seemed to detect them correctly and set the internal 64 bit timestamp value correctly, but, again, not sure if this actually solves the issue. Besides, the code is a bit hacky right now and with a lot of LOG_WARN here and there to highlight what I was interested in. Feedback welcome!

kris-lab:
This issues has been solved in #472 

Closing. Please reopen if needed.

PR #472, opened 2016-02-23
Post-processing recordings with more than a single timestamp reset

lminiero:
This PR follows the discussion started in #471 and tries to address the issue, by detecting multiple timestamp resets and fixing the problem. Highly experimental and so to be tested carefully, also to make sure it doesn't introduce any regression with existing recordings (particularly those containing, for instance, out of order packets and the like).

The patch also introduces a new command line parameter for the post-processor, ```--parse```, which besides reading the header (what ```---header``` did already) also goes through the recording to simply parse the file and re-order the packets. Once done, it does not actually process the recording and stops there. May be useful for testing and debugging without having to go through the whole processing, which might take longer.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -342,7 +342,6 @@ int main(int argc, char *argv[])
 			/* Take into account the number of resets when setting the internal, 64-bit, timestamp */
 			p->ts = (times_resetted*max32)+ntohl(rtp->timestamp);
 		}
-		JANUS_LOG(LOG_INFO, "%"SCNu32" --> %"SCNu64"\n", ntohl(rtp->timestamp), p->ts);
 		last_ts = ntohl(rtp->timestamp);
 		post_reset_pkts++;
 		/* Fill in the rest of the details */
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -79,12 +79,15 @@ int main(int argc, char *argv[])
 	/* Evaluate arguments */
 	if(argc != 3) {
 		JANUS_LOG(LOG_INFO, "Usage: %s source.mjr destination.[opus|webm]\n", argv[0]);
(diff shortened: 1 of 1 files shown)


kris-lab:
I can confirm that conversion works perfectly for dump with single time resets. In the future I will try to verify that it would work also with longer dumps > 30h (> 40GB in my case).

postprocessing/janus-pp-rec.c
```diff
@@ -300,21 +317,35 @@ int main(int argc, char *argv[])
...
+					reset = ntohl(rtp->timestamp);
+					JANUS_LOG(LOG_WARN, "Timestamp reset: %"SCNu32"\n", reset);
+					times_resetted++;
+					post_reset_pkts = 0;
+				}
 			}
+			/* Take into account the number of resets when setting the internal, 64-bit, timestamp */
+			p->ts = (times_resetted*max32)+ntohl(rtp->timestamp);
 		}
+		JANUS_LOG(LOG_INFO, "%"SCNu32" --> %"SCNu64"\n", ntohl(rtp->timestamp), p->ts);
```
kris-lab: I would try to avoid this log. It dumps far to much info and is not really useful in the production. wdyt?

lminiero:
Cool, thanks for the update, keep us posted!

postprocessing/janus-pp-rec.c
lminiero: Yeah, as I meant it was there orly for my tests in debugging, the plan is to get rid of them all when done. I'll start with this one :smile:

lminiero:
@kris-lab any update on the very long recordings? If you can confirm it seems to be working as expected with all the recordings you tested it with so far and there are no regressions, I'll merge.

kris-lab:
I can confirm that it works with recordings up to 18h long. Unfortunately, I don't have any longer at the moment. I haven't experience any regression so far. I would suggest to merge this.

wdyt?

lminiero:
Thanks for the feedback! Merging then, if we notice any regression or problem with longer recordings along the road we can handle them in a new issue/PR.

PR #476, opened 2016-03-03
Device selection in janus.js and new demo

lminiero:
This PR is heavily based on the excellent contribution by @leonuh in #446. More specifically, it integrates the device selection stuff within ```janus.js``` (and ```janus.nojquery.js```) itself, and adds a new demo that is basically a variant of the Echo Test demo that allows you to exploit the new feature.

There are a few differences from the original patch, so I'll summarize the changes here:

1. There's a new static function in janus.js, ```Janus.listDevices```, which does the enumeration for you: you pass to it a callback function, which the method will then pass an array of devices. These devices can be used to build a list somewhere. In order to get this working well with Firefox, I had to put a ```getUserMedia``` before ```enumerateDevices```, as otherwise the label for each device is empty.

2. As in the original patch, if you pass an object with a deviceId to createOffer/createAnswer for audio/video, it is used in place of whatever you specified. I got rid of the ```jQuery.extend``` the original patch had because i) it seemed to break the device selection, and ii) we couldn't use it in ```janus.nojquery.js``` anyway. All of this means that you cannot combine, for instance, a deviceId with a desired resolution. Not sure this can be done at all, but we can worry about this later.

3. The new demo is called "Device Selection" and is in a different group from the other demos, although it's referenced in the Echo Test demo itself. When you start it, it only lists the devices, and a new button allows you to start capturing the selected ones. As in the original patch, changing them while in a session allows you to change devices: rather than binding this to a change in the select, though, I preferred using the button for that instead, so that you can change both audio and video devices at the same time before re-capturing (before you needed two different changes for that), and you don't risk re-capturing after changing the selected device by mistake.

That's basically it. It's still not mergeable as this apparently breaks screen sharing as it is (although it may be because I'm testing on M50). Please test and provide feedback!

### DIFF
--- a/html/adapter.js
+++ b/html/adapter.js
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2014 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2016 The WebRTC project authors. All Rights Reserved.
  *
  *  Use of this source code is governed by a BSD-style license
  *  that can be found in the LICENSE file in the root of the source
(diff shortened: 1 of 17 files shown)


lminiero:
> It's still not mergeable as this apparently breaks screen sharing as it is

Found the issue, I forgot to use promises for the screen sharing getUserMedia as well...

lminiero:
Any feedback? Objections to a merge?

leonuh:
I haven't had time to check the demo yet, although I went through the code once and it seems to be good.  Regarding to the jquery.extend part: you can consider to use Object.assign. http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign

lminiero:
Thanks, I'll look into that.

lminiero:
This has been around long enough and seems to be working as expected, merging.

PR #478, opened 2016-03-09
Configurable video codec to force in VideoRoom plugin

lminiero:
This pull request adds a new setting for rooms in the VideoRoom plugin, called ```videocodec```, both in the configuration file and in ```create``` requests. This setting can be used to force the video codec to be used in the specified room: that is, the only codec the room will accept for publishing video, and as such the only codec subscribers will be able to use for getting video from them. By default this is VP8 (```videocodec=vp8```), but you can choose VP9 (```videocodec=vp9```) and H.264 (```videocodec=h264```) as well.

Hybrid scenarios (rooms where you accept more than one video codec) are not supported yet, we can worry about this later. Besides, due to how the recording and postprocessing stuff works at the moment, recording will NOT work if you choose a codec different than VP8. Again, we can worry about this later.

I verified that this seems to be working as expected with both VP9 (Chrome, Firefox falls back to audio only as it doesn't support VP9 apparently) and H.264 (Firefox, Chrome fails when receiving H.264 in the subscriber offer). If this effort interests you, please test and provide feedback.

### DIFF
--- a/conf/janus.plugin.videoroom.cfg.sample
+++ b/conf/janus.plugin.videoroom.cfg.sample
@@ -7,6 +7,7 @@
 ;              conference or 1 for a webinar)
 ; bitrate = <max video bitrate for senders> (e.g., 128000)
 ; fir_freq = <send a FIR to publishers every fir_freq seconds> (0=disable)
+; videocodec = vp8|vp9|h264 (video codec to force on publishers, default=vp8)
 ; record = true|false (whether this room should be recorded, default=false)
 ; rec_dir = <folder where recordings should be stored, when enabled>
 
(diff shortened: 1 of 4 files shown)


mtdxc:
Great. I just implement this on my own repo

lminiero:
Any feedback? Objections to a merge?

cacheworks:
I've merged this and configured a VP9 room 5 days ago, tested a bit and didn't notice any problems with the gateway
Great option to have, nice that it's configurable for rooms so we can test (It seems that VP9 in Chrome is still unstable for some users)

lminiero:
Thanks! If no one else objects, I'll merge this later on.

arjenveenhuizen:
I think this is a very clean solution! Would be great if you merge it into master, so by all means Lorenzo :)

lminiero:
Thanks for the additional feedback! Merging, then: later on we can improve this further, e.g., to do the same for audio codecs as well.

Issue #485, opened 2016-03-14
Streaming mountpoint stats


marchaase:
I'd like to be able to tell if a remote RTP stream is "active".  I have a remote RTP stream that is sometimes active and sometimes not.  Currently if I try to stream that mountpoint it will attempt it and from the client's perspective it just never loads the video.  I don't see any easy way to detect that the remote RTP stream is active or not from the client's perspective.

I was wondering if it would make sense to add some statistics for mountpoints.  Something that could be queried with plugin message API to allow a client to have some logic to detect if a stream is active.  It looks like something could be added to the struct janus_streaming_mountpoint.  The mountpoint receiving thread could update information into the struct to allow the message handling thread to dole out that information.

I've hacked together something really quickly to see how easy it would be and made the receiving thread update a new struct member with the value of last_ts.  I then changed the message API to just add this information alongside the mountpoint info in the "list" command.

last_ts doesn't seem ideal for use here as it looks like it's for keeping the timestamp sensible if source streams are switched.  Potentially the sequence number might work.

I think the biggest question I have is what types of data would be the easiest to add that could cover all of the different stream types.  Maybe mirroring the admin API for the stats giving bytes/packets for audio/video/data would be a good target for stats covered.  Getting bytes seems easy and would work for network or 'live' file sources, but wouldn't work too well with 'ondemand' though as you have multiple clients with different stats.

I think there is a question about where in the message API this would fit best as well.  'list' makes the most sense to me to fit into the existing messages, but is this too much information for that response?  Should there be another message like 'info' that could give you more detailed info on the mountpoint?

Thoughts?

marchaase:
Maybe as this is an enhancement idea this is better suited for the for the google forum?  You can close this if you like and I'll post there.

amnonbb:
Here 
https://github.com/cargomedia/janus-gateway-rtpbroadcast
i found some extended modified stream plugin.
but did not check yet

lminiero:
I think having an issue here as a "desiderata" is not a bad thing, so no need to move it to the group. Feel free to post there to talk about it there as well, if not just to point to the issue here and see if anybody else is interested.

[...]

PR #486, opened 2016-03-16
RTP stream timestamps

marchaase:
#485 

I added rtp timestamps to the streaming plugin.  It adds two variables to the janus_streaming_mountpoint struct: last_received_video and last_received_audio.  It uses janus_get_monotonic_time to timestamp when an RTP packet has arrived and updates the struct's variable.  In the message handling for the 'list' command it outputs the timestamps if it's an RTP stream and a 'now' json member which is the current monotonic time.  This allows a client to calculate the time since the last packet (audio or video) has arrived and decide if the stream has timed out/stopped sending packets.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -1020,8 +1020,10 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session 
 			if(mp->streaming_source == janus_streaming_source_rtp) {
 				janus_streaming_rtp_source *source = mp->source;
 				gint64 now = janus_get_monotonic_time();
-				json_object_set_new(ml, "audio_age_ms", json_integer((now - source->last_received_audio) / 1000));
-				json_object_set_new(ml, "video_age_ms", json_integer((now - source->last_received_video) / 1000));
+				if ( source->audio_fd != -1 )
+					json_object_set_new(ml, "audio_age_ms", json_integer((now - source->last_received_audio) / 1000));
+				if ( source->video_fd != -1 )
+					json_object_set_new(ml, "video_age_ms", json_integer((now - source->last_received_video) / 1000));
 			}
 			json_array_append_new(list, ml);
 		}
@@ -2771,8 +2773,8 @@ janus_streaming_mountpoint *janus_streaming_create_rtp_source(
 	live_rtp_source->vrc = NULL;
 	live_rtp_source->audio_fd = audio_fd;
 	live_rtp_source->video_fd = video_fd;
-	live_rtp_source->last_received_audio = 0;
-	live_rtp_source->last_received_video = 0;
+	live_rtp_source->last_received_audio = janus_get_monotonic_time();
+	live_rtp_source->last_received_video = janus_get_monotonic_time();
 	live_rtp_source->keyframe.enabled = bufferkf;
 	live_rtp_source->keyframe.latest_keyframe = NULL;
 	live_rtp_source->keyframe.temp_keyframe = NULL;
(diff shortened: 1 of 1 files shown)


ploxiln:
It might be a nicer api/interface if you calculate the age in milliseconds (`(now - last_recieved)/1000000`) In this case the json properties would be something like "audio_age_ms" and "video_age_ms", and no "now" element.

marchaase:
I do like that idea - that way the json member is self descriptive. An
integer is precise enough in ms for this purpose.

lminiero:
I'd move the timing info to the ```janus_streaming_rtp_source``` struct instead, since they only make sense for RTP based feeds.

lminiero:
As a side note, if you haven't signed our [CLA](https://cla.conf.meetecho.com/janus/) yet, please do so before we can merge your contribution.

marchaase:
Signed the CLA.

Implemented ms API facing changes as @ploxiln suggested.
Moved last_received memebrs to rtp struct.
Also changed the API to only output the age in ms if that part of the stream is enabled.

Changing the location of the last_received_* members also means they aren't in the mutex protected section.  There is not mutex for the rtp struct, so I didn't add one for now.  @lminiero it sounds like you are of the mind that writing to a gint64 in one thread and reading the gint64 in another thread shouldn't be an issue?  I think 64-bit values are atomic on a 64-bit architecture if they are memory aligned, but they are not on a 32-bit arch.  Probably not that bad of a problem since it's a monotonic counter - there is one possible race condition when the carry over from the 32nd bit to the 33rd bit occurs and if the read happens right there on a 32-bit arch.

lminiero:
You're right that this can happen, but I think it's unlikely enough that we can ignore that. Anyway, if you feel more comfortable with having a lock anyway, I think you can use the same lock you were using before, that is the lock of the mountpoint itself: after all, the source is a property of the mountpoint.

lminiero:
@marchaase as a side note, are you sure you signed the CLA? I can't find you in the list, vould you try again? Thanks!

plugins/janus_streaming.c
```diff
@@ -1015,6 +1017,14 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session
 			json_object_set_new(ml, "id", json_integer(mp->id));
 			json_object_set_new(ml, "description", json_string(mp->description));
 			json_object_set_new(ml, "type", json_string(mp->streaming_type == janus_streaming_type_live ? "live" : "on demand"));
+			if(mp->streaming_source == janus_streaming_source_rtp) {
+				janus_streaming_rtp_source *source = mp->source;
+				gint64 now = janus_get_monotonic_time();
+				if ( source->audio_fd != -1 )
+					json_object_set_new(ml, "audio_age_ms", json_integer((now - source->last_received_audio) / 1000));
+				if ( source->video_fd != -1 )
+					json_object_set_new(ml, "video_age_ms", json_integer((now - source->last_received_video) / 1000));
```
lminiero: Just for our coding conventions, could you remove those extra spaces in the two ifs?

    if(source->audio_fd != -1)

For the rest, the patch looks fine to me and, apart from the considerations on whether or not we need locking (something we can fix along the way, if needed) we can merge it. Thanks for working on this!

lminiero: Nevermind that, it's such a minor fix that I can do this myself after we merge. Thanks!

marchaase:
Cool, sounds good Lorenzo.  Hard to force yourself into a different coding style sometimes =)

I really like the janus-gateway software, so I'm glad that I'm able to contribute what little small parts I can.

And the locking can be addressed easily if needed later, just got to move that update into the mutex protected section at the bottom of the video/audio if branches.

Issue #484, opened 2016-03-14
Janus SIP plugin hangs up when there is issue SIP signalling


nbansalvox:
While doing some basic testing with Janus SIP plugin, I noticed that for certain call flows, Janus SIP plugin starting spitting the following error and becomes unresponsive:
`[plugins/janus_sip.c:janus_sip_relay_thread:2883] [SIP-click2vox] Error polling: POLLERR.`
It seems that issue is reproducible, here is the SIP call flow resulting in the error:
Janus     SIP INVITE --------> Remote end point
Janus     2xx       <----------    Remote end point (2xx contain Record  Route header address)
Janus    ACK      ------------> SIP address in Contact header of 2xx
Issue in this case is that  ACK should have been sent to the address in Record-Route header rather than IP address in Contact header of 2xx.
This SIP signalling always triggers the above said issue with Janus.

saghul:
I believe there is some problem with Sofia SIP or with how we use it. Do you get the same issue if you specify the proxy? (do so by using an IP address, if you put a domain it will make a query for each request)

lminiero:
That's what I suggested [here](https://groups.google.com/forum/#!topic/meetecho-janus/dbmAkHaug3w), referring you your comment in #451, I don't know if the POLLERR thing also happens when doing that. My guess is that the missing ACK leads to the local RTP/RTCP sockets being closed by the code somewhere, but the media loop stays up: I don't know if that's the case though.

nbansalvox:
The issue doesn't happen when I use configure the proxy IP address while sending REGISTER reqeust to Janus. 
But in the failing case, I think error isn't handled graceful, it stays in the loop of POLLERR, whereas I think it should just hangup the call gracefully

lminiero:
That's what the patch I made last weekend should do: https://github.com/meetecho/janus-gateway/commit/48967c3e69d5cbbdeff6e3f96d3640bc0129ae10
Have you tried if that commit fixed the issue for you?

lminiero:
Can this have anything to do with it? https://sourceforge.net/p/sofia-sip/mailman/message/8147409/
I'm not a 100% expert on SIP, but I guess they're saying you have to use the Contact address whenever you get one (which is what is happening in the 200 OK, at least in my dumps), unless strict routing is involved (in your case it's lr instead).

lminiero:
[...]

Issue #451, opened 2016-01-29
Janus not respecting SIP Record-Route


s-schumann:
Hi

Janus does not seem to respect Record-Route when sending in-dialog SIP messages.

We are using the SIP Webphone and found the following happen:
- Out of dialog INVITE/200 OK contain Record-Route headers
- ACK message builds Route headers accordingly and correctly
- ACK is sent directly to the other party (to Contact IP:Port) instead of the IP destination of the first Route 

Edit: If "proxy" is filled it works as expected, but I would interpret this as outbound proxy. If the proxy is found based on domain I would expect similar behavior.

Best regards
Sebastian

saghul:
Hi Sebastian!

Generally speaking I found Sofia SIP acts quite weird: if you pass a non-resolved proxy, it will do the resolution *for every request*! So, I ended up making the resolution myself and passing the resulting IP address as the proxy.

s-schumann:
Hi Saul
Thank you for replying.
For me, adding the proxy manually also worked. I've put the IP now and it is well usable :)
Strange though that Sofia SIP does not support Record-Route for my scenario w/o proxy, it isn't used that rare I think :(
Sebastian

saghul:
Oh, it most certainly supports it. Alas the documentation is hard to follow, and the project is currently abandoned AFAIK.

lminiero:
Do you guys think we should handle this "resolve the address" part in the plugin code, rather than forcing users to resolve externally themselves and then pass the proxy IP address?

saghul:
Ideally yeah, I think so. I thought about using libunbound for it, but
never found the time to actually jump in and do it.

lminiero:
Why another library instead of just using ```getaddrinfo```?

saghul:
`getaddrinfo` doesn't do NAPTR nor SRV. We need a library capable of doing NAPTR and SRV lookups, and implement RFC 3263 with it.

lminiero:
Got it, tnx for the clarification! I thought it did as there's a service as part of the arguments, but I never actually played with that so I was just guessing.

lminiero:
Would c-ares work as well? Just wondering about which library is most likely to be already available, or more in general most likely to be available in distro repos.

lminiero:
Answering myself after reading a bit on this, I guess you suggested libunbound for its ability to cache results which c-ares apparently doesn't have, am I right?

saghul:
[...]

PR #490, opened 2016-03-18
Fix Sofia SIP when both Record-Route and Contact are there

lminiero:
This PR tries to address issues #484 and, potentially, #451 too.

The main problem is that apparently Sofia SIP is not respecting the Record-Route set in responses: in particular, @nbansalvox found out that, when sending an INVITE to a Proxy registering as a guest, a ```Contact``` header present in the ```200 OK``` would override any ```Record-Route``` set in the same response, thus leading our stack to send an ACK to the wrong place and breaking the call.

Long story short, if you're too lazy to read my explaination a few lines below, the only easy way I could find to fix this broken behaviour was to:

1. disable the automatic ACK when receiving a 200 OK, so that we could send it ourselves;
2. check if there's a ```Record-Route``` header present and, if so, use ```NTATAG_DEFAULT_PROXY``` to force the Sofia stack to pass through that proxy from there on for that dialog.

Now, ```NTATAG_DEFAULT_PROXY``` is probably an ugly way to do that, and it may cause problems if we don't handle this properly (e.g., re-adjusting this in case the route is changed again later and so on), but it apparently is the _only_ way to get this done using Sofia SIP's NUA stack.

Coming to the more detailed explaination (feel free to zone out if you don't care, but please don't if you want to provide feedback and suggestions), the reason for this relies in how the internals of Sofia SIP handle the whole transaction/routing process. Digging in the code these past few days, trying to figure out where the broken behaviour might happen and why, I think I found out where the problem occurs, and more specifically in a method called ```leg_route``` in ```nta.c```:

https://github.com/jart/sofia-sip/blob/master/libsofia-sip-ua/nta/nta.c#L4991

Looking at the code there, it appears that the route is indeed handled, but that if a ```Contact``` header has been provided then it takes precedence, as it was outlined in the original issue that was reported. Now, that method is called in a few parts, more importantly from a method called ```nta_leg_client_reroute```, which in turn is called by ```nua_dialog_uac_route```. That method, interestingly enough, is called in ```nua_client_response```, which is where a SIP response is handled (e.g., a 200 OK), and apparently only when (surprise surprise!) a ```Contact``` header is found:

https://github.com/jart/sofia-sip/blob/master/libsofia-sip-ua/nua/nua_client.c#L1049

Now, I initially thought that fixing this behaviour could be done by using those same methods, by managing the transmission of the ACK ourselves (hence the ```NUTAG_AUTOACK(0)```, or otherwise the ACK is automatically sent to the broken route), intercepting a ```Record-Route``` header, and enforcing it using ```nua_dialog_uac_route``` or something like that. Unfortunately, that's not possible, as while the leg/agent info is obviously part of the ```nua_t``` Sofia SIP uses as the NUA stack, every structure in Sofia SIP is completely opaque and hidden to the application: this means there is no means for us, from the plugin, to directly access the leg or the agent needed as arguments to those methods, and there isn't even any complementar method to get the same information from the ```nua_t``` instance itself.

As such, I thought about a few alternatives, namely:

1. copying the private structures from ```nua_stack.h``` within our plugin as well, in order to have access to those definitions in a brutally hacky way (which is a mess not only because it's hacky, but also because there are tons of cross references to other private structures, all of which should be added if you need to figure out the proper offsets of the members);
2. moving from the NUA to NTA, that is stop using the UserAgent helpers Sofia SIP provides and handle all the SIP transactions and states ourselves directly, something I have absolutely no intention of doing (I like SIP, but not that much);
3. using a different approach to force Sofia SIP to go where I want it to.

After realizing how 1. and 2. were not an option, I tried looking for a way to do 3., and that's how I found ```NTATAG_DEFAULT_PROXY```, which is documented here:

http://sofia-sip.sourceforge.net/refdocs/nta/nta__tag_8h.html#a470c3d7c598ed2d4514836d3bb916fc7

    URL for (default) proxy.

    The requests are sent towards the default outbound proxy regardless the values of
    request-URI or Route headers in the request. The URL of the default proxy is not
    added to the request in the Route header or in the request-URI (against the
    recommendation of RFC 3261 section 8.1.2).

    The outbound proxy set by NTATAG_DEFAULT_PROXY() is used even if the dialog
    had an established route set or registration provided User-Agent with a
    Service-Route set.

So, probably a bit hacky and dirty, but it seems to work in my case when trying to replicate the sample scenario described in #484.

Any feedback on this? Any suggestion on how we could do this better? Any hint on where we should do this, besides ```nua_r_invite``` as I'm doing now for this proof of concept? Any places/scenarios this could go horribly wrong?

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -1646,6 +1646,7 @@ static void *janus_sip_handler(void *data) {
 				NUTAG_PROXY(session->account.proxy),
 				TAG_IF(strlen(custom_headers) > 0, SIPTAG_HEADER_STR(custom_headers)),
 				NUTAG_AUTOANSWER(0),
+				NUTAG_AUTOACK(0),
 				TAG_END());
 			g_free(sdp);
 			sdp_parser_free(parser);
@@ -2408,6 +2409,13 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				sdp_parser_free(parser);
 				break;
 			}
+			/* Send an ACK */
+			char *route = sip->sip_record_route ? url_as_string(session->stack->s_home, sip->sip_record_route->r_url) : NULL;
+			JANUS_LOG(LOG_WARN, "Sending ACK (route=%s)\n", route ? route : "none");
+			nua_ack(nh,
+				TAG_IF(route, NTATAG_DEFAULT_PROXY(route)),
+				TAG_END());
+			/* Parse SDP */
 			JANUS_LOG(LOG_VERB, "Peer accepted our call:\n%s", sip->sip_payload->pl_data);
 			session->status = janus_sip_call_status_incall;
 			char *fixed_sdp = g_strdup(sip->sip_payload->pl_data);


nbansalvox:
Hi Lorenzo,
Can't we modify the Sofia-sip code to disable that either, we can define a config flag, SIP clients who want Contact to take precedence over Record-Route, they can do so in their configuration, others can give priority to Record-Route. I really don't see many scenarios where you would like to give precedence to Contact header. 
NTATAG_DEFAULT_PROXY is quite common too, Jain SIP has similar option, in case, we can't modify Sofia-SIP, you could consider not enforcing NTATAG_DEFAULT_PROXY for all the Janus users, it might be a good idea to make it opt-in?

Nitesh

lminiero:
We can't make changes to Sofia SIP, it's a library most users install from repos, and I'd rather not maintain a fork somewhere just for that and force users to use that instead (especially considering they may be using it in other projects as well).

As to NTATAG_DEFAULT_PROXY, despite what the name suggests setting it doesn't override it for _all_ users, but only for the dialog you set it on. In this case, we're setting it before ACKing, which means that until you end the session with a BYE it will be active only for that session.

nbansalvox:
Well, if you enable it for all the dialogs before ACKing, doesn't it mean it is enabled for all the SIP dialogs 5SIP outbound dialogs initiating the call), because all SIP dialogs need an ACK.

lminiero:
That's not what the patch does. It only enables it, before ACKing, to that specific dialog, if and only if a Record-Route is present in that dialog's 200 OK. If it's not present, the ACK is sent without that argument (```TAG_IF``` is false). All other dialogs are not affected no matter what.

lminiero:
Merging this as it fixed that broken scenario without affecting other use cases. In case this is needed on a wider scale (e.g., for other messages, or call topologies), we can worry about that later on.

PR #492, opened 2016-03-23
Support for other codecs and formats in recorder and post-processor

lminiero:
As you know, both the recorder and the post-processor in master currently assume Opus is always used for audio and VP8 is always used for video. With the multiple codecs WebRTC supports, that's not always the case, and that's what this PR tries to address.

This PR changes the way plugins can create recorders, so that they can specify what the .mjr files will contain (e.g., Opus vs. G.711, or VP8 vs. VP9 vs. H.264). At the same time, the PR also extends the post-processor, so that these new codecs can be handled and post-processed to an external media file as well. More precisely, VP9 is converted to a .webm file as VP8 was before, H.264 to a .mp4 file, and G.711 audio is converted to a plain .wav file instead. Again, with the exception of G.711 (which is decoded to raw frames), no transcoding happens: the payload is simply extracted from the ordered RTP packets, and dumped to the right media container.

This is a PR rather than a direct merge as I've had to modify a few things in the way recorders are created, which means this could break existing plugins that still rely on the old way of creating recorders. Besides, I took advantage of this "breaking" change to also simplify the way payload types associated to specific codecs are read from the SDP, as the number of methods "exploded" when we added support for configurable codecs in the VideoRoom plugin.

If you're a plugin developer, you need to take care of these changes:

1. ```janus_recorder_create``` now expects a string with the codec name, rather than a simple integer stating whether the content will be audio or video: so, for recording Opus audio, for instance, you'll need to do something like ```janus_recorder_create(dir, "opus", filename);``` instead of ```janus_recorder_create(dir, 0, filename);``` (notice ```"opus"``` where ```0``` was).
2. ```janus_get_opus_pt``` and ```janus_get_vp8_pt``` don't exist anymore: now there's a single method called ```janus_get_codec_pt```, which besides the ```sdp``` expects a string with the ```codec``` name to look for (e.g., "opus" or "h264").
3. there's a new method called ```janus_get_codec_from_pt``` that can do the other way around: you pass the SDP and the payload type, and it tells you which codec it refers to; useful if you see media flowing in your plugin after a negotiation occurred, and you want to know which codec ended up being used. Still not used in any plugin but I'll probably do that (e.g., in the SIP plugin where recording still assumes Opus/VP8 because I'm lazy).

That's about it. If you're playing with different codecs, especially in the VideoRoom plugin, please test this new changes and let me know if they work as expected. I did some test recordings and processing for all the new codecs (G.711, VP9, H.264) and they all seemed to be ok, but there's only so much I try do in a few runs.


utils.c
```diff
@@ -424,36 +282,51 @@ int janus_get_vp9_pt(const char *sdp) {
...
-					return pt;
-				}
-			} else if(strstr(line, "a=rtpmap") && strstr(line, "h264/90000")) {
-				/* Gotcha! */
-				int pt = 0;
-				if(sscanf(line, "a=rtpmap:%d h264/90000", &pt) == 1) {
-					*next = '\n';
-					return pt;
+					if(strstr(name, "vp8") || strstr(name, "VP8"))
+						return "cp8";
```
ploxiln: typo "cp8"

lminiero: D'oh, thanks for spotting this, I'll fix this in the morning.

pallab-gain:
Thank you for adding G.711 support. We are have been testing it for two days, and we are happy with the quality of recording, and overall performance with PCMU audio codec. Our QA team didn't encounter any bugs so far during their test so far with recording. Thank you so much for the efforts, and time.

lminiero:
@pallab-gain thanks for the feedback! I'll keep this as a PR for a little while longer due to the few changes required on plugins, and to give 3rd-party plugin developers time to handle this.

lminiero:
Just as a heads up, I'm planning to merge this soon.

lminiero:
As anticipated, merging! The plugin API bump should make sure that third party plugins have to be recompiled, and so those who are using the `janus_recorder_create` stuff are made aware of the change.

PR #494, opened 2016-03-26
Removed duplicate calls to janus_mutex_unlock(&rooms_mutex)

andreasg123:
I noticed duplicate calls to janus_mutex_unlock(&rooms_mutex) in janus_videoroom.c and removed them.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -561,7 +561,7 @@ int janus_pidfile_remove(void) {
 
 void janus_get_json_type_name(int jtype, gboolean positive_non_empty, char *type_name) {
 	/* Longest possible combination is "a non-empty boolean" plus one for null char */
-	gsize req_size = 20
+	gsize req_size = 20;
 	if(positive_non_empty) {
 		g_strlcpy(type_name,
 			jtype == JSON_INTEGER || jtype == JSON_REAL ?
--- a/utils.c
+++ b/utils.c
@@ -558,3 +558,76 @@ int janus_pidfile_remove(void) {
 	g_free(pidfile);
 	return 0;
 }
+
+void janus_validate_json_object(json_t *root, struct janus_json_parameter *params, unsigned int nparams, int *error_code, char *error_cause, int error_cause_size, int error_missing_element, int error_invalid_element) {
+	*error_code = 0;
+	unsigned int i;
+	for(i = 0; i < nparams; i++) {
+		json_t *val = json_object_get(root, params[i].name);
+		if(!val) {
+			if(params[i].required) {
+				JANUS_LOG(LOG_ERR, "Missing element (%s)\n", params[i].name);
+				*error_code = error_missing_element;
+				if(error_cause != NULL)
+					g_snprintf(error_cause, error_cause_size, "Missing element (%s)", params[i].name);
+				return;
+			}
+			continue;
+		}
+		/* Use JSON_TRUE instead of the non-existing JSON_BOOLEAN */
+		gboolean is_valid = (json_typeof(val) == params[i].jtype ||
+												 (params[i].jtype == JSON_TRUE && json_typeof(val) == JSON_FALSE));
+		if(is_valid && params[i].positive_non_empty)
+			switch(params[i].jtype) {
+			case JSON_INTEGER:
+				is_valid = (json_integer_value(val) >= 0);
+				break;
+			case JSON_REAL:
+				is_valid = (json_real_value(val) >= 0);
+				break;
+			case JSON_STRING:
+				is_valid = (json_string_length(val) > 0);
+				break;
+			case JSON_ARRAY:
+				is_valid = (json_array_size(val) > 0);
+				break;
+			default:
+				break;
+			}
+		if(!is_valid) {
+			char type_name[20] = "";
+			if(params[i].positive_non_empty)
+				strcat(type_name,
+							 params[i].jtype == JSON_INTEGER || params[i].jtype == JSON_REAL ?
+							 "positive " : "non-empty ");
+			switch(params[i].jtype) {
+			case JSON_TRUE:
+				strcat(type_name, "boolean");
+				break;
+			case JSON_INTEGER:
+				strcat(type_name, "integer");
+				break;
+			case JSON_REAL:
+				strcat(type_name, "real");
+				break;
+			case JSON_STRING:
+				strcat(type_name, "string");
+				break;
+			case JSON_ARRAY:
+				strcat(type_name, "array");
+				break;
+			case JSON_OBJECT:
+				strcat(type_name, "object");
+				break;
+			default:
+				break;
+			}
+			gboolean is_vowel = (strchr("aeiou", type_name[0]) != NULL);
+			JANUS_LOG(LOG_ERR, "Invalid element (%s should be %s %s)\n", params[i].name, is_vowel ? "an" : "a", type_name);
+			*error_code = error_invalid_element;
+			if(error_cause != NULL)
+				g_snprintf(error_cause, 512, "Invalid element (%s should be %s %s)", params[i].name, is_vowel ? "an" : "a", type_name);
+			return;
+		}
+	}
+}
--- a/utils.c
+++ b/utils.c
@@ -559,24 +559,25 @@ int janus_pidfile_remove(void) {
 	return 0;
 }
 
-void janus_validate_json_object(json_t *root, struct janus_json_parameter *params, unsigned int nparams, int *error_code, char *error_cause, int error_cause_size, int error_missing_element, int error_invalid_element) {
+void janus_validate_json_object(json_t *root, struct janus_json_parameter *params, unsigned int nparams, int *error_code, char *error_cause, int error_cause_size, gboolean log_error, int error_missing_element, int error_invalid_element) {
 	*error_code = 0;
 	unsigned int i;
 	for(i = 0; i < nparams; i++) {
 		json_t *val = json_object_get(root, params[i].name);
 		if(!val) {
 			if(params[i].required) {
-				JANUS_LOG(LOG_ERR, "Missing element (%s)\n", params[i].name);
 				*error_code = error_missing_element;
+				/* This message was "Missing element" in the videoroom plugin. */
+				if(log_error)
+					JANUS_LOG(LOG_ERR, "Missing mandatory element (%s)\n", params[i].name);
 				if(error_cause != NULL)
-					g_snprintf(error_cause, error_cause_size, "Missing element (%s)", params[i].name);
+					g_snprintf(error_cause, error_cause_size, "Missing mandatory element (%s)", params[i].name);
 				return;
 			}
 			continue;
 		}
 		/* Use JSON_TRUE instead of the non-existing JSON_BOOLEAN */
-		gboolean is_valid = (json_typeof(val) == params[i].jtype ||
-												 (params[i].jtype == JSON_TRUE && json_typeof(val) == JSON_FALSE));
+		gboolean is_valid = (json_typeof(val) == params[i].jtype || (params[i].jtype == JSON_TRUE && json_typeof(val) == JSON_FALSE));
 		if(is_valid && params[i].positive_non_empty)
 			switch(params[i].jtype) {
 			case JSON_INTEGER:
@@ -595,6 +596,7 @@ void janus_validate_json_object(json_t *root, struct janus_json_parameter *param
 				break;
 			}
 		if(!is_valid) {
+			*error_code = error_invalid_element;
 			char type_name[20] = "";
 			if(params[i].positive_non_empty)
 				strcat(type_name,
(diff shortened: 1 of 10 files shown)


lminiero:
Thanks for spotting these, I'll review the changes early next week and merge in case.

andreasg123:
I noticed that there was much code duplication for validating properties of JSON objects.  I created a data structure describing those properties (type, positive integer, required) and a function for validating them (janus_validate_json_object).  I only converted some of the code to illustrate and test the concept.  If you like, I could replace the rest of the validation, too.

lminiero:
The idea is nice, and if we could make this generic enough (say, in utils.c/h) so that both core and plugins can use it that would be even greater. Anyway, I'll look into this early next week as well (sorry, holiday period here). A first, quick feedback I can give is to please adhere to our coding style, especially with respect to indentation, whitespaces, etc.: we have an .editorconfig file as part of the project, in case yur IDE supports it.

That said, please also sign our [CLA](https://cla.conf.meetecho.com/janus/) if you haven't done already, as we can't merge anything without that.

Thanks!

andreasg123:
I addressed your requests and moved janus_validate_json_object into utils.c. Because each plugin uses different error codes to indicate missing and invalid elements, I added those error codes as parameters. I expanded the validation to (non-empty) arrays. It would also work for objects and reals but those aren't used in the videoroom plugin. I applied the new validation approach to the rest of janus_videoroom.c.

andreasg123:
I changed my validation function so that it also can work in janus.c and added a sample. Because janus.c uses different error messages (e.g., "Invalid element type" instead of "Invalid element"), I changed the error messages to match those. Thus, the video room plugin would produce slightly different error messages. In janus.c, I had to add the function janus_process_error_string to avoid compile warnings.

I'll leave it at that. I don't think that further changes would be beneficial before you have a chance to review this approach in detail.

lminiero:
Thanks for making the changes so quickly!

I gave a quick look at the patch and it seems fine. I'm a bit concerned by the fact that, on the one hand, this seems to be making things a bit more complex, as since it needs to be generic enough we need "templates" for all messages we'll be handling (and I think we have a couple of methods somewhere that can have a different "signature" according to some parameters: you can blame the lazyness on me!). On the other hand, though, it makes things much cleaner in the message parsing/processing code, and having templates in place would probably help disciplining a bit the API, both existing and future (and might also make API study easier for newcomers).

In the next few days I'll go deeper in the code and provide more specific comments, thanks! (we'll fly to Buenos Aires for the IETF, tomorrow, and will be there for about 10 days, so I may be a bit slow with comments, I apologize in advance).

andreasg123:
I started writing a function janus_validate_json_object_format that would take two format strings for the two error cases. Unfortunately, I ran into the problem that JANUS_LOG only accepts string constants as format string due to the concatenation of that string in the call to JANUS_PRINT. This would require expanding the JANUS_LOG macro by hand in the new function.

Also, even with my current implementation, the logged file and line would be in utils.c and not where the JSON object was validated. To address this issue, part of janus_validate_json_object would have to be a macro.

Still, I think that the data-driven validation approach makes it much clearer what parameters are expected. It helped me understand what I had to provide on the JavaScript side.

Enjoy your trip to Buenos Aires.

lminiero:
> Still, I think that the data-driven validation approach makes it much clearer what parameters are expected. It helped me understand what I had to provide on the JavaScript side.

Yep, I agree, and that's what I meant when I mentioned making the API easier to study and understand for newcomers, especially where it's not thouroughly documented. It's definitely a very welcome contribution, so thanks again for working on this!

andreasg123:
I think that I got it now. As I suggested earlier, I replaced janus_validate_json_object with the macro JANUS_VALIDATE_JSON_OBJECT. Now error messages list the correct file, e.g.:

    [ERR] [plugins/janus_videoroom.c:janus_videoroom_handle_message:1184] Missing mandatory element (request)

I also provide the macro JANUS_VALIDATE_JSON_OBJECT_FORMAT for different error messages. I applied it to JSEP errors to illustrate the concept.

	static struct janus_json_parameter jsep_parameters[] = {
		{"type", JSON_STRING, FALSE, TRUE},
		{"trickle", JSON_TRUE, FALSE, FALSE},
		{"sdp", JSON_STRING, FALSE, TRUE}
	};
	
	JANUS_VALIDATE_JSON_OBJECT_FORMAT("JSEP error: missing mandatory element (%s)",
		"JSEP error: invalid element type (%s should be %s)",
		root, jsep_parameters, error_code, error_cause, FALSE,
		JANUS_ERROR_MISSING_MANDATORY_ELEMENT, JANUS_ERROR_INVALID_ELEMENT_TYPE);
	if(error_code != 0) {
		ret = janus_process_error_string(request, session_id, transaction_text, error_code, error_cause);
		goto jsondone;
	}

lminiero:
Thanks for these updates! Sorry for the lack of feedback, but I'm currently abroad and won't have much time to look into this. I'll do a deeper review when I come back.

andreasg123:
I only now noticed that _janus_videoroom_handle_message_ and _janus_videoroom_handler_ use different patterns for allocating _error_cause_. When switching to the macro, I simplified the parameter list by assuming that _error_cause_ would always be allocated on the stack.  Otherwise, _sizeof_ won't work. If that's not a valid assumption, _error_cause_size_ should be added to the parameter list.

lminiero:
I think the assumtpion is fine (sorry for the late reply, just back and still a bit jetlagged!).

utils.c
```diff
@@ -558,3 +558,58 @@ int janus_pidfile_remove(void) {
 	g_free(pidfile);
 	return 0;
 }
+
+void janus_get_json_type_name(int jtype, gboolean positive_non_empty, char *type_name) {
+	if(positive_non_empty)
```
lminiero: Can you add brackets ```{...}``` to this? It makes it less ambiguous, since the action is on multiple lines.

utils.c
```diff
@@ -558,3 +558,58 @@ int janus_pidfile_remove(void) {
...
+		strcpy(type_name,
```
lminiero: Use the Glib alternatives where possible (so ```g_strcpy``` instead of ```strcpy```, ```g_strcat``` instead of ```strcpy```) as they allow for easier portability and are ostensibly safer too.

utils.c
```diff
@@ -558,3 +558,58 @@ int janus_pidfile_remove(void) {
...
+			jtype == JSON_INTEGER || jtype == JSON_REAL ?
+			"a positive " : "a non-empty ");
+	else if(jtype == JSON_INTEGER || jtype == JSON_ARRAY || jtype == JSON_OBJECT)
```
lminiero: Better to add brackets to all those elses as well, then.

utils.c
```diff
@@ -558,3 +558,58 @@ int janus_pidfile_remove(void) {
...
+		strcpy(type_name, "an ");
+	else
+		strcpy(type_name, "a ");
+	switch(jtype) {
+	case JSON_TRUE:
```
lminiero: As part of our code style, we indent all cases and their content.

utils.c
```diff
@@ -558,3 +558,58 @@ int janus_pidfile_remove(void) {
...
+		strcat(type_name, "object");
+		break;
+	default:
+		break;
+	}
+}
+
+gboolean janus_json_is_valid(json_t *val, json_type jtype, gboolean positive_non_empty) {
+	gboolean is_valid = (json_typeof(val) == jtype || (jtype == JSON_TRUE && json_typeof(val) == JSON_FALSE));
+	if(is_valid && positive_non_empty)
```
lminiero: See above for brackets.

utils.c
```diff
@@ -558,3 +558,58 @@ int janus_pidfile_remove(void) {
...
+		switch(jtype) {
+		case JSON_INTEGER:
+			is_valid = (json_integer_value(val) >= 0);
```
lminiero: See above for indentation.

utils.h
```diff
@@ -139,4 +147,40 @@ int janus_pidfile_create(const char *file);
 /*! \brief Unlock and remove a previously created PID file
  * @returns 0 if successful, a negative integer otherwise */
 int janus_pidfile_remove(void);
+
+void janus_get_json_type_name(int jtype, gboolean positive_non_empty, char *type_name);
```
lminiero: Can you add a doxygen style documentation for the new methods in this header? You can use the other methods as an example. This would have your new methods automatically be added to the docs, when generated. It doesn't have to be 100% detailed, as I can edit it later on: it's enough to clarify what's what and what the method does in principle.

lminiero:
I've started a review of the code and, form what I could see so far, good work! I've just added a few comments, mostly on coding guidelines to better adhere to our "style". Apart from that, it does indeed seem quite easy to re-use in all the other plugins without too much effort, and should make the code cleaner.

I'm just wondering whether in the docs we should suggest any size for the involved strings you modify? Since there are strcpy and strcat involved, strings can be larger than once might expect. Anyway, since in the original code we're already using large strings buffers for that (e.g., 512 bytes) that might not be needed,

andreasg123:
I believe that I covered all your comments.

To support other developers that try to follow your coding style while using Emacs, I added a function for ~/.emacs below (https://www.emacswiki.org/emacs/IndentingC).

    (defun janus-mode ()
      (interactive)
      ; c-basic-offset and tab-width can be set to personal preference as long as
      ; they are the same. They will expand to the single tab required by the Janus
      ; coding style.
      (setq c-basic-offset 2
            tab-width 2
            indent-tabs-mode t)
      (c-set-offset 'arglist-intro '+)
      (c-set-offset 'arglist-cont-nonempty '+)
      (c-set-offset 'case-label '+))

lminiero:
Thanks! I'll give a more thorough review tomorrow. I'll probably merge next Monday or Tuesday, as I'll be away all week, and I'd like to first change the other plugins myself to use your stuff so that I can push them immediately after I merge your contribution.

andreasg123:
I changed the other plugins so that you can focus on double-checking everything. I cherry-picked your commits after I forked so that pulling should not run into any problems.

I created the new macros JANUS_CHECK_PIN and JANUS_CHECK_SECRET to combine the validation of the "pin" and "secret" parameters with the comparison against the required pin or secret.  That happened a total of 11 times in the audiobridge, streaming, and videoroom plugins.

ploxiln:
Cherry-picking commits from master into this branch is not a great idea. Even if git can cleanly merge this branch with master, there will be multiple distinct copies of these commits in the same git history. The two better alternatives are to either merge master into this branch _or_ rebase this branch on master (_or_ if changes to master are not significant and there are no merge conflicts, don't do anything).

Also, your proposed changes would likely be easier for Lorenzo to review and would get merged sooner, if logically separate changes were put in separate pull requests. Yes they depend on each other in a series so they wouldn't be super easy to manage, and you'd have to wait for one to be merged before the next could be put into final form for clean review... but the original somewhat serious double-mutex-unlock fix would probably be merged already. (emacs.el would be very easy to put in a separate PR to just make this one a bit shorter.)

andreasg123:
I thought that git would be smarter about merging the same commit.  I'll know better in the future.

Also, I'll try to remember to create a branch for each distinct change so that it can be a separate PR. I removed the commit with emacs.el.

ploxiln:
I do appreciate the net reduction of 800 lines of code, just in request json validation :smile:

ploxiln:
My peanut-gallery code-style tip is to consider using a flags int instead of a pair of booleans for positive, non-empty, required properties of parameters. So when defining the list of parameters it would look more like

```
 static struct janus_json_parameter listener_parameters[] = {
	{"feed", JSON_INTEGER, PARAM_POSITIVE | PARAM_REQUIRED},
	{"audio", JSON_TRUE, 0},
	{"video", JSON_TRUE, 0},
	{"data", JSON_TRUE, 0}
};
static struct janus_json_parameter feeds_parameters[] = {
	{"feeds", JSON_ARRAY, PARAM_NONEMPTY}
};
```

also, you could put the comment `/* Use JSON_TRUE instead of the non-existing JSON_BOOLEAN */` in a more central location

andreasg123:
Yes, using flags would allow to separate positive and non-empty and allow for more constraints later.  I could make that change in the next couple of days.

Regarding JSON_TRUE, it might be best to add another constant that could be used in the parameter descriptions:

    #define JANUS_JSON_BOOL JSON_TRUE

andreasg123:
I changed janus_json_parameter to use flags.

lminiero:
This has become even larger while I was away! I'll need some more time to review this... thanks, I'll get back to you as soon as I've gone through this.

andreasg123:
I just want to check if you found any issues with my PR or if you noticed areas where additional changes would be beneficial.

lminiero:
Hi Andreas, sorry, I'm still looking into it as it's quite a huge effort, and I've been a bit busy lately! I'll come back to you ASAP.

leonklingele:
+1 for this.

andreasg123:
I'm closing this PR because #516, #517, and #521 cover it.

PR #497, opened 2016-04-07
janus-pp-rec should always janus_log_destroy() at exit

### LINKED ISSUE #496: janus-pp-rec has no stdout
There is no stdout, when using janus-pp-rec:
1) Without options
2) With --header option

Janus from "master branch"
Cloned this time

ploxiln:
otherwise JANUS_LOG() messages may not be written out before the process exits

should fix #496


ploxiln:
no rush to merge, just looked like a rather easy fix so I pushed it up in case it helps someone

lminiero:
Just as I do find some time to work on #496, I find out you already fixed that! :smile:
I'll merge right away as it's exactly the same thing I wanted to do anyway.

Thanks a lot for the help!

PR #502, opened 2016-04-13
emacs.el to set the Janus coding style in Emacs

andreasg123:
To support other developers that try to follow your coding style while using Emacs, I added a function for ~/.emacs (https://www.emacswiki.org/emacs/IndentingC).


lminiero:
Thanks, merging!

PR #504, opened 2016-04-20
core: use RTLD_LOCAL when loading plugins and transports

saghul:
It's the default on Linux, but might not be that way on other systems
(hello OSX!), so make future portability work easier.


lminiero:
Makes sense, thanks! Merging then.

PR #507, opened 2016-04-24
fix janus build on mac os x, add openssl CFLAGS

mabu-github:
Include directory should be set in addition to library path for Darwin. Otherwise it won't compile e.g. on Mac OS X El Capitan because the openssl header files cannot be found. Make will produce the following output without the fix:

`/usr/local/include/libwebsockets.h:100:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
         ^
3 warnings and 1 error generated.
make[2]: *** [transports/transports_libjanus_websockets_la-janus_websockets.lo] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -17,6 +17,7 @@ LT_INIT
 
 case "$host_os" in
 darwin*)
+	CFLAGS="$CFLAGS -I/usr/local/opt/openssl/include"
 	LDFLAGS="$LDFLAGS -L/usr/local/lib -L/usr/local/opt/openssl/lib"
 	AM_CONDITIONAL([DARWIN_OS], true)
 	AM_CONDITIONAL([HAS_DTLS_WINDOW_SIZE], false)


lminiero:
Shouldn't this be ```CFLAGS="$CFLAGS -I/usr/local/include -I/usr/local/opt/openssl/include"``` to match the existing ```LDFLAGS="$LDFLAGS -L/usr/local/lib -L/usr/local/opt/openssl/lib"```? Just asking, never even used a Mac OS :wink:

mabu-github:
The code is just about *extending* possibly existing CFLAGS. If it were wrong, other includes should fail. But this is not the case. The following build went well:
`sh autogen.sh \
            && PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig/ \
               ./configure --disable-rabbitmq --prefix=/usr/local/janus \
            && make \
            && make install \
            && make configs`
I have already successfully tried the echo test.

mabu-github:
I have tried leaving out `/usr/local/lib/` in LDFLAGS too. When I check the binary with `otool -L` (ldd equivalent on mac os), it seems as if everything was linked correctly:
`mabu$ otool -L /usr/local/janus/bin/janus
/usr/local/janus/bin/janus:
	/usr/local/opt/libnice/lib/libnice.10.dylib (compatibility version 15.0.0, current version 15.1.0)
	/usr/local/opt/glib/lib/libgthread-2.0.0.dylib (compatibility version 4601.0.0, current version 4601.2.0)
	/usr/local/opt/glib/lib/libgio-2.0.0.dylib (compatibility version 4601.0.0, current version 4601.2.0)
	/usr/local/opt/glib/lib/libgobject-2.0.0.dylib (compatibility version 4601.0.0, current version 4601.2.0)
	/usr/local/opt/glib/lib/libglib-2.0.0.dylib (compatibility version 4601.0.0, current version 4601.2.0)
	/usr/local/opt/gettext/lib/libintl.8.dylib (compatibility version 10.0.0, current version 10.4.0)
	/usr/local/opt/jansson/lib/libjansson.4.dylib (compatibility version 12.0.0, current version 12.0.0)
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/sofia-sip/lib/libsofia-sip-ua.0.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
	/usr/local/opt/libusrsctp/lib/libusrsctp.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libcurl.4.dylib (compatibility version 7.0.0, current version 8.0.0)
	/usr/local/opt/srtp/lib/libsrtp.1.dylib (compatibility version 0.0.0, current version 0.0.0)`

Most of the things should be correctly autodetected by pkg-config during the process. The exception is in fact openssl because apple keeps around an openssl v0.9.8 for compatibility in `/usr/lib/libssl.0.9.8.dylib /usr/lib/libcrypto.0.9.8.dylib` which would be used in the build process. Notice that we need a higher version to build janus. To prevent wrong openssl configuration, it should IMHO be sufficent to install a recent version of openssl (1.0.2 as of the time of writing) via brew, fix the `PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig/` and to extend `CFLAGS` and `LDFLAGS` with `/usr/local/opt/openssl/...`.

lminiero:
Thanks for the explaination. As anticipated, I'm no expert at all of building on Mac OS, so any help on getting Janus to work easily on those systems is always welcome.

Do you know if this change causes problems on pre-El Capitan machines? I guess it shouldn't, since you simply added an alternative include directory, but I thought I'd ask anyway as apparently this worked on older systems so far (the Mac OS stuff you see was provided in older PR, so I guess it worked for the authors).

If I hear no objection in the next few days, I'll merge this, thanks!

mabu-github:
I cannot say anything about pre El-Capitan but generally adding an include directory never hurts if you're not overwriting some default location. 
But in this case it is explicitly wanted that a default location is overwritten because otherwise Apple's version of openssl is included.
In my opinion it should *not break* anything.

If Janus used OpenSSL 0.9.8 some time ago, this would be explain why the error has not occured because the header's in the systems default location would match.

lminiero:
> If Janus used OpenSSL 0.9.8 some time ago, this would be explain why the error has not occured because the header's in the systems default location would match.

No, I believe we always needed at least 1.0.1 in the past, as older versions didn't have DTLS-SRTP support IIRC. Now 1.0.2 is pretty much needed (or anyway anything older than 1.0.1f), as starting from Chrome 51/52 `ec_point_formats` and `elliptic_curves` will be required and not optional.

Merging then!

PR #510, opened 2016-04-30
Fix Janus.isWebrtcSupported

leonklingele:
Previous function did not work as expected.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -185,10 +185,7 @@ Janus.init = function(options) {
 
 // Helper method to check whether WebRTC is supported by this browser
 Janus.isWebrtcSupported = function() {
-	if(RTCPeerConnection === null || getUserMedia === null) {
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks! You're right, I meant to change that (uncompliant browsers would have those undefined, not null) and always forgot to do that.

Before I can merge this, can you sign our [CLA](https://cla.conf.meetecho.com/janus/)?

leonklingele:
Done. :)

lminiero:
> Done. :)

Can you try again? I still don't see it, might be a hiccup in our server :open_mouth: 
You need to authenticate via github, and then compile the data at the end of the page.

Thanks!

leonklingele:
I see, sorry. Missed to fill out the form at the confirmation page.

lminiero:
Thanks, merging!

PR #514, opened 2016-05-01
janus.js: Try to share screen when share windows fails

### LINKED ISSUE #513: Error when share screen with Firefox
When I use `janus.plugin.videoroom` with janus.js to share screen with Firefox, I receive an error:
```javascript
MediaStreamError {
  name: "NotFoundError",
  message: "The object can not be found here.",
  constraint: "",
  stack: ""
}
```

After some investigation I found that my Firefox (45.0.1) don't support `mediaSource: 'window'`. When I execute `navigator.mediaDevices.getUserMedia({video: {mediaSource: 'window'}})` I get the same error, but `navigator.mediaDevices.getUserMedia({video: {mediaSource: 'screen'}})` works perfectly.

I have no idea where the error comes from. My current setup is:
- **OS**: Arch Linux
- **Browser**: Firefox 45.0.1
- **Window Manager**: bspwm (Tiling Window Manager) _<- I suspect this is the cause_

Wherever the error comes, janus.js should be able to deal with this error, trying share `screen` when share `window` fails.

magarcia:
For Firefox when try to share window and this fails with a
`MediaStreamError NotFound`, try to share screen.

Closes #513

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1401,7 +1401,16 @@ function Janus(gatewayCallbacks) {
 						Janus.debug(constraints);
 						navigator.mediaDevices.getUserMedia(constraints)
 							.then(function(stream) { gsmCallback(null, stream); })
-							.catch(function(error) { pluginHandle.consentDialog(false); gsmCallback(error); });
+							.catch(function(error) {
+								if (error.name === 'NotFoundError' && constraints.video.mozMediaSource === 'window' && constraints.video.mediaSource === 'window') {
+									constraints.video.mozMediaSource = 'screen'
+									constraints.video.mediaSource = 'screen'
+									getScreenMedia(constraints, gsmCallback)
+								} else {
+									pluginHandle.consentDialog(false);
+									gsmCallback(error);
+								}
+							});
 					};
 					if(window.navigator.userAgent.match('Chrome')) {
 						var chromever = parseInt(window.navigator.userAgent.match(/Chrome\/(.*) /)[1], 10);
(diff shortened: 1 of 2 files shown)


lminiero:
What if `constraints.video` is undefined? That would cause an exception.

saghul:
What if the user has some window open which doesn't want the remote person to see?

lminiero:
> What if the user has some window open which doesn't want the remote person to see?

Good point... I'd say that we should honor what the user asked for. If that failed, the user (or the application) can then decide whether they want to try sharing the screen in general or not.

magarcia:
I'm not completely agree with that. Currently the "share screen" action has different behaviours for each browser (Chrome/Chromium shares screen but Firefox share Windows). I think is better create two actions "share screen" and "share window", I see most natural this behaviour.

However, a change like this can cause issues to users who are using "share screen" as "share window".

lminiero:
I wasn't aware that Firefox allowed for different ways of capturing things, just seen this on this demo here: https://mozilla.github.io/webrtc-landing/gum_test.html

I think the right way to handle this is extending `janus.js` so that we allow for `video: "window"` besides the `video: "screen"` we have right now. This will be a synonim for Chrome, while it will result in a different constraint for Firefox instead. With this approach, the fallback I mentioned should work: if "screen" fails for you, just do a new request with "window" instead in your application, without any automated fallback that could cause the issue Saùl mentioned.

I'll make the change shortly and post it as a new PR you can check, just to make sure it would fix it for you.

lminiero:
Just done, feedback welcome!

lminiero:
Closing as I just merged #518.

PR #515, opened 2016-05-05
fixbug postprocessing for opus using DTX

jing3018:
where opus using DTX and the client mute the audio, the seq is continuous，but client sent less packets with rate (1 packet per second) 。in this condition， need use rtp timestamp to insert slience frame.

### DIFF
--- a/postprocessing/pp-opus.c
+++ b/postprocessing/pp-opus.c
@@ -72,16 +72,18 @@ int janus_pp_opus_process(FILE *file, janus_pp_frame_packet *list, int *working)
 	uint64_t pos = 0;
 	uint8_t *buffer = g_malloc0(1500);
 	while(*working && tmp != NULL) {
-		if(tmp->prev != NULL && (tmp->seq - tmp->prev->seq > 1)) {
+		if(tmp->prev != NULL && ((tmp->ts - tmp->prev->ts)/48/20 > 1)) {
 			JANUS_LOG(LOG_WARN, "Lost a packet here? (got seq %"SCNu16" after %"SCNu16", time ~%"SCNu64"s)\n",
 				tmp->seq, tmp->prev->seq, (tmp->ts-list->ts)/48000);
 			/* FIXME Write the silence packet N times to fill in the gaps */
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for this. I seem to recall we had a similar issue with a recording ourselves, I'll check if we still have it and see if that fixes it, and in case merge.

I only have one doubt: your patch fixes the `pos` part, but it still only inserts a single silence frame, which is 960 samples, so 20ms. Does this still result in a correctly playable file for you?

jing3018:
@lminiero , it's not inserts a single silence frame, will insert `silence_count` silence frames. the pos calculate is not correct in this pull request, I will fix it now , my record file audio and video is sync now.

lminiero:
Thanks for the updates! I'm currently abroad for a conference, but I'll review this in more detail as soon as I can, probably when I come back.

lminiero:
Just found out some time to look into this and it looks fine to me.

The only thing I noticed is that your code seems to always assume 20ms for audio packets when computing the `pos`, while we simply relied on the sequence number for that, no matter the duration of the packet. For silence this should be fine (our "silence" packet is 20ms anyway), but there may be cases where some clients are using packets with different lenghts, which are supported by the standard.

I'm not completely familiar with what the `pos` value means in the ogg specification, though. Is it just some sort of internal sequence number, unrelated to the actual length and temporal "positioning" of the packet?

jing3018:
yeah, I always assume 20ms for audio packets。I'm also not familiar with ogg。It's not a perfect solution。My application always use 20ms PT, It's works for me。BTW, the code in master also assume the PT = 20ms。

lminiero:
> BTW, the code in master also assume the PT = 20ms

Well, not really: only the silence managament assumed 20ms, because the sample we have is of that length. The rest of the code never looked at timestamps, and only relied on sequence numbers to increase the `pos`. We saved whatever packet was there, which may have included more ms of data. Your code does the same, but it's the `pos` management that may "break" in case packets of 10ms, 40ms or 60ms appear: in that case, you may end up with either multiple packets with the same `pos`, or with gaps in the `pos` itself.

Anyway, your patch does fix an error, and we can always worry about a proper management of audio packets of different duration later on, should people encounter issues.

So merging for now, thanks!

PR #516, opened 2016-05-05
videoroom: remove duplicate janus_mutex_unlock of rooms_mutex

ploxiln:
just in case it helps ... split-out from #494 (and rebased because it helps with the next part I'm going to try to split-out and rebase/squash)

(@andreasg123 still gets credit for the commit)


lminiero:
It does help, thanks! Anyway, looking at the patch again, rather than removing those extra unlocks, I'd move the first one from line 1636 to 1674 (before the other lock), so that we're sure the room is still there while we check things.

ploxiln:
Changed as you suggested, and I found another identical case close by.

lminiero:
Looks cool, merging, thanks!

PR #517, opened 2016-05-05
json validation helpers

ploxiln:
the core part of #494 split out, again just in case it helps, along with two tiny tweaks

(but following the commit in #516 - if that is merged, and this branch is updated in some way to get github to refresh the commit list, "Removed duplicate calls to janus_mutex_unlock" will no longer show up in this PR)

If it is decided to go with this broken-out version, I'll prepare another PR with the rest of the changes for the rest of the plugins.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -79,6 +79,22 @@ gchar *janus_get_server_key(void) {
 /* API secrets */
 static char *api_secret = NULL, *admin_api_secret = NULL;
 
+/* JSON parameters */
+static int janus_process_error_string(janus_request *request, uint64_t session_id, const char *transaction, gint error, gchar *error_string);
+
+static struct janus_json_parameter incoming_request_parameters[] = {
+	{"transaction", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
+	{"janus", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
+	{"id", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE}
+};
+static struct janus_json_parameter attach_parameters[] = {
+	{"plugin", JSON_STRING, JANUS_JSON_PARAM_REQUIRED}
+};
+static struct janus_json_parameter jsep_parameters[] = {
+	{"type", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
+	{"trickle", JANUS_JSON_BOOL, 0},
+	{"sdp", JSON_STRING, JANUS_JSON_PARAM_REQUIRED}
+};
 
 /* Admin/Monitor helpers */
 json_t *janus_admin_stream_summary(janus_ice_stream *stream);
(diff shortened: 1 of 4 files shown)


lminiero:
This is really helpful, thanks.

Apologies to both you and @andreasg123 for not having provided sufficient feedback so far, but the effort is huge and I still didn't have much time to look into it, and I want to make sure I don't half-look through it. In particular, I want to make sure I understand how memory is handled there, so that I can subsequently merge this easily/seamlessly in the reference counters branch as well.

This might take a few more days (I have two different events to attend abroad the next two weeks), but I promise I'll get there! :smile:

saghul:
@ploxiln great work! this is really useful.

ploxiln:
rebased on updated #516

andreasg123:
The only memory allocated by the validation function is 20 bytes on the stack for the type name.  I suppose that number could be defined as a constant that would be used as the value of `req_size` in `janus_get_json_type_name`. Otherwise, the macros `JANUS_VALIDATE_JSON_OBJECT` and `JANUS_VALIDATE_JSON_OBJECT_FORMAT` require that `error_cause` is allocated on the stack so that `sizeof` works.  Parameters also cannot be allocated on the heap for the same reason.

@ploxiln thanks for fixing the misspelled "secret". What's wrong with [json_string_length](http://jansson.readthedocs.io/en/2.7/apiref.html#string) that you replaced with `json_string_value`?

ploxiln:
`json_string_length()` was introduced in jansson-2.7, and when building janus so far I've used the system package manager to install jansson, and I still use ubuntu-14.04 sometimes, which includes jansson-2.5. http://jansson.readthedocs.io/en/stable/changes.html

I can upgrade if the consensus is to use json_string_length(), not that big a deal, but I figured it was still pretty easy and minimal to stay compatible with jansson-2.5 for now.

ploxiln:
did a simple little rebase to make the videorooms mutex fix (merged to master) disappear from this PR

lminiero:
Sorry for the late feedback (again!), I looked through this and it seems great.

Just a quick question before I merge, though (as I want to integrate its usage in the other plugins anyway, so that I can sync them ASAP): why using `JANUS_VALIDATE_JSON_OBJECT` almost anywhere, while the lower level `JANUS_VALIDATE_JSON_OBJECT_FORMAT` for parsing JSEP stuff? Was it just to showcase the two possible different approaches in doing things, the latter giving more control on the returned text?

andreasg123:
`JANUS_VALIDATE_JSON_OBJECT` produces _standard_ error message while still returning the error code appropriate for the plugin.  `JANUS_VALIDATE_JSON_OBJECT_FORMAT` is intended for the situations where a non-standard error message is needed such as the JSEP error messages.  I haven't looked closely enough to tell if that's the only place where different error messages are used.

Note that this PR created by @ploxiln does not include my changes for the other plugins.  For that, my original PR is still needed.

lminiero:
@andreasg123 thanks for the clarification!

I'll merge this PR then. If you already integrated this in some other plugins, could you then prepare a new PR that takes care of that based on what will become the new merged master? I'll take care of whatever's missing after that.

Thanks!

ploxiln:
I also did promise to prepare a PR for the rest of the plugins after this was merged :) it'll be a day before I get to it though

lminiero:
No problem, there's no hurry, thanks!

PR #518, opened 2016-05-05
Differentiate screen and window sharing in Firefox

lminiero:
This PR tries to address the issue raised in #514, specifically with respect to the way Firefox handles screensharing, which is different from Chrome. In Firefox, you have to specify in advance whether you want to share the whole screen or a single window, while Chrome puts them all together in the picker later on.

As such, now `janus.js` allows for both `video: "screen"` and `video: "window"`. Both will end up in the same place for Chrome, while for Firefox this will drive the choice accordingly. The screen sharing demo has been updated as well, in order to show a dialog ("Screen or Window?") when you open it from Firefox.

This seems to be working as expected for me, so I plan to merge shortly. If you hate this or would like other improvements first, feel free to shout!

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1297,7 +1297,7 @@ function Janus(gatewayCallbacks) {
 			}
 			var videoSupport = isVideoSendEnabled(media);
 			if(videoSupport === true && media != undefined && media != null) {
-				if(media.video && media.video != 'screen') {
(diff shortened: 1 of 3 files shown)


magarcia:
Nice solution, looks cleaner than #514 and should fix #513

lminiero:

PR #521, opened 2016-05-13
json validation helpers - use for all plugins

ploxiln:
also add `JANUS_CHECK_SECRET()` and `JANUS_CHECK_PIN()` helpers

this is a follow-up of #517 (the rest of #494)

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -441,6 +441,39 @@ janus_plugin *create(void) {
 	return &janus_audiobridge_plugin;
 }
 
+/* Parameter validation */
+static struct janus_json_parameter request_parameters[] = {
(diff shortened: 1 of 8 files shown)


lminiero:
Wow, that was much faster than a day! :smile:  I'll check this tomorrow in the morning with fresh eyes and in case everything's fine I'll merge right away.
Thanks!

lminiero:
Looks fine to me. Just as a side note, since `JANUS_CHECK_SECRET` and `JANUS_CHECK_PIN` are basically the same, apart from the name of the element to check, why not make it a generically appliable method that has the element name to verify as part of its signature?

ploxiln:
yeah looks like those could be combined, I'll add a commit for that

lminiero:
Thanks!

ploxiln:
updated

lminiero:
Thanks! I'll do a quick further revision check and then merge.

PR #523, opened 2016-05-14
Reduce code duplication in videoroom plugin with several new functions.

andreasg123:
As promised, now that the request verification is merged, I made a first attempt to address code duplication in the videoroom plugin.  I used the [Simian Similarity Analyzer](http://www.harukizaemon.com/simian/) to locate promising areas to reduce code duplication.

I made the following changes for a net reduction of 270 lines:
- Combined SDP templates for VP8 and VP9.
- `janus_videoroom_notify_participants` sends a message to all other participants in a room.
- `janus_videoroom_leave_or_unpublish` notifies all other participants that somebody left or unpublished.
- `janus_videoroom_access_room` gets a room by ID and checks secret or pin if requested.
- `janus_videoroom_recorder_close` closes recorders for a participant.
- `janus_videoroom_sdp_v_format_vp8_vp9` formats the SDP for VP8 and VP9.
- `janus_videoroom_sdp_v_format_h264` formats the SDP for H.264.
- In `janus_videoroom_handler`, use a loop to create the two similar SDP variants.

In two of those functions, I wrote a comment that a mutex should be locked.  I didn't lock the mutex inside the function because in some cases the mutex is supposed to stay locked past the function call.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -986,6 +986,42 @@ void janus_videoroom_create_session(janus_plugin_session *handle, int *error) {
 	return;
 }
 
+static void janus_videoroom_notify_participants(janus_videoroom_participant *participant, char *msg) {
+	/* participant->room->participants_mutex has to be locked. */
+	GHashTableIter iter;
+	gpointer value;
+	g_hash_table_iter_init(&iter, participant->room->participants);
+	while (!participant->room->destroyed && g_hash_table_iter_next(&iter, NULL, &value)) {
+		janus_videoroom_participant *p = value;
+		if(p && p->session && p != participant) {
+			JANUS_LOG(LOG_VERB, "Notifying participant %"SCNu64" (%s)\n", p->user_id, p->display ? p->display : "??");
+			int ret = gateway->push_event(p->session->handle, &janus_videoroom_plugin, NULL, msg, NULL, NULL);
+			JANUS_LOG(LOG_VERB, "  >> %d (%s)\n", ret, janus_get_api_error(ret));
+		}
+	}
+}
+
+static void janus_videoroom_leave_or_unpublish(janus_videoroom_participant *participant, gboolean is_leaving) {
+	json_t *event = json_object();
+	json_object_set_new(event, "videoroom", json_string("event"));
+	json_object_set_new(event, "room", json_integer(participant->room->room_id));
+	json_object_set_new(event, is_leaving ? "leaving" : "unpublished", json_integer(participant->user_id));
+	char *leaving_text = json_dumps(event, JSON_INDENT(3) | JSON_PRESERVE_ORDER);
+	json_decref(event);
+	/* we need to check if the room still exists, may have been destroyed already */
+	if(participant->room) {
+		if(!participant->room->destroyed) {
+			janus_mutex_lock(&participant->room->participants_mutex);
+			janus_videoroom_notify_participants(participant, leaving_text);
+			if (is_leaving) {
+				g_hash_table_remove(participant->room->participants, GUINT_TO_POINTER(participant->user_id));
+			}
+			janus_mutex_unlock(&participant->room->participants_mutex);
+		}
+	}
+	g_free(leaving_text);
+}
+
 void janus_videoroom_destroy_session(janus_plugin_session *handle, int *error) {
 	if(g_atomic_int_get(&stopping) || !g_atomic_int_get(&initialized)) {
 		*error = -1;
@@ -1021,33 +1057,7 @@ void janus_videoroom_destroy_session(janus_plugin_session *handle, int *error) {
 			if(participant->recording_base)
 				g_free(participant->recording_base);
 			participant->recording_base = NULL;
-			json_t *event = json_object();
-			json_object_set_new(event, "videoroom", json_string("event"));
-			json_object_set_new(event, "room", json_integer(participant->room->room_id));
-			json_object_set_new(event, "leaving", json_integer(participant->user_id));
-			char *leaving_text = json_dumps(event, JSON_INDENT(3) | JSON_PRESERVE_ORDER);
-			json_decref(event);
-			GHashTableIter iter;
-			gpointer value;
-			/* we need to check if the room still exists, may have been destroyed already */
-			if(participant->room) {
-				if(!participant->room->destroyed) {
-					janus_mutex_lock(&participant->room->participants_mutex);
-					g_hash_table_iter_init(&iter, participant->room->participants);
-					while (!participant->room->destroyed && g_hash_table_iter_next(&iter, NULL, &value)) {
-						janus_videoroom_participant *p = value;
-						if(p == participant) {
-							continue;	/* Skip the leaving publisher itself */
-						}
-						JANUS_LOG(LOG_VERB, "Notifying participant %"SCNu64" (%s)\n", p->user_id, p->display ? p->display : "??");
-						int ret = gateway->push_event(p->session->handle, &janus_videoroom_plugin, NULL, leaving_text, NULL, NULL);
-						JANUS_LOG(LOG_VERB, "  >> %d (%s)\n", ret, janus_get_api_error(ret));
-					}
-					g_hash_table_remove(participant->room->participants, GUINT_TO_POINTER(participant->user_id));
-					janus_mutex_unlock(&participant->room->participants_mutex);
-				}
-			}
-			g_free(leaving_text);
+			janus_videoroom_leave_or_unpublish(participant, TRUE);
 		} else if(session->participant_type == janus_videoroom_p_type_subscriber) {
 			/* Detaching this listener from its publisher is already done by hangup_media */
 		} else if(session->participant_type == janus_videoroom_p_type_subscriber_muxed) {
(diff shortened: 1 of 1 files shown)


lminiero:
I appreciate the code reduction, but IMHO in some places it is not worth the effort as we loose flexibility. The SDP templates are an example of that: I'd rather keep them separated, as tomorrow Chrome or Firefox or another browser may require something different for, let's say, VP9, and I'd have to split the templates again. Anything that is SDP related I'd like to keep as much control on, and avoid worrying about loops, steps etc...

andreasg123:
I restored the separate SDP templates for VP8 and VP9.  With respect to the loop, please have another look.  The second loop iteration produces almost the same result, just replacing "recvonly" with "sendonly" and using the static values for payload.

There are a few additional differences that didn't make sense to me.  For Opus and VP8, you completely remove an unsupported audio or video codec in the second pass, for others you send the same as in the first pass.  For some unsupported codecs, you send payload 0, for others you still send the static payload for that codec.  Such differences become more apparent with the duplicate code removed.

lminiero:
Thanks for the change. Yep, of course I'll have another look :smile: The considerations I wrote yesterday were more general, and only deriving from a quick glance I had given to the code. I plan to review the code in more detail in the next few days.

Not sure as to what you're referring to with respect to codecs we remove or keep. The plugin allows you to choose which codec to force on all users, publishers and viewers alike, to make sure everything's consistent. This means filtering out stuff the publisher may have sent, and preparing an offer that we'll send all viewers with only what we need/want. The only thing we remove after that is the bitrate attribute if it's zero. Anyway, I'll check your updated code to see if there were any assumptions we made that were lost there, or if there were indeed bugs that were fixed here.

andreasg123:
I believe that my version produces the same output as before but I'm confused about some of that output.  In particular, I'm talking about codecs set for the room that are not offered by the publisher. For OPUS and VP8, the SDP media line sent back to the publisher and stored for the listeners is different.

| Codec | Publisher | Listener |
|---|---|---|
| OPUS | "m=audio 111 RTP/SAVPF 0\r\n" | "" |
| ISAC 32K | "m=audio 104 RTP/SAVPF 0\r\n" | same |
| ISAC 16K | "m=audio 103 RTP/SAVPF 0\r\n" | same |
| PCMU | "m=audio 0 RTP/SAVPF 0\r\n" | same |
| PCMA | "m=audio 0 RTP/SAVPF 0\r\n" | same |
| VP8 | "m=video 0 RTP/SAVPF 0\r\n" | "" |
| VP9 | "m=video 0 RTP/SAVPF 0\r\n" | same |
| H264| "m=video 0 RTP/SAVPF 0\r\n" | same |

I would think that is should be "m=audio 0 RTP/SAVPF 0\r\n" instead of using the payload type as the port number.  If that is correct, I could simplify the code.  Also, I don't know why the listener would not receive a video or audio media line if the room is set to OPUS or VP8, respectively, and the publisher does not offer that.

andreasg123:
I rebased my PR to use @ploxiln's `JANUS_CHECK_SECRET`.

andreasg123:
I tried to determine what the correct values for the media lines are that I put into the table 1-1/2 weeks ago.  [RFC 3264](https://www.ietf.org/rfc/rfc3264.txt) seems to provide the relevant information.

> A port number of zero in the offer indicates that the stream is offered but MUST NOT be used.  This has no useful semantics in an initial offer, but is allowed for reasons of completeness, since the answer can contain a zero port indicating a rejected stream (Section 6).  Furthermore, existing streams can be terminated by setting the port to zero (Section 8).  In general, a port number of zero indicates that the media stream is not wanted.
> 
> For each "m=" line in the offer, there MUST be a corresponding "m=" line in the answer.  The answer MUST contain exactly the same number of "m=" lines as the offer.  This allows for streams to be matched up based on their order.  This implies that if the offer contained zero "m=" lines, the answer MUST contain zero "m=" lines.
> 
> If the answerer has no media formats in common for a particular offered stream, the answerer MUST reject that media stream by setting the port to zero.

That would indicate that "m=audio 111 ..." above is a typo in the code.  For the answer, the media line has to remain and the port has to be zero.  For sending the offer to the listener, one could either drop the media line or set the port to zero.

I think that this should be handled consistently for all codecs.  Was there a reason why you treated OPUS and VP8 differently?  If it can be handled the same for all codecs, I can update the PR to do that more concisely.

lminiero:
I think it doesn't matter, as the actual SDP is then updated by the core itself, when it merges the template from the plugin with the WebRTC stuff it needs to get it working, and I believe we set all of that to 0 in sdp.c anyway no matter what the plugin says.

andreasg123:
I performed a rebase to include your recent changes.  I also made the mline code more consistent so that it returns the same for all rejected audio or video.  Finally, I added the function `janus_videoroom_recorder_create` to combine the two places where recorders are created.

With all the changes you are making, it is difficult to keep this current.  I'm willing to do another rebase for the pending plugin optimization.  At that time, you should decide if you want it.

lminiero:
Yep, sorry for all the changes, but as I anticipated there were a lot of things I was working on at the same time, which is why I was hesitant to integrate the code refactoring until then, as I would have had to work around them. As soon as I merge most of them (which should be soon) I'll definitely integrate your changes too!

plugins/janus_videoroom.c
```diff
@@ -2321,6 +2330,43 @@ void janus_videoroom_hangup_media(janus_plugin_session *handle) {
 	}
 }
 
+static void janus_videoroom_sdp_v_format_vp8(char *mline, int mline_size, int pt, int b, const char *video_mode) {
```
lminiero: Since all these methods have the same parameters, wouldn't it be simpler to have a `janus_videoroom_sdp_v_format` method that also has the codec as one of the arguments, and then do a switch in there to apply the correct format? We already have enums for the purpose: https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_videoroom.c#L346

lminiero: VP8 and VP9 may even belong to the same case, as they have the same number of arguments to the format string, and they only differ in `sdp_v_template_vp8` and `sdp_v_template_vp9`.

plugins/janus_videoroom.c
```diff
@@ -1127,6 +1137,40 @@ char *janus_videoroom_query_session(janus_plugin_session *handle) {
...
+	/* rooms_mutex has to be locked */
+	int error_code = 0;
+	char error_cause[512];
+	json_t *room = json_object_get(root, "room");
+	guint64 room_id = json_integer_value(room);
+	*videoroom = g_hash_table_lookup(rooms, GUINT_TO_POINTER(room_id));
+	if(*videoroom == NULL) {
+		JANUS_LOG(LOG_ERR, "No such room (%"SCNu64")\n", room_id);
+		error_code = JANUS_VIDEOROOM_ERROR_NO_SUCH_ROOM;
+		g_snprintf(error_cause, 512, "No such room (%"SCNu64")", room_id);
```
lminiero: What's the point of printing the error, if you only return the code?

plugins/janus_videoroom.c
```diff
@@ -3467,355 +3417,168 @@ static void *janus_videoroom_handler(void *data) {
...
-								JANUS_LOG(LOG_ERR, "Couldn't open an audio recording file for this publisher!\n");
-							}
+						switch(videoroom->acodec) {
+							case JANUS_VIDEOROOM_OPUS:
+								if(opus_pt < 0) {
+									if(pass == 1)
+										JANUS_LOG(LOG_WARN, "Videoroom is forcing OPUS, but publisher didn't offer any... rejecting audio\n");
+								} else {
+									int pt = (pass == 1 ? opus_pt : OPUS_PT);
+									g_snprintf(audio_mline, 256, sdp_a_template_opus,
```
lminiero: Why not make a function for audio formats as well, as you did for video? Even though the number of arguments there is always the same, it would provide a nice interface in case any of them requires more or less in the future. I'd add the audio codec enum as one of the arguments there as well.

plugins/janus_videoroom.c
```diff
@@ -2208,6 +2189,68 @@ void janus_videoroom_slow_link(janus_plugin_session *handle, int uplink, int vid
 	}
 }
 
+static void janus_videoroom_recorder_create(janus_videoroom_participant *participant, gboolean audio, gboolean video) {
```
lminiero: I guess that both `recorder_create` and `recorder_close` could lock the `rec_mutex` themselves. Or would you rather leave this to the caller?

plugins/janus_videoroom.c
ploxiln: A way to be able to use the same `*printf()` call, despite there being a different number of substitutions, is to specify the argument position in the format string. For every instance of `pt` after the first, the format string could have `%1$d` which means "take the first value argument as type d" (instead of "take the next argument"), and you would not need to provide the subsequent `pt` parameters to `g_snprintf()`

lminiero: Ah, cool, I didn't know about this. Is it available on all compilers, though, or is this a feature that would require something quite new?

ploxiln: That particular feature is pretty old and available everywhere c89 is.
(you're right that some format replacement types are gcc-specific or relatively new)

plugins/janus_videoroom.c
andreasg123: Both are called in a [longer critical section](https://github.com/andreasg123/janus-gateway/blob/7b7ab2e0fd3e7415dfd0372018ac61ad5be0195e/plugins/janus_videoroom.c#L2843) protected by `rec_mutex`. Currently, `rec_mutex` also protects `recording_active` and `recording_base` that are changed outside those functions.  Do mutexes properly deal with being locked a second time by the same thread (and then unlocked twice)?

plugins/janus_videoroom.c
andreasg123: My mistake, I need the contents of `error_cause` outside that function after an error.  I'll check in a new version with the additional parameters `error_cause` and `error_cause_size`.

plugins/janus_videoroom.c
ploxiln: So called "recursive" or "re-entrant" mutexes can deal with that, but the default kind of mutexes cannot.
(You can initialize a pthread mutex with an attr `PTHREAD_MUTEX_RECURSIVE`, but I'd recommend not starting to use recursive mutexes in this project).

plugins/janus_videoroom.c
andreasg123: I didn't know about that feature, either.  According to [this page](http://linux.die.net/man/3/printf), it is **not** part of the C99 standard but comes from the [Single Unix Specification](https://en.wikipedia.org/wiki/Single_UNIX_Specification).

I'm fine to switch all SDP templates to it because it would make using them simpler as long as there is no concern about portability.  As a caveat, all parameters have to use the positional notation and all parameters have to be referred to at least once.

ploxiln: I see you're right, it does warn for mixed, I haven't used it in a while (though it does work ;)

```
$ gcc -Wall -o test test.c
test.c:4:35: warning: cannot mix positional and non-positional arguments in format string [-Wformat]
        printf("count=%d cmd=%s again=%1$d reprise=%1$d\n", argc, argv[0]);
                                      ~~~^
1 warning generated.
$ ./test two three four
count=4 cmd=./test again=4 reprise=4
```

Anyway I think all positional in these format strings isn't too ugly.

andreasg123:
I checked in a new commit with the new functions `janus_videoroom_sdp_a_format` and `janus_videoroom_sdp_v_format`.  They don't use the positional arguments yet.  The switch statements could just assign the appropriate template to a variable once the number of arguments is the same but `gcc` will warn if the `printf` pattern isn't a constant.  I also added a comment showing how the [log statements could be combined](https://github.com/andreasg123/janus-gateway/blob/a046cf754eb296c06b38cd19bc4616e967e2a2ae/plugins/janus_videoroom.c#L3528).

This version isn't fully tested.  If you like this approach, I can make sure that it doesn't cause any problems.

andreasg123:
I combined the log messages as I suggested yesterday.  I also verified that the VideoRoom plugin still runs with these changes.

I'll leave the use of positional arguments in the SDP templates for a separate PR.  Now, the audio and video templates are only used in a single place so that the change should be easy.

lminiero:
Yep, I agree positional arguments deserve a separate conversation, especially if portability or compatibility issues arise. Let me give another look and test to the patch and I'll merge, thanks!

lminiero:
Looks good, thanks!

PR #526, opened 2016-05-17
Reduce code duplication in janus.c with new functions.

andreasg123:
I worked on code duplication in janus.c by validating more requests with the new validation frameworks.  I also introduced four new functions:
- `janus_request_check_secret` checks the "apisecrect" and "token" parameters.
- `janus_request_ice_pending_trickles` processes the pending ICE trickle candidates.
- `janus_json_list_token_plugins` lists the plugins accessible with a token as a JSON object.
- `janus_request_allow_token` handles the "allow_token" and "disallow_token" requests.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -93,10 +93,6 @@ static struct janus_json_parameter jsep_parameters[] = {
 	{"trickle", JANUS_JSON_BOOL, 0},
 	{"sdp", JSON_STRING, JANUS_JSON_PARAM_REQUIRED}
 };
-static struct janus_json_parameter allow_token_parameters[] = {
-	{"token", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
-	{"plugins", JSON_ARRAY, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_NONEMPTY}
-};
 static struct janus_json_parameter add_token_parameters[] = {
 	{"token", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"plugins", JSON_ARRAY, 0}
@@ -1462,7 +1458,7 @@ static json_t *janus_json_list_token_plugins(const char *token_value, const gcha
 	return reply;
 }
 
-static int janus_request_allow_token(janus_request *request, guint64 session_id, const gchar *transaction_text, gboolean allow) {
+static int janus_request_allow_token(janus_request *request, guint64 session_id, const gchar *transaction_text, gboolean allow, gboolean add) {
 	/* Allow/disallow a valid token valid to access a plugin */
 	int ret = -1;
 	int error_code = 0;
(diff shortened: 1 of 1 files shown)


andreasg123:
There was more duplicate code around the call to one of my new functions.  Here is the replacement function:

- `janus_request_ice_handle_answer` handles the ICE answer.

lminiero:
Thanks as usual for this tremendous effort!

Anyway, I'd like to keep this under ice for a little longer, if possible. In fact, I'm currently working on a new set of modules (Events API) which of course involves the core as it is as well. Besides, I still have to incorporate all the changes you and Pierce made in the reference counters branch, which is already a titanic task, and having to worry about this as well right now worries me (I'd like to be able to merge THAT sooner or later! :smile: ).

I'll still check this as soon as I get the chance, of course, if not just to give you some feedback in case I find things that may be safer or easier to keep the way they are.

andreasg123:
I'll rebase the PR from time to time to keep it current with the master.

It's unfortunate that git doesn't provide better diffs that would show that my changes are actually pretty minor.  I created a [diff file](https://github.com/meetecho/janus-gateway/files/278026/diff.txt) that shows the differences between my new functions and the old code after renaming some variables.

andreasg123:
I did a rebase now that the validation part is merged.  The diff files that I previously produced are now slightly out of date but I verified that you did not make any changes in the affected sections.  I can produce the diff files again if needed.

janus.c
```diff
@@ -2658,215 +2556,16 @@ json_t *janus_plugin_handle_sdp(janus_plugin_session *plugin_session, janus_plug
...
-					nice_address_set_port(&c->addr, janus_ice_get_rtcpmux_blackhole_port());
-					c->username = g_strdup(ice_handle->video_stream->ruser);
-					c->password = g_strdup(ice_handle->video_stream->rpass);
-					if(!nice_agent_set_selected_remote_candidate(ice_handle->agent, ice_handle->video_stream->stream_id, 2, c)) {
-						JANUS_LOG(LOG_ERR, "[%"SCNu64"] Error forcing dummy candidate on RTCP component of stream %d\n", ice_handle->handle_id, ice_handle->video_stream->stream_id);
-						nice_candidate_free(c);
-					}
-				}
-			}
 			janus_mutex_lock(&ice_handle->mutex);
```
lminiero: Have you checked if putting everything (including the above part) within this lock doesn't cause any deadlock? Not sure why it was in the middle like that, but it may have been to avoid deadlock issues where the handle mutex was locked again within the same thread (there are several ICE related methods there).

andreasg123: The only functions called from `janus_request_ice_handle_answer` with the `handle` as a parameter are `janus_ice_trickle_parse` and `janus_ice_setup_remote_candidates`.  Neither of those accesses `handle->mutex`.  Also, those calls were previously in the locked section.

The other place that contained the code moved to `janus_request_ice_handle_answer` [locked a much larger section](https://github.com/andreasg123/janus-gateway/blob/janus-duplicate-code/janus.c#L1112).  I don't see any problems with extending the locked sections.

andreasg123:
Have you had any more time to review this PR?  Did the diff files that I produced help?

lminiero:
@andreasg123 sorry, not yet. As I anticipated in one of the first comments, this particular PR I'd like to keep under ICE for a while, since I have three major PRs waiting that may all require considerable changes should we merge this right away.

I'll definitely let you know as soon as I'm ready to integrate this effort: no need for you to rebase on a regular basis, don't worry! :smile:

andreasg123:
I performed a rebase to include the changes from #544 (no conflicts).  I also noticed another small duplicated section.

andreasg123:
@lminiero, would it help if I retargeted this PR for the refcount branch or do you want to wait until both the refcount and  event-handlers branches are merged?

lminiero:
I think using refcount as a base would help, as we have a lot of fixes and optimizations there, thanks! Unless it would be too much of a burden to adapt your code reduction to a different code base?

andreasg123:
I rebased for refcount.  That mostly involved changing `janus_ice_component_free` to `janus_ice_component_destroy`.  There was also the following code fragment that I had to add to the new function `janus_request_ice_remove_rtcp_component`.

	if(!video) {
		handle->audio_stream->video_ssrc = 0;
		handle->audio_stream->video_ssrc_peer = 0;
		g_free(handle->audio_stream->video_rtcp_ctx);
		handle->audio_stream->video_rtcp_ctx = NULL;
	}

lminiero:
Thanks! I'll be abroad next week for a conference, so I'll only be able to look into this when I come back.

lminiero:
Just found some time to review this, and it looks fine to me. I'll give this code a go in some scenarios here to make sure nothing breaks, and in case it's all fine I'll merge, thanks!

PR #528, opened 2016-05-19
ICE video packets fix

andreasg123:
I believe this is a typo in ice.c.


lminiero:
Yep, good catch, thanks!

PR #531, opened 2016-05-20
Avoid race condition when closing recorders in videoroom

### LINKED ISSUE #530: Race condition when ending recording and closing stream in VideoRoom plugin
There seems to be a race condition when the web browser sends a message to end recording for a publisher in a video room and immediately after closes the WebRTC stream.  It crashes the server in `janus_recorder_free` (see [stack trace](https://github.com/meetecho/janus-gateway/files/275191/stack-trace.txt)).  The [log file](https://github.com/meetecho/janus-gateway/files/275192/log.txt) indicates that the "configure" request for turning off recording is received first and then the DTLS alert.  After the DTLS alert, the audio recording is closed with a file name.  Immediately after, it is closed again, this time without a file name.

My colleague set up that server and he runs this version of Janus:

    commit 5f1437cf49d8051a6021a0332e2fac425716fb44
    Merge: 31ae4f2 85f4116
    Author: Lorenzo Miniero <lminiero@gmail.com>
    Date:   Sat Apr 23 00:40:02 2016 +0200

andreasg123:
This attempts to fix #530.  Recorders for a publisher in a video room have to be protected against parallel closing as can happen if a "configure" request and a hangup arrive at almost the same time.  

Room participants only have the `listeners_mutex`.  Even though the name isn't appropriate here, the mutex is otherwise fine for this purpose.  Only if a listener wants to be added to or removed from a publisher at the same time as the publisher turning recording on or off, there would be a very slight delay due to the use of the same mutex.  I don't consider that to be significant enough to favor a second mutex.

I also protected the creation of a recorder in response to receiving an SDP.  It may be possible that a hangup would arrive shortly after the SDP so that the recorders should be protected here, too.

This PR is more important than  #523 because it fixes a server crash.  I'll rebase the other PR and deal with the merge conflicts if you merge this PR.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -2120,6 +2120,9 @@ void janus_videoroom_hangup_media(janus_plugin_session *handle) {
 	if(session->participant_type == janus_videoroom_p_type_publisher) {
 		/* This publisher just 'unpublished' */
 		janus_videoroom_participant *participant = (janus_videoroom_participant *)session->participant;
+		/* Lock listeners_mutex to protect recorders from race conditions. */
+		/* This mutex doesn't have an appropriate name but it is less coarse than participant->room->participants_mutex. */
+		janus_mutex_lock(&participant->listeners_mutex);
 		if(participant->sdp)
 			g_free(participant->sdp);
 		participant->sdp = NULL;
@@ -2143,7 +2146,6 @@ void janus_videoroom_hangup_media(janus_plugin_session *handle) {
 			janus_recorder_free(participant->vrc);
 		}
 		participant->vrc = NULL;
-		janus_mutex_lock(&participant->listeners_mutex);
 		while(participant->listeners) {
 			janus_videoroom_listener *l = (janus_videoroom_listener *)participant->listeners->data;
 			if(l) {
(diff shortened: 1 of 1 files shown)


lminiero:
The patch looks fine to me, although I agree a dedicated mutex just for recorders (as you suggested also in other related PRs) would be clearer and more apt. I can take care of that for all plugins where this is needed, using your source as a basis.

Just to be clear, so the fix is just to avoid the risk of freeing a recorder twice, right? And not in actually using the recorder for recording after it has been freed, otherwise this would require the mutex to be locked for those operations as well (which we shouldn't do for performance reasons).

andreasg123:
Yes, the fix protects the recorder from being freed twice.  For using a freed recorder, there is the mutex in the recorder that may protect against most of it.  There may still be a scenario that could cause problems, but hopefully less severe than freeing the recorder twice.  Just like in #533, one can remove the test for the recorder before calling `janus_recorder_save_frame` but I'm not sure if that actually would make a difference.

Regarding the mutex to be used, I see three options:

1. Use `listeners_mutex` the way I did it.
2. Rename `listeners_mutex` to `mutex`.  That would make it look like #533.
3. Add a new member `recorder_mutex` to be used for this case.

Which one do you prefer?  I can change the fix to the chosen option.

lminiero:
Sorry for this late reply. I think we can keep the mutex you're using, for now: I can take care of adding the new mutexes to this and other plugins just for recorders in another commit or PR.

Thanks!

PR #533, opened 2016-05-23
Avoid race condition when closing recorders in SIP

andreasg123:
Based on my earlier comments in #530, here is how I would fix the SIP plugin.  Note that I removed the `if` statement in places like this:

	if(session->vrc)
		janus_recorder_save_frame(session->vrc, buf, len);

Protecting that section with a mutex might be better but this is pretty safe because the value of `session->vrc` is only checked once.  Even if it is set to `NULL` in the middle of the call to `janus_recorder_save_frame`, saving the frame should succeed.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -1000,8 +1000,7 @@ void janus_sip_incoming_rtp(janus_plugin_session *handle, int video, char *buf, 
 			}
 			if(session->media.has_video && session->media.video_rtp_fd) {
 				/* Save the frame if we're recording */
-				if(session->vrc)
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! Removing the if is fine, as we check the validity of the pointer within `janus_recorder_save_frame` anyway. We probably don't even need a lock, but just an atomic int we can use to check whether we've freed something already or not (although a lock is easier and quicker).

I'll act on this tomorrow as right now Github is acting crazy (it's been stuck on "Checking for ability to merge automatically…" for a while).

lminiero:
Merging this as well, and same considerations on the mutex name I can change later on. Thanks!

PR #536, opened 2016-05-23
Event handler plugins

lminiero:
Hi all,

this is a pretty major change in Janus, as it introduces a new category of plugins: event handlers. More specifically, now Janus (and plugins) can originate events of specific types (e.g., session created, handle detached, peerconnection going up, media available, packet loss, plugin-specific data, etc.). These events are then passed to all the event handler plugins that subscribe to that type, so that they can be, indeed, "handled" somehow.

Since the Admin API is pull-based, this should be useful/helpful for several reasons. You may want, for instance, to use some events as a "trigger" to do an Admin API pull, rather than polling continuously; or you may want to save these events to a database, so that you have a view of what happened in each media session; or send them to an external component for processing; etc. I expect a typical usage will actually be implementing plugins that interface to external services (e.g., Homer, callstats.io, etc.) in order to have them analyzed/processed there.

At the moment I only implemented a very simple sample plugin, that basically forwards any event it receives to a web backend via HTTP POST. If you want to give it a quick test, enable it in `janus.eventhandler.sampleevh.cfg` (it's disabled by default), set a valid backend (e.g., `http://localhost:8080`), and wait for events to be received by your web server. A very simple way to set one up just for testing purposes is this node.js code, that will just print whatever it receives on standard output:

```
var http = require('http');

http.createServer(function (req, res) {
    var body = "";
    req.on('data', function (chunk) {
        body += chunk;
    });
    req.on('end', function () {
        console.log(body);
        res.writeHead(200);
        res.end();
    });
}).listen(8080);
```

That said, this is a very rough first attempt at defining how this new interface could be implemented. For instance, the available event types are limited to what came to my mind when I started writing the code, and so can definitely be extended. At the moment the kind of events (macrocategory, if you will) you can receive in an event handler are:
- session related events (e.g., session created/destroyed, etc.)
- handle related events (e.g., handle attached/detached, etc.)
- JSEP related events (e.g., got/sent offer/answer)
- WebRTC related events (e.g., PeerConnection up/down, ICE updates, DTLS updates, etc.)
- media related events (e.g., media started/stopped flowing, stats on packets/bytes, etc.)
- events originated by plugins (at the moment, all of them, no way to pick)
- events originated by transports (at the moment, all of them, no way to pick)

Not sure if we can have more, or if they can fit in there. The idea is that the core generates events of a certain macrocategory, each with different details. Event handler plugins can then subscribe to those macrocategories (none, some, all) in order to be notified. I realize this is pretty crude (you can't filter which plugins you're interested in, for instance: it's all or none), but for the sake of simplicity (handler plugins edit a mask, the core checks it before relaying) I chose to do it the way it is now for the moment.

Events are always provided as `json_t` objects: I chose to do that instead of creating a new struct for the purpose as (i) a JSON object is generic enough to cover pretty much everything, and (ii) its `json_incref`/`json_decref` mechanism allowed for a much easier management of referencing these events (the same event may be shared across multiple handlers, which may handle them at any time). Besides, it makes it easier for plugins that just want to serialize the data and shoot it somewhere (as my sample plugin does, incidentally). I tried to define a common syntax to refer to, though, and specifically:

```
{
    "type" : <numeric event type identifier>,
    "timestamp" : <monotonic time of when the event was generated>,
    "session_id" : <unique session identifier>,
    "handle_id" : <unique handle identifier, if provided/available>,
    "event" : {
        <event body, custom depending on event type>
    }
}
```

The monotonic timer is there in case event handler plugins need to know how much time passed since the event was generated and when they actually managed to handle it. You can see the exact format of the events I implemented so far by looking at the messages my sample plugin handler shoots, as explained above.

As to which events are available, I only implemented only a few of them, in this first version, and specifically:
- session created/destroyed/timeout
- handle attached/detached
- JSEP local/remote offer/answer
- ICE states, DTLS states, selected ICE pair, PeerConnection up/down
- media started/stopped flowing

Relevant messages that are missing are, for instance, statistics of some sort, and other will definitely come to mind, but I guess it's a good start for the moment. As a side note, for what concerns originating events in other plugins I only integrated the "notify event" feature in the EchoTest plugin, for the moment: specifically, any time you send the EchoTest plugin a new message, it sends the core the current internal configuration (audio/video/bitrate) as an event. Of course the plan is to use the feature for many more events that may be relevant to specific plugins, e.g., SIP calls in the SIP plugin, people joining/leaving/publishing/etc. in any of the communication plugins, and so on. This really will depend on what people will think is necessary, and what isn't.

I believe that's all, for now. I don't expect it to be very optimized right now, and I haven't even started figuring out if it has any impact on performance, and how much. At the moment each event is generated as soon as the "thing" happens, and as such is originated by whatever thread is in control of the activity. No message bus or anything of that sort is implemented: this means that, when receiving an event, handler plugins MUST just increase the reference to the object and just enqueue it, to then process it ASAP in a thread of their own (pretty much as most of the media plugins do anyway). This should help avoid slowing Janus down, as doing any I/O or event processing in the callback itself would instead.

Any comment? Feedback? Suggestion?


lminiero:
... no opinion at all on this?! :wink:

saghul:
I had a quick glance. Do you have anything other than monitoring in mind? As for sending the events, maybe it's worth it to have an auxiliary thread pulling the events out of a queue and dispatching then? This way senders can send an event from anywhere, but receivers will always get them from the same thread, serialized.

lminiero:
Monitoring and statistics collection, e.g., for ex-post troubleshooting, are the main things that come to mind. The auxiliary thread you mention could be something like the "message bus" I cited, and is one of the enhancements I could make along the road. I'd still require handler plugins to handle events separately, though, as while they would not delay Janus by itself now, they'd still be able to slow or even lock (in the worst case) the auxiliary thread itself.

lminiero:
Added new events (SIP messages, some RTCP and media related stuff, and others), and disabled the feature by default. It can be enabled either by explicitly setting `broadcast=yes` in `[events]` in `janus.cfg`, or by passing `-e` on the command line.

lminiero:
Also added a thread which does the actual broadcasting: core and plugins just enqueue events, and the thread shoots them out.

nowylie:
This looks like a pretty good idea, the only change I might suggest is not hard coding event types.

lminiero:
> This looks like a pretty good idea, the only change I might suggest is not hard coding event types.

Not sure what you mean by that? Without a formal definition of events, event handlers would not be able to recognize them reliably and handle them. If you mean the fact they're numeric (and more specifically a bit map), that was a choice to make them easier/lighter to subscribe to/unsubscribe from, and filter when notifying to just the interested handlers.

nowylie:
Your formal definition for events doesn't need to be in the event dispatching code though. Let the sender structure the event, the dispatcher can pass around arbitrary JSON data and a numerical event type, and the receiver can interpret the structure. The sender and receiver can agree on the format out of band through the event type.

Then if you wanted to make things really flexible, let the numeric values assigned to event types be dynamic by introducing some kind of event type registry.

lminiero:
But not all event handlers will just send stuff out exactly as it is like our demo plugin does. Other event handlers will receive these events, parse them expecting to know how the core formatted the stuff, and then use the info in order to do something (e.g., send out aggregated stuff, saving processed data somewhere, etc.). For this to work, the JSON data these handlers receive cannot be arbitrary (apart from the plugin specific info, of course), otherwise handlers won't know, for instance, how to know when a handle was attached, a PeerConnection went up, or media stopped flowing.

nowylie:
This is what I meant by agreeing on the format out of band. For example if the audiobridge plugin started dispatching events of thte type `AUDIOBRIDGE_PARTICIPANT_JOINED`. It's reasonable to assume that my event handler, that registered interest in `AUDIOBRIDGE_PARTICIPANT_JOINED` events, will understand the format of any events it receives with that type.

lminiero:
I feel very dumb but I stil don't understand the problem :smile: 

Plugins already send events that are opaque to this mechanism. Type is always `JANUS_EVENT_TYPE_PLUGIN` there and plugins can choose what to notify and how (event handlers are supposed to be familiar with the specific plugin itself to know how to process those events). For events defined in the core, type is more constrained, and already works the way you mean: e.g., if the core sends an event of type `JANUS_EVENT_TYPE_WEBRTC`, the event handler knows what this is about and the format it will have.

The main reason to have these types defined in advance, is to allow event handlers to tell the core what they're interested in and what not. The core may (and will) generate many events for each call, so if your handler is not interested in all of them, no point in flooding it with info it will discard, it's a waste of resources.

Or did you mean something else?

nowylie:
I think our confusion might be in the definition of 'event type'. I'm imagining an event type to describe 'what happened', where I think you're saying that it means 'what kind of event happened'. Is this right?

I only advocate moving the formal definitions out of the event dispatching code and into the event generating code. Mainly as a means for separation of concerns.

lminiero:
What do you mean by event dispatching, specifically?

nowylie:
In events.c:
`void janus_events_notify_handlers(int type, guint64 session_id, ...)` could be `void janus_events_notify_handlers(int type, json_t *data)`

Then somewhere else you could have something like:
```
json_t *new_session_event(guint64 session_id, char *name) {
	json_t *data = json_object();
	json_object_set_new(data, "type", json_integer(JANUS_EVENT_TYPE_SESSION));
	json_object_set_new(data, "timestamp", json_integer(janus_get_monotonic_time()));
	json_object_set_new(data, "session_id", json_integer(session_id));
	json_object_set_new(data, "name", json_string(name));
}
...
janus_events_notify_handlers(JANUS_EVENT_TYPE_SESSION, new_session_event(session_id, name));
```

lminiero:
Ah, got it. The reason why we use variable arguments is that it makes it much easier to trigger event from within the core. We call what looks like a function, without having to worry about generating a `json_t` ourselves, which is something `events.c` can do for us.

nowylie:
That makes sense and is the purpose of the `new_session_event()` function in my previous comment. Sorry if I'm not making much sense, it's late over here. I might mock up an example of what I'm imagining tomorrow to give you an idea :)

lminiero:
No problem at all! Your feedback is very welcome, actually, the more discussion there is about this new feature, the better, as I want people to find it useful and meaningful, obviously.

cacheworks:
This is awesome, getting info on when events happen without polling the Admin API is very useful and allows me to keep track of sessions and handle info a lot easier.
And with media / ice /dtls info it makes monitoring and debugging a lot easier!

One thing I ran into is that I wanted to have an extra field (display) added to the "joined" event in janus_videoroom.c to keep track of who's session it is, so I added:
`json_object_set_new(info, "display", json_string(display_text));`

Do you think adding extra information like this should be done by modifying the plugins as I did, or would it be possible to include more information to start with, and let the event handler plugin decide what to send out?

I could imagine forwarding unused data to the handler plugin isn't very desirable for performance, would it be possible to only listen for certain fields as is possible with the event types? (And would that actually save resources or are the events still being sent then)

Thanks for your continuous work on this project

lminiero:
Thanks for playing with this! If you use it for a while and eventually have any data to share, in particular with respect to performance with and without the feature enabled, please let us know.

As to the data, of course if more info is needed, it's the source that needs to add more (in this case, the plugin), as otherwise the handler can't make it up. Anyway, the plugin you tested is just an example of a handler, and this one happens to forward everything it receives, with all the fields out there. If you want less info, aggregated info, or a different management of events, rather than messing with the sample handler (which is an option but a dangerous one, as you risk not being able to merge updates along the way) you can write a new plugin that does whatever you want with the events.

cacheworks:
Are there any specific metrics you'd be interesting in seeing in regards to performance? I was just looking at total CPU and memory usage whether it was around average and couldn't tell much difference there with a handful of videorooms, but I'm not really measuring in a proper way with the same amount of clients

I understand the source will need to add more data if needed, I just wondered if more information could be included by default, so someone writing a handler plugin can choose what to use without creating both a normal and handler plugin (and perhaps core changes if they lack data in those events too?)

I agree that changing existing plugins too much will be a pain for merging updates, but adding a whole new plugin to add a single line to it for the handler plugin seems like a long-winded way of achieving the same thing.
So would you consider adding `json_object_set_new(info, "display", json_string(display_text));` to janus_videoroom.c in the "joined" event, so that the handler plugin can choose whether or not to display it, or do you think I should duplicate the videoroom plugin to add that for myself?

I would think display is used more often by admins to identify videoroom users, without it my list of sessions isn't very useful to tell who's who, so I want to receive it at least once so I can associate it with the session id in my interface

Thanks

arjenveenhuizen:
> I would think display is used more often by admins to identify videoroom users, without it my list of sessions isn't very useful to tell who's who, so I want to receive it at least once so I can associate it with the session id in my interface

I concur, we (mis-)use the display field to this purpose as well. Would be a good addition :).

cacheworks:
Well that was quick, thanks @lminiero !

lminiero:
Just added.

As to the performance, any info you have available would be cool to have! Each handle may generate several events (both from the core and the plugins), which means more things to do in general (e.g., allocating event data to pass to handlers, and the activity in the handlers themselves), which multiplied by the number of handles can be something to take into account.

If it turns up that, even with several rooms and several participants, the load is pretty much the same, then even better, but I'd like to be sure that's indeed the case. In general, the more you guys play with this feature the better, as it brings us a step closer to merging it! :+1:

joshdickson40:
Could we get a merge in from master with this branch? Would like to give this a go and it looks like there are existing conflicts.

lminiero:
Probably small conflicts due to very recent changes, as the last merge dates back to just last week. Will do that early next week.

joshdickson40:
Wonderful. Thank you as always, Lorenzo.

lminiero:
@joshdickson40 done, looking forward to your feedback!

lmangani:
Thanks @lminiero for the awesome extension and for the live session helping us bootstrap - Homer integration is in the works!

vr000m:
@karthikbr82 PTAL.

joshdickson40:
Would you mind merging this again please? Looks like a conflict in the videocall plugin.

lminiero:
Sure, I'll do that tomorrow.

joshdickson40:
Great, thanks!

lminiero:
@joshdickson40 just did that, looking forward to your feedback!

joshdickson40:
One feedback point: I really think this needs event queueing (at least the option) considering the amount of data that can come from this. Something like queuing up 100 events / 1 second, whichever occurs sooner would definitely be a positive addition here.

README.md
```diff
@@ -229,7 +229,7 @@ or on the command line:
 
 	<installdir>/bin/janus --help
 	
-	janus 0.2.0
```
joshdickson40: Seeing as libcurl is required to build the event handler, we should probably note that in the readme. On Ubuntu I am using libcurl4-openssl-dev

joshdickson40:
This is very cool! A great start. Right now my feedback is definitely:

* Batching... for instance, there are *way* too many POSTs sent during call join for my liking. Unfortunately it looks like the API here is definitely set up for single events. I would much rather see the JSON structure implement an array of events by default, and if you opt to not have batching, that array will simply always have length 1. 
* Using 'display' field as often as possible, looks like it was added to the video room only (looking at audio room & text room for starters)
* This definitely requires some kind of authentication. Could there be an optional token in `janus.eventhandler.samplevh.cfg` that would be transmitted for each POST? I don't really have a huge opinion about what kind of auth this should use seeing as every integration with third party services in my API is different - but there should definitely be an option.
* Outgoing POST requests hung longer than I would expect them to, I was able to get quite a few queued up when my dev API was offline.
* I'm unclear what protocol (if any) there is for retrying events or if this is necessary. Maybe it's worth a discussion.
* I'm not sure the 'type' event field makes sense. I'm left wondering why it isn't just filled with the string that its describing internally. It would make much more sense for a media event to have type `media` than type 32.

I actually think the POST request hang time, event retries, and batching are interrelated here. We don't have to reinvent the wheel here because plenty of real time apps already have this figured out. 

The idea would be to retry POST requests should one fail (fail in this sense is either a timeout or a non 2XX response). A ~10s timeout also seems like a better idea (not sure what it is now, but it's got to be over that.

If we see a failure, we should also stop sending additional POST requests and queue them until one succeeds. It's too easy to accidentally DOS yourself otherwise. Obviously if this queue runs over some number of elements, you would not continue to queue should the POST destination remain offline for an extended amount of time. Assuming that the failure is temporary, we'd then empty the queue once the destination API is back online.

Again, really, really good start.

README.md
lminiero: Ack.

lminiero:
@joshdickson40 thanks for the detailed feedback! Before addressing the individual points, though, there's an important distinction to make, between the sample plugin and the internal mechanism itself. The sample plugin is not THE aysnchronous event mechanism, it's just an example implementation of a very (maybe too) simple way to use the information the real mechanism provides, specifically by forwarding it as it is via plain HTTP requests. Many improvements on the actual forwarding/aggregation/formatting can be done in different plugins that implement different behaviours (e.g., a different transport protocol like using RabbitMQ queues, aggregating requests, changing the syntax of the messages, dropping some events, etc.) As such, while we're of course interested in presenting a good enough reference implementation there, I'm mostly interested in being sure the internal stuff is enough and works as expected, so that other developers have all they need to then work on new event handler plugins that do it better than the sample we provide, and the way they need it to work.

Coming to your points:

1. **Batching:** see above, not really internal-related, but hopefully only related to the way the sample plugin is implemented. Specifically, we have an async queue that is awaken any time a new event arrives. As such, grouping would mean introducing some kind of prediction mechanism (how do you know another message will arrive soon thereafter?), or changing the wake up process by making it time based (e.g., every 500ms) so that you can send all those you find in the queue at a specified time (which can also delay the delivery of some events, though).
2. **Display:** sure, I guess this can be done. In general I tend to prefer avoiding redundant information (the display is something you get the first time, so you can recover it from other unique identifiers in subsequent events), but sometimes it's harmless and lightweight enough that it's better to have it there than forcing you to obtain it locally, e.g., via maps.
3. **Auth:** see above, sample plugin related. Cool to have but not key to the purpose of the PR. An enhanced event handler plugin could add that, along other things. Anyway, if it's trivial to do with libcurl no problem in doing that.
4. **Hung POSTs:** probably caused by the fact we use a single thread for everything, and HTTP POSTs as well. Not sure how we can improve this for the sample, I guess there are ways to configure timeouts in libcurl (don't remember if we're doing this already). An alternative would be using different threads for outgoing requests, even though that could get out of hand very soon, especially if tons of events are being sent. Maybe a thread pool of a limited size can help limit the damage. Anyway, this is probably too advanced for the purpose of the sample implementation, which is only mostly demonstrative.
5. **Retrying:** the sample doesn't care, it receives an event, it serializes as it is, and it sends it via HTTP. If it failed, it failed. I guess we can just re-add it to the queue if it failed, maybe with a counter and/or an exponential backoff kind of approach, so that we don't try forever and don't hammer in a loop. Again, maybe too advanced for the sample itself.
6. **Type:** see above, the sample serializes exactly what it receives, and this includes the internal `type` identifier we use for our own purposes. Using a string internally would be way less efficient, and that's why that's a number instead. Turning this into a string when notifying it externally is a serialization matter, and so related to the way the specific event handler plugin will decide to process incoming events. As such, I'd rather not change the way the sample plugin works, as the purpose there is to show exactly what each event looks like, with no manipulation at all. Different plugins can do whatever they want with events instead.

Thanks again for fostering the discussion on this PR, it really helps! Hopefully we're a step closer to finally be able to merge this :smile:

lminiero:
@joshdickson40 these commits address most of the points you made: you can find a couple more options in the sample event handler configuration, namely with respect to event grouping (up to 100 events in a single POST) and HTTP authentication. I also added a 10s hard limit for POSTs. Tried all this briefly and it seems to work, but feedback would be welcome of course.

No change on retries, instead, that's more complicated than it sounds the way events are handled/queued right now.

lminiero:
PS: no thread pool used yet either, as I'm not sure about this now. Specifically, it may result in events getting to the backend completely out of order, depending on how threads are scheduled and executed (although all requests do have a timestamp that should help with that). Anyway, I guess that's an issue that can happen with separate HTTP requests as well. A different plugin employing a different transport/delivery mechanism may handle this better.

joshdickson40:
Okay, I definitely better understand what you're trying to do with this now per your replies. Everything makes sense.

The updated changes you have here will roll out onto our EC2 cluster later today, so I should have some feedback soon for whether or not it's causing us any issues compared to master. So far so good! Thanks as always, Lorenzo.

lminiero:
That would be awesome, thanks!

joshdickson40:
Trying this out now. It looks like all POSTs are being sent twice. First it's sent without auth (which my API is replying 401 to), and then the same payload arrives bearing the correct credentials. It does look like you're setting them, so I'm not sure what's going on here.

lminiero:
> First it's sent without auth (which my API is replying 401 to), and then the same payload arrives bearing the correct credentials.

Isn't this how it should work? It may be because we set an autodetect in libcurl to check which authentication mechanism is in place. Since it doesn't know what the server will ask for, it first tries normally, gets a 401 (maybe with digest) and answers to that. This might be circumvented by explicitly setting the authentication mechanism, I guess, but not sure.

joshdickson40:
Well, I suppose technically yes, but I don't think that's the desired behavior that you'd want to see from something calling your API. I would expect that if auth is set via config, every attempt would specify the credentials, not just after a 401. 

Maybe this is an artifact of this not really being the right mechanism for auth here?

A more sensible routine for API auth would be:
* Config specifies public & private keys, not username & password (so, basically what it is now)
* Janus sense public key as header, something like 'X-Janus-Key'
* Janus calculates a HMAC SHA256 hex digest from signing the payload with the private key, and sends the digest as 'X-Janus-Signature'

Now we have the benefit of 1) never sending the secret key, 2) not sending all POSTs twice for auth, and 3) following better API design guidelines.

lminiero:
Yeah, but that's something libcurl needs to do, not us. As I said, it's probably caused by the "autodetect" mechanism we use. Checking the docs for curl:

    --anyauth

    (HTTP) Tells curl to figure out authentication method by itself, and use the most secure
    one the remote site claims to support. This is done by first doing a request and checking
    the response-headers, thus possibly inducing an extra network round-trip. This is used
    instead of setting a specific authentication method, which you can do with --basic,
    --digest --ntlm, and --negotiate.

This means that setting the method in advance would do that, something that complicates things though as I have to expose additional configuration options, implement different calls to the library, etc., something I'd hoped to avoid. For now I'll change the code to use basic authentication by default, as that's quite immediate. Later on we can worry about other methods.

lminiero:
Just done that.

I also added a new category of events for plain core-related events. Right now this only includes an event when Janus has been started, and one for when it's being shutdown. In the future this might have other Janus related stuff, like health info, cpu/ram stuff, and whatever will make sense to see there.

joshdickson40:
> In the future this might have other Janus related stuff, like health info, cpu/ram stuff, and whatever will make sense to see there.

I had meant to put this on my feedback list so +1 for that.

I'm still a bit confused about why you're using username & password authentication for an API route in the first place, though. I agree that it should be on libcurl to figure that out, and it sort of feels like this is putting an incorrect hack (forcing basic) on top of another incorrect hack (using username & password authentication for an outbound API route).

It [looks like](http://stackoverflow.com/questions/29862306/encoding-a-string-in-hmac-sha256) it would be just as simple to do this in a more correct way that's consistent with how other integrations would handle this, and we wouldn't then have to change things in the future. OpenSSL is already a dependency and that's all we'd need.

(I'd be happy to make the change, by the way, I'm just a little unsure about exactly how I'd push that change into this PR)

joshdickson40:
I'm not exactly sure how to best show the diff, but I played with this and [implemented](https://github.com/joshdickson40/janus-gateway/blob/77f315b283df3832dbe2890d66176f72c9455324/events/janus_sampleevh.c) an HMAC SHA256 hex digest for API authentication instead of trying to work around the username & password issues. Maybe we could give this a try instead.

(I am sure my C needs some work, though this does work.)

This has quite a few advantages over the username and password configuration, and seeing as auth has been supported for ~24 hours, it seems like it would be a good idea to do this correctly now.

[Pusher implements](https://pusher.com/docs/webhooks#authentication) a similar approach for how they authenticate webhooks.

lminiero:
Isn't all this very dependant on what the server supports? If the backend someone uses only does basic auth and they still want to do that, this won't work. I don't want to put too many constraints on what you can use to receive events. Again, this is meant to be a very dump skeleton plugin people can use as an inspiration...

joshdickson40:
I suppose to an extent, but I don't think I know of any backend frameworks that support basic auth but don't have support for making the HMAC calculation simple, nor has that fear hasn't stopped a lot of services from having HMAC-only authentication for webhooks. If you're going to go through the process of creating new API routes for this, I don't think the work to verify the HMAC is an issue.

It does look like [Stripe](https://stripe.com/docs/webhooks) supports only basic auth as do a few others I came across. Basic auth is definitely not "wrong," I just don't know why we wouldn't pick the best solution given that this is brand new. [The topic](https://groups.google.com/a/lists.stripe.com/forum/#!topic/api-discuss/-zqruS0lTig) has come up a few times on Stripe's forums with people requesting the signature, which is what Braintree already does.

I am more comfortable with the HMAC solution, so I'm going to continue tracking any changes you make here and run that. I'd like to merge that back in, but, if you want to keep the basic version, that's fine too. Maybe we can get some additional feedback on this?

lminiero:
That's a good point. I guess we can make this configurable, and keep things super simple for those who want to. I'll try and merge your changes in here tomorrow. One quick question, as I'm not an expert on this stuff: is the value there always the same, or does it change for every time you call the HMAC stuff? I guess that being the hash of a fixed secret, it's not going to change, which means it's probably best to generate that only once and not for every request. Or am I missing something?

joshdickson40:
The value of the HMAC is not the hash of a fixed secret, it's different every time (well, assuming the inputs are different.) There's basically three parts, there's the hashing algorithm (SHA256), the secret key, and the message that's hashed (which is the `event_text`). So assuming `event_text` changes, the hash changes as well.

What you end up with is a signature that proves the sender has the secret key, while also verifying the integrity of the message (i.e. if either the message or the key were different, the API-side verification will fail). The hash is one-way, so you cannot decipher the secret key from a number of hashes. And (maybe most importantly) the secret is never actually transmitted.

To authenticate the request on the user's API side, they:
* Verify that the request carries the `X-Janus-Key` and `X-Janus-Signature` headers and a JSON body
* Check that the key is valid (for me, this will carry info on which Janus instance was the event originator)
* Calculate the hash again using the request body and the secret key:
```
crypto.createHmac('sha256', JANUS_AUTH_SECRET)
    .update(JSON.stringify(body))
    .digest('hex') !== signature)
```
* If the hash matches the value of `X-Janus-Signature`, then you know the request is legitimate.

This also necessitated the change to:
```
char *event_text = json_dumps(output, JSON_COMPACT);
``` 
... because the whitespace would affect the payload length, which would generate the wrong hash.

drichardson:
I agree with @lminiero that any auth is likely to be application specific, and should not be included in Janus mainline.

If, however, some sort of auth should be included, it should be based on well understood standards rather than custom solutions (e.g., JWT).

In particular, the mechanism @joshdickson40 has proposed will not work for a couple reasons:

1. HMACs require symmetric keys. You can't sign with a public and verify with a private, or vice versa. There are some digital signature algorithms that use asymmetric keys, but HMAC is not one of them.
2.  [JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) doesn't guarantee the order of non-array objects, so even if the other parts of the algorithm were correct, different JavaScript implementations (and JSON deserializers/serializers) could produce different output which would result in the signatures being different even though the JSON data looks identical.

There are probably a number of other problems too, because developing cyrotographic protocols is difficult and needs peer review. Composing well understood and secure cryptographic algorithms provides no guarantee that the cryptographic protocol you've created with them is secure. I think it's better if Janus avoids doing anything in this area other than showing a reference implementation and perhaps providing hooks to an application specific authentication mechanism.

joshdickson40:
Doug,

1. The HMAC implementation I linked to does not sign with a public key and verify with a private key or vice versa. The private key is used Janus-side to compute the hash, and it's used again API-side to re-compute the hash. The public key is sent as `X-Janus-Key`, but it does not affect the hash operation at all.

2. In v8/Node object property order is preserved provided you are not using integer object keys (which you should not be), so JSON.stringify is safe here (and this has nothing to do with Janus, of course).

Furthermore, this does work - we are currently using it in production.

I am not developing a cryptographic protocol here, the method that this is using is a well-designed, well-understood, off-the-shelf approach, and we're using OpenSSL to do the heavy lifting. As I've cited earlier, this is the exact approach that [Pusher](https://pusher.com/docs/webhooks) takes with its webhook authentication. There are dozens of other examples of APIs using this kind of authentication for webhooks (exactly what is signed varies, but the procedure is essentially the same).

JWT is definitely out of scope for Janus here - that I agree with you on. I also agree that it's fine that Janus implements a bare-bones approach. But, if the idea is to get things like this pull request thoroughly tested in a real environment doing real work, it has to support some level of real usage from a feature standpoint. In my mind, batching (lowers DOS risk) and some kind of basic authentication (I'm uncomfortable running an exposed, auth-less API route) are required. And if we're going to go through the trouble of implementing username & password auth, all I'm saying is that it makes sense to use a different approach that is just as easy to implement while offering numerous advantages.

Josh

drichardson:
@joshdickson40 oh I misunderstood. Is the public part of the key used anywhere? If not, why use a public/private keypair at all?

joshdickson40:
The key (i.e. the "public" key) is doing two things:

1) It's basically token authentication (so essentially performing the same role as username:password, through the protocol is slightly different obviously) and
2) It can be an identifier to figure out where the request originated from

So, the API-side verification becomes 1) was this sent from a valid location (i.e. is the key right), and 2) does that location know the secret key (i.e. is the hash right).

Sometimes all of the callers will use the same identifier, sometimes not. For a third party integration, you're going to see that filled with your account key most often regardless of the machine that physically sends the request. If we had two Janus instances A & B, we could also provide an identifier in that field, so you know where the webhook originated from.

lminiero:
Now I'm confused again :wink:  The fact itself we're using a non-standard header like `X-Janus-Key` confuses and scares me. No matter how common this might be out there (again, not really an expert so I have no idea), this looks like something that has not a wide enough scope to be considered standard to me.

Again, for this simple demo I really don't believe we should do fancy stuff. Relying on plain (auth/digest, I'll expand what we have depending on what libcurl exposes) and standard authentication is more than enough. If anybody is interested in anything more complex than that without writing a new plugin of their own, they can just add their own mechanism to the existing plugin, as Josh did.

Does this make sense?

joshdickson40:
I mean, I guess it depends on what you mean by "non-standard." I don't think putting an app key as `X-Janus-Key` is unusual at all. Is it to some kind of spec like username & password basic auth? Obviously not. Is this approach common for authenticating webhooks given that there is no standard for trying to do this simply? Very. 

I obviously don't see an issue just going with the HMAC route, but I'm more than happy to just sit my patch on top of it. I'd much rather you spend time on other things so we can get this merged than me take up your time talking about auth schemes for this :)

joshdickson40:
Okay, finally a non-auth question.

I'm trying to sort through the event firehose to detect 'room open' and 'room close' events. Say that I have one user, and I am going to have them join an audio call, and then leave the call some time later.

When they dynamically create the room and join, I see:
```javascript
{ type: 64,
  timestamp: xxx,
  session_id: xxx,
  event: 
   { plugin: 'janus.plugin.audiobridge',
     data: { event: 'created', room: xxx } },
  handle_id: xxx 
}
``` 

I never see the close event, even when waiting for a while.

My message to set up the room is:
```javascript
message: {
    'request' : 'create',
    'room' : xxx,
    'secret': xxx,
    'permanent' : false,
    'is_private' : true,
    'sampling': 48000,
    'publishers': xx,
    'record': true,
    'id': xxx
},
```

My understanding was that setting 'permanent' to false meant that the room would be destroyed when it was empty. Is my understanding off, or is the event not fired, etc?

lminiero:
> << I'm moving this issue as I realized it doesn't have anything to do with this PR >>

I had a mail notification about a missing event when the room was destroyed. Wasn't this the case?

joshdickson40:
Hi Lorenzo. Do you have a timeframe in mind for this getting merged? Has worked very well for me so far and I would like to open PR's for some of the tertiary things I've built on top of this.

lminiero:
I hope soon! I'm currently abroad for a conference, so if you guys can
discuss any shortcomings or limitations that should be addressed, in the
meanwhile, I'll think about fixes and consider merging when I get back next
week.

Il 17/ott/2016 21:05, "Joshua Dickson" <notifications@github.com> ha
scritto:

Hi Lorenzo. Do you have a timeframe in mind for this getting merged? Has
worked very well for me so far and I would like to open PR's for some of
the tertiary things I've built on top of this.

—
You are receiving this because you were mentioned.
<https://github.com/meetecho/janus-gateway/pull/536#issuecomment-254369835>,
or mute the thread

joshdickson40:
The one thing I would definitely like to see (and intend to submit a PR for if you don't get to it before then) is a short period where requests are retried if the first one either 1) times out or 2) returns a non 2xx response. I am using this to maintain state information and it's just too much to ask for that connection to never have an issue across thousands of API calls.

What I would propose would be either a short period (~2 min) where failing requests are retried, or a number (if the request timeout is 10 sec, this would be ~10-12 attempts) where the request would be retried.

In reality this is a bit more complicated though. If we sent message A at time 0, and 6 seconds later message A fails to send, we will resend that at time 6, but it's possible that we sent messages B and C that received 200 OK at time 3 and time 5, respectively. It would be very simple to envision a case where this leads you to assume incorrect state information.

The correct fix for this would be to block on sending further events and enqueue them until we get a 200 OK on a previous outbound call. I haven't looked upstream to see if this is being handled at all, but I think you'd have to implement this queueing along with the retry logic.

And again, I am completely sympathetic to the fact that this is experimental and a proof of concept that's not intended to be 100% robust, and you might want to wait on doing this until after you've merged the existing functionality. But! I want to be depending on this, and for that, retries definitely need to be in there at some point :)

lminiero:
An incorrect state is indeed one of the potential issues that may arise, although we have a timestamp in the JSON that helps you reorder stuff if needed (but that may not be enough). In the current version it's not easy to simply retry: I'll add a mechanism to retry the last transmission a fixed (configurable) number of times.

It's not easy to understand what should trigger a retry, though. Is it a 404? Something else? What is a failure, exactly? And how do we differentiate it from broken backends or configurations, e.g., from requests that would never succeed?

joshdickson40:
It sounds like we agree on timestamp. The main issue there is that you may retroactively realize you are out of sync, but you might have already acted on what you thought was the next state change in real time. I'm sure you could build some difficult, convoluted logic to sort things out API-side and get back in state, but that's obviously not a good way to do this and would be quite error-prone.

A retry happens for anything that's not a 2XX response inside the allowed request period (I'm forgetting specifics but I think you set it to 10 seconds). If you get a 404, 500, 503, 504, etc, those all trigger a retry. If the timeout period halts the request (i.e. the request is hanging for a substantial amount of time), the retry occurs.

I am not 100% sure what you mean regarding broken backends / etc. My assumption here is that:
- At any one point in time, Janus is sending at most one set of events to the API backend and enqueueing any other events that are ready to be sent while waiting for the result of the outbound message (I'm assuming batching here)
- If the outbound message succeeds, Janus checks for any enqueued events, and if some exist, sends them immediately - if not, it stops for a short amount of time before re-checking the queue.
- If the outbound message fails to see a valid 2XX response within the timeout period (10 seconds), Janus retries it as is (i.e. it does NOT rebuild the message), and continues to retry it so long as it continues to fail up to X times (~10 is probably fair... ) *Note: after each failing try, Janus should wait a short amount of time, this way if Janus sees 10 failing gateway errors in ~3 seconds, we don't necessarily conclude that the backend is down.... there should be a time component, not just retry number to this logic*
- If the outbound message eventually succeeds, Janus checks for any enqueued events, and if some exist, sends them immediately - if not, it stops for a short amount of time before re-checking the queue (it might be smart here to have an upper bound on how many events we can send in each request, so that a 1 min outage is not followed by a message that is *very* large - perhaps 50-100 events/message is the batching upper bound. Obviously the number must be high enough that we know that even with all events enabled, we're able to work through a large queue)
- If the outbound message never succeeds over ~10x retries over the course of ~1-2 mins, assume that the backend is down. Delete the events that we were trying to send, and empty the queue. This ensures that for very long outages or situations where files are misconfigured, we don't have a permanently increasing internal queue of pending events. At this point, Janus can wait a short amount of time and begin checking the queue, attempting to send the events, etc....

Basically Janus has no idea of how "down" / "broken" the backend is. The message send logic is just centered around trying to keep the outgoing event queue at 0.

lminiero:
Added a simple retransmission mechanism that is configurable. You can choose:

1. how many times the last event (or bunch of events) should be retransmitted;
2. the exponential backoff step (in milliseconds) to retry again.

By default it's 5 and 100ms, so retransmissions in case of errors are done after 100ms (1st time), 200ms (2nd time), 400ms (3rd time) and so on. Of course, in case a transmission or retransmission succeeds, this mechanism is reset. The fact that it's configurable should allow people to adapt it to their needs.

Anything more complex than that can be done in custom plugins.

joshdickson40:
Haven't gotten a chance to look at this yet but thank you very much for looking at the retry mechanism. I will try to pull this into staging in the next day or two.

joshdickson40:
We've been running this last build with event retransmission and everything has looked good on our end. Just wanted to report back that all looks OK! This PR has made life infinitely easier for how I like to manage these calling instances. Perhaps we can get this merged back into mainline soon :) Great work, Lorenzo.

lminiero:
@joshdickson40 tnx for the feedback! I'm abroad for the IETF and then will be away for some more days after that, but if everything's still fine I'll merge when I get back to the office, then.

mirkobrankovic:
Just started testing this branch.

Just as a quick thought, is there a need to enable plugin but also to enable it in main janus.cfg config file?
*one should be enough :)

lminiero:
Yes there actually is. Core and plugins are different things: janus.cfg drives whether or not notifications are globally enabled, while each plugin can then be configured to send notifications or not. If you only left janus.cfg, you could end up with plugins spamming the core with info nobody needs. If you only left the configuration in plugins, there would be no way to control whether or not the core should notify, as the core has no view of plugin internals.

mirkobrankovic:
aaah yes, i see the point, since I was only comparing janus.eventhandler.cfg and janus.cfg.

I see now the sampleevh is just an http post example how to get them from core/A to B.

lminiero:
Just FYI, I plan to merge this either tomorrow or the day after that, so if there's anything that you feel should be fixed before please shout! This will also result in the current master to be tagged as v.0.2.1

joshdickson40:
I have not yet pulled 876304f and 3fb6eb0 into our code base, are those low risk enough that they shouldn't affect the stability of this PR relative to previous commits? I have been running a custom build based on 55f6c13 and have nothing but positive things to say 👍

lminiero:
They aggregate multiple pull requests and features so I can't say for sure, but it's all relatively minor changes anyway: support for RTP extensions in a couple of plugins (merged today), better support for RTSP restreaming (authentication, timeout management), "smarter" checks in configure, and a few fixes here and there. None of this should impact any existing application.

Thanks for testing this branch for so long by the way!

joshdickson40:
Oh no problem, it made API design here so much easier, and again, thanks for taking a look at the retry mechanism so I could get this out into production usage. I can probably get this pushed out tomorrow for production usage. If you'd like I could report back with a final sanity check that this all is working okay on my end. If you're not too worried about the minor changes since then (I've quickly browsed the PRs) feel free not to wait for that. Either way, great work on this.

lminiero:
If you're willing to give the final version a go merging can wait: more testing can only help! In a couple of days it's going to be a holiday in Italy anyway, so I can merge this (if everything's ok) when I then get back to the office on Monday.

joshdickson40:
Done, will report back with info when I have it.

lminiero:
@joshdickson40 any (good) news for us? :smile:

joshdickson40:
Ah I'm sorry, I thought I had already replied to this. Yes, I pulled this latest build along with the updated janus.js and adapter.js components and everything seems fine. LGTM. 

On Mon, Dec 12, 2016 at 05:39 Lorenzo Miniero

<
mailto:Lorenzo Miniero <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

https://github.com/joshdickson40
any (good) news for us?

😄

—

You are receiving this because you were mentioned.

Reply to this email directly,
https://github.com/meetecho/janus-gateway/pull/536#issuecomment-266395807
, or

lminiero:
Thanks! I'll work on a couple more pending fixes and possibly later today I'll merge then :+1:

lminiero:
Merging right now. For those that don't want to use this new version yet, I've tagged the latest version as v0.2.1: https://github.com/meetecho/janus-gateway/tree/v0.2.1

soulfly:
Are there any versions changelogs?

lminiero:
No, sorry, the easiest way is to look at the commits or at the discussions here. Anyway, I'm writing a post on meetecho-janus that summarizes the changes and how people can use the new feature.

lminiero:
https://groups.google.com/forum/#!topic/meetecho-janus/oMO32588sMs

brabo-hi:
Hi,
i have the rabbitMq event handler plugin activated.
However all events are being received to the queue "**from-janus**" instead of "**janus-events**".
How can i fix that ?

PR #543, opened 2016-05-25
Don't warn in response to a "detached" event

andreasg123:
When `plugin.detach()` is called from JavaScript, a "detach" request is sent to the server and then the plugin is deleted from `pluginHandles`.  In response, the server sends a "detached" event but there is no plugin to receive it.  While the warning in the console causes no harm, if gives the impression that something went wrong.

I only observed this with the WebSockets transport but I believe that it would happen with the HTTP transport, too.


lminiero:
Fair point, merging... thanks!

Issue #540, opened 2016-05-24
json_pack would make JSON object creation more compact and readable


andreasg123:
[json_pack](http://jansson.readthedocs.io/en/2.5/apiref.html#json_pack) offers compact object creation.  Before:

```
json_t *event = json_object();
json_object_set_new(event, "videoroom", json_string("event"));
json_object_set_new(event, "room", json_integer(participant->room->room_id));
json_object_set_new(event, "leaving", json_integer(participant->user_id));
```

After:

```
json_t *event = json_pack("{sssisi}, "videoroom", "event", "room", participant->room->room_id, "leaving", participant->user_id);
```

It would be nice if JSON object creation could use `json_pack` where appropriate.

lminiero:
That may be just me, but I actually find it less readable! :stuck_out_tongue_winking_eye: 
Anyway, I didn't know about it and it looks like a cool feature we can have a look at for the future, assuming it doesn't break backwards compatibility for those with too old Jansson library versions.

saghul:
For that little JSON object maybe, try the ones with an SDP inside!

It is, however a good idea to use this in production, to make the packets smaller. Maybe a compile time switch?

lminiero:
Actually I don't think this makes the packets any smaller, right? AFAIU, it's just a different, more compact way of filling a json_t object. If size is an issue, you can make packets smaller by removing the `JSON_INDENT(3)` we have when getting a string out of that.

saghul:
Well, spaces and newlines do count! Not a big issue though.

lminiero:
It's a good idea, though. I might add a new configuration setting (e.g., prettyjson or something like that) that defines the number to have in `JSON_INDENT` (0 disables it), and/or involves `JSON_COMPACT` to push it even further.

saghul:
👍

andreasg123:
Yes, I like the idea of having a configuration setting to remove indentation.  Otherwise, I would be in favor of just removing the indentation because one can always look at the parsed JSON in the browser debugger.

Regarding the readability of `json_pack`, one can still define the object in multiple lines:

	json_t *event = json_pack("{sssisi},
		"videoroom", "event",
		"room", participant->room->room_id,
		"leaving", participant->user_id);

`json_pack` was added in [version 2.0](https://jansson.readthedocs.io/en/2.0/changes.html).

lminiero:
[...]

PR #544, opened 2016-05-26
Optimization of core-to-plugin communication

lminiero:
This pull request is another big change, mostly because it changes some of the APIs plugins and core use to communicate with each other.

Specifically, the way core-to-plugin communication works right now is the following:

1. transport reveives a JSON string and parses it to a `json_t` object;
2. transport sends the `json_t` object to the core via `incoming_message`;
3. if it's meant for a plugin, the core serializes the json_t object of the body to a string, and passes it to the plugin via `handle_message`;
4. the plugin receives the string, parses it to a `json_t` object again, and then handles it, synchronously or asynchronously;
5. when a response/event is ready as a `json_t` object, the plugin serializes it to a string, and passes it to the core as either a response via `janus_plugin_result_new` or an event via `push_event`, with or without an SDP string;
6. the core parses the string to a `json_t` object, and uses it (and the SDP, if any) to prepare a Janus API message to send back to the user;
7. this edited `json_t` is passed to the transport, which serializes it to a string and sends it on the wire.

As you can see, while this works (and has worked so far) nicely, it is not very optimized, considering we're parsing/serializing `json_t` objects internally when there's actually no need at all to do that. It would be much easier, and faster, to just route `json_t` objects internally between points 3. and 6., and leave parsing and serialization exclusively to transport plugins. Core and plugins should never have to worry about that (unless they're involving JSON for things of their own, unrelated to the API itself).

This PR tries to address this very issue, and as you can see changes the way the core can pass messages the plugins, and the way plugins can send stuff back to the core (`query_session` included). More specifically, now plugins receive from and pass to the core a `json_t` object directly. When a negotiation is involved, another `json_t` object representing the JSEP data (type and sdp) is exchanged between core and plugins as well. Proper management of `json_t` objects and their reference counters is done in order to avoid leaks or access to broken data: more specifically, it's the plugin's responsibility to `json_decref` any `json_t` object it sends or receives (except for responses created with `janus_plugin_result_new`: those are decremented in the related destroy method by the core). Some additional cleanup is done as well as part of the patch, but nothing major or fancy: the bulk of the work is what I mentioned. As a side note, the patch also uses here and there the nice `json_pack` method suggested in #540, something that should make @andreasg123 happy :smile: 

In terms of actual benefit from this effort, I haven't done extensive testing, but I did a few quick measurements on the EchoTest when pushing an answer back, and I did notice this new approach got me a ~25% faster response. This may not seem a lot, but if we look at chatty plugins where a lot of communication happens, this may have a significant impact on making Janus even lighter and faster than it currently is.

That said, please have a look and let me know if you feel this is helpful. We can't merge right away as this impacts existing plugins quite heavily, requiring some changes here and there: nothing drastic, as you'll see from the changes to the stock plugins, but still something that needs to be addressed. That also explains why the patch changes the version of Janus to 0.2.0, and bumps the plugin API version as well.

Feedback welcome!


janus.c
```diff
@@ -1231,21 +1233,8 @@ int janus_process_incoming_request(janus_request *request) {
...
-			}
-			if(!json_is_object(event)) {
-				JANUS_LOG(LOG_ERR, "[%"SCNu64"] Cannot send response from plugin (JSON error: not an object)\n", handle->handle_id);
-				json_decref(event);
-				ret = janus_process_error(request, session_id, transaction_text, JANUS_ERROR_PLUGIN_MESSAGE, "Plugin returned an invalid JSON response");
-				janus_plugin_result_destroy(result);
-				goto jsondone;
-			}
+			/* Reference the content, as destroying the result instance will decref it */
+			json_incref(result->content);
```
andreasg123: Shouldn't this still test if `result->content` is a JSON object?

lminiero: You mean, as opposed to an array for instance? Good point, I'll add it back, thanks!

janus.c
```diff
@@ -1231,21 +1233,14 @@ int janus_process_incoming_request(janus_request *request) {
...
-			if(!event) {
-				JANUS_LOG(LOG_ERR, "[%"SCNu64"] Cannot send response from plugin (JSON error: on line %d: %s)\n", handle->handle_id, error.line, error.text);
-				ret = janus_process_error(request, session_id, transaction_text, JANUS_ERROR_PLUGIN_MESSAGE, "Plugin returned an invalid JSON response");
-				janus_plugin_result_destroy(result);
-				goto jsondone;
-			}
-			if(!json_is_object(event)) {
-				JANUS_LOG(LOG_ERR, "[%"SCNu64"] Cannot send response from plugin (JSON error: not an object)\n", handle->handle_id);
-				json_decref(event);
+			if(!json_is_object(result->content)) {
```
andreasg123: The body of this `if` statement is almost the same as for the previous `if` statement.  This could be combined:

	if(result->content == NULL || !json_is_object(result->content)) {
		/* Missing content or not a JSON object */
		ret = janus_process_error(request, session_id, transaction_text, JANUS_ERROR_PLUGIN_MESSAGE, result->content == NULL ? "Plugin didn't provide any content for this synchronous response" : "Plugin returned an invalid JSON response");
		janus_plugin_result_destroy(result);
		goto jsondone;
	}

janus.c
```diff
@@ -1263,14 +1255,15 @@ int janus_process_incoming_request(janus_request *request) {
...
+				json_object_set_new(reply, "hint", json_string(result->text));
 			json_object_set_new(reply, "transaction", json_string(transaction_text));
 			/* Send the success reply */
 			ret = janus_process_success(request, reply);
 		} else {
 			/* Something went horribly wrong! */
-			ret = janus_process_error(request, session_id, transaction_text, JANUS_ERROR_PLUGIN_MESSAGE, "%s", result->content ? g_strdup(result->content) : "Plugin returned a severe (unknown) error");
+			const char *error_text = result->text ? result->text : "Plugin returned a severe (unknown) error";
+			ret = janus_process_error(request, session_id, transaction_text, JANUS_ERROR_PLUGIN_MESSAGE, "%s", error_text);
 			janus_plugin_result_destroy(result);
```
andreasg123: It would be better to use `janus_process_error_string` here.

janus.c
```diff
@@ -2479,22 +2470,18 @@ int janus_plugin_push_event(janus_plugin_session *plugin_session, janus_plugin *
...
-	}
-	if(!json_is_object(plugin_event)) {
+	/* Make sure this is a JSON object */
+	if(!json_is_object(message)) {
 		JANUS_LOG(LOG_ERR, "[%"SCNu64"] Cannot push event (JSON error: not an object)\n", ice_handle->handle_id);
 		return JANUS_ERROR_INVALID_JSON_OBJECT;
 	}
 	/* Attach JSEP if possible? */
-	json_t *jsep = NULL;
+	const char *sdp_type = json_string_value(json_object_get(jsep, "type"));
```
andreasg123: There should be a check if `jsep` is `NULL` or not an object.

plugins/janus_audiobridge.c
```diff
@@ -1474,36 +1421,28 @@ struct janus_plugin_result *janus_audiobridge_handle_message(janus_plugin_sessio
...
 			/* Prepare JSON error event */
 			json_t *event = json_object();
 			json_object_set_new(event, "audiobridge", json_string("event"));
 			json_object_set_new(event, "error_code", json_integer(error_code));
 			json_object_set_new(event, "error", json_string(error_cause));
-			char *event_text = json_dumps(event, JSON_INDENT(3) | JSON_PRESERVE_ORDER);
-			json_decref(event);
-			janus_plugin_result *result = janus_plugin_result_new(JANUS_PLUGIN_OK, event_text);
-			g_free(event_text);
+			janus_plugin_result *result = janus_plugin_result_new(JANUS_PLUGIN_OK, NULL, event);
```
andreasg123: If you change the order of the `plugin_response` and `error` labels, you can set `response=event` and fall through from the `error` to the `plugin_response` label.  The cleanup is the same for both cases.

andreasg123:
I commented on the order of labels in the Audiobridge plugin.  In general, I would change that function in all plugins as follows and replace `goto error` with `goto plugin_response`.

Before:

	plugin_response:
	{
		if (!response) {
			error_code = JANUS_VIDEOROOM_ERROR_UNKNOWN_ERROR;
			g_snprintf(error_cause, 512, "Invalid response");
			goto error;
		}
		if(root != NULL)
			json_decref(root);
		if(jsep != NULL)
			json_decref(jsep);
		g_free(transaction);

		janus_plugin_result *result = janus_plugin_result_new(JANUS_PLUGIN_OK, NULL, response);
		return result;
	}

	error:
	{
		if(root != NULL)
			json_decref(root);
		if(jsep != NULL)
			json_decref(jsep);
		g_free(transaction);

		/* Prepare JSON error event */
		json_object_set_new(event, "videoroom", json_string("event"));
		json_object_set_new(event, "error_code", json_integer(error_code));
		json_object_set_new(event, "error", json_string(error_cause));
		janus_plugin_result *result = janus_plugin_result_new(JANUS_PLUGIN_OK, NULL, event);
		return result;
	}

After:

	plugin_response:
	{
		if(error_code == 0 && !response) {
			error_code = JANUS_VIDEOROOM_ERROR_UNKNOWN_ERROR;
			g_snprintf(error_cause, 512, "Invalid response");
		}
		if(error_code != 0) {
			/* Prepare JSON error event */
			json_object_set_new(event, "videoroom", json_string("event"));
			json_object_set_new(event, "error_code", json_integer(error_code));
			json_object_set_new(event, "error", json_string(error_cause));
			response = event;
		}
		if(root != NULL)
			json_decref(root);
		if(jsep != NULL)
			json_decref(jsep);
		g_free(transaction);
		return janus_plugin_result_new(JANUS_PLUGIN_OK, NULL, response);
	}

janus.c
lminiero: Ack, will do that tomorrow.

janus.c
lminiero: Yep, the assignment was faster that way.

lminiero:
Makes sense and is more concise, thanks, I'll modify those as part of this PR tomorrow.

ploxiln:
I like the idea, nice.

lminiero:
Since this is going to be a change in the Janus version, I took advantage of that to also change all the `json_integer(1 or 0)` and `json_string("true" or "false")` to actually use `json_true()` and `json_false()`, which is something all versions of Jansson support. In fact, the main reason I resorted to those tricks was because of the lack of global support for `json_boolean()`.

This ended up mainly affecting some of the admin requests (e.g., to enable/disable some features dynamically, or handle info responses), while not much in the plugins (e.g., AudioBridge now has muted as a boolean instead of a string, which was silly), so it shouldn't be hard to address in existing deployments, if needed at all.

lminiero:
Just FYI, I think I'll merge this soon, no later than next week. Since this changes the way plugins interact with the core there will obviously be a version (and API version) change.

lminiero:
Merged.

PR #547, opened 2016-05-26
Add calls to janus_videoroom_message_free

andreasg123:
When reviewing your new PR, I noticed missing calls to `janus_videoroom_message_free`.

It might be easier to put the call to `janus_videoroom_message_free` at the beginning of the loop and right after the loop:

	while(g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping)) {
		janus_videoroom_message_free(msg);
		msg = g_async_queue_pop(messages);
		...
	}
	janus_videoroom_message_free(msg);

That would avoid the need to call it before every `continue` statement.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -3159,6 +3159,7 @@ static void *janus_videoroom_handler(void *data) {
 					g_snprintf(error_cause, 512, "Error subscribing!");
 					goto error;
 				}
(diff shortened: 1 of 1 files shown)


lminiero:
Good catch: I guess we never found out becuse it's for the plan-b stuff nobody uses (and that I should really get rid of, as it's broken). Good idea about moving the free stuff as well, I'll think about that for the optimization PR.

Issue #390, opened 2015-11-23
VideoTrack mute / unmute delay


leonuh:
Hi there,

The stream is delayed on the "receiver side" after disable and enable the videoTrack.
The delay has gone after page refresh (rejoin)
To test:
Change: getAudioTracks to getVideoTracks in https://github.com/meetecho/janus-gateway/blob/master/html/videoroomtest.js#L350

Any idea?

lminiero:
Nothing we can do, that's a browser thing. When a track is muted, silence/black video is sent, and when unmuted the stream is restored: as such, something is always flowing. Nothing Janus has possibility to have effect on, since as far as Janus is concerned the media never stopped flowing.

Closing as not an issue.

PS: you may want to look into the new mute-related methods in janus.js as they make things simpler than having to manipulate tracks directly: https://github.com/meetecho/janus-gateway/commit/59d9ba8521c51140630a6a7fe3dad8aa2b73b784

leonuh:
Thank you for your answer. I am using these methods already.

Just to avoid any misunderstanding:
The audio keeps the synchronicity. The video is not in sync with the audio after it was restored. It is not a problem for the audio regardless of the mute / unmute.

Do you have any suggestion to trigger a kind of reset for the stream (eg. a new stream instance after mute / unmute for the remote users)?

lminiero:
Just checked with apprtc.appspot.com, which is peer-to-peer, and there the problem doesn't happen... just open a session among two peers, and on the JavaScript console of one of them use this command to mute/unmute the video:

    document.getElementById('mini-video')._srcObject.getVideoTracks()[0].enabled = false; // mute
    document.getElementById('mini-video')._srcObject.getVideoTracks()[0].enabled = true; // unmute

Apparently there there's no delay at all, which means that there's indeed something that is happening withing Janus that is confusing viewers. I'll have to check what's happening.

leonuh:
As a workaround (until the bugfix) it works without any delay:
```
plugin.send({message: { request: 'configure', video: true/false }});
```

lminiero:
This still has the publisher send stuff, though (even though it's discarded by the plugin).

leonuh:
Unfortunately the previous workaround doesn't work either properly cause of the earlier bug: If the video is false then the audio doesn't work for the new feeds. I works only if I toggle the video at least for a moment from the publisher's side. It seems like the audio depends on the video data somehow.

[...]

PR #548, opened 2016-05-27
Max NACK queue in seconds, instead of packets, plus some other RTCP tweaks

lminiero:
While investigating the infamous #390 (which apparently is a problem in Chrome, tracking an issue there), I fixed some things here and there related to RTCP management.

This is only a PR and not a direct commit as I also changed the way the Max NACK queue is handled, though, which means the existing configuration changes meaning in that respect. The way it has worked so far was in terms of packets: e.g., a `MAX_NACK_QUEUE=300` meant we'd store at max 300 packets to be ready to NACK, and then get rid of older ones when newer arrived. This kinda worked, but is not very effective and can actually be counterproductive, considering it doesn't take into account that different media may have a different number of packets per second (e.g., 50 for audio, 30 or more/less for video), which made it hardly configurable for more fine tuning. Besides, the default value of 300 stored way too many packets, possibly leading to excessive buffering and delayed playout on slow links.

This patch changes that, and turns the Max NACK queue value to a timing one: now the value you can configure in the settings is in seconds, and means "only store RTP packets if they're not older than X seconds", no matter how many there are and no matter what media they refer to. The default is one second, which should be reasonable.

Played with it a bit and it seems to be working fine, but feedback is welcome. As soon as I have an ack I'll merge.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3175,8 +3175,8 @@ void *janus_ice_send_thread(void *data) {
 			}
 			video_rtcp_last_sr = now;
 		}
-		/* Should we clean up old NACK buffers? */
-		if(max_nack_queue > 0 && (now-last_nack_cleanup >= (max_nack_queue/4))) {
+		/* Should we clean up old NACK buffers? (we check each 1/4 of the max_nack_queue time) */
+		if(max_nack_queue > 0 && (now-last_nack_cleanup >= (max_nack_queue*250))) {
 			/* Check if we do for both streams */
 			janus_cleanup_nack_buffer(now, handle->audio_stream);
 			janus_cleanup_nack_buffer(now, handle->video_stream);
--- a/ice.c
+++ b/ice.c
@@ -280,18 +280,43 @@ typedef struct janus_ice_queued_packet {
 static janus_ice_queued_packet janus_ice_dtls_alert;
 
 
-/* Maximum value, in seconds, for the NACK queue/retransmissions (default=1s) */
-#define DEFAULT_MAX_NACK_QUEUE	1
+/* Maximum value, in milliseconds, for the NACK queue/retransmissions (default=1000ms=1s) */
+#define DEFAULT_MAX_NACK_QUEUE	1000
 /* Maximum ignore count after retransmission (100ms) */
 #define MAX_NACK_IGNORE			100000
 
 static uint max_nack_queue = DEFAULT_MAX_NACK_QUEUE;
 void janus_set_max_nack_queue(uint mnq) {
 	max_nack_queue = mnq;
-	JANUS_LOG(LOG_VERB, "Setting max NACK queue to %ds\n", max_nack_queue);
+	if(max_nack_queue == 0)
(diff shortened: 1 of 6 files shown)


lminiero:
Mentioning @ploxiln in particular has he contributed a lot to NACK management in the code.

ploxiln:
I think this is a good idea.

Second granularity is a bit coarse when the typical value is 1 second, I suggest milliseconds (default can be the same, but 1000 milliseconds).

ice.c
```diff
@@ -3145,6 +3150,48 @@ void *janus_ice_send_thread(void *data) {
...
+						g_free(p);
+						first = g_list_first(component->retransmit_buffer);
+						p = (janus_rtp_packet *)(first ? first->data : NULL);
+					}
+				}
+				janus_mutex_unlock(&component->mutex);
+			}
+			if(handle->video_stream && handle->video_stream->rtp_component) {
+				janus_ice_component *component = handle->video_stream->rtp_component;
+				janus_mutex_lock(&component->mutex);
```
ploxiln: Notice that starting from this line, the block of code til the end of this scope is identical to the block above. You can make a function for this, used something like this:

```c
janus_cleanup_nack_buffer(handle->video_stream->rtp_component);
```

ice.c
```diff
@@ -3145,6 +3150,48 @@ void *janus_ice_send_thread(void *data) {
 			}
 			video_rtcp_last_sr = now;
 		}
+		/* Should we clean up old NACK buffers? */
+		if(now-last_nack_cleanup >= 500000) {
```
ploxiln: I think you should base the cleanup frequency on `max_nack_queue`, I think a period of one quarter the `max_nack_queue_time` would be reasonable.

lminiero:
Good points, I'll make the change ASAP, thanks for looking into this!

lminiero:
Just made the changes you suggested, but added a minimum acceptable value to 200ms anyway: first, because it would checking stuff too often, and second, less than that it really doesn't make any sense if we think of round trip times and the like. A value of 0 is still acceptable to disable NACK support entirely (which means we don't store packets).

ice.c
```diff
@@ -3145,6 +3175,13 @@ void *janus_ice_send_thread(void *data) {
 			}
 			video_rtcp_last_sr = now;
 		}
+		/* Should we clean up old NACK buffers? */
+		if(max_nack_queue > 0 && (now-last_nack_cleanup >= (max_nack_queue/4))) {
```
ploxiln: I think you need to multiply `max_nack_queue` by 1000 when comparing to `now` etc, like you do in `janus_cleanup_nack_buffer()`

ploxiln:
looks good to me

ice.c
lminiero: You're right, I missed that... I'll fix that tomorrow in the morning.

lminiero: Rather than doing a `*1000/4`, I did a `*250` directly.

lminiero:
Merging.

PR #549, opened 2016-05-27
getUserMedia belongs to navigator.mediaDevices, not to window

mabu-github:
References:
* https://developer.mozilla.org/en/docs/Web/API/Navigator/getUserMedia
* https://github.com/webrtc/adapter/blob/v1.1.0/src/js/firefox/getusermedia.js

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -162,7 +162,7 @@ Janus.init = function(options) {
 				}
 			}
 			if(src === 'adapter.js') {
-				if(window.getUserMedia && window.RTCPeerConnection) {
+				if(navigator.mediaDevices.getUserMedia && window.RTCPeerConnection) {
 					// Already loaded
 					done();
 					return;
@@ -185,7 +185,7 @@ Janus.init = function(options) {
 
 // Helper method to check whether WebRTC is supported by this browser
 Janus.isWebrtcSupported = function() {
-	return window.RTCPeerConnection && window.getUserMedia;
+	return window.RTCPeerConnection && navigator.mediaDevices.getUserMedia;
 };
 
 function Janus(gatewayCallbacks) {
--- a/html/janus.nojquery.js
+++ b/html/janus.nojquery.js
@@ -209,7 +209,7 @@ Janus.init = function(options) {
 		}
 		function addJs(src,done) {
 			if(src === 'adapter.js') {
-				if(window.getUserMedia && window.RTCPeerConnection) {
+				if(navigator.mediaDevices.getUserMedia && window.RTCPeerConnection) {
 					// Already loaded
 					done();
 					return;
@@ -232,7 +232,7 @@ Janus.init = function(options) {
 
 // Helper method to check whether WebRTC is supported by this browser
 Janus.isWebrtcSupported = function() {
-	return window.RTCPeerConnection && window.getUserMedia;
+	return window.RTCPeerConnection && navigator.mediaDevices.getUserMedia;
 };
 
 


mabu-github:
adapter.js would always be loaded when it is already provided because window.getUserMedia does not exist

andreasg123:
`navigator.mediaDevices.getUserMedia` returns a promise and `navigator.getUserMedia` doesn't.  One cannot just replace one with the other.  In my own code, I handle it like this:

	var success_callback = function() {
		// success action
	};
	var error_callback = function() {
		// error action
	};
	var constraints = {};
	if (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
		navigator.mediaDevices.getUserMedia(constraints).then(success_callback, error_callback);
	}
	else {
		var getUserMedia = (navigator.getUserMedia || navigator.mozGetUserMedia || navigator.webkitGetUserMedia);
		getUserMedia.call(navigator, constraints, success_callback, error_callback);
	}

I agree that one should use the modern, standard-compliant version but only if it doesn't break anything.

mabu-github:
The current implementation using window.getUserMedia is definitively broken. It just never happens in the demo because adapter.js is not already provided.

In fact all the code is in if conditions being capability checks. Thus being defined is sufficient and implementation does not matter.

lminiero:
We use whatever the `adapter.js` we ship exposes, and currently it does a `navigator.getUserMedia = getUserMedia;`, where `getUserMedia` is not defined as Promises based. As such, I agree we need to update the check we have, but it should look like this:

`if(navigator.getUserMedia && window.RTCPeerConnection) {`

If you're not using our own method to access devices (that is, you pass your own MediaStream to createOffer/createAnswer in janus.js, which you can do), you can do your own checks to make sure stuff works the way you need.

lminiero:
Actually, correcting myself, we DO use `navigator.mediaDevices.getUserMedia` already in the code... happened in some PR some time ago, not sure when as I have too many things to keep track of :smile:

andreasg123:
I would suggest just to drop the test and "adapter.js".  By now, browsers that support WebRTC do it the correct way.  There isn't anything that one can do for Safari, IE, or Edge with the adapter.

`navigator.getUserMedia` [will be dropped](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getUserMedia) so that there isn't much point to test for that.

I agree with Lorenzo that one should get the media stream outside of Janus and pass it in.  That's what I do.  :smirk:

lminiero:
Well, `adapter.js` is actually quite helpful in some scenarios, as we do use some helpers defined in there. Besides, all the Edge support is in there (although I'm wondering if it still works, by now, as I haven't tested that in a while).

When I mentioned `navigator.getUserMedia`, I meant that that's what `adapter.js` creates for us.

Finally, getting the MediaStream outside is an option, but I'd rather not make it the only way. A lot of people like to just let `janus.js` worry about that (me included in the demos :wink: ).

mabu-github:
Please have a look at the old code:
`if(window.getUserMedia && window.RTCPeerConnection) {`

getUserMedia is **never defined on window**. It is either defined on navigator or navigator.mediaDevices. Thus it is a constant condition: false.

This is why you always load adapter.js in the same directory, even when one provides his own adapter js, let's say v1.10 from somewhere else.

lminiero:
Yeah, that's why I meant we should change the check to `if(navigator.getUserMedia && window.RTCPeerConnection) {` instead (s/window/navigator).

mabu-github:
Hm, above you mentioned that the code is already using navigator.mediaDevices.getUserMedia

e.g. in 
`function getScreenMedia(constraints, gsmCallback) {
						Janus.log("Adding media constraint (screen capture)");
						Janus.debug(constraints);
						navigator.mediaDevices.getUserMedia(constraints)
							.then(function(stream) { gsmCallback(null, stream); })
							.catch(function(error) { pluginHandle.consentDialog(false); gsmCallback(error); });
					};`

As **adapter.js already shims navigator.mediaDevices.getUserMedia** and the support for **navigator.getUserMedia will be dropped** in the future, I think it would be better to stick with navigator.mediaDevices.getUserMedia.

Excerpt from adapter.js:
`// Shim for mediaDevices on older versions.
  if (!navigator.mediaDevices) {
    navigator.mediaDevices = {getUserMedia: requestUserMedia,
      addEventListener: function() { },
      removeEventListener: function() { }
    };
  }
// Returns the result of getUserMedia as a Promise.
function requestUserMedia(constraints) {
  return new Promise(function(resolve, reject) {
    getUserMedia(constraints, resolve, reject);
  });
}`

mabu-github:
I see the problem now. I am using adapter.js v1.1.0 which will not expose variables on the window object anymore (which is totally reasonable). Either an AMD-style define() function is available or it will expose its variables under 'adapter.*'.

Unfortunately I cannot see which adapter.js version was used for janus because adapter.js is just provided in the html sources.

Would you mind adding dependencies into the build process by using a tool like 'bower'? Furthermore I cannot find jasmine tests or something similar for the JS sources. I think it would be a good idea because then the code could be testet against newer versions of adapter.js and so it would be possible to use newer versions without any risk.

I could offer to start work on this. But then it would be great to have a short Skype session before, because putting a lot of work into it without knowing if it will get integrated into the software is not a good idea.

This PR should probably closed as won't fix. Nevertheless it would be good to know which version of adapter.js you are using.

lminiero:
The adapter.js we use in the demos is available in the [HTML folder](https://github.com/meetecho/janus-gateway/blob/master/html/adapter.js). It's probably not the latest but it's not ancient.

As to bower, sorry but I've said multiple times in other issues and PRs as well that I don't want to be constrained to a specific technology or framework for the demos on the client side. We have a sample, simple, library people can use if they like (janus.js or janus.nojquery.js) but that's about it, everybody is free to do what they want on the client side. Nolan for instance recently contributed a [different library](https://github.com/ndarilek/node-janus) people can use to do the same thing, using a different technology. Should you be interested in working on something like this, we'd definitely welcome the contribution! It doesn't necessarily have to be part of our repo, it could very well be an external effort that we can link along the other [resources](https://janus.conf.meetecho.com/docs/resources) we're aware of.

That said, we do have a [bower file](https://github.com/meetecho/janus-gateway/blob/master/bower.json) somebody provided, available in the repo, that could help those interested in it (I think it references a specific version of adapter.js, not sure which one), but personally never used and don't intend to, so I'm not familiar with it.

mabu-github:
Thanks a lot for the hint about the other library and the good discussion! I really appreciate your work on Janus and the effort you put in open sourcing and maintaining it. 

If you don't mind, I'd like to share some opinions:

In my opinion external software should be clearly marked as such including a version. Dependency management solutions pull in all necessary deps at build time. And JavaScript should also be tested.
A common workflow is to "build" js files and produce variants like minified, with and without external dependencies. Multiple files can also be combined into one for inclusion in the browser. Then there are automated test frameworks that assert that the software is correct (this gives a lot of confidence for refactoring).

When I first looked at the streaming demo js, it honestly took me a while (to refactor) to see what's going on because display logic is mixed with the janus interaction and so on.
Personally, I like short concise functions that do one thing and are well testable because often test expectations serve as a kind of documentation.

lminiero:
I agree on most of what you say, but the demos are really just meant to be a simple example of what you can do with Janus. I don't expect people to just modify the demos as they are, as I'm quite a crappy JavaScript developer (I guess that's clear :smile: ) and the demos are ostensibly ugly and not very "modern".

That said, I'd rather not go for any of the new fancy stuff there either, as 1. it might drive inexperienced people away (me being one of them: if I have to learn a new framework just to try or learn something that is related to a project I shy away), and 2. that might lead people into thinking that's the best or only way to go to use Janus, which I strongly want them not to think. The idea is that you can do whatever you want on the client side, with or without janus.js, and basic JavaScript is proof of that.

Of course, we'll have to make sure janus.js doesn't use or refer to the wrong stuff, thus leading to broken demos in general, so the checks you mention will have to be fixed anyway. Not sure if we can just get rid of adapter.js entirely as Andreas suggested, as, while we may redefine the few methods we use (which at the moment may be limited to attachMediaStream and maybe a couple others?), it still has very useful stuff to take care of ORTC stuff for Edge. I'll have to check if Edge still works after we started using `navigator.mediaDevices.getUserMedia`: if not, I'll have to add a condition in janus.js to have it use the `adapter.js` version in case Edge is involved.

lminiero:
Anyway, I don't want to give the impression I don't care about the stuff. I obviously do, but in my opinion that's all orthogonal to Janus itself. I'm mostly focused on the server side stuff and on the Janus internals, the JavaScript code and the demos just being what I use to play with it and see if and how it works. The reason for them being so basic is also related to the mantaining aspect you mentioned before: if I had to learn tons of new frameworks for testing, building and using the JavaScript code I'd go crazy, I'd go even crazier any time I had to change/add something, and that would take precious time away from focusing on what I care about the most, that is Janus itself.

That's why I also encourage people to work on new ways to talk to Janus, Nolan's library being one notable example. Should you decide to start doing something similar using the tools you're most familiar with, in order to have something you're more comfortable with to interact with Janus, that would be very welcome and I'd be happy to point users to it. The more variety, the better!

PR #558, opened 2016-06-09
Support for recording data channel text messages

lminiero:
This PR extends the recording functionality in Janus, allowing you to also record data channel data, and then post-process the recording to an external file as you do with audio and video. Not sure if this is something that can be of interest, but it was not a big effort and so it seemed reasonable to have it there.

Since we currently only support text data, that's what we do record. The internal recorder adds timing related info to the data (`janus_get_real_time()`) that you can use in conjunction with the timing info in the header to figure out when exactly the data "appeared".

As to the post-processing, just for fun I created a `.srt` converter, which basically generates a subtitle file you can use with most players. I tested this with the EchoTest plugin (the only plugin where I added the data channel recording functionality) and it seems to be working fine: the subtitles appeared exactly when they were meant to, when played along the processed audio+video. Of course, different export formats can be envisaged for the future, especially for scenarios/contexts where a `.srt` file doesn't make much sense.

One noticeable thing that needs to be improved is how to handle large buffers. In the post-processor we usually dump whatever's larger than 1500 bytes, as that will typically never happen for RTP. Data channels don't have the same constraint, something that needs to be addressed in the code.

Feedback?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -313,6 +313,8 @@ janus_pp_rec_SOURCES = \
 	postprocessing/pp-opus.h \
 	postprocessing/pp-opus-silence.h \
 	postprocessing/pp-rtp.h \
+	postprocessing/pp-srt.c \
+	postprocessing/pp-srt.h \
 	postprocessing/pp-webm.c \
 	postprocessing/pp-webm.h \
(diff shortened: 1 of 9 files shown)


lminiero:
Merging.

PR #560, opened 2016-06-10
Added new plugin (and demo) for datachannel based text broadcasting

lminiero:
Hi all,

this PR adds a new plugin that implements a feature a few people have been asking about. This plugin allows you to exchange text-based public and private messages via datachannels within the context of text rooms. More precisely, it works this way:

1. you attach to the plugin;
2. you ask for a PeerConnection (`request:"setup"`);
3. you get an SDP offer from the plugin which only contains data channels;
4. you send your SDP answer (`request:"ack"`);
5. after that, you only send and receive JSON messages through this data channel.

Within the context of the plugin, you can create/destroy/join/leave multiple rooms at the same time, and send/receive public/private messages while in them. You can send messages either to the whole room, to a single participant, or to multiple recipient in the room. Messages addressed to the room itself can also be forwarded to external applications via HTTP POST, if so configured.

The following are a few examples of messages you can send, as how you create a room:

    {
        textroom: "create",
        transaction: "123"
        room: 1234,
        description: "My room"
    }

How you join:

    {
        textroom: "join",
        transaction: "123"
        room: 1234,
        username: "a1b2c3d4",
        display: "Lorenzo Miniero"
    }

Sending a message to the whole room:

    {
        textroom: "message",
        transaction: "123"
        room: 1234,
        text: "Hey there"
    }

Sending a whisper to a single user:

    {
        textroom: "message",
        transaction: "123"
        room: 1234,
        to: "e5f6g7h8",
        text: "Hey there"
    }

Sending a whisper to multiple users:

    {
        textroom: "message",
        transaction: "123"
        room: 1234,
        tos: [ "e5f6g7h8", "i9j0k1l2" ],
        text: "Hey there"
    }

etc., quite simple, you can figure out more by looking at the code or the demo.

The demo itself is basically a very simple chatroom: you join a preconfigured room, can chat publicly, and send/receive whispers. Nothing fancy, but it's functional, and will hopefully provide people with more interesting ideas on how the plugin can be used.

The main idea is of course not to have yet another way of doing chat in a browser (there are better tools for the purpose, although this might be useful for that too in case you want to just use Janus and nothing else), but rather to provide an easy way to bridge data channel content with external applications. Right now you can only pass messages addressed to the room to an HTTP backend, so it's a start, but in the future I'd like to allow bidirectional communication, e.g., allow external applications to inject text addressed to one or more recipients in a room. This should allow for a very easy interaction with applications that are not WebRTC compliant.

Any feedback on this? The idea is to merge this soon, so please play with it and let me know what you feel.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -297,6 +297,16 @@ conf_DATA += conf/janus.plugin.voicemail.cfg.sample
 EXTRA_DIST += conf/janus.plugin.voicemail.cfg.sample
 endif
 
+if ENABLE_PLUGIN_TEXTROOM
+plugin_LTLIBRARIES += plugins/libjanus_textroom.la
+plugins_libjanus_textroom_la_SOURCES = plugins/janus_textroom.c
+plugins_libjanus_textroom_la_CFLAGS = $(plugins_cflags)
+plugins_libjanus_textroom_la_LDFLAGS = $(plugins_ldflags)
+plugins_libjanus_textroom_la_LIBADD = $(plugins_libadd)
+conf_DATA += conf/janus.plugin.textroom.cfg.sample
+EXTRA_DIST += conf/janus.plugin.textroom.cfg.sample
+endif
+
 ##
 # Post-processing
 ##
--- a/conf/janus.plugin.textroom.cfg.sample
+++ b/conf/janus.plugin.textroom.cfg.sample
@@ -1,0 +1,13 @@
+; [<unique room ID>]
+; description = This is my awesome room
+; is_private = yes|no (whether this room should be in the public list, default=yes)
+; secret = <optional password needed for manipulating (e.g. destroying) the room>
+; pin = <optional password needed for joining the room>
+; post = <optional backend to contact via HTTP post for all incoming messages>
+
+[1234]
+description = Demo Room
+; is_private = yes
+secret = adminpwd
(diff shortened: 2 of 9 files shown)


saghul:
😍 👍 🍰

lminiero:
Wow, cake! :smile:

andreasg123:
I'm wondering if this would also be useful in a video room.  I know that it has data channel support but it doesn't have whisper.

saghul:
You can create an application which uses the two plugins for that.

amnonbb:
it will be nice if we can pass role/type (moderator/administrotor/user) of participant or something... :grimacing:

lminiero:
But that would require an authentication of some sorts in the plugin, something I don't plan to add. Everyone's free to use the username they wish, there's no password or stuff like that. I really don't want to make a (worse) clone of Jabber....

Some sort of admin features are already there by means of the `secret` you can specify when creating rooms, so I guess some new requests might exploit that part.

lminiero:
Since this is a separate feature that doesn't mess with existing stuff, I'll merge. Maybe this will encourage more people to test and play with it. More advanced feature can always be added later on.

cacheworks:
Thanks for this,  it's a great plugin example, easier to understand than the ones with video/audio

PR #561, opened 2016-06-13
sip: add ability to override User Agent per account

saghul:
This is useful because it allows us to pass in a custom crafted UA string including the browser version, for example.

/cc @lminiero


lminiero:
:+1:

saghul:
❤️ 🍰

PR #563, opened 2016-06-14
Autogenerate DTLS cert/key if not specified

saghul:
The ultimate goal here is to have Janus working without any configuration. This brings us one step closer!

/cc @lminiero

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,8 +246,8 @@ or on the command line:
 	-P, --plugins-folder=path     Plugins folder (default=./plugins)
 	-C, --config=filename         Configuration file to use
 	-F, --configs-folder=path     Configuration files folder (default=./conf)
-	-c, --cert-pem=filename       HTTPS/DTLS certificate
-	-k, --cert-key=filename       HTTPS/DTLS certificate key
+	-c, --cert-pem=filename       DTLS certificate
+	-k, --cert-key=filename       DTLS certificate key
 	-S, --stun-server=filename    STUN server(:port) to use, if needed (e.g.,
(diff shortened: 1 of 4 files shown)


lminiero:
This looks useful, thanks for working on this!

I'd make it much more explicit when you're generating new certificates, though. Folks may have improperly configured their own certificates and end up using random ones without being aware of it otherwise. Maybe a `LOG_WARN` in the `if (!server_pem && !server_key) {` check that leads to `janus_dtls_generate_keys`?

lminiero:
Or maybe a command line option that says "autogenerate" that will override the check for existing ones? Not sure what the best approach is.

saghul:
I thought of that. My take is the following: if you specify either but not both we give an error. If you give both we load them, and if you give neither, then we generate them.

So the only way to generate them by mistake would be to leave the options commented out.

Alternatively, we cold have a special token `<auto>` which must be set in order to generate them, but it needs to be set for both at the same time. Thoughts?

saghul:
> Maybe a LOG_WARN in the if (!server_pem && !server_key) { check that leads to janus_dtls_generate_keys?

I'd be ok with that to avoid further config complications. Remember my ultimate goal is to have zero configuration 😎

lminiero:
I think a LOG_WARN is enough then!

saghul:
Done!

lminiero:
Cool, tnx! Let me do a quick check and I'll merge :+1:

lminiero:
Looks good, merging!

PR #567, opened 2016-06-20
Fix VideoRoom SDP compose error

jswirl:
A misplaced `break;` in the new `janus_videoroom_sdp_v_format()` function caused H.264 video not being sent.


lminiero:
Good catch, thanks!

jswirl:
Wow, that was fast, thanks 😄

PR #568, opened 2016-06-27
Helper threads for relaying packets to multiple recipients (Streaming plugin)

lminiero:
This PR is a first step at addressing the feedback provided [here](https://groups.google.com/forum/#!topic/meetecho-janus/HR0f3XN-_i8), that is the difficulties the Streaming and VideoRoom plugins have at serving a very high number of viewers for the same source.

The root of the problem is that each source (mountpoint thread for the Streaming plugin, incoming RTP callback for the VideoRoom plugin) currently has a loop that iterates on all viewers, and this loop, while theoretically lightweight (it simply passes the packet to the core, which copies the buffer and adds it to a sync queue for the packet to be sent later, so no I/O involved), still obviously requires a tiny amount of time for each participant (allocation of new buffer + queue lock/unlock). When a lot of viewers are involved, this sum of tiny times can easily surpass the inter-packet arrival times, meaning the source can't keep up anymore (we don't read incoming packets from the network soon and frequently enough) and packets are lost, which leads to several NACKs being sent by all participants at the same time, since they all loose the same stuff.

This PR tries to solve this by adding helper threads in the plugin (Streaming plugin only at the moment: should this work fine, I'll add to the VideoRoom too). More specifically, each of these helper threads is made responsible of a sublist of the source viewers, for which we have a threshold. Anytime all threads have reached their peak (currently 100 viewers), a new thread is allocated. New viewers are always added on the thread that is less loaded. This way, the source will still have a loop, but instead of looping on all participants, it will loop on a limited set of helper threads, who will in turn make sure the media is forwarded to their participants. Since the number of helper threads will typically be quite low, we'll never risk the source can't keep up.

This is not perfect. At the moment, these threads are only destroyed when the mountpoint is. This means that if you got a high peak of viewers that then went away, all the new threads you created will remain there until you remove the mountpoint. This may or may not be desirable. Besides, a helper thread when you know the numbers will be low may actually be a waste of resources, so we may want to consider making this optional/configurable (especially for the VideoRoom) in future commits.

That said, the patch is experimental, so I encourage everybody interested in streaming to a lot of participants to play with it and test intensively. I've verified that the threshold works as expected using a very low value (e.g., 2 instead of 100), but I haven't gone beyond that, since right now I don't have the means nor the time to start stress tests for that.

Feedback welcome.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -294,7 +294,9 @@ static GThread *watchdog;
 static void *janus_streaming_handler(void *data);
 static void *janus_streaming_ondemand_thread(void *data);
 static void *janus_streaming_filesource_thread(void *data);
+static void janus_streaming_queue_rtp_packet(gpointer data, gpointer user_data);
 static void janus_streaming_relay_rtp_packet(gpointer data, gpointer user_data);
+static void *janus_streaming_queue_thread(void *data);
 static void *janus_streaming_relay_thread(void *data);
 static gboolean janus_streaming_is_keyframe(gint codec, char* buffer, int len);
 
@@ -355,6 +357,17 @@ typedef struct janus_streaming_codecs {
 	char *video_fmtp;
 } janus_streaming_codecs;
 
+#define MAX_VIEWERS_PER_RELAY	100
+typedef struct janus_streaming_relay {
+	GAsyncQueue *packets;
+	GList *recipients;
+	GThread *thread;
+	gint64 destroyed;
+	janus_mutex mutex;
+} janus_streaming_relay;
+static GList *old_relays;
+janus_mutex relays_mutex;
+
 typedef struct janus_streaming_mountpoint {
 	gint64 id;
 	char *name;
@@ -369,7 +382,8 @@ typedef struct janus_streaming_mountpoint {
 	void *source;	/* Can differ according to the source type */
 	GDestroyNotify source_destroy;
 	janus_streaming_codecs codecs;
-	GList/*<unowned janus_streaming_session>*/ *listeners;
+	GList *listeners;
+	GList *relays;
 	gint64 destroyed;
 	janus_mutex mutex;
 } janus_streaming_mountpoint;
@@ -456,6 +470,7 @@ typedef struct janus_streaming_rtp_relay_packet {
 	uint32_t timestamp;
 	uint16_t seq_number;
 } janus_streaming_rtp_relay_packet;
+janus_streaming_rtp_relay_packet exit_packet;
 
 
 /* Error codes */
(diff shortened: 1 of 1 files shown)


lminiero:
Any news? I'm starting to wonder whether this is useful at all. I've been doing some tests on my own, and I managed to get 1000 viewers without too many issues, so I'm not sure this two-tier approach is really needed.

trankkilou:
Hi Lorenzo,

Here some updates.

I have made 2 kind of test (for both 20 clients / relay) :

1/ I 'm using webcam (720p@15 fps) with videoroom then RTP forwarding. I'm
using VP8/OPUS as codec in the videoroom plugin . After I'm connecting
client with streaming plugin to stream

2/ I'm using Full hd  mp4 trailer. I reencode to 1080p vp8 @24 fps, 3
Mbits, only video. I'm using ffmpeg to send it to port 5002 via rtp.  After
I'm connecting client with streaming plugin to stream


Regarding the first test :

Packets dropped start around 270 clients. This introduce drop down of  FPS
AND high latency (up to 40 secondes)  ->
https://www.youtube.com/watch?v=tOcCQy1ewzs



Regarding the second test

Video start to stutter around 150 clients and is stuck at 170 clients but
janus is 70% IDLE !  ->https://youtu.be/DG3-V1-0fMc
After about 15 min, janus crash (core dump attached), each time



The second test show that janus is able to deal with 450 Mbits and 65 000
packets / sec. So regarding the first test, network can't be involved.


I don't know you are testing using your "home" clients but maybe it is not
representative.

To sum up 3 issues detcted :
- latency
- packets drops -> stuttering and FPS drop down
- core dump

I think what we are talking here is very important: stability and fairness
-> As long as janus is not 100% CPU and there is no network issue, a
customer who stream shouldn't be bored by others, which is clearly not the
case actually.


So I propose you to focus on the core dump first.

My colleague alexandre will try to help me tonight regarding the "packets
losses" issue

kr,

J-

2016-07-05 10:33 GMT+02:00 Lorenzo Miniero <notifications@github.com>:

> Any news? I'm starting to wonder whether this is useful at all. I've been
> doing some tests on my own, and I managed to get 1000 viewers without too
> many issues, so I'm not sure this two-tier approach is really needed.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/meetecho/janus-gateway/pull/568#issuecomment-230418328>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ARCTYLLzXR7mjQfZnu7KC_77xAdrJjhMks5qShbsgaJpZM4I-_xz>
> .
>



-- 
cdt,

Juju

trankkilou:
Update:

I made some iperf test between my VMs and notice a high packets lost ratio (no losses < 30 mbits, 30% à 50 mbits 70% at 500 Mbits).

So my main troubles is coming from this ATM,  I will have to check with teams how to fix this on their 1 Billions dollars equipment..

cheers

J-

lminiero:
Closing as unrelated to the issue.

PR #570, opened 2016-06-30
Resolves #569 On MacOS X, libraries can be in /opt/local/lib

### LINKED ISSUE #569: Configure script cannot find libsrtp
I'm using MacPorts and I have installed port `srtp`, which provides libsrtp, in version 1.4.4.

However, `./configure` gives me the following error:
```
checking for srtp_init in -lsrtp... no
configure: error: libsrtp not found.
```

Yoric:


ploxiln:
Another option is to mention in the instructions, that you can specify this stuff outside configure, for wherever you installed the libraries:

```
export PKG_CONFIG_PATH="/opt/local/lib/pkgconfig"
export CFLAGS="-I/opt/local/include"
export LDFLAGS="-L/opt/local/lib"
./configure ...
```

this is what I did in my (linux-centric) build script https://gist.github.com/ploxiln/a78867604cde1c6b7e0e

lminiero:
I agree with @ploxiln that it makes more sense to be clearer in the README, rather than adding all possible folders in there. I'll edit it later today.

Yoric:
Well, that is the default folder for MacPorts. I believe that making it build out-of-the-box with MacPorts would be quite useful.

(alright, not entirely out-of-the-box, MacPort's libusrsctp is too old to build, but I guess I'll be submitting a patch to MacPorts when I have time)

lminiero:
Ok, then I'll merge this and also add the info on environment variables for the future later on :+1: 
Thanks!

PR #571, opened 2016-07-04
fix processing vp8 with no extended bit

hasbean:
My gstreamer generated VP8 files with no extended bit for some reason, and this is a problem because the postprocessor skips anything without an xbit.  This pull request amends that and my VP8 files work fine now.  I haven't tried it with other mjr files though, so it might need some testing.

### DIFF
--- a/postprocessing/pp-webm.c
+++ b/postprocessing/pp-webm.c
@@ -4,7 +4,7 @@
  * \brief    Post-processing to generate .webm files
  * \details  Implementation of the post-processing code (based on FFmpeg)
  * needed to generate .webm files out of VP8/VP9 RTP frames.
- * 
+ *
  * \ingroup postprocessing
  * \ref postprocessing
  */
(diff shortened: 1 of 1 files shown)


lminiero:
I'm not sure I can merge this. I use GStreamer a lot and I never experienced this issue. The post-processor definitely does not skip everything without an extension bit: it just skips the first byte if there's no exception, and then moves on to the payload. That check is only needed when control bits are set, which means info on the resolution is provided.

hasbean:
In my case, all the frames have no xbit for some reason.

Here's my mjr file
http://cam.bojacob.net/test.mjr

lminiero:
I guess it's because no key frame was received in the short time the video lasted.

lminiero:
Do you get a meaningful video from your patch?

hasbean:
Seems to happen even with 10 minute long videos.

And yes I'm using it right now with no issues.  However, I don't have any mjr files that actually DO contain an xbit, so I can't test it there.

lminiero:
The sample we provide in the recording folder should.

hasbean:
Played the output in Firefox and saw a guy with long curly black hair.  If that's what I'm supposed to see, then yup it's working :)

lminiero:
My hair's not black but we can't blame the post processor for that! :smile: 
I'll check the patch to see what it does differently from the original code, especially to figure out how it manages to get working video from your sample, and get back to you, thanks.

lminiero:
By the way, could you fix the indentation? You're using spaces, while we use tabs for that. There's an `.editorconfig` file with the settings, if your editor supports that.

lminiero:
I think I understood the issue: the S-bit is checked in the X-bit branch, but it does NOT depend on that. In fact, the S-bit is in the main octed header, which means it can indeed exist by itself, and so your changes are actually correct!

As soon as you fix the indentation I'll merge the fix, thanks.

hasbean:
Felt like an episode of Silicon Valley, but all indented now!

lminiero:
Thanks! :+1:

PR #573, opened 2016-07-06
Fixes for 64-bit identifiers

lminiero:
Somebody recently made me notice that, although most of our identifiers are `guint64`, we use `g_random_int()` to assign them values, which is a method that only returns 32-bit unsigned integers. While this was brought to my attention as a security concern (it's easier to brute-force a 32-bit value), this actually raised some other issues when I looked into it, which is what this patch tries to address.

In fact, I soon found out it wasn't just a matter of generating actual 64-bit values. Throughout the code we often use `GHashTable` tables as maps for sessions, handles, rooms, mountpoints and so on. In order to use those numeric identifiers as keys to those maps, we used `GUINT_TO_POINTER` a lot. Unfortunately, that only works as expected for numbers as high as 2^32, and doesn't when you go beyond that. This means that when you use higher values (even when you pass a value yourself via API), assuming that a 64-bit value is ok, things go crazy, map lookups fail, and things like that.

In order to fix that, I had to do several things in the code.

1. I added two new methods, called `janus_random_uint32()` and `janus_random_uint64()`, to generate `guint32` and `guint64` random numbers respectively. `janus_random_uint32()` is simply `g_random_int()` in disguise. `janus_random_uint64()` uses `g_random_int()` twice to generate a random 64-bit unsigned integer. Besides, since JavaScript apparently can't handle any integer beyond 2^53, I had to make sure the generated value here wouldn't go beyond that (or otherwise nothing works in browsers).
2. I had to modify all the `GHashTable` management when those values were involved. Specifically, `g_int64_hash` and `g_int64_equal` are now passed as arguments when creating the table, in order to make sure keys are properly addressed and compared. All lookups and removes now address the identifiers by reference, as the key needs to be a pointer and, as anticipated, `GUINT_TO_POINTER` is a no go. Finally, for inserts we needed to duplicate (as in, allocate and copy) the identifiers in order to use them as keys. In fact, using temporary values and their references, there, would soon leave the key as garbage, rendering the respective values unreachable; using the properties of existing structs wasn't reliable either, as those could be freed before the key. In order to do so, I added a new `janus_uint64_dup` to take care of that. Of course, since these keys need to be automatically freed when they're not needed anymore, I made sure `g_free` was passed as the first `GDestroyFunction` when creating the tables.

This is basically it. If you were using 64-bit identifiers in a similar way you're encouraged to align to this patch, as soon as we merge it. This is not a direct commit as I want to make sure this doesn't break any existing app out there. In fact, we now end up often generating larger identifiers for both sessions and handles, which is something that may confuse applications that are not prepared for that (although they should be, those have always been supposed to be 64-bit).

In the future we may work on making the random generation of those identifiers more "secure". At the moment, as I said, it's still just using `g_random_int` anyway with some ands/ors to make sure it's a 0<x<2^53 value. So, no crypto or anything fancy is involved yet. 

That said, I'd like to merge this soon, so feedback on this, hints on things I may have forgotten/mistyped/whatever is more than welcome!

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -380,7 +380,7 @@ static gboolean janus_check_sessions(gpointer user_data) {
 				session->timeout = 1;
 				/* FIXME Is this safe? apparently it causes hash table errors on the console */
 				g_hash_table_iter_remove(&iter);
-				g_hash_table_insert(old_sessions, GUINT_TO_POINTER(session->session_id), session);
+				g_hash_table_insert(old_sessions, janus_uint64_dup(session->session_id), session);
 
 				/* Schedule the session for deletion */
 				GSource *timeout_source = g_timeout_source_new_seconds(3);
@@ -415,7 +415,7 @@ static gpointer janus_sessions_watchdog(gpointer user_data) {
 janus_session *janus_session_create(guint64 session_id) {
 	if(session_id == 0) {
 		while(session_id == 0) {
-			session_id = g_random_int();
+			session_id = janus_random_uint64();
 			if(janus_session_find(session_id) != NULL) {
 				/* Session ID already taken, try another one */
 				session_id = 0;
@@ -435,28 +435,28 @@ janus_session *janus_session_create(guint64 session_id) {
 	session->last_activity = janus_get_monotonic_time();
 	janus_mutex_init(&session->mutex);
 	janus_mutex_lock(&sessions_mutex);
-	g_hash_table_insert(sessions, GUINT_TO_POINTER(session_id), session);
+	g_hash_table_insert(sessions, janus_uint64_dup(session->session_id), session);
 	janus_mutex_unlock(&sessions_mutex);
 	return session;
 }
 
 janus_session *janus_session_find(guint64 session_id) {
 	janus_mutex_lock(&sessions_mutex);
-	janus_session *session = g_hash_table_lookup(sessions, GUINT_TO_POINTER(session_id));
+	janus_session *session = g_hash_table_lookup(sessions, &session_id);
 	janus_mutex_unlock(&sessions_mutex);
 	return session;
 }
 
 janus_session *janus_session_find_destroyed(guint64 session_id) {
 	janus_mutex_lock(&sessions_mutex);
-	janus_session *session = g_hash_table_lookup(old_sessions, GUINT_TO_POINTER(session_id));
+	janus_session *session = g_hash_table_lookup(old_sessions, &session_id);
 	janus_mutex_unlock(&sessions_mutex);
 	return session;
 }
 
 void janus_session_notify_event(guint64 session_id, json_t *event) {
 	janus_mutex_lock(&sessions_mutex);
-	janus_session *session = sessions ? g_hash_table_lookup(sessions, GUINT_TO_POINTER(session_id)) : NULL;
+	janus_session *session = sessions ? g_hash_table_lookup(sessions, &session_id) : NULL;
 	if(session != NULL && !session->destroy && session->source != NULL && session->source->transport != NULL) {
 		janus_mutex_unlock(&sessions_mutex);
 		/* Send this to the transport client */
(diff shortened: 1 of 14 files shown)


saghul:
I had a quick glance, and here is a random idea (pun intended): how about using some UUIDs as identifiers?

The identifier could be created with [uuid_generate_random](http://linux.die.net/man/3/uuid_generate_random) and converted into a hex string with [uuid_unparse_lower](http://linux.die.net/man/3/uuid_unparse). This way, the ids would be alphanumeric, random enough (we are using UUID v4) and JS has no problem with that.

lminiero:
That's definitely an idea for future version, but it's going to be a very breaking change (it changes the Janus API completely), and so should be done along more disruptive PRs... I'd like to fix the current behaviour with existing identifiers for the time being.

saghul:
Would it really be that breaking? Is there code that relies on the fact that identifiers are integer numbers? 😮

utils.c
```diff
@@ -69,6 +69,32 @@ gboolean janus_strcmp_const_time(const void *str1, const void *str2) {
...
+	 * PS: JavaScript only supports integer up to 2^53, so we need to
+	 * make sure the number is below 9007199254740992 for safety
+	 */
+	guint64 num = g_random_int() & 0x1FFFFF;
+	num = (num << 32) | g_random_int();
+	return num;
+}
+
+guint64 *janus_uint64_dup(guint64 num) {
+	guint64 *dup = g_malloc0(sizeof(guint64));
```
saghul: dup is a libc function name, maybe avoid shadowing it?

utils.c
```diff
@@ -69,6 +69,32 @@ gboolean janus_strcmp_const_time(const void *str1, const void *str2) {
...
+	 * more strongly random stuff... using /dev/urandom is probably not
+	 * a good idea, as we don't want to make it harder to cross compile Janus
+	 *
+	 * TODO Look into what libssl and/or libcrypto provide in that respect
+	 *
+	 * PS: JavaScript only supports integer up to 2^53, so we need to
+	 * make sure the number is below 9007199254740992 for safety
+	 */
+	guint64 num = g_random_int() & 0x1FFFFF;
+	num = (num << 32) | g_random_int();
```
saghul: Is it or or plus? I have similar code written as (high << 32) + low.

ploxiln: bitwise-or and plus are equivalent because the one-bits never overlap
(but bitwise-or is conceptually what is intended)

lminiero: Yeah it's more of a bitwise operation here, although a plus would get you the same result.

utils.c
lminiero: Ack, will fix tomorrow.

lminiero:
> Would it really be that breaking? Is there code that relies on the fact that identifiers are integer numbers?

Well, languages that don't care about types may be fine, but that's not always the case, especially in API wrappers. If you expect a session_id to be an integer and then you get a string, things could get messy. Same applies to room ids, participant ids, and so on and so forth. It's a transition that needs to be advertized in proper advance, and put together with other stuff that will need to be taken care of so that people really pay attention. This patch is supposed to be transparent to all that instead, which is why I want to fix that until then.

lminiero:
Merging, as it seems to fix the problem for me.

janus.c
```diff
@@ -435,28 +435,28 @@ janus_session *janus_session_create(guint64 session_id) {
...
-	g_hash_table_insert(sessions, GUINT_TO_POINTER(session_id), session);
+	g_hash_table_insert(sessions, janus_uint64_dup(session->session_id), session);
 	janus_mutex_unlock(&sessions_mutex);
 	return session;
 }
 
 janus_session *janus_session_find(guint64 session_id) {
 	janus_mutex_lock(&sessions_mutex);
-	janus_session *session = g_hash_table_lookup(sessions, GUINT_TO_POINTER(session_id));
+	janus_session *session = g_hash_table_lookup(sessions, &session_id);
```
jswirl: Here we're using the address of `session_id`, which is located on function local stack.
But when inserting, we used the returned address of `janus_uint64_dup()`, which is allocated on the heap.
These may or may not be pointing to the memory location of our `session_id`, is this the expected behaviour?

lminiero: It is expected, as we're using `g_hash_table_new_full(g_int64_hash, g_int64_equal`, which uses a dedicated method is used to compare them to find the right one. The reason why we're allocating memory for the key is because the pointer you use for inseerting is the one that is used for actually addressing the key: all other calls on the key instead (lookup, remove) are hashed and compared using the specified methods, and so can be temporary. If you use a temporary `guint64` and use a pointer to that for inserting, as soon as the calling method returns the key points to garbage and the value is lost.

jswirl: Okay. I see how `GHashTable` is used now. Thanks.

PR #575, opened 2016-07-08
RTSP PLAY request URL should not have a slash appended

foxxyz:
While testing some RTSP streaming, I ran into [this interesting line](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_streaming.c#L2962), where a slash is being appended to the given stream URL before issuing the PLAY request. This lead to the following request/response:
```
PLAY rtsp://192.168.0.67/b5.mp4/ RTSP/1.0
CSeq: 4
Range: 0.000-
> 400 BAD REQUEST
```

Removing the slash as in the patch seemed to fix the issue, returning a 200 status code and starting playback, although I'm curious how this has been working for others. Perhaps I have a picky server! [The IETF RTSP RFC](https://www.ietf.org/rfc/rfc2326.txt) suggests there should not be a slash appended.

Let me know if you have any comments and thanks for a great piece of software!

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -2959,8 +2959,7 @@ janus_streaming_mountpoint *janus_streaming_create_rtsp_source(
 	g_free(data.buffer);
 	data.buffer = g_malloc0(1);
 	data.size = 0;		
-	sprintf(uri, "%s/", url);
-	curl_easy_setopt(curl, CURLOPT_RTSP_STREAM_URI, uri);
+	curl_easy_setopt(curl, CURLOPT_RTSP_STREAM_URI, url);
 	curl_easy_setopt(curl, CURLOPT_RANGE, "0.000-");
 	curl_easy_setopt(curl, CURLOPT_RTSP_REQUEST, (long)CURL_RTSPREQ_PLAY);		
 	res = curl_easy_perform(curl);


ploxiln:
The confusion may be because often an RTSP server will respond to a DESCRIBE request with a "Content-Base" which adds a slash, and multiple streams which can be specified as a component after the slash. An example in annoying gstreamer rtspsrc debug output format because it's convenient for me to copy/paste at the moment:

```
RTSP request message 0x700000428c60
 request line:
   method: 'DESCRIBE'
   uri:    'rtsp://127.0.0.1:8554/test'
   version: '1.0'
 headers:
   key: 'User-Agent', value: 'GStreamer/1.8.2'
   key: 'Accept', value: 'application/sdp'

RTSP response message 0x700000428c00
 status line:
   code:   '200'
   reason: 'OK'
   version: '1.0'
 headers:
   key: 'CSeq', value: '2'
   key: 'Content-Type', value: 'application/sdp'
   key: 'Content-Base', value: 'rtsp://127.0.0.1:8554/test/'
   key: 'Server', value: 'GStreamer RTSP server'
   key: 'Date', value: 'Fri, 08 Jul 2016 06:19:00 GMT'
   key: 'Content-Length', value: '489'
 body: length 490

(the SDP which describes two streams)

RTSP request message 0x700000428aa0
 request line:
   method: 'SETUP'
   uri:    'rtsp://127.0.0.1:8554/test/stream=0'
   version: '1.0'
 headers:
   key: 'User-Agent', value: 'GStreamer/1.8.2'
   key: 'Transport', value: 'RTP/AVP;unicast;client_port=53368-53369'

RTSP response message 0x700000428a40
 status line:
   code:   '200'
   reason: 'OK'
   version: '1.0'
 headers:
   key: 'CSeq', value: '3'
   key: 'Transport', value: 'RTP/AVP;unicast;client_port=53368-53369;server_port=56258-56259;ssrc=D8453CC8;mode="PLAY"'
   key: 'Server', value: 'GStreamer RTSP server'
   key: 'Session', value: 'jZb3bh10KyEItfUF'
   key: 'Date', value: 'Fri, 08 Jul 2016 06:19:00 GMT'

(... other stream is setup ...)

RTSP request message 0x700000428c20
 request line:
   method: 'PLAY'
   uri:    'rtsp://127.0.0.1:8554/test/'
   version: '1.0'
 headers:
   key: 'User-Agent', value: 'GStreamer/1.8.2'
   key: 'Range', value: 'npt=0-'
 body:
```

lminiero:
I admit I never played much with the RTSP support. I guess that the base is used automatically by the stack for the SETUP lines, and so there's no need for us to do that ourselves.

@foxxyz before I can merge, can you sign our CLA at the bottom of [this](https://cla.conf.meetecho.com/janus/) page? Thanks!

foxxyz:
Thanks @ploxiln. So to be more correct, the PLAY request URL should be taken from a `Content-Base` header (if returned during the DESCRIBE operation) and otherwise be identical to the URL given in the configuration? For example, my server does not return a `Content-Base`.

@lminiero Thanks! I have signed the agreement.

ploxiln:
I'm not 100% sure about that anymore, maybe that slash is something gstreamer adds which isn't really needed, even if the Content-Base has the slash ...

see the examples if you scroll up and down a bit at https://tools.ietf.org/html/rfc2326#page-67

ploxiln:
I checked with a physical ip camera, and it's DESCRIBE response had a `Content-Base` without the trailing slash added, and gstreamer didn't add the trailing slash in the PLAY request.

If you read that part of the RFC I linked above, that's the only example exchange in that RFC which includes the Content-Base header, and in that case the Content-Base header had the trailing slash, but the PLAY request did not.

So, according to the RFC from 1998 you never add the trailing slash, but according to gstreamer you add it if it's in the Content-Base. ¯\_(ツ)_/¯

ploxiln:
(and `Content-Base` is not defined in that RFC, it refers to the HTTP RFC, but `Content-Base` was removed in HTTP/1.1 because it was not consistently used. In HTTP/1.0 it was the base for relative URLs in the page. and all this stuff is pre-21st-century)

foxxyz:
@ploxiln Thanks for the research! Seems like my change is correct then.

Let me know if you any other reservations before merging! Thanks!

lminiero:
I'll merge tomorrow when I get back to the office, thanks.
Il 10/lug/2016 02:27, "foxxyz" <notifications@github.com> ha scritto:

lminiero:
:+1:

foxxyz:
Thanks @lminiero!

PR #577, opened 2016-07-14
Rudimentary handling of SIP session-refresh (keepalive) added.

tuijldert:
Further to thread: https://groups.google.com/forum/#!searchin/meetecho-janus/reinvite/meetecho-janus/8S3Fuj4RR3s/KwzmIMDHBwAJ, this change should solve the issue.

On receiving a re-INVITE it compares the sdp payload with the original agreed-upon sdp and if not different, concludes it's a keepaliv and just responds with a 200 Ok.
Otherwise, behaviour is as before.

Please review and commit.

Hope this helps.

TIA,
Tom

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -327,6 +327,7 @@ typedef struct janus_sip_session {
 	janus_sip_account account;
 	janus_sip_call_status status;
 	janus_sip_media media;
+	sdp_parser_t *raw_media;
 	char *transaction;
 	char *callee;
 	janus_recorder *arc;		/* The Janus recorder instance for this user's audio, if enabled */
@@ -2276,8 +2277,13 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			}
 			if(session->stack->s_nh_i != NULL) {
 				if(session->stack->s_nh_i == nh) {
-					/* re-INVITE, we don't support those. */
-					nua_respond(nh, 488, sip_status_phrase(488), TAG_END());
+					if (sdp_session_cmp(sdp_session(session->raw_media), sdp_session(parser)) == 0) {
+						/* re-INVITE that is basically a session-refresh (no media-change needed), accept. */
+						nua_respond(nh, 200, sip_status_phrase(200), TAG_END());
+					} else {
+						/* re-INVITE, we don't support those. */
+						nua_respond(nh, 488, sip_status_phrase(488), TAG_END());
+					}
 				} else if(session->status >= janus_sip_call_status_inviting) {
 					/* Busy with another call */
 					JANUS_LOG(LOG_VERB, "\tAlready in a call (busy, status=%s)\n", janus_sip_call_status_string(session->status));
@@ -2464,6 +2470,7 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				JANUS_LOG(LOG_VERB, "Detected video codec: %d (%s)\n", session->media.video_pt, session->media.video_pt_name);
 			}
 			session->media.ready = 1;	/* FIXME Maybe we need a better way to signal this */
+			session->raw_media = parser;
 			GError *error = NULL;
 			g_thread_try_new("janus rtp handler", janus_sip_relay_thread, session, &error);
 			if(error != NULL) {
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -815,7 +815,7 @@ int janus_videoroom_init(janus_callbacks *callback, const char *config_path) {
 				videoroom->fir_freq = atol(firfreq->value);
 			videoroom->acodec = JANUS_VIDEOROOM_OPUS;
 			if(audiocodec && audiocodec->value) {
-				if(!strcasecmp(audiocodec->value, ""))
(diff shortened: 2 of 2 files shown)


plugins/janus_sip.c
```diff
@@ -2464,6 +2470,7 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				JANUS_LOG(LOG_VERB, "Detected video codec: %d (%s)\n", session->media.video_pt, session->media.video_pt_name);
 			}
 			session->media.ready = 1;	/* FIXME Maybe we need a better way to signal this */
+			session->raw_media = parser;
```
lminiero: This made me notice that we're missing a `sdp_parser_free(parser);` there: we do have it for errors and not for when things go fine instead. As a side note, you're not freeing that reference either, which is a leak.

plugins/janus_sip.c
```diff
@@ -2276,8 +2277,13 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			}
 			if(session->stack->s_nh_i != NULL) {
 				if(session->stack->s_nh_i == nh) {
-					/* re-INVITE, we don't support those. */
-					nua_respond(nh, 488, sip_status_phrase(488), TAG_END());
+					if (sdp_session_cmp(sdp_session(session->raw_media), sdp_session(parser)) == 0) {
```
lminiero: How does this method work? Does it only inspect media, ports and stuff like that, or everything in the SDP? Just wondering as my guess that even a tiny difference in the SDP may actually trigger a `!=0` there.

tuijldert:
I observed the -potential- leak as well and was wondering about that.

Given the consistency with which this is done in `janus_sip.c` and my unfamiliarity with the code I presumed this might have been intentional.

Furthermore, not sure what the compare does exactly but it should indeed fail on the tiniest of differences since the gateway doesn't support any media changes (and they *are* media changes in case of a diff, we checked that).

A possible future extension for this btw., is an indication that refreshes can be done via UPDATEs (include an `Allow: UPDATE` and catch update requests). Is there a wishlist somewhere?

Anyway, I can work on getting rid of the leak when closing the session:
1. are we Ok with the addition of the `raw_media` pointer?
2. Should I do that in a separate pull request?

lminiero:
> I observed the -potential- leak as well and was wondering about that.
> 
> Given the consistency with which this is done in janus_sip.c and my unfamiliarity with the code I presumed this might have been intentional.

Actually we freed this when handling messages originated by users (call, answer), and not in callbacks coming from Sofia (r_invite, i_invite).

I fixed the leak for the SIP plugin in master, so you should first merge with that. Then you only need to remove the `sdp_parser_free(parser)` when keeping the reference (as otherwise of course your pointer points to garbage), and make sure your `raw_media` is freed properly when cleaning up the session.

> Furthermore, not sure what the compare does exactly but it should indeed fail on the tiniest of differences since the gateway doesn't support any media changes (and they are media changes in case of a diff, we checked that).

Fair point. Since no ICE or fancy stuff is involved, in the future we should really make sure we can adapt to IP/ports changes, there, as no change would be needed on the WebRTC side and so no renegotiation on that side. New media being added/removed instead are a different matter.

> A possible future extension for this btw., is an indication that refreshes can be done via UPDATEs (include an Allow: UPDATE and catch update requests). Is there a wishlist somewhere?
> 
> Anyway, I can work on getting rid of the leak when closing the session:
> 1. are we Ok with the addition of the raw_media pointer?
> 2. Should I do that in a separate pull request?

No need to do that in a separate PR, I guess, as it is related to the functionality you're adding (unless the `Allow: UPDATE` implies support for many more things we don't really provide).

By the way, forgot to say this before: thanks for working on this!

tuijldert:
Done.
Let's conclude this and leave the UPDATE implementation as a separate exercise as that implies more than just refreshing.

+1 on port changes but no priority.

And your welcome :)

Please review and commit.

lminiero:
Sorry for the lack of feedback, I've been abroad for the IETF and have just come back to work. I'll review by today, promise!

lminiero:
Looks fine, thanks :+1: 
The SDP parser stuff will probably become unneeded as soon as we merge #578 and make the SIP plugin use the new SDP utilities as well.

tuijldert:
Thanks for the merge.

Will there be a "next stable release" of the gateway somewhere soon?

Bit skeptical on the SDP utils... Aren't you re-inventing the wheel a bit? Wouldn't improving the Sofia library be easier and more efficient? (I know, I'm lazy like that, I hate duplicate work... :)

lminiero:
I'm most definitely not getting to work on improving Sofia SIP, which is an old library only FreeSwitch is mantaining, I believe. Any change I make would never end up in a timely way in the version of the libraries repos ship. Besides, I'd have to dig into someone else's code, something I have absolutely no intention of doing.

We currently depend on Sofia SIP exclusively for SDP (apart from the SIP plugin, of course), which is overkill. The new SDP utils provide the same features without an unneeded dependency, and with just a few lines of code.

saghul:
> Bit skeptical on the SDP utils... Aren't you re-inventing the wheel a bit? Wouldn't improving the Sofia library be easier and more efficient? (I know, I'm lazy like that, I hate duplicate work... :)

Sofia in unmaintained at this point.

PR #584, opened 2016-07-23
Add support for 'ack' field in text room messages (Textroom plugin)

joshdickson40:
This PR adds support for an 'ack' field in messages sent to the text room plugin per conversation [here](https://groups.google.com/forum/#!topic/meetecho-janus/2EcoJajOzmU).

For rooms with high message send frequency, the additional ack messages are unnecessary overhead, as we can instead assume that if (1) the connection is up and (2) we do not see an error event, that the server was able to see and parse the message. This is consistent with event handling from other real-time data API providers from an API implementation perspective (i.e. Pusher, Pubnub, etc). 

Per discussion, this is a non-breaking change. If no field is provided, the current behavior (ack message is sent) is continued. If the ack field is provided, the plugin will observe the value specified.

This PR also updates the text room demo to show usage of the 'ack' field, along with a note that this is *not a client ack*, as in the ack event is confirmation only that the server saw the event and not that it was actually distributed to any clients. The text room has never done client delivery acks (you can build it on your own, if you like), and I'd suggest it remains that way to keep things simple :).

Tested on Ubuntu 14.04 and 16.04. Feedback welcome, of course.

### DIFF
--- a/html/textroomtest.js
+++ b/html/textroomtest.js
@@ -338,7 +338,7 @@ function sendData() {
 		transaction: randomString(12),
 		room: myroom,
 		text: data,
-    ack: true // acknowledge message reciept by server? (does NOT acknowledge receipt by other clients)
+        ack: true // acknowledge message reciept by server? (does NOT acknowledge receipt by other clients)
 	};
 	textroom.data({
 		text: JSON.stringify(message),
--- a/html/textroomtest.js
+++ b/html/textroomtest.js
@@ -337,7 +337,7 @@ function sendData() {
 		textroom: "message",
 		transaction: randomString(12),
 		room: myroom,
-		text: data,
+        text: data,
         ack: true // acknowledge message reciept by server? (does NOT acknowledge receipt by other clients)
 	};
 	textroom.data({
--- a/html/textroomtest.js
+++ b/html/textroomtest.js
@@ -159,11 +159,11 @@ $(document).ready(function() {
 										if(whisper === true) {
 											// Private message
 											$('#chatroom').append('<p style="color: purple;">[' + dateString + '] <b>[whisper from ' + participants[from] + ']</b> ' + msg);
-											$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;				
+											$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;
 										} else {
 											// Public message
 											$('#chatroom').append('<p>[' + dateString + '] <b>' + participants[from] + ':</b> ' + msg);
-											$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;				
+											$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;
 										}
 									} else if(what === "join") {
 										// Somebody joined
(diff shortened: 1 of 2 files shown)


joshdickson40:
Also, this was tested on Ubuntu 14.04 and 16.04. Feedback welcome, of course.

lminiero:
Thanks for working on this, I'll review it better when I get back to the office. A couple of quick notes:

1. we use tabs for indenting, apparently you're using spaces instead?
2. the ack property should be added to the message definition, as the JSON validatio needs to know what this is (boolean and optional, in this case); https://github.com/joshdickson40/janus-gateway/blob/784b9c1cee490d5e91ea18201536a8c43c14b0e1/plugins/janus_textroom.c#L153

joshdickson40:
Those are just trailing spaces being removed that shouldn't be there anyway.

lminiero:
> Those are just trailing spaces being removed that shouldn't be there anyway.

That's not what I'm talking about. I'll hightlight what I mean by individual comments on the code.

html/textroomtest.js
```diff
@@ -337,7 +337,8 @@ function sendData() {
 		textroom: "message",
 		transaction: randomString(12),
 		room: myroom,
-		text: data
+		text: data,
+    ack: true // acknowledge message reciept by server? (does NOT acknowledge receipt by other clients)
```
lminiero: Tabs, not spaces.

plugins/janus_textroom.c
```diff
@@ -154,7 +154,8 @@ static struct janus_json_parameter message_parameters[] = {
 	{"room", JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE},
 	{"text", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"to", JSON_STRING, 0},
-	{"tos", JSON_ARRAY, 0}
+	{"tos", JSON_ARRAY, 0},
+  {"ack", JANUS_JSON_BOOL, 0}
```
lminiero: Tabs, not spaces.

plugins/janus_textroom.c
```diff
@@ -803,7 +804,9 @@ void janus_textroom_handle_incoming_request(janus_plugin_session *handle, char *
 		}
 		g_free(msg_text);
 		janus_mutex_unlock(&textroom->mutex);
-		if(!internal) {
+    /* Should we send ack event? Yes if no ack field provided, else follow field directive */
+    json_t *ack = json_object_get(root, "ack");
```
lminiero: Tabs, not spaces.

joshdickson40:
Ah, I'm sorry, yes you're right, weird Atom issue with inconsistent tab lengths.

joshdickson40:
Well this is weird. Definitely not using spaces locally, looks correct (tabs) in Sublime & Atom... but they are definitely showing up as spaces here.

joshdickson40:
Well that was annoying - was not using tabs here, but they were sure showing up as tabs on the commit. Should be fixed.

lminiero:
Thanks :+1:

lminiero:
I'll probably fix a couple of comments in some additional commits.

PR #589, opened 2016-07-28
First take at supporting re-invites/updates in SIP plugin (uses #578)

lminiero:
As @saghul suggested, SIP re-INVITE support is added as a separate pull request that is based on #578. Copying text I wrote there here.

It basically updates the socket connections after the peer re-invites us, and makes sure no media is forwarded from the browser to the SIP peer if the media direction changed. Still no way for the browser to trigger a re-invite as we don't support renegotiation on the Janus core side yet.

This seems to work for me, but I only tried with Asterisk and forwards (Asterisk answering and playing something back, and then forwarding the call to a SIP endpoint), so you guys may want to make further tests to make sure this works as expected. Mentioning @tuijldert in particular as he worked on #577 in master.

Still incomplete as this does NOT renegotiate SRTP contexts too if renegotiated: it currently only works with plain RTP. Updating a session with new crypto stuff will require more work. _Note:_ although just tried and with Asterisk it worked nevertheless: since SRTP is involved, Asterisk doesn't force media end-to-end after in the re-INVITE, but stays on the media path to ensure secure media between Janus and itself, and allowing plain RTP on the other end. Anyway, with proxies that behave differently this is still something that needs effort on our end.

Feedback welcome!

### DIFF
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -156,6 +156,14 @@ $(document).ready(function() {
 											$('#password').removeAttr('disabled');
 											$('#register').removeAttr('disabled').click(registerUsername);
 											$('#registerset').removeAttr('disabled');
+										} else {
+											// Reset status
+											sipcall.hangup();
+											$('#dovideo').removeAttr('disabled').val('');
+											$('#peer').removeAttr('disabled').val('');
+											$('#call').removeAttr('disabled').html('Call')
+												.removeClass("btn-danger").addClass("btn-success")
+												.unbind('click').click(doCall);
 										}
 										bootbox.alert(error);
 										return;
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -3012,7 +3012,7 @@ static void *janus_sip_relay_thread(void *data) {
 	socklen_t addrlen;
 	struct sockaddr_in remote;
 	int resfd = 0, bytes = 0;
-	struct pollfd fds[4];
+	struct pollfd fds[5];
 	int pipe_fd = session->media.pipefd[0];
 	char buffer[1500];
 	memset(buffer, 0, 1500);
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2359,6 +2359,7 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				break;
 			}
 			/* Send SDP to the browser */
+			char *fixed_sdp = g_strdup(sip->sip_payload->pl_data);
 			json_t *jsep = json_pack("{ssss}", "type", "offer", "sdp", fixed_sdp);
 			json_t *call = json_object();
 			json_object_set_new(call, "sip", json_string("event"));
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -295,6 +295,14 @@ typedef struct janus_sip_account {
 	janus_sip_registration_status registration_status;
 } janus_sip_account;
 
+typedef struct janus_sip_rtp_context {
+	/* Needed to fix seq and ts in case of re-INVITEs/UPDATEs that result in a RTP stream */
+	uint32_t a_last_ssrc, a_last_ts, a_base_ts, a_base_ts_prev,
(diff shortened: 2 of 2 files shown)


lminiero:
@tuijldert I've merged the fixes in this PR as well, which I believe is what you're testing (I do see the poll at line 3068, here). Please let me know if the fix works.

tuijldert:
[janus-log.txt](https://github.com/meetecho/janus-gateway/files/449395/janus-log.txt)

No change, see the log. Makes sense?

tuijldert:
[janus-log.txt](https://github.com/meetecho/janus-gateway/files/449402/janus-log.txt)

Same result. See log. Makes any sense?

lminiero:
Doesn't seem to provide any additional info. I did fix an issue that may have caused memory corruptions, namely the pointers to freed lists that we were not setting to NULL before reusing them, so that's not it. Not sure as to what may cause an ABORT on just polling a few file descriptors. Besides, I don't see any reference to buffer overflows in those logs. You may have to check the core dump with gdb to see the values of the involved variables, in this case I think only fds.

Have you already tried doing a make clean and new install?

tuijldert:
Hmmm, that is my standard operating procedure: do a '`make clean`, get the changes and perform the `make` and `sudo make install`...
Should I bring in AddressSanitizer to get some more info? any tips on that?

I'm also surprised you don't have these same issues. All we do is connect a Janus client and call it with audio and video. As soon as the ACK comes in (and media should start flowing from gatway to our server), the gateway crashes. So we can crash Janus within 30 sec.
Performing the same test with the master branch succeeds.

I've included a wireshark (sip-) capture from the gateway (192.168.51.18) to illustrate. As soon as the ACK is received, Janus crashes.

[JanusCrashTest.zip](https://github.com/meetecho/janus-gateway/files/449681/JanusCrashTest.zip)

Hope this helps,
Tom.

lminiero:
I'm not able to replicate it, it works as expected for me. I'm only trying with audio, though, while from a quick glance to your dump video is enabled too, which means more file descriptors are involved in the poll. I'll make a test with video too to see if it happens for me there. If you can try with an audio only call that would help too.

AddressSanitizer is worth a try, if not just to check if it spots some memory corruption that eventually leads to the problem you're experiencing.

lminiero:
Aah, maybe I got it... we define the array of file descriptors for the poll as `struct pollfd fds[4];`, but we have _5_ descriptors when audio and video are available: 2 for audio, 2 for video, and one for the pipe that is needed to be notified about re-invites. I'll push a commit in a minute that should address this.

lminiero:
@tuijldert can you try again now please?

tuijldert:
Jackpot!
Stay tuned for the next level; re-invite tests... :)

Thanks,
Tom.

tuijldert:
> Feedback welcome!
Keepalive re-Invites look ok. See the trace of the test.

[JanusReInviteTest.zip](https://github.com/meetecho/janus-gateway/files/449824/JanusReInviteTest.zip)

lminiero:
Cool! Although I see your re-invite doesn't change the ports. This new branch also supports changing the remote ports in-call now.

lminiero:
Merging this with the SDP refactoring stuff.

saghul:
@lminiero Oh, why?! This was something to be considered after the SDP refactor, which just landed... I guess we'll see how it goes then.

lminiero:
@saghul you're right, sorry about that! Since it was a patch on something we were merging I thought we'd just merge that one too... @tuijldert had done some testing on it and so had I in different scenarios, so I felt safe enough doing that, mea culpa! Anyway, it should be easy enough to disable it if you'd rather it not mess with your sessions, as we wheck it [here](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_sip.c#L2321), where you can just reply with a 200 as it did before instead of handling the re-INVITE. Maybe we can make this configurable in the settings? Let me know and I'll work on a patch to do that.

Feel free to insult me when we meet next week! :smile:

saghul:
Heh, no worries Lorenzo! Let's brainstorm next week, everything will be clearer with Argentinian steak and wine :-)

PR #592, opened 2016-08-02
Fix for post-processing (timestamp resets + retransmissions)

lminiero:
This is an attempt to fix an issue reported in a [post](https://groups.google.com/forum/#!topic/meetecho-janus/bM8_4iATTjw) on the group. Vladislav provided a a link to a .mjr file, which although being only 20 Mb size, would be converted to a 1.1 gb Opus file by janus-pp-rec.

I replicated the issue and studied the logs, and the cause was a confusion in the tool when handling retransmitted packets that occurred in the vicinity of timestamp resets. The patch this PR provides seems to fix the issue (the resulting opus is now ~16mb), but I dediced to publish it as a PR rather than directly committing to master as I want to make sure this doesn't introduce regressions with recordings that worked fine before instead.

Please test this with your recordings as much as possible, and let us know if you experience issues. Mentioning @kris-lab in particular as he was involved in #471 and #472 that addressed a similar problem with timestamp resets, and so we definitely don't want that problem to come back to bite us.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -143,9 +143,6 @@ int main(int argc, char *argv[])
 	long offset = 0;
 	uint16_t len = 0;
 	uint32_t count = 0;
-	uint32_t last_ts = 0, reset = 0;
-	int times_resetted = 0;
-	uint32_t post_reset_pkts = 0;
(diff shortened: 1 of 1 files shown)


lminiero:
This has been around long enough, merging.

PR #594, opened 2016-08-03
Use `var` keyword before declaring charSet var.

rpadovani:
While JavaScript does not require this, if the script is merged
with other scripts that have "use strict", the code stops working.

Also there is no reason the scope of charSet is global.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -270,7 +270,7 @@ function Janus(gatewayCallbacks) {
 
 	// Private method to create random identifiers (e.g., transaction)
 	function randomString(len) {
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks for this! Could you update the `janus.nojquery.js` file as well, which probably suffers from the same issue?

As a side note, can you sign our [CLA](https://cla.conf.meetecho.com/janus/) before we can merge this?

Thanks!

rpadovani:
Thanks to you for your awesome work on this!

Done, I signed the CLA and updated `janus.nojquery.js` as well

lminiero:
Thanks, merging then!

PR #606, opened 2016-09-06
Changes to DTLS BIO filter for OpenSSL 1.1.0

### LINKED ISSUE #602: OpenSSL 1.1.0 support
Janus no longer builds against OpenSSL 1.1.0.
OpenSSL 1.1.0 was released last week and made some structures opaque.

lminiero:
This is an attempt to fix #602.

OpenSSL 1.1.0, among other changes, made some structures we use opaque. This affected the DTLS BIO filter we have to handle automatic fragmentation, in `dtls-bio.c`. The OpenSSL developers were [really helpful](https://mta.openssl.org/pipermail/openssl-users/2016-September/004364.html) in providing some info on how to handle the transition, so I prepared the patch you can find in this pull request. The patch basically uses the `OPENSSL_VERSION_NUMBER` define to check if we're on OpenSSL 1.1.0 or not. If we're on an older version, we do nothing different than what we did before; if we are on 1.1.0, we use the new setters/getters instead.

I don't have OpenSSL 1.1.0 installed so I can't check if this even compiles. If you do, please compile/test and provide feedback.

### DIFF
--- a/dtls-bio.c
+++ b/dtls-bio.c
@@ -35,6 +35,8 @@ long janus_dtls_bio_filter_ctrl(BIO *h, int cmd, long arg1, void *arg2);
 int janus_dtls_bio_filter_new(BIO *h);
 int janus_dtls_bio_filter_free(BIO *data);
 
+/* Filter initialization */
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 static BIO_METHOD janus_dtls_bio_filter_methods = {
 	BIO_TYPE_FILTER,
 	"janus filter",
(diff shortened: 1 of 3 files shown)


seirotis:
Will test this now.

> I don't have OpenSSL 1.1.0 installed so I can't check if this even compiles. If you do, please compile/test and provide feedback.

You can still compile Janus against a custom OpenSSL:
```
./configure CFLAGS='-I/path/to/.openssl/include/' LDFLAGS='-L/path/to/.openssl/lib/'
```

seirotis:
Compile output:

```
...
dtls.c:148:13: warning: 'janus_dtls_cb_openssl_threadid' defined but not used [-Wunused-function]
 static void janus_dtls_cb_openssl_threadid(CRYPTO_THREADID *tid) {
             ^
dtls.c:160:13: warning: 'janus_dtls_cb_openssl_lock' defined but not used [-Wunused-function]
 static void janus_dtls_cb_openssl_lock(int mode, int type, const char *file, int line) {
             ^
  CC       janus-dtls-bio.o
dtls-bio.c: In function 'janus_dtls_bio_filter_write':
dtls-bio.c:135:2: error: too few arguments to function 'BIO_next'
  long ret = BIO_write(BIO_next(), in, inl);
  ^
In file included from /path/to/.openssl-1-1-0/include/openssl/err.h:21:0,
                 from dtls-bio.h:17,
                 from dtls-bio.c:16:
/path/to/.openssl-1-1-0/include/openssl/bio.h:561:6: note: declared here
 BIO *BIO_next(BIO *b);
...
```

cacheworks:
Does it compile when you replace line 135 of dtls-bio.c with the following:
```long ret = BIO_write(BIO_next(bio), in, inl)```

lminiero:
Yeah, sorry, I forgot the argument to `BIO_next`, the suggestion by @cacheworks should fix that.

lminiero:
Fixed, could you let me know if that works now?

seirotis:
Compilation works, but after starting Janus and the Echo test, janus crashes: Segmentation fault (core dumped)

lminiero:
Just found out another missing line: I forgot to actually initialize the BIO filter at startup, which for OpenSSL 1.1.0 meant it always being NULL, hence the crash. This should fix it.

lminiero:
I installed OpenSSL 1.1.0 myself in a custom path, and this patch seems to work for me. Anyway, apparently the way OpenSSL handles threading changed in 1.1.0, and stuff like `CRYPTO_THREADID_set_callback` does nothing now (which explains the warnings you also pasted some comments ago). I'll have to check if safe threading is implicit, now, or if it needs to be enabled some other way: we have those calls for a reason, as otherwise bad things can (and did) happen when numbers increased.

lminiero:
According to [this changelog](https://www.openssl.org/news/cl110.txt), `It is no longer necessary to set locking callbacks to use OpenSSL in a multi-threaded environment`, so I hid the related `CRYPTO_THREADID` stuff behind an `ifdef` that is only used for OpenSSL < 1.1.0. Besides, `DTLSv1_method()` is deprecated as well, as it's simply `DTLS_method()` now, so I fixed that too.

If you guys can confirm this works for you both with OpenSSL 1.0.x and 1.1.0 I'll merge.

cacheworks:
This has compiled and is running without problems as far as I can tell with OpenSSL 1.0.2g and 1.1.0 

I verified that OpenSSL 1.1.0 was being used using ```nm /opt/janus/bin/janus | grep OPENSSL``` and tested the Echotest and Videoroom plugins

lminiero:
Thanks, merging then!

PR #403, opened 2015-12-03
Reference counters

lminiero:
This patch introduces a new file that provides a reference counter implementation that should hopefully be used across the whole Janus code, so that we can get rid of all the timed watchdogs/garbage collectors we have in place now (and the problems they can cause any time a race condition occurs), in both core and plugins/transports.

This PR does nothing but add this new implementation to the codebase. It's still not integrated in any part of Janus, as I first wanted to make sure this was a step in the right direction, and whether or not the mechanism itself as it is implemented right now can be improved. I didn't want to make use of GLib GObjects for several reasons, so I decided to take a lighter approach with something more "homemade". This excellent [blog post](http://nullprogram.com/blog/2015/02/17/) provided some really nice info on how this could be done in an effective way, and that's what lead to the code I just added, which is quite obviously heavily based on that: atomic operations on counters, and a callback function that the application sets and is called when the counter gets to 0.

According to how the implementation works, now, this is how you'd use it. Assuming this is the struct we want to handle through the RC mechanism:

```
struct mystruct {
    int number;
    char *string;
}
```

that's how you'd modify it:

```
struct mystruct {
    int number;
    char *string;
    janus_refcount rc;
}
```

When allocating a new instance of that struct, you have to initialize the reference counter:

```
struct mystruct *instance = [..];
janus_refcount_init(&instance->rc, mystruct_free_cb);
```

This automatically sets the counter to 1 (as you obviously own yourself), and instructs the reference counter about which function must be invoked when the counter gets to 0 and so you can free the instance. The `mystruct_free_cb` method must have a signature like this:

```
void mystruct_free_cb(janus_refcount *counter);
```

and its implementation might look like this:

```
void mystruct_free_cb(janus_refcount *counter) {
    struct mystruct *instance = janus_refcount_containerof(counter, struct mystruct, rc);
    if(instance->string)
        free(instance->string);
    free(instance);
}
```

The important part here is `janus_refcount_containerof`, which is a macro we use to get the pointer to the instance that needs to be freed out of its reference counter. We pass it, in order:
1. a pointer to the reference counter itself;
2. the type of the instance (`struct mystruct` in this case);
3. the name of the reference counter as member of the instance struct (`rc` in our example).

This is needed because, as explained in the mentioned blog post, you can make use of `offsetof` to programmatically calculate the pointer to the container instance starting from one of its members (our reference counter, in this case). This saves us the need to have another pointer (and so more bytes) in `janus_refcount` itself to keep track of the object instance we're handling.

[...]

PR #611, opened 2016-09-12
Include refcount.h in files that use janus_refcount.

andreasg123:
For #403, I found two header files that use `janus_refcount` but do not include `refcount.h` (one already reported by @cacheworks).  `dtls.h` could get it via `sctp.h` but only if `HAVE_SCTP` is defined.  `janus.h` always includes `refcount.h` indirectly via `ice.h` but I think that a direct inclusion is less error prone.

### DIFF
--- a/dtls.h
+++ b/dtls.h
@@ -20,6 +20,7 @@
 #include <srtp/srtp.h>
 
 #include "sctp.h"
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks Andreas, merging!

PR #612, opened 2016-09-13
Bug in janus_sip.c when the sip stack receives an INVITE with no SDP

matthewmgamble:
Found a bug in janus_sip.c when the sip stack receives an INVITE without SDP after the initial invite.  In this call flow, Janus was assuming the invite would always have an SDP and would segfault when receiving an invite without one.

Added this quick patch to check if the received invite actually has an SDP before attempting to parse it, and if it doesn't, just respond 200 OK and let the call continue.

Basic testing shows that this solves the issue and prevents the segfault.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2253,7 +2253,7 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				break;
 			}
 			if (!sip->sip_payload) {
-				JANUS_LOG(LOG_ERR,"\tReceived re-invite without SDP - responding 200 OK\n");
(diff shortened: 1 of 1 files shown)


lminiero:
Cool, thanks for spotting this! Before we can merge, could you sign our [CLA](https://cla.conf.meetecho.com/janus/) at the end of this page here? Thanks!

matthewmgamble:
Done!

lminiero:
Thanks! Just as a last thing before the merge, could you change the `LOG_ERR` to something like `LOG_WARN` or even just `LOG_VERB`? It's not really an error, if things do work after that. Probably `LOG_WARN` is better as it indicates something that the user may not have intended.

lminiero:
Tnx, merging!

matthewmgamble:
Yep - that makes sense.  Change committed.

PR #618, opened 2016-09-19
SIP and SDP fixes for sdp-home

saghul:
From the https://github.com/meetecho/janus-gateway/pull/614 discussion.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2297,6 +2297,13 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			char *fixed_sdp = g_strdup(sip->sip_payload->pl_data);
 			JANUS_LOG(LOG_VERB, "Someone is inviting us in a call:\n%s", sip->sip_payload->pl_data);
 			janus_sip_sdp_process(session, sdp, FALSE);
+			/* Check if offer has neither audio nor video, fail with 488 */
+			if (!session->media.has_audio && !session->media.has_video) {
+				nua_respond(nh, 488, sip_status_phrase(488), TAG_END());
+				janus_sdp_free(sdp);
+				g_free(fixed_sdp);
+				break;
+			}
 			/* Send SDP to the browser */
 			json_t *jsep = json_pack("{ssss}", "type", "offer", "sdp", fixed_sdp);
 			json_t *call = json_object();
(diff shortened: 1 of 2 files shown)


sdp.c
```diff
@@ -600,6 +600,7 @@ int janus_sdp_anonymize(janus_sdp *anon) {
 		} else if(m->type == JANUS_SDP_VIDEO && m->port > 0) {
 			video++;
 			m->port = video == 1 ? 9 : 0;
+#ifdef HAVE_SCTP
```
lminiero: I think I had this ifdef removed because it resulted in missing media lines in case data channels support was not compiled. Or more likely this was happening in the merge method... anyway, whatever the case I'm not sure it's needed here, as the merge process sets the port to 0 anyway if you don't have DCs.

sdp.c
```diff
@@ -654,7 +656,7 @@ int janus_sdp_anonymize(janus_sdp *anon) {
 		GList *purged_ptypes = NULL;
 		while(tempA) {
 			janus_sdp_attribute *a = (janus_sdp_attribute *)tempA->data;
-			if(strstr(a->value, "red/90000") || strstr(a->value, "ulpfec/90000") || strstr(a->value, "rtx/90000")) {
+			if(a->value && (strstr(a->value, "red/90000") || strstr(a->value, "ulpfec/90000") || strstr(a->value, "rtx/90000"))) {
```
lminiero: I had not incorporated this change of yours from another PR as in my SDP utils you can't have a line like `a=name:`, that is, without a value after the semicolon: the parser returns an error. Anyway, now that I think about it this iterator may actually also traverse regular attributes with just a name (e.g., `a=sendrecv`), which means that check is indeed needed :+1:

plugins/janus_sip.c
```diff
@@ -2297,6 +2297,13 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			char *fixed_sdp = g_strdup(sip->sip_payload->pl_data);
 			JANUS_LOG(LOG_VERB, "Someone is inviting us in a call:\n%s", sip->sip_payload->pl_data);
 			janus_sip_sdp_process(session, sdp, FALSE);
+			/* Check if offer does not have audio or video, fail with 488 */
```
lminiero: Shouldn't this be, "fail with 488 in case neither is available"?

sdp.c
saghul: Ok, then we need another fix, actually declare the data variable :-) Will do.

sdp.c
saghul: Yep, we have some a=x-blinksomethingsomething :-)

plugins/janus_sip.c
saghul: Indeed!

saghul:
Updated, PTAL.

Issue #598, opened 2016-08-22
Janus.isWebrtcSupported() failing


revmischa:
I'm using Chrome viewing a page served from localhost (not file:///) which should be considered a secure origin. `Janus.isWebrtcSupported()` returns false, because window.getUserMedia isn't set. It should be checking `navigator.mediaDevices.getUserMedia` instead. This got renamed.

revmischa:
in janus.js it varies between getUserMedia(), window.getUserMedia(), navigator.mediaDevices.getUserMedia(), and navigator.mozGetUserMedia().

lminiero:
"Chrome" doesn't say much. Which version? I use it on localhost all the time and no problem here.

revmischa:
Chromium Version 51.0.2704.79 Ubuntu 15.10 (64-bit)
Are you

lminiero:
Port 8000.

revmischa:
Well window.getUserMedia is undef for me, navigator.mediaDevices.getUserMedia is not.

serdardurbaris:
I also had problem with Chrome. I take time out warning at server side.  I can't send video by chrome. But it works on yandex seamless. My chrome version on my computer 52.0.2743.116m (64bit)

karth295:
The `isWebrtcSupport()` check fails for me in Chrome version "52.0.2743.116 m" on Windows. If I ignore that check, I get an error in the console: "janus.js:197 WebRTC not supported by this browser", but I can send/receive video just fine.

lminiero:
@revmischa @karth295 @serdardurbaris please check the PR I referenced.

karth295:
@lminiero that PR didn't solve the issue for me--I just realized that I have a different issue than @revmischa . For me, the issue is that `window.RTCPeerConnection` is not set when I call `Janus.isWebrtcSupported()`.

It does get set later (I guess when `adapter.js` runs?), and `isWebrtcSupported()` returns true.

I'm calling `isWebrtcSupported` in the `Janus.init` callback, which is in the same place as `videoroomtest.js` (https://github.com/meetecho/janus-gateway/blob/master/html/videoroomtest.js#L73). The difference is that I'm calling it immediately, while videoroomtest waits for user input.

```
Janus.init
  callback: () ->
    if not Janus.isWebrtcSupported()
      alert("You must use Chrome, Firefox or Edge")
```

lminiero:
[...]

PR #619, opened 2016-09-19
Updates to janus.js/adapter.js

lminiero:
This patch tries to address issues like #598, those that are mostly caused by the outdated `adapter.js` we're using right now, and the fact `janus.js` doesn't always use what's defined in the adapter but uses some methods directly. The updated `adapter.js` should take care of that. Anyway, this forced us to change a couple of things in `janus.js` as well, that is:

1. redefining `Janus.isWebrtcSupported()` (which shouldn't fail anymore, now);
2. creating new `Janus.attachMediaStream()` and `Janus.reAttachMediaStream()`, since `attachMediaStream` and `reAttachMediaStream` have been removed from the adapter itself;
3. relying on the `adapter.browserDetails` object the adapter creates for info on browser and version, rather than using the old `webrtcDetectedBrowser` and parsing version ourselves.

2. and 3. forced us to update the JavaScript code of the demos too, as you can see, but nothing major. This seems to be working as expected on the browsers I have installed (Chrome 54, Firefox 48, Firefox Nightly 51), but that's all I have, so I leave it to you guys to test other browsers/versions (maybe even some older ones, if they're still shipped and around) and make sure it's still working fine. I'm particularly interested in checking whether or not Edge works again now, for instance, as I do believe some recent changes to `janus.js` broke that at the time.

Feedback welcome!

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -154,7 +154,8 @@ Janus.init = function(options) {
 				to.srcObject = from.srcObject;
 			}
 		};
-		// Detect tab close
+		// Detect tab close: make sure we don't loose existing onbeforeunload handlers
+		var oldOBF = window.onbeforeunload;
 		window.onbeforeunload = function() {
 			Janus.log("Closing window");
 			for(var s in Janus.sessions) {
@@ -164,6 +165,8 @@ Janus.init = function(options) {
 					Janus.sessions[s].destroy();
 				}
 			}
+			if(oldOBF && typeof oldOBF == "function")
+				oldOBF();
 		}
 		function addJsList(srcArray) {
 			if (!srcArray || !Array.isArray(srcArray) || srcArray.length == 0) {
--- a/html/janus.js
+++ b/html/janus.js
@@ -264,7 +264,7 @@ function Janus(gatewayCallbacks) {
 	}
 	var iceServers = gatewayCallbacks.iceServers;
 	if(iceServers === undefined || iceServers === null)
-		iceServers = [{"url": "stun:stun.l.google.com:19302"}];
+		iceServers = [{urls: "stun:stun.l.google.com:19302"}];
 	// Whether IPv6 candidates should be gathered
 	var ipv6Support = gatewayCallbacks.ipv6;
 	if(ipv6Support === undefined || ipv6Support === null)
--- a/html/janus.js
+++ b/html/janus.js
@@ -186,6 +186,7 @@ Janus.init = function(options) {
 			if(src === 'jquery.min.js') {
 				if(window.jQuery) {
 					// Already loaded
+					Janus.debug(src + " already loaded, skipping");
 					done();
 					return;
 				}
@@ -193,6 +194,7 @@ Janus.init = function(options) {
 			if(src === 'adapter.js') {
 				if(navigator.getUserMedia && navigator.mediaDevices.getUserMedia && window.RTCPeerConnection) {
 					// Already loaded
+					Janus.debug(src + " already loaded, skipping");
 					done();
 					return;
 				}
--- a/html/janus.js
+++ b/html/janus.js
@@ -99,7 +99,8 @@ Janus.init = function(options) {
 		Janus.listDevices = function(callback) {
 			callback = (typeof callback == "function") ? callback : Janus.noop;
 			if(navigator.mediaDevices) {
-				getUserMedia({ audio: true, video: true }, function(stream) {
+				navigator.mediaDevices.getUserMedia({ audio: true, video: true })
+				.then(function(stream) {
 					navigator.mediaDevices.enumerateDevices().then(function(devices) {
 						Janus.debug(devices);
 						callback(devices);
@@ -116,7 +117,8 @@ Janus.init = function(options) {
 							}
 						} catch(e) {}
 					});
-				}, function(err) {
+				})
+				.catch(function(err) {
 					Janus.error(err);
 					callback([]);
 				});
@@ -125,6 +127,33 @@ Janus.init = function(options) {
 				callback([]);
 			}
 		}
+		// Helper methods to attach/reattach a stream to a video element (previously part of adapter.js)
+		Janus.attachMediaStream = function(element, stream) {
+			if(adapter.browserDetails.browser === 'chrome') {
+				var chromever = adapter.browserDetails.version;
+				if(chromever >= 43) {
+					element.srcObject = stream;
+				} else if(typeof element.src !== 'undefined') {
+					element.src = URL.createObjectURL(stream);
+				} else {
+					Janus.error("Error attaching stream to element");
+				}
+			} else {
+				element.srcObject = stream;
+			}
+		};
+		Janus.reattachMediaStream = function(to, from) {
+			if(adapter.browserDetails.browser === 'chrome') {
+				var chromever = adapter.browserDetails.version;
+				if(chromever >= 43) {
+					to.srcObject = from.srcObject;
+				} else if(typeof element.src !== 'undefined') {
+					to.src = from.src;
+				}
+			} else {
+				to.srcObject = from.srcObject;
+			}
+		};
 		// Detect tab close
 		window.onbeforeunload = function() {
 			Janus.log("Closing window");
@@ -162,7 +191,7 @@ Janus.init = function(options) {
 				}
 			}
 			if(src === 'adapter.js') {
-				if(window.getUserMedia && window.RTCPeerConnection) {
+				if(navigator.getUserMedia && navigator.mediaDevices.getUserMedia && window.RTCPeerConnection) {
 					// Already loaded
 					done();
 					return;
@@ -185,7 +214,8 @@ Janus.init = function(options) {
 
 // Helper method to check whether WebRTC is supported by this browser
 Janus.isWebrtcSupported = function() {
-	return window.RTCPeerConnection && window.getUserMedia;
+	return window.RTCPeerConnection !== undefined && window.RTCPeerConnection !== null &&
+		navigator.getUserMedia !== undefined && navigator.getUserMedia !== null;
 };
 
 function Janus(gatewayCallbacks) {
@@ -1186,7 +1216,7 @@ function Janus(gatewayCallbacks) {
 		Janus.log("Preparing local SDP and gathering candidates (trickle=" + config.trickle + ")");
 		config.pc.onicecandidate = function(event) {
 			if (event.candidate == null ||
-					(webrtcDetectedBrowser === 'edge' && event.candidate.candidate.indexOf('endOfCandidates') > 0)) {
+					(adapter.browserDetails.browser === 'edge' && event.candidate.candidate.indexOf('endOfCandidates') > 0)) {
 				Janus.log("End of candidates.");
 				config.iceDone = true;
 				if(config.trickle === true) {
@@ -1414,8 +1444,8 @@ function Janus(gatewayCallbacks) {
 							.then(function(stream) { gsmCallback(null, stream); })
 							.catch(function(error) { pluginHandle.consentDialog(false); gsmCallback(error); });
 					};
-					if(window.navigator.userAgent.match('Chrome')) {
-						var chromever = parseInt(window.navigator.userAgent.match(/Chrome\/(.*) /)[1], 10);
+					if(adapter.browserDetails.browser === 'chrome') {
+						var chromever = adapter.browserDetails.version;
 						var maxver = 33;
 						if(window.navigator.userAgent.match('Linux'))
 							maxver = 35;	// "known" crash in chrome 34 and 35 on linux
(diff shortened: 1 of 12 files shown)


html/janus.js
```diff
@@ -125,6 +127,33 @@ Janus.init = function(options) {
...
+					to.srcObject = from.srcObject;
+				} else if(typeof element.src !== 'undefined') {
+					to.src = from.src;
+				}
+			} else {
+				to.srcObject = from.srcObject;
+			}
+		};
 		// Detect tab close
 		window.onbeforeunload = function() {
```
istolga: It is possible to have such function somewhere else. Please call window.onbeforeunload if such function is already defined.

html/janus.js
```diff
@@ -162,7 +191,7 @@ Janus.init = function(options) {
 				}
 			}
 			if(src === 'adapter.js') {
-				if(window.getUserMedia && window.RTCPeerConnection) {
+				if(navigator.getUserMedia && navigator.mediaDevices.getUserMedia && window.RTCPeerConnection) {
 					// Already loaded
```
istolga: Instead of comment //Already loaded can you please add some Janus.debug log that adapter is already loaded. The same for jquery. It is always nice to have such things for validation purposes especially when loading scripts in some sophisticated order and condition.

html/janus.js
lminiero: You mean something like `oldcb = window.onbeforeunload` if it exists, and then invoking `oldcb()` at the end of our callback?

html/janus.js
lminiero: Good point, tnx, I'll add that tomorrow!

html/janus.js
istolga: correct. From my experience using different libraries and js files in a big project it is always some onbeforeunload function somewhere.

istolga: Btw, I'm thinking it can be a loop(you call the existing function from your method, somebody does the same way), but I might be wrong. Need to spend more time to make sure no issues with the known frameworks.
Do you think it will be safe to just provide destroy method inside Janus and leave how they want to call it to them?

lminiero: I see the point you make, that might indeed cause trouble. I guess this can be an option. For instance, we already have a `destroyOnUnload`, that dictates whether you should automatically destroy a new session on unload. We might add an option to the `init` where you specify the behaviour in that regard, e.g.:

1. just set the `unbeforeunload`, as we do now;
2. set the `unbeforeunload` and invoke the previous one at the end;
3. don't set the `unbeforeunload` at all (which makes `destroyOnUnload` useless).

lminiero: That said, I don't believe the loop you mentioned should be an issue in our case: in fact, we set the `unbeforeunload` interlally in `janus.js`, so that's not something that developers are expected to modify, or even be able to overload, as the method we set it to is not public and not exposed via API. I believe the original change you suggested should be fine.

istolga: Good, thank you Lorenzo

joshdickson40:
Worked these changes into our front end code and looks good so far - will report back with any issues.

istolga:
I'm also using them at our initial project for WebRTC. Might be some time I'll report back, too many things now:)

cacheworks:
+1 👍 , I've played around with the demo in this commit and used the included adapter.js in an internal videoroom project without issues so far.

I did a quick test with Edge and the audiobridge plugin, the getUserMedia request and listing the participants worked, but it failed with ```{janus: "hangup", reason: "ICE failed", sender: 651918656720...", session_id: 489088885259..."}```

The server logs showed ```Skipping unsupported transport 'TCP' for media```

So it's worth noting that ice_tcp needs to be enabled in janus.cfg for ICE to work in Edge
```
ice_lite = true
ice_tcp = true
```

Unfortunately DTLS then fails, I'm not sure what the cause is there
When connecting it starts showing ```[11637394312190] DTLSv1_get_timeout: xxxx``` in the log, here's an example: https://gist.github.com/cacheworks/cd614622392687dffbb35b7caa66a9ec


After a while Edge then shows the following in the console;
```ORTC18615: ORTC RTCDtlsTransport: DTLS handshake failed. hr=c004e00f.```

**Comment by lminiero** [comment]
@cacheworks thanks, testing with Edge is one of the things I was really interested in. The TCP stuff is not really important: the `skipping` part is absolutely normal, and not even a warning anymore, so definitely no need to enable TCP or anything. If the updated janus/adapter works with Edge, then the real issue is either ICE or DTLS failing when negotiating the session. If you could share JavaScript logs and admin API for the failing handle that would help!

**Comment by cacheworks** [comment]
No problem! I'm not too familiar with ORTC / WebRTC interoperability yet but with H.264 and VP8 in development I figured I should start testing 👍 

Without enabling TCP Edge will receive a ```ICE failed``` hangup event right after joining the audiobridge, so I wasn't able to test without enabling it.

Here's the handle info from the admin API; https://gist.github.com/cacheworks/e37f41ebcff508e7116257d5c4815320
and here's the Edge console output;
https://gist.github.com/cacheworks/2248789004a6704d60bcc5670830811d (DTLS fails on [line 680](https://gist.github.com/cacheworks/2248789004a6704d60bcc5670830811d#file-edge-adapterjs-icelite-console-L680))





I tried the demo once more without ICE-lite / TCP as well, here's the handle info;
https://gist.github.com/cacheworks/7545167d149ed41d3e468e0e3ea05f87 
and here's the console output;
https://gist.github.com/cacheworks/41e392d73cca3b3cbff0e488e1c48088 (hangup occurs around [line 685](https://gist.github.com/cacheworks/41e392d73cca3b3cbff0e488e1c48088#file-edge-adapterjs-console-L685))

**Comment by lminiero** [comment]
The DTLS failure is a red herring, the real problem is ICE failing or anyway not connecting (the Admin API confirms that), it never gets to DTLS. It's most likely caused by the fact Edge is not sending srflx (STUN) candidates. I only see a host candidate (ptivate IP so useless), TCP candidates (that are discarded), IPv6 candidates (Janus started with IPv6 disabled so discarded) and, who knows why, prflx candidates (that we discard as they should be discovered by the recipient, not advertized by the client).

I guess Edge is not using the STUN/TURN candidates that are specified, which explains why srflx canditates are missing. I'll look into this on Monday.

**Comment by revmischa** [comment]
This appears to unbreak isWebrtcSupported for me

**Comment by revmischa** [comment]
I'm getting mixed results with ICE - Chromium on Ubuntu seems to stream my video okay, Chrome on my macbook streams ok, Chrome on my mac desktop will not display video no matter what, locally or remotely via ICE. I have no idea what the issue is.
```
(process:1): libnice-DEBUG: agent_recv_message_unlocked: Received 1 valid messages of length 70 from base socket 0x7f9b9c016070.
(process:1): libnice-DEBUG: Agent 0x7f9b9c01a020 : Packet received on local socket 23 from [67.169.38.x]:55781 (70 octets).
(process:1): libnice-stun-DEBUG: STUN error: RTP or other non-protocol packet!
(process:1): libnice-DEBUG: component_io_cb: 0x7f9b9c01a020: received 1 valid messages with 70 bytes
(process:1): libnice-DEBUG: agent_recv_message_unlocked: Received 0 valid messages of length 0 from base socket 0x7f9b9c016070.
```

**Comment by revmischa** [comment]
Tried enabling ice_lite

```
[2016/09/26 05:47:45:7893] NOTICE: lws_server_socket_service: wsi 0x7f9fb0000ad0 read 0
[2016/09/26 05:47:45:8324] NOTICE: lws_server_socket_service: wsi 0x7f9fb0000ad0 read 0
Creating new session: 2532931738991520
Creating new handle in session 2532931738991520: 3968487203402115
[3968487203402115] Creating ICE agent (ICE Lite mode, controlled)
[ERR] [ice.c:janus_ice_cb_component_state_changed:1473] [3968487203402115] ICE failed for component 1 in stream 1...
No WebRTC media anymore
```
ok??

lminiero:
@revmischa is this related to this PR in particular, or to your usage of Janus in general? If it's the latter, please open another issue or write a post to the Google group, as this discussion here is only for the updates to the JavaScript code.

lminiero:
@cacheworks can you check if the latest commit fixes the Edge issue for you? Looking at the adapter code, the Edge shim fixes the STUN/TURN servers syntax, but only goes through those identified by `urls`. Since that's the correct way of addressing them, while we still used the old `url` one, I fixed that, so that should work now.

cacheworks:
I just tested again, I made sure janus.js updated and cache was clear, but it seems to behave like it did before, it also still works in Firefox and Chrome so didn't seem to break anything either.

Here's the handle info after the ICE Failed hangup was received;
https://gist.github.com/cacheworks/24933f37093a94f5fe77656b62d29b63
And here's the Edge console output to go along with that;
https://gist.github.com/cacheworks/4115ccfd81109fadda0cf2a638d234a2

Here's the same handle info with ICE lite enabled on the server; (This does include the client's IP address 9.x.x.x as a remote-candidate this time)
https://gist.github.com/cacheworks/59802f9a65d3429601080374773b0108
and the console output;
https://gist.github.com/cacheworks/54e7db690464453f37a4b6e1aa68068e

lminiero:
I still don't see any `srflx` candidate there, so that explains the failure.

One thing I noticed, though, is that I do see [trickle candidates being sent](https://gist.github.com/cacheworks/4115ccfd81109fadda0cf2a638d234a2#file-edge_console-232545f-L532) in the Edge console, but according to the output they're all [empty](https://gist.github.com/cacheworks/4115ccfd81109fadda0cf2a638d234a2#file-edge_console-232545f-L539). Can you add something like this:

    Janus.log("Got trickle candidate:", event)

more ore less before [this line here](https://github.com/meetecho/janus-gateway/blob/232545f9f7056d6c524982db02046bb7c60bcb41/html/janus.js#L1237)? I'm wondering whether the syntax they use in the `onicecandidate` callback is different, and so we're not getting the info correctly for Edge.

cacheworks:
Looks like the candidates are indeed all empty;
https://gist.github.com/cacheworks/c821a0c518c8892366cb4061f77b1b65

lminiero:
No, it's just not printing the content for you:

    Got trickle candidate: [object Event]

Try changing the line I pasted before to:

    Janus.log("Got trickle candidate:", JSON.stringify(event))

cacheworks:
Ah indeed sorry.. I should've had a proper look before copying, I thought
```   "Got trickle candidate:"
   {
      [functions]: ,
      __proto__: { },
      AT_TARGET: 2,
      bubbles: false,
      BUBBLING_PHASE: 3,
      cancelable: false,
      cancelBubble: false,
      candidate: { },
      CAPTURING_PHASE: 1,
      currentTarget: null,
      defaultPrevented: false,
      eventPhase: 0,
      isTrusted: false,
      returnValue: true,
      srcElement: { },
      target: { },
      timeStamp: 1474976066360,
      type: "icecandidate"
   }
``` was already fully expanded.

Here's the stringified output;
https://gist.github.com/cacheworks/929d7965b6c9f61c0c2136b3b5f739d9

Looks like the only strflx candidate contains the client's internal IP address
```Got trickle candidate: {"candidate":{"sdpMid":"pewvgir6kf","sdpMLineIndex":0,"candidate":"candidate:4 1 TCP 1684797951 192.168.1.133 22150 typ srflx raddr 192.168.1.133 rport 22150 tcptype active"}}```

lminiero:
Yeah, that's actually a TCP candidate and not UDP. I guess it's still not using the STUN server as it should. Can you also add a few more lines to the adapter.js so that we can see why it's ignoring them?

Before [this](https://github.com/meetecho/janus-gateway/blob/232545f9f7056d6c524982db02046bb7c60bcb41/html/adapter.js#L1134):

    console.log("iceServers:", config.iceServers)

and before [this](https://github.com/meetecho/janus-gateway/blob/232545f9f7056d6c524982db02046bb7c60bcb41/html/adapter.js#L1157)

    console.log("iceOptions:", this.iceOptions)

It might also help adding this line before [this](https://github.com/meetecho/janus-gateway/blob/232545f9f7056d6c524982db02046bb7c60bcb41/html/adapter.js#L1292)

    console.log("iceOptions(2):", self.iceOptions)

just to see what's it actually using when it's time to create the ICE/DTLS transport. Sorry for asking you all these debug lines, but I don't have a VM with Edge installed anymore, and I don't have much time to install and configure one right now. Thanks!

cacheworks:
No worries, happy to help, it's the least I can do with how helpful you've been with issues / pull requests 👍 

Here's the ouput;
https://gist.github.com/cacheworks/3ac69427813993ea52960a9a974c253d

```iceServers: [{"urls":"stun:stun.l.google.com:19302"}]```

```iceOptions: {"gatherPolicy":"all","iceServers":[{"urls":"stun:stun.l.google.com:19302"}]}```

```iceOptions(2): {"gatherPolicy":"all","iceServers":[{"urls":"stun:stun.l.google.com:19302"}]}```

lminiero:
Then no idea on why it doesn't like it. Have you checked if there's some firewall that's preventing stuff to go through? I haven't used Windows in like 10 years...

cacheworks:
Yeah looking at https://msdn.microsoft.com/en-us/library/mt433107(v=vs.85).aspx the output all seems pretty sane.

I've disabled Winodws firewall and tried using several stun servers I found but but it still doesn't return a proper srflx candidate.

I also tried a machine with an Edge version from before the W10 anniversary update, and tried tethering over 4G but so far it doesn't change aanything
I'll poke around a bit more adding debug lines to see if I can spot something that doens't get set properly

lminiero:
This seems to be working as expected, so I'll merge. We can track Edge problems/support in a separate issue/PR.

PR #621, opened 2016-09-20
Added missing session lock and expanded another lock.

andreasg123:
There should be a lock for all manipulations of `session->ice_handles`.  I added the missing lock.  I expanded another lock because the test `g_hash_table_size(session->ice_handles) > 0` should be protected by the lock, too.


lminiero:
Thanks, merging!

PR #623, opened 2016-09-21
Moved janus_mutex_lock(&session->mutex) outside of janus_ice_handle_destroy

andreasg123:
`janus_ice_handle_destroy` locked `session->mutex` again after it was locked in #621.  As the new lock is needed, it is easiest to require the lock before calling `janus_ice_handle_destroy`.  `janus_ice_handle_destroy` locks `old_plugin_sessions_mutex` but the same lock nesting already happens in `janus_ice_handle_attach_plugin` so that this shouldn't cause a deadlock.


lminiero:
Good catch, tnx!

PR #625, opened 2016-09-22
Functions for manipulating ice_handles

andreasg123:
I mentioned earlier that it would be nice if manipulation of `ice_handles` happened in the same file.  I created new functions in `ice.c` that do that (and also remove some code duplication).

I didn't change anything other than moving the code fragments into function so that this shouldn't cause any problems.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -826,10 +826,10 @@ janus_ice_handle *janus_ice_handle_create(void *core_session) {
 	guint64 handle_id = 0;
 	while(handle_id == 0) {
 		handle_id = janus_random_uint64();
-		handle = janus_ice_handle_find(core_session, handle_id);
+		handle = janus_session_handles_find(session, handle_id);
 		if(handle != NULL) {
 			/* Handle ID already taken, try another one */
-			janus_refcount_decrease(&handle->ref);	/* janus_ice_handle_find increases it */
+			janus_refcount_decrease(&handle->ref);	/* janus_session_handles_find increases it */
 			handle_id = 0;
 		}
 	}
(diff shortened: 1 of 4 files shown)


lminiero:
This looks helpful, although from a logical perspective I'm not sure whether it makes more sense to do all the `ice_handles` manipulations in `ice.c` or in `janus.c` instead. In fact, that's a table that's actually part of the `janus_session` object, and as such not really a manipulation on a handle, which would indeed belong to `ice.c`. This ambiguity is also reflected in the name of the methods, as for instance `janus_ice_handle_clear` seems to suggest this clears a `janus_ice_handle` instance, when instead, it's a method for removing all the handles of a session.

This is of course just a matter of taste, as the functionality would be the same wherever we put it, and I believe you did the right thing by grouping the features and reducing the code in general! What do you think about this?

andreasg123:
I think that it would make more sense to manipulate `ice_handles` in `janus.c` because it is a member of `janus_session` and to call `janus_ice_handle_destroy` defined in `ice.c`.  I only moved the additional functions to `ice.c` because the current design seemed to call for that.

If you like, I can move the newly defined functions and some others to `janus.c`.  What should the prefix for those functions be, `janus_ice_handles_`?

lminiero:
I'd keep `janus_session_` as a prefix in general, as it's sessions related, and then add a `handles` there, so maybe `janus_session_handles_` --> `clear/remove/ecc.`. Makes sense?

andreasg123:
I renamed and moved the functions as discussed.  I amended my commit to avoid a confusing history.

lminiero:
Thanks! I'll review when I get back to the office and merge!

lminiero:
Thanks for this further fix, I'll definitely review tomorrow in the morning! (and if I forget again, feel free to insult me :smile: )

lminiero:
Looks good, thanks, merging!

PR #628, opened 2016-09-26
Maintain refcount for members of janus_videoroom.participants

andreasg123:
This PR increases and decreases `refcount` for `janus_videoroom_publisher` as it is inserted and removed from `janus_videoroom.participants`.  Otherwise, `janus_videoroom_notify_participants` may be called for publishers that are already freed.

This may address the [crash](https://github.com/meetecho/janus-gateway/pull/403#issuecomment-249386886) reported by @cacheworks.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -454,6 +454,12 @@ static void janus_videoroom_subscriber_free(const janus_refcount *s_ref) {
 	g_free(s);
 }
 
+static void janus_videoroom_publisher_dereference(janus_videoroom_publisher *p) {
+	if(!p)
+		return;
+	janus_refcount_decrease(&p->ref);
+}
+
 static void janus_videoroom_publisher_destroy(janus_videoroom_publisher *p) {
 	if(!p)
 		return;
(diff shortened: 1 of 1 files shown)


lminiero:
Looks fine to me! Thinking about it, maybe a cleaner way of addressing removals from the hashtable is to use a `GDestroyNotify` that decrements the reference in `g_hash_table_new_full`? This way, however the element is removed from the table, it's reference is decremented, and it can only happen once. I believe we use the same approach in some other tables in the code (but not all of them).

andreasg123:
That's a good idea.  I also just did that in my PR for `ice_handles`.

lminiero:
Thanks! I'll take care of doing the same for all the other places where this might be useful. I'm leaving office now so I'll do a final review on both your patches when I get back tomorrow.

lminiero:
Love this more compact approach! I guess it's even safer as it's basically all in a single check. A further fix that may be worthwhile to spread around in other areas of the code, I'll work on that.

lminiero:
Looks good here too, merging!

PR #634, opened 2016-09-28
Increase publisher refcount before unlocking videoroom->mutex

andreasg123:
In inspecting the code, I noticed a missing publisher refcount decrease when trying to remove a non-existing forwarder.  Also, there is no need to check first if the forwarder is there because `g_hash_table_remove` tells you that.

I moved two refcount increases closer to the place where the publisher is referenced to reduce confusion.

Finally, I found two places with potential race conditions where `videoroom->mutex` was unlocked before the publisher refcount was increased.  That would allow another thread to remove the publisher from `room->participants` and to free it before the increased refcount would prevent that.  An alternative approach would be to create a function similar to `janus_session_handles_find` that would increase the refcount before returning the publisher.

However, these changes do not address the crash reported by @cacheworks.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1542,20 +1542,22 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 		guint32 video_handle = 0;
 		if(audio_port > 0) {
 			audio_handle = janus_rtp_forwarder_add_helper(publisher, host, audio_port, 0);
+			if(audio_handle)
+				janus_refcount_increase(&publisher->ref);
 		}
 		if(video_port > 0) {
 			video_handle = janus_rtp_forwarder_add_helper(publisher, host, video_port, 1);
+			if(video_handle) {
+				janus_refcount_increase(&publisher->ref);
 		}
 		janus_mutex_unlock(&videoroom->mutex);
 		response = json_object();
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for looking into this, it definitely makes it much more stable. And I take the opportunity to thank you again for all the help you're giving on the reference counters branch: these last few weeks are the first time in months I've seen it really closer to become mergeable, which is great!

plugins/janus_videoroom.c
```diff
@@ -1542,20 +1542,22 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session
 		guint32 video_handle = 0;
 		if(audio_port > 0) {
 			audio_handle = janus_rtp_forwarder_add_helper(publisher, host, audio_port, 0);
+			if(audio_handle)
+				janus_refcount_increase(&publisher->ref);
 		}
 		if(video_port > 0) {
 			video_handle = janus_rtp_forwarder_add_helper(publisher, host, video_port, 1);
+			if(video_handle) {
+				janus_refcount_increase(&publisher->ref);
```
cacheworks: ```if(video_handle) {``` opens a conditional here but I don't see it being closed, should this be ```if(video_handle)``` instead?

lminiero: Good catch! I fixed the typo in the PR as I had merged it already.

PR #639, opened 2016-09-30
Use g_clear_pointer to decrease refcounts

andreasg123:
I replaced almost all calls to `janus_refcount_decrease` with `g_clear_pointer` with the appropriate `GDestroyNotify`.

I see that every time a subscriber is added, the refcount for `publisher->session` is increased (and decreased when the subscriber is removed).  I don't know why that happens because I don't see a reference from the subscriber to the publisher's session.

Even if this is needed, I would suggest that both subscriber and publisher should increase the refcount for their own session when they get created.  That should happen anyway to avoid having an uncounted reference.


andreasg123:
I found a [missing room refcount decrease](https://github.com/meetecho/janus-gateway/pull/639/commits/9f8805f7e9ad289ee785c0cd8ea2879dfe6266ca#diff-f302a3ccc39e037fccf44e207332553cR2380).  You should put that in even if you don't want to merge the PR.

andreasg123:
Might as well take advantage of refcounts by decreasing the refcount for the video room in the error block of `janus_videoroom_handle_message`.  This requires that the `videoroom` variable is set to `NULL` and doesn't get used after it is assigned to the publisher or subscriber.

PR #640, opened 2016-10-01
Handle race condition for subscribers during janus_videoroom_hangup_media

andreasg123:
In the crash reported by @cacheworks, a packet was sent to subscribers in `janus_videoroom_incoming_rtcp` while `janus_videoroom_hangup_media` was running in another thread.  Subscribers have to be locked in both threads to avoid a race condition.  There are other fields that have to be protected, e.g, `video_active`, that are now modified inside the lock of `subscribers_mutex`.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1893,6 +1893,8 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char 
 	janus_videoroom_publisher *participant = (janus_videoroom_publisher *)session->participant;
 	if(participant == NULL || g_atomic_int_get(&participant->destroyed))
 		return;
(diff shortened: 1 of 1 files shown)


cacheworks:
I gave this commit a try and ran into the following crashes last night.
Leaving this here as it's not merged with the other PRs yet but it might indicate an issue to keep in mind when merging them @lminiero 

It crashed twice with the following output;
https://gist.github.com/cacheworks/bc0fc1bfd1f3f7640eba8ceaaf666931

and twice with;
https://gist.github.com/cacheworks/d5d33397a8739229cd4bb8a15650c2c3

andreasg123:
The second crash is addressed by #642.  The first crash is new.  It uses a previously freed web socket.  I'm not familiar with that part of the code.

PR #642, opened 2016-10-02
Protect session publisher with a lock and increase refcount

andreasg123:
Use a mutex to protect access to the session publisher. `janus_videoroom_session_get_publisher` returns the session publisher and increases the refcount.  This should prevent the publisher from getting freed while it is still being used.

I didn't do the same for the session subscriber.  So far, no race conditions have appeared for that part but one should have a closer look to see if it needs the same treatment.

There should be a merge conflict with #640 in `janus_videoroom_incoming_rtp` (the new approach is better) and there may also be merge conflicts with #639.


lminiero:
@andreasg123 which PR (639, 640, 642) do you suggest me review/merge first? So that I don't make your life harder with handling potential conflicts there.

andreasg123:
I kept the PRs separate for easier reviewing.  If you want them all, I can merge them (or let you do it).

PR #643, opened 2016-10-03
Combine recent PRs for videoroom

andreasg123:
This is the merge of #639, #640, and #642.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -365,6 +365,7 @@ typedef struct janus_videoroom_session {
 	gboolean stopping;
 	volatile gint hangingup;
 	volatile gint destroyed;
+	janus_mutex mutex;
 	janus_refcount ref;
 } janus_videoroom_session;
 static GHashTable *sessions;
@@ -441,6 +442,11 @@ typedef struct janus_videoroom_rtp_relay_packet {
 
 
 /* Freeing stuff */
+static void janus_videoroom_subscriber_dereference(janus_videoroom_subscriber *s) {
+	/* No need to test for NULL in conjunction with g_clear_pointer. */
+	janus_refcount_decrease(&s->ref);
+}
+
 static void janus_videoroom_subscriber_destroy(janus_videoroom_subscriber *s) {
 	if(s && g_atomic_int_compare_and_exchange(&s->destroyed, 0, 1))
 		janus_refcount_decrease(&s->ref);
@@ -453,8 +459,8 @@ static void janus_videoroom_subscriber_free(const janus_refcount *s_ref) {
 }
 
 static void janus_videoroom_publisher_dereference(janus_videoroom_publisher *p) {
-	if(p)
-		janus_refcount_decrease(&p->ref);
+	/* This is used by g_pointer_clear and g_hash_table_new_full so that NULL is only possible if that was inserted into the hash table. */
+	janus_refcount_decrease(&p->ref);
 }
 
 static void janus_videoroom_publisher_destroy(janus_videoroom_publisher *p) {
@@ -482,6 +488,10 @@ static void janus_videoroom_publisher_free(const janus_refcount *p_ref) {
 	g_free(p);
 }
 
+static void janus_videoroom_session_dereference(janus_videoroom_session *session) {
+	janus_refcount_decrease(&session->ref);
+}
+
 static void janus_videoroom_session_destroy(janus_videoroom_session *session) {
 	if(session && g_atomic_int_compare_and_exchange(&session->destroyed, 0, 1))
 		janus_refcount_decrease(&session->ref);
@@ -492,7 +502,12 @@ static void janus_videoroom_session_free(const janus_refcount *session_ref) {
 	/* Remove the reference to the core plugin session */
 	janus_refcount_decrease(&session->handle->ref);
 	/* This session can be destroyed, free all the resources */
+	janus_mutex_destroy(&session->mutex);
 	g_free(session);
+}
+
+static void janus_videoroom_room_dereference(janus_videoroom *room) {
+	janus_refcount_decrease(&room->ref);
 }
 
 static void janus_videoroom_room_destroy(janus_videoroom *room) {
@@ -919,6 +934,7 @@ void janus_videoroom_create_session(janus_plugin_session *handle, int *error) {
 	g_atomic_int_set(&session->hangingup, 0);
 	g_atomic_int_set(&session->destroyed, 0);
 	handle->plugin_handle = session;
+	janus_mutex_init(&session->mutex);
 	janus_refcount_init(&session->ref, janus_videoroom_session_free);
 
 	janus_mutex_lock(&sessions_mutex);
@@ -926,6 +942,15 @@ void janus_videoroom_create_session(janus_plugin_session *handle, int *error) {
 	janus_mutex_unlock(&sessions_mutex);
 
 	return;
+}
+
+static janus_videoroom_publisher *janus_videoroom_session_get_publisher(janus_videoroom_session *session) {
+	janus_mutex_lock(&session->mutex);
+	janus_videoroom_publisher *publisher = (janus_videoroom_publisher *)session->participant;
+	if(publisher)
+		janus_refcount_increase(&publisher->ref);
+	janus_mutex_unlock(&session->mutex);
+	return publisher;
 }
 
 static void janus_videoroom_notify_participants(janus_videoroom_publisher *participant, json_t *msg) {
@@ -944,21 +969,19 @@ static void janus_videoroom_notify_participants(janus_videoroom_publisher *parti
 }
 
 static void janus_videoroom_leave_or_unpublish(janus_videoroom_publisher *participant, gboolean is_leaving) {
+	/* we need to check if the room still exists, may have been destroyed already */
+	if(!participant->room || participant->room->destroyed)
+		return;
 	json_t *event = json_object();
+	janus_mutex_lock(&participant->room->mutex);
 	json_object_set_new(event, "videoroom", json_string("event"));
 	json_object_set_new(event, "room", json_integer(participant->room->room_id));
 	json_object_set_new(event, is_leaving ? "leaving" : "unpublished", json_integer(participant->user_id));
-	/* we need to check if the room still exists, may have been destroyed already */
-	if(participant->room) {
-		if(!participant->room->destroyed) {
-			janus_mutex_lock(&participant->room->mutex);
-			janus_videoroom_notify_participants(participant, event);
-			if(is_leaving) {
-				g_hash_table_remove(participant->room->participants, &participant->user_id);
-			}
-			janus_mutex_unlock(&participant->room->mutex);
-		}
+	janus_videoroom_notify_participants(participant, event);
+	if(is_leaving) {
+		g_hash_table_remove(participant->room->participants, &participant->user_id);
 	}
+	janus_mutex_unlock(&participant->room->mutex);
 	json_decref(event);
 }
 
@@ -987,11 +1010,15 @@ void janus_videoroom_destroy_session(janus_plugin_session *handle, int *error) {
 		if(session->participant_type == janus_videoroom_p_type_publisher) {
 			/* Get rid of publisher */
 			janus_videoroom_publisher *p = (janus_videoroom_publisher *)session->participant;
+			janus_mutex_lock(&session->mutex);
 			session->participant = NULL;
+			janus_mutex_unlock(&session->mutex);
+			janus_videoroom_leave_or_unpublish(p, TRUE);
 			if(p->room) {
+				/* Don't clear p->room.  Another thread calls janus_videoroom_leave_or_unpublish,
+					 too, and there is no mutex to protect this change. */
 				janus_refcount_decrease(&p->room->ref);
 			}
-			janus_videoroom_leave_or_unpublish(p, TRUE);
 			janus_videoroom_publisher_destroy(p);
 		} else if(session->participant_type == janus_videoroom_p_type_subscriber) {
 			janus_videoroom_subscriber *s = (janus_videoroom_subscriber *)session->participant;
@@ -1025,7 +1052,7 @@ json_t *janus_videoroom_query_session(janus_plugin_session *handle) {
 			json_object_set_new(info, "type", json_string("none"));
 		} else if(session->participant_type == janus_videoroom_p_type_publisher) {
 			json_object_set_new(info, "type", json_string("publisher"));
-			janus_videoroom_publisher *participant = (janus_videoroom_publisher *)session->participant;
+			janus_videoroom_publisher *participant = janus_videoroom_session_get_publisher(session);
 			if(participant) {
 				janus_videoroom *room = participant->room; 
 				json_object_set_new(info, "room", room ? json_integer(room->room_id) : NULL);
@@ -1048,6 +1075,7 @@ json_t *janus_videoroom_query_session(janus_plugin_session *handle) {
 						json_object_set_new(recording, "video", json_string(participant->vrc->filename));
 					json_object_set_new(info, "recording", recording);
 				}
+				g_clear_pointer(&participant, janus_videoroom_publisher_dereference);
 			}
 		} else if(session->participant_type == janus_videoroom_p_type_subscriber) {
 			json_object_set_new(info, "type", json_string("subscriber"));
@@ -1071,7 +1099,7 @@ json_t *janus_videoroom_query_session(janus_plugin_session *handle) {
 	}
 	json_object_set_new(info, "hangingup", json_integer(g_atomic_int_get(&session->hangingup)));
 	json_object_set_new(info, "destroyed", json_integer(g_atomic_int_get(&session->destroyed)));
-	janus_refcount_decrease(&session->ref);
+	g_clear_pointer(&session, janus_videoroom_session_dereference);
 	return info;
 }
 
@@ -1461,8 +1489,8 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 			janus_refcount_increase(&room->ref);
 			if(room->is_private) {
 				/* Skip private room */
-				janus_refcount_decrease(&room->ref);
 				JANUS_LOG(LOG_VERB, "Skipping private room '%s'\n", room->room_name);
+				g_clear_pointer(&room, janus_videoroom_room_dereference);
 				continue;
 			}
 			if(!g_atomic_int_get(&room->destroyed)) {
@@ -1480,7 +1508,7 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 				json_object_set_new(rl, "num_participants", json_integer(g_hash_table_size(room->participants)));
 				json_array_append_new(list, rl);
 			}
-			janus_refcount_decrease(&room->ref);
+			g_clear_pointer(&room, janus_videoroom_room_dereference);
 		}
 		janus_mutex_unlock(&rooms_mutex);
 		response = json_object();
@@ -1520,8 +1548,8 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 		janus_mutex_lock(&videoroom->mutex);
 		janus_videoroom_publisher *publisher = g_hash_table_lookup(videoroom->participants, &publisher_id);
 		if(publisher == NULL) {
-			janus_refcount_decrease(&videoroom->ref);
 			janus_mutex_unlock(&videoroom->mutex);
+			g_clear_pointer(&videoroom, janus_videoroom_room_dereference);
 			JANUS_LOG(LOG_ERR, "No such publisher (%"SCNu64")\n", publisher_id);
 			error_code = JANUS_VIDEOROOM_ERROR_NO_SUCH_FEED;
 			g_snprintf(error_cause, 512, "No such feed (%"SCNu64")", publisher_id);
@@ -1531,9 +1559,9 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 		if(publisher->udp_sock <= 0) {
 			publisher->udp_sock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
 			if(publisher->udp_sock <= 0) {
-				janus_refcount_decrease(&publisher->ref);
-				janus_refcount_decrease(&videoroom->ref);
+				g_clear_pointer(&publisher, janus_videoroom_publisher_dereference);
 				janus_mutex_unlock(&videoroom->mutex);
+				g_clear_pointer(&videoroom, janus_videoroom_room_dereference);
 				JANUS_LOG(LOG_ERR, "Could not open UDP socket for rtp stream for publisher (%"SCNu64")\n", publisher_id);
 				error_code = JANUS_VIDEOROOM_ERROR_UNKNOWN_ERROR;
 				g_snprintf(error_cause, 512, "Could not open UDP socket for rtp stream");
@@ -1571,8 +1599,9 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 			json_object_set_new(rtp_stream, "video_stream_id", json_integer(video_handle));
 			json_object_set_new(rtp_stream, "video", json_integer(video_port));
 		}
-		janus_refcount_decrease(&publisher->ref);	/* These two unrefs are related to the message handling */
-		janus_refcount_decrease(&videoroom->ref);
+		/* These two unrefs are related to the message handling */
+		g_clear_pointer(&publisher, janus_videoroom_publisher_dereference);
+		g_clear_pointer(&videoroom, janus_videoroom_room_dereference);
 		json_object_set_new(rtp_stream, "host", json_string(host));
 		json_object_set_new(response, "publisher_id", json_integer(publisher_id));
 		json_object_set_new(response, "rtp_stream", rtp_stream);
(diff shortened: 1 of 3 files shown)


lminiero:
@andreasg123 thanks for that! I'll try to review either tomorrow or the day after that, and come back to you guys ASAP.

cacheworks:
I got curious and checked out out the videoroom-combined branch in @andreasg123 's repo.

When trying to publish in the videoroom demo it gets stuck while publishing
I went back through the commits and see that this starts happening in commit  e154a152fafe51028e68ebda4136058cfec0bbef 
in commit ef0e54f804a38f9f6ab799ed913c37d498a26314 publishing does work

Here's a full log of trying to publish with 2 users using the videoroomtest demo using commit e154a152fafe51028e68ebda4136058cfec0bbef 
https://gist.github.com/cacheworks/7c40ae7c8e77d63eb1b987f56db832e4

If I keep the tabs open keepalives will still show up in the server log, but it never gets past publishing

andreasg123:
@cacheworks, good catch.  I found my mistake.

cacheworks:
:+1: Publishing indeed works again with that commit, I'll start testing some more

Edit: I now get the following crash when leaving the videoroom;
https://gist.github.com/cacheworks/fbb8e9fdd5b818a7819625e75cd78968

andreasg123:
It looks like I was a little overzealous in clearing the publisher room before I was done with it.  There may be other places like but I won't look before tomorrow.

On the plus side, this revealed a problem in the test in `janus_videoroom_leave_or_unpublish`.

andreasg123:
I reviewed all other places where I call `g_clear_pointer` with a member of a struct.  In none of those places, the member gets used after it is cleared.  There shouldn't be any more problems like the two that @cacheworks discovered.

plugins/janus_videoroom.c
```diff
@@ -441,6 +442,11 @@ typedef struct janus_videoroom_rtp_relay_packet {
 
 
 /* Freeing stuff */
+static void janus_videoroom_subscriber_dereference(janus_videoroom_subscriber *s) {
+	if(s)
```
saghul: What can lead to `s` being NULL here? According to [the documentation](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-clear-pointer) `g_clear_pointer` won't call the destroy function if the passed pointer is NULL, so how is it possible to end in that situation?

andreasg123: I checked the source for [GHashTable](https://github.com/GNOME/glib/blob/8c9be18bd3dff42a2352c2dac6b8f17b243d2980/glib/ghash.c#L903) and there is nothing that prevents `NULL` values.  Still, we could remove the test if we are sure that we would never insert a `NULL` publisher into the participants table.

saghul: IMHO I'd replace the test with an assert.  When I saw it it made me wonder how it would be possible.  I'm all for defensive programming, but being overly defensive can make one a bit confused :-)

lminiero:
Oh God please no, no asserts... I hate it enough when libraries decide to assert and abort my application without giving me the chance to handle the error case. Not sure about removing those checks, they seem quick and harmless enough.

plugins/janus_videoroom.c
lminiero: Oh God please no, no asserts... I hate it enough when libraries decide to assert and abort my application without giving me the chance to handle the error case. Not sure about removing those checks, they seem quick and harmless enough.

cacheworks:
I got the following crash after 10 hours using commit 1bb12c382cd1cb53abd79d9536db14164f5c2b23
https://gist.github.com/cacheworks/4a11526b5fd9817ecf50743b500f6e4e

Looks like it's pointing to the same line as before:
```json_object_set_new(event, "room", json_integer(participant->room->room_id));```

Here's output including the reference counter debug;
https://gist.github.com/cacheworks/d525acfb3eac84ecdbb70bafb8db922d'

Thank you all for taking time to look at yet another log. I started reading and watching as many C tutorials as I can so I can help reviewing later. It's starting to make sense but these locks and threads are hard to keep track of 👍  😄

**Comment by lminiero** [comment]
I'll try to review as soon as possible, although from what I've understood there are other issues this patch needs to address first?

**Comment by andreasg123** [comment]
The crash reported by @cacheworks is still the same race condition between "detach" and "unpublish".  I now don't clear `publisher->room` when decreasing the refcount.  This could cause a problem if the same chain of events happens after the room is destroyed because the room refcount would be zero.  I don't know if the client would send "unpublish" after the room is destroyed.

An alternative would be to protect access to `publisher->room` with a mutex so that clearing it wouldn't cause a crash in `janus_videoroom_leave_or_unpublish`.

Other than this crash I'm not aware of any issues that need to be addressed.

**Comment by cacheworks** [comment]
Thanks for clarifying @andreasg123, I don't currently delete videorooms with users in them so I shouldn't run into that issue.
But if there's a possibility for a client to trigger a crash when they are deleted, albeit accidentally or maliciously, it's probably a good idea to be on the safe side and add a lock right? Or are there performance considerations with using a lock in there to opt for not using it?

I recompiled and have been running this last commit for an hour now, I'll keep you posted on how it handles the peak hours 👍

**Comment by andreasg123** [comment]
It might make sense to protect access to `janus_videoroom_publisher.room` with a mutex, especially if there isn't much performance impact.  This could be done in a fashion similar to `janus_videoroom_session_get_publisher` [introduced in this PR](https://github.com/andreasg123/janus-gateway/blob/9690a9f1621a9c5ebb2a901fdd00caaf12e934f1/plugins/janus_videoroom.c#L946), together with a clear function:

	static void janus_videoroom_publisher_clear_room(janus_videoroom_publisher *publisher) {
		janus_mutex_lock(&publisher->mutex);
		g_clear_pointer(&publisher->room, janus_videoroom_room_dereference);
		janus_mutex_unlock(&publisher->mutex);
	}

I tried to find some data on the performance of a mutex.  Supposedly, [an uncontested mutex has little overhead](http://stackoverflow.com/questions/1277627/overhead-of-pthread-mutexes).  The current version of GLib [implements GMutex with futex](https://bugzilla.gnome.org/show_bug.cgi?id=731986).

**Comment by cacheworks** [comment]
Great, I appreciate you taking the time to explain, seeing these examples help to get familiar with the code

Unfortunately I've returned with a new crash using commit 9690a9f1621a9c5ebb2a901fdd00caaf12e934f1
https://gist.github.com/cacheworks/3e0114ae689f66a7f25d42be56ca8c34

This time it points to a g_main_loop_quit in [ice.c:962](https://github.com/andreasg123/janus-gateway/blob/9690a9f1621a9c5ebb2a901fdd00caaf12e934f1/ice.c#L962) though, and searching through the diff for 'handle' I don't see any changes here that would invoke that ```janus_ice_handle_destroy``` directly, or could this be due to a dereference?

Do you think it's a coincidence that this issue only popped up in this commit and should I create a new issue for it. Or do you see anything in this PR that might've caused this?

I'll keep it using this for now to see if it crashes again with the same output in the next few hours

**Comment by andreasg123** [comment]
It looks like the `janus_ice_handle` points to a bad address.  I don't know why AddressSanitizer doesn't show it as previously freed.  This could be related to 16d1e08.  @cacheworks, do you have the refcount log for this?

**Comment by cacheworks** [comment]
Ah sure do, forgot to include that this time; https://gist.github.com/cacheworks/31ce06407a68027632356bb7b5b0e8c7

**Comment by andreasg123** [comment]
Here is the relevant part of the log:

	[janus.c:janus_session_handles_find:508:increase] 0x612003b23ad0 (6)   ***
	Detaching handle from JANUS VideoRoom plugin
	[janus.plugin.videoroom-0x604001f40210] No WebRTC media anymore
	[plugins/janus_videoroom.c:janus_videoroom_destroy_session:1026:decrease] 0x60d00000c630 (260)
	[plugins/janus_videoroom.c:janus_videoroom_subscriber_destroy:452:decrease] 0x60b0019e43a0 (0)
	[plugins/janus_videoroom.c:janus_videoroom_session_destroy:497:decrease] 0x608000d9d3f0 (0)
	[plugins/janus_videoroom.c:janus_videoroom_session_free:503:decrease] 0x604001f40228 (1)
	[ice.c:janus_ice_handle_destroy:976:decrease] 0x612003b23ad0 (5)   ***
	[janus.c:janus_ice_handle_dereference:355:decrease] 0x612003b23ad0 (4)   ***
	[janus.c:janus_process_incoming_request:1439:decrease] 0x612003b23ad0 (3)   ***
	...
	[ice.c:janus_ice_send_thread:3519:decrease] 0x612003b23ad0 (2)   ***
	...
	[ice.c:janus_ice_component_destroy:1170:decrease] 0x611002185c90 (1)
	[ice.c:janus_ice_component_destroy:1173:decrease] 0x614002087bd8 (0)
	[dtls.c:janus_dtls_srtp_destroy:800:decrease] 0x613003d143c0 (1)
	[ice.c:janus_ice_component_free:1185:decrease] 0x613003d143c0 (0)
	[ice.c:janus_ice_stream_destroy:1131:decrease] 0x612003b23ad0 (1)   ***
	[ice.c:janus_ice_stream_destroy:1134:decrease] 0x611002185c90 (0)   ***
	...
	[7975052193732121] WebRTC resources freed
	[ice.c:janus_ice_thread:2092:decrease] 0x612003b23ad0 (0)
	[ice.c:janus_ice_free:985:decrease] 0x604001f40228 (0)
	[7975052193732121] WebRTC resources freed
	[7975052193732121] Handle and related resources freed
	[ice.c:janus_ice_free:992:decrease] 0x60b00002bfe8 (14)

The refcount for the `janus_ice_handle` goes to zero after ""Detaching handle from JANUS VideoRoom plugin" (ice.c:932) and probably before the crash at ice.c:962 .  `janus_ice_webrtc_free` is called twice on it, as can be seen by the log message "WebRTC resources freed".   I guess that makes sense because `janus_ice_webrtc_free` is called both from `janus_ice_thread` and from `janus_ice_free`. 

The problem is that that `janus_videoroom_session_free` is called too early.  With the change in 16d1e08, destroying the `ice_handles` hash table decreases the refcount for the `janus_ice_handle`.  

Unfortunately, I don't have time now to figure out a solution.

**Comment by andreasg123** [comment]
There isn't enough information in the log for me to figure out how `handle->iceloop` got corrupted.  I rebased the PR and included pointer addresses in some log messages.  That should provide more information if this crash happens again.

**Comment by cacheworks** [comment]
I just got the following ouput using commit 9690a9f1621a9c5ebb2a901fdd00caaf12e934f1 (I merged in 34d96615ffc7dbafab17c6c544b8aa98c35a108e and #644 )
I was running it with the extra log output on log_level 5, here's the crash and 1000 lines leading up to that.

Looking at this though, it seems to be caused by an incoming request on the admin interface, I poll the server for sessions every 10 seconds but hadn't run into this yet.
https://gist.github.com/cacheworks/718c0b63a4cce83ac8d1f4f65b915c3f

I'm leaving it here just in case I'm missing something and it could be related to this PR, but I'd assume this is a separate issue for the refcount or master branch?

I've merged 58d9393373586d45bfb38c6c1dc1a79febddeb15 now so I'll leave that running overnight, and I'll change the session polling at intervals to manual for now to prevent the other crash from popping up as soon

**Comment by andreasg123** [comment]
@cacheworks, the most recent crash has nothing to do with refcounts or my changes.  `component->selected_pair` is freed in response to a signal from `NiceAgent`.  Ideally, access to `component` would be protected by a mutex.  Assuming that pointer accesses are atomic, there would probably be no problem if the old value of `component->selected_pair` were freed after assigning the new value.

**Comment by andreasg123** [comment]
I rebased this branch so that all the recently merged fixes, including #652, are included.  @cacheworks, please let me know if you get any more crashes.

**Comment by cacheworks** [comment]
Ah cheers! I had manually merged those but this is a bit easier to keep track of.
I'm currently still running the instance I started 22 hours ago so that already looks promising so far.

I'll recompile the rebased PR to deploy tomorrow and will keep you posted on the uptime 👍

**Comment by cacheworks** [comment]
I got the following crash today after deploying 7a3f51799a8dcd01007ecab6434f724855ba1968 earlier;
```heap-use-after-free``` [transports/janus_websockets.c:830](https://github.com/andreasg123/janus-gateway/blob/7a3f51799a8dcd01007ecab6434f724855ba1968/transports/janus_websockets.c#L830) ```janus_websockets_send_message```

Looks like they're not directly related to these changes though as it's pointing to  [janus_websockets.c](https://github.com/andreasg123/janus-gateway/blob/7a3f51799a8dcd01007ecab6434f724855ba1968/transports/janus_websockets.c#L830) 

Here's the gateway log output;
https://gist.github.com/cacheworks/356649d9cb6ef3cd63d4104c108f60c5

**Comment by andreasg123** [comment]
@cacheworks, this is the same [crash](https://github.com/meetecho/janus-gateway/pull/640#issuecomment-251051376) that you previously reported.  I attempted to fix it with #644 by mistakenly believing that the problem was with `client->wsi` because it was allocated inside libwebsockets, even though the line number was slightly off.  I now realize that `client` also gets allocated inside libwebsockets.  There is a race condition between `janus_websockets_send_message` and `janus_websockets_destroy_client`.

Here are some statements of both function in an order that would cause the problem.  In this case, `janus_websockets_send_message` would get past the test `transport->transport_p == NULL` and would then wait at `client->mutex`.  In the meantime, `janus_websockets_destroy_client` would acquire the mutex, set `ws_client->ts->transport_p` to `NULL`, and destroy `ws_client->ts` (same as `transport` in `janus_websockets_send_message`).  Presumably, libwebsockets would free `client` after the callback indicating that it is destroyed.  Then, `janus_websockets_send_message` would continue with a now freed `client`.

	// janus_websockets_send_message
	if(message == NULL)
		return -1;
	if(transport == NULL || transport->transport_p == NULL || g_atomic_int_get(&transport->destroyed)) {
		g_free(message);
		return -1;
	}
	janus_websockets_client *client = (janus_websockets_client *)transport->transport_p;
	// janus_websockets_send_message wait

	// janus_websockets_destroy_client
	janus_mutex_lock(&ws_client->mutex);
	JANUS_LOG(LOG_INFO, "[%s-%p] Destroying WebSocket client\n", log_prefix, wsi);
	#ifndef HAVE_LIBWEBSOCKETS_NEWAPI
	ws_client->context = NULL;
	#endif
	ws_client->wsi = NULL;
	/* Notify core */
	gateway->transport_gone(&janus_websockets_transport, ws_client->ts);
	ws_client->ts->transport_p = NULL;
	janus_transport_session_destroy(ws_client->ts);
	/* Remove messages queue too, if needed */
	if(ws_client->messages != NULL) {
		char *response = NULL;
		while((response = g_async_queue_try_pop(ws_client->messages)) != NULL) {
			g_free(response);
		}
		g_async_queue_unref(ws_client->messages);
	}
	/* ... and the shared buffers */
	g_free(ws_client->incoming);
	ws_client->incoming = NULL;
	g_free(ws_client->buffer);
	ws_client->buffer = NULL;
	ws_client->buflen = 0;
	ws_client->bufpending = 0;
	ws_client->bufoffset = 0;
	janus_mutex_unlock(&ws_client->mutex);

	// janus_websockets_send_message continue
	janus_mutex_lock(&client->mutex);
	/* Convert to string and enqueue */
	char *payload = json_dumps(message, json_format);
	g_async_queue_push(client->messages, payload);
	#ifdef HAVE_LIBWEBSOCKETS_NEWAPI
	lws_callback_on_writable(client->wsi);
	#else
	libwebsocket_callback_on_writable(client->context, client->wsi);
	#endif
	janus_mutex_unlock(&client->mutex);

I don't have a good idea how to fix this.  I'll give it some thought.  Adding a mutex to `janus_transport_session` would affect all transports.

**Comment by lminiero** [comment]
> there is no more janus_transport_session in master

There has never been :wink: 
It's something we added in the refcount branch, at the moment transport plugins would simply pass an opaque pointer to their internal stuff.

**Comment by andreasg123** [comment]
> > there is no more janus_transport_session in master

> There has never been 😉

Yes, I noticed that when I started working on #655.

@lminiero, do you have any comments on this PR?  It mostly seems to work for @cacheworks now.  [One remaining crash](https://github.com/meetecho/janus-gateway/pull/643#issuecomment-252447843) is related to libwebsockets and the cause of [the other crash](https://github.com/meetecho/janus-gateway/pull/643#issuecomment-251835895) may only be discovered with the addresses that I added to the log messages.

**Comment by lminiero** [comment]
I plan to have a good look at the patch tomorrow, along 655, sorry if this is taking a while! I did notice the several debugging lines printing the %p, so I guess that before merging it makes sense to keep on debugging here anyway.

File: plugins/janus_videoroom.c [hunk 2]
```diff
@@ -944,21 +969,19 @@ static void janus_videoroom_notify_participants(janus_videoroom_publisher *parti
 }
 
 static void janus_videoroom_leave_or_unpublish(janus_videoroom_publisher *participant, gboolean is_leaving) {
+	/* we need to check if the room still exists, may have been destroyed already */
+	if(!participant->room || participant->room->destroyed)
```
lminiero: Shouldn't this destroyed check be atomic? I noticed I still left some non-atomic checks of destroyed myself, which is wrong, as we don't use the destroyed value as a trigger for lazy frees anymore in this branch.

plugins/janus_videoroom.c
```diff
@@ -944,21 +969,19 @@ static void janus_videoroom_notify_participants(janus_videoroom_publisher *parti
...
+		return;
 	json_t *event = json_object();
+	janus_mutex_lock(&participant->room->mutex);
```
lminiero: Maybe move this lock before the call to `json_object()`? Just to be (more) sure no destruction happens right after our destroyed check.

plugins/janus_videoroom.c
```diff
@@ -987,11 +1010,15 @@ void janus_videoroom_destroy_session(janus_plugin_session *handle, int *error) {
 		if(session->participant_type == janus_videoroom_p_type_publisher) {
 			/* Get rid of publisher */
 			janus_videoroom_publisher *p = (janus_videoroom_publisher *)session->participant;
```
lminiero: This should use your new `janus_videoroom_session_get_publisher(..)` method, right?

plugins/janus_videoroom.c
```diff
@@ -987,11 +1010,15 @@ void janus_videoroom_destroy_session(janus_plugin_session *handle, int *error) {
...
+			janus_mutex_lock(&session->mutex);
 			session->participant = NULL;
+			janus_mutex_unlock(&session->mutex);
+			janus_videoroom_leave_or_unpublish(p, TRUE);
 			if(p->room) {
+				/* Don't clear p->room.  Another thread calls janus_videoroom_leave_or_unpublish,
+					 too, and there is no mutex to protect this change. */
 				janus_refcount_decrease(&p->room->ref);
```
lminiero: But if this `janus_refcount_decrease` here frees the room, `p->room` points to garbage: isn't it better to be conservative, rather than risking `janus_videoroom_leave_or_unpublish` to crash?

plugins/janus_videoroom.c
```diff
@@ -1916,6 +1948,7 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
 		packet.timestamp = ntohl(packet.data->timestamp);
 		packet.seq_number = ntohs(packet.data->seq_number);
 		/* Go */
+		janus_mutex_lock(&participant->subscribers_mutex);
```
lminiero: While this check is probably needed, I'm afraid it might affect performance, considering how often it's called...

plugins/janus_videoroom.c
```diff
@@ -2024,16 +2059,21 @@ void janus_videoroom_incoming_data(janus_plugin_session *handle, char *buf, int
...
+	if(participant == NULL || g_atomic_int_get(&participant->destroyed)) {
+		g_clear_pointer(&participant, janus_videoroom_publisher_dereference);
 		return;
+	}
 	/* Get a string out of the data */
 	char *text = g_malloc0(len+1);
 	memcpy(text, buf, len);
 	*(text+len) = '\0';
 	JANUS_LOG(LOG_VERB, "Got a DataChannel message (%zu bytes) to forward: %s\n", strlen(text), text);
+	janus_mutex_lock(&participant->subscribers_mutex);
```
lminiero: See above.

plugins/janus_videoroom.c
```diff
@@ -2046,9 +2086,11 @@ void janus_videoroom_slow_link(janus_plugin_session *handle, int uplink, int vid
 	/* Check if it's an uplink (publisher) or downlink (viewer) issue */
 	if(session->participant_type == janus_videoroom_p_type_publisher) {
 		if(!uplink) {
-			janus_videoroom_publisher *publisher = (janus_videoroom_publisher *)session->participant;
-			if(publisher == NULL || g_atomic_int_get(&publisher->destroyed))
+			janus_videoroom_publisher *publisher = janus_videoroom_session_get_publisher(session);
+			if(publisher == NULL || g_atomic_int_get(&publisher->destroyed)) {
```
lminiero: See above.

plugins/janus_videoroom.c
```diff
@@ -2427,11 +2471,11 @@ static void *janus_videoroom_handler(void *data) {
 				publisher->vrc = NULL;
 				janus_mutex_init(&publisher->rec_mutex);
 				publisher->firefox = FALSE;
-				publisher->bitrate = videoroom->bitrate;
+				publisher->bitrate = publisher->room->bitrate;
```
lminiero: Why changing this? Isn't it more likely that `publisher->room` is broken (user left) than `videoroom` is (room destroyed)?

plugins/janus_videoroom.c
andreasg123: It fixes [this crash](https://github.com/meetecho/janus-gateway/pull/403#issuecomment-248583617) reported by @cacheworks.  [Locks in glibc were improved](https://lwn.net/Articles/534758/) a few years ago so that this may not have too much of a performance impact.

plugins/janus_videoroom.c
andreasg123: Yes, it should be.  About half the tests for `room->destroyed` were not atomic.  I'll change them all.

plugins/janus_videoroom.c
andreasg123: There is no need to increase the refcount for the publisher here as it is gettting cleared.  However, the session mutex locked one line later should protect this assignment, too.

plugins/janus_videoroom.c
andreasg123: I changed this back to this to avoid the crash that I got when I cleared it.  This is the lesser of two evils as long as there is no mutex to manage changes to `janus_videoroom_publisher.room`.

andreasg123:
I had to rush my changes at the end because I have to leave.  Please let me know if I missed something.

plugins/janus_videoroom.c
lminiero: Maybe it's @cacheworks code that does this? I think our demo only does a detach when closing a publisher, unpublish is done via a button on demand by the user and never automatically. Anyway, I agree it's good to fix this on the server side, as that's not a behaviour that should lead to a crash.

plugins/janus_videoroom.c
```diff
@@ -2024,16 +2059,21 @@ void janus_videoroom_incoming_data(janus_plugin_session *handle, char *buf, int
 	janus_videoroom_session *session = (janus_videoroom_session *)handle->plugin_handle;
 	if(!session || g_atomic_int_get(&session->destroyed) || session->participant_type != janus_videoroom_p_type_publisher)
 		return;
-	janus_videoroom_publisher *participant = (janus_videoroom_publisher *)session->participant;
-	if(participant == NULL || g_atomic_int_get(&participant->destroyed))
+	janus_videoroom_publisher *participant = janus_videoroom_session_get_publisher(session);
+	if(participant == NULL || g_atomic_int_get(&participant->destroyed)) {
+		g_clear_pointer(&participant, janus_videoroom_publisher_dereference);
```
lminiero: Yep, but we may be doing an `&NULL`. I assume this is safe then? (probably a dumb question :smile: )

lminiero:
Looks good! Let's see if the debugging `%p` you added allows us to spot what's left so that we can merge :+1:

plugins/janus_videoroom.c
cacheworks: I call UnpublishOwnFeed() and .destroy() on each session onBeforeUnload so perhaps that causes it sometimes. It's not something I was able to consistently replicate yet though so it might occasionally pop up when sending 'detach' or 'unpublish' gets delayed.
this change fixed that indeed though so we don't have to trust the client to not crash the gateway 👍

plugins/janus_videoroom.c
cacheworks: Neither me nor any of the users connecting to the gateway have noticed any delays after this commit was added so this lock seems fine for performance with around 700 handles at a time

lminiero: Although this might be more of an issue when there are many viewers for a specific publisher, we'll make this test ourselves later on.

plugins/janus_videoroom.c
lminiero: Yep! Anyway, just FYI, you don't need to unpublish when you destroy a session or simply detach the publishing handle, the VideoRoom plugin does all that for you: `hangup_media` is automatically called, so all users are notified anyway.

plugins/janus_videoroom.c
andreasg123: `&participant` is a pointer to the local variable that may contain the value `NULL`.  `participant == NULL` tests whether the local variable contains the value `NULL`.

andreasg123:
The `%p` that I added are always useful in conjunction with `refcount_debug`.  They don't do any harm without it other than adding a few extra characters to some log lines.

My suggestion is to either just leave them in or to add `if` statements that check for `refcount_debug` when selecting a variant for a call to `JANUS_LOG`.  Lorenzo, what do you think?

The situation where the extra log would have been useful only happened once.  In all likelihood, it wasn't caused by the changes in this PR but either by 16d1e08 (already merged) or something that has always been there.  I don't think that this should hold up merging unless there are other remaining issues.

lminiero:
Fair point, we do have the `REFCOUNT_DEBUG` enabled anyway in the refcount branch for the same reason (verbose debugging) so it definitely makes sense having your additional info there. Merging then!

PR #644, opened 2016-10-03
Handle LWS_CALLBACK_WSI_DESTROY

andreasg123:
The [crash](https://github.com/meetecho/janus-gateway/pull/640#issuecomment-251051376) reported by @cacheworks could be due to libwebsockets not calling back with `LWS_CALLBACK_CLOSED`, maybe because the socket didn't get fully established.  Additionally checking for `LWS_CALLBACK_WSI_DESTROY` should make this safer.  The `destroyed` flag is used to do the cleanup only once.

@cacheworks, please provide more the websockets log the next time one of the call stacks includes "libwebsockets".

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -739,12 +739,19 @@ void janus_websockets_destroy(void) {
 	JANUS_LOG(LOG_INFO, "%s destroyed!\n", JANUS_WEBSOCKETS_NAME);
 }
 
-static void janus_websockets_destroy_client(janus_websockets_client *ws_client) {
+static void janus_websockets_destroy_client(
+	janus_websockets_client *ws_client,
+#ifdef HAVE_LIBWEBSOCKETS_NEWAPI
+	struct lws *wsi,
+#else
+	struct libwebsocket *wsi,
+#endif
+	const char *log_prefix) { 
 	if(!ws_client || !g_atomic_int_compare_and_exchange(&ws_client->destroyed, 0, 1))
 		return;
 	/* Cleanup */
 	janus_mutex_lock(&ws_client->mutex);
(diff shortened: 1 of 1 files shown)


cacheworks:
Thanks for looking into this crash as well. I'll run this with more websockets debugging :+1:

lminiero:
This looks like a reasonable and clean way to handle the issue effectively, thanks! Not sure if we need to wait for more tests from @cacheworks before merging this, as the issue apparently only happened once and so may be quite hard to replicate. Anyway, it may be good to still have it running for a while to make sure it doesn't cause any regression (which I don't think should happen).

As soon as you guys feel it's safe to merge, I can do that right away :+1:

cacheworks:
I'm indeed not sure how to replicate this, I only saw it happen twice in that one commit, which also had a couple of other use-after-free issues in those 10 hour, so I started testing #643 instead and haven't seen it since.

It also never popped up in the 67 hours using 8289183ebe35fedeeadfc70432cd5c92ff087b94 so it'd be hard to tell with certainty that it's solved

I'll merge and use this internally to make sure it doesn't break anything though

lminiero:
Thanks! No need to waste hours and hours on this, if it seems to be working fine for what looks like a reasonable while, just let me know and I'll merge.

Thinking about this, shouldn't this be a fix for the master as well, rather than the refcount branch alone? I can see this happening in the existing branches too, if the cause is related to WebSockets and not the reference counter mechanism. I can take care of that after a merge.

andreasg123:
When creating #647, I noticed that this should be applied to the admin web socket, too.

There is a lot of code duplication between `janus_websockets_callback` and `janus_websockets_admin_callback` that could be reduced by creating a common function that takes `log_prefix` as a parameter.

lminiero:
I noticed a couple of nits (indentation and a duplicate message) in both PRs, but I can fix them myself after the merge. Thanks!

PR #645, opened 2016-10-04
Add support of MQTT transport

manifest:
Message Queue Telemetry Transport (MQTT) is a popular and lightweight publish and subscribe messaging transport. It's also an open standard, and there is already exist a lot of open source brokers/clients. It would be nice to have build-in support for this protocol within Janus Gateway.

If it's ok, I could make another commit with Docker file describing a container with all needed development environment and installed dependencies including MQTT broker within it. I also would glad to fix any code-style and other issues. Just let me know about it

### DIFF
--- a/conf/janus.transport.mqtt.cfg.sample
+++ b/conf/janus.transport.mqtt.cfg.sample
@@ -1,0 +1,24 @@
+; Configuration of the MQTT additional transport for the Janus API.
+[general]
+enable = no												; Whether the support must be enabled
+json = indented										; Whether the JSON messages should be indented (default),
+																	; plain (no indentation) or compact (no indentation and no spaces)
+
+url = tcp://localhost:1883				; The connection URL of the MQTT broker
+;client_id = guest								; Client identifier
+;username = guest									; Username to use to authenticate, if needed
+;password = guest									; Password to use to authenticate, if needed
+;keep_alive_interval = 20					; Keep connection for N seconds
+;cleansession = 0									; Clean session flag
+;disconnect_timeout = 100					; Seconds to wait before destroying client
+subscribe_topic = to-janus				; Topic for incoming messages
+;subscribe_qos = 1								; QoS for incoming messages
+publish_topic = from-janus				; Topic for outgoing messages
+;publish_qos = 1									; QoS for outgoing messages
+
+[admin]
+;admin_enable = no								; Whether the support must be enabled
+subscribe_topic = to-janus-admin	; Topic for incoming admin messages
+;subscribe_qos = 1								; QoS for incoming admin messages
+publish_topic = from-janus-admin	; Topic for outgoing admin messages
+;publish_qos = 1									; QoS for outgoing admin messages
--- a/transports/janus_mqtt.c
+++ b/transports/janus_mqtt.c
@@ -1,0 +1,637 @@
+/*! \file   janus_mqtt.c
+ * \author Andrei Nesterov <ae.nesterov@gmail.com>
+ * \copyright GNU General Public License v3
+ * \brief  Janus MQTT transport plugin
+ * \details  This is an implementation of a MQTT transport for the Janus API,
+ * using the Eclipse Paho C Client library (https://eclipse.org/paho/clients/c).
+ * This means that this module adds support for MQTT based messaging as
+ * an alternative "transport" for API requests, responses and notifications.
+ * This is only useful when you're handling the communication with
+ * clients your own way. Right now, you can only configure
+ * the address of the MQTT broker to use, and the queues to
+ * make use of to receive (to-janus) and send (from-janus) messages
+ * from/to an external application. As with WebSockets, considering that
+ * requests wouldn't include a path to address some mandatory information,
+ * these requests addressed to Janus should include as part of their payload,
+ * when needed, additional pieces of information like \c session_id and
+ * \c handle_id. That is, where you'd send a Janus request related to a
+ * specific session to the \c /janus/<session> path, with RabbitMQ
+ * you'd have to send the same request with an additional \c session_id
+ * field in the JSON payload.
+ * \note When you create a session using RabbitMQ, a subscription to the
+ * events related to it is done automatically through the outgoing queue,
+ * so no need for an explicit request as the GET in the plain HTTP API.
+ *
+ * \ingroup transports
+ * \ref transports
+ */
+
+#include "transport.h"
+
+#include <MQTTAsync.h>
+
+#include "../debug.h"
+#include "../config.h"
+#include "../utils.h"
+
+/* Transport plugin information */
+#define JANUS_MQTT_VERSION        1
+#define JANUS_MQTT_VERSION_STRING "0.0.1"
+#define JANUS_MQTT_DESCRIPTION    "This transport plugin adds MQTT support to the Janus API via Paho client library."
+#define JANUS_MQTT_NAME           "JANUS MQTT transport plugin"
+#define JANUS_MQTT_AUTHOR         "Andrei Nesterov <ae.nesterov@gmail.com>"
+#define JANUS_MQTT_PACKAGE        "janus.transport.mqtt"
+
+/* Transport methods */
+janus_transport *create(void);
+int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path);
+void janus_mqtt_destroy(void);
+int janus_mqtt_get_api_compatibility(void);
+int janus_mqtt_get_version(void);
+const char *janus_mqtt_get_version_string(void);
+const char *janus_mqtt_get_description(void);
+const char *janus_mqtt_get_name(void);
+const char *janus_mqtt_get_author(void);
+const char *janus_mqtt_get_package(void);
+gboolean janus_mqtt_is_janus_api_enabled(void);
+gboolean janus_mqtt_is_admin_api_enabled(void);
+int janus_mqtt_send_message(void *context, void *request_id, gboolean admin, json_t *message);
+void janus_mqtt_session_created(void *context, guint64 session_id);
+void janus_mqtt_session_over(void *context, guint64 session_id, gboolean timeout);
+
+/* Transport setup */
+static janus_transport janus_mqtt_transport_ =
+	JANUS_TRANSPORT_INIT (
+		.init = janus_mqtt_init,
+		.destroy = janus_mqtt_destroy,
+
+		.get_api_compatibility = janus_mqtt_get_api_compatibility,
+		.get_version = janus_mqtt_get_version,
+		.get_version_string = janus_mqtt_get_version_string,
+		.get_description = janus_mqtt_get_description,
+		.get_name = janus_mqtt_get_name,
+		.get_author = janus_mqtt_get_author,
+		.get_package = janus_mqtt_get_package,
+
+		.is_janus_api_enabled = janus_mqtt_is_janus_api_enabled,
+		.is_admin_api_enabled = janus_mqtt_is_admin_api_enabled,
+
+		.send_message = janus_mqtt_send_message,
+		.session_created = janus_mqtt_session_created,
+		.session_over = janus_mqtt_session_over,
+	);
+
+/* Transport creator */
+janus_transport *create(void) {
+	JANUS_LOG(LOG_VERB, "%s created!\n", JANUS_MQTT_NAME);
+	return &janus_mqtt_transport_;
+}
+
+/* API flags */
+static gboolean janus_mqtt_api_enabled_ = FALSE;
+static gboolean janus_mqtt_admin_api_enabled_ = FALSE;
+
+/* JSON serialization options */
+static size_t json_format_ = JSON_INDENT(3) | JSON_PRESERVE_ORDER;
+
+/* MQTT client context */
+typedef struct janus_mqtt_context {
+	janus_transport_callbacks *gateway;
+	MQTTAsync client;
+	struct {
+		int keep_alive_interval;
+		int cleansession;
+	} connect;
+	struct {
+		int timeout;
+	} disconnect;
+	struct {
+		char *topic;
+		int qos;
+	} subscribe;
+	struct {
+		char *topic;
+		int qos;
+	} publish;
+	struct {
+		struct {
+			char *topic;
+			int qos;
+		} subscribe;
+		struct {
+			char *topic;
+			int qos;
+		} publish;
+	} admin;
+} janus_mqtt_context;
+
+/* Transport client methods */
+void janus_mqtt_client_connection_lost(void *context, char *cause);
+int janus_mqtt_client_message_arrived(void *context, char *topicName, int topicLen, MQTTAsync_message *message);
+void janus_mqtt_client_delivery_complete(void *context, MQTTAsync_token token);
+int janus_mqtt_client_connect(janus_mqtt_context *ctx);
+void janus_mqtt_client_connect_success(void *context, MQTTAsync_successData *response);
+void janus_mqtt_client_connect_failure(void *context, MQTTAsync_failureData *response);
+int janus_mqtt_client_reconnect(janus_mqtt_context *ctx);
+void janus_mqtt_client_reconnect_success(void *context, MQTTAsync_successData *response);
+void janus_mqtt_client_reconnect_failure(void *context, MQTTAsync_failureData *response);
+int janus_mqtt_client_disconnect(janus_mqtt_context *ctx);
+void janus_mqtt_client_disconnect_success(void *context, MQTTAsync_successData *response);
+void janus_mqtt_client_disconnect_failure(void *context, MQTTAsync_failureData *response);
+int janus_mqtt_client_subscribe(janus_mqtt_context *ctx, gboolean admin);
+void janus_mqtt_client_subscribe_success(void *context, MQTTAsync_successData *response);
+void janus_mqtt_client_subscribe_failure(void *context, MQTTAsync_failureData *response);
+void janus_mqtt_client_admin_subscribe_success(void *context, MQTTAsync_successData *response);
+void janus_mqtt_client_admin_subscribe_failure(void *context, MQTTAsync_failureData *response);
+int janus_mqtt_client_publish_message(janus_mqtt_context *ctx, char *payload, gboolean admin);
+void janus_mqtt_client_publish_janus_success(void *context, MQTTAsync_successData *response);
+void janus_mqtt_client_publish_janus_failure(void *context, MQTTAsync_failureData *response);
+void janus_mqtt_client_publish_admin_success(void *context, MQTTAsync_successData *response);
+void janus_mqtt_client_publish_admin_failure(void *context, MQTTAsync_failureData *response);
+void janus_mqtt_client_destroy_context(janus_mqtt_context **ctx);
+
+/* We only handle a single client */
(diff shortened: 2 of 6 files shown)


lminiero:
@manifest amazing effort, thanks for sharing! New transport plugins are always welcome, especially if they're widely used. I'm personally not familiar with MQTT, but if it's a popular way of exchanging messages then it's definitely nice to have. Is the way it works now similar to the RabbitMQ implementation we have? That is, a single client allowed writing/reading to/from a server's queue?

I'll review the code itself later to provide specific feedback. What I'd add is some info to the README and mainpage.dox, as we have for the other transports, but those can wait. No Docker file please, though: that's orthogonal to Janus itself, and if I were you I'd just publish it as an external repo we can then point to in the [resources](https://janus.conf.meetecho.com/docs/resources) documentation page.

manifest:
@lminiero yes, it works similar to RabbitMQ implementation (a single client allowed writing/reading to/from a server's queue). I'll take a look on README and mainpage.dox and than update this PR. Looking forward to code review feedback.

conf/janus.transport.mqtt.cfg.sample
```diff
@@ -0,0 +1,36 @@
...
+
+[subscribe]
+topic = to-janus              ; Topic for incoming messages
+;qos = 1                      ; QoS for incoming messages
+
+[publish]
+topic = from-janus            ; Topic for outgoing messages
+;qos = 1                      ; QoS for outgoing messages
+
+[admin.subscribe]
```
lminiero: Why so many groups? Isn't it simpler to just use a couple, e.g., one for Janus API and one for Admin API, and put the related stuff in there, as the other transport plugins do?

transports/janus_mqtt.c
```diff
@@ -0,0 +1,648 @@
...
+ *
+ * \ingroup transports
+ * \ref transports
+ */
+
+#include "transport.h"
+#include "../debug.h"
+#include "../config.h"
+#include "../utils.h"
+#include <MQTTAsync.h>
```
lminiero: Can you add an empty line here? We try to separate local includes from includes related to dependencies. Probably silly, but just coding style :smile:

transports/janus_mqtt.c
```diff
@@ -0,0 +1,648 @@
...
+
+	/* Read configuration */
+	char filename[255];
+	g_snprintf(filename, 255, "%s/%s.cfg", config_path, JANUS_MQTT_PACKAGE);
+	JANUS_LOG(LOG_VERB, "Configuration file: %s\n", filename);
+	janus_config *config = janus_config_parse(filename);
+	if(config != NULL)
+		janus_config_print(config);
+
+	/* Handle configuration */
```
lminiero: All this should happen in the `if(config != NULL)` check. It's not that all unlikely to load a plugin without a configuration file available (e.g., a user forgetting to do a `make configs`).

lminiero: As a side note, I guess this wouldn't cause issues as the default values would be used as a fallback in the following code... still, a missing configuration should still result in a warning on the console, as that may not be what the user wanted.

transports/janus_mqtt.c
```diff
@@ -0,0 +1,648 @@
...
+const char *janus_mqtt_get_author(void);
+const char *janus_mqtt_get_package(void);
+gboolean janus_mqtt_is_janus_api_enabled(void);
+gboolean janus_mqtt_is_admin_api_enabled(void);
+int janus_mqtt_send_message(void *context, void *request_id, gboolean admin, json_t *message);
+void janus_mqtt_session_created(void *context, guint64 session_id);
+void janus_mqtt_session_over(void *context, guint64 session_id, gboolean timeout);
+
+/* Transport setup */
+static janus_transport janus_mqtt_transport_ =
```
lminiero: I can't say I'm really a fan of variable names that end with an underscore, but that may be just me... :wink:

transports/janus_mqtt.c
```diff
@@ -0,0 +1,648 @@
...
+		} else if(!strcasecmp(json_item->value, "compact")) {
+			/* Compact, so no spaces between separators */
+			json_format_ = JSON_COMPACT | JSON_PRESERVE_ORDER;
+		} else {
+			JANUS_LOG(LOG_WARN, "Unsupported JSON format option '%s', using default (indented)\n", json_item->value);
+			json_format_ = JSON_INDENT(3) | JSON_PRESERVE_ORDER;
+		}
+	}
+
+	{ // connect configuration
```
lminiero: Use `/* ... */` instead of `// ...` for comments please (same for other occurrences).

lminiero: Besides (again, just coding style, you can blame me for being so nitpicking! :smile: ), why putting all those in brackets? The way we typically do that in our code is to just prepend the comment line.

transports/janus_mqtt.c
```diff
@@ -0,0 +1,648 @@
...
+	if(context == NULL) {
+		json_decref(message);
+		return -1;
+	}
+
+	janus_mqtt_context *ctx = (janus_mqtt_context *)context;
+	char *payload = json_dumps(message, json_format_);
+	json_decref(message);
+
+	JANUS_LOG(LOG_HUGE, "Sending %s API message via MQTT: %s\n", admin ? "admin" : "Janus", payload);
```
lminiero: I see you're sending messages directly from within the `janus_mqtt_send_message` method. Can `janus_mqtt_client_publish_admin_message` and `janus_mqtt_client_publish_janus_message` block while sending those messages, or do they queue messages internally in the MQTT library? The reason why I ask is that we definitely don't want to keep the core busy, and so all the existing transport libraries only queue the message, while a dedicated thread works on the queue to actually dispatch them.

transports/janus_mqtt.c
```diff
@@ -0,0 +1,648 @@
...
+	/* We don't care */
+}
+
+void janus_mqtt_session_over(void *context, guint64 session_id, gboolean timeout) {
+	/* We don't care, not even if it's a timeout (should we?), our client is always up */
+}
+
+void janus_mqtt_client_connection_lost(void *context, char *cause) {
+	JANUS_LOG(LOG_INFO, "MQTT connection lost cause of %s. Reconnecting...\n", cause);
+	// automatic reconnect
```
lminiero: Shouldn't this notify the core the transport is gone, and then notify it again when it comes back up? I guess the idea is not to loose existing sessions because of an occasional disconnection, but a missing connection would result in notifications/events sent by the core being just dropped I believe.

transports/janus_mqtt.c
```diff
@@ -0,0 +1,648 @@
...
+	JANUS_LOG(LOG_HUGE, "MQTT client has been successfully published to MQTT topic: %s\n", ctx->publish.topic);
+}
+
+void janus_mqtt_client_publish_janus_failure(void *context, MQTTAsync_failureData *response) {
+	janus_mqtt_context *ctx = (janus_mqtt_context *)context;
+	int rc = response ? response->code : 0;
+	JANUS_LOG(LOG_ERR, "MQTT client has been failed publishing to MQTT topic: %s, return code: %d\n", ctx->publish.topic, rc);
+}
+
+int janus_mqtt_client_publish_admin_message(janus_mqtt_context *ctx, char *payload) {
```
lminiero: Can't some of this methods, which are basically the same code, just envisage something like an additional `gboolean admin` flag, and then use the Janus or Admin stuff according to its value? I guess it makes sense for some callbacks to remain separated, especially if you don't have anything they tell you to allow you to retrieve the original context, for active methods should allow for that.

transports/janus_mqtt.c
```diff
@@ -0,0 +1,648 @@
...
+
+	janus_mqtt_client_destroy_context(context);
+}
+
+int janus_mqtt_client_subscribe(janus_mqtt_context *ctx) {
+	MQTTAsync_responseOptions options = MQTTAsync_responseOptions_initializer;
+	options.onSuccess = janus_mqtt_client_subscribe_success;
+	options.onFailure = janus_mqtt_client_subscribe_failure;
+	options.context = ctx;
+	return MQTTAsync_subscribe(ctx->client, ctx->subscribe.topic, ctx->subscribe.qos, &options);
```
lminiero: Doesn't this mean that you can't just do the Admin API, if you don't enable the Janus API support as well?

conf/janus.transport.mqtt.cfg.sample
manifest: There is similar properties per connection type (publish/subscribe) in MQTT. For instance, QoS. So I separated them by type for the purpose - do not use prefixes.

```ini
[subscribe]
topic = to-janus
qos = 1

[publish]
topic = from-janus
qos = 1
```

Do you think prefixes is a better solution? Should I replace the config above to something like this?

```ini
[general]
subscribe_topic = to-janus
subscribe_qos = 1
publish_topic = from-janus
publish_qos = 1
```

transports/janus_mqtt.c
manifest: Done

transports/janus_mqtt.c
manifest: We already have an error message:
```
[ERR] [config.c:janus_config_parse:91]   -- Error reading configuration file 'janus.transport.mqtt.cfg'... error 2 (No such file or directory)
```

transports/janus_mqtt.c
manifest: I use the async API of client library. The messages are internally being queued by the library, and then being dispatched by dedicated thread.

transports/janus_mqtt.c
manifest: With QoS equal to 1 and 2, MQTT protocol guarantees message delivery. Although the client is configured to use in-memory persistence, I think it would be ok in our case. To lose messages we should have entire system or janus itself failed.

```c
 * ::MQTTCLIENT_PERSISTENCE_NONE: Use in-memory persistence. If the device or 
 * system on which the client is running fails or is switched off, the current
 * state of any in-flight messages is lost and some messages may not be 
 * delivered even at QoS1 and QoS2.
```
https://github.com/eclipse/paho.mqtt.c/blob/master/src/MQTTAsync.h

transports/janus_mqtt.c
manifest: It doesn't. For now, you can enable just Janus API or just Admin API or both.

manifest:
I've added some info to `README` and `mainpage.dox`.

conf/janus.transport.mqtt.cfg.sample
lminiero: I'd prefer the prefixed approach there, yep, mostly because it makes it easier for who configures the plugin to know what belongs where (and because it's consistent with the way you configure other transports).

transports/janus_mqtt.c
lminiero: Nah, I'm not that picky!

transports/janus_mqtt.c
lminiero: Got it.

transports/janus_mqtt.c
lminiero: No need, I was mostly curious on why you had gone for that.

transports/janus_mqtt.c
lminiero: I was mostly thinking of more extreme cases, e.g., the server goes away for some reason, and you're there trying to reconnect forever to something that doesn't exist anymore. Anyway, that's probably an edge case that's not really important.

transports/janus_mqtt.c
lminiero: But the way it works now, aren't you trying to first subscribe to the Janus API, and only in the success callback, subscribe to the Admin API after that? If the Janus API is disabled, wouldn't the first subscription fail? Not really familiar with the library, so please forgive me if the question is silly :smile:

transports/janus_mqtt.c
lminiero: :+1:

lminiero:
Thanks for the quick addition, the changes there look good to me!

conf/janus.transport.mqtt.cfg.sample
manifest: Will do.

transports/janus_mqtt.c
lminiero: So the question WAS silly... thanks for the clarification!

conf/janus.transport.mqtt.cfg.sample
manifest: Done.

manifest:
Looks like I fixed all the issues. Please, take a look.

lminiero:
Thanks, I'll review tomorrow (day almost over here), and I'll also try to install the library and play with it myself.

As a side note, before we can merge, have you signed our [CLA](https://cla.conf.meetecho.com/janus/) at the end of this page? We do need this before we can accept any external contribution.

Thanks again for this very nice effort!

manifest:
I've described how you can set up development environment, build and start playing with the MQTT transport very quickly using Docker, in that [Gist](https://gist.github.com/manifest/aa2cfb19b596669b710bbc0f8885d64f). As for CLA, I will sign it tomorrow. Just need to ask my employer the right contact information you requested for the document.

oscarvady:
Wow, good job!

I'd like to ask you which are the scenarios when MQTT can be useful. I
thought that MQTT is an IoT protocol used on Wireless Sensor Devices, so
that, maybe MQTT protocol is not efficient with WebRTC media streams
(video/audio). What do you think about that?

Best regards



2016-10-04 16:39 GMT+02:00 Andrei Nesterov <notifications@github.com>:

manifest:
@oscarvady I think, you've misunderstood the differences between signaling and media streaming protocols within WebRTC framework. The transport here implements signaling. Actually, signaling could be implemented in any bidirectional protocol, but some of them is better fits than other. MQTT is very lightweight and efficient, that make it ideal for IoT as you already said. Many brokers support WebSockets over MQTT. So that, by implementing WebRTC signaling based on it we could use same siglaning across variety of clients including web, work stations, mobile and even embedded devices.

oscarvady:
I understand, thank you very much Andrei! ;)

2016-10-06 10:01 GMT+02:00 Andrei Nesterov <notifications@github.com>:

manifest:
@lminiero I've got this error trying to sign your company's CLA.

screen shot 2016-10-07 at 10 04 29 am

lminiero:
Can you try again? It's probably the backend that had a temporary hiccup.

manifest:
I've tried it for few times yesterday and done it just now one more time. The result is the same: "Error inserting database data". Maybe there is another way to accomplish that?

lminiero:
I'll notify the systems guy and keep you posted, sorry about this issue.

manifest:
@lminiero Yes, I've filled up all the fields. I haven't used any weird characters. I've tried different formats for my phone +71234567890, 71234567890 and without the county code 1234567890. And I'm still getting an error on response.

I've tried in actual versions of FF and Chrome on macOS.

lminiero:
That's weird, can you make a screenshot of what you've filled before hitting the button to do the post, so that we can try to replicate this ourselves?

lminiero:
(or if it's easier, just sending us the info, e.g., via mail, lminiero[at]gmail.com )

manifest:
I sent the info and screenshot by email :-)

lminiero:
Giving this a try right now. I didn't have any MQTT server installed, and tried running Janus anyway. Besides, I didn't modify the configuration file, which lead to:

    [WARN] MQTT support disabled for both Janus and Admin API, giving up
    JANUS MQTT transport plugin destroyed!

which is expected. The usual stuff worked (regular EchoTest demo via HTTP), but when I tried shutting Janus down via CTRL+C I got:

    Disconnecting MQTT client...
    Bus error (core dumped)

Maybe caused by the fact that `janus_mqtt_client_destroy_context` was called but `context_` was not reset to NULL after the failed setup?

lminiero:
For some reason, the second time I started Janus I had to do a `LD_LIBRARY_PATH=/usr/local/lib ./janus` or it wouldn't find the library. Worked fine the first time, so not sure what caused it (maybe installing mosquitto, not sure though). It may be worth to add to the README that either the path to install paho.mqtt.c must be set to where libraries usually go (e.g., /usr/lib64) or that the preload must be set accordingly when starting Janus, as it defaults the installation to /usr/local/lib which, at least in my Fedora setup, isn't searched for libraries.

Apart from that, just tried with a simple node-js application and mosquitto, and it seems to be working nicely! All requests handled, and both responses and events arrive as they should. I haven't done extensive testing or anything like that, but that will probably come as soon as people start using it.

As a side note, can you use tabs instead of spaces for the comments in the configuration file? I just noticed this when enabling support for the module myself.

All that said, as soon as you fix the crash I mentioned above and the nits cited here, this is good to merge for me!

manifest:
Couldn't repeat the crash by myself. But I think you're right in your assumption. I've updated the function.

manifest:
I've added a note about install paths similar to that is for RabbitMQ and also replaced spaces by tabs in the config file.

manifest:
I've added a note about install paths and also replaced spaces by tabs in the config file.

lminiero:
Thanks! I still don't see the fix to the broken `context_`, though, unless I'm looking at the wrong commit. `context_` is set to the value of `ctx` here:

https://github.com/meetecho/janus-gateway/pull/645/commits/7f613b920266b4caecc3fd49c16c33bea8fc57e0#diff-1738368af93e523883890a4d3c5c62e3R165

If something goes wrong, `ctx` is destroyed here via `janus_mqtt_client_destroy_context`:

https://github.com/meetecho/janus-gateway/pull/645/commits/7f613b920266b4caecc3fd49c16c33bea8fc57e0#diff-1738368af93e523883890a4d3c5c62e3R327

The broken pointer is still in `context_` though, as it's not set to NULL. This means that when Janus is shutting down and the module with it, `janus_mqtt_client_disconnect` is called on `context_`, which is pointing to garbage:

https://github.com/meetecho/janus-gateway/pull/645/commits/7f613b920266b4caecc3fd49c16c33bea8fc57e0#diff-1738368af93e523883890a4d3c5c62e3R340

and so the code in `janus_mqtt_client_disconnect` segfaults.

manifest:
I've changed the type of argument for [janus_mqtt_client_destroy_context](https://github.com/meetecho/janus-gateway/pull/645/commits/899f4ef11054328255350096836018a7f7865a02#diff-1738368af93e523883890a4d3c5c62e3R633) function to solve this.

lminiero:
That doesn't solve the issue. It's quite easy to replicate. Set `enabled=no` in `[general]` (the default after a fresh install anyway). Start Janus, see the `[WARN] MQTT support disabled for both Janus and Admin API, giving up` warning, and then shutdown, and it will crash:

    Disconnecting MQTT client...
    Bus error (core dumped)

The only way to fix this, is setting `context_ = NULL` in the `error:` label after you destroy the original context.

I actually noticed that there's a fix to make in the core as well. In fact, we're ignoring the result of the init, which leads to the destroy for your plugin (and others that fail to load, I guess) to be called even if it failed to load. I guess the other plugins don't complain as we clean the stuff related to their setup when it fails the first time, and so the destroy is harmless.

I'll fix this in the core right away, but if you could add that `context_ = NULL` at the end of the `error:`label before returning -1 it would help!

lminiero:
Nevermind that, I just fixed the behaviour in the code so your issue should be fixed as a consequence (destroy won't be invoked). Merging, thanks!

PR #652, opened 2016-10-06
Assign new value before freeing old value to avoid state with freed value

andreasg123:
This should work around the [crash](https://github.com/meetecho/janus-gateway/pull/643#issuecomment-252077398) reported by @cacheworks, assuming that pointer access is atomic.

There are other cases where a pointer is set to `NULL` after being freed.  Using `g_clear_pointer` in those cases should be safer because the pointer would be set to `NULL` **before** being freed.

	g_free(a->b);
	a->b = NULL;

	g_clear_pointer(&a->b, g_free);

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1516,8 +1516,6 @@ void janus_ice_cb_new_selected_pair (NiceAgent *agent, guint stream_id, guint co
 		JANUS_LOG(LOG_ERR, "[%"SCNu64"]     No component %d in stream %d??\n", handle->handle_id, component_id, stream_id);
 		return;
 	}
-	if(component->selected_pair)
-		g_free(component->selected_pair);
 	char sp[200];
 #ifndef HAVE_LIBNICE_TCP
 	g_snprintf(sp, 200, "%s <-> %s", local, remote);
@@ -1565,7 +1563,9 @@ void janus_ice_cb_new_selected_pair (NiceAgent *agent, guint stream_id, guint co
 		laddress, lport, ltype, local->transport == NICE_CANDIDATE_TRANSPORT_UDP ? "udp" : "tcp",
 		raddress, rport, rtype, remote->transport == NICE_CANDIDATE_TRANSPORT_UDP ? "udp" : "tcp");
 #endif
+	gchar *prev_selected_pair = component->selected_pair;
 	component->selected_pair = g_strdup(sp);
+	g_clear_pointer(&prev_selected_pair, g_free);
 	/* Now we can start the DTLS handshake (FIXME This was on the 'connected' state notification, before) */
 	JANUS_LOG(LOG_VERB, "[%"SCNu64"]   Component is ready enough, starting DTLS handshake...\n", handle->handle_id);
 	/* Have we been here before? (might happen, when trickling) */
@@ -3363,9 +3363,10 @@ void *janus_ice_send_thread(void *data) {
 					/* Append REMB */
 					memcpy(rtcpbuf+rrlen, pkt->data, pkt->length);
 					/* Free old packet and update */
-					g_free(pkt->data);
+					char *prev_data = pkt->data;
 					pkt->data = rtcpbuf;
 					pkt->length = rrlen+pkt->length;
+					g_clear_pointer(&prev_data, g_free);
 				}
 				/* FIXME Copy in a buffer and fix SSRC */
 				char sbuf[JANUS_BUFSIZE];


cacheworks:
Wow, I was just filling out a new issue and I saw your PR pop up before I hit send.

Once again thank you for taking the time to debug and explain issues for so many of the logs I come back with, I appreciate to be able to debug and patch things so quickly with the help of you guys despite lacking in knowledge here

andreasg123:
I didn't think this all the way through.  My change ensures that the pointer stored in `component->selected_pair` isn't freed, assuming atomic pointer access (all eight bytes are read or written at once).  Before, the freed pointer stayed there for 48 lines in `janus_ice_cb_new_selected_pair`, including calls to `nice_address_to_string`, `nice_address_get_port`, `g_snprintf`, and `g_strdup`.  However, the new version doesn't prevent one thread from reading the pointer and another thread from replacing and freeing it immediately.

In this particular situation, it should be fine.  `component->selected_pair` is only passed to `json_string` and not used after that.  Having a different value for the `if` statement and the call isn't a problem, either.

	if(component->selected_pair) {
		json_object_set_new(c, "selected-pair", json_string(component->selected_pair));
	}

The [Jansson library](http://www.digip.org/jansson/) isn't annotated by AddressSanitizer so that it wouldn't notice access to a freed pointer.  Accessing a freed pointer isn't inherently fatal.  It's only bad if new data is written to that memory location while somebody is still using that pointer.

lminiero:
I thank you again as well for all the love you're giving to the project! You've been much more productive on Janus than me lately :smile: 
The change looks fine to me and the suggestion reasonable: I understand the point you make about messing with the pointer from different threads, but as you noticed the only place where we set the value for that variable is that callback, the others only read it.

lminiero:
I also updated the refcount branch to incorporate the fix as well :+1:

andreasg123:
Here is a long discussion on [atomic pointer access in C](http://stackoverflow.com/questions/879077/is-changing-a-pointer-considered-an-atomic-action-in-c).  The conclusion is that it probably works but is not guaranteed by the language.  In this case, there is also the issue that there may be pointers to an old value that gets freed in another thread.

While my change _probably_ works, the only **safe** approach is to protect setting the pointer, freeing the value, and reading the value of the pointer in a mutex.

	janus_mutex_lock(&component->mutex);
	if(component->selected_pair) {
	    json_object_set_new(c, "selected-pair", json_string(component->selected_pair));
	}
	janus_mutex_unlock(&component->mutex);

	janus_mutex_lock(&component->mutex);
	g_free(component->selected_pair);
	component->selected_pair = g_strdup(sp);
	janus_mutex_unlock(&component->mutex);

lminiero:
IIRC, the fact that on 64 bit machines pointers are 8 bytes, means you can never give for granted that setting them is atomic. Not sure if I'm remembering this wrong... anyway, most of our sets/resets is already protected by mutexes. I don't know why that's not the case for selected-pair, but it may have been intentional to avoid deadlocks, since sometimes IIRC libnice callbacks are invoked as a direct consequence of actions made by the user (e.g., setting new remote candidates), which means that if something was locked there, locking again may deadlock the code.

andreasg123:
Just with respect to 64-bit pointers, here are the C statements and the corresponding assembly code (note [AT&T syntax](http://stackoverflow.com/questions/2397528/mov-src-dest-or-mov-dest-src)):

	g_snprintf(sp, 200, "%s <-> %s", local, remote);
	gchar *prev_selected_pair = component->selected_pair;
	component->selected_pair = g_strdup(sp);
	g_clear_pointer(&prev_selected_pair, g_free);

	.LC10:
		.string	"%s <-> %s"
		.section	.rodata.str1.8
		.align 8

		.loc 1 1418 0
		movq	(%rsp), %rcx	# %sfp,
		leaq	288(%rsp), %rdi	#, tmp329
		movq	%r13, %r8	# remote,
		movl	$.LC10, %edx	#,
		movl	$200, %esi	#,
		xorl	%eax, %eax	#
	.LVL77:
		call	g_snprintf	#
	.LVL78:
		.loc 1 1463 0
		movq	56(%r14), %rbp	# component_52->selected_pair, prev_selected_pair.205_54
	.LVL79:
		.loc 1 1464 0
		leaq	288(%rsp), %rdi	#, tmp330
		call	g_strdup	#
	.LVL80:
	.LBB311:
		.loc 1 1465 0
		testq	%rbp, %rbp	# prev_selected_pair.205_54
	.LBE311:
		.loc 1 1464 0
		movq	%rax, 56(%r14)	# tmp236, component_52->selected_pair
	.LVL81:
	.LBB312:
		.loc 1 1465 0
		je	.L71	#,
	.LVL82:
		.loc 1 1465 0 is_stmt 0 discriminator 1
		movq	%rbp, %rdi	# prev_selected_pair.205_54,
		call	g_free	#
	.LVL83:
	.L71:

I would assume that `movq %rax, 56(%r14)` is atomic (moving the result of `g_strdup` as a quad-word from register `%rax` to an offset of 56 bytes past the pointer stored in register `%r14` that holds `component`).

As an aside, `g_clear_pointer` is expanded inline so that there is no disadvantage in using it with `g_free`.  `prev_selected_pair` is checked for `NULL` with `testq` and `je` jumps to `.L71` if it is `NULL`.

PR #653, opened 2016-10-07
Remove code duplication between regular and admin web sockets

andreasg123:
As I mentioned a couple of days ago, there is quite a bit of code duplication between regular and admin web sockets.

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -988,7 +988,7 @@ static int janus_websockets_callback_https(
 }
 
 /* This callback handles Janus API requests */
-static int janus_websockets_callback(
+static int janus_websockets_common_callback(
 #ifdef HAVE_LIBWEBSOCKETS_NEWAPI
 		struct lws *wsi,
 		enum lws_callback_reasons reason,
@@ -997,8 +997,9 @@ static int janus_websockets_callback(
 		struct libwebsocket *wsi,
 		enum libwebsocket_callback_reasons reason,
 #endif
-		void *user, void *in, size_t len)
+		void *user, void *in, size_t len, gboolean admin)
 {
+	const char *log_prefix = admin ? "AdminWSS" : "WSS";
 	janus_websockets_client *ws_client = (janus_websockets_client *)user;
 	switch(reason) {
 		case LWS_CALLBACK_ESTABLISHED: {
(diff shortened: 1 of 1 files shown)


lminiero:
I don't think this will work. The signature of `janus_websockets_common_callback` is dictated by the libwebsockets library, not us, as it's a callback they invoke automatically after some events. Besides, the removal of `libwebsocket_callback_on_writable` with a manual invocation of the callback is probably a bad idea too, as `libwebsocket_callback_on_writable` tells the library that it's ok/required to write now, and so it should let us know as soon as the socket becomes writeable and notify us with an event so that we can take care of that.

andreasg123:
Git didn't do a good job in showing the changes that I made.  Here is a diff of the two callback functions after I replaced "WSS" and "AdminWSS", respectively, with "%s".

	--- /tmp/regular.c      2016-10-07 08:26:39.265543838 -0700
	+++ /tmp/admin.c        2016-10-07 08:27:14.575326623 -0700
	@@ -1,5 +1,5 @@
	-/* This callback handles Janus API requests */
	-static int janus_websockets_callback(
	+/* This callback handles Admin API requests */
	+static int janus_websockets_admin_callback(
	 #ifdef HAVE_LIBWEBSOCKETS_NEWAPI
	                struct lws *wsi,
	                enum lws_callback_reasons reason,
	@@ -21,8 +21,8 @@
	                        libwebsockets_get_peer_addresses(this, wsi, libwebsocket_get_socket_fd(wsi), name, 256, ip, 256);
	 #endif
	                        JANUS_LOG(LOG_VERB, "[%s-%p] WebSocket connection opened from %s by %s\n", wsi, ip, name);
	-                       if(!janus_websockets_is_allowed(ip, FALSE)) {
	-                               JANUS_LOG(LOG_ERR, "[%s-%p] IP %s is unauthorized to connect to the WebSockets Janus API interface\n", wsi, ip);
	+                       if(!janus_websockets_is_allowed(ip, TRUE)) {
	+                               JANUS_LOG(LOG_ERR, "[%s-%p] IP %s is unauthorized to connect to the WebSockets Admin API interface\n", wsi, ip);
	                                /* Close the connection */
	 #ifdef HAVE_LIBWEBSOCKETS_NEWAPI
	                                lws_callback_on_writable(wsi);
	@@ -109,7 +109,7 @@
	                        g_free(ws_client->incoming);
	                        ws_client->incoming = NULL;
	                        /* Notify the core, passing both the object and, since it may be needed, the error */
	-                       gateway->incoming_request(&janus_websockets_transport, ws_client, NULL, FALSE, root, &error);
	+                       gateway->incoming_request(&janus_websockets_transport, ws_client, NULL, TRUE, root, &error);
	                        return 0;
	                }
	                case LWS_CALLBACK_SERVER_WRITEABLE: {

The main difference is passing a different Boolean value to `janus_websockets_is_allowed` and `gateway->incoming_request`.  There is also a difference in including "Janus" or "Admin" in one log message.  The rest is covered by the log prefix that I changed to "%s" so that it is replaced by the value of `log_prefix` as the first parameter for the `JANUS_LOG` calls.

I didn't change the signature of the callbacks for the libwebsockets library but have them call the common function.

	/* This callback handles Janus API requests */
	static int janus_websockets_callback(
	#ifdef HAVE_LIBWEBSOCKETS_NEWAPI
			struct lws *wsi,
			enum lws_callback_reasons reason,
	#else
			struct libwebsocket_context *this,
			struct libwebsocket *wsi,
			enum libwebsocket_callback_reasons reason,
	#endif
			void *user, void *in, size_t len)
	{
	#ifdef HAVE_LIBWEBSOCKETS_NEWAPI
		return janus_websockets_common_callback(wsi, reason, user, in, len, FALSE);
	#else
		return janus_websockets_common_callback(this, wsi, reason, user, in, len, FALSE);
	#endif
	}
	
	/* This callback handles Admin API requests */
	static int janus_websockets_admin_callback(
	#ifdef HAVE_LIBWEBSOCKETS_NEWAPI
			struct lws *wsi,
			enum lws_callback_reasons reason,
	#else
			struct libwebsocket_context *this,
			struct libwebsocket *wsi,
			enum libwebsocket_callback_reasons reason,
	#endif
			void *user, void *in, size_t len)
	{
	#ifdef HAVE_LIBWEBSOCKETS_NEWAPI
		return janus_websockets_common_callback(wsi, reason, user, in, len, TRUE);
	#else
		return janus_websockets_common_callback(this, wsi, reason, user, in, len, TRUE);
	#endif
	}

lminiero:
Thanks for the clarification, that makes sense now! Let me give it a quick go and in case it's ok I'll merge then :+1:

lminiero:
I'll work on merging these changes in the refcount branch now.

PR #655, opened 2016-10-09
Put the mutex for websockets into janus_transport_session

andreasg123:
This is an attempt to fix the [race condition](https://github.com/meetecho/janus-gateway/pull/643#issuecomment-252447843) in websockets.  By putting the mutex into `janus_transport_session`, `janus_websockets_client` can be freed by libwebsockets without causing access to freed memory.  As there is no `janus_transport_session` in master, this PR wouldn't apply there.

The change from `g_free(message)` to `json_decref(message)` in `janus_websockets_send_message` should be made in any case.

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -979,8 +979,6 @@ static int janus_websockets_common_callback(
 {
 	const char *log_prefix = admin ? "AdminWSS" : "WSS";
 	janus_websockets_client *ws_client = (janus_websockets_client *)user;
-	if(reason != LWS_CALLBACK_ESTABLISHED && g_atomic_int_get(&ws_client->destroyed))
-		return 0;
 	switch(reason) {
 		case LWS_CALLBACK_ESTABLISHED: {
 			/* Is there any filtering we should apply? */
@@ -1038,6 +1036,8 @@ static int janus_websockets_common_callback(
 				JANUS_LOG(LOG_ERR, "[%s-%p] Invalid WebSocket client instance...\n", log_prefix, wsi);
 				return -1;
 			}
+			if(g_atomic_int_get(&ws_client->destroyed))
+				return 0;
 			/* Is this a new message, or part of a fragmented one? */
 #ifdef HAVE_LIBWEBSOCKETS_NEWAPI
 			const size_t remaining = lws_remaining_packet_payload(wsi);
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -125,7 +125,6 @@ typedef struct janus_websockets_client {
 	int buflen;								/* Length of the buffer (may be resized after re-allocations) */
 	int bufpending;							/* Data an interrupted previous write couldn't send */
 	int bufoffset;							/* Offset from where the interrupted previous write should resume */
-	janus_mutex mutex;						/* Mutex to lock/unlock this session */
 	volatile gint session_timeout;			/* Whether a Janus session timeout occurred in the core */
 	volatile gint destroyed;				/* Whether this libwebsockets client instance has been closed */
 	janus_transport_session *ts;			/* Janus core-transport session */
@@ -750,7 +749,7 @@ static void janus_websockets_destroy_client(
 	if(!ws_client || !g_atomic_int_compare_and_exchange(&ws_client->destroyed, 0, 1))
 		return;
 	/* Cleanup */
-	janus_mutex_lock(&ws_client->mutex);
+	janus_mutex_lock(&ws_client->ts->mutex);
 	JANUS_LOG(LOG_INFO, "[%s-%p] Destroying WebSocket client\n", log_prefix, wsi);
 #ifndef HAVE_LIBWEBSOCKETS_NEWAPI
 	ws_client->context = NULL;
@@ -759,7 +758,6 @@ static void janus_websockets_destroy_client(
 	/* Notify core */
 	gateway->transport_gone(&janus_websockets_transport, ws_client->ts);
 	ws_client->ts->transport_p = NULL;
-	janus_transport_session_destroy(ws_client->ts);
 	/* Remove messages queue too, if needed */
 	if(ws_client->messages != NULL) {
 		char *response = NULL;
@@ -776,7 +774,8 @@ static void janus_websockets_destroy_client(
 	ws_client->buflen = 0;
 	ws_client->bufpending = 0;
 	ws_client->bufoffset = 0;
-	janus_mutex_unlock(&ws_client->mutex);
+	janus_mutex_unlock(&ws_client->ts->mutex);
+	janus_transport_session_destroy(ws_client->ts);
 }
 
 int janus_websockets_get_api_compatibility(void) {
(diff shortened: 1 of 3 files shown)


transports/janus_websockets.c
```diff
@@ -776,7 +774,8 @@ static void janus_websockets_destroy_client(
 	ws_client->buflen = 0;
 	ws_client->bufpending = 0;
 	ws_client->bufoffset = 0;
-	janus_mutex_unlock(&ws_client->mutex);
+	janus_mutex_unlock(&ws_client->ts->mutex);
+	janus_transport_session_destroy(ws_client->ts);
```
andreasg123: This has a tricky dependency.  Libwebsockets cannot free `ws_client` until `janus_websockets_destroy_client` returns.  By calling `janus_transport_session_destroy` last, we make sure that `ws_client->ts` doesn't get freed earlier by `janus_refcount_decrease`.  Anybody else acquiring `ws_client->ts->mutex` will find that `ws_client->ts->transport_p` is now `NULL`.  However, if they didn't increase the refcount, they may access `ws_client->ts->transport_p` via a freed pointer.

One could instead use the following sequence (but this still doesn't protect somebody waiting on the mutex from having it freed under them).  This twists my brain.

	janus_refcount_increase(&ws_client->ts-ref);
	janus_transport_session_destroy(ws_client->ts);
	janus_mutex_unlock(&ws_client->ts->mutex);
	janus_refcount_decrease(&ws_client->ts-ref);

lminiero:
Thanks for working on this! Although to be honest I don't like the idea of a mutex there much... I'll have to check the crash @cacheworks experienced to have a better idea of the issue at hand, as I couldn't investigate this myself yet, and then I'll review your code too.

cacheworks:
Thanks for your efforts, I'd be lost debugging all of this without you guys 👍 
I gave this commit a try and get the following output while the gateway is starting;

https://gist.github.com/cacheworks/4e909d66601c0892e4c4c076a0b6c834

andreasg123:
@cacheworks, it was a bad assumption on my part that the callback with `LWS_CALLBACK_ESTABLISHED` would be first.  There is at least a callback with `LWS_CALLBACK_FILTER_PROTOCOL_CONNECTION` before it.  Thus, there is no easy test before the `switch` statement to see whether `ws_client` is allocated and `ws_client->destroyed` is initialized.

@lminiero, I still believe that the mutex needs to be outside the memory allocated as user space by libwebsockets to avoid using that mutex in `janus_websockets_send_message` after libwebsockets freed it.  I guess one could create a wrapper around `janus_websockets_client` that is allocated and freed in `janus_websockets.c`.

transports/janus_websockets.c
andreasg123: `janus_request_new` increases the refcount for `instance` (`janus_transport_session`) and `janus_websockets_send_message` checks whether `transport_p` is `NULL` inside the mutex, so that my original solution should be fine.

lminiero:
I agree that we may need a wrapper for `janus_websockets_client`, since as you noticed the struct itself is automatically allocated (and freed) by libwebsockets itself, making the management of reference counting and when to get rid of the stuff troublesome. What I'm not sure I like is adding the mutex to the transport stuff, which means in cascade force it on all transports. Regular plugins don't have anything like that, they handle mutexes internally if needed. Unfortunately I still couldn't look into all this due to tons of stuff to do lately, I promise I'll check it tomorrow!

andreasg123:
I understand your reluctance to force a mutex in `janus_transport_session` on all transports.  However, the other transports could just ignore the mutex.  It only adds a little extra memory.  Just a thought...

lminiero:
Just looked to the PR and the issue it addresses, and I think that, once for WebSockets we implement the wrapper struct we discussed, the right way to address this is not a new mutex in the transport, but using the reference counter mechanism better. In fact, the issue here is mostly caused by the core uses a transport that doesn't exist anymore. I don't think we're increasing the reference to the transport when we use it in the core, which we should. We could then exploit the `transport_gone` callback to let the core know when it's time to let go. Coupled with the aforementioned wrapper for the libwebsockets struct (which would give us control over frees), this should prevent the core from contacting a transport that doesn't exist anymore, and at the same time free only when it's time to do that.

Do you think this makes sense? If so, I can work on a PR to do that.

andreasg123:
Sure, that would work.  However, I think that it would be easier to use `janus_transport_session` as this wrapper because it's already there so that little code has to be changed.  If you think that that puts too much burden on the other transports (40 bytes in each session for the mutex), your suggestion is the way to go.

andreasg123:
Also, waiting for the `transport_gone` callback may be too late to address the race condition of 
[janus_websockets_destroy](https://github.com/meetecho/janus-gateway/blob/79f3d87dea4a989cf24f1bd0171c9a9b2773f308/transports/janus_websockets.c#L753) and [janus_websockets_send_message](https://github.com/meetecho/janus-gateway/blob/79f3d87dea4a989cf24f1bd0171c9a9b2773f308/transports/janus_websockets.c#L827) both waiting for `ws_client->mutex`.  That happens when the websocket is closing while `janus_process_success` sends "ack" (from "trickle" in `janus_process_incoming_request`) or `janus_session_notify_event` sends "detached" (from `janus_ice_handle_destroy`).

After acquiring the mutex, `janus_websockets_send_message` should be able to tell that the websocket is closed without accessing freed memory.

andreasg123:
> I don't think we're increasing the reference to the transport when we use it in the core, which we should.

It gets increased in `janus_request_new` [here](https://github.com/meetecho/janus-gateway/blob/79f3d87dea4a989cf24f1bd0171c9a9b2773f308/janus.c#L572).  I didn't check if there are other places where it needs to be increased.  In both crashed reported by @cacheworks, `janus_websockets_send_message` was called indirectly by `janus_process_incoming_request` so that the refcount would have been increased, as far as I can tell.

lminiero:
The problem is not the 40 bytes, obviously, but the concept I'd like to avoid... adding a mutex just to make a single transport work feels like a hack, so that's why I'm thinking of how to better use the reference counter mechanism to avoid the issue you guys found out. I'll give a better look at how we're handling transports from a core level in the next few days to see if I can come up with something.

andreasg123:
I don't think that having a mutex in `janus_transport_session` is a hack.  As I wrote in my documentation, it protects changes to `transport_p`, including setting it to `NULL`.  The latter indicates that the transport session is closed.  In that sense, `janus_transport_session.destroyed` may not be needed anymore.

lminiero:
Sorry for the lack of feedback, but as I wrote in a few other parts I've been away all last week. I'll catch up between tomorrow and the next few days, in particular on the websockets client wrapper stuff (I might prepare a PR for that).

lminiero:
I'm definitely terrible at keeping track of things! Sorry about that... I'll definitely look into this today or tomorrow at the latest. I'm less hesitant now at adding the mutex at the container level as you suggested, so I'll make a couple of tests, and also try to think of something about the tricky thing you mentioned in the inline comment.

transports/janus_websockets.c
```diff
@@ -819,12 +818,17 @@ gboolean janus_websockets_is_admin_api_enabled(void) {
 int janus_websockets_send_message(janus_transport_session *transport, void *request_id, gboolean admin, json_t *message) {
 	if(message == NULL)
 		return -1;
-	if(transport == NULL || transport->transport_p == NULL || g_atomic_int_get(&transport->destroyed)) {
-		g_free(message);
+	if(transport == NULL || g_atomic_int_get(&transport->destroyed)) {
+		json_decref(message);
```
lminiero: Good catch here, very likely a leftover... it's in master as well, I'll fix that.

transports/janus_websockets.c
```diff
@@ -843,18 +847,22 @@ void janus_websockets_session_created(janus_transport_session *transport, guint6
 }
 
 void janus_websockets_session_over(janus_transport_session *transport, guint64 session_id, gboolean timeout) {
-	if(!timeout || transport == NULL || transport->transport_p == NULL || g_atomic_int_get(&transport->destroyed))
+	if(!timeout || transport == NULL || g_atomic_int_get(&transport->destroyed))
```
lminiero: Are we always sure that `transport_p` can be always trusted to be valid? I'd keep the check just for safety...

lminiero: Ok, I see you're actually just doing the check later, in order to wrap it in the lock/unlock, so please disregard the comment above!

lminiero:
Merging, thanks again!

PR #658, opened 2016-10-12
Set JANUS_THREAD_DEBUG to log the number of threads

andreasg123:
This counts the number of threads and logs them to find the cause for the [crash](https://github.com/meetecho/janus-gateway/pull/403#issuecomment-253330636) reported by @cacheworks.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -32,6 +32,10 @@
 #include "rtcp.h"
 #include "apierror.h"
 
+#ifdef JANUS_THREAD_DEBUG
+static volatile gint send_thread_count = 0;
+#endif
+
 /* STUN server/port, if any */
 static char *janus_stun_server = NULL;
 static uint16_t janus_stun_port = 0;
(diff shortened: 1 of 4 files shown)


cacheworks:
Thanks @andreasg123! I'll get this rolled out later and keep you posted

andreasg123:
I amended the commit because I set the wrong log level for the messages in VideoRoom.

lminiero:
See https://github.com/meetecho/janus-gateway/pull/403#issuecomment-253441377, I'm not sure we need this in the code, it can be monitored from outside.

andreasg123:
As @lminiero wrote, most of the same information can be accessed from the outside:

	ps -To pcpu,tid,comm -C janus | sort -r -k1

The problem seems to be due to AddressSanitizer that keeps track of every thread ever created.  Using more thread pools would address that issue.

PR #661, opened 2016-10-13
Use nodebug variants of session_get_publisher and publisher_dereference in high-traffic functions

andreasg123:
`janus_videoroom_session_get_publisher_nodebug` and `janus_videoroom_publisher_dereference_nodebug` call `janus_refcount_increase_nodebug` and `janus_refcount_decrease_nodebug`, respectively.  They are used in high-traffic functions such as `janus_videoroom_incoming_rtp` to keep the size of the log manageable.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -462,6 +462,10 @@ static void janus_videoroom_publisher_dereference(janus_videoroom_publisher *p) 
 	janus_refcount_decrease(&p->ref);
 }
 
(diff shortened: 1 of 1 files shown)


lminiero:
In that case we can assume it's no problem debugging references, right? The ref is increased and decreased in the same method.

andreasg123:
It shouldn't be a problem for debugging as long as we make sure that calls to the new functions are properly balanced.  We lose some information but hopefully not anything critical.

PR #664, opened 2016-10-14
Combine handing of boolean settings

andreasg123:
This combines handling of requests that set a Boolean setting into a single function.  Each setting has to have a simple setter function and optionally a getter function.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -108,17 +108,8 @@ static struct janus_json_parameter admin_parameters[] = {
 	{"transaction", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"janus", JSON_STRING, JANUS_JSON_PARAM_REQUIRED}
 };
-static struct janus_json_parameter debug_parameters[] = {
-	{"debug", JANUS_JSON_BOOL, JANUS_JSON_PARAM_REQUIRED}
-};
 static struct janus_json_parameter level_parameters[] = {
 	{"level", JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE}
-};
-static struct janus_json_parameter timestamps_parameters[] = {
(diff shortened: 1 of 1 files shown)


janus.c
```diff
@@ -108,18 +108,9 @@ static struct janus_json_parameter admin_parameters[] = {
 	{"transaction", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"janus", JSON_STRING, JANUS_JSON_PARAM_REQUIRED}
 };
-static struct janus_json_parameter debug_parameters[] = {
```
lminiero: I understand a helper to set a boolean, but why remove the templates? They're helpful for validation, and we can't always assume that parsing and setting will always be combined. I'd keep the two logically separated.

lminiero:
Did a quick gance and added a comment: I'll only be able to make a full review after I come back, though, sorry!

janus.c
andreasg123: I removed the template because validation happens inside `janus_request_set_boolean_setting` without using `JANUS_VALIDATE_JSON_OBJECT`.  Because I combined different parameters ("debug", "timestamps", "colors") in the same function, one can't use a single template.  My goal was to move duplicate code and the validation for the parameters is the same or very similar.

andreasg123: Setting a Boolean implies that the parameter has to be there and be of type Boolean so that validation in the helper function is appropriate.

andreasg123:
@lminiero, now that you're back, do you have more comments on this PR?

lminiero:
I'm not really sure about this PR, to be honest: it doesn't seem to reduce that many lines, while it removes flexibility, especially in case I want to modify (see, expand) any of those requests later on, so I don't see any real benefit. I'd rather focus on #526 instead now that you've rebased it to refcount (and which has been around for quite a while, my fault!)

andreasg123:
I created #672 to combine some of the code that was targeted by this PR.

PR #666, opened 2016-10-18
Fix compile failed on Mac

Akagi201:
I already tested it compiling successfully on mac and linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,14 +239,15 @@ for more info.
 ### Install dependencies
 
 ```
-brew install jansson libnice openssl libusrsctp libmicrohttpd libwebsockets cmake rabbitmq-c sofia-sip opus libogg glib pkg-config gengetopt
+brew tap homebrew/boneyard
+brew install jansson libnice openssl libusrsctp libmicrohttpd libwebsockets cmake rabbitmq-c sofia-sip opus libogg glib pkg-config gengetopt autoconf automake libtool
 ```
 
 ```
 wget https://github.com/cisco/libsrtp/archive/v1.5.4.tar.gz
 tar xvf v1.5.4.tar.gz
 cd libsrtp-1.5.4
-./configure --prefix=/usr/local/libsrtp
+./configure --prefix=/usr/local
 make
 sudo make install
 ```
--- a/README.md
+++ b/README.md
@@ -253,9 +253,11 @@ sudo make install
 
 ### Build janus gateway
 
+Note: Unix Sockets is not supported on macOS, so you must disable it.
+
 ```
 cd /path/to/janus-gateway
-./configure --disable-unix-sockets --disable-rabbitmq --disable-mqtt --prefix=/usr/local/janus PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig
+./configure --disable-unix-sockets --prefix=/usr/local/janus PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig
 make
 sudo make install
 ```
--- a/README.md
+++ b/README.md
@@ -251,6 +251,8 @@ make
 sudo make install
 ```
 
+### Build janus gateway
+
 ```
 cd /path/to/janus-gateway
 ./configure --disable-unix-sockets --disable-rabbitmq --disable-mqtt --prefix=/usr/local/janus PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig
@@ -267,7 +269,7 @@ things you can configure, either in a configuration file:
 or on the command line:
 
 	<installdir>/bin/janus --help
-	
+
 	janus 0.2.1
 
 	Usage: janus [OPTIONS]...
--- a/README.md
+++ b/README.md
@@ -234,6 +234,30 @@ plugins you don't care about). Use the --help option when configuring
 for more info.
 
 
+## Compile on macOS
+
+### Install dependencies
+
+```
+brew install jansson libnice openssl libusrsctp libmicrohttpd libwebsockets cmake rabbitmq-c sofia-sip opus libogg glib pkg-config gengetopt
+```
+
+```
+wget https://github.com/cisco/libsrtp/archive/v1.5.4.tar.gz
+tar xvf v1.5.4.tar.gz
+cd libsrtp-1.5.4
+./configure --prefix=/usr/local/libsrtp
+make
+sudo make install
+```
+
+```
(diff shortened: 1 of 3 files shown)


lminiero:
Not sure we should just remove that `clockid_t`. If it was there, somebody added it for a reason. Maybe it should only be defined if it's not already? Anybody with a mac can confirm this doesn't break anything?

As a side note, please sign our [CLA](https://cla.conf.meetecho.com/janus/) at the end of the page before we can merge. Thanks!

Akagi201:
@lminiero As I already said that, I have tested it my macOS and Linux. `clockid_t` must not be redefined, and it is for linux too.

README.md
```diff
@@ -234,6 +234,32 @@ plugins you don't care about). Use the --help option when configuring
...
+./configure --prefix=/usr/local/libsrtp
+make
+sudo make install
+```
+
+### Build janus gateway
+
+```
+cd /path/to/janus-gateway
+./configure --disable-unix-sockets --disable-rabbitmq --disable-mqtt --prefix=/usr/local/janus PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig
```
lminiero: Why the `--disable-unix-sockets --disable-rabbitmq --disable-mqtt` here? I understand that's how you're compiling it, but other people may be interested in the features.

Akagi201: unix-sockets is not supported on macOS.

OK, I will remove the other 2 disable options

lminiero:
I know it works for you, since you proposed the patch :smile:
I was asking for more Mac users to test and confirm, before merging, as I don't own one and have absolutely no experience on OSX development. As soon as we have some more feedback I'll merge.

Tnx for signing the CLA!

andreasg123:
Clearly, `clockid_t` [isn't defined on some Macs](http://stackoverflow.com/questions/13472761/unknown-type-name-clockid-t).  One comment suggests that `_POSIX_C_SOURCE < 199309L` may be a good test to wrap around the definition.

lminiero:
Thanks for the suggestion @andreasg123! @Akagi201 can you incorporate it in the PR and conditionally restore the definition?

Akagi201:
@lminiero @andreasg123 My macOS version is 10.12.1. And I can build it successfully.

I have no environment to test the old macOSs.

So I am not sure to define some "strange" macros to the source codes.

I think all mac users will tend to use the newest OS version. And we should not the redefine the types.

andreasg123:
I have a Mac and I'm running 10.11.6, disproving that Mac users tend to run the newest OS version.  I'm pretty sure that there are users running 10.10 or even earlier.

I looked for the definition of `clockid_t` and found it in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/time.h`.

It's wrapped in this:

    #if !defined(_DARWIN_FEATURE_CLOCK_GETTIME) || _DARWIN_FEATURE_CLOCK_GETTIME != 0
    #if __DARWIN_C_LEVEL >= 199309L

It shouldn't be too difficult to figure out what to wrap around the definition of `clockid_t` in Janus to negate that.  I would suggest this:

    #if !defined(_DARWIN_C_LEVEL) || _DARWIN_C_LEVEL < 199309L
    typedef ... clockid_t;
    #endif

I also looked for `_DARWIN_FEATURE_CLOCK_GETTIME` in all include files in that directory and subdirectories and didn't find it elsewhere.  I don't know where it would be defined, maybe in the compiler.

Akagi201:
@andreasg123 Thanks for your deep research. Could you give the exact `typedef` code on Old mac machines. Because I don't have the environment to test.

andreasg123:
It looks like I didn't submit my comment that I wrote a couple of days ago.  I don't have an old Mac, either.  However, the following should work as far as I know.

    #if !defined(__DARWIN_C_LEVEL) || __DARWIN_C_LEVEL < 199309L
    typedef int clockid_t;
    #endif

IvRRimum:
I don't think it's possible to build this on OSX. libnice is not supported anymore.

```
configure: error: Package requirements (
                    glib-2.0 >= 2.32
                    nice
                    jansson
                    libssl >= 1.0.1
                    libcrypto
                  ) were not met:

No package 'nice' found

```

```
==> Installing libnice from homebrew/boneyard
==> Downloading https://nice.freedesktop.org/releases/libnice-0.1.7.tar.gz
Already downloaded: /Users/karliscudars/Library/Caches/Homebrew/libnice-0.1.7.tar.gz
==> ./configure --prefix=/usr/local/Cellar/libnice/0.1.7 --disable-silent-rules
Last 15 lines from /Users/karliscudars/Library/Logs/Homebrew/libnice/01.configure:
configure: WARNING: 'missing' script is too old or missing
checking for a thread-safe mkdir -p... ./install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports nested variables... (cached) yes
checking for style of include used by make... GNU
checking for gcc... clang
checking whether the C compiler works... no
configure: error: in `/private/tmp/libnice-20161110-55947-1cur7r4/libnice-0.1.7':
configure: error: C compiler cannot create executables
See `config.log' for more details

Error: libnice was moved to homebrew-boneyard because it has unfixable issues.
Please do not file any issues about this. Sorry!

```

andreasg123:
I don't plan to build Janus for the Mac but the `libnice` issue may be fixable.  It was moved to the Homebrew boneyard for [no particular reason](https://github.com/Homebrew/homebrew-core/pull/6510).  Version 0.1.7 is pretty old and the current version is [0.1.13](https://cgit.freedesktop.org/libnice/libnice/) (still 18 months old).

IvRRimum:
I just ran VM with linux and ran janus that way. It's working

Akagi201:
@andreasg123 @andreasg123 
I add the typedef for old macOSs.

Please review it and merge.

nareix:
Nice PR. I buld it on macOS successfully

lminiero:
Sorry for the late reply, I've been abroad more than two weeks... I'll review and merge when I get back to the office on Monday, thanks!

lminiero:
Merging, thanks! Just as a heads up, I'll edit the README to make it more verbose in some parts and concise in others, and in line with the way the rest of the text is written. I'll also add libcurl to the `brew` command, so let me know if that will break anything.

PR #670, opened 2016-10-25
fix documentation for audiobridge plugin

seb3s:


lminiero:
Thanks for fixing the typo!

PR #672, opened 2016-10-25
Function to create reply with common parameters

andreasg123:
This adds the function `janus_create_reply` for the common pattern of creating a JSON reply object and setting the properties "janus", "session_id", and "transaction".

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -150,15 +150,21 @@ gint janus_is_stopping(void) {
 /* Public instance name */
 static gchar *server_name = NULL;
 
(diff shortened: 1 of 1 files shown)


andreasg123:
I rebased to the current refcount branch.

lminiero:
Isn't `janus_create_message` more suited than `janus_create_reply`? We do send notifications as well from the core that are not a response to anything.

andreasg123:
That makes sense.  Initially, I only used the new function for assignments to `reply` variables but I've found additional uses since.

lminiero:
Looks good, tnx!

PR #673, opened 2016-10-26
Add script to setup janus on ubuntu quickly

jaychsu:
Tested and worked on Ubuntu {14.04,16.04}

### DIFF
--- a/setup-on-ubuntu.sh
+++ b/setup-on-ubuntu.sh
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+# REF: https://groups.google.com/forum/#!msg/meetecho-janus/RYP4FBaeUi0/bIGSPZlpEQAJ
+# Test env: ubuntu {14.04,16.04} LTS
+
+# 0. Get infos and params without empty
+# ======
+# Default setting
+SOURCE_LOCATE=$HOME
+COMPILE_FOLDER=/opt/janus
+ENABLE_POST_PROCESS=false
+ENABLE_DATA_CHANNEL=false
+ENABLE_WEB_SOCKET=false
+
+while getopts "s:c:pdw" OPT; do
+	case $OPT in
+		s)
+			SOURCE_LOCATE=$OPTARG
+			echo "Set 'SOURCE_LOCATE' to '${OPTARG}'."
+			;;
+		c)
+			COMPILE_FOLDER=$OPTARG
+			echo "Set 'COMPILE_FOLDER' to '${OPTARG}'."
+			;;
+		p)
+			ENABLE_POST_PROCESS=true
+			echo "Enable feature: 'post-process'."
+			;;
+		d)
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for the contribution, but I don't think this needs to be merged. This is something that can be easily provided as an external helper tool. Advertizing on the group, or adding to the resources docs, is advised.

njam:
One idea could be to provide the configuration for proper "debianization": the `control` file defines what dependencies need to be installed, the `rules` file defines how to build the binaries, you can even have acceptance tests.
Here's a blueprint to create a janus-gateway .deb for Ubuntu vivid: [vivid/janus/debian/](https://github.com/cargomedia/debian-packages/tree/d94289570d8e7d21e654c9c58cf1a3d38551ac71/packages/vivid/janus/debian).

I've seen repositories having a `debian/` folder in their root. But agree, probably it's better to do it outside. Maybe just move it to a separate repo.

jaychsu:
It's OK.

I just share a worked process to setup `Janus`. Since there are some steps in `README` make me confused.

joshdickson40:
Does the `svn co http://sctp-refimpl.googlecode.com/svn/trunk/KERN/usrsctp usrsctp` step succeed for you? That step fails for me and the site appears to be down.

joshdickson40:
Also Ubuntu 16.04 and 14.04 ship with an outdated version of `libsrtp`. So when its included as `libsrtp-dev` for the dependency installs, you get 1.4.X and we want 1.5.X.

jaychsu:
Since I do installation in `README` previously. So I updated `libsrtp` to 1.5 manually without script.

I updated in following commit.

joshdickson40:
I'm still confused about the `libsrtp-dev` install step (I actually opened a Google Groups thread about this today, coincidentally). So if we install it from the Ubuntu PPA, we get 1.4.X. The readme suggests that we in fact *uninstall* this version (or don't install it in the first place, presumably) and install 1.5.0 via the provided instructions. However, when following the instructions to install 1.5.0 (or 1.5.4), Janus does not see libsrtp. Are you sure that 1.5.0 is actually being used with your script? My tests seem to indicate Janus is actually using 1.4.x despite following the install steps for 1.5.X.

lminiero:
Answered on the group. The target for 1.5.4 is `shared_library`.

joshdickson40:
Still seeing `configure: error: libsrtp not found` from Janus when using the instructions you just added to the Readme on Ubuntu 16.04.

lminiero:
Those are the steps we use for our 16.04 docker images, and they work. You're probably doing something wrong, e.g., forgetting the --libdir or something.

joshdickson40:
Yeah I was in the process of editing my comment as I saw this comment pop up, that was in fact the problem. Thank you, Lorenzo.

jaychsu:
Closed and for information only

PR #675, opened 2016-11-01
Dump old refcounts once per hour

andreasg123:
Here is the PR [that I suggested](https://github.com/meetecho/janus-gateway/pull/403#issuecomment-257616026).  It compiles but I haven't tested it otherwise.  The interval and age of refcounts can be set via `REFCOUNT_DUMP_INTERVAL` and `REFCOUNT_DUMP_AGE` in `refcount.h`.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -65,9 +65,11 @@ static int pipefd[2];
 
 
 #ifdef REFCOUNT_DEBUG
+#include <time.h>
 /* Reference counters debugging */
 GHashTable *counters = NULL;
 janus_mutex counters_mutex;
+time_t last_refcount_dump = 0;
 #endif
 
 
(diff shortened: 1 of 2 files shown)


cacheworks:
Thanks for the quick suggestion Andreas! I'll recompile with this PR and keep you posted

andreasg123:
@cacheworks, you should make sure to save the whole log.  I can write a script that extracts a dump and the refcount changes for those pointers.  That hopefully will reveal parts of the code where increases outweigh decreases.

andreasg123:
This probably won't reveal much more than the Python script that I wrote.  Lorenzo, feel free to pull it if you think that it would be beneficial.

To locate the source of the memory usage, one probably has to use a heap profiler such as [gperftools](https://github.com/gperftools/gperftools).  I don't know if that would work with the GLib memory allocation but the description seems to indicate that only linking is needed to get the malloc replacement.

PR #683, opened 2016-11-16
RTP Extensions support

AjayChoudary:
Tried to support WebRTC RTP Extensions as explained in below links.
https://tools.ietf.org/html/rfc5285
https://tools.ietf.org/html/rfc6464
https://webrtc.org/experiments/rtp-hdrext/abs-send-time/

Currently i hard coded the extension uri's & id's,  as am unable to find the right place to do extmap negotiation.

RTP Header Extension for Client-to-Mixer Audio Level is very useful in improving audio bridge quality :+1:

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -96,6 +96,14 @@ gboolean janus_ice_is_bundle_forced(void) {
 	return janus_force_bundle;
 }
 
+static gboolean janus_extmap;
+void janus_ice_extmap_support(gboolean extmap) {
+	janus_extmap = extmap;
+	JANUS_LOG(LOG_INFO, "EXTMAP %s going to be enabled\n", janus_extmap ? "is" : "is NOT");
+}
+gboolean janus_ice_is_extmap_enabled(void) {
+	return janus_extmap;
+}
 /* Whether rtcp-mux support is mandatory or not (false by default) */
 static gboolean janus_force_rtcpmux;
 static gint janus_force_rtcpmux_blackhole_port = 1234;
@@ -2712,6 +2720,8 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 		audio_stream->audio_rtcp_ctx->tb = 48000;	/* May change later */
 		audio_stream->video_rtcp_ctx = g_malloc0(sizeof(rtcp_context));
 		audio_stream->video_rtcp_ctx->tb = 90000;
+		memset(audio_stream->aextmaps, 0x00, sizeof(audio_stream->aextmaps));
+		memset(audio_stream->vextmaps, 0x00, sizeof(audio_stream->vextmaps));
 		janus_mutex_init(&audio_stream->mutex);
 		audio_stream->components = g_hash_table_new(NULL, NULL);
 		g_hash_table_insert(handle->streams, GUINT_TO_POINTER(handle->audio_id), audio_stream);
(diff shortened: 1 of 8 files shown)


lminiero:
Thanks for this. I'm currently abroad and will only be able to look into this next week, though. From a quick glance, I'd only ask you to fix a few indentantations in the code, as I see some spaces in place of tabs.

AjayChoudary:
To test RTP extensions support, we have to enable "extmap-support" in [media] context in janus.cfg.

@lminiero to make this work, i made changes in both SDP & RTP layers, 
tried my best to fit it in janus code style, but used some ugly hacks to communicate from plugin to SDP layer. (In those places i mentioned "FIXME").

Tested audio_bridge and video_room plugins both were working fine.

Please review and let me know if any further required from my side to make smooth merge to master.

lminiero:
Sorry for the time this review took, but I've been abroad about two weeks and I've only now gotten back to work...

While I appreciate the contribution and seems in general quite helpful, this seems to be doing more than it should, especially considering the core doesn't really care about what you negotiate, plugins do. Besides, we already have some of the things this patch adds, e.g., SDP direction indications. I don't believe there's a need for a global extmap enable/disable either, plugins should decide what they need or not.

The hacks are mostly in plugins calling `janus_ice_` stuff which should never happen: since the extmap stuff is related to SDP negotiation, that should probably go in the sdp-utils or RTP code instead, as that's what plugins refer to when there's stuff to do with the SDP and/or RTP/RTCP.

I'd actually rather do it differently, though, in a way where we can even avoid the extmap array entirely. If we don't strip those attributes in the core and then have plugins care about that, reusing what's available already in the SDP tools, then there's no need for all that. E.g., audioroom gets an offer, it can crawl the attributes using the SDP utils and see if it finds the extension it needs; if it does, it can add the same thing to the answer it generates so that it's negotiates, and if needed set a boolean value to true (if it needs it for any reason, for instance knowing whether to look for extensions later on). Once you do that, there really is no need for any change (apart possibly from helpers to search for attributes that avoid the crawling part, but that's orthogonal), as the core will automatically integrate the extmap the plugin sent in the answer back to the WebRTC user. 

For this reason, I'd only keep in the code you added to properly handle/skip extensions in plugins (even though the values don't seem to be actually used in your patch, only parsed?), and change the way you check for the presence of those extensions using the already available SDP utils instead. The RTP extension parsing could be done in rtp.h instead (which might require a new rtp.c for the actual implementation) so that all plugins have a simpler way to read a value from there (pretty much as we do already for RTCP, e.g., `janus_rtcp_get_remb`), and for this we can probably re-use the code you already added to the patch.

I do realize this seems quite a heavy refactoring from what you've done, so I can also prepare a new PR myself that establishes the groundwork and that you can extend with the additional features you care about. Does this make sense?

Thanks!

AjayChoudary:
@lminiero thanks for your valuable time and am very glad to your response.

Initially i tried preparing sdp (with extmap's) in plugin only, but those new lines are not carried in final sdp sent to endpoint. Then i added them directly in core.

Regarding extmaps array, RTP extensions id's are dynamic (unlike RTCP headers).
Currently all browsers are using common id for extensions, but RFC is not specified any fixed id for extensions. We should always respect the dynamic id (specified in offer) to parse the Extensions.
So extension_id + extension_uri combination should be available in both SDP & RTP levels.

In my patch i didn't used the parsed extensions, we have to implement an [algorithm](https://tools.ietf.org/html/rfc6464#section-5) to take a decision and need to implement [Mixer-Client Audio Level extension](https://tools.ietf.org/html/rfc6465).
In Video-Room all end points will the power of P2P communication (as we are relaying the video-orientation/delay/abs-time extension headers in RTP).  

Please let me know once you create new PR.

lminiero:
Yeah I know they're dynamic, but I wouldn't worry about that now. Anyway, I've started making some changes in a dedicated PR, I'll publish it in the next few hours or upcoming days.

lminiero:
@AjayChoudary just created, let me know if this addresses your original requirements.

lminiero:
Closing this as we merged #697 eventually, thanks for the help on this!

PR #694, opened 2016-11-26
Fix mutex unlocking

seb3s:
Fix mutex unlocking that is causing segmentation fault on my server.
The mutex is already unlocked a bit earlier in the function.

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -2058,9 +2058,6 @@ void janus_audiobridge_hangup_media(janus_plugin_session *handle) {
 	}
 	participant->last_drop = 0;
 	janus_mutex_unlock(&participant->qmutex);
-	if(audiobridge != NULL) {
-		janus_mutex_unlock(&audiobridge->mutex);
-	}
 	janus_mutex_unlock(&rooms_mutex);
 }
 


lminiero:
Good catch about the extra unlock but:

1. it should never result in a crash, and
2. the one to remove should actually be this other one here https://github.com/seb3s/janus-gateway/blob/83364e217290b6599de80cc371f9266c60e3b992/plugins/janus_audiobridge.c#L2016 as you want to keep the room locked while doing the rest of the stuff on the recording

Can you let me know if restoring the unlock you removed and removing the one I mentioned fixes it for you? In case I'll integrate the fix myself.

seb3s:
Hi Lorenzo,
Yep it is ok, it also fixes the crash if I remove the first unlock instead of the last one.
I had no trouble before until I installed yesterday my app on a new server. It then crashes miserably every time I quit a room !
Debian Jessie on Intel(R) Xeon(R) CPU E3-1245 v5 @ 3.50GHz.
Cheers,
Sébastien.

lminiero:
I just made a commit that removes that first unlock instead of the second one, so we can close this PR. Thanks for the heads up and for spotting this!

seb3s:
Great thanks for the commit

PR #695, opened 2016-11-28
Support for setting an iceTransportPolicy

agclark81:
Introduces support for setting an iceTransportPolicy parameter on gatewayCallbacks to force a specific policy, such as "relay" which would require the use of a TURN relay. In some restrictive environments, STUN might return ICE candidates that can't successfully receive audio/video. If no parameter is specified, there is no change to the existing behavior.


lminiero:
Is this supported on all browsers? Just wondering if this can break things when you specify the setting and the browser doesn't do that (although I guess that's up probably to the application to check).

Anyway, looks like a reasonable addition to me, and something I wasn't aware of, thanks! I see you already signed the CLA so we can merge right away :+1:

PR #697, opened 2016-11-30
Added support for (some) RTP extensions

lminiero:
This PR revisits the effort made by @AjayChoudary in #683. Specifically, it introduces support for RTP header extensions in Janus, trying to make it a bit more integrated with what was already available.

Unlike the original patch, this PR does not keep a map/array of extension IDs, though. Just as the core doesn't mess with payload types, it doesn't interfere with extension IDs either, and leaves negotiation and management of those to plugins. An example of how this can be done is available in both the AudioBridge and VideoRoom plugins, which have been extended to negotiate a few extensions, if provided by the offerer. The AudioBridge negotiates the audio level extension to check if a frame is silence and to keep track of the current audio level. The VideoRoom plugin negotiates audio level and video orientation to just forward those to the viewer (no manipulation done on the server side).

The PR introduces a few new methods in rtp.h that plugins can use:

1. `janus_rtp_payload` allows you to quickly access the RTP payload: it automatically skips extensions and other stuff beyond the common RTP header, and so might be useful to avoid issues like assuming the data is simply 12 bytes away (which is what we did in a couple of plugins).
2. `janus_rtp_header_extension_get_id` is a quick'n'dirty method to find the extension ID associated with a namespace (e.g., `"urn:ietf:params:rtp-hdrext:ssrc-audio-level"`) from an SDP string; it might be useful as a quick alternative to parsing the SDP and traversing it using the SDP utils (which is what the VideoRoom plugin does instead) to get the ID, which is needed when you have to negotiate the feature and make sure you use the same ID in an answer.
3. `janus_rtp_header_extension_get_from_id` does the opposite: you pass the numeric ID and you get back the extension namespace (assuming the extension is defined in rtp.h, otherwise we fail).

There also are a couple of methods to actually parse an RTP extension:

1. `janus_rtp_header_extension_parse_audio_level` parses an audio level RTP extension, returning the value (in dBov) reported in there.
2. `janus_rtp_header_extension_parse_video_orientation` parses a video orientation RTP extension, returning the values (camera, flip, rotation) reported in there (couldn't test this one as even negotiating it I never got any of those, so feedback welcome on the accuracy).

I plan to merge this quite soon, so please let me know if there's anything that needs to be fixed or improved. I'm not aware of any issue with the existing demos and none should appear, but you never know if this can end up breaking something else (e.g., external applications making assumptions on the RTP they get from Janus plugins).

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -481,8 +481,8 @@ record_file =	/path/to/recording.wav (where to save the recording)
 
 
 /* Plugin information */
-#define JANUS_AUDIOBRIDGE_VERSION			9
-#define JANUS_AUDIOBRIDGE_VERSION_STRING	"0.0.9"
+#define JANUS_AUDIOBRIDGE_VERSION			10
+#define JANUS_AUDIOBRIDGE_VERSION_STRING	"0.0.10"
 #define JANUS_AUDIOBRIDGE_DESCRIPTION		"This is a plugin implementing an audio conference bridge for Janus, mixing Opus streams."
 #define JANUS_AUDIOBRIDGE_NAME				"JANUS AudioBridge plugin"
 #define JANUS_AUDIOBRIDGE_AUTHOR			"Meetecho s.r.l."
@@ -695,6 +695,8 @@ typedef struct janus_audiobridge_participant {
 	gint64 last_drop;		/* When we last dropped a packet because the imcoming queue was full */
 	janus_mutex qmutex;		/* Incoming queue mutex */
 	int opus_pt;			/* Opus payload type */
+	int extmap_id;			/* Audio level RTP extension id, if any */
+	int dBov_level;			/* Value in dBov of the audio level (last value from extension) */
 	janus_audiobridge_rtp_context context;	/* Needed in case the participant changes room */
 	/* Opus stuff */
 	OpusEncoder *encoder;		/* Opus encoder instance */
@@ -713,6 +715,7 @@ typedef struct janus_audiobridge_rtp_relay_packet {
 	uint32_t ssrc;
 	uint32_t timestamp;
 	uint16_t seq_number;
+	gboolean silence;
 } janus_audiobridge_rtp_relay_packet;
 
 /* RTP forwarder instance: address to send to, and current RTP header info */
@@ -777,8 +780,8 @@ static gint janus_audiobridge_rtp_sort(gconstpointer a, gconstpointer b) {
 		"m=audio 1 RTP/SAVPF %d\r\n"		/* Opus payload type */ \
 		"c=IN IP4 1.1.1.1\r\n" \
 		"a=rtpmap:%d opus/48000/2\r\n"		/* Opus payload type */ \
-		"a=fmtp:%d maxplaybackrate=%"SCNu32"; stereo=0; sprop-stereo=0; useinbandfec=0\r\n" \
-											/* Opus payload type and room sampling rate */
+		"a=fmtp:%d maxplaybackrate=%"SCNu32"; stereo=0; sprop-stereo=0; useinbandfec=0\r\n" /* Opus payload type and room sampling rate */ \
+		"%s"								/* extmap(s), if any */
 
 /* Helper struct to generate and parse WAVE headers */
 typedef struct wav_header {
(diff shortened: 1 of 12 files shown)


lminiero:
Added `playout-delay` as well, even though just as `video-orientation` I never seem to see it in RTP packets received by the VideoRoom plugin. Not sure if it's only triggered by some specific behaviour.

plugins/janus_audiobridge.c
```diff
@@ -695,6 +695,8 @@ typedef struct janus_audiobridge_participant {
 	gint64 last_drop;		/* When we last dropped a packet because the imcoming queue was full */
 	janus_mutex qmutex;		/* Incoming queue mutex */
 	int opus_pt;			/* Opus payload type */
+	int extmap_id;			/* Audio level RTP extension id, if any */
```
AjayChoudary: These two variables can be **uint8_t**

plugins/janus_audiobridge.c
```diff
@@ -1919,8 +1924,25 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 		pkt->ssrc = 0;
 		pkt->timestamp = ntohl(rtp->timestamp);
 		pkt->seq_number = ntohs(rtp->seq_number);
+		/* Any audio level extension to quickly check if this is silence? */
+		pkt->silence = FALSE;
+		if(participant->extmap_id > 0) {
+			int level = 0;
+			if(janus_rtp_header_extension_parse_audio_level(buf, len, participant->extmap_id, &level) == 0) {
+				pkt->silence = (level == 127);
```
AjayChoudary: For time being we use some threshold to decide silence
pkt->silence = (level > configurable_threshold)?TRUE:FALSE;

the famous [hark](https://github.com/otalk/hark/blob/master/hark.bundle.js#L58) script is using -50dBov as threshold
-127 to -50 is silence, -50 to -0 is voice, we can put 100dBov as less impacted threshold.

Here we need to maintain some history of previous volume levels, then we can make more accurate decisions.

AjayChoudary:
Great @lminiero, changes are working fine for me.
Tested audiobridge & videoroom plugins.

plugins/janus_audiobridge.c
lminiero: Good point, I'll update the value and the methods that use them.

plugins/janus_audiobridge.c
lminiero: I'm personally not familiar enough with the way this level is formatted: I still haven't figured out how to convert that dBov value to a generic and linear volume identifier, for instance, something that may actually be very hard if it is indeed dependant on the saturation of codecs themselves. Working on the dBov value might be quicker, if one knows how to interpret them.

Not sure on what we should do with thresholds. I guess we could make that a property of an audioroom when you create it, so that you can decide per-room whether to be more aggressive or not. Another alternative is to make it a per-participant property, as different users may have different captures (some louder, some not). Of course nothing prevents us to have them both: if a participant doesn't have a custom value, the global room one is used; if the room doesn't have it either, 127 is used. Does this make sense per your use case?

lminiero: As a side note, this can also be dynamic, if as you suggested one introduces history in this. Anyway, that's way more complex that I can handle in an effective way, I believe, so if you have more experience on these aspects a PR after we merge this would be very welcome!

AjayChoudary:
[Post](https://groups.google.com/forum/#!topic/discuss-webrtc/KLNlIAslawQ) is something related to this PR

New [Extension](https://tools.ietf.org/html/draft-holmer-rmcat-transport-wide-cc-extensions-01) is coming in chrome-56 :)

 a=extmap:5 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions

lminiero:
We'll worry about new extensions when they arrive :wink: 
I already do see it as I always use the latest unstable build on my Fedora. Anyway, I don't plan to look into it in the near future, contributions on this aspect will be welcome after we merge this.

mirkobrankovic:
I got some warnings on incompatible types on make, but no problems on test:

```
  CC       janus-ice.o
ice.c: In function ‘janus_ice_set_stun_server’:
ice.c:775:102: warning: passing argument 3 of ‘stun_message_find_xor_addr’ from incompatible pointer type [enabled by default]
  StunMessageReturn ret = stun_message_find_xor_addr(&msg, STUN_ATTRIBUTE_XOR_MAPPED_ADDRESS, (struct sockaddr_storage *)&address, &addrlen);
                                                                                                      ^
In file included from /usr/include/stun/stunagent.h:73:0,
                 from /usr/include/stun/usages/bind.h:61,
                 from ice.c:22:

```

lminiero:
@mirkobrankovic actually I think that warning appears on master as well, and is related to the version of libnice. If you're using an older version of libnice (0.1.8 or 0.1.4, maybe?) then the arguments in `stun_message_find_xor_addr` had a different type at the time and hence the warning.

lminiero:
That said, no issue testing with Chrome and Firefox? I'm wondering if it might be a good idea to add configuration properties for rooms to decide whether you want to disable the negotiation of extensions: disabling them would have those plugins behave as they did before, and might help those scared that something might break for them otherwise.

lminiero:
Just did that, and now people can disable the usage of those RTP extensions when creating rooms, either in configuration and via the API. The default is still to negotiate those extensions, when offered by publishers, so each feature must be explicitly disabled.

Anyway, this should make it safe enough to merge now.

PR #706, opened 2016-12-09
IInclude publisher's audio_ssrc and video_ssrc in plugin_videoroom listparticipants

Sean-Der:
Right now I have each user join a room 1 publisher only, no other participants. Then they choose to forward RTP traffic as they please. The destination box has just a single audio and a single video listener, and I use the ssrc to filter streams (instead of managed ports)

I couldn't find another way to get the ssrc, and this seemed the most sensible place to put it (I think).

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1717,6 +1717,10 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 			if(p->display)
 				json_object_set_new(pl, "display", json_string(p->display));
 			json_object_set_new(pl, "publisher", (p->sdp && p->session->started) ? json_true() : json_false());
+			if ((p->sdp && p->session->started)) {
+				json_object_set_new(pl, "internal_audio_ssrc", json_integer(p->audio_ssrc));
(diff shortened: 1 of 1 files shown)


lminiero:
That's not going to work on the client side, Janus overwrites all SSRCs with the ones it generated at the core level when it gets RTP/RTCP packets from plugins. This means that whatever you'll write there is not what will arrive at the media leel to the browser.

lminiero:
Or is this just in the Admin API to provide you with an insight of the internal SSRCs the plugin sees, so that your external RTP receiver can discriminate the traffic? If so, please rename the name of the property as it's likely going to be misleading to those who use it, e.g., to something like "internal_audio_ssrc" etc.

Sean-Der:
Thanks for the review, renamed the keys!

Also wanted to say janus is FANTASTIC. I built some small things using these libs and it was a mess. Really impressive what Janus can do, and the code itself (very easy for someone to jump in and understand)

lminiero:
Thanks for the fix and the kind words too!
I see you already signed the CLA so I believe this is ok to merge now :+1:

PR #715, opened 2016-12-22
Check participant->room before using it at janus_videoroom_leave_or_unpublish

hijaq:
Probably there's a reason for checking `participant->room` for NULL, so the creation of json object that uses `participant->room->room_id` should be after that check, not before.


lminiero:
@hijaq thanks for this PR and for #716 as well. I'm on vacation until the first week of January, so I don't have time to review and test these patches until then unfortunately. I'll definitely get back to you when I'm back! Please feel free to ping me again here if I forget to do that.

lminiero:
LGTM :+1:

PR #716, opened 2016-12-22
Check for videoroom listener at janus_videoroom_incoming_rtcp

hijaq:
Check for listener `l` once and before using `l->video`.


lminiero:
LGTM :+1:

PR #722, opened 2016-12-29
fix typo thanks to lintian

linuxmaniac:
> I: janus: spelling-error-in-binary usr/bin/janus Recieved Received


lminiero:
Good catch, thanks!

Issue #723, opened 2016-12-29
provide manpage for janus


linuxmaniac:
Please provide a good old manpage for janus executable

lminiero:
Will work on that after the holidays :wink:

lminiero:
Just implemented as a PR in #726, can you check it's in line with what you'd expect out of a manpage? It's my first attempt to write one ever, so there may very well be something wrong or missing.

linuxmaniac:
Fixed by  8485d09

PR #726, opened 2017-01-02
Added manpages for janus and janus-pp-rec (addresses #723)

lminiero:
This PR adds two manpages, one for `janus` and one for `janus-pp-rec`. Since I never wrote a man-page in my life and this is my first take at this, feedback is welcome before I merge.


lminiero:
Tnx, merging then!

PR #731, opened 2017-01-05
Fix test pipeline for the streams plugin

hfiguiere:
error-resilient expect a flag, not a boolean.


lminiero:
Tnx for spotting this :+1:

PR #734, opened 2017-01-05
Fix setting a XHR.status property that has only a getter

hijaq:
Google chrome throws the following error when `janus.nojquery.js` tries to set XHR.status value:
```
TypeError: setting a property that has only a getter  
    Janus.init/</Janus.ajax/XHR.onreadystatechange
```


lminiero:
Looks like a good fix, thanks for that, merging right away!

PR #745, opened 2017-01-10
ACL and kick support in AudioBridge, VideoRoom and TextRoom

lminiero:
This PR introduces some more control you can have on participation in the room-based plugins, so AudioBridge, VideoRoom and TextRoom. More precisely, it introduces, in all of them, a few methods you can use to configure an ACL (access control list) of people you want to be able to join, so that you can limit access to only authorized users. The PR also adds a kicking feature, which can be used with or without the ACL. Both editing the ACL and kicking users is of course only possible for the person who knows the room secret, typically who created the room.

The ACL is based on a list of string based tokens, which work in a similar way to how the core-level tokens do. Basically, you associate a list of tokens to a room, either at creation time or later on: if a list is available, then each user joining must provide one of those tokens, doesn't matter which one. If it fails to do so, then the join request is rejected. The fact that users can provide just any token in the list, and is not strictly associated to a specific user, is to provide maximum flexibility: this way you can also, for instance, just assume a secret invitation code shared with all users. Of course, adding a separate token for each user allows you to have more control over who can and cannot join, e.g., to implement features like banning users (which simply means removing the related token from the ACL and kicking the user).

If you look at the updated audiobridge documentation, you'll find more information on the syntax for these new requests. You can find a few examples below just for the sake of clarity, e.g.:

Adding users to the ACL of a room:

    {
        request: "allowed",
        action: "add",
        secret: "adminpwd",
        room: 1234,
        allowed: ["ciccio", "pippo", "jimbo"]
    }

Removing users from the ACL of a room:

    {
        request: "allowed",
        action: "remove",
        secret: "adminpwd",
        room: 1234,
        allowed: ["pippo"]
    }

Temporarily disable the ACL support in a room (allows everybody in):

    {
        request: "allowed",
        action: "disable",
        secret: "adminpwd",
        room: 1234
    }

Re-enable (or enable) the ACL support in a room (needs a list, either pre-existing or provided here):

    {
        request: "allowed",
        action: "disable",
        secret: "adminpwd",
        room: 1234
    }

Kick a user from a room:

    {
        request: "kick",
        secret: "adminpwd",
        room: 1234,
        id: 1000
    }

The syntax is pretty much the same for all plugins (apart from the textroom where kicking expects a string username and not a numeric ID). Besides, notice that in the VideoRoom ACL joins and kicking only work for the publisher, but not its subscriptions (#741 might help here).

Feedback welcome!

### DIFF
--- a/html/textroomtest.js
+++ b/html/textroomtest.js
@@ -188,6 +188,19 @@ $(document).ready(function() {
 										$('#chatroom').append('<p style="color: green;">[' + getDateString() + '] <i>' + participants[username] + ' left</i></p>');
 										$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;
 										delete participants[username];
+									} else if(what === "kicked") {
+										// Somebody was kicked
+										var username = json["username"];
+										var when = new Date();
+										$('#rp' + username).remove();
+										$('#chatroom').append('<p style="color: green;">[' + getDateString() + '] <i>' + participants[username] + ' was kicked from the room</i></p>');
+										$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;
+										delete participants[username];
+										if(username === myid) {
+											bootbox.alert("You have been kicked from the room", function() {
+												window.location.reload();
+											});
+										}
 									} else if(what === "destroyed") {
 										// Room was destroyed, goodbye!
 										Janus.warn("The room has been destroyed!");
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -759,7 +759,6 @@ typedef struct janus_audiobridge_participant {
 	janus_audiobridge_room *room;	/* Room */
 	guint64 user_id;		/* Unique ID in the room */
 	gchar *display;			/* Display name (opaque value, only meaningful to application) */
-	gchar *token;			/* Invitation token, needed if the room has a list of people that can join */
 	gboolean prebuffering;	/* Whether this participant needs pre-buffering of a few packets (just joined) */
 	gboolean active;		/* Whether this participant can receive media at all */
 	gboolean working;		/* Whether this participant is currently encoding/decoding */
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -39,14 +39,15 @@ record_file =	/path/to/recording.wav (where to save the recording)
  * (invalid JSON, invalid request) which will always result in a
  * synchronous error response even for asynchronous requests. 
  * 
- * \c create , \c destroy , \c exists, \c list, \c listparticipants
- * and \c resetdecoder are synchronous requests, which means you'll
- * get a response directly within the context of the transaction.
- * \c create allows you to create a new audio conference bridge
+ * \c create , \c destroy , \c exists, \c allowed, \c kick, \c list,
+ * \c listparticipants and \c resetdecoder are synchronous requests,
+ * which means you'll get a response directly within the context of the
+ * transaction. \c create allows you to create a new audio conference bridge
  * dynamically, as an alternative to using the configuration file;
  * \c destroy removes an audio conference bridge and destroys it, kicking
  * all the users out as part of the process; \c exists allows you to
- * check whether a specific audio conference exists; \c list
+ * check whether a specific audio conference exists; \c allowed allows
+ * you to edit who's allowed to join a room via ad-hoc tokens; \c list
  * lists all the available rooms, while \c listparticipants lists all
  * the participants of a specific room and their details; finally,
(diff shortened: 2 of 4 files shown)


lminiero:
Merging.

izemize:
When "pippo" already started listening "jimbo" (video and audio stream), and i send remove request to server in background, user "pippo" dropped from room and listening stops?

lminiero:
If you're kicking "jimbo" (the publisher), his PeerConnection is closed. As such, all his subscribers ("pippo" included) are notified that "jimbo" is not available anymore, pretty much as how it normally happens when a publisher leaves the room.

izemize:
Oh, it's not good for me, because to many listeners kicked too when publisher "jimbo" is kicked. You have any idea how to drop / kick only one listener "pippo" by token/name/ or any filtering options?

lminiero:
Not sure what you mean: if the publisher goes away, the listener is useless, and exactly the same thing happens when the publisher leaves normally and is not kicked. We're not talking about the person, here, just the PeerConnection used by this person to subscribe to the kicked user. Anyway, this is not the place for this discussion. Please bring it to the list.

AGenchev:
Hi there, I tested the "kick" call. The effect was strange: my **kicked** user could continue to hear and watch the media streams from the room he was subscribed to. Only his published streams got disconnected. Did I omitted something or it works this way by design ? I implemented disconnect on the client side but all we know that this is not ... enforceable.

lminiero:
This has also been discussed [here](https://groups.google.com/forum/#!topic/meetecho-janus/DCqQAcTWpQg).

As I explained there, subscribers and publishers are very loosely related. As a matter of fact, while the typical use case is to create a publisher handle first (whether you'll actually send media or not) in order to be notified about available feeds to subscribe to, it's perfectly legitimate to *only* create subscriber handles, if you know about the feed identifiers somehow (e.g., using out of band mechanisms). This was done by design, to allow for the maximum flexibility in how the pub/sub approach works.

That said, we do have a value that could be used for the purpose. Specifically, we recently added a property called `private_id` that is returned by the plugin anytime you publish. This is an identifier that, in theory, you should set in all subscriptions you create, so that subscriptions and publisher can be associated. It is currently not used for anything except, in some cases, by Event Handlers (although `opaque_id` is a [better option there](http://www.meetecho.com/blog/correlating-janus-event-handlers/)). So we could use it for the kick feature.

The problem here, though, is that `private_id` is completely optional. We don't require it, and I'll never force people to require it, exactly to still allow people to only subscribe should their scenario need it (if you don't need a publisher handle, why force you to create it?). What we can do, though, is add a new property when creating rooms called something like `require_private_id: <true|false>`, the default being `false`. This way, if this value is `true`, we reject any attempt to subscribe if it doesn't contain a valid `private_id` we know about. As a result, when kicking we can check the `private_id` of the publisher you just kicked, and we can close the related subscriptions as well. Again, this would only work with `require_private_id: true`, which would NOT be the default: that is, you use it only if you really need it, as it seems you do. The default behaviour would still be the one we have now, for the flexilibity reasons I already mentioned that I have no intention of loosing.

If that sounds like a reasonable idea, I can add it to the TODO list. Not sure when I'll be able to work on this, though, as we're quite busy these days.

AGenchev:
Thanks for this explanation ! I don't want to cripple the usage cases, or break the existing clients, so IMO this should go in the TODO list as long as it will allow us to effectively kick a given user as this has been done in the IRC chat apps.  I need what I think the users would need.
For now, the "kick_from_listening" function is on my client side which allows an "evil" programmer to hack it and listen as long as he wants. I think that the rest of the users won't like such app logic, so I'm just trying to make the things work right.

lminiero:
Implemented in the above mentioned PR, please test carefully.

pachikichioyaji:
こんにちは
kickのテストはprivate_idをroomに付加することで可能になりました
私は質問があります、issuに書かれているaclの記述ですが
このallowdのパラメーターはdisplay:の値のリストですか、それともid:の値のリストですか?私はdisplayの値のリストを記述してみましたがうまくいきませんでした
	var rejectofcustomers = {
		request: "allowed",
		action: "add",
		secret:***",
		room: 2222,
		allowed: ["ABC","abc"],
	}
	sfutest.send({"message": rejectofcustomers});

翻訳が変でごめんなさい

lminiero:
No, the values in allowed are tokens. You need to present a valid VideoRoom token when joining, when the ACL is in place.

```
{
        "request" : "join",
        "ptype" : "publisher",
        "room" : <unique ID of the room to join>,
        "id" : <unique ID to register for the publisher; optional, will be chosen by the plugin if missing>,
        "display" : "<display name for the publisher; optional>",
        "token" : "<invitation token, in case the room has an ACL; optional>"
}
```

PR #748, opened 2017-01-11
Opaque identifier to contextualise handles

lminiero:
This PR continues the discussion started in #741.

More precisely, it adds another property you can set when creating a handle and attaching to a plugin, namely a opaque string that can be used for different purposes: it may be used to aggregate handles of the same user or type, for instance, or to uniquely identify a user across different sessions, depending what you need to do in terms of monitoring and debugging. A very simple use cases is modifying the VideoRoom demo so that a unique random ID is generated everytime the page is loaded, and you pass it to every handle you create for the participant (the one they use to publish and to subscribe): this way, when debugging you know all those handles belong to the same person. The same approach could be used in other applications or for different purposes (e.g., to identify in event handlers or admin API a specific user using your own service identifiers).

This opaque identifier is simply a string, and Janus never uses it for anything: it just saves it, and returns it when handle info is requested via the Admin API or handle attach events are triggered to event handlers. As such, they're only useful if you plan to use those features for your own monitoring and troubleshooting applications. It's a generic string, meaning that you can put whatever you want in it: a number formatted as a string, a base64 encoded piece of information, or xml/json/whatever as long as you serialize that to a string.

I've modified the `janus.js` and `janus.nojquery.js` libraries so that you can specify it by adding a `opaqueId: "<string>"` property when attaching, so it should be very straightforward to use and test.

Just as #741, I plan to merge this very soon, so feedback before that happens is welcome.

### DIFF
--- a/events.c
+++ b/events.c
@@ -111,6 +111,11 @@ void janus_events_notify_handlers(int type, guint64 session_id, ...) {
 			json_object_set_new(body, "name", json_string(name));
 			char *plugin = va_arg(args, char *);
 			json_object_set_new(body, "plugin", json_string(plugin));
+			/* Handle-related events may include an opaque ID provided by who's using the plugin:
+			 * in event handlers, it may be useful for inter-handle mappings or other things */
+			char *opaque_id = va_arg(args, char *);
+			if(opaque_id != NULL)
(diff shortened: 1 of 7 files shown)


lminiero:
Works for us, so merging.

PR #750, opened 2017-01-17
Transport-related events

lminiero:
This PR implements the transport-related events for Event Handlers, as they were the only ones left missing so far. It does this at two different levels:

1. on one side, transports themselves generate events; as different modules might implement different technologies in different ways, these events are transport- and module-specific (e.g., the HTTP transport might notify you about new HTTP requests that arrive, the WebSockets transport about connections/disconnections, etc.);
2. on another side, the existing "session created" event has been enriched with info on the transport that originated the creation.

This is a simple example of how a transport-related event for HTTP could look like:

```
   {
      "type": 128,
      "timestamp": 1484656548988936,
      "event": {
         "transport": "janus.transport.http",
         "id": "0x7f1fc00008c0",
         "data": {
            "event": "request",
            "admin_api": false,
            "ip": "::1",
            "port": 59294
         }
      }
   }
```

and this is how the new "session created" event looks like instead:

```
   {
      "type": 1,
      "timestamp": 1484656548989422,
      "session_id": 4723806468072828,
      "event": {
         "name": "created",
         "transport": {
            "transport": "janus.transport.http",
            "id": "0x7f1fc00008c0"
         }
      }
   }
```

Since the session has been created by an HTTP transport instance with ID `0x7f1fc00008c0`, you can see previous transport-related events to have an idea of where this session came from (e.g., remote address, if available).

Feedback welcome!

### DIFF
--- a/events.c
+++ b/events.c
@@ -68,10 +68,22 @@ void janus_events_notify_handlers(int type, guint64 session_id, ...) {
 			/* Core events also allocate a json_t object for its data, unref it */
 			json_t *body = va_arg(args, json_t *);
 			json_decref(body);
-		} else if(type == JANUS_EVENT_TYPE_PLUGIN || type == JANUS_EVENT_TYPE_TRANSPORT) {
+		} else if(type == JANUS_EVENT_TYPE_SESSION) {
+			/* Session events may allocate a json_t object for transport-related info, unref it */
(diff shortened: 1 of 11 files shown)


lminiero:
Played with this myself a bit and looks useful already in its current stage. Merging, we can always improve them along the road.

PR #754, opened 2017-01-20
Reconnect sockets to new IP as well

chornyitaras:
If remote media info changes janus sip plugin will update only remote ports.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -3323,7 +3323,7 @@ static void *janus_sip_relay_thread(void *data) {
 
 		if(session->media.updated) {
 			/* Apparently there was a session update */
-			if(have_server_ip)
+			if(have_server_ip) && ((inet_aton(session->media.remote_ip, &server_addr.sin_addr)) <= 0)
 				janus_sip_connect_sockets(session, &server_addr);
 			session->media.updated = FALSE;
 		}


lminiero:
I'll have a look at this on Monday, thanks.

lminiero:
I'm not sure I completely understand the fix. We already do something when the IP (and not only the ports) changes, as identified by this [line here](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_sip.c#L2982). The only reason I can see for the additional check you made is if `have_server_ip` was originally `FALSE` (offerless INVITE?), because we didn't have an IP and a renegotiation after that gave us one. Even in that case, though, the change you made is not enough, as we have to make sure the address is valid, meaning the whole if+gethostbyname check made above should be done again when `session->media.updated==TRUE` in order to update the value of `have_server_ip`, and keep the `if(have_server_ip)` as it is.

Is this what you wanted to address? Can you elaborate on your scenario, and what led you to this fix?

taraschornyi:
this is  a preparation for following: re-invire w/o sdp then ACK with sdp contains different remote media IP .  next step will be to handle sdp in ACK

chornyitaras:
this is a preparation for following scenario: re-invire w/o sdp then ACK with sdp contains different remote media IP . next step will be to handle sdp in ACK

chornyitaras:
As i understand now upon receiving re-invite with sdp with new ip janus will update only ports, Cause  [loop ](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_sip.c#L3320) is running and server_addr is not updated with new ip. So janus will update only remote ports, but not IP

lminiero:
As I said pointing to the code, we DO process an update to IP too already. Unless you can prove it doesn't work for you, I don't see the purpose of this patch.

chornyitaras:
Yes you are extracting updated ip and storing it in session->media.remote_ip. this works great. 
However in janus_sip_relay_thread line 3209 server_addr is set only once and would not change before calling janus_sip_connect_sockets in line 3327. This will case situation when sip relay thread will update only remote ports.

lminiero:
Apologies, I was misreading the patch...  I see now that your addition is not checking if the IP changed, but actually *updating* the value of server_addr. In this case yes, I see the point of this patch :wink: 
Anyway, I'd move the `inet_aton` before the `if(have_server_ip)`: this way, if `inet_aton` fails (it can happen) we can avoid an endless loop by setting `updated` to `FALSE` and print an error.

chornyitaras:
Ok I'll change it later today

lminiero:
> Anyway, I'd move the inet_aton before the if(have_server_ip): this way, if inet_aton fails (it can happen) we can avoid an endless loop by setting updated to FALSE and print an error.

Nevermind that, we set `updated=FALSE` even in that case... getting back to work after weekends shouldn't harm my judgmenent that much!

That said, could you sign our [CLA](https://cla.conf.meetecho.com/janus/) before we can merge?

lminiero:
> Ok I'll change it later today

No need, the endless loop is actually never going to happen, as we update the status of `updated` whether we update the sockets or not. I'll take care of adding an error message myself. At this point, all I need is the CLA signed and I can merge right away, thanks!

lminiero:
Merging, thanks :+1:

lminiero:
Only now noticed that the check you added was missing brackets: fixed that and added the error message.

chornyitaras:
sorry for that, was editing a file using built in GitHub editor and did not test a change thought it is to small change to make a mistake(

PR #757, opened 2017-01-26
Allow configuring SSRC when creating RTP forwarders (AudioBridge, VideoRoom)

lminiero:
As the title says. This PR allows you to configure the SSRC and/or payload type the RTP packets sent via RTP forwarder should contain. Useful if the recipient is configured to discriminate traffic on either of the two, to only accept some values, or simply expecting something specific.


lminiero:
Merging.

PR #758, opened 2017-01-30
Allow Streaming plugin to relay datachannels, and VideoRoom to forward them

lminiero:
This PR involves datachannels more in a couple plugins. Specifically, it allows you to configure a Streaming mountpoint to also relay text via datachannels, and not only audio/video as it does now. Besides, it enhances the RTP forwarding functionality in the VideoRoom plugin to allow it to also forward text received via datachannels externally. The two are of course intertwined in case you're chaining the VideoRoom input to a Streaming input, as some of you do.

For what concerns the Streaming plugin, this is particularly useful in case you want to add context to a mountpoint without using out-of-band mechanisms (e.g., you're streaming a TV channel or radio broadcast, the datachannel might be used to tell the viewer when the song or program changes). Beware, though, that the datachannel support here is currently limited to the live/RTP mountpoints: considering I'm not aware of many people using the other types, that shouldn't be an issue.

As a side note, I also added the possibility to record datachannel messages to both plugins. The syntax here reflect the one already in place for audio and video and so should be easy to infer.

As to the new features, the changes API- and configuration-wise are minimal.

## Streaming plugin

For the Streaming plugin, to create a mountpoint that includes data it's a matter of adding:

    [my-configuration]
    ...
    data = yes
    dataport = <port number>
    databuffermsg = <boolean>

to the mountpoint in the configuration file, or:

    {
        request: "create",
        ...
        data: true,
        dataport: <port number>,
        databuffermsg: <true|false>
    }

when creating via API. The `databuffermsg` you see is whether the plugin should store the latest received message, and send it to new viewers as soon as they attach: as such, it works pretty much as `videobufferkf` does for keyframes in video. Anyway, due to the fact datachannels may be effectively only set up a bit later than the WebRTC PeerConnection is actually up and running, this particular setting is currently experimental and may not always work as expected.

On the client side, you just need to make sure the viewer negotiates datachannels (e.g., adding `data:true` to `createAnswer` with `janus.js`) and can read the messages (e.g., adding the `ondataopen` and `ondata` callbacks using `janus.js`).

## VideoRoom plugin
In the VideoRoom plugin, no change is needed in the configuration or API to use datachannels. In fact, if a publisher negotiates datachannels (e.g., because they added `data:true` to their `createOffer` with `janus.js`), then subscribers will receive those as well.

Anyway, should you want to RTP-forward a publisher, you now have the possibility of forwarding data as well to the external recipient. In order to do that, the only change to the `rtp_forward` request is the fact you can add a `dataport` setting:

    {
        request: "rtp_forward",
        ...
        dataport: <port number>
    }

which, if successful, will have the response include a `data_stream_id` (forwarder stream identifier) and `data` (the port you specified) too.

## That's all folks

Should be straightforward enough to understand and test. Did a bit of testing myself and it seems to work as expected, but of course there may be a few things I'm missing. Please test and provide feedback, so that I can merge this ASAP.

### DIFF
--- a/conf/janus.plugin.streaming.cfg.sample.in
+++ b/conf/janus.plugin.streaming.cfg.sample.in
@@ -19,6 +19,7 @@
 ; audio = yes|no (do/don't stream audio)
 ; video = yes|no (do/don't stream video)
 ;    The following options are only valid for the 'rtp' type:
+; data = yes|no (do/don't stream text via datachannels)
 ; audioport = local port for receiving audio frames
 ; audiomcast = multicast group port for receiving audio frames, if any
 ; audiopt = <audio RTP payload type> (e.g., 111)
@@ -29,6 +30,8 @@
 ; videortpmap = RTP map of the video codec (e.g., VP8/90000)
 ; videobufferkf = yes|no (whether the plugin should store the latest
 ;		keyframe and send it immediately for new viewers, EXPERIMENTAL)
+; databuffermsg = yes|no (whether the plugin should store the latest
+;		message and send it immediately for new viewers)
 ; url = RTSP stream URL (only for restreaming RTSP)
 ;
 ; To test the [gstreamer-sample] example, check the test_gstreamer.sh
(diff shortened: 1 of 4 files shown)


atoppi:
Everything seems to work.

lminiero:
Merging then!

PR #760, opened 2017-01-31
Added audiolevel events to videoroom, audiobridge

mirkobrankovic:
…be calculated for X amount of packets.

This can be done also, probably, by converting int janus_rtp_header_extension_parse_audio_level from into to void to return both v and level.
videoroom incoming_rtp modified to calculate average level of dBov but event to users is not working.


lminiero:
Thanks! I did something similar talk-detection mechanism based on audio-levels in a custom project, even though that was to drive a mixer. I'll give this a better look in the next few days, but a few quick notes:

1. please fix the indentation, we don't use spaces, only tabs;
2. I don't see the point of `janus_rtp_header_extension_parse_audio_level_and_return`, we already have `janus_rtp_header_extension_parse_audio_level`;
3. why 60 packets (there are 50 in a second) and why 1200 as threshold?
4. this doesn't work if packets are not sent by the user for a while and then sent again (the count will include packets that are too old and as such not relevant).

mirkobrankovic:
Hi Lorenzo,
1. indeed i see now, will fix it :)
2. I splited it since I could not the 'value' or the level it self returned with `janus_rtp_header_extension_parse_audio_level` to plugin (probably because of my C knowledge :D)
3. I was plaing with all sort of different values for number of packets and threshold, since when I talk to mic directly packet dBov is between 0-10, but if I have a background noise it is between 30-60. So I thought maybe value of <20 is the indicator that I'm talking to mic (so 60x20 = 1200).
4. This one I didn't think of, and i'm not sure how to tackle it, I guess I need to check if there is no big difference in sequence numbers between previous and new rtp packet ?

lminiero:
The audio level in `janus_rtp_header_extension_parse_audio_level` is an input/output value, which means you'd do something like this:

```
int level = 0;
if(janus_rtp_header_extension_parse_audio_level(buf, len, id, &level) != 0) {
    JANUS_LOG(LOG_INFO, "Audio level is %d\n", level);
}
```

The return value instead tells you if the extension was found or not.

Anyway, your notifications to participants doesn't work as you're only notifying the participant himself, while it should be done in some sort of loop: see the muted/unmuted event for an example. Besides, it would be useful to have the same event passed to event handlers as well.

mirkobrankovic:
Great tjanks for pointers, 
Let me work on that more

mirkobrankovic:
I tried your example but my level is never set after parsing audio header:

```
[ERR] [plugins/janus_videoroom.c:janus_videoroom_incoming_rtp:2275] Audio level is 0
[ERR] [plugins/janus_videoroom.c:janus_videoroom_incoming_rtp:2275] Audio level is 0
[ERR] [plugins/janus_videoroom.c:janus_videoroom_incoming_rtp:2275] Audio level is 0
[ERR] [plugins/janus_videoroom.c:janus_videoroom_incoming_rtp:2275] Audio level is 0
```

lminiero:
Well, your method does exactly the same, but simply returns the value too:

	if(level)
		*level = value;
	return value;

so they're pretty much identical. I notice now though that in the snippet I gave you we check if the return value is different from 0: it should actually check if it is 0, as that's a success for the method:

    if(janus_rtp_header_extension_parse_audio_level(buf, len, id, &level) == 0) {

In your case you're getting 0 because no extension is found I believe.

mirkobrankovic:
Indeed you are right, and I forgot it now, 0 is returned if found and -1 on not found:

```
Audio level is 61
Audio level is 62
Audio level is 71
Audio level is 73
```
Then no need for another method :D

mirkobrankovic:
ok now to add event trigger :d

mirkobrankovic:
I tried to push event like all other examples from videoroom with:
`gateway->notify_event(&janus_videoroom_plugin, session->handle, info);`

but the problem is that I don't see any of the event of type 64 (plugin event type)
and events are enabled on plugin level and global level,

mirkobrankovic:
When I set:
`events = all`, I see the events but they get rejected after 5 retries and timeout since I guess I'm not ack-ing them.
and If I set `events = plugin` I don't see any plugin event at all, so maybe the filtering is not ok.

lminiero:
> If I set events = plugin I don't see any plugin event at all, so maybe the filtering is not ok.

It's `plugins`, not `plugin`: https://github.com/meetecho/janus-gateway/blob/master/events/janus_sampleevh.c#L187

> I see the events but they get rejected after 5 retries and timeout since I guess I'm not ack-ing them

If you're using the sample handler, your backend must return a 200 OK.

mirkobrankovic:
Thanks again, I need to be more carefull :)

audio_level

Event is transmited to the endpoint

lminiero:
@mirkobrankovic sorry for the lack of feedback on this, but I'm really busy these days. I'll try to give this a better review in the next few days, as it might a be an interesting feature to merge sooner or later.

plugins/janus_videoroom.c
```diff
@@ -2292,6 +2269,36 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
 	if(!session || session->destroyed || !session->participant || session->participant_type != janus_videoroom_p_type_publisher)
 		return;
 	janus_videoroom_participant *participant = (janus_videoroom_participant *)session->participant;
+	if(participant->audio_active) {
```
lminiero: I believe all this should be optional, that is, when you create a room you should be able to decide whether you want to be notified about who's speaking or not. The way it works now, this seems to be automatically done when the extension is found, am I correct?

plugins/janus_videoroom.c
```diff
@@ -2292,6 +2269,36 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
...
+		int level = 0;
+		if(janus_rtp_header_extension_parse_audio_level(buf, len, participant->audio_level_extmap_id, &level) == 0) {
+			// JANUS_LOG(LOG_INFO, "Audio level is %d\n", level);
+			participant->audio_dBov_sum = participant->audio_dBov_sum + level;
+			participant->audio_active_packets = participant->audio_active_packets + 1;
+		// 2 seconds of talking (100 packets) with average of ~25 dBow
```
lminiero: Wrong indentation. Besides, please use `/* Comment */` instead of `// Comment`.

plugins/janus_videoroom.c
```diff
@@ -2292,6 +2269,36 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
...
+			if(participant->audio_active_packets == 100) {
+				if(participant->audio_dBov_sum < 2500) {
```
lminiero: The two settings above should be made configurable as well, so that people can fine tune the behaviour themselves depending on their application, their needs or their taste.

plugins/janus_videoroom.c
```diff
@@ -2292,6 +2269,36 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
...
+					janus_videoroom_notify_participants(participant, event);
+					json_decref(event);
+					JANUS_LOG(LOG_ERR, "AVG audio_level %f\n", (float)participant->audio_dBov_sum/(float)participant->audio_active_packets);
+					/* Also notify event handlers */
+					if(notify_events && gateway->events_is_enabled()) {
+						json_t *info = json_object();
+						json_object_set_new(info, "event", json_string("talking"));
+						json_object_set_new(info, "room", json_integer(participant->room->room_id));
+						json_object_set_new(info, "user", json_string(participant->display));
+						gateway->notify_event(&janus_videoroom_plugin, session->handle, info);
```
lminiero: I'm wondering if we do need the `session->handle` here: without it, it becomes an event the plugin generates by itself and not strictly associated with a specific user (e.g., as a room being created). On the other end, it is triggered by incoming media from a specific user, so it's probably fine to keep it too.

rtp.c
```diff
@@ -158,6 +158,19 @@ int janus_rtp_header_extension_parse_audio_level(char *buf, int len, int id, int
 	return 0;
 }
 
+int janus_rtp_header_extension_parse_audio_level_and_return(char *buf, int len, int id, int *level) {
```
lminiero: I don't think we need this, as you're using the method that existed already.

rtp.h
```diff
@@ -107,6 +107,14 @@ const char *janus_rtp_header_extension_get_from_id(const char *sdp, int id);
...
 int janus_rtp_header_extension_parse_audio_level(char *buf, int len, int id, int *level);
 
+/*! \brief Helper to parse a ssrc-audio-level RTP extension (https://tools.ietf.org/html/rfc6464)
+ * @param[in] buf The packet data
+ * @param[in] len The packet data length in bytes
+ * @param[in] id The extension ID to look for
+ * @param[out] level The level value in dBov (0=max, 127=min)
+ * @returns dBov level of audio, 0 otherwise */
+int janus_rtp_header_extension_parse_audio_level_and_return(char *buf, int len, int id, int *level);
+
```
lminiero: See above.

mirkobrankovic: Hi @lminiero, thanks for the feedback, I guess merge done wrong, I'll try to rebase to master.

Strange thing is that I don't receive notifications on replies, i need to enable this.

lminiero:
I won't be able to review this before I come back next week. Please ping me if I forget to do that.

notedit:
cool,  need some thing like this in audio bridge,  waiting for your merge.

mirkobrankovic:
@lminiero I just pushed commit with configurable vars from file and config json (I tested only json create with audiolevel_event = true/false), also tested average level of 25 and 50 and correctly got events.
Problem might be that `janus_rtp_header_extension_parse_audio_level` is called for every packet and that one have longing, so when running on DEBUG level it is hard to read anything.

PS. I think I synced everithing with latest master, since there were some removals of 
`memset(buf, 0, 20);` and only one left, so please double check if I missed anything.
My Meld said everything is in sync. 

Thanks,
mirko

lminiero:
@mirkobrankovic thanks! I'll look into this when I get back home next week. I'm still not sure about a simple count of packets, as if packets don't arrive for a while you may take into account stuff that is too old. I'll try and see how I did it in my prototype, which took into account the time it passed since the last check instead of a count, and come back to you with more details to possibly improve this.

mirkobrankovic:
@lminiero great, thanks :)

notedit:
@mirkobrankovic @lminiero  have any thought  on implement this on audio bridge?

lminiero:
I actually still have to come back to @mirkobrankovic with some additional info I promised (my fault, too many things to do). What I can try to do next week is implementing something similar for the AudioBridge using the mechanisms I mentioned (in particular, the time based collection of audio levels, instead of number of events) so that Mirko can adapt this PR to do the same.

notedit:
@lminiero  thanks,  you are a hero.

lminiero:
Actually, checked my old code and it works pretty much like Mirko's, so there's no need to adapt to any timing related info... which means we can re-use the same codebase for the AudioBridge too, probably. @mirkobrankovic could you just solve the conflicts then, so that I can make a final review and then merge?

mirkobrankovic:
@lminiero done, hopefully i didn't do any nasty thoings anymore :D

mirkobrankovic:
@notedit i understood that it is already present in audio_room, if it is not, i can give it a try, follow same logic i did in video one

lminiero:
> @lminiero done, hopefully i didn't do any nasty things this time :D

Thanks! I'll give it a quick check on Monday and then merge :+1: 

> @notedit i understood that it is already present in audio_room, if it is not, i can give it a try, follow same logic i did in video one

IIRC (I should really remember better the code I wrote myself) we only use audio levels as an helper to decide whether to drop a packet or not if it's silence, and so not mix it. No talker indication.

notedit:
@mirkobrankovic  as @lminiero  said, just helper code  No talker indication

mirkobrankovic:
@notedit, @lminiero, great then all packets that are not dripped can be mesured for avg level, lets see from monday

mirkobrankovic:
@lminiero , i think it is not working when set from config file, i might be missing something. Create room event is ok, but static config is not set.
I implemented similar on audiobridge and now i see it is not creating events

mirkobrankovic:
@lminiero in join and start commands there is no properties for this:
```
{"janus":"message","body":{"request":"start","room":1234},"transaction":"QSulkzekAT8Z","jsep":{
```
but only for create and also in config file:
```
[janus.plugin.audiobridge.cfg]
    [general]
    [1234]
        description: Demo Room
        secret: adminpwd
        sampling_rate: 16000
        audiolevel_event: yes
        audio_active_packets: 100
        audio_level_average: 25
        record: false
```
so I'm confused if it needs to be added on join/start, since `participant->room->...` is empty

mirkobrankovic:
Gbut sme on audiobridge code, when i create room with create command and pass the config I get it working ok:
`[ERR][plugins/janus_audiobridge.c:janus_audiobridge_incoming_rtp:2289] AVG audio_level 12.350000`

mirkobrankovic:
I found a problem, I used `janus_is_true(audio_active_packets->value);`
for int/guint values, not sure if there is `janus_is_int`

mirkobrankovic:
@lminiero , my code was the problem, parsing the integer options from files with `janus_is_true()`. 
I'm not sure what should i use atm

lminiero:
If you mean from the config file, do an `atoi` and then check if it's greater than zero. Other plugins do the same (the AudioBridge plugin itself has similar checks). If you mean from API, instead, you shouldn't do validation yourself, as that's responsibility of `JANUS_VALIDATE_JSON_OBJECT`.

mirkobrankovic:
exactly, just found atoi. Thanks.
Provisioned from config file is still not ok, just need to check what is null.

mirkobrankovic:
@lminiero should I push audiobridge as seperate pull ? or to add it to same PR

lminiero:
I think we can have it in the same PR, thanks! As it's basically the same feature implemented the same way :+1:

mirkobrankovic:
talking
Should be fixed also for config file provisioned rooms.

lminiero:
Looks like you're pushing the display instead of the numeric user ID, though? The user ID should be the only one present, with the display optionally there if it's available, as applications may want to identify the user based on the ID instead (display is not unique).

mirkobrankovic:
good point, changing it now

mirkobrankovic:
talking
"Mirko, take a snickers.... " :D

conf/janus.plugin.audiobridge.cfg.sample
```diff
@@ -19,5 +22,8 @@
 description = Demo Room
 secret = adminpwd
 sampling_rate = 16000
+audiolevel_event = no
+audio_active_packets = 100
+audio_level_average = 25
```
lminiero: If the default for `audiolevel_event` is "no", I'd remove these three lines from the sample room configuration. They're already documented above so no need for a placeholder there.

plugins/janus_audiobridge.c
```diff
@@ -86,6 +86,9 @@ record_file =	/path/to/recording.wav (where to save the recording)
 	"allowed" : [ array of string tokens users can use to join this room, optional],
 	"sampling" : <sampling rate of the room, optional, 16000 by default>,
 	"audiolevel_ext" : <true|false, whether the ssrc-audio-level RTP extension must be negotiated for new joins, default true>,
+	"audiolevel_event" : yes|no (whether to emit event to other users or not),
+    "audio_active_packets" : 100 (number of packets with audio level, default=100, 2 seconds),
+    "audio_level_average" : 25 (average value of audio level, 127=muted, 0='too loud', default=25),
```
lminiero: Broken indentation in the two lines above (should be tabs).

plugins/janus_audiobridge.c
```diff
@@ -623,6 +626,9 @@ static struct janus_json_parameter create_parameters[] = {
 	{"record_file", JSON_STRING, 0},
 	{"permanent", JANUS_JSON_BOOL, 0},
 	{"audiolevel_ext", JANUS_JSON_BOOL, 0},
+	{"audiolevel_event", JANUS_JSON_BOOL, 0},
+    {"audio_active_packets", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
+    {"audio_level_average", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
```
lminiero: Broken indentation as above.

plugins/janus_audiobridge.c
```diff
@@ -725,6 +731,9 @@ typedef struct janus_audiobridge_room {
 	gboolean is_private;		/* Whether this room is 'private' (as in hidden) or not */
 	uint32_t sampling_rate;		/* Sampling rate of the mix (e.g., 16000 for wideband; can be 8, 12, 16, 24 or 48kHz) */
 	gboolean audiolevel_ext;	/* Whether the ssrc-audio-level extension must be negotiated or not for new joins */
+	gboolean audiolevel_event;	/* Whether to emit event to other users about audiolevel */
+    int audio_active_packets;	/* amount of packets with audio level for checkup */
+    int audio_level_average;	/* average audio level */
```
lminiero: Broken indent.

plugins/janus_audiobridge.c
```diff
@@ -1034,6 +1045,9 @@ int janus_audiobridge_init(janus_callbacks *callback, const char *config_path) {
 			janus_config_item *priv = janus_config_get_item(cat, "is_private");
 			janus_config_item *sampling = janus_config_get_item(cat, "sampling_rate");
 			janus_config_item *audiolevel_ext = janus_config_get_item(cat, "audiolevel_ext");
+			janus_config_item *audiolevel_event = janus_config_get_item(cat, "audiolevel_event");
+            janus_config_item *audio_active_packets = janus_config_get_item(cat, "audio_active_packets");
+            janus_config_item *audio_level_average = janus_config_get_item(cat, "audio_level_average");
```
lminiero: Broken indent.

plugins/janus_audiobridge.c
```diff
@@ -1070,6 +1084,15 @@ int janus_audiobridge_init(janus_callbacks *callback, const char *config_path) {
...
 				audiobridge->audiolevel_ext = janus_is_true(audiolevel_ext->value);
+			audiobridge->audiolevel_event = FALSE;
+            if(audiolevel_event != NULL && audiolevel_event->value != NULL)
+            	audiobridge->audiolevel_event = janus_is_true(audiolevel_event->value);
+            audiobridge->audio_active_packets = 100;
+            if(audio_active_packets != NULL && audio_active_packets->value != NULL)
+                audiobridge->audio_active_packets = atol(audio_active_packets->value);
+            audiobridge->audio_level_average = 25;
+            if(audio_level_average != NULL && audio_level_average->value != NULL)
+                audiobridge->audio_level_average = atol(audio_level_average->value);
```
lminiero: Broken indent.

plugins/janus_audiobridge.c
```diff
@@ -1070,6 +1084,15 @@ int janus_audiobridge_init(janus_callbacks *callback, const char *config_path) {
 			audiobridge->audiolevel_ext = TRUE;
 			if(audiolevel_ext != NULL && audiolevel_ext->value != NULL)
 				audiobridge->audiolevel_ext = janus_is_true(audiolevel_ext->value);
+			audiobridge->audiolevel_event = FALSE;
+            if(audiolevel_event != NULL && audiolevel_event->value != NULL)
+            	audiobridge->audiolevel_event = janus_is_true(audiolevel_event->value);
+            audiobridge->audio_active_packets = 100;
+            if(audio_active_packets != NULL && audio_active_packets->value != NULL)
+                audiobridge->audio_active_packets = atol(audio_active_packets->value);
```
lminiero: Why `atol` for those two values, if they're both `int`? `atoi` is the right method for that.

mirkobrankovic:
maybe I added back indentation problems, since i'm switching a lot of editors on different projects

plugins/janus_audiobridge.c
```diff
@@ -1256,6 +1279,21 @@ void janus_audiobridge_destroy_session(janus_plugin_session *handle, int *error)
 	return;
 }
 
+static void janus_audiobridge_notify_participants(janus_audiobridge_participant *participant, json_t *msg) {
+	/* participant->room->participants_mutex has to be locked. */
```
lminiero: The AudioBridge plugin doesn't have any `participants_mutex`. I guess this should refer to the `mutex` property instead, which should be locked when you call `janus_audiobridge_notify_participants` below.

plugins/janus_audiobridge.c
```diff
@@ -1372,6 +1410,9 @@ struct janus_plugin_result *janus_audiobridge_handle_message(janus_plugin_sessio
 		json_t *allowed = json_object_get(root, "allowed");
 		json_t *sampling = json_object_get(root, "sampling");
 		json_t *audiolevel_ext = json_object_get(root, "audiolevel_ext");
+		json_t *audiolevel_event = json_object_get(root, "audiolevel_event");
+        json_t *audio_active_packets = json_object_get(root, "audio_active_packets");
+        json_t *audio_level_average = json_object_get(root, "audio_level_average");
```
lminiero: Broken indent.

plugins/janus_audiobridge.c
```diff
@@ -1452,6 +1493,9 @@ struct janus_plugin_result *janus_audiobridge_handle_message(janus_plugin_sessio
 		else
 			audiobridge->sampling_rate = 16000;
 		audiobridge->audiolevel_ext = audiolevel_ext ? json_is_true(audiolevel_ext) : TRUE;
+		audiobridge->audiolevel_event = audiolevel_event ? json_is_true(audiolevel_event) : FALSE;
+        audiobridge->audio_active_packets = audio_active_packets ? json_integer_value(audio_active_packets) : 100;
+        audiobridge->audio_level_average = audio_level_average ? json_integer_value(audio_level_average) : 25;
```
lminiero: Broken indent.

plugins/janus_audiobridge.c
```diff
@@ -2219,11 +2263,43 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
 			int level = 0;
-			if(janus_rtp_header_extension_parse_audio_level(buf, len, participant->extmap_id, &level) == 0) {
+			if (janus_rtp_header_extension_parse_audio_level(buf, len, participant->extmap_id, &level) == 0) {
+				participant->audio_dBov_sum = participant->audio_dBov_sum + level;
+				participant->audio_active_packets = participant->audio_active_packets + 1;
 				pkt->silence = (level == 127);
 				participant->dBov_level = level;
+				if (participant->room->audiolevel_event) {
+					if (participant->audio_active_packets == participant->room->audio_active_packets) {
+
```
lminiero: Why the empty space here?

plugins/janus_audiobridge.c
```diff
@@ -2219,11 +2263,43 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+						if ((float) participant->audio_dBov_sum / (float) participant->audio_active_packets <
```
lminiero: Shouldn't this check if `participant->audio_active_packets` is different than zero? Otherwise that math operation might result in a divide by zero and crash Janus.

plugins/janus_audiobridge.c
```diff
@@ -2219,11 +2263,43 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+							participant->room->audio_level_average) {
+							// Notify participants
+							json_t *event = json_object();
+							json_object_set_new(event, "event", json_string("talking"));
+							json_object_set_new(event, "room", json_integer(participant->room->room_id));
+							json_object_set_new(event, "user", json_integer(participant->user_id));
+							janus_audiobridge_notify_participants(participant, event);
```
lminiero: This should be wrapped in the mutex above.

plugins/janus_videoroom.c
```diff
@@ -834,6 +848,15 @@ int janus_videoroom_init(janus_callbacks *callback, const char *config_path) {
 			videoroom->audiolevel_ext = TRUE;
 			if(audiolevel_ext != NULL && audiolevel_ext->value != NULL)
 				videoroom->audiolevel_ext = janus_is_true(audiolevel_ext->value);
+			videoroom->audiolevel_event = FALSE;
+			if(audiolevel_event != NULL && audiolevel_event->value != NULL)
+				videoroom->audiolevel_event = janus_is_true(audiolevel_event->value);
+			videoroom->audio_active_packets = 100;
+			if(audio_active_packets != NULL && audio_active_packets->value != NULL)
+				videoroom->audio_active_packets = atol(audio_active_packets->value);
```
lminiero: Same points on `atoi`.

plugins/janus_videoroom.c
```diff
@@ -2212,6 +2241,38 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
...
+	janus_videoroom *videoroom = participant->room;
+
+	if(videoroom->audiolevel_event && participant->audio_active) {
+		int level = 0;
+		if(janus_rtp_header_extension_parse_audio_level(buf, len, participant->audio_level_extmap_id, &level) == 0) {
+			/* JANUS_LOG(LOG_INFO, "Audio level is %d\n", level); */
+			participant->audio_dBov_sum = participant->audio_dBov_sum + level;
+			participant->audio_active_packets = participant->audio_active_packets + 1;
+			if(participant->audio_active_packets == videoroom->audio_active_packets) {
+				if((float)participant->audio_dBov_sum/(float)participant->audio_active_packets < videoroom->audio_level_average) {
```
lminiero: Same point on the divide by zero.

lminiero:
> maybe I added back indentation problems, since i'm switching a lot of editors on different projects

Yep, just reviewed the changes and highlighted that too :wink:

plugins/janus_audiobridge.c
mirkobrankovic: Yes , in videoroom I used this one:
`janus_audiobridge_notify_participants`
but in audiobridge there was no such helper but on every place notification code part was done separately, so  Idon't know if this will be ok

mirkobrankovic:
@lminiero  can you do 'one' more review ... with a note on one :D

lminiero:
Looks good now, thanks for the quick fixes! There still are a couple of things I'd ask you to fix, though:

1. To keep everything consistent and avoid a false positive, the levels list and the timer should be reset when you do a `changeroom` in the AudioBridge (the PeerConnection lives on but you're attached to a different room). This should also be done at the both plugins' `hangup_media` notification for the publishers, as otherwise if you unpublish and then publish again after a while, your levels from the previous PeerConnection still count when it comes at evaluating your speech detection.
2. When parsing the audio level configuration, there should be checks to see if they're <= 0; if they are, a warning should be displayed that says we'll use the default values instead (this applies to the VideoRoom too). Not sure what happens if one provides negative values for instance, which is still something we should prevent.

mirkobrankovic:
Good point.
I've done first point and i will test second one tomorrow and then push it. Cheers

mirkobrankovic:
ok,
Config file:
```
[WARN] audio_active_packets shouldn't be <= 0, will use default value: 100!!!
[WARN] audio_level_average shouldn't be <= 0, will use default value: 25!!!
```
From json `create` request:
```
Creating new session: 1496897427839280
Creating new handle in session 1496897427839280: 4154945220672631
[WARN] audio_active_packets shouldn't be <= 0, will use default value: 100!!!
[WARN] audio_level_average shouldn't be <= 0, will use default value: 25!!!
[4154945220672631] Creating ICE agent (ICE Full mode, controlled)
[4154945220672631] The DTLS handshake has been completed
```
and ofcourse i rtested if i got event after both, and was ok.

@lminiero can you review it again ?

mirkobrankovic:
I need to be more carefull with this close/open button, there should be a yes/no popup

lminiero:
Thanks! This is still missing a reset on `changeroom` though? In the meanwhile I'll give this a go myself so that I can merge right after that last fix :+1:

conf/janus.plugin.videoroom.cfg.sample
```diff
@@ -10,6 +10,9 @@
 ; audiocodec = opus|isac32|isac16|pcmu|pcma (audio codec to force on publishers, default=opus)
 ; videocodec = vp8|vp9|h264 (video codec to force on publishers, default=vp8)
 ; audiolevel_ext = yes|no (whether the ssrc-audio-level RTP extension must
+; audiolevel_event = yes|no (whether to emit event to other users or not, default=no)
+; audio_active_packets = 100 (number of packets with audio level, default=100, 2 seconds)
+; audio_level_average = 25 (average value of audio level, 127=muted, 0='too loud', default=25)
 ;		be negotiated/used or not for new publishers, default=yes)
```
lminiero: Looks like the `be negotiated` part belongs to after `audiolevel_ext = yes|no`, possibly a copy/paste error when you added your new settings?

mirkobrankovic:
@lminiero done :)

mirkobrankovic:
As Corn would say: https://youtu.be/xmOOGeZE-aE?t=138

lminiero:
> As Corn would say: https://youtu.be/xmOOGeZE-aE?t=138

Did you mean Korn? :wink: 
Yep, playing with the PR right now to test it myself :+1: 
I noticed that in the VideoRoom `incoming_rtp` callback, you do:

    if(videoroom->audiolevel_event && participant->audio_active) {

I'd add an additional check to see if it's an audio packet, because otherwise you process video packets too with the existing code: parsing the extension would always fail, but it's unneeded operations. So something like:

        if(!video && videoroom->audiolevel_event && participant->audio_active) {

Back to testing, will let you know if it's free to go, thanks again!

mirkobrankovic:
great, I forgot that :) added now.

Thanks @lminiero

mirkobrankovic:
This PR is more looking like:
https://twitter.com/iamdevloper/status/851445899008106496 
Hehe, my fault :D

lminiero:
Seems to be working nicely! The thresholds will be tricky to configure but that's an issue in every conferencing system.

A couple of very quick fixes, if possible:

1. in both plugins, use `audiobridge` and `videoroom` instead of `event` when sending a "talking" event: this is more consistent with the way we send all other event in there, and in how we process those events in the JavaScript code;
2. in the same code, use `id` instead of `user` for the user ID.

As soon as you can do that, I'll merge as it seems to be working nicely!

I'll probably change the feature a bit, in order to also notify when a user stops talking: this would make it easier to render on the client side, as otherwise your only choice in JavaScript is, e.g., showing an element when you get the event, and automatically hide it after an arbitrary amount of time (e.g., 2 seconds) that may or may not be in sync with how the room was configured (what if you chose to check every second instead of every two seconds?). If you keep track of whether you notified a user talking last time, if the user is not talking anymore you can send the "not talking anymore" event that the client can use. Again, nothing you have to do here, I'll do it myself based on the code you added...

mirkobrankovic:
@lminiero , thanks for quick responses :D

lminiero:
This is a very good start, so any enhancement can be done later along the way... thanks again, merging then!

atyenoria:
@mirkobrankovic @lminiero 
This functionality works well. I'm very thankful for this.
I have a question. Is it possible to detect the change from the state of speaking to the state of not speaking? 
At now, I can get only the  the change from the state of  not speaking to the state of speaking.
This is very similar to https://github.com/otalk/hark .
Btw, for mobile app development, react-native-webrtc can't get the audio level ( https://github.com/oney/react-native-webrtc/issues/276 )
So, This feature is very good solution for detecting the speaker for both web and mobile at now.

lminiero:
@atyenoria that's a good point, and something I wanted to add but never had the time to do. Can you create a new issue so that we can keep track of this feature request? Thanks!

atyenoria:
@lminiero DONE

PR #765, opened 2017-02-07
Make JSON reponse for created textrooms consistent with other plugins

thehunmonkgroup:
All other plugins which deal with programmatic creation of rooms
respond with the plugin name and 'created', while the textroom
plugin reponds with 'success'. This fixes the inconsistency.

### DIFF
--- a/plugins/janus_textroom.c
+++ b/plugins/janus_textroom.c
@@ -1441,7 +1441,7 @@ void janus_textroom_handle_incoming_request(janus_plugin_session *handle, char *
 		if(!internal) {
 			/* Send response back */
 			json_t *reply = json_object();
-			json_object_set_new(reply, "textroom", json_string("success"));
+			json_object_set_new(reply, "textroom", json_string("created"));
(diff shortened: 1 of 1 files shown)


lminiero:
While it's good to have consistency (although it's different plugins and each of them can do as they want), I'm afraid this would break applications that check for `success` in the response there. As such, I'd integrate this in a PR that is much more disruptive and will force people to update their applications anyway, like the reference counters branch.

thehunmonkgroup:
I think it's particularly helpful to have consistency in the plugins shipped with Janus core.

So I think I hear you saying that you'll accept the change, but only in a cycle where there are other breaking API changes? If so, that sounds fine to me. Would you consider merging it with the reference counters branch?

lminiero:
Sure, I can make the change myself on the branch there.

thehunmonkgroup:
Great! Shall we leave this issue open as a placeholder until you merge it there?

lminiero:
Yep, I'll take care of that later on.

lminiero:
Closing as addressed in my PR. Thanks!

PR #767, opened 2017-02-07
Fix typo in textroom plugin log message for list command

thehunmonkgroup:
Says 'video room' when it should say 'text room'


lminiero:
Thanks for spotting this.

As a side note, have you signed our [CLA](https://cla.conf.meetecho.com/janus/)? That's required before we can merge any external contribution.

thehunmonkgroup:
CLA so signed

lminiero:
Thanks! Merging then.

Issue #766, opened 2017-02-07
textroom plugin 'list' command doesn't seem to work


thehunmonkgroup:
I'm able to successfully execute 'list' commands to the videoroom and audiobridge plugin, however using the same code with the textroom plugin fails.

Here's the output when I execute list on a videoroom plugin handle:

```
debug: Executing Janus videoroom plugin command with message: {"request":"list"}
Sending message to plugin (handle=312643734989790):
{ janus: 'message',
body: { request: 'list' },
transaction: 'cuUwzTs1E2DR' }
Got a success on session 1169148762667865
{ janus: 'success',
session_id: 1169148762667865,
sender: 312643734989790,
transaction: 'cuUwzTs1E2DR',
plugindata:
{ plugin: 'janus.plugin.videoroom',
data: { videoroom: 'success', list: [Object] } } }
Message sent!
{ janus: 'success',
session_id: 1169148762667865,
sender: 312643734989790,
transaction: 'cuUwzTs1E2DR',
plugindata:
{ plugin: 'janus.plugin.videoroom',
data: { videoroom: 'success', list: [Object] } } }
Synchronous transaction successful (janus.plugin.videoroom)
{ videoroom: 'success',
list:
[ { room: 1234,
description: 'Demo Room',
max_publishers: 6,
bitrate: 128000,
fir_freq: 10,
audiocodec: 'opus',
videocodec: 'vp8',
record: false,
num_participants: 0 } ] }
RETURN DATA
{ videoroom: 'success',
list:
[ { room: 1234,
description: 'Demo Room',
max_publishers: 6,
bitrate: 128000,
fir_freq: 10,
audiocodec: 'opus',
videocodec: 'vp8',
record: false,
num_participants: 0 } ] }
```

And then running the same list command on a textroom plugin handle:

```
debug: Executing Janus textroom plugin command with message: {"request":"list"}
Sending message to plugin (handle=759819753124802):
{ janus: 'message',
body: { request: 'list' },
transaction: '6p1uVgMZAz14' }
Got an ack on session 4303262646572199
{ janus: 'ack',
session_id: 4303262646572199,
hint: 'I\'m taking my time!',
transaction: '6p1uVgMZAz14' }
Message sent!
{ janus: 'ack',
session_id: 4303262646572199,
hint: 'I\'m taking my time!',
transaction: '6p1uVgMZAz14' }
RETURN DATA
undefined
```

[...]

PR #769, opened 2017-02-08
Allow some TextRoom commands to be sent via Janus API

lminiero:
So far, all TextRoom commands had to be sent via datachannels. As explained in #766, it might make sense to have some room-management requests available to the regular plugin messaging as well, in order to be able, e.g., to create rooms on the server side without having to create a datachannel.

This PR implements this, and allows for the usage of `create`, `destroy`, `exists` and `list` to be invoked via plugin messaging. All the other requests still need a datachannel. Notice that when using plugin messaging the syntax for the request name will have to be `request:"name"` and not `textroom:"name"` (e.g., `request:"create"` and not `textroom:"create"`); all other properties are named exactly the same.

Tested briefly and it seems to be working for me, but please play with this more than I just did, in order to make sure I didn't break anything. If everything works fine and no leaks are found (try spotting those as well), the plan is to merge soon.

### DIFF
--- a/html/textroomtest.js
+++ b/html/textroomtest.js
@@ -205,6 +205,8 @@ $(document).ready(function() {
 											});
 										}
 									} else if(what === "destroyed") {
+										if(json["room"] !== myroom)
+											return;
 										// Room was destroyed, goodbye!
 										Janus.warn("The room has been destroyed!");
 										bootbox.alert("The room has been destroyed", function() {
--- a/plugins/janus_textroom.c
+++ b/plugins/janus_textroom.c
@@ -1525,7 +1525,8 @@ janus_plugin_result *janus_textroom_handle_incoming_request(janus_plugin_session
 		if(!internal) {
 			/* Send response back */
 			reply = json_object();
-			json_object_set_new(reply, "textroom", json_string("success"));
+			/* Notice that we reply differently if the request came via Janus API */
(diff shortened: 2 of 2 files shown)


thehunmonkgroup:
Tested all of ```create```, ```destroy```, ```list```, ```exists```, all seem to work perfectly. I'm not clear how to spot leaks, would need guidance on that.

Since these are new API commands, and thus will not break backwards compat, I'd strongly suggest implementing the same consistency in responses that I referred to in #765 before merging this PR.

For example, this patch currently returns ```{ textroom: 'success' }``` for both the create and destroy calls, when to be consistent with the other plugins, it should return ```{ textroom: 'created' }``` for create calls, and ```{ textroom: 'destroyed' }``` for destroy calls.

lminiero:
These are not new API commands, it's exactly the same commands that are channeled through a different API: the method used to process the request and return the results is the same.

lminiero:
While it's true that the message processing is shared, the processor does know where the request comes from: so, for requests coming from the Janus API (which as you said, can indeed be considered a new API) we reply with `created` and `destroyed` instead of `success`. The old approach is still used for creates/destroys over datachannels.

Does that work for you @thehunmonkgroup? This should indeed fix #765 even without doing that in the reference counters.

thehunmonkgroup:
Yes, works for me!

lminiero:
Looks like we can merge, so going forward. In case we spot issues we can fix them later on.

Issue #771, opened 2017-02-09
[renamed] Configurable media timeouts


bhakimi:
we could use this feature also, any chance this can be added to the project?

https://groups.google.com/forum/#!msg/meetecho-janus/U47gFCFreqw/t0PtQjgpBAAJ

lminiero:
Closing as we only use Github for issues. Please keep discussions like this on the group.

lminiero:
Thinking about it, it makes sense to have an issue here as a reminder, as that might be a useful feature. Could you rename the issue to something that refers to its purpose, e.g., "Configurable media timeouts"?

lminiero:
Done, please test and let us know if it works as expected.

bhakimi:
will test out in a few days, thank you for the prompt response

bhakimi:
just to be clear, i can set this in janus.cfg right?

lminiero:
Yeah https://github.com/meetecho/janus-gateway/pull/773/files#diff-92061b7f60d74c096d2180c059d27a15R66

lminiero:
Closing as addressed in #773 which we merged.

PR #773, opened 2017-02-09
Configurable timeout for the 'not receiving audio/video' events

lminiero:
This PR allows for a configuration of the timeout value for the "not receiving audio/video" events. By default, these events are kicked after 1 seconds of no media being received (`{janus:"media", .. , receiving:false}`): a new event of the same type is notified only when media is being received again  (`{janus:"media", .. , receiving:true}`).

This patch allows you to change that value of 1 seconds, which was hardcoded in the core, to something else. A value of 0 disables the "not receiving..." events entirely (but not the "now receiving.." ones). The configuration can be done in `janus.cfg`, command line, or even in real-time using the Admin API.

Addresses #771, feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,8 +342,11 @@ or on the command line:
                                   (default=off)
 	-u, --rtcp-mux                Whether to force rtcp-mux or not (whether RTP
                                   and RTCP will always be muxed)  (default=off)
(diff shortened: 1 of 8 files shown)


lminiero:
Merging as confirmed it works [here](https://groups.google.com/forum/#!topic/meetecho-janus/U47gFCFreqw).

PR #774, opened 2017-02-09
Configurable session timeout value

lminiero:
The value of the session timeout dictates how long we should tolerate inactivity (as in no Janus API messages received) on a session. If no messages were received for a number of seconds that exceeds the timeout, we assume the session is dead and so we free the resources.

Time and again I've been asked to make the session timeout configurable (by default it was hardcoded to 60 seconds). This PR addresses that, and makes it configurable in `janus.cfg`, command line, and even via Admin API.

A value of `0` is possible, and completely disables session timeout checking. Anyway, you don't want that, as that will easily cause orphaned sessions (clients that died and didn't clean up) whose resources are never freed.

Feedback welcome, as I hope to merge soon.


lminiero:
Merging.

Issue #729, opened 2017-01-04
Naming convention for mjr files safe access.


Horsetopus:
Following [this discussion](https://groups.google.com/forum/#!topic/meetecho-janus/WDwC-oyM0jk), I'd like to propose some conventions regarding the naming of the mjr files.
During the recording, the files should have a .tmp extension.
This is so that there is a clear way to know when the files can be accessed for post processing.
I would make a pull request, but the cleanest way to do it is to change the janus_recorder structure to keep some info regarding the file path, and I am confident enough in my skills in C or my nowledge of hwo this might affect Janus.

lminiero:
Sorry, I saw this last week and forgot to answer... while a reasonable approach to somehow signal applications that are unaware of the signalling (and that so don't know a "stop recording" has just been issued, for instance), I'll have to give this some more thought, as the recording and post-processing is a feature heavily used in production by several people (I know we use it A LOT), and so I don't want to risk breaking any feature that relies on that and works right now.

Horsetopus:
Well, my issue is I couldn't find a way to be 100% sure that the file was closed with the signalling.
In the case of the videoRoom plugin can I assume, or more importantly be sure, that by the time I get the success event for the configure request, the files have been closed?

lminiero:
The event is sent after we invoke the methods to close the recording, and
it's done in the same thread, so it should be ok. Anyway, I'm on my phone
and I don't have the code in front of me so I'll have to double check to be
sure.

Il 11 gen 2017 8:30 PM, "Horsetopus" <notifications@github.com> ha scritto:

lminiero:
Just checked the code (it took a while! :smile: ) and I can confirm the call to `janus_videoroom_recorder_close` precedes the event, and the calls to `janus_recorder_close` and `janus_recorder_free` are made in the same thread. Anyway, `janus_videoroom_recorder_close` is also automatically called by `janus_videoroom_hangup_media`, which means that if the PeerConnection is hangup, then you get no explicit event of the recording being done.

lminiero:
Thinking about this a bit more: what do you think about a configuration value in `janus.cfg`? e.g., something that tells the recorder whether the temporary name should be used or not (default=false for backwards compatibility). This way, applications relying on the old mechanism would still work, and you'd have the freedom to use the feature instead should you want to.

If that works for you, I can work on a PR.

lminiero:
Went ahead and created one anyway... please let me know if that works for you.

[...]

PR #775, opened 2017-02-09
Option to add temporary extension while recording

lminiero:
@Horsetopus expressed this request in #729, as it would make it easier for him to monitor the state of recording in external scripts. Rather than forcing it for everybody, I thought I'd make this an option available in `janus.cfg`. If you set `recordings_tmp_ext` to something, e.g.:

    recordings_tmp_ext = tmp

then while recording the `.mjr` file will also have an additional `.tmp` extension. When the recording is closed, the extra extension is removed. Without that option, the file has the `.mjr` extension since the beginning.

Tested briefly and it seems to do the job. Anyway, I noticed that in some cases, if recordings are not closed normally (e.g., PeerConnection just closed in the EchoTest), the extra extensions are never removed. The more you test this and spot inconsistencies like these, the sooner we can get these sorted out and merge.


lminiero:
Merging.

PR #776, opened 2017-02-10
Permit user configurable network device selection for multicast RTP and RTSP streams

cmacq2:
Permit user configurable network device selection for listening to multicast RTP and RTSP streams in Janus. Since multicast addresses only make sense in the context of specific network interfaces, this change enables Janus to stream RTP/RTSP received on abitrary interfaces reliably to its WebRTC clients/peers.

To enable explicit configuration of network interfaces, three new configuration parameters are introduced (supported both as static configuration and as part of JSON payload in the REST API):

 * The `x-multicast-iface` option for RTSP streams.
 *  The `x-audio-multicast-iface` and `x-video-multicast-iface` options for RTP streams. (As a side effect, for RTP streams audio and video components may be bound to different network devices.)
    
The same value type/syntax is used for each option: either a device name or its IP address. Currently when IP addresses are specified, an IPv4 address should be specified because the streaming plugin itself does not yet sufficiently support IPv6 (relying, as it does, on raw IPv4 address types). As a result IPv6 addresses are not honoured yet even though the main configuration code paths are fully IPv6 compatible.


lminiero:
Whoa, that's a lot of code, thanks for the contribution! I'll need some time to review this, although it looks like something that can be used in a few other parts of the code too, and not only the Streaming plugin. I hope I'll manage to give it a look between today and tomorrow, as on Wednesday I'll go abroad for an event, and would only be able to inspect this next Monday.

In the meanwhile, have you signed our [CLA](https://cla.conf.meetecho.com/janus/)?

cmacq2:
I've signed just now.

cmacq2:
@lminiero any chance you could find the time to review this today?

lminiero:
That's the plan! Sorry if that's taking a while, I've been abroad last week.

ip-utils.c
```diff
@@ -0,0 +1,159 @@
+/*! \file    ip-utils.c
+ * \author   Johan Ouwerkerk <jm.ouwerkerk@gmail.com>
+ * \copyright GNU General Public License v3
+ * \brief    IP address related utility functions (headers)
```
lminiero: The `(headers)` part should be removed, here, this is the implementation.

ip-utils.c
```diff
@@ -0,0 +1,159 @@
...
+
+#include <errno.h>
+#include <sys/types.h>
+#include <arpa/inet.h>
+#include <net/if.h>
+#include <string.h>
+
+#include "ip-utils.h"
+
+static int compare_byte_arrays(const uint8_t * b1, const uint8_t * b2, const size_t size) {
```
lminiero: Please prefix all methods, even those that are static and only used internally, with something `janus_`. In this case, I guess `janus_iputils_` or `janus_ip_` would be the way to go? Feel free to think of something cooler :smile:

ip-utils.c
```diff
@@ -0,0 +1,159 @@
...
+			((query->mode & IPv4OnlyMatch) && iface_matches_ipv4(ifa, &query->ipv4) == 1) ||
+			((query->mode & IPv6OnlyMatch) && iface_matches_ipv6(ifa, &query->ipv6) == 1) ||
+			((query->mode & NameOnlyMatch) && iface_matches_name(ifa, query->device_name) == 1) ? 1 : 0;
+	}
+	else {
+		return -EINVAL;
+	}
+}
+
+const struct ifaddrs * janus_query_network_devices(const struct ifaddrs * ifa, const janus_network_query_config * query) {
```
lminiero: Extra space after the asterisks (per coding style, we write pointers like `type *pointer`). Please fix wherever it applies.

ip-utils.c
```diff
@@ -0,0 +1,159 @@
...
+		}
+	}
+	else {
+		return -EINVAL;
+	}
+}
+
+void janus_network_address_nullify(janus_network_address * a) {
+	if(a)
+	{
```
lminiero: Per coding style, please put the opening bracket on the same line as the `if`, in other parts of the code too if it happens.

ip-utils.h
```diff
@@ -0,0 +1,139 @@
...
+#ifndef _JANUS_IP_UTILS_H
+#define _JANUS_IP_UTILS_H
+
+#include <ifaddrs.h>
+#include <netinet/in.h>
+
+/** @name Janus helper methods to match names and addresses with network interfaces/devices.
+ */
+///@{
+typedef enum JanusNetworkQueryOptions {
```
lminiero: Coding style: we don't use upper camel case types or variable names. Please call the enum something like `janus_network_query_options`, or anyway something that reflects the common prefix you'll choose for your new code (e.g., `janus_iputils_network_query_options` or `janus_ip_network_query_options`, to quote the examples made above).

ip-utils.h
```diff
@@ -0,0 +1,139 @@
...
+	NoMatch = 0, NameOnlyMatch = 1, IPv4OnlyMatch = 2, IPv6OnlyMatch = 4, AnyIPMatch = 6,  AnyMatch = 7
```
lminiero: Same as above for enum names, which should at least in part include the enum type (e.g., `NoMatch` --> `janus_iputils_network_query_option_no_match`).

ip-utils.h
```diff
@@ -0,0 +1,139 @@
...
+typedef struct janus_network_query_config {
+	const char * device_name;
+	JanusNetworkQueryOptions mode;
+	struct in_addr ipv4;
+	struct in6_addr ipv6;
+} janus_network_query_config;
+
+/*!
+ * \brief Structure to hold network addresses in a tagged union which should be IPv4 and IPv6 compatible.
+ * Use the `family` member (either `AF_INET` or `AF_INET6`) to determine which type of address is contained.
```
lminiero: I don't think doxygen understands the reverse quote syntax for formatting as code, as that's markdown syntax. If you want those to be handled as markdown does, then you should use the `\c` directive wherever it occurs, e.g.:

\`family\` --> \c family

ip-utils.h
```diff
@@ -0,0 +1,139 @@
...
+	int family;
+	union {
+		struct in_addr ipv4;
+		struct in6_addr ipv6;
+	};
+} janus_network_address;
+
+/*!
+ * \brief Initialise a network device query.
+ * \param user_value the user-supplied string which is supposed to describe either the device name or its IP address.
```
lminiero: The description of each parameter should start with a capital letter, as that's the text doxygen automatically adds for each of them.

ip-utils.h
```diff
@@ -0,0 +1,139 @@
...
+ * \param query_mode (a mask of) options describing the supoorted types of matches which should be accepted when performing a look up with this query.
+ * This can be used to restrict the query to 'by device name' or 'IPv4 only' type searches.
+ * \param query the query object to configure.
+ * \return 0 on success or -EINVAL if any of the arguments are NULL or the given value to look for does not correspond to a valid IPv4/IPv6 address and
```
lminiero: Missing comma after "on success".

plugins/janus_streaming.c
```diff
@@ -79,11 +79,14 @@ videobufferkf = yes|no (whether the plugin should store the latest
 	keyframe and send it immediately for new viewers, EXPERIMENTAL)
 databuffermsg = yes|no (whether the plugin should store the latest
 	message and send it immediately for new viewers)
+x-audio-multicast-iface = network interface IP address or device name to listen on when receiving multicast audio
+x-video-multicast-iface = network interface IP address or device name to listen on when receiving multicast video
```
lminiero: Please use a naming that is consistent with the others, e.g., `audioiface`. Besides, any reason why this should be limited to multicast, and not apply to unicast as well? There may be several reasons to limit the interfaces a mountpoint listens on (e.g., to make sure only another process in your lan can feed a mountpoint, and not anybody on the internet who knows the port). As a side note, there should be one for data channels too, as they're also a possible property in mountpoints now.

plugins/janus_streaming.c
```diff
@@ -575,6 +584,32 @@ void *janus_streaming_watchdog(void *data) {
 	return NULL;
 }
 
+static int janus_streaming_lookup_multicast_interface(const struct ifaddrs * ifas, const char * iface, janus_network_address * result) {
+	if(result) {
```
lminiero: Please return immediately if `result` is invalid, no need to add an indentation layer just for that.

plugins/janus_streaming.c
```diff
@@ -575,6 +584,32 @@ void *janus_streaming_watchdog(void *data) {
...
+		if(iface) {
+			janus_network_query_config q;
+			if(janus_prepare_network_device_query(iface, IPv4OnlyMatch | NameOnlyMatch, &q)) {
+				JANUS_LOG(LOG_ERR, "Unable to prepare query of network devices by name and/or IPv4 address for values matching: %s\n", iface);
+				return -1;
+			}
```
lminiero: Please put the else right after the closing bracket (same in other parts of the code where it happens too).

plugins/janus_streaming.c
```diff
@@ -590,6 +625,11 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 		return -1;
 	}
 
+	struct ifaddrs * ifas = NULL;
+	if(getifaddrs(&ifas) || ifas == NULL) {
+		JANUS_LOG(LOG_ERR, "Unable to acquire list of network devices/interfaces; some configurations may not work as expected...\n");
```
lminiero: Considering this is not returning an error, maybe it should be a `LOG_WARN` instead?

plugins/janus_streaming.c
```diff
@@ -1271,6 +1364,11 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session
 			if(error_code != 0)
 				goto plugin_response;
 		}
+
+		if(getifaddrs(&ifas) || ifas == NULL) {
+			JANUS_LOG(LOG_ERR, "Unable to acquire list of network devices/interfaces; some configurations may not work as expected...\n");
```
lminiero: See above.

plugins/janus_streaming.c
```diff
@@ -1589,7 +1730,7 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session
 			mp->pin = g_strdup(json_string_value(pin));
 		if(save) {
 			/* This mountpoint is permanent: save to the configuration file too
-			 * FIXME: We should check if anything fails... */
+			* FIXME: We should check if anything fails... */
```
lminiero: Why remove the space there? It was indented with the one above.

plugins/janus_streaming.c
```diff
@@ -1741,7 +1882,7 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session
 		janus_mutex_unlock(&mp->mutex);
 		if(save) {
 			/* This change is permanent: save to the configuration file too
-			 * FIXME: We should check if anything fails... */
+			* FIXME: We should check if anything fails... */
```
lminiero: See above.

plugins/janus_streaming.c
```diff
@@ -2361,9 +2506,9 @@ static void *janus_streaming_handler(void *data) {
 			}
 		} else if(!strcasecmp(request_text, "switch")) {
 			/* This listener wants to switch to a different mountpoint
-			 * NOTE: this only works for live RTP streams as of now: you
-			 * cannot, for instance, switch from a live RTP mountpoint to
-			 * an on demand one or viceversa (TBD.) */
+			* NOTE: this only works for live RTP streams as of now: you
```
lminiero: See above (please fix everywhere it applies).

plugins/janus_streaming.c
```diff
@@ -2520,15 +2665,34 @@ static int janus_streaming_create_fd(int port, in_addr_t mcast, const char* list
...
-			struct ip_mreq mreq;
-			memset(&mreq, 0, sizeof(mreq));
+			struct ip_mreqn mreq;
+			memset(&mreq, '\0', sizeof(mreq));
 			mreq.imr_multiaddr.s_addr = mcast;
-			if(setsockopt(fd, IPPROTO_IP, IP_ADD_MEMBERSHIP, &mreq, sizeof(struct ip_mreq)) == -1) {
+			if(!janus_network_address_is_null(iface)) {
+				if(iface->family == AF_INET) {
+					mreq.imr_address = iface->ipv4;
+                    mreq.imr_ifindex = 0;
```
lminiero: This whole block uses spaces instead of tabs for indentations, please fix.

lminiero:
As a side note, has this been tested on Mac OS too? Unfortunately I don't have access to one, so can never check we don't break anything there myself.

cmacq2:
I'll rework this PR to include your feedback. 
As for OSX: that is a good point. the `struct mreqn` is Linux-only. I'll use the `imr_interface` member of `struct mreq` instead which is POSIX. In either case, there should be no regressions for existing Janus deployments as this is an opt-in feature.

plugins/janus_streaming.c
cmacq2: Probably an artifact of rebasing & resolving a merge conflict as the guts of this patch are roughly a month old now. ;)

plugins/janus_streaming.c
cmacq2: Well it means the OS is not returning any network interfaces, i.e. claiming there *are* not network interfaces or we're not able to see them for some other reason. This is one of those things that one might file under "should not ever happen".

plugins/janus_streaming.c
cmacq2: OK, I'll change the name into something more uniform with existing config keys. Currently it does not make sense in a unicast context because the code explicitly uses these interfaces to populate the structuring describing the multicast in the `IP_ADD_MEMBERSHIP` operation.

I do not know if this could possibly make (any) sense for unicast  addresses: as I understand it, unicast addresses are bound to a particular interface anyway by definition.

lminiero: Well, if you have a mountpoint bind to port 5002, and you have multiple interfaces, currently the socket will bind to them all. We should allow binding only to, e.g., eth0, and not eth1 too, for increased protection and control. This is what we already provide for the HTTP and WebSockets servers for instance.

cmacq2:
Continuing from [your comment](https://github.com/meetecho/janus-gateway/pull/776#discussion_r102058702)

> Well, if you have a mountpoint bind to port 5002, and you have multiple interfaces, currently the socket will bind to them all. We should allow binding only to, e.g., eth0, and not eth1 too, for increased protection and control. This is what we already provide for the HTTP and WebSockets servers for instance.

I think I finally understand what you are getting at. With the latest update to the PR, unicast streams may also be restricted to a specific IP (using this configured IP as the IP of the network address to bind to).

--------------------

No data channels yet.

plugins/janus_streaming.c
cmacq2: Well it means the OS is not returning any network interfaces, i.e. claiming there *are* no network interfaces or we're not able to see them for some other reason. This is one of those things that one might file under "should not ever happen".

cmacq2:
Introduced the data channel variant as the `datasctpnetwork` option.

cmacq2:
@lminiero the only remaining issue that I see is the logging level (`ERR` vs `WARN`). Apart from that I think the new state of the PR could do with an accompanying update to the initial review to match, so we know whether the PR is basically OK or needs more work.

(Thing is: with some of these source files being so large, it's kind of hard to ensure everything matches the code style -- doubly so because I cannot find documentation outlining it or test scripts to check for it.)

lminiero:
@cmacq2 tnx for the quick changes! Yep, this definitely gives me something I can now test, and we can worry about the nits later. I'm a bit swamped with things to do right now, but I should be able to give it a go tomorrow: please don't hesitate to remind me in case I forget (which happens all the time...)

cmacq2:
@lminiero OK, thanks! Looking forward to the test results.

cmacq2:
@lminiero since I'm free to remind you... ;) Did you manage to find the time for testing yet?

lminiero:
@cmacq2 not yet, sorry, but I still have a few hours ahead of me :wink:

lminiero:
Seems to crash with me with a: `./janus: symbol lookup error: /home/lminiero/test/ip-utils/lib/janus/plugins/libjanus_streaming.so: undefined symbol: janus_network_address_is_null`

I tried modifying the sample mountpoint by adding:

    audiortpnetwork = eno1
    videortpnetwork = eno1

to my config (eno1 is my ethernet adapter). I verified that setting a non-existing interface results in a failure to create the mountpoint, which is good.

lminiero:
Added an `extern` before the inline methods in `ip-utils.h` and it went through. More testing to follow...

lminiero:
You probably want to reverse the order of the IP fields you print here: https://github.com/meetecho/janus-gateway/pull/776/files#diff-b6451ee93281bd18cf90a556862c894eR2739

    [gstreamer-sample] Audio listener restricted to interface address: 80.1.168.192
    [gstreamer-sample] Video listener restricted to interface address: 80.1.168.192

(my local IP address is 192.168.1.80)

plugins/janus_streaming.c
```diff
@@ -575,6 +587,29 @@ void *janus_streaming_watchdog(void *data) {
...
+		return -1;
+	} else if(iface) {
+		janus_network_query_config q;
+		if(janus_prepare_network_device_query(iface, janus_ip_network_query_options_ipv4 | janus_ip_network_query_options_name, &q)) {
+			JANUS_LOG(LOG_ERR, "Unable to prepare query of network devices by name and/or IPv4 address for values matching: %s\n", iface);
+			return -1;
+		} else {
+			const struct ifaddrs *found = janus_query_network_devices(ifas, &q);
+			if(!found || janus_get_network_device_address(found, result)) {
+				JANUS_LOG(LOG_INFO, "Unable to retrieve address from matching network device for: %s\n", iface);
```
lminiero: I guess this should be a `LOG_ERR`?

lminiero:
The `extern` stuff may not be the best solution, as this way I can't get the code to compile if I use the ip-utils from the core (e.g., in janus.c), something that might be useful as we do have parts of the code (e.g., ICE related) where we parse interfaces and things like that. Any suggestion on how to fix this?

Apart from this (and a couple of naming related nits I'll bother you with later :smile: ) this looks fine to me, great work! I'll probably add some wrappers after merging to make it even simpler to use: e.g., `janus_streaming_lookup_network_interface` could actually be something provided by ip-utils directly, etc.

cmacq2:
About the crash: did you do a `make clean` before `make && make install`? 

Unless I am much mistaken the `extern` bit should not be needed, precisely because the core is the one doing the loading (therefore the symbol should be visible even before the streaming plugin is loaded). Or I may be misunderstanding something more profound here and barking up the wrong tree.

cmacq2:
Good catch about the order of the IP fields.It probably should just use `inet_ntop()` instead anyway.

lminiero:
I was starting fresh, so no need to make clean there. Without the extern, the symbol is apparently not exported to the module, at least with my gcc.

cmacq2:
What is your gcc version?

lminiero:
gcc (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1)

plugins/janus_streaming.c
cmacq2: This is not that serious: it may be caused by user error (misconfigured IP or device name).
Hence `LOG_INFO`; alternatively: perhaps `LOG_WARN` instead? What do you think?

lminiero: In this case this would result in no address being returned and so a failure to create the mountpoint, so I think LOG_ERR is more appropriate. But nevermind that, it's things I can change later on (also because as I mentioned I plan to implement some helper methods in the ip-utils anyway).

cmacq2:
I used `gcc (Debian 6.3.0-6) 6.3.0 20170205`

cmacq2:
I cannot reproduce the crash:
```
    mkdir -p "$HOME/projects/janus-import/dist"
    cd "$HOME/projects/janus-import"
    git clone git@github.com:cmacq2/janus-gateway.git
    cd janus-gateway
    git checkout multicast-multiple-nics
    ./autogen.sh
    ./configure --prefix="$HOME/projects/janus-import/dist"
    make
    make install
    make configs
    cd "$HOME/projects/janus-import/dist"
    bin/janus
```

Output of starting Janus:

> ````
> ---------------------------------------------------                                                                                                                                                                                                                            
>  Starting Meetecho Janus (WebRTC Gateway) v0.2.2                                                                                                                                                                                                                              
> ---------------------------------------------------                                                                                                                                                                                                                            
>                                                                                                                                                                                                                                                                               
> Checking command line arguments...
> Debug/log level is 4
> Debug/log timestamps are disabled
> Debug/log colors are enabled
> Adding 'vmnet' to the ICE ignore list...
> Using 192.168.178.101 as local IP...
> [WARN] Token based authentication disabled
> Initializing ICE stuff (Full mode, ICE-TCP candidates disabled, IPv6 support disabled)
> ICE handles watchdog started
> TURN REST API backend: (disabled)
> [WARN] Janus is deployed on a private address (192.168.178.101) but you didn't specify any STUN server! Expect trouble if this is supposed to work over the internet and not just in a LAN...
> BUNDLE is NOT going to be forced
> rtcp-mux is NOT going to be forced
> [WARN] OpenSSL >= 1.1.0
> Fingerprint of our certificate: D2:B9:31:8F:DF:24:D8:0E:ED:D2:EF:25:9E:AF:6F:B8:34:AE:53:9C:E6:F3:8F:F2:64:15:FA:E8:7F:53:2D:38
> [WARN] Data Channels support not compiled
> Event handler plugins folder: /home/prometheus/projects/janus-import/dist/lib/janus/events
> [WARN] Event handlers support disabled
> Sessions watchdog started
> Plugins folder: /home/prometheus/projects/janus-import/dist/lib/janus/plugins
> Loading plugin 'libjanus_videoroom.so'...
> JANUS VideoRoom plugin initialized!
> VideoRoom watchdog started
> Loading plugin 'libjanus_audiobridge.so'...
> JANUS AudioBridge plugin initialized!
> AudioBridge watchdog started
> Loading plugin 'libjanus_videocall.so'...
> JANUS VideoCall plugin initialized!
> Loading plugin 'libjanus_recordplay.so'...
> VideoCall watchdog started
> JANUS Record&Play plugin initialized!
> Loading plugin 'libjanus_streaming.so'...
> Record&Play watchdog started
> JANUS Streaming plugin initialized!
> Loading plugin 'libjanus_echotest.so'...
> Streaming watchdog started
> JANUS EchoTest plugin initialized!
> EchoTest watchdog started
> Loading plugin 'libjanus_textroom.so'...
> JANUS TextRoom plugin initialized!
> Loading plugin 'libjanus_voicemail.so'...
> JANUS VoiceMail plugin initialized!
> VoiceMail watchdog started
> Loading plugin 'libjanus_sip.so'...
> TextRoom watchdog started
> JANUS SIP plugin initialized!
> SIP watchdog started
> Transport plugins folder: /home/prometheus/projects/janus-import/dist/lib/janus/transports
> Loading transport plugin 'libjanus_http.so'...
> HTTP webserver started (port 8088, /janus path listener)...
> [WARN] HTTPS webserver disabled
> [WARN] Admin/monitor HTTP webserver disabled
> [WARN] Admin/monitor HTTPS webserver disabled
> JANUS REST (HTTP/HTTPS) transport plugin initialized!
> Loading transport plugin 'libjanus_pfunix.so'...
> HTTP/Janus sessions watchdog started
> Configuring SOCK_SEQPACKET Unix Sockets server (Janus API)
> [WARN] Unix Sockets server disabled (Admin API)
> JANUS Unix Sockets transport plugin initialized!
> Loading transport plugin 'libjanus_websockets.so'...
> Unix Sockets thread started
> libwebsockets >= 1.6 available, using new API
> WebSockets server started (port 8188)...
> [WARN] Secure WebSockets server disabled
> [WARN] Admin WebSockets server disabled
> [WARN] Secure Admin WebSockets server disabled
> JANUS WebSockets transport plugin initialized!
> WebSocket (Janus API) thread started
````

**Comment by cmacq2** [comment]
Rebased the PR to latest master.

**Comment by cmacq2** [comment]
I've revisited `ip-utils.{h,c}` a bit and updated the streaming plugin to match:

- Remove `inline` from functions, move definitions to `ip-utils.c`. This is a blunt instrument type of attempt to fix your crash. But it's not like there's any particular reason why these functions must be `inline` so there seems to be no downside. Please retest & let me know if this resolves the issue?
- Added additional utilities to convert `janus_network_address` structs to a plain C-string via a intermediary 'buffer' type format that is IPv4 & IPv6 compatible/aware called `janus_network_address_string_buffer`. The actual conversion to strings is done by the trusty `inet_ntop()` function call. (I.e the proper way to do it.)

**Comment by lminiero** [comment]
Thanks, will try ASAP and come back to you!

**Comment by lminiero** [comment]
Seems to work fine, thanks! Anyway, there's no checks in some places where there should be, e.g., whenever we access `ifaddrs` properties like `ifa->ifa_addr->sa_family`. I'm mentioning this as, when trying to play with your methods, this crashed Janus:

```
	struct ifaddrs ifas;
	janus_network_query_config q;
	janus_prepare_network_device_query(iface, janus_ip_network_query_options_ipv4 | janus_ip_network_query_options_name, &q));
	const struct ifaddrs *found = janus_query_network_devices(&ifas, &q);
```

The cause is that, while `janus_query_network_devices` and `janus_ip_iface_matches` do check whether `ifa` is a valid pointer (which it is, in this case), `janus_ip_iface_matches_name` assumes `ifa->ifa_addr` is a valid pointer too, and also does a `strncmp` on `ifa->ifa_name` assuming that exists too.

Considering developers are undisciplined (me included!) and can pass things that may be invalid, I'd check for the validity of those pointers before using them in the code.

cmacq2:
Sure I can add additional defensive checks. One thing worth mentioning though is that you pass in an uninitialised `struct`, so even with the checks the particular pointer values could happen to look legit and lead to who knows what. So it won't be an absolue defense.

cmacq2:
Added additional checks.

lminiero:
> One thing worth mentioning though is that you pass in an uninitialised struct, so even with the checks the particular pointer values could happen to look legit and lead to who knows what.

Yep, that's true! Thanks for the additional checks though, they never hurt :+1: 
I'll give another quick round later and plan to merge soon thereafter!

cmacq2:
Rebased to latest master. @lminiero do you think you can find some time to test (and hopefully merge) this today?

lminiero:
Will do, sorry again if this is taking a while but it has been a very busy week! I'll check right now and then merge in the morning.

lminiero:
Looks, good, thanks, merging! :+1:

lminiero:
Just as a note, I'll be making a small series of changes, just for "editorial" reasons, e.g.:

    audiortpnetwork --> audioiface
    videortpnetwork --> videoiface
    datasctpnetwork --> dataiface

I'll try to also integrate your IP utils in some parts of the core (and other plugins) where we try to do similar things.

Issue #778, opened 2017-02-13
please include source for code not authored here


jonassmedegaard:
Hi,

Some code is included from external projects, but has been included here in a non-source form.

A simple example is html/jquery.min.js which has been "minified": It is technically possible but not easy to edit that code.

Another less obvious example is html/js/bootstrap.js which is not minified but still concatenated together, which is not the format upstream uses for development: It is easy to edit that code, but such changes are not easy to pass upstream.

Some distributors - Debian in particular - insist on redistributing as "source" only code which the authors consider _their_ source.

Please - if not for easing development without fragmentation, then for ease of redistribution - include _upstream_ source for all included code not authored here.

lminiero:
Thanks for the feedback.

The web demos are there only for ease of playing with the server: the "actual" Janus is the C core and its plugins, nothing else. That's why the JS files are the way they are: the first time I worked on some demos, I used the libraries I already had available at the time in some test setups and was more familiar with (bootstrap and jquery, mostly, as I'm a lazy JS developer!) and they stayed pretty much the time since. I can't remember where I got most of them from, but I think jquery.min.js came from their site (from some time ago, obviously), while bootstrap.js is the version shipped, I believe, with the [template](https://bootswatch.com/cerulean/) I used (not sure if they did something different with the sources).

Not exactly sure on what I should do to fix this: does upstream in this context mean updating, e.g., jquery in this repo everytime it's updated itself? I personally don't have the energy nor the time to follow up on everything we use in the demos (which don't need anything but the basic stuff anyway). Or did you mean something different?

PRs to fix the current state would be very welcome of course.

jonassmedegaard:
Quoting Lorenzo Miniero (2017-02-13 14:33:09)
> Thanks for the feedback.
> 
> The web demos are there only for ease of playing with the server: the 
> "actual" Janus is the C core and its plugins, nothing else.

Understood.  And those demos are highly appreciated!


> That's why the JS files are the way they are: the first time I worked 
> on some demos, I used the libraries I already had available at the 
> time in some test setups and was more familiar with (bootstrap and 
> jquery, mostly, as I'm a lazy JS developer!) and they stayed pretty 
> much the time since. I can't remember where I got most of them from, 
> but I think jquery.min.js came from their site (from some time ago, 
> obviously), while bootstrap.js is the version shipped, I believe, with 
> the [template](https://bootswatch.com/cerulean/) I used (not sure if 
> they did something different with the sources).

[...]

PR #780, opened 2017-02-13
Reference third party js/css files externally (see #778)

lminiero:
All external css and js files (apart from adater.js) now point to files on https://cdnjs.com
Please test that it doesn't break any demo.


jonassmedegaard:
Sorry, but this PR does not really address issue #781.

The copyright and license which I believe you should add at the top of the file is this: https://github.com/webrtc/adapter/blob/master/LICENSE.md

What you did was repeat a phrase which I quoted from _your_ copy of adapter.js: https://github.com/meetecho/janus-gateway/blob/master/html/adapter.js#L495 - a phrase _referencing the missing copyright and licensing.

I apologize if my way of presenting the issue was confusing.

lminiero:
I copied the header from here: https://github.com/webrtc/adapter/blob/master/src/js/adapter_core.js#L1
I read only know it refers to the external license file at the root, which should then be copied inline. I'll take care of this next week, I have a full day tomorrow working on other things tomorrow, and then I'll be abroad for an event.

jonassmedegaard:
Quoting Lorenzo Miniero (2017-02-13 20:48:50)
> I copied the header from here: https://github.com/webrtc/adapter/blob/master/src/js/adapter_core.js#L1
> I read only know it refers to the external license file at the root, which should then be copied inline. I'll take care of this next week, I have a full day tomorrow working on other things tomorrow, and then I'll be abroad for an event.

Makes sense.

Enjoy your journey - and the other stuff :-)

 - Jonas

-- 
 * Jonas Smedegaard - idealist & Internet-arkitekt
 * Tlf.: +45 40843136  Website: http://dr.jones.dk/

 [x] quote me freely  [ ] ask before reusing  [ ] keep private

jonassmedegaard:
Quoting Jonas Smedegaard (2017-02-13 21:31:41)
> Quoting Lorenzo Miniero (2017-02-13 20:48:50)
>> I copied the header from here: https://github.com/webrtc/adapter/blob/master/src/js/adapter_core.js#L1
> Makes sense.

For the record:, above part is not sensible, I suspect - unless what you 
forked was _only_ that adapter_core.js file.

I guess you forked the product of concatenating multiple files 
_including_ adapter_core.js (or however things were organized when you 
forked it), and then cherry-picking a header from one of their source 
files makes little sense.

...but the other part sounds sensible, regardless.


 - Jonas

-- 
 * Jonas Smedegaard - idealist & Internet-arkitekt
 * Tlf.: +45 40843136  Website: http://dr.jones.dk/

 [x] quote me freely  [ ] ask before reusing  [ ] keep private

thehunmonkgroup:
on a very related note, would you accept a pull request to UMD-ify janus.js and janus.nojquery.js?

like this, so it can be used easily across different environments: https://github.com/umdjs/umd/blob/master/templates/commonjsStrictGlobal.js

lminiero:
I think this was asked before, or something similar. I'd rather not go that road. I'm not at all familiar with that stuff, and I have to maintain and keep everything up-to-date, something I'd rather avoid. Demos (and janus.js, for that matter) are supposed to be a simple example of how you can use Janus, not the de-facto standard. Nolan Darilek's and Cargomedia's JavaScript stacks, both availabile in the [resources](https://janus.conf.meetecho.com/docs/resources) documentation page, do follow that approach instead, so they might be a better option than janus.js if this is something you care about.

jonassmedegaard:
Quoting Lorenzo Miniero (2017-02-18 11:24:44)
> I think this was asked before, or something similar. I'd rather not go that road. I'm not at all familiar with that stuff, and I have to maintain and keep everything up-to-date, something I'd rather avoid. Demos (and janus.js, for that matter) are supposed to be a simple example of how you can use Janus, not the de-facto standard. Nolan Darilek's and Cargomedia's JavaScript stacks, both availabile in the [resources](https://janus.conf.meetecho.com/docs/resources) documentation page, do follow that approach instead, so they might be a better option than janus.js if this is something you care about.

Please note that Cargomedia has apparently given up: 
https://github.com/cargomedia/janus-gateway-js/pull/102

 - Jonas

-- 
 * Jonas Smedegaard - idealist & Internet-arkitekt
 * Tlf.: +45 40843136  Website: http://dr.jones.dk/

 [x] quote me freely  [ ] ask before reusing  [ ] keep private

lminiero:
True, but they clarified they're willing to let somebody maintain it, so it doesn't mean the project is dead: https://groups.google.com/d/msg/meetecho-janus/bSMyGgLeMiE/SCsF4f3XCQAJ

thehunmonkgroup:
@lminiero I've got janus.nojquery.js working fine for me, so I can certainly let go the idea of making it UMD compatible.

I will add that I don't think there's really much to maintain with a UMD definition -- it's a very simple wrapper that makes things easier for consumers, and unless some drastic new way of loading javascript emerges, it won't change.

lminiero:
In that case I guess a PR won't harm, if not to see what that changes that would entail.

thehunmonkgroup:
Ok, cool. I'll work one up soon

thehunmonkgroup:
For interested parties, #792 is a draft of the UMD for janus[.nojquery].js

lminiero:
Any feedback on this PR?

docs/header.html
```diff
@@ -7,12 +7,12 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
 <!--BEGIN PROJECT_NAME--><title>$title</title><!--END PROJECT_NAME-->
 <!--BEGIN !PROJECT_NAME--><title>$title</title><!--END !PROJECT_NAME-->
-<script type="text/javascript" src="jquery.js"></script>
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/1.7.2/jquery.min.js"></script>
```
thehunmonkgroup: Maybe a more modern version of jQuery? They're on to 3.x now...

lminiero: As long as it doesn't break any of the demos or `janus.js` functionality, why not... I only specified that as that's the version we're currently shipping and the one I used myself.

lminiero:
Looks nobody will test this before I merge it... going the hard way then :wink:

PR #784, opened 2017-02-14
Several new helper methods for SDP utilities

lminiero:
This PR adds a whole lot of new helper methods to the SDP utilities I wrote some time ago. While working on some test plugins, in fact, I noticed how hard it could be to quickly prepare an answer to an offer, for instance, or even to create an offer from scratch without plain string manipulation, all things easily prone to error.

While the PR adds a lot of helpers to quickly find payload types, codec names, m-lines and the like, the main addition are two methods to generate an answer to an offer, and to generate an offer, both as `janus_sdp` objects. In order to drive how the offer or the answer should look like, these methods have variable arguments, and we have defines in place to make this easier for the caller. Here's an example of how you can create a new offer now:

	janus_sdp *offer = janus_sdp_generate_offer("My session", "127.0.0.1",
		JANUS_SDP_OA_AUDIO, TRUE,
		JANUS_SDP_OA_AUDIO_PT, 100,
		JANUS_SDP_OA_AUDIO_DIRECTION, JANUS_SDP_SENDONLY,
		JANUS_SDP_OA_AUDIO_CODEC, "opus",
		JANUS_SDP_OA_VIDEO, FALSE,
		JANUS_SDP_OA_DATA, FALSE,
		JANUS_SDP_OA_DONE);

This creates an offer with only audio as media: audio must be Opus, sendonly and use payload type 100. Video and datachannels are not added to the offer instead. As you can see, it's just a matter of adding a bunch of name/value settings, concluded by a `JANUS_SDP_OA_DONE`.

Generating an answer to a received offer is pretty much similar:

	janus_sdp *answer = janus_sdp_generate_answer(offer,
		JANUS_SDP_OA_AUDIO, TRUE,
		JANUS_SDP_OA_AUDIO_DIRECTION, JANUS_SDP_RECVONLY,
		JANUS_SDP_OA_AUDIO_CODEC, "opus",
		JANUS_SDP_OA_VIDEO, FALSE,
		JANUS_SDP_OA_DATA, FALSE,
		JANUS_SDP_OA_DONE);

Here we're saying the offer should only accept audio, and only accept opus as a negotiated codec, and to try and only receive media; depending on the direction set by the media, this might result in a different attribute in the SDP (e.g., if you ask for sendrecv, but the offer proposed sendonly, you'll end up with recvonly). We're also saying that, if present, video and data must be rejected. Payload types obviously are not provided here, as we'll use whatever the offerer proposed.

You can find a list of all the properties you can set in the `sdp-utils.h` header. By default (that is, if you terminate immediately with a `JANUS_SDP_OA_DONE`) the code generates/accepts audio, video and data, tries to do sendrecv for everything, negotiates Opus and VP8 as codecs (the most common choices), and adds some common attributes.

I've already integrated this new approach in the EchoTest, AudioBridge and VideoRoom plugins and, as you can see, it greatly simplified the way negotiations are done done there (especially in the VideoRoom plugin, that had a complex SDP template mechanism in place to account for different room codecs).

Since this is quite a substantial change, please test extensively and let me know if you spot issues, especially in how the modified plugins behave right now. If you're writing your own plugin, also try using these new helpers and let me know if they're easy enough to use for you, and if you find them flexible enough for your SDP processing needs.

Feedback welcome!


amnonbb:
How can be set audio bitrate in videroom in this PR?
Default bitrate of audio in videroom is very low (about 40k). I was solve this by puting in
[sdp_a_template_opus](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_videoroom.c#L550) 
 "b=AS:96\r\n"
Maybe it's not good way to do this, but it worked :alien:

lminiero:
The same way the VideoRoom now sets the video bitrate:

	janus_sdp_mline *m = janus_sdp_mline_find(answer, JANUS_SDP_AUDIO);
	if(m != NULL) {
		m->b_name = g_strdup("AS");
		m->b_value = 96;
	}

I may add some helper methods to quickly set/get the bitrate too, but this seems compact enough already.

lminiero:
This has been around enough, I tested this a bit and I think I got rid of all leaks. Besides, it made things much easier for me in new plugins, so it's nice to have. Merging.

PR #785, opened 2017-02-14
fix adapter loaded logic to properly report errors

thehunmonkgroup:
the callback for the scenario where adapter.js is already loaded eventually
executes the Janus.init callback -- if any errors are thrown the load sequence
should abort and report the error, instead of trying to load another script.

### DIFF
--- a/html/janus.nojquery.js
+++ b/html/janus.nojquery.js
@@ -252,7 +252,10 @@ Janus.init = function(options) {
 					done();
 					return;
 				}
-			} catch(e) {};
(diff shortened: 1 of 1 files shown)


lminiero:
Not sure this will be needed, as I'm externalizing all the dependencies in #780, and that will apply to adapter.js as well. This means janus.js will assume both jquery and adapter have already been loaded by the web page, and possibly only check if they're indeed available or not.

thehunmonkgroup:
yep that would make this obsolete.

PR #786, opened 2017-02-18
Allow updating display value via configure command

thehunmonkgroup:
This adds functionality to the audiobridge and videoroom plugins to allow updating the 'display' value that is optionally passed on registration, via the handle's 'configure' message

I hope you'll go easy on me, I'm not the strongest C programmer, and this is my first attempt to accomplish anything of substance in the Janus codebase :) I'm not clear if I took the best approach, and haven't added any doc yet, pending your feedback and possible adjustments.

I will say that I tested both of them in my local install, and they appear to work perfectly. Sending this configure command:
```javascript
      const configure = {
        "request": "configure",
        "display": "foo",
      };
      handle.send({
        "message": configure,
      });
```
Yields this response for the audiobridge:
```json
{
	"audiobridge": "event",
	"room": 3604791866433732600,
	"participants": [{
		"id": 2677844740813301,
		"display": "foo",
		"muted": false
	}]
}
```
And this for the videoroom:
```json
{
 	"videoroom": "event",
 	"id": "4708612212021077",
 	"display": "foo"
}
```

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -654,7 +654,8 @@ static struct janus_json_parameter configure_parameters[] = {
 	{"quality", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"volume", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"record", JANUS_JSON_BOOL, 0},
-	{"filename", JSON_STRING, 0}
+	{"filename", JSON_STRING, 0},
+	{"display", JSON_STRING, 0}
 };
 static struct janus_json_parameter rtp_forward_parameters[] = {
 	{"room", JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE},
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -2771,8 +2771,8 @@ static void *janus_audiobridge_handler(void *data) {
 					}
 				}
 				if (display) {
-					const char *display_text = json_string_value(display);
-					participant->display = g_strdup(display_text);
+				  g_free(participant->display);
+					participant->display = g_strdup(json_string_value(display));
 					JANUS_LOG(LOG_VERB, "Setting display property: %s (room %"SCNu64", user %"SCNu64")\n", participant->display, participant->room->room_id, participant->user_id);
 				}
 				/* Notify all other participants about the mute/unmute */
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -2731,6 +2731,7 @@ static void *janus_audiobridge_handler(void *data) {
 			json_t *gain = json_object_get(root, "volume");
 			json_t *record = json_object_get(root, "record");
 			json_t *recfile = json_object_get(root, "filename");
+			json_t *display = json_object_get(root, "display");
 			if(gain)
 				participant->volume_gain = json_integer_value(gain);
 			if(quality) {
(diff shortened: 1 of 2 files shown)


plugins/janus_audiobridge.c
```diff
@@ -2731,6 +2731,7 @@ static void *janus_audiobridge_handler(void *data) {
 			json_t *gain = json_object_get(root, "volume");
 			json_t *record = json_object_get(root, "record");
 			json_t *recfile = json_object_get(root, "filename");
+			json_t *display = json_object_get(root, "display");
```
lminiero: You need to add display to `configure_parameters` as well, for automatic JSON validation.

plugins/janus_videoroom.c
```diff
@@ -3282,6 +3282,7 @@ static void *janus_videoroom_handler(void *data) {
 				json_t *bitrate = json_object_get(root, "bitrate");
 				json_t *record = json_object_get(root, "record");
 				json_t *recfile = json_object_get(root, "filename");
+				json_t *display = json_object_get(root, "display");
```
lminiero: See above, in this plugin you have to update the `publish_parameters`.

plugins/janus_videoroom.c
```diff
@@ -3366,6 +3367,22 @@ static void *janus_videoroom_handler(void *data) {
...
+				if(display) {
+					const char *display_text = json_string_value(display);
+					participant->display = g_strdup(display_text);
+					janus_mutex_lock(&participant->room->participants_mutex);
+					json_t *display_event = json_object();
+					json_object_set_new(display_event, "videoroom", json_string("event"));
+					json_object_set_new(display_event, "id", json_integer(participant->user_id));
+					json_object_set_new(display_event, "display", json_string(participant->display));
+					if(participant->room) {
+						if(!participant->room->destroyed) {
```
lminiero: This two ifs can be put together with an  `&&`.

thehunmonkgroup:
ok, fixed up the stuff from your feedback, hopefully this is about ready to go!

lminiero:
Thanks! I'll give it a quick check myself and merge later.

plugins/janus_audiobridge.c
```diff
@@ -2745,26 +2747,33 @@ static void *janus_audiobridge_handler(void *data) {
...
+							g_free(pkt);
+							pkt = NULL;
+						}
+						janus_mutex_unlock(&participant->qmutex);
 					}
-					janus_mutex_unlock(&participant->qmutex);
+				}
+				if (display) {
+					const char *display_text = json_string_value(display);
+					participant->display = g_strdup(display_text);
```
lminiero: Just noticed that you're not freeing the old display: this will result in a leak.

plugins/janus_videoroom.c
```diff
@@ -3386,6 +3388,20 @@ static void *janus_videoroom_handler(void *data) {
 					}
 				}
 				janus_mutex_unlock(&participant->rec_mutex);
+				if(display) {
+					const char *display_text = json_string_value(display);
+					participant->display = g_strdup(display_text);
```
lminiero: Same here.

thehunmonkgroup:
Yeah this is where my lack of C programming skills shows :)

I think I've fixed it with the latest commit, let me know if you need anything else.

lminiero:
> I think I've fixed it with the latest commit, let me know if you need anything else.

Indentation looks a bit weird, spaces instead of tabs? That said, I'll have to check if this can cause race conditions, e.g., what happens if somebody joins exactly while you're changing the display name, meaning the old display name might be used for that notification while it's being freed, and possibly cause havoc. I don't remember if we use mutexes for that. Maybe one way to overcome that is first allocating the new string in a temporary variable, store the old one in another, and then set the actual display to the new value, e.g.,

    char *old_display = participant->display;
    char *new_display = g_strdup(display_text);
    participant->display = new_display;
    g_free(old_display);

This may not really be atomic, but should at least ensure a bit of protection against that happening.

thehunmonkgroup:
*sighs*...yes, probably spaces instead of tabs -- i'll try to clean that up in my next round -- almost all my coding workflows use spaces, so vim does that for me automatically.

do you want me to make the change you suggested for both plugins? i'm not clear from your comments.

lminiero:
Yeah, this should happen in both plugins, considering the display property may be accessed at any time for broadcasting participants in both of them.

lminiero:
Anyway, that's something I can do myself, as I have to update the audiobridge file anyway to update the example we have there. For the time being, merging, so thanks for the feature!

PR #787, opened 2017-02-20
add custom libnice install instructions to README

thehunmonkgroup:
Newer versions of libnice are more stable, this additional documentation
covers the basics of a custom compile of libnice that will work with
Janus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,15 @@ On Ubuntu or Debian, it would require something like this:
 on Ubuntu or Debian, unless you're using a recent version (e.g., Ubuntu
 14.04 LTS). In that case, you'll have to [install it manually](http://www.opus-codec.org).
 
+* *Note:* distribution versions of libnice may not be up to date with
+the latest releases, and older releases of the library are not as stable.
+You can [install it manually](https://nice.freedesktop.org/wiki). For
+custom installations of libnice, you can run
+```pkg-config --cflags --libs nice``` to make sure Janus can find the
+installation. If not, you may need to set the ```PKG_CONFIG_PATH```
+environment variable prior to compiling Janus, eg.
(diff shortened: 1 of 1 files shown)


lminiero:
Wouldn't `export PKG_CONFIG_PATH=/path/to/libnice/lib/pkgconfig` break `pkg-config` looking for other stuff? We don't use it for libnice only.

thehunmonkgroup:
From ```man pkg-config```

```
pkg-config retrieves information about packages from special metadata files.
These files are named after the package, and has a .pc extension.  On most
systems, pkg-config looks in /usr/lib/pkgconfig,  /usr/share/pkg‐config,
/usr/local/lib/pkgconfig and /usr/local/share/pkgconfig for these files.  It will
additionally look in the colon-separated (on Windows, semicolon-separated)
list  of  directories  specified  by  the  PKG_CONFIG_PATH environment variable.
```

IOW, setting PKG_CONFIG_PATH augments, not overwrites, the locations pkg-config searches for files. For people doing more advanced build scenarios that are already leveraging PKG_CONFIG_PATH, I believe they would already know how to handle that env variable properly with respect to a Janus compile, and as such I think the documentation as written is effective.

lminiero:
Got it, thanks for the clarification. That said, I'm not sure I'm ok with the updated text in the README, as it may encourage people to just install master that, as you found out, won't work. I'd just modify the text to say that, if libnice was not installed in a standard location, a `PKG_CONFIG_PATH` variable must be provided so that the correct one can be used instead.

thehunmonkgroup:
Sure, that's reasonable. I cleaned it up to remove anything that looks like a recommendation to install a newer version.

lminiero:
Much better, thanks! Merging :+1:

PR #789, opened 2017-02-21
enhancements to BoringSSL handling in autoconf

thehunmonkgroup:
- The BoringSSL header check didn't work at all, it just looked
  for OpenSSL headers in their standard install location. Now the
  check specifically looks in the include directory of the
  BoringSSL installation directory.

- Allows a custom location to be passed for the installation, eg.
  --enable-boringssl=/path/to/boringssl

- Document custom location feature in README

I'm far from an autoconf expert -- these changes are based on the autoconf documentation I read and some examples I found online. I tested all cases (disable/enable/custom install location), and they work correctly, including properly erroring out the configure script if BoringSSL can't be found. Even if it's not perfect, it's certainly an improvement over what's there now :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ to manually install it to a specific location. Use the following steps:
 Once the library is installed, you'll have to pass an additional
 ```--enable-boringssl``` flag to the configure script, as by default
 Janus will be built assuming OpenSSL will be used. By default, Janus
-expects BoringSSL to be installed in ```/opt/local``` -- if it's
+expects BoringSSL to be installed in ```/opt/boringssl``` -- if it's
 installed in another location, pass the path to the configure script
 as such: ```--enable-boringssl=/path/to/boringssl``` If you were using
 OpenSSL and want to switch to BoringSSL, make sure you also do a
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks, I'll give it a try later. There's a typo in the README, though, as you say the default location in `/opt/local` while it's `/opt/boringssl`

thehunmonkgroup:
typo corrected.

lminiero:
Looks good, thanks! I'll give it a quick try later and in case merge :+1:

lminiero:
Notice I slightly changed the BoringSSL related check to have it create a define when BoringSSL is detected. This way we can print on the console about whether Janus is using BoringSSL or not. Tiny little change so handling the conflict in this PR should be trivial: https://github.com/meetecho/janus-gateway/commit/1cab80cd3b673192617283d8df0058131b6ee681

thehunmonkgroup:
Yep, easy fix indeed

lminiero:
Thanks. Since I assume you tested this with custom locations, let me try if the default behaviour still works and in case I'll merge :+1:

lminiero:
Works nicely, thanks, merging!

thehunmonkgroup:
oh yes, definitely checked with custom location, that's what i'm using :)

PR #792, opened 2017-02-23
wrap janus[.nojquery].js in a Universal Module definition

thehunmonkgroup:
This is a draft PR to illustrate wrapping the janus[.nojquery].js library in a Univeral Module Definition: https://github.com/umdjs/umd

There are various definition patterns, I chose what I believe is the most common: allowing users to consume the library as an AMD module, a Node.js-like module, or the old school global.

Although there is a high line count to this patch, the changes are actually quite simple:

1. Make the Janus object local to the module definition (this involved moving all the object properties hung off the Janus function object to after its declaration)
2. Wrap that in a self executing function that allows jQuery to be used locally as either $ or jQuery (this could be eliminated if one or the other was used exclusively)
3. Wrap all that in the UMD, which is basically just a few lines of environment detection code inside a self-executing function, and should hardly ever be expected to change (only if a new loading environment became popular).

The way adapter.js is currently loaded is weird, and doesn't work well with the scoping, so I included it specifically in the videoroom demo HTML for example purposes -- I believe this weirdness is already being addressed in the web-refs branch.

Taking this approach will ease consumption of this library for more modern tooling, ala Webpack/Browserify, without compromising it's traditional use case.

If this draft is acceptable, I'll wait until the web-refs branch is merged, clean it up, and also convert janus.nojquery.js, which will be straightforward.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -22,186 +22,39 @@
 	OTHER DEALINGS IN THE SOFTWARE.
  */
 
-// List of sessions
-Janus.sessions = {};
-
-// Screensharing Chrome Extension ID
-Janus.extensionId = "hapfgfdkleiggjjpfpenajgdnfckjpaj";
-Janus.isExtensionEnabled = function() {
-	if(window.navigator.userAgent.match('Chrome')) {
-		var chromever = parseInt(window.navigator.userAgent.match(/Chrome\/(.*) /)[1], 10);
-		var maxver = 33;
-		if(window.navigator.userAgent.match('Linux'))
-			maxver = 35;	// "known" crash in chrome 34 and 35 on linux
-		if(chromever >= 26 && chromever <= maxver) {
-			// Older versions of Chrome don't support this extension-based approach, so lie
-			return true;
-		}
-		return ($('#janus-extension-installed').length > 0);
+/**
+ * UMD wrapper.
+ */
+(function (root, factory) {
+	if (typeof define === "function" && define.amd) {
+		define([
+			"adapter",
+			"jquery",
+		], function(adapter, jQuery) {
+			return (root.Janus = factory(adapter, jQuery));
+		});
+	} else if (typeof module === "object" && module.exports) {
+		module.exports = (root.Janus = factory(require("adapter"), require("jquery")));
 	} else {
-		// Firefox of others, no need for the extension (but this doesn't mean it will work)
-		return true;
+		root.Janus = factory(root.adapter, root.jQuery);
 	}
-};
+}(this, function(adapter, jQuery) {
+var Janus;
+/**
+ * UMD wrapper.
+ */
 
-Janus.noop = function() {};
+/**
+ * jQuery wrapper.
+ * Allows $ to be used with proper namespacing.
+ */
+(function($) {
+/**
+ * jQuery wrapper.
+ */
 
-// Initialization
-Janus.init = function(options) {
-	options = options || {};
-	options.callback = (typeof options.callback == "function") ? options.callback : Janus.noop;
-	if(Janus.initDone === true) {
-		// Already initialized
-		options.callback();
-	} else {
-		if(typeof console == "undefined" || typeof console.log == "undefined")
-			console = { log: function() {} };
-		// Console logging (all debugging disabled by default)
-		Janus.trace = Janus.noop;
-		Janus.debug = Janus.noop;
-		Janus.vdebug = Janus.noop;
-		Janus.log = Janus.noop;
-		Janus.warn = Janus.noop;
-		Janus.error = Janus.noop;
-		if(options.debug === true || options.debug === "all") {
-			// Enable all debugging levels
-			Janus.trace = console.trace.bind(console);
-			Janus.debug = console.debug.bind(console);
-			Janus.vdebug = console.debug.bind(console);
-			Janus.log = console.log.bind(console);
-			Janus.warn = console.warn.bind(console);
-			Janus.error = console.error.bind(console);
-		} else if(Array.isArray(options.debug)) {
-			for(var i in options.debug) {
-				var d = options.debug[i];
-				switch(d) {
-					case "trace":
-						Janus.trace = console.trace.bind(console);
-						break;
-					case "debug":
-						Janus.debug = console.debug.bind(console);
-						break;
-					case "vdebug":
-						Janus.vdebug = console.debug.bind(console);
-						break;
-					case "log":
-						Janus.log = console.log.bind(console);
-						break;
-					case "warn":
-						Janus.warn = console.warn.bind(console);
-						break;
-					case "error":
-						Janus.error = console.error.bind(console);
-						break;
-					default:
-						console.error("Unknown debugging option '" + d + "' (supported: 'trace', 'debug', 'vdebug', 'log', warn', 'error')");
-						break;
-				}
-			}
-		}
-		Janus.log("Initializing library");
-		// Helper method to enumerate devices
-		Janus.listDevices = function(callback, config) {
-			callback = (typeof callback == "function") ? callback : Janus.noop;
-			if (config == null) config = { audio: true, video: true };
-			if(navigator.mediaDevices) {
-				navigator.mediaDevices.getUserMedia(config)
-				.then(function(stream) {
-					navigator.mediaDevices.enumerateDevices().then(function(devices) {
-						Janus.debug(devices);
-						callback(devices);
-						// Get rid of the now useless stream
-						try {
-							stream.stop();
-						} catch(e) {}
-						try {
-							var tracks = stream.getTracks();
-							for(var i in tracks) {
-								var mst = tracks[i];
-								if(mst !== null && mst !== undefined)
-									mst.stop();
-							}
-						} catch(e) {}
-					});
-				})
-				.catch(function(err) {
-					Janus.error(err);
-					callback([]);
-				});
-			} else {
-				Janus.warn("navigator.mediaDevices unavailable");
-				callback([]);
-			}
-		}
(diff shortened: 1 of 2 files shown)


lminiero:
I'll leave others to discuss this for the time being. This already looks more complicated than needed for me, and something that may live as a different project/repo (translated: something I shouldn't have to worry about :wink: ), so I'll have to think about this.

thehunmonkgroup:
Fair enough. Just for clarity, I'll add as long as you:

 - Keep all methods/properties on the Janus object (which it already is)
 - Leave the beginning and ending wrapper code
 - Don't add any new dependencies to the library

Then you can do whatever you want with the code I've wrapped, and it will work fine.

lminiero:
@thehunmonkgroup can you update the branch so that it's in sync with the current status? It would help to have something up-to-date for me (and others) to test and familiarize with this, so that we can consider merging. Not sure if merging #894 later on might cause another conflict, but I plan to merge that soon anyway in case you'd rather wait for that to happen.

thehunmonkgroup:
yeah, because of all the whitespace changes, it's best to clean this up after any other patches that affect .js files, so i'll wait

lminiero:
> yeah, because of all the whitespace changes, it's best to clean this up after any other patches that affect .js files, so i'll wait

I merged the PR I referenced to, so there shouldn't be other changes waiting to be merged. FYI, if you add `?w=1` to the `/files` diff section of a PR, it ignores whitespace changes and focuses on the "meat" (although not sure how well it performs, as it still showes a lot in your patch). As a side note, I noticed some indentation issues in your current patch, so please fix those as well in future revisions.

thehunmonkgroup:
Can you be more specific about the indentation issues you want fixed? I've purposefully *not* indented the code inside the wrappers, but if that's the thing you're pointing to, and you want that changed, then let me know.

lminiero:
I'm talking about the UMD wrapper code you added, the part that starts at `(function (root, factory) {`. By the way, you probably already know this, but we don't load jquery and adapter programmatically anymore, we expect both to be available (which means the web page is supposed to load them). Just mentioning this as that code still seems to do it, so it was probably done before we changed that.

lminiero:
PS: when I mean indenting, I only mean the use of spaces instead of tabs.

thehunmonkgroup:
OK, I've updated the PR, and tested both the jquery and nojquery cases, they seem to be working fine. To avoid stale code, I simply re-wrapped the .js files from latest master and rebased my PR, so anything inside the wrapper is the latest JS code in master.

Also fixed the tab stuff I think. Lemme know if you need anything else.

html/janus.js
```diff
@@ -22,185 +22,39 @@
...
-		return ($('#janus-extension-installed').length > 0);
+/**
+ * UMD wrapper.
+ */
+(function (root, factory) {
+	if (typeof define === "function" && define.amd) {
+		define([
+			"adapter",
+			"jquery",
+		], function(adapter, jQuery) {
```
lminiero: This stuff is not needed anymore, we don't load external JavaScript files programmatically. Both are linked externally in html files, and currently point to their cdnjs link.

html/janus.nojquery.js
```diff
@@ -22,248 +22,29 @@
...
+/**
+ * UMD wrapper.
+ */
+(function (root, factory) {
+	if (typeof define === "function" && define.amd) {
+		define([
+			"adapter",
+		], function(adapter) {
+			return (root.Janus = factory(adapter));
+		});
```
lminiero: Same as above on adapter.

thehunmonkgroup: @lminiero this is the exact point of the UMD wrapper -- those sections you're referencing above handle the case when a user wants to load the library via RequireJS. The next case supports if a user wants to load the library via Node.js/CommonJS, and the final case if the user wants to load the library in the old school global scope (which is what the demos do). Removing those lines would defeat the purpose of the wrapper :)

lminiero: What I mean is that we don't load those libraries in the demos anymore, support for that has been removed from janus.js. Now you have to load them yourself in the webpage (or load them programmatically yourself). The reason for this is that we don't ship any external js now, they're all loaded from cdnjs.

thehunmonkgroup: @lminiero yep, I'm aware that they are loaded globally in the demos, the wrapper handles the globally loaded case, translating the globally namespaced variables to locally scoped variables in the factory function.

The demos themselves will need no adjustment with this approach, they can continue to load their support libraries from the CDN.

abustany:
In the meanwhile, here's a Webpack 2 configuration that achieves a similar effect:

```
var path = require('path')
var webpack = require('webpack')

module.exports = {
  ...
  module: {
    rules: [
      {
        // Shim the Janus module: make it export the Janus global variable,
        // and automatically import jQuery.
        test: /vendor\/janus-.+.js$/, // path to the Janus module here, I have it in the "vendor" folder
        use: [
          'imports-loader?$=jquery,jQuery=jquery,adapter=webrtc-adapter',
          'exports-loader?Janus'
        ]
      },
      ...
  },
}

**Comment by lminiero** [comment]
As I said in other discussions, I'm not going to make `janus.js` so heavy by involving nodejs-like stuff and imports on other dependencies.

**Comment by abustany** [comment]
Yep, this is not a proposed code change for Janus, it's just a reference for downstream users that use webpack 2 for their web apps. Those webpack users can use the snippet above, and janus.js in its current state will then behave as a module: webpack itself will wrap janus.js locally with what this PR proposes.

**Comment by lminiero** [comment]
Ah, got it, sorry for the misunderstanding! Proof enough my JS voodoo mostly sucks... :smile:

**Comment by cmacq2** [comment]
Thinking about issue #879 raises the related issue of "it would be really nice if `janus.js` could be a proper import-able module". This PR addresses it by explicitly hardcoding a preferred solution, but:

 - It ties `janus.js` to UMD, which may 'get' in the way.
 - Module format/syntax kind of depends on the project.
 - It is inherently boilerplate!
 - It can be auto-generated these days, supporting numerous module formats to boot.

**Please note**: this would not touch the `janus.js` code itself, and `janus.js` code would continue to function by itself. Rather during build a couple of alternatives are auto-generated which basically wrap the existing code in a format that is easier to consume for people who *do* use the various JavaScript build tools & module systems.

-----------------

What I would propose is to (optionally) execute `rollup.js` and emit importable modules from the plain `janus.js` code during `make`. We could use a simple facade `janus.module.js` which consists of just this:

```
export default Janus;
```

And then have a `rollup` config which knows to pull in `Janus` from `janus.js`, with a couple of make target variants to auto-generate CommonJS, ES andUMD module formats by varying the `rollup.js` output format (`-f`). 

That will require some `autoconf` & `make` -fu to make building this conditional on `node` & `rollup` being available. We'd need to document this as part of the build instructions as well.

Alternatively, we could split the build into a two-stage process, and in that case you could set up a `npm` package for it and require people do e.g. `npm build` as well as `make` before `make install`.

At a later stage explicit `--disable`/`--enable` flags for  `./configure` would be desirable.

**Comment by lminiero** [comment]
I haven't understood a single thing you said... :wink: 
As discussed, I'm all but a JS guru, especially when it comes at these advanced points. This is one of the reasons why I'm fundamentally against making janus.js way more complicated than it currently is (considering I'm the one that would have to maintain that). Again, I'd definitely welcome a more flexible client framework proposed as a different repo we can reference in the resources page: the more options people can choose from, the better.

**Comment by cmacq2** [comment]
> I haven't understood a single thing you said... :wink:
As discussed, I'm all but a JS guru, especially when it comes at these advanced points. 

Well the basic point of this PR is that many people would prefer to write modular code. Using and abusing global variables to get the job done is no longer considered such a great idea, instead a number of competing module definitions/syntaxes have emerged in JavaScript land.

This is somewhat unfortunate as there is not necessarily any 'right' way to expose `janus.js`  as a proper module.

This PR proposes to settle on using UMD as a compatibility wrapper around 2, but there is also an increasingly popular support for the newer JavaScript standard syntax which uses `import` and `export` statements instead. To translate it to a C/C++ analog: `export` gives you control over what to expose to an `import`, a bit like using `-fvisibiliy=hidden` as GCC compiler flag and then manually adding `__attribute__()` annotations in C/C++.

> This is one of the reasons why I'm fundamentally against making janus.js way more complicated than it currently is (considering I'm the one that would have to maintain that). Again, I'd definitely welcome a more flexible client framework proposed as a different repo we can reference in the resources page: the more options people can choose from, the better.

What I proposed would not alter `janus.js` at all. Instead there would be a wrapper file which just provides the basic glue syntax to look like a module of its own *and* there would be a build step in which e.g. `rollup` would basically splice `janus.js` into that and emit a js output file which people could use as a modular alternative to `janus.js`.

So people can choose: either `janus.js` or a modular alternative which just wraps the existing `janus.js` code in the 'right' module syntax for what the client code/environment expects. This is particular useful for environments where you cannot simply include a `<script>` tag in HTML in order to load `janus.js` alongside your own application code.

Taking a hypothetical `example.js` as example:

```
// example.js
Example = {
  doSomething: function() { console.log('hello, world'); }
};
```

Combined with a separate `example.module.js` which looks like this:

```
// example.module.js
import Example from './example.js';
export default Example;
```

Then, using rollup during build, another `example.build.js` file could be generated which is equivalent to:
```
// example.build.js
Example = {
  doSomething: function() { console.log('hello, world'); }
};
export default Example;
```

And people could use that as an alternative to the plain `example.js` like this:

```
import Example from './example.build.js'
Example.doSomething();
```

cmacq2:
Please review #1029 which takes an alternative approach and auto-generates the module definition during build time (i.e. during `make`). This does away with the need for changes to `janus.js` code, and also lets us support a few more flavours of popular JavaScript module syntax.

lminiero:
Closing in light of the new, more flexible, proposal.

PR #796, opened 2017-02-27
RTP context/rewriting as part of core instead of plugins

lminiero:
This PR integrates the stuff some plugins (AudioBridge/SIP/VideoRoom/Streaming) are doing exactly the same way in the rtp.c/.h core stuff. Specifically, the management of RTP contexts in order to properly overwrite sequence numbers and timestamps when the source (SSRC) changes. Ideally, this should make it easier for other plugins needing the same to use. Besides, in the future it will help the core take care of that of its own too (i.e., not rely too much on what plugins push and fix things accordingly).


lminiero:
Merging.

PR #797, opened 2017-02-28
check for existence of peerconnection object in oniceconnectionstatechange callback

thehunmonkgroup:
It's possible for the ```oniceconnectionstatechange``` callback to be fired
after the ```janus.js``` library has already cleaned up its peer connection
object, so check for its existence before accessing it.


lminiero:
I had already committed it yesterday but forgotten to push, done it now: https://github.com/meetecho/janus-gateway/commit/8107965ae0f09a71eaa5a24ca484799a86361c7e

PR #803, opened 2017-03-02
Better integration of new IP tools in Janus core and plugins

lminiero:
This PR tries to follow up the excellent work started by @cmacq2 in #776. It renames some of the methods and stuff to have a shared prefix for everything (`janus_network_`), adds a new method to detect your local IP (basically what we already had in `janus.c` and `janus_sip.c`, can be improved), and moves `janus_streaming_lookup_network_interface` to the ip-utils instead (as `janus_network_lookup_interface`, which now tries to query IPv6 too) so that other plugins in the future will be able to use it to.

I may do some more things with this, as I'm figuring out, for instance, if the "ignore interface" stuff we have in `ice.c` can be generalized somehow and moved to the ip-utils too, but that's pretty much it. It shouldn't break anything so I plan to merge soon.

### DIFF
--- a/ip-utils.c
+++ b/ip-utils.c
@@ -276,7 +276,7 @@ int janus_network_detect_local_ip(janus_network_query_options addr_type, janus_n
 char *janus_network_detect_local_ip_as_string(janus_network_query_options addr_type) {
 	janus_network_address addr;
 	janus_network_address_string_buffer buf;
-    int res = janus_network_detect_local_ip(addr_type, &addr) || janus_network_address_to_string_buffer(&addr, &buf);
+	int res = janus_network_detect_local_ip(addr_type, &addr) || janus_network_address_to_string_buffer(&addr, &buf);
 	if(res != 0)
 		return NULL;
 	return g_strdup(janus_network_address_string_from_buffer(&buf));
--- a/ip-utils.c
+++ b/ip-utils.c
@@ -166,6 +166,23 @@ int janus_network_address_is_null(const janus_network_address *a) {
 	return !a || a->family == AF_UNSPEC;
 }
 
+int janus_network_address_from_sockaddr(struct sockaddr *s, janus_network_address *a) {
+	if(!s || !a)
+		return -EINVAL;
+	if(s->sa_family == AF_INET) {
+		a->family = AF_INET;
+		struct sockaddr_in *addr = (struct sockaddr_in *)s;
+		a->ipv4 = addr->sin_addr;
+		return 0;
+	} else if(s->sa_family == AF_INET6) {
+		a->family = AF_INET6;
+		struct sockaddr_in6 *addr = (struct sockaddr_in6 *)s;
+		a->ipv6 = addr->sin6_addr;
+		return 0;
+	}
+	return -EINVAL;
+}
+
 int janus_network_address_to_string_buffer(const janus_network_address *a, janus_network_address_string_buffer *buf) {
 	if(buf && !janus_network_address_is_null(a)) {
 		janus_network_address_string_buffer_nullify(buf);
@@ -197,6 +214,28 @@ const char *janus_network_address_string_from_buffer(const janus_network_address
 	} else {
 		return b->family == AF_INET ? b->ipv4 : b->ipv6;
 	}
+}
+
+int janus_network_string_is_valid_address(janus_network_query_options addr_type, const char *user_value) {
+	janus_network_address a;
+	return janus_network_string_to_address(addr_type, user_value, &a) == 0;
+}
+
+int janus_network_string_to_address(janus_network_query_options addr_type, const char *user_value, janus_network_address *result) {
+	if((addr_type != janus_network_query_options_ipv4 &&
+			addr_type != janus_network_query_options_ipv6 &&
+			addr_type != janus_network_query_options_any_ip) || !user_value || !result) {
+		return -EINVAL;
+	}
+	if((addr_type & janus_network_query_options_ipv4) && inet_pton(AF_INET, user_value, &result->ipv4) > 0) {
+		result->family = AF_INET;
+		return 0;
+	}
+	if((addr_type & janus_network_query_options_ipv6) && inet_pton(AF_INET6, user_value, &result->ipv6) > 0) {
+		result->family = AF_INET6;
+		return 0;
+	}
+	return -EINVAL;
 }
 
 int janus_network_lookup_interface(const struct ifaddrs *ifas, const char *iface, janus_network_address *result) {
--- a/ip-utils.c
+++ b/ip-utils.c
@@ -24,33 +24,6 @@
 
 #include "ip-utils.h"
 
-int janus_network_detect_local_ip(char *buf, size_t buflen) {
-	struct sockaddr_in addr;
-	socklen_t len;
-	int fd = socket(AF_INET, SOCK_DGRAM, 0);
-	if(fd == -1)
-		goto error;
-	addr.sin_family = AF_INET;
-	addr.sin_port = htons(1);
-	inet_pton(AF_INET, "1.2.3.4", &addr.sin_addr.s_addr);
-	if(connect(fd, (const struct sockaddr*) &addr, sizeof(addr)) < 0)
-		goto error;
-	len = sizeof(addr);
-	if(getsockname(fd, (struct sockaddr*) &addr, &len) < 0)
-		goto error;
-	if(getnameinfo((const struct sockaddr*) &addr, sizeof(addr),
-			buf, buflen,
-			NULL, 0, NI_NUMERICHOST) != 0)
-		goto error;
-	close(fd);
-	return 0;
-error:
-	if(fd != -1)
-		close(fd);
-	g_strlcpy(buf, "127.0.0.1", buflen);
-	return -1;
-}
-
 static int janus_ip_compare_byte_arrays(const uint8_t *b1, const uint8_t *b2, const size_t size) {
 	size_t i;
 	for(i = 0; i < size; ++i) {
@@ -92,8 +65,7 @@ static int janus_ip_iface_matches(const struct ifaddrs *ifa, const janus_network
 			((query->mode & janus_network_query_options_ipv4) && janus_ip_iface_matches_ipv4(ifa, &query->ipv4) == 1) ||
 			((query->mode & janus_network_query_options_ipv6) && janus_ip_iface_matches_ipv6(ifa, &query->ipv6) == 1) ||
 			((query->mode & janus_network_query_options_name) && janus_ip_iface_matches_name(ifa, query->device_name) == 1) ? 1 : 0;
-	}
-	else {
+	} else {
 		return -EINVAL;
 	}
 }
@@ -126,8 +98,7 @@ int janus_network_prepare_device_query(const char *user_value, const janus_netwo
 		}
 
 		return ((query_mode == janus_network_query_options_none) == (query->mode == janus_network_query_options_none)) ? 0 : -EINVAL;
-	}
-	else {
+	} else {
 		return -EINVAL;
 	}
 }
@@ -144,8 +115,7 @@ static int janus_ip_copy_ipv4(const struct sockaddr_in *iface, struct in_addr *r
 int janus_network_get_devices_ipv4(const struct ifaddrs *ifa, const janus_network_query_config *query, struct in_addr *result) {
 	if(ifa && ifa->ifa_addr && (ifa->ifa_addr->sa_family == AF_INET) && result && query && (query->mode & janus_network_query_options_ipv4)) {
 		return janus_ip_copy_ipv4((struct sockaddr_in *) ifa->ifa_addr, result);
-	}
-	else {
+	} else {
 		return -EINVAL;
 	}
 }
@@ -163,8 +133,7 @@ static int janus_ip_copy_ipv6(const struct sockaddr_in6 *iface, struct in6_addr 
 int janus_network_get_devices_ipv6(const struct ifaddrs *ifa, const janus_network_query_config *query, struct in6_addr *result) {
 	if(ifa && ifa->ifa_addr && (ifa->ifa_addr->sa_family == AF_INET6) && result && query && (query->mode & janus_network_query_options_ipv6)) {
 		return janus_ip_copy_ipv6((struct sockaddr_in6 *) ifa->ifa_addr, result);
-	}
-	else {
+	} else {
 		return -EINVAL;
 	}
 }
@@ -181,8 +150,7 @@ int janus_network_get_device_address(const struct ifaddrs *ifa, janus_network_ad
 			default:
 				return -EINVAL;
 		}
-	}
-	else {
+	} else {
 		return -EINVAL;
 	}
 }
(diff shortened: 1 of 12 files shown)


ip-utils.c
```diff
@@ -12,13 +12,45 @@
...
+#include <netdb.h>
 #include <net/if.h>
 #include <string.h>
+#include <unistd.h>
+
+#include <glib.h>
 
 #include "ip-utils.h"
 
+int janus_network_detect_local_ip(char *buf, size_t buflen) {
```
cmacq2: The purpose of the ip-utils set of functions is to provide abstractions that makes it easier to write IPv4 & IPv6 compatible code. This function is not IPv6 compatible and does not mark itself as being IPv4. 

Additionally you pass a raw char buffer around for receiving output, which is contrary to conventions for the other functions:
 * It puts the onus on the caller to properly allocate the buffer and pass size (error prone)
 * It is a leaky abstraction, and the 'proper' way to determine the `buflen` creates namespace pollution in the caller context because the correct `buflen` values are should be determined by predefined header constants.
 * This API is inherently hard to use in an IPv4 & IPv6 compatible manner. It is not really fire and forget.

A better (API) would be something like this:
```
   /*
    * addr_type must be one of:
    * - janus_ip_network_query_options_ipv4 => caller wants to support IPv4 only
    * - janus_ip_network_query_options_ipv6 => caller wants to support IPv6 only
    * - janus_ip_network_query_options_any_ip => caller wants to support both IPv4 and IPv6
    */ 
   janus_network_detect_local_ip(janus_ip_network_query_options addr_type, janus_network_address *result);
```

The callee (function) is now free to implement logic for both IPv4 and IPv6 and even better it can reset the `result` to an invalid (null-valued) address which can be tested for in other parts of the code. This allows other parts of the code to recover somewhat from lack of error checking in the caller context, because a null valued result is invalid by definition. So it's more robust.

Then with additional socket/bind/multicast operations also converted to `ip-utils` style the caller could:

 1. Possibly: populate a `struct sockaddr` as follows: `void janus_network_address_to_sockaddr(janus_network_address *addr, int port, struct sockaddr *result)`
 2. Directly open a socket and bind to the address if it wanted to, passing the returned address (if available) and a port number: `int janus_network_address_to_socket(janus_network_address *addr, int port);` and `int janus_network_address_bind_to_socket(janus_network_address *addr, int port);`
 3. Convert to presentation format (strings) via the `janus_network_address_to_string_buffer` call, and then obtain the correct string with: `janus_network_address_string_from_buffer` (already available).

ip-utils.c
```diff
@@ -191,10 +223,32 @@ int janus_network_address_string_buffer_is_null(const janus_network_address_stri
...
+const char *janus_network_address_string_from_buffer(const janus_network_address_string_buffer *b) {
 	if(janus_network_address_string_buffer_is_null(b)) {
 		return NULL;
 	} else {
 		return b->family == AF_INET ? b->ipv4 : b->ipv6;
 	}
 }
+
+int janus_network_lookup_interface(const struct ifaddrs *ifas, const char *iface, janus_network_address *result) {
+	if(result == NULL)
```
cmacq2: You forget to check `ifas` for `NULL`, and do not signal the error explicitly. At the top of the function body, you probably should include something like this:

```
   if(!ifas || !result || !iface) {
      return -EINVAL;
   }
```

ip-utils.c
```diff
@@ -191,10 +223,32 @@ int janus_network_address_string_buffer_is_null(const janus_network_address_stri
...
+		return -1;
+	janus_network_address_nullify(result);
+	if(iface == NULL)
+		return -1;
+	janus_network_query_config q;
+	/* Let's see if iface is an IPv4 address, an IPv6 address, or possibly an interface name */
+	if(janus_network_prepare_device_query(iface,
+			janus_network_query_options_ipv4 | janus_network_query_options_ipv6 | janus_network_query_options_name, &q)) {
+		/* None of them..? */
+		return -1;
```
cmacq2: This should be `return -EINVAL` so the error type (invalid input arguments) is signaled as well as the bare fact that it is an error.

cmacq2: That goes for other places where a simple `return -1` is performed: it is better to use `- <errno>` type error codes with values from `errno.h`. That way the caller can inspect the return code and take more informed decisions based on the type of error being signaled, even if that is something trivial like:
```
    int error = janus_network_lookup_interface(...);
    if(error) {
        JANUS_LOG(LOG_ERR, "Oh noes: %s\n", strerror(-error));
    }
```

ip-utils.h
```diff
@@ -176,18 +183,27 @@ void janus_network_address_string_buffer_nullify(janus_network_address_string_bu
...
- * \return a pointer to the human readable representation of the network address inside the given buffer, or NULL if the buffer is invalid or NULL.
+ * \return A pointer to the human readable representation of the network address inside the given buffer, or NULL if the buffer is invalid or NULL.
  * \see \c janus_network_address_to_string_buffer
  */
-const char * janus_network_address_string_from_buffer(const janus_network_address_string_buffer *b);
+const char *janus_network_address_string_from_buffer(const janus_network_address_string_buffer *b);
+
+/*!
+ * \brief Look for an interface name to prepare a janus_network_address instance
+ * \param ifas The list of interfaces to lookup into (e.g., as returned from getifaddrs)
```
cmacq2: Nitpick: "to lookup into" should probably be "to look into" ?

ip-utils.h
```diff
@@ -176,18 +183,27 @@ void janus_network_address_string_buffer_nullify(janus_network_address_string_bu
...
  * \param b The buffer containing the given network
- * \return a pointer to the human readable representation of the network address inside the given buffer, or NULL if the buffer is invalid or NULL.
+ * \return A pointer to the human readable representation of the network address inside the given buffer, or NULL if the buffer is invalid or NULL.
  * \see \c janus_network_address_to_string_buffer
  */
-const char * janus_network_address_string_from_buffer(const janus_network_address_string_buffer *b);
+const char *janus_network_address_string_from_buffer(const janus_network_address_string_buffer *b);
+
+/*!
+ * \brief Look for an interface name to prepare a janus_network_address instance
```
cmacq2: Nitpick: "Look for an interface name to prepare a janus_network_address instance" should probably be: "Convert an interface name or IP address to a janus_network_address instance" ?

ip-utils.h
```diff
@@ -176,18 +183,27 @@ void janus_network_address_string_buffer_nullify(janus_network_address_string_bu
...
+ * \param ifas The list of interfaces to lookup into (e.g., as returned from getifaddrs)
+ * \param iface The interface name or IP address to look for
+ * \param result Pointer to a valid janus_network_address instance that will contain the result
+ * \return 0 in case of success, a negative integer otherwise
```
cmacq2: And this should become `0 in case of success, -EIVAL otherwise` because the error would be invalid arguments...

ip-utils.c
```diff
@@ -12,13 +12,45 @@
...
+int janus_network_detect_local_ip(char *buf, size_t buflen) {
+	struct sockaddr_in addr;
+	socklen_t len;
+	int fd = socket(AF_INET, SOCK_DGRAM, 0);
+	if(fd == -1)
+		goto error;
+	addr.sin_family = AF_INET;
+	addr.sin_port = htons(1);
+	inet_pton(AF_INET, "1.2.3.4", &addr.sin_addr.s_addr);
+	if(connect(fd, (const struct sockaddr*) &addr, sizeof(addr)) < 0)
```
cmacq2: What exactly are you trying to do? If all you want to know is the local IP address you could simply walk a `struct * ifaddrs`, filtering for the correct `AF_*` type and maybe checking the address is not link-local (in case of IPv6)?

Also: this function lacks sanity checks for multihomed setups (multiple network interfaces) if you are interested in being able to restrict traffic from Janus to peers to specific network interfaces

ip-utils.c
lminiero: Excellent feedback, thanks for looking this through! Yes, this method is currently a bit hammered in and needs to be aligned in style and scope. I'll follow your suggestion on how to fix that. Anyway, the purpose here is not to constrain anything, but just to get a valid IP address that can theoretically reach the outside, so that you can advertise it around (e.g., in the SDP), so I'd still keep a string as the return value somehow to keep things easier.

ip-utils.c
lminiero: Good catch, thanks!

ip-utils.c
lminiero: Ack, you're right, the documentation mentioned that for other methods :+1:

ip-utils.h
lminiero: Yep, sorry, my English still sucks at times :smile: 
In my defence, I speak an excellent Italian!

ip-utils.h
lminiero: Will fix :+1:

ip-utils.h
lminiero: Ack :+1:

ip-utils.c
lminiero: The only purpose of this method is not just to find a local IP address, but to find an IP address that is likely to be used for communicating with the outside. We create a file descriptor and we point it to an external address: the UDP connect doesn't really connect, but still involves the OS routing table to see which interface would be able to get there. We then get the local address from the socket.

I think this was contributed a long time ago in a PR, and seems to work fine. But yes, it's currently limited to IPv4 and should be improved to support IPv6 too.

ip-utils.c
cmacq2: > so I'd still keep a string as the return value somehow to keep things easier.

That is actually what makes the code *harder* to use because now the caller has to pick the correct buffer size to use (which is one opportunity for errors) and should zero it (which is another). You simplify the API by taking away the possibility for making the wrong decisions, so pushing towards the proper structs actually helps. For example this is how you would use the function to obtain the string afterwards:

```
    janus_network_address a;
    janus_network_address_string_buffer b;
    int error = janus_network_detect_local_ip(janus_ip_network_query_options_any_ip, &a);
    if(error) {
       // handle it.
    }
    janus_network_address_to_string_buffer(&a, &b);
    const char * formatted_ip_address = janus_network_address_string_from_buffer(&b);
    if(formatted_ip_address) {
        // use it.
        printf("Your IP address is: %s\n", formatted_ip_address);
    }
```

Now if the argument is that the primary (or sole) use of the function is to obtain such strings, then that calls for a higher level convenience function which does precisely that:
```
    const char *janus_network_detect_local_ip_as_string(janus_ip_network_query_options addr_type);
```

It's body would be equivalent to this:
```
    const char *janus_network_detect_local_ip_as_string(janus_ip_network_query_options addr_type) {
        janus_network_address a;
        janus_network_address_string_buffer b;
        return janus_network_detect_local_ip(addr_type, &a) || janus_network_address_to_string_buffer(&a, &b) ? NULL : janus_network_address_string_from_buffer(&b);
    }
```

That way you keep the API clean and unambiguous, plus you obtain the convenience function which you were really after.

lminiero: Fair points, I'll address this tomorrow :+1:

ip-utils.c
cmacq2: > The only purpose of this method is not just to find a local IP address, but to find an IP address that is likely to be used for communicating with the outside. 

Yes I thought so too; but unless I am much mistaken the actual IP address you are going to find will be one of the interface pointing to the default gateway. This is the correct answer if and only if the client communicates through Janus via the same network interface. It looks like in multihomed setups you can get the 'wrong' answer, for example.

ip-utils.h
cmacq2: Make that `-EINVAL`. Missed an N.

ip-utils.c
lminiero: That's what the ability to override the method in janus.cfg is for.  I don't think there's an autodetect approach that works for everybody. The best we can do is make it work for the internet in most cases.

lminiero:
This latest commit should address the points raised by @cmacq2. Seems to be working as expected when we call it in `janus.c` and `janus_sip.c` (will integrate it in `janus_nosip.c` as well when we merge this).

ip-utils.c
```diff
@@ -232,23 +200,84 @@ const char *janus_network_address_string_from_buffer(const janus_network_address
...
+	}
+	if(fd != -1)
+		close(fd);
+	return found ? 0 : -EINVAL;
+}
+
+char *janus_network_detect_local_ip_as_string(janus_network_query_options addr_type) {
+	janus_network_address addr;
+	janus_network_address_string_buffer buf;
+    int res = janus_network_detect_local_ip(addr_type, &addr) || janus_network_address_to_string_buffer(&addr, &buf);
```
cmacq2: Small nitpick: you have an indentation issue here.

ip-utils.c
```diff
@@ -232,23 +200,84 @@ const char *janus_network_address_string_from_buffer(const janus_network_address
...
+		return -EINVAL;
+	janus_network_address_nullify(result);
+	gboolean found = FALSE;
+	int fd = -1;
+	if(addr_type == janus_network_query_options_ipv4 || addr_type == janus_network_query_options_any_ip) {
+		/* Let's try IPv4 (FIXME Should probably use other internal methods) */
+		struct sockaddr_in addr;
+		socklen_t len;
+		fd = socket(AF_INET, SOCK_DGRAM, 0);
+		if(fd > -1) {
```
cmacq2: Socket is documented to return -1 on error, so it's better to test explicitly for that.

ip-utils.c
lminiero: :+1:

ip-utils.c
lminiero: We only care about a success anyway, so as long as it's 0 or a positive integer we're fine.

janus.c
```diff
@@ -3680,6 +3667,7 @@ gint main(int argc, char *argv[])
 		gboolean private_address = FALSE;
 		const char *test_ip = nat_1_1_mapping ? nat_1_1_mapping : local_ip;
 		struct sockaddr_in addr;
+		/* FIXME This ugly check only works with IPv4... */
```
cmacq2: What you need here is a helper function in `ip-utils` like:

```
// addr_type may be one of *_ipv4, *_ipv6 or *_any_ip
// user_value is the string to convert
// result will be populated if successful.
// returns 0 on success, -EINVAL otherwise (if any argument is invalid).
int janus_network_string_to_address(janus_ip_network_query_options addr_type, const char *user_value, janus_network_address *result);

// convenience version of the above function prototype, passes *_any_ip to the full version.
int janus_network_string_to_address_default(const char *user_value, janus_network_address *result);
```

The guts for that function could be copied from `janus_prepare_network_device_query` and adapted as needed.

ip-utils.c
```diff
@@ -232,23 +200,84 @@ const char *janus_network_address_string_from_buffer(const janus_network_address
...
+	return found ? 0 : -EINVAL;
+}
+
+char *janus_network_detect_local_ip_as_string(janus_network_query_options addr_type) {
+	janus_network_address addr;
+	janus_network_address_string_buffer buf;
+    int res = janus_network_detect_local_ip(addr_type, &addr) || janus_network_address_to_string_buffer(&addr, &buf);
+	if(res != 0)
+		return NULL;
+	return g_strdup(janus_network_address_string_from_buffer(&buf));
```
cmacq2: Possibly consolidate this in a ternary statement?
```
return res ? NULL : g_strdup(janus_network_address_string_from_buffer(&buf));
```

janus.c
cmacq2: The first of two could look like this:
```
int janus_network_string_to_address(janus_ip_network_query_options addr_type, const char *user_value, janus_network_address *result) {
    if((addr_type != janus_ip_network_query_options_ipv4 && addr_type != janus_ip_network_query_options_ipv6 && addr_type != janus_ip_network_query_options_any_ip) || !user_value || !result) {
        return -EINVAL;
    }
    else {
        if(inet_pton(AF_INET, user_value, &result->ipv4) > 0) {
            result->family = AF_INET;
            return 0;
        }
        if(inet_pton(AF_INET6, user_value, &result->ipv6) > 0) {
            result->family = AF_INET6;
            return 0;
        }
        return -EINVAL;
    }
}
```

The other one is:

```
int janus_network_string_is_valid_address(janus_ip_network_query_options addr_type, const char *user_value) {
    janus_network_address a;
    return janus_network_string_to_address(addr_type, user_value, &a);
}
```

plugins/janus_sip.c
```diff
@@ -176,7 +176,7 @@ static volatile gint initialized = 0, stopping = 0;
 static gboolean notify_events = TRUE;
 static janus_callbacks *gateway = NULL;
 
-static char local_ip[INET6_ADDRSTRLEN];
+static char *local_ip;
```
cmacq2: You may want to initialise this to NULL, just to be safe.

plugins/janus_sip.c
```diff
@@ -760,45 +760,32 @@ int janus_sip_init(janus_callbacks *callback, const char *config_path) {
 		gboolean local_ip_set = FALSE;
```
cmacq2: It's hard to be sure when looking at the diff, but if you do initialise the `local_ip` pointer to NULL, you may no longer need this local variable?

janus.c
lminiero: Thanks, I'll integrate this: we do indeed have some redundant methods in `utils.c` that should be removed/replaced, and this seem to do the trick (besides better handling the scenario you referred to).

ip-utils.c
lminiero: It was ternary in the code you proposed: I preferred to split it to make it better readable.

plugins/janus_sip.c
lminiero: Ack.

plugins/janus_sip.c
lminiero: Good point: we had it because before `local_ip` was an array of characters, now that it's a pointer it's useless.

lminiero:
This last commit addresses the comments by @cmacq2, and makes a couple of changes to get rid of spurious network-related methods in 'utils.c`.

Specifically, `janus_address_to_ip` converted a `struct sockaddr` to an IP string, so I added a new method called `janus_network_address_from_sockaddr` that converts a `struct sockaddr` to a `janus_network_address` (only for the address part), allowing us to use the existing modules for that, e.g.:

	janus_network_address naddr;
	janus_network_address_string_buffer naddr_buf;
	if(janus_network_address_from_sockaddr((struct sockaddr *)addr, &naddr) != 0 ||
			janus_network_address_to_string_buffer(&naddr, &naddr_buf) != 0) {
		/* Handle error */
	}
	const char *ip = janus_network_address_string_from_buffer(&naddr_buf);

This has been integrated in the code that used the old approach, so `ice.c`, `turnrest.c` and `janus_http.c`.

The `janus_address_to_ip` method, instead, is a different matter... in fact, the stuff we have in the new `ip-utils`as of now only cares about addresses, but not ports. The `janus_network_address`, for instance, doesn't contain any info related to ports, meaning the `janus_network_address_from_sockaddr` method loses that in the process. Not sure adding a port property would be much useful, as we wouldn't use it for anything else right now. As such, since `janus_http.c` is the only code that currently used that, I simply moved the method from `utils.c` to there as a static method under a different name. Should there be a need to make this more globally available, we can worry about that later on.

We still have some `inet_pton` scattered across the code of some plugins, but not sure if we can remove those as of now, and they have a low priority anyway.

lminiero:
Merging, if issues arise we can address them accordingly. Thanks again to @cmacq2 for the tons of useful feedback! (besides providing the tools in the first place)

PR #804, opened 2017-03-02
Moved most of SRTP-related stuff to rtp.h/.c (cleans dtls and janus_sip)

lminiero:
This PR is motivated by the efforts in #799, where the NoSIP plugin shares a lot of code with the SIP plugin. It moves a few SRTP-related things around, specifically by putting it all in `rtp.h` and `rtp.c` (where it probably belongs) rather than in the DTLS code or the SIP plugin (where we duplicated a lot of stuff).

I guess that the `srtp_policy_t` stuff the SIP plugin does could be moved too in the future (we do something similar, although not the same thing, in `dtls.c`) but for now I think this is enough.

Should be harmless and I don't expect this to break anything, so I plan to merge soon.


lminiero:
Merging.

PR #807, opened 2017-03-06
Correctly handle parse error in janus.nojquery.js

bbfsdev:
We are getting parse errors due to empty string being passed to JSON parse.
Added better error handling to:
a) Not fail.
b) Understand why we get empty strings there.

The error:
Uncaught SyntaxError: Unexpected end of JSON input

### DIFF
--- a/html/janus.nojquery.js
+++ b/html/janus.nojquery.js
@@ -192,7 +192,13 @@ Janus.init = function(options) {
 						return;
 					}
 					// Got payload
-					params.success(JSON.parse(XHR.responseText));
+					try {
+						params.success(JSON.parse(XHR.responseText));
+					} catch(e) {
+						params.error(XHR, XHR.status, 'Could not parse response, error: ' + e +
+									 ', text: ' + XHR.responseText);
+						return;
+					}
 				};
 			}
 			try {


lminiero:
Thanks for spotting this and for providing a fix. There's another identical `JSON.parse` at line 207, so that should be addressed as well.

That said, before we can merge we need you to sign our [CLA](https://cla.conf.meetecho.com/janus/).

Thanks!

lminiero:
A couple additional `JSON.parse` that might have a try/catch is here:

https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L571
https://github.com/meetecho/janus-gateway/blob/master/html/janus.nojquery.js#L622

which is the part of the code that handles incoming messages (responses/events) for WebSockets instead of HTTP/longpoll. I guess that's not really a problem there, though, as your issue may be with HTTP communication and not with what Janus sends you (which is always going to be valid JSON).

lminiero:
Added this myself in [this commit here](https://github.com/meetecho/janus-gateway/commit/bafe1a7e01abedc21d1e0adab2bd540857e0aa2d), so no need for you to update the PR. Thanks!

PR #810, opened 2017-03-10
Fix: missing directory (full path) when renaming from temp extension

neilkinnish:


lminiero:
Thanks for the fix, I'll review on Monday. In the meanwhile, could you sign our [CLA](https://cla.conf.meetecho.com/janus/), so that I can merge when it's ready?

neilkinnish:
Done :)

record.c
```diff
@@ -248,8 +248,16 @@ int janus_recorder_close(janus_recorder *recorder) {
...
 		char newname[1024];
 		memset(newname, 0, 1024);
-		g_snprintf(newname, strlen(recorder->filename)-strlen(rec_tempext), "%s", recorder->filename);
-		if(rename(recorder->filename, newname) != 0) {
+		char tmpname[1024];
+		memset(tmpname, 0, 1024);
+		if(recorder->dir) {
+			g_snprintf(newname, strlen(recorder->filename)-strlen(rec_tempext), "%s/%s", recorder->dir, recorder->filename);
+			g_snprintf(tmpname, strlen(recorder->filename), "%s/%s", recorder->dir, recorder->filename);
+		}else{
```
lminiero: Can you add spaces there? `} else {`

record.c
```diff
@@ -248,8 +248,16 @@ int janus_recorder_close(janus_recorder *recorder) {
 		/* We need to rename the file, to remove the temporary extension */
 		char newname[1024];
 		memset(newname, 0, 1024);
-		g_snprintf(newname, strlen(recorder->filename)-strlen(rec_tempext), "%s", recorder->filename);
-		if(rename(recorder->filename, newname) != 0) {
+		char tmpname[1024];
+		memset(tmpname, 0, 1024);
+		if(recorder->dir) {
+			g_snprintf(newname, strlen(recorder->filename)-strlen(rec_tempext), "%s/%s", recorder->dir, recorder->filename);
```
lminiero: We should actually check if `strlen(recorder->filename)-strlen(rec_tempext)` (or `strlen(recorder->filename)` when the temporary extension is not used) is higher than `sizeof(newname)`, as otherwise any of those `g_snprintf` may overflow. It's unlikely but adding the check won't hurt. This is actually my fault as it was part of the original patch, but I guess fixing it here makes sense.

neilkinnish: Yes, just realised I need to recalculate the length based on the recording dir also?

record.c
neilkinnish: yes, apologies

lminiero:
Thanks, merging!

neilkinnish:
@lminiero there was an issue – I sent another commit, just before you merged, but it's not included.

lminiero:
Yep, I should have checked if it compiled first... :wink: 
Fixed in https://github.com/meetecho/janus-gateway/commit/1fa9d739e499ee1492814c9cb0cf8883a03b85bc

neilkinnish:
Great - thanks!

Thanks for being so responsive.

PR #821, opened 2017-03-17
MQTT username/password fix

stormbkk87:
Username/Password fix for MQTT transport


lminiero:
Isn't this leaking username and password? They're allocated and never freed.

stormbkk87:
My bad. Is this the best place to release? Not much of a C guy.

```
void janus_mqtt_client_destroy_context(janus_mqtt_context **ptr) {
	janus_mqtt_context *ctx = (janus_mqtt_context *)*ptr;
	if(ctx) {
		MQTTAsync_destroy(&ctx->client);
		g_free(ctx->subscribe.topic);
		g_free(ctx->publish.topic);
		g_free(ctx->connect.username);
		g_free(ctx->connect.password);
		g_free(ctx->admin.subscribe.topic);
		g_free(ctx->admin.publish.topic);
		g_free(ctx);
		*ptr = NULL;
	}

	JANUS_LOG(LOG_INFO, "%s destroyed!\n", JANUS_MQTT_NAME);
}
```

lminiero:
I think so (I'll have to check as I didn't author the plugin). Not sure if `janus_mqtt_client_destroy_context` can be called even on reconnect attempts, though? In that case, a reconnect attempt would find garbage where username/password are and break. Anyway, that shouldn't be the case, as `MQTTAsync_create` is only called in the init, and the `MQTTAsync_destroy` destroys that too, so credentials would be the last of the problems. I'll give this a better look when I get back to the office on Monday, but if you can confirm the PR fixes the credentials issue for you I'll note this as a +1 for merging!

Thanks again for the fix!

stormbkk87:
Yes, I've tested it and connection works and I also tested messaging and works.

lminiero:
Merging then, thanks!

PR #823, opened 2017-03-17
Attempt to provide re-based SIP plugin (WIP)

lminiero:
Hi all,

this is an attempt to provide a new SIP plugin as an alternative to the already available one. In fact, the one we have is based on Sofia SIP, which isn't the favourite choice for some people I've chatted with. As such, besides the effort on the NoSIP plugin (see #799  for details), I've decided to try and implement a new plugin that is based on [re](http://creytiv.com/re.html) instead, a popular library that provides stacks for several VoIP protocols. The idea is to keep the API from a user perspective pretty much the same as the existing plugin, which should allow for a very easy testing of both choices to see which one fits the use cases one has the best.

That said, unfortunately I haven't been very successful so far... after overcoming an initial problem with the loop that didn't seem to work for me (fixed, I believe, by simply invoking `poll_method_set(poll_method_best());` in the plugin initialization), I'm having quite a hard time figuring out what the best approaches to use the library are. Some considerations I've expressed [on their mailing list](http://lists.creytiv.com/pipermail/re-devel/2017-March/001177.html).

Now I've started integrating the new library in the new plugin (which is obviously stongly based on the existing `janus_sip.c`), but even the simplest procedure, registering, is failing for me, as it crashes deep within the code for reasons I can't explain. I tried extracting the code into a separate application and it works there, so I really don't know what may be going wrong, and am not familiar enough with the library. From how I can see it, I'm doing the register+authentication the way it should be done, but fresh eyes may spot problems I'm failing to see.

Rather than just give up, I decided to bring this to the community's attention and publish this broken PR anyway, as I believe this library has a great potential to help us improve things, not only on the SIP side. If you're somehow familiar with [re](http://creytiv.com/re.html) and want to help, please do! Even just nudging me in the right direction, e.g., clarifying what I may be doing wrong and where, would definitely help.

Thanks!


lminiero:
Hey guys, don't rush or you'll run over each other... :wink:

plugins/janus_sipre.c
```diff
@@ -0,0 +1,2893 @@
...
+			}
+			g_free(user_host);
+			g_free(user_id);
+
+			session->account.registration_status = janus_sipre_registration_status_registering;
+			if(send_register) {
+				char ttl_text[20];
+				g_snprintf(ttl_text, sizeof(ttl_text), "%d", ttl);
+					/* TODO Any way to specify TTL in sipreg_register? */
+				int err = sipreg_register(&session->stack.reg, sipstack,
```
alfredh: is this function called from the same thread that is running the re-main loop?

plugins/janus_sipre.c
```diff
@@ -0,0 +1,2893 @@
...
+			if(session->status != janus_sipre_call_status_invited) {
+				JANUS_LOG(LOG_ERR, "Wrong state (not invited? status=%s)\n", janus_sipre_call_status_string(session->status));
+				error_code = JANUS_SIPRE_ERROR_WRONG_STATE;
+				g_snprintf(error_cause, 512, "Wrong state (not invited? status=%s)", janus_sipre_call_status_string(session->status));
+				goto error;
+			}
+			if(session->callee == NULL) {
+				JANUS_LOG(LOG_ERR, "Wrong state (no caller?)\n");
+				error_code = JANUS_SIPRE_ERROR_WRONG_STATE;
+				g_snprintf(error_
```
alfredh: re_cancel must be called from the same thread as is running the re_main loop.
you can use struct mqueue to achieve this.

plugins/janus_sipre.c
lminiero: Nope, this is called by the plugin message handler thread.

When you say the same thread that is running the re-main loop, do you mean the thread that calls `re_main(..)`? We have that call here:

https://github.com/meetecho/janus-gateway/blob/f0c374b524c514d175496040bc7017ffebb25333/plugins/janus_sipre.c#L2778

but my understanding was that the `re_main(..)` call is blocking, and so the thread can't do anything else, which is why I'm devoting a separate thread just for that. Is this the wrong way to use the method?

plugins/janus_sipre.c
lminiero: Thanks, I'll look into it!

lminiero:
@alfredh using mqueue turned out to be surprisingly easy, thanks for the hint! Unfortunately it looks I'm still getting the same crash in the memory management...

https://gist.github.com/lminiero/19b065caf98c3d1bcacdfa8181084595

Do you have any insight on where I should look for a potential cause? Not sure what I'm doing wrong, as I verified I'm passing valid, allocated (and not freed ahead of time) strings when handling the authentication challenge, which is where it seems to be happening for some reason on the internal `struct mbuf`.

alfredh:
one more thing, you need to call `re_thread_init` and `re_thread_close` in the
re_main thread.

please look at this example in the `retest` testprogram:

https://github.com/creytiv/retest/blob/master/src/remain.c#L59

this is basically needed to create an instance of re_main for that thread,
otherwise you will have undefined behavior.

plugins/janus_sipre.c
```diff
@@ -0,0 +1,2934 @@
...
+	RAND_poll();
+#endif
+
+	/* Setup libre */
+	int err = libre_init();
+	if(err) {
+		JANUS_LOG(LOG_ERR, "libre_init() failed: %d (%s)\n", err, strerror(err));
+		return -1;
+	}
+	poll_method_set(poll_method_best());
```
alfredh: this should not be needed. if you have to set the polling-method explicitly, it indicates that
something is wrong.

it is done internally in libre main.c:

https://github.com/creytiv/re/blob/master/src/main/main.c#L539

also you need to call `sip_transp_add` from the re-main thread (using mqueue)
otherwise the system will poll the wrong file-descriptor.

lminiero: Thanks for these quick pointers, I'll fix those ASAP :+1:

lminiero:
Mh, updated the code and tried this but apparently using `re_thread_init()` causes the loop to never start working. If I uncomment it, the events are managed again (but it still crashes).

Considering we only have a single thread responsible for launching `re_main()`, do we really need `re_thread_init()`, though? Since we'd do everything from the main loop launched in that thread (and mqueues would ensure all requests would be executed in that context too, and not from other threads), there shouldn't be any thread-unsafeness. Or am I missing something?

alfredh:
I suggest you move the whole init sequence to the `janus_sipre_stack_thread`

Pseudo code:

```
thread()
{
    libre_init();

    re_thread_init();

    mqueue_alloc();

    sip_stack_init();

    re_main();

    re_thread_close();

    libre_close();
}
```

as a general principle, all RE functions should be called from the same thread.
this also includes mqueue_alloc which internally calls `fd_listen`.
fd_listen uses the current thread to look up the RE_MAIN instance. if this logic
is shared between multiple threads, nothing will work :)

the crash that you are seeing could be caused by data corruption.
if you are using Linux, perhaps run the whole process in Valgrind to check for
illegal memory access.

lminiero:
Thanks for the suggestion! Just updated the code to do that.

On the crash, yep, I had already done a valgrind check some time ago, and couldn't spot anything obvious, as it seemed to appear within libre callbacks. Just did a quick run again and the only problems I see seem indeed to be related to the crash, as they're related to `md5_printf` which is in the backtrace:

https://gist.github.com/lminiero/316d8d7ec28539702bbd94a5da88ed96

Specifically, it seems to happen when I return from the auth callback, so after I've provided the credentials to use to authenticate the REGISTER. This should result in the code to hash those credentials, the realm, etc., via `md5_printf`, which is where hell breaks loose.

From what I could remember trying to debug the libre internals at the time, it looked like the cause was a size that for some reason would become huge, and thus the realloc would crash (which might explain the `General Protection Fault`). I can't figure out what in my code could cause that, though, as the `struct mbuf` object that is used there is internal to those methods and not provided by the application, and the credentials are copied from valid strings via `str_dup` exactly as the examples I've seen do.

Is there anything that jumps at your eyes looking at the valgrind snippet? Of course I'm not asking you to debug this for me, but even just a hint on what may be breaking would be very helpful, as this looks like the last big obstacle to move on. After the IETF I'll try to dig deeper myself.

alfredh:
I had a look at the stack-trace and it looks indeed very strange:

are you by any chance using libusrsctp which also defines mbuf_init ?

```
int md5_printf(uint8_t *md, const char *fmt, ...)
{
	struct mbuf mb;
	va_list ap;
	int err;

	mbuf_init(&mb);    < ------- !

	va_start(ap, fmt);
	err = mbuf_vprintf(&mb, fmt, ap);
	va_end(ap);

	if (!err)
		md5(mb.buf, mb.end, md);

	mbuf_reset(&mb);

	return err;
}
```

if mbuf_init is not done here, it can explain the crash.

I have seen with other projects that there was a conflict with libre mbuf_init
and usrsctp mbuf_init. The latter project should have renamed it by now.

lminiero:
Thanks for looking into this! Yes, I'm indeed using `usrsctp`, for datachannels support. I haven't updated it in a while so it may very well be that I'm still using a version where the symbol is still available. Let me try...

lminiero:
YES! That was it! @alfredh you sir are truly a saviour, and have a beer promised in Berlin :smile:

notedit:
i do not use sip plugin,  but want to know if libre will be used in other plugin.

lminiero:
@notedit there's no plan to use it in other plugins, did you have any idea? In the future I may look into its ICE stack to see if it can fit the Janus core, though, as a possible alternative to libnice.

notedit:
libre seems promising,   i am studying this lib.  hope to use it in my personal project.

lminiero:
@alfredh sorry to bother you again, but I've been bashing my head the last few days trying to understand a couple of things that don't seem to be working the way I thought they would. Please let me know if these are questions I should ask on the `re-devel` mailing instead.

I did get both outgoing and incoming calls to work, somehow, but incoming calls only work for a short time. The cause apparently is that libre seems to be ignoring/dumping the ACK from the peer, and so the peer keeps on sending the 200 OK over and over. As a result, the peer eventually gives up and hangs up the call due to a timeout. The ACK is definitely there, and I'm doing all testing locally, so not sure why it's ignored... the only thing that comes to mind is that I had to figure out the best way to handle provisional responses, as `sip_treply` couldn't be used (by looking at the code, it needs a valid transaction object to work when code<200, and I don't think I have it). As such, following the example from some other code I found online, I'm now using `sipsess_accept` to send a "180 Ringing" back when I get an incoming call: the 1xx response is needed as I then forward the call to the web page, and it's up to the web user to decide whether to accept or deny the call; as such, I need to inform the peer that I got the request and I'm taking care of it. If the web user accepts, I do a new `sipsess_accept` on the same `struct sipsess` object with a 200 OK and the SDP to return back. Not sure if this double `sipsess_accept` is what is causing the issue, possibly resulting in a SIP dialog confusion? I remember checking the messages and the call-id looked the same, but I may have missed something. In case that's not the proper approach, what is the suggested way to handle this process? Possibly a different method to send the SDP back, as we already accepted the connection before?

Another thing I'm having trouble understanding is how to teardown a session. There doesn't seem to be any way to issue a BYE. I did find a `sipsess_terminate` request in the headers, but it seems to be a private interface that is not exposed to end users. I tried using `sipsess_close_all` but it seems not to have any visible effect (my callbacks would still be notified). Looking around it seems that some simply `mem_deref` the `struct sipsess` object, which is what I'm now doing as well, but this doesn't seem to result in a BYE: from what I could see doing some dumps, it seems to result in a `486 Busy` whether the call has started or not. Is there some API I'm not aware of to send a regular BYE instead?

Thanks in advance for any hint you can give me to head in the right direction!

lminiero:
I think I finally sorted out most of the issues I mentioned in my previous post: the problem was most likely caused by me invoking `sipsess_accept` more than once, instead of just doing it the first time, and then call `sipsess_progress`, `sipsess_answer` or `sipsess_reject` after that. This way, calling `mem_deref` on the session object does indeed generate a BYE and not an error code as I was experiencing (there I was getting a 486 because for the SIP stack the call was indeed not up and running yet, so it made sense).

I still have some things to sort out, code to clean and features to add, but this now at least works as expected for basic stuff (call, be called), so that's a good step forward!

lminiero:
Note to self: the fix made in #885 will need to be integrated in here as well.

lminiero:
Made several changes to the plugin, and now it's very close, in terms of functionality, to the Sofia-based plugin that is in master. The only thing missing is the ability to pass a HA1 secret to do a `HA1+Digest`, instead of a plain one: couldn't find a way to do the same with re, so for now I've disabled the setting in the web demo.

Feedback welcome, as at this stage I'll probably merge soon.

plugins/janus_sipre.c
```diff
@@ -0,0 +1,3832 @@
...
+				error_code, error_cause, TRUE,
+				JANUS_SIPRE_ERROR_MISSING_ELEMENT, JANUS_SIPRE_ERROR_INVALID_ELEMENT);
+			if(error_code != 0)
+				goto error;
+			json_t *uri = json_object_get(root, "uri");
+			/* Check if we need to ACK manually (e.g., for the Record-Route hack) */
+			json_t *autoack = json_object_get(root, "autoack");
+			gboolean do_autoack = autoack ? json_is_true(autoack) : TRUE;
+			/* Check if the INVITE needs to be enriched with custom headers */
+			char custom_h
```
sschmidt: I think we should (optionally/configurable) pass a DNS server as second argument for `sip_alloc`. When passing `NULL` only connections to an fixed IP are supported.

lminiero: Are you sure that's the case? I thought that omitting that would simply fallback to the system resolving functionality. Besides, the second argument is not a DNS server, but a libre DNS client instance, and not sure how that's supposed to work.

sschmidt: Pretty sure, I think... :)

When I use a domain name as DNS server with this branch/PR, I'm always receiving "22 invalid parameter" from the register method. I traced that down:

The error comes from RE's DNS Client.. before invoking a DNS query, the client checks if dnsc is set, i.e. here: https://github.com/creytiv/re/blob/30a963b8748f16febdec78e49d92b522875dba69/src/dns/client.c#L753

That method is invoked when the library tries to send the request to the destination server, i.e. in the addr_lookup-method here: https://github.com/creytiv/re/blob/30a963b8748f16febdec78e49d92b522875dba69/src/sip/request.c#L552 which will be invoked in the `sip_request` method in the same file. The `sip_request` call is invoked from the INVITE command, using the `dnsc` from `session->stack.sipstack` which we set to NULL during alloc. 

If I try to connect to the server using the IP, that call won't get invoked and the connection will be established.

Shoot me a mail if I can be of any help debugging that. I can send you some dumps or access to my test environment.

I think the `dnsc` is just a struct which we can "create", but haven't succeeded in doing so yet.

lminiero: Thanks for the detailed analysis! I'm glad someone is looking into this, as I'd really like to merge this plugin soon (currently working on some other features it still lacks from the other SIP plugin), but without proper feedback it would be hard to see if this is actually in a usable state or not.

As to `dnsc`, we should probably check how baresip creates and handles it.

sschmidt: Good idea.
We've tested the PR with a callcenter software of us. I think the two points we found is the DNS issue stated here and there is a delay once someone answers a call. People tend to say their name multiple times and get confused, because the connection is "up but not really there" :) Don't know how to better describe it. For example, if you call 08009377546, which is just a service that reads out your calling number, you won't here the first digits but join the call slightly later. Haven't found the time to analyze that issue yet, just wanted to point it out already.

lminiero: That's probably because, while on the SIP side media can start flowing right away, as soon as the answer arrives, on the WebRTC side of things takes longer: when the answer gets to the browser, ICE connectivity checks start flowing, and after that the DTLS handshake needs to take place. Nothing that can be done there, I'm afraid. I think the same happens with the Sofia-based SIP plugin as well, as the same approach is followed.

sschmidt: I'm not that deep into the field right now, but isn't ICE connectivity checks a thing which we can do prior to establishing the call, i.e. https://tools.ietf.org/html/rfc5245#section-12.1.1 ?
Would be willing to jump in and help with this work if you see a way to fix this issue. At least it feels good that the RFC chapter starts with "To reduce post-dial delays" :) Do you have an alternative idea how we could reduce the disorientation? e.g. send a tone to the callee as long as we haven't established the connection webrtc-side?

lminiero: No, you need the answer available, as otherwise you don't know the ICE ufrag and password the peer uses, and ICE authentication would fail. Same for DTLS handshakes (the answer has the fingerprint).

lminiero:
@sschmidt just committed a change that should add a DNS client when allocating the stack. Haven't tested it, though, so you may want to check if that indeed works as expected.

alfredh:
libre is using its own resolver (the `dns` module) and does not use any operating
system functions to resolve. It is only using the operating system to fetch a list of
DNS servers.

you need to create a (global) instance of `struct dnsc` using `dnsc_alloc` and then
pass it into the SIP stack creation.


/alfred

lminiero:
@alfredh yep, that's what I ended up doing yesterday, and I verified it worked in another setup. Thanks for confirming that was the way to go! I'm using a `struct dnsc` instance per user, though, not a global one (I followed [this example](https://github.com/alfredh/pylibre/blob/master/src/sip.c#L81) which acted on a "self", and I assumed a per-context instance) , so I may have to change that.

lminiero:
I think we're finally very close to merging this... basically almost all of the features available in the SIP plugin are available here as well, and I've managed to test the SIPre plugin in many scenarios successfully already!

There are a few things that are still missing, though, and while some are not really that important, a couple seem to be. @alfredh, could you please give me some feedback on the points below, to see if they're possible with re? Let me know if you'd rather I wrote this on the re mailing list instead.

1. We have a configuration property called `keepalive_interval` in the Sofia plugin, which basically triggers some keepalives via OPTIONS requests on a regular basis when registering. By looking at the code, it seems like re should do this automatically, so we probably don't need this.
2. I still found no way to tell re I'm behind a NAT, though... the Contact header seems to always use a private address. Is there any way to tell re to use a STUN request for the Contact header, or a way to use rport when registering?
3. With the other plugin we also have the possibility to do an HA1-based authentication (the password is already hashed in JavaScript by the client). Is this possible with re? This is something I don't really care about, as I was never able to use it myself, but not sure if there are people out there using it instead.
4. What's the proper way to cancel a request? To make a practical example, with the current implementation, if I do a REGISTER to which I don't get an answer, re automatically retries for about a minute before giving up. If I have my user exit, I still see the retries going on, although I do a `mem_deref` on the `struct sipreg` instance, and after a few seconds even a `sip_close` on the `struct sip` instance the register client is associated with (forcing the transactions to interrupt). Eventually this results in a callback being invoked with an argument pointer that is now invalid (user long gone), and can at times result in a crash. Anything I'm forgetting to do here?

Thanks!

lminiero:
Just as the NoSIP branch, this effort has been around some time already. While it's not perfect yet, and it's known to cause issues under some circumstances, I don't have time to work on it right now, and so I'd rather merge it as it is and fix it later. Hopefully this will result in more feedback and contributions from users, to improve how it currently works.

PR #834, opened 2017-03-28
add ssl support for the rabbitmq transport

flaviogrossi:
This pull request adds ssl support for the rabbitmq transport.

It introduces a few new configuration keys (only taken into consideration if ssl is enabled) and uses them when doing the amqp socket setup.

### DIFF
--- a/conf/janus.transport.rabbitmq.cfg.sample
+++ b/conf/janus.transport.rabbitmq.cfg.sample
@@ -25,6 +25,13 @@ host = localhost			; The address of the RabbitMQ server
 ;vhost = /					; Virtual host to specify when logging in, if needed
 to_janus = to-janus			; Name of the queue for incoming messages
 from_janus = from-janus		; Name of the queue for outgoing messages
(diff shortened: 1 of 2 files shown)


lminiero:
Cool, thanks for working on this! I'm abroad for the whole week so won't be able to review until I get back, though. In the meantime, can you sign our [CLA](https://cla.conf.meetecho.com/janus/)?

flaviogrossi:
done

lminiero:
Looks good!

certs and key files seem to only be freed in case of issues, though (when we return -1), otherwise they're never deallocated. Is this done by the library, or should we do it? If so, they should either be global, or (maybe better) part of the `janus_rabbitmq_client` structure: we do the same, for instance, for the names of the queues we'll use, which we `g_strdup` when parsing. This way, when getting rid of the client, we can deallocate certs and keys as well.

Anyway, looking at the code, I just noticed we're not doing that for the existing `rmq_client` properties either, which means this is something I have to fix, and not you... merging, thanks again for your contribution!

PR #838, opened 2017-04-01
Chrome-only System Audio w/ Screenshare (Tab audio only for Linux/Mac, Tab or Desktop audio on Windows)

chadfurman:
Needs testing:

1. Cross-browser compatibility.  Not sure if I broke anything in Firefox/Safari/Edge/iPhone etc
2. Making sure nojquery version matches -- I can't just "copy/paste" the whole file, but I think I brought in the important bits.
3. Indenting -- I've had trouble with this in the past.  Looks like I got it, and even improved it in some places.

*Note* If `isAudioSendEnabled()` evaluates to "true", audio will be included in the screenshare prompt.  I had to use "webkitGetUserMedia" instead of the `navigator.mediaDevices.getUserMedia` because the latter was causing an error when you selected an item to share which didn't have an audio stream (i.e. Linux desktop, or even uncheck the saftey "share audio" box in the Chrome UI) whereas the `webkitGetUserMedia` handled that case fine.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1395,7 +1395,7 @@ function Janus(gatewayCallbacks) {
 			}
 			var videoSupport = isVideoSendEnabled(media);
 			if(videoSupport === true && media != undefined && media != null) {
-				if(media.video && media.video != 'screen' && media.video != 'window') {
+				if(media.video && !isScreenshare(media)) {
 					var width = 0;
 					var height = 0, maxHeight = 0;
 					if(media.video === 'lowres') {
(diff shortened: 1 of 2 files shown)


chadfurman:
referenced here; https://groups.google.com/forum/#!topic/meetecho-janus/9NVhNF_rRqE

html/janus.js
```diff
@@ -2238,4 +2275,18 @@ function Janus(gatewayCallbacks) {
...
+
+	function isScreenshare(media) {
+		return (
+			media.video === 'screen' ||
+			media.video === 'window' ||
+			media.video === 'tab' ||
+			(
+				typeof media.video === "object" &&
+				media.length &&
+				(( media.indexOf('screen') + media.indexOf('window') + media.indexOf('tab')) > -3)
```
lminiero: I don't understand this check: `media` is an object, why would you check its `indexOf`?

chadfurman: I think this was supposed to check media.video.length and media.video.indexOf() so that you could pass in an array of "window/tab/screen".  However even if you did pass in an array here I don't think it would change anything.  So I'm going to remove this conditional

chadfurman: @lminiero ^

lminiero:
Closing as janus.js has wildly changed since when this PR was created. Please feel free to create a new one if you believe it can still be relevant.

PR #841, opened 2017-04-04
open RTCP port for RTSP streams

### LINKED ISSUE #832: RTSP streaming stops after a few seconds (no RTCP)
Hi,
I'm trying to re-stream with janus the output of an NVR.
The session janus-nvr is started correctly and the timeout is set to 60 seconds, however the session is closed by the nvr after a few seconds (about 10).
The problem seems to be that the nvr wants to communicate via **RTCP** but janus does not.
So, during the SETUP phase, janus reports 2 client ports (one for RTP and one for RTCP I suppose), but thet it does not listen on the second port.
Here's the SETUP message:
* janus -> nvr
```
SETUP rtsp://192.168.1.103/cam/realmonitor?channel=1&subtype=0/trackID=0 RTSP/1.0
CSeq: 3
Transport: RTP/AVP;unicast;client_port=45795-45796
Authorization: Basic YWRtaW46YWRtaW4=
```

* nvr -> janus
```
RTSP/1.0 200 OK
Server: Rtsp Server/2.0
CSeq: 3
Session: 13801393866;timeout=60
Transport: RTP/AVP;unicast;client_port=45795-45796;server_port=20000-20001;ssrc=66A362C2
x-dynamic-rate: 1
```
After 5 seconds the nvr sends an RTCP **Sender Report** message to port 45796 but the response is an ICMP **Destination port unreachable**

In VLC the SETUP is analogous but after 2 seconds (then every 5 seconds) VLC sends an RTCP **Receiver Report** message, after 5 seconds (then every 5 seconds) the NVR sends a **Sender Report** message.

I can share dumps if needed.

Thanks,
Fabrizio

fbertone:
this opens the rtcp port for RTSP streams (rtp port+1)
close #832


fbertone:
please note that this changes the return type of janus_streaming_rtsp_parse_sdp (shouldn't be a problem as it is used only for rtsp)
my editor messed a bit with whitespaces, sorry about that

lminiero:
Thanks for working on this I'll review in more detail as soon as I can.

> please note that this changes the return type of janus_streaming_rtsp_parse_sdp (shouldn't be a problem as it is used only for rtsp)

From a quick look, I don't think the return type should be changed, also considering you allocate a `multiple_fds` struct and never free it. I'd modify the signature of the method to add an input/output `multiple_fds` pointer, and still return an int, e.g.:

```int janus_streaming_rtsp_parse_sdp(const char *buffer, const char *name, const char *media, int *pt, char *transport, char *rtpmap, char *fmtp, char *control, const janus_network_address *iface, multiple_fds *fds) {```

that you can call with a:

```
multiple_fds fds = {0, 0};
janus_streaming_rtsp_parse_sdp(...., &fds);
```

> my editor messed a bit with whitespaces, sorry about that

Actually I think it fixed some extra whitespaces we had :smile:

**Comment by fbertone** [comment]
> you allocate a multiple_fds struct and never free it.

uh, you're right, I forgot it, I'm just closing the socket in

```
	#ifdef HAVE_LIBCURL
	if(source->audio_rtcp_fd > 0) {
		close(source->audio_rtcp_fd);
	}
	if(source->video_rtcp_fd > 0) {
		close(source->video_rtcp_fd);
	}
	#endif
```

sure, modify as you think it's better

**Comment by fbertone** [comment]
Also, I don't handle the case if the socket could not be opened.
Since it is just a "quick and dirty" fix we can just not care about it, but maybe at least log a warning.

To do a better job this should be somehow handled because if the socket can not be opened, this probably means that the port is already allocated to another process, and during the SETUP we are sending to the RTSP server a port that belongs to someone else, which is not very nice.

However this should be quite rare, so yeah, let's just ignore it for now :)

**Comment by lminiero** [comment]
Makes sense, and quick&dirty works for me :wink: 
As soon as you can address my previous comment I'll merge, thanks!

**Comment by lminiero** [comment]
Merging, thanks!

**Comment by lminiero** [comment]
One weird thing I just noticed though... as local ports we bind to, we're using the ports we parse from the remote SDP, instead of picking random ones ourselves to then advertize back? That doesn't sound right... or am I misunderstanding something?

**Comment by fbertone** [comment]
I've noticed that, I can tell you that in my case the remote SDP indicates port 0, the result apparently is that the port is actually picked randomly.
I haven't dug too much on this detail so I don't really know for sure, but never had problems related to this so far, so I guess it works fine.

**Comment by lminiero** [comment]
It's still puzzling, as if the remote party specifies something, then this will cause problems. I'll have to force 0 as a port. We may want to set 0 as a port for RTCP as well: this will end up with RTCP not being RTP+1, but if the client is not making assumptions that should work.

**Comment by fbertone** [comment]
I think it's better to check the RFCs, maybe the client is actually required to use the specified port?
Dunno, that would seem strange.
I'll take a look at the specifications.

**Comment by fbertone** [comment]
check [rfc 2326](https://tools.ietf.org/html/rfc2326#page-12), it states that:
```
Besides the media parameters, the network destination address and
   port need to be determined. Several modes of operation can be
   distinguished:

   Unicast:
          The media is transmitted to the source of the RTSP request,
          with the port number chosen by the client. Alternatively, the
          media is transmitted on the same reliable stream as RTSP.

   Multicast, server chooses address:
          The media server picks the multicast address and port. This is
          the typical case for a live or near-media-on-demand
          transmission.

   Multicast, client chooses address:
          If the server is to participate in an existing multicast
          conference, the multicast address, port and encryption key are
          given by the conference description, established by means
          outside the scope of this specification.
```

**The media server picks the multicast address and port.**
Seems to be the case, what do you think?
I'm also checking the newer v2 [rfc 7826]()

**Comment by fbertone** [comment]
In [rtsp 2.0](https://tools.ietf.org/html/rfc7826#appendix-D.1.2), 
```
If the session is over multicast, the port number
indicated SHOULD be used for reception.  The client MAY try to
override the destination port, through the Transport header.  The
servers MAY allow this: the response will indicate whether or not
this is allowed.  If the session is unicast, the port numbers are the
ones RECOMMENDED by the server to the client, about which receiver
ports to use; the client MUST still include its receiver ports in its
SETUP request.  The client MAY ignore this recommendation.  If the
server has no preference, it SHOULD set the port number value to
zero.
```
So, since we are not really using a full standard implementation I think we can use whatever port we want.
I guess that the session would be almost always unicast?

**Comment by lminiero** [comment]
I think so, yes, which means the ports are chosen by us (the RTSP client) and not the server, which still means using the port we parsed from the server's message is weird. Anyway, not a high priority thing to fix but I'll keep that in mind. Thanks for looking through the documents!

**Comment by fbertone** [comment]
No problem.
I can fix it if you like more the idea of randomly choose both ports.

**Comment by lminiero** [comment]
If that's a quick fix that doesn't take you much time that would help, thanks! Especially because right now the RTCP=RTP+1 thing may end up sending stuff to ports we use on the WebRTC side of things. Otherwise I can fix that in the next few days myself.

**Comment by fbertone** [comment]
Done, but I have to further investigate the thing because the result is not what I expected.
Here's a session with the new code:
* SETUP sent by Janus
```
SETUP rtsp://192.168.0.10/video.pro1/trackID=0 RTSP/1.0
CSeq: 3
Transport: RTP/AVP;unicast;client_port=53853-57228
Authorization: Basic YWRtaW46YWRtaW4=
```
as you can see, the ports are random.
* and, here's the REPLY:
```
RTSP/1.0 200 OK
CSeq: 3
Date: Wed Apr  5 15:41:43 2017 GMT
Session: 7b34dfe83cd22b79;timeout=80
Reconnect: true
Transport: RTP/AVP;unicast;client_port=53853-53854;server_port=5286-5287
```
as you can see, the server picks the correct port for RTP but RTCP is set to RTP+1 :/
Now, I don't know if it's this server that is buggy and does not implement RTCP itself (i tested with a different cam, it's not the other NVR that required the RTCP port to be open) or if we are supposed to always use adjacent ports.
I'll do more research tomorrow.

**Comment by fbertone** [comment]
Damn, this RTSP is a headache, check this out: 
* SETUP
```
SETUP rtsp://192.168.1.103/cam/realmonitor?channel=1&subtype=0/trackID=0 RTSP/1.0
CSeq: 3
Transport: RTP/AVP;unicast;client_port=41869-47333
Authorization: Basic YWRtaW46YWRtaW4=
```
* REPLY
```
RTSP/1.0 200 OK
Server: Rtsp Server/2.0
CSeq: 3
Session: 2074984567;timeout=60
Transport: RTP/AVP;unicast;client_port=41869-47333;server_port=20000-20001;ssrc=37F63243
x-dynamic-rate: 1
```
So, apparently the server accepted the ports proposed.
But where do you think it will send the RTCP Sender Report?!
```
User Datagram Protocol, Src Port: 20001, Dst Port: 41870
```
on port RTP+1!!!
Obviously port RTP+1 is closed so it terminates the stream :/
Maybe those ports are supposed to be a range and not two unique ports?
Or most probably everybody is implementing his own interpretation of RTSP and there is really no standard out there :/
Anyways, seems clear that we have to stick with port and port+1.
How would you handle the binding error on port+1?
Error and drop the whole stream or just ignore it and still use port+1 in the SETUP?

lminiero:
Yeah, hard to deal with softwares that make these assumptions... but that was true for many softphones and VoIP servers up to not that much time ago, so is unfortunately a known problem one has to leave with. I guess they'll also expect the RTP port to be even, and RTCP odd, or bad things may happen as well otherwise.

Considering it's important that the RTCP port is reachable and it doesn't mess with other stuff, I think the only reasonable approach is that if RTP+1 fails, we have to drop the RTP port we created before and just try again with a new RTP port, hoping that this time RTP+1 will work. Giving up the whole stream entirely seems overkill, considering we have the freedom to choose the port ourselves.

PR #844, opened 2017-04-06
use random rtp port for rtsp streams

fbertone:
As discussed in #841, this picks a random couple of ports <RTP, RTP+1> instead of use the port proposed by the server in the remote SDP


fbertone:
I'm a bit afraid that this will break streams using multicast (see #841 for references) but it should be easy to add again a simple check and use the port indicated by the server

lminiero:
Thanks for working on this so quickly, you really adopted the RTSP stuff! :smile: 

Not really sure how many people use the RTSP support plus multicasting, actually: I believe multicast support was added mainly for the plain RTP mountpoints, but yeah, you raise a valid points... we'll worry about this later!

I'll try to review ASAP :+1:

lminiero:
Looks good to me... merging, thanks!

PR #846, opened 2017-04-06
Added git commit + compile time information to the Janus logs

lminiero:
For a while I wanted to add information like which revision of Janus a specific instance was built against, or when it was built, as it's stuff that is always useful to have in the logs. I found a very useful suggestion [here](http://stackoverflow.com/a/1843783) and so implemented it, which means that now, with this patch, everytime you start Janus you should get something like this:

```
Janus commit: 10eea99203f30cd74b058f7b542439a6e03a8d59
Compiled on:  Thu  6 Apr 16:17:26 CEST 2017
```

I've published it as a pull request rather than on master directly as I want to make sure it works as expected not only on other systems (e.g., Mac OS), but also on other Linux distros besides mine (Fedora 25). The part I'm worried the most about is the way we generate the `version.c` file:

```
version.c: FORCE
	`which git` rev-parse HEAD | awk 'BEGIN {print "#include \"version.h\""} {print "const char *janus_build_git_sha = \"" $$0"\";"} END {}' > version.c
	date | awk 'BEGIN {} {print "const char *janus_build_git_time = \""$$0"\";"} END {} ' >> version.c
```

as I'm not sure the \`which git\` part fares well everywhere. Besides, both lines are probably very much bash-oriented. Can you guys please test and let me know? If you have suggestions on how to improve/fix in case I'd very much welcome it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 cmdline.c
 cmdline.h
+version.c
 docs/html/
 
 /janus
--- a/Makefile.am
+++ b/Makefile.am
@@ -128,6 +128,8 @@ janus_SOURCES = \
 	turnrest.h \
 	utils.c \
 	utils.h \
+	version.c \
+	version.h \
 	plugins/plugin.c \
 	plugins/plugin.h \
 	transports/transport.h \
@@ -152,13 +154,20 @@ janus_LDADD = \
 
 man1_MANS = janus.1
 
-BUILT_SOURCES = cmdline.c cmdline.h
+BUILT_SOURCES = cmdline.c cmdline.h version.c
 
 cmdline.c: janus.ggo
 	gengetopt --set-package="janus" --set-version="$(VERSION)" < $^
 
+version.c: FORCE
+	`which git` rev-parse HEAD | awk 'BEGIN {print "#include \"version.h\""} {print "const char *janus_build_git_sha = \"" $$0"\";"} END {}' > version.c
+	date | awk 'BEGIN {} {print "const char *janus_build_git_time = \""$$0"\";"} END {} ' >> version.c
+
+.PHONY: FORCE
+FORCE:
+
 EXTRA_DIST += janus.ggo
-CLEANFILES += cmdline.c cmdline.h
+CLEANFILES += cmdline.c cmdline.h version.c
 
 ##
 # Transports
--- a/janus.c
+++ b/janus.c
@@ -26,6 +26,7 @@
 #include <poll.h>
 
 #include "janus.h"
+#include "version.h"
 #include "cmdline.h"
 #include "config.h"
 #include "apierror.h"
@@ -3135,6 +3136,9 @@ gint main(int argc, char *argv[])
 	struct rlimit core_limits;
 	core_limits.rlim_cur = core_limits.rlim_max = RLIM_INFINITY;
 	setrlimit(RLIMIT_CORE, &core_limits);
+
+	g_print("Janus commit: %s\n", janus_build_git_sha);
+	g_print("Compiled on:  %s\n\n", janus_build_git_time);
 
 	struct gengetopt_args_info args_info;
 	/* Let's call our cmdline parser */
--- a/version.h
+++ b/version.h
@@ -1,0 +1,19 @@
+/*! \file   version.h
+ * \author Lorenzo Miniero <lorenzo@meetecho.com>
+ * \copyright GNU General Public License v3
+ * \brief  Janus GitHub versioning (headers)
+ * \details This exposes a quick an easy way to display the commit the
+ * compiled version of Janus implements, and when it has been built. It 
+ * is based on this excellent comment: http://stackoverflow.com/a/1843783
+ *
+ * \ingroup core
+ * \ref core
+ */
+ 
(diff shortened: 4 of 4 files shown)


thehunmonkgroup:
Isn't `which git` just going to look in $PATH to find it?

it does work on my Sierra machine:
```
which git
/usr/local/bin/git
```
and Bash is the default shell on OS X.

however, running those commands above from the root of the Janus git dir does throw an error on my OS X machine:

```
[hunmonk:~/git/janus-gateway] staging+ ± `which git` rev-parse HEAD | awk 'BEGIN {print "#include \"version.h\""} {print "const char *janus_build_git_sha = \"" $$0"\";"} END {}' > version.c
[hunmonk:~/git/janus-gateway] staging+* ± date | awk 'BEGIN {} {print "const char *janus_build_git_time = \""$$0"\";"} END {} ' >> version.c
awk: illegal field $(Mon Apr 10 11:53:14 PDT 2017), name "0"
 input record number 1, file 
 source line number 1
```

pretty sure OS X has a non-GNU awk going, which is probably causing the issue.

Running it on Debian 8 doesn't throw errors, but the commit hash is missing from the resulting file:
```
root@circleanywhere-video.stirlab.local$/var/local/git/janus-gateway: `which git` rev-parse HEAD | awk 'BEGIN {print "#include \"version.h\""} {print "const char *janus_build_git_sha = \"" $$0"\";"} END {}' > version.c
root@circleanywhere-video.stirlab.local$/var/local/git/janus-gateway: date | awk 'BEGIN {} {print "const char *janus_build_git_time = \""$$0"\";"} END {} ' >> version.c
root@circleanywhere-video.stirlab.local$/var/local/git/janus-gateway: cat version.c 
#include "version.h"
const char *janus_build_git_sha = "";
const char *janus_build_git_time = "Mon Apr 10 11:55:51 PDT 2017";
```

lminiero:
Thanks for the feedback! Yeah, I guess `awk` is not too reliable, due to the different implementations out there... I just copied the code suggested on that question, so I may have to change the way that info is extracted. The hard part will be finding something that consistently works everywhere...

lminiero:
PS: if you try those linees from bash directly, you will indeed get the empty commit in version.c, unless you replace `$$0` with `$1` (happens for me as well)... are you also getting the empty commit hash when doing a `make`, on Debian?

lminiero:
PPS: I guess the same applies to your Sierra machine too: it fails to evaluate `$$0` when you try it manually, while `$1` should work.

thehunmonkgroup:
yes that seems better. on Debian 8:
```
$/var/local/git/janus-gateway: cat version.c 
#include "version.h"
const char *janus_build_git_sha = "41f88c0f9151467e017f70b182b017b43b0bdaa1";
const char *janus_build_git_time = "Tue Apr 11 09:04:20 PDT 2017";
```
i've never built on OS X before, i get this error when i try with that branch:
```
[hunmonk:~/git/janus-gateway] commit-date+ ± make
CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/bash /var/local/git/janus-gateway/missing aclocal-1.14 -I m4
/bin/bash: /var/local/git/janus-gateway/missing: No such file or directory
make: *** [aclocal.m4] Error 127
```
BTW, not everybody's bash lives at /bin/bash!
```
[hunmonk:~/git/janus-gateway] commit-date+ 2 ± which bash
/usr/local/bin/bash
```
IIRC, that would also be a problem for FreeBSD installs.

lminiero:
We don't have that `/bin/bash` anywhere, so not sure where that's coming from. There are custom instructions for building Janus on OS X that I can't vouch for as I don't own one.

thehunmonkgroup:
all i did was go to the commit-date branch and type 'make', so i'm curious what generates that line that has /bin/bash in it, autotools?

hopefully you'll find somebody else to test the OS X build, it's not something that i want to spend any more time on.

lminiero:
> all i did was go to the commit-date branch and type 'make', so i'm curious what generates that line that has /bin/bash in it, autotools?

If you copied stuff from a Linux machine, you'd probably need to do a new `/.autogen` to regenerate the configure script, I guess.

> hopefully you'll find somebody else to test the OS X build, it's not something that i want to spend any more time on.

Ack, you already provided useful feedback, thx!

thehunmonkgroup:
ah, yes, autogen.sh fixed that issue.

tried again, but ran into missing deps. ah well :)

lminiero:
Well, since this seems to work fine on both Linux and OS X (thanks to @thehunmonkgroup for the feedback!), I'll merge. Looking forward to your hate mail when it doesn't compile for you :wink:

willmorgan:
> Well, since this seems to work fine

Not on CentOS 7. See #846 

```
which git` rev-parse HEAD | awk 'BEGIN {print "#include \"version.h\""} {print "const char *janus_build_git_sha = \"" $0"\";"} END {}' > version.c
fatal: Not a git repository (or any of the parent directories): .git
```

thehunmonkgroup:
@willmorgan are you compiling that from a git checkout of Janus? b/c that's a pretty weird error if not

willmorgan:
We're doing something a little special, see messages in #860

PR #847, opened 2017-04-06
Use cmake compile rabbitmq-c  in README

peterwillcn:
use cmake rabbitmq-c compile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,8 +203,9 @@ following steps:
 	cd rabbitmq-c
 	git submodule init
 	git submodule update
-	autoreconf -i
-	./configure --prefix=/usr && make && sudo make install
+	mkdir build && cd build
+	cmake -DCMAKE_INSTALL_PREFIX=/usr ..
+	make -j$(nproc) && sudo make install
 
 * *Note:* you may need to pass `--libdir=/usr/lib64` to the configure
 script if you're installing on a x86_64 distribution.


lminiero:
Are you sure this is needed? I think these are the instructions I used at the time and worked for me... Is this a recent change, that would not work with older versions? And (apologies for the probably very dumb question...), where does the `$(nproc)` come from?

peterwillcn:
Hi, lminiero

> The [nproc](http://www.gnu.org/software/coreutils/manual/html_node/nproc-invocation.html#nproc-invocation) command prints the number of processors (cores).

Not all distributions provide this command, but most (all?) of the ones that do incorporate this from the [GNU Coreutils](http://www.gnu.org/software/coreutils/manual/html_node/nproc-invocation.html#nproc-invocation)

Use autoreconf compile get this error

> librabbitmq/.libs/librabbitmq.so: undefined reference to `pthread_once'
autoreconf

User cmake compile successful
> cmake

lminiero:
@peterwillcn ack, thanks for the clarification. Could you sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

lminiero:
@peterwillcn eventually added the steps myself, but without the `nproc` part: since you mentioned it may not be universally available, and it's only used as a flag to `-j` to do simultaneous jobs when compiling, I simply removed it to keep it easier. Thanks for bringing this issue to our attention!

Issue #848, opened 2017-04-06
Possible leak with textroom datachannels 


Sean-Der:
Ever since I started using textrooms I have been having some memory issues, I believe I found the [culprit](https://hastebin.com/buboxovixe.pas)

This leak does grow over time, and always the same place. I am using Janus @ 786c7ffc12bf6190099cbf5ebafed06a6d331510 and usrsctp @ 2f6478eb8d40f1766a96b5b033ed26c0c2244589

If this isn't enough data I can try to get the most minimal reproduce, record data about the instances that are leaked. Also more then happy to try and fix it myself if you don't have time (but you are going to be WAY faster then me).

thanks!

Sean-Der:
And [here](https://hastebin.com/fibuxuluyo.pas) is another one with some more relevant lines

lminiero:
Since this is happening within the usrsctp library I'm not sure there's anything we can do. In the snippet you provided, this is caused by something called here:

`#9 0x4cd259 in janus_sctp_thread /home/ubuntu/src/janus-gateway/sctp.c:1296`

which is this line here:

`usrsctp_conninput((void *)sctp, message->buffer, message->length, 0);`

The only allocated stuff we provide ourselves is `message->buffer`, and we free it right after that call:

`janus_sctp_message_destroy(message);`

It may be a leak that AddressSanitizer improperly notifies about, possibly because of the way usrsctp handles its own memory?

lminiero:
Looking at the usrsctp code in the part where it complains about the leak:

`    #1 0x7f5d32f3a7a1 in sctp_invoke_recv_callback netinet/sctputil.c:4830`

it's related to this line here:

`if ((buffer = malloc(length)) == NULL) {`

where the buffer is eventually passed to a callback:

`inp->recv_callback(so, addr, buffer, length, rcv, flags, inp->ulp_info);`

Maybe this `recv_callback` is something that then feeds data to our code, and we're responsible for freeing the data? Not sure as I'm not that familiar with the usrsctp internals.

lminiero:
Just in case this is related to notifications about incoming data, can you add this line:

```if(data) free(data);```

before returning [here](https://github.com/meetecho/janus-gateway/blob/master/sctp.c#L333) and let me know if it fixes it for you?

**Comment by Sean-Der** [comment]
Hey @lminiero 

I think you got it perfectly! I just loaded up the textroom demo and sent a few messages with that free and now I have
```
=================================================================
[...]

PR #850, opened 2017-04-07
Fix leak in janus_sctp_incoming_data, callback used by usrsctp_socket didn't free passed data

Sean-Der:
Hey @lminiero this is the follow up from issue #848 

Feel free to close this PR if you planned on doing more work with it, just wanted to make this as easy as possible on you (hopefully)

thanks!


lminiero:
Please add a check to see if data is a valid pointer. AFAIK not all compilers return when free is done on NULL and might crash on it.

lminiero:
Whoops, my bad, it's in the `if(data)` already... merging, thanks!

PR #854, opened 2017-04-10
Support for on-hold in SIP plugin

lminiero:
This PR adds support for the on-hold feature to the SIP plugin. It basically allows you to put a call on-hold, and then resume it later on.

The way it works is that it sends a re-INVITE to the peer by asking them not to send anything anymore using the media direction. So, if the call was `sendrecv`, we'll ask for a `sendonly`; if the call was `sendonly`, we'll send `inactive`; and so on. When resuming, we simply restore the direction we had before putting on-hold. Notice that we do NOT set the c-line to `0.0.0.0`: while this was the proper way to put on-hold a long time ago, this is now discouraged, and the media direction is what should be used instead.

If you want to give this a try, the messages to use in the API are `hold` for putting on-hold, and `unhold` for resuming: nothing else is needed, as simple as that. Here are a couple of examples that would work in the SIP demo page:

_Put the current call on-hold:_

    sipcall.send({message: {request: "hold"}});

_Resume the call:_

    sipcall.send({message: {request: "unhold"}});

Please let us know if this works as expected for you as well, so that we can merge this soon.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -33,12 +33,13 @@
  * which means all responses (successes and errors) will be delivered
  * as events with the same transaction.
  *
- * The supported requests are \c register , \c call , \c accept and
+ * The supported requests are \c register , \c call , \c accept, \c hold , \c unhold and
  * \c hangup . \c register can be used, as the name suggests, to register
  * a username at a SIP registrar to call and be called; \c call is used
  * to send an INVITE to a different SIP URI through the plugin, while
  * \c accept is used to accept the call in case one is invited instead
- * of inviting; finally, \c hangup can be used to terminate the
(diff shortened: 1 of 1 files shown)


lminiero:
Got positive feedback in provate by somebody using it, so I'm going to merge.

shrhoads:
This is a good feature, but shouldn't the outgoing audio be paused as well and not just inbound when you put the call on "hold"?   The current "hold" stops inbound audio but keeps outbound (which is the opposite of what I am looking for, a "mute" feature, I still want to listen, but I want to mute my audio.

Do you think it makes sense to have a "mute" and "unmute" feature?

lminiero:
You can mute the local stream on the client side and that will just send silenze.

PR #859, opened 2017-04-12
Allow video/audio port of 0 for RTP streaming, also add 'audio_port' and 'video_port' to find the randomly generated port

Sean-Der:
This is a WIP, I am not super familiar with the streaming plugin so I don't know everything needed to make this change merge-able.

I am creating streaming mountpoints at runtime and don't want to have to worry about what port is chosen. This allows me to just set 0, and then query for the ports later.

thanks

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -182,6 +182,7 @@ void janus_streaming_incoming_rtcp(janus_plugin_session *handle, int video, char
 void janus_streaming_hangup_media(janus_plugin_session *handle);
 void janus_streaming_destroy_session(janus_plugin_session *handle, int *error);
 json_t *janus_streaming_query_session(janus_plugin_session *handle);
+static int janus_streaming_get_fd_port(int fd);
 
 /* Plugin setup */
 static janus_plugin janus_streaming_plugin =
@@ -690,7 +691,7 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 					continue;
 				}
 				if(doaudio &&
-						(aport == NULL || aport->value == NULL ||
+						(aport == NULL || aport->value == NULL || atoi(aport->value) == 0 ||
 						acodec == NULL || acodec->value == NULL ||
 						artpmap == NULL || artpmap->value == NULL)) {
 					JANUS_LOG(LOG_ERR, "Can't add 'rtp' stream '%s', missing mandatory information for audio...\n", cat->name);
@@ -710,14 +711,14 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 					}
 				}
 				if(dovideo &&
-						(vport == NULL || vport->value == NULL ||
+						(vport == NULL || vport->value == NULL || atoi(vport->value) == 0 ||
 						vcodec == NULL || vcodec->value == NULL ||
 						vrtpmap == NULL || vrtpmap->value == NULL)) {
 					JANUS_LOG(LOG_ERR, "Can't add 'rtp' stream '%s', missing mandatory information for video...\n", cat->name);
 					cl = cl->next;
 					continue;
 				}
-				if(dodata && (dport == NULL || dport->value == NULL)) {
+				if(dodata && (dport == NULL || dport->value == NULL || atoi(dport->value) == 0)) {
 					JANUS_LOG(LOG_ERR, "Can't add 'rtp' stream '%s', missing mandatory information for data...\n", cat->name);
 					cl = cl->next;
 					continue;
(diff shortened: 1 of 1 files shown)


lminiero:
Definitely a good idea, thanks for working on this and sharing it! In the future we might even envisage some kind of a port range for the Streaming plugin, so that one can constrain which are chosen when done randomly.

The only nit I have by looking quickly at the code is that `janus_streaming_get_fd_port` seems to assume the address will be IPv4, and that wouldn't work in case it's IPv6 instead. I guess the ip-utils stuff we merged recently might help in making this more flexible.

Sean-Der:
Stupid mistake on my part @lminiero sockaddr_in doesn't even need to be populated since we are just filling it with data from the fd. Amended the commit.

re: ip-utils, is this a task I could take on? I would love to help with Janus stuff outside of just what I use it for.

lminiero:
Thanks for the quick fix! I'll review quickly and let you know if we can merge as is :+1: 

Considering it doesn't really matter if it's IPv4 or IPv6, the ip-utils usage is not needed anymore. If you're just wondering if it might help in other contexts of for future PRs, I'd definitely give it a check, as it's a really nice contribution we received recently. Within the Streaming plugin, for instance, we use it to optionally only bind to specific interfaces/address, rather than all the available ones, which can be really useful in case you want to protect your mountpoint source. It's used in some other parts of the code as well, although not as much as it should.

Sean-Der:
Awesome!

I was also wondering if you had other tasks I could help out with. I would love to hang out on IRC/mailing list and fix bugs. Janus is a super enjoyable code base to work with.

lminiero:
There are several issues and PRs, so we do indeed have our hands full :smile: 
Any contribution like the one you just made, or fixes to bugs you or others find out about, are always more than welcome and appreciated!

lminiero:
@Sean-Der just noticed this only does the 0-port stuff for audio and video: can you add the same for data as well, since the Streaming plugin now supports it?

Sean-Der:
Well that was a bad oversight on my part, the data_fd is 2 lines below! Thanks for the review, pushed those changes.

lminiero:
Thanks for the quick fix!

Just found some time to test it and it seems to work fine, but I don't like the idea of `info` returning the port numbers... `info` is a public request, and this means that it exposes private mountpoint information, thus leading to potential abuse (whoever can now send the mountpoint data, unless you add a constraint on the interface).

I'd return the ports as part of the `create` response instead, as this way only who created the mountpoint has this information (pretty much as it happened before as well, where whocreated the mountpoint provided them in the first place). This would mean that setting port=0 from configuration file would not be allowed (you'd have no way to retrieve them after that) but it doesn't look like a big deal to me, as this port=0 thing is really a dynamic approach, and so belongs to the dynamic API.

As a side note, I noticed `audioport`, `videoport` and `dataport` are still required, which means that if you want to have them random, it's not enough to just omit them (as I thought), but you have to specify them and set to 0. Not a problem for me, just highlighting this as users will need to be aware of that!

Sean-Der:
Fixed! The create now returns the ports, info is unchanged and added assertions so that port != 0 when reading the cfg file.

I agree with you that random port should be explicit, I think it would be confusing for people if not passing the port would cause a random one to be generated (like if you have a script and the videoport key doesn't get set before sending to Janus that would be a pain to debug)

plugins/janus_streaming.c
```diff
@@ -691,8 +692,8 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 				}
 				if(doaudio &&
 						(aport == NULL || aport->value == NULL ||
-						acodec == NULL || acodec->value == NULL ||
-						artpmap == NULL || artpmap->value == NULL)) {
+						atoi(aport->value) == 0 || acodec == NULL ||
```
lminiero: Can you move the `atoi(aport->value) == 0` part a couple lines above, so that it's next to the `aport->value == NULL` check? And bring the two `acodec` checks on the same line again? It's cleaner from a code perspective.

plugins/janus_streaming.c
```diff
@@ -711,13 +712,13 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 				}
 				if(dovideo &&
 						(vport == NULL || vport->value == NULL ||
-						vcodec == NULL || vcodec->value == NULL ||
-						vrtpmap == NULL || vrtpmap->value == NULL)) {
+						atoi(vport->value) == 0 || vcodec == NULL ||
+						vcodec->value == NULL || vrtpmap == NULL || vrtpmap->value == NULL)) {
```
lminiero: Same as above.

plugins/janus_streaming.c
```diff
@@ -1815,6 +1816,18 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session
 		json_object_set_new(ml, "description", json_string(mp->description));
 		json_object_set_new(ml, "type", json_string(mp->streaming_type == janus_streaming_type_live ? "live" : "on demand"));
 		json_object_set_new(ml, "is_private", mp->is_private ? json_true() : json_false());
+		if(mp->streaming_source == janus_streaming_source_rtp) {
```
lminiero: While testing I noticed this activates when creating RTSP mountpoints too, which shouldn't happen as in that case it's the RTSP server that feeds those addresses automatically, and there's no need to spawn scripts yourselves. I think an easy way to fix that is to change the check you added to something like `if(!strcasecmp(type_text, "rtp")) {` instead, as this way we're sure it only triggers for plain RTP mountpoints. An alternative might be accessing the `source` property within this check, cast it to `janus_streaming_rtp_source`, and only return the ports if the `rtsp` boolean property is `FALSE` (notice that, in case, this check would need to be wrapped in a `#ifdef HAVE_LIBCURL`, as that boolean wouldn't exist otherwise).

Sean-Der:
Thanks again for the review! I fixed all the comments you had, feel free to amend my commit also doesn't bother me.

lminiero:
Noticed you added the `janus_streaming_rtp_source` part, but forgot to do the check on the `rtsp` property :smile: 
No problem, I'll merge and add it myself. Thx!

PR #870, opened 2017-04-26
Use MHD_USE_AUTO if libmicrohttpd is recent enough

lminiero:
This should fix an issue we had, where if libmicrohttpd was too new, the default configuration would not start an HTTP server in the HTTP plugin. The issue was first reported here:

https://groups.google.com/forum/#!topic/meetecho-janus/76dBFVhaCb0

The libmicrohttpd developers helped highlight what could be the issue:

http://lists.gnu.org/archive/html/libmicrohttpd/2017-04/msg00014.html

This PR tries to address that, so please test (works for me) as I plan to merge soon.

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -424,9 +424,9 @@ static struct MHD_Daemon *janus_http_create_daemon(gboolean admin, char *path,
 					admin ? "Admin" : "Janus", secure ? "HTTPS" : "HTTP");
 				daemon = MHD_start_daemon(
 #if MHD_VERSION >= 0x00095208
-					MHD_USE_TLS | MHD_USE_THREAD_PER_CONNECTION | MHD_USE_AUTO_INTERNAL_THREAD | MHD_USE_AUTO | MHD_USE_DUAL_STACK,
+					MHD_USE_SSL | MHD_USE_THREAD_PER_CONNECTION | MHD_USE_AUTO_INTERNAL_THREAD | MHD_USE_AUTO | MHD_USE_DUAL_STACK,
 #else
-					MHD_USE_TLS | MHD_USE_THREAD_PER_CONNECTION | MHD_USE_POLL_INTERNALLY | MHD_USE_POLL | MHD_USE_DUAL_STACK,
+					MHD_USE_SSL | MHD_USE_THREAD_PER_CONNECTION | MHD_USE_POLL_INTERNALLY | MHD_USE_POLL | MHD_USE_DUAL_STACK,
 #endif
 					port,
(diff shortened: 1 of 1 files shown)


mirkobrankovic:
AUTO instead of POL and TLS instead of SSL, seems legit to me ;)

lminiero:
Added some additional changes after some more clarifications from the MHD developers (who've been incredibly helpful!)

ruddell:
This works fine on Fedora, but when I try this PR on Ubuntu 16, I get the following error (note that before this PR it works fine):
### System Info:
```
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.2 LTS
Release:	16.04
Codename:	xenial

libmicrohttpd-dev:
  Installed: 0.9.44+dfsg-1ubuntu2
  Candidate: 0.9.44+dfsg-1ubuntu2
```

### Error:
```make
  CC       transports/transports_libjanus_http_la-janus_http.lo
transports/janus_http.c: In function ‘janus_http_create_daemon’:
transports/janus_http.c:429:6: error: ‘MHD_USE_TLS’ undeclared (first use in this function)
      MHD_USE_TLS | MHD_USE_THREAD_PER_CONNECTION | MHD_USE_POLL_INTERNALLY | MHD_USE_POLL | MHD_USE_DUAL_STACK,
      ^
transports/janus_http.c:429:6: note: each undeclared identifier is reported only once for each function it appears in
Makefile:1504: recipe for target 'transports/transports_libjanus_http_la-janus_http.lo' failed
make[2]: *** [transports/transports_libjanus_http_la-janus_http.lo] Error 1
make[2]: Leaving directory '/root/janus-gateway'
Makefile:2200: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/root/janus-gateway'
Makefile:1028: recipe for target 'all' failed
make: *** [all] Error 2
root@janus:~/janus-gateway# 
```

mirkobrankovic:
Make sure you checkout this branch before you do autogen.sh.
I think it is:
git checkout mhd_fix

ruddell:
Sorry I forgot that step in the commands to reproduce, just added it to the list.  Clarifying what I said above, master works fine.

mirkobrankovic:
@lminiero seems like the backwards compatibility is not ok for USE_MHD_TLS, since i see this is 9.44 libmicrohttpd

lminiero:
Good catch, I'll revert that change.

lminiero:
Reverted that change, which should fix things then. Merging as the rest should be working as expected.

PR #872, opened 2017-04-27
Added audio output device selection, if available, to devicetest demo

### LINKED ISSUE #869: Need a way to set "output" in demo
Unless my local copy is outdated, the device selection demo needs to include audio output device selection also.  It's different than input selection, as it's not a media constraint but a sink attached to each video element.

Edit: Just putting this here until I get around to making a PR for it.

lminiero:
Will fix #869, when merged. Notice that this will only work on Chrome >= 49: apparently it's not available anywhere else (not even Firefox). Couldn't test much as I don't have other output devices to choose from.

### DIFF
--- a/html/devicetest.html
+++ b/html/devicetest.html
@@ -67,11 +67,11 @@
 				<div class="row hide" id="devices">
 					<form class="form-inline">
 						<div class="form-group" style="margin-right: 20px;">
-							<label for="audio-device">Audio device:</label>
+							<label for="audio-device">Audio device (input):</label>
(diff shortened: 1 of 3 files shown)


lminiero:
What join? The devicetest demo uses the EchoTest, not the VideoRoom. Each time you change the input devices, the call is reset, and so is the output device selection.

chadfurman:
:+1: okay.  I'll give it a run later, but looks right

lminiero:
Merging, we can fix along the way if something's wrong.

PR #873, opened 2017-04-27
fix #871 detaching of already detached plugin

### LINKED ISSUE #871: Janus detaches already detached plugin on session destroy
When plugin got detached on the Janus side, front end javascript is calling plugin's detach method to the already detached plugin, which causes `not found` error. More information is here - https://groups.google.com/forum/#!topic/meetecho-janus/YU_wu9xMS4c

hijaq:
Unfortunately Lorenzo's commit 8274fe87fcd8a3fbdf4ef82934edcd41828cf007 doesn't fix that problem and still calls plugin's `detach()` method on already detached plugin. At the same time, there's a WebRTC cleanup procedure inside of `destroyHandle()`, so we can't just skip `detach()` call in handling of `detached` message, so I did a small fix that works solves the issue (tested it only with REST).

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -412,6 +412,7 @@ function Janus(gatewayCallbacks) {
 				// Don't warn here because destroyHandle causes this situation.
 				return;
 			}
+			pluginHandle.detached = true;
 			pluginHandle.ondetached();
 			pluginHandle.detach();
 		} else if(json["janus"] === "media") {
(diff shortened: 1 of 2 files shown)


lminiero:
Why the new property? Can't you just check if `pluginHandles[handleId]` exists and, if not, do the success callback automatically as you added?

lminiero:
Ah ok, got it, the value would still be in the hashtable so the check I suggested wouldn't work.

lminiero:
Looks good to me, although I'd move the check you suggested to before `if(!connected) {`, as it doesn't make sense to start building a request we'll never send anyway. The important thing is doing it after `cleanupWebRTC`, so before `if(!connected) {` would do the trick.

hijaq:
done

lminiero:
Thanks, merging then!

PR #878, opened 2017-05-01
Remove warning as post-processor now supports VP9

Joe-Palmer:


lminiero:
Good catch! Can you sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge?

lminiero:
Since the change was minimal, I removed the lines myself and committed. Thanks for spotting this!

PR #880, opened 2017-05-02
Fix: broken check for whether or not data channels are needed

cmacq2:
Fix: broken check for whether or not data channels were actually requested.

Previously, explicitly deconfigured data=no streams would trigger a false positive here if Janus was compiled without support for SCTP.


lminiero:

PR #883, opened 2017-05-03
fix TextRoom plugin name in launching handler error

fbertone:


lminiero:
This will teach me to be more careful when doing copy/paste in the future... :wink: 
Thanks, merging!

PR #884, opened 2017-05-04
Fix for OSX websockets not able send messages back after connect

yultide:
On OSX build, after connect the first callback LWS_CALLBACK_SERVER_WRITEABLE tries to call g_async_queue_try_pop() and returns a NULL.  Since lws_callback_on_writable() or libwebsocket_callback_on_writable() is not called in that condition, subsequent writes are not processed.

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -1272,6 +1272,11 @@ static int janus_websockets_common_callback(
 					janus_mutex_unlock(&ws_client->mutex);
 					return 0;
 				}
+#ifdef HAVE_LIBWEBSOCKETS_NEWAPI
+				lws_callback_on_writable(wsi);
+#else
+				libwebsocket_callback_on_writable(this, wsi);
+#endif
 				janus_mutex_unlock(&ws_client->mutex);
 			}
 			return 0;


lminiero:
Whoops, I completely missed this, sorry about that... before I can merge, can you sign our [CLA](https://cla.conf.meetecho.com/janus/)?

yultide:
Signed

lminiero:
Thanks! I've added your PR to a list I've just posted on the group, just to have some additional feedback on all the PR activity on the repo: https://groups.google.com/forum/#!topic/meetecho-janus/KoujdV0cqfA

I don't expect your PR to give any issue, and I plan to merge this really soon (I'll give it a check myself later), but the more feedback we can get to make sure it doesn't break anything the better!

lminiero:
@yultide  not sure we can merge this. Looked better at the code and this basically results in a "writable" loop. Normally, when you mark the connection as writable, you write what you have to, and only mark it as writable again if you didn't manage to write everything. In this case, though, we're always marking the connection as writable again even when there was nothing to write: this means we get in the "writable" event again immediately, which marks the connection writable again, and so on, causing a loop. Tested this locally and in fact I got a 100% CPU usage in top for Janus.

What was this trying to fix in your specific case?

lminiero:
Closing.

yultide:
Sorry.  Was on vacation and just got back.

I'm building on OSX with the v2.2-stable libwebsockets on OSX.  When I connect using websockets, the messages from the janus server are never sent back to the client.  It's probably a bug in the websockets library on OSX and I think the code in janus is correct.

BTW, I didn't get 100% cpu on the libwebsockets v2.2 on linux and mac, but I didn't test it with the older websockets library.

BHamrick1:
I can confirm that this makes janus function on OSX with websockets, but it also causes janus to run at 100% cpu and janus also does not shutdown cleanly anymore. I really need this to work as we have written all of our janus communication using websockets instead of the REST api.

lminiero:
I'm not experiencing this issue and I can't replicate it, so I'm afraid I can't help. If you can come up with a patch that solves this, a pull request is more than welcome of course.

BHamrick1:
Are you testing this on OSX? I am not seeing this issue on linux either. We are probably going to be running janus in vagrant to get around this issue.

lminiero:
I don't have OSX machines to test with.

bigboyaaa:
I'm testing onOSX,this bug still exist.In the latest code,how to solve this?

lminiero:
For a solution to this problem, please see [this discussion](https://groups.google.com/forum/#!topic/meetecho-janus/HsFaEXBz4Cg).

PR #886, opened 2017-05-11
New configure options to disable all plugins

lminiero:
This PR makes it simpler and cleaner to only enable a few plugins, rather than installing them all. In fact, so far if you wanted to only install the VideoRoom plugin and the HTTP transport and not build anything else, you had to do something like:

```
./configure --prefix=/opt/janus --disable-plugin-echotest --disable-plugin-streaming
    --disable-plugin-videocall --disable-plugin-sip --disable-plugin-audiobridge
    --disable-plugin-recordplay --disable-plugin-textroom --disable-plugin-voicemail
    --disable-rabbitmq --disable-mqtt --disable-unix-sockets --disable-websockets
```

Besides not being very readable, and being very easily error-prone, this would break when new plugins or transports are merged, as they'd not be listed explicitly as not to be built.

This new PR makes it much easier by adding new flags called:

    --disable-all-plugins
    --disable-all-transports
    --disable-all-handlers

As the names suggest, each of them disables compilation for all plugins of a specific category, meaning you can then simply enable those you're interested in. For the VideoRoom + HTTP scenario above, the configure line would become:

```
./configure --prefix=/opt/janus --disable-all-plugins --enable-plugin-videoroom
    --disable-all-transports --enable-rest
```

Played with this a bit and it seems to work fine, not interfering with the default behaviour and so on, but of course feedback is welcome in case you find it breaking anything it shouldn't. If nobody stops me, I'll merge this soon.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,47 +40,11 @@ AC_ARG_ENABLE([docs],
               [],
               [enable_docs=no])
 
-AC_ARG_ENABLE([rest],
-              [AS_HELP_STRING([--disable-rest],
-                              [Disable REST (HTTP/HTTPS) support])],
-              [],
-              [enable_rest=maybe])
-
-AC_ARG_ENABLE([websockets],
-              [AS_HELP_STRING([--disable-websockets],
-                              [Disable WebSockets support])],
-              [],
-              [enable_websockets=maybe])
-
 AC_ARG_ENABLE([data-channels],
               [AS_HELP_STRING([--disable-data-channels],
                               [Disable DataChannels])],
               [],
               [enable_data_channels=maybe])
-
-AC_ARG_ENABLE([rabbitmq],
-              [AS_HELP_STRING([--disable-rabbitmq],
-                              [Disable RabbitMQ integration])],
-              [],
-              [enable_rabbitmq=maybe])
-
-AC_ARG_ENABLE([mqtt],
-              [AS_HELP_STRING([--disable-mqtt],
-                              [Disable MQTT integration])],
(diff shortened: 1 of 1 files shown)


joshdickson40:
Makes sense to have your plugins/transports/etc configured this way so when new ones are landed (i.e. mqtt) they are automatically disabled unless you want to explicitly enable them. Especially useful for cases like mqtt that require additional packages.

Ideally this could also include some doc updates to suggest this usage :)

lminiero:
Notice that all plugins are already configured on a "dependency available" basis. All of them are optional, and only built in case the dependency is available. This means that if you don't have the MQTT library installed, the related transport plugin will be skipped. Of course some plugins don't depend on anything that Janus itself doesn't depend on, and so by default they're always built.

What this PR adds is an ability to fine tune what you really want built. For instance, why build the Record&Play or EchoTest plugins if you'll never use them? They'd just sit there in the memory doing nothing. Normally you'd explicitly disable all the plugins you don't care about, and enable those that you want. In case new plugins come up later (e.g., the NoSIP and SIPre plugins we're working on) you'd need to know you have to manually disable them as well, unless there's a way to disable everything and only enable what you care about, which is what we do here.

joshdickson40:
> Notice that all plugins are already configured on a "dependency available" basis. All of them are optional, and only built in case the dependency is available. This means that if you don't have the MQTT library installed, the related transport plugin will be skipped.

Didn't know this!

lminiero:
Looks harmless so merging.

Issue #751, opened 2017-01-17
crash keeps happening since version 0.1.1 and also on latest git checkout 


bhakimi:
let me know if you need anything else, this is the core dump 

```
[New LWP 22784]
[New LWP 34672]
warning: Could not load shared library symbols for linux-vdso.so.1.
Do you need "set solib-search-path" or "set sysroot"?
Warning: couldn't activate thread debugging using libthread_db: Cannot find new threads: generic error
warning: File "/lib64/libthread_db-1.0.so" auto-loading has been declined by your `auto-load safe-path' set to "$debugdir:$datadir/auto-load".
To enable execution of this file add
        add-auto-load-safe-path /lib64/libthread_db-1.0.so
line to your configuration file "/root/.gdbinit".
To completely disable this security protection add
        set auto-load safe-path /
line to your configuration file "/root/.gdbinit".
For more information about this security protection see the
"Auto-loading safe path" section in the GDB manual.  E.g., run from the shell:
        info "(gdb)Auto-loading safe path"
warning: Unable to find libthread_db matching inferior's thread library, thread debugging will not be available.
Warning: couldn't activate thread debugging using libthread_db: Cannot find new threads: generic error
warning: File "/lib64/libthread_db-1.0.so" auto-loading has been declined by your `auto-load safe-path' set to "$debugdir:$datadir/auto-load".
warning: Unable to find libthread_db matching inferior's thread library, thread debugging will not be available.
Core was generated by `/usr/bin/janus -F/etc/janus -N -L /var/log/janus.log -p /var/run/janus.pid'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fc705ccc2d0 in sha1_update () from /usr/lib64/libsrtp.so.1
[Current thread is 1 (LWP 12108)]
```

lminiero:
The gdb trace is incomplete, you should provide (in gist/pastebin, not pasted here) the full stacktrace. Make sure your libsrtp and openssl are up-to-date, libsrtp 1.4.x is known to give issues.

bhakimi:
we use net-libs/libsrtp-1.5.2-r1:0 , what command do i run to get the complete output

lminiero:
`bt` is usually enough
https://sourceware.org/gdb/onlinedocs/gdb/Backtrace.html

lminiero:
Any info on the backtrace?

bhakimi:
[...]

PR #887, opened 2017-05-11
Fix on nasty race condition (SRTP contexts for outgoing traffic)

lminiero:
This PR aims at fixing a nasty race condition we only recently found out about, after a deep investigation on reports mostly related to #751.

The issue was basically that, at times, Janus would crash when encrypting traffic with SRTP, apparently deep within the crypto code, e.g., when hashing the packets via OpenSSL. After a lot of debugging, and messing with the libraries a lot for more meaningful information, we found out that the cause was the SRTP context being messed with while it was used. At first we thought it was because of the call being hung up while in use, but we confirmed it was not the case. Besides, all outgoing traffic for a specific user has a dedicated thread, which meant there could be no concurrent access to the libsrtp encrypting features from separate threads, which would indeed break libsrtp (you need to do your own locking for that).

It turned out that this concurrent access was indeed what was happening: at times, the above mentioned outgoing-traffic thread for the user was not properly shut down, and so, even though we had checks in place, when creating a new PeerConnection on the same handle you could end up with a new thread for the user accessing the same stuff. Since they'd both share the same queue, they could both end up encrypting SRTP traffic from the same context, and cause the crash we experienced.

This PR modifies the way the outgoing-traffic is managed. Now, the outgoing-traffic thread is created the first time you create a PC just as before, but is not shutdown when you close the PC. It stays idle instead, and is re-used for any further PC you create on the same handle (e.g., publish/unpublish/publish again in the videoroom, or other SIP calls after you end the first one). This way, the concurrent access can never happen, and the start of the media exchange is actually faster after the first time, as you already have a thread available and don't have to wait for it to spawn.

We tried this patch in some dedicated tests and it seems to be working as expected for us, and it does fix the crash in the scenario I described. Anyway, since it touches a delicate part of the core, I'd ask you all to also test this patch themselves, especially with own applications. As soon as I get feedback that it works as expected and doesn't break anything, I'd like to merge, which should hopefully happen soon as it fixes a serious issue. As usual, if I don't get any feedback I'll go on merge, so it's in your best interest to make sure you're not surprised :wink:.

Thanks!

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -987,6 +987,7 @@ janus_ice_handle *janus_ice_handle_create(void *gateway_session, const char *opa
 	handle->handle_id = handle_id;
 	handle->app = NULL;
 	handle->app_handle = NULL;
+	handle->queued_packets = g_async_queue_new();
 	janus_mutex_init(&handle->mutex);
 
 	/* Set up other stuff. */
@@ -1087,7 +1088,7 @@ gint janus_ice_handle_destroy(void *gateway_session, guint64 handle_id) {
 		return 0;
 	}
 	JANUS_LOG(LOG_INFO, "Detaching handle from %s\n", plugin_t->get_name());
-	/* TODO Actually detach handle... */
+	/* Actually detach handle... */
 	int error = 0;
 	janus_mutex_lock(&old_plugin_sessions_mutex);
 	/* This is to tell the plugin to stop using this session: we'll get rid of it later */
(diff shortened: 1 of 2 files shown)


joshdickson40:
All good here running on Ubuntu 16.04, LGTM.

lminiero:
Added another fix, as we noticed a possible race condition that could lead to the handle's ICE agent to be created twice. This could lead to crashes, as two different loops would live, each generating events. Cleaning the WebRTC resources would close the latest, but not the first, and so events could be triggered that would access a handle that doesn't exist anymore.

@joshdickson40 could you do another check just to be sure I didn't mess anything up? After this, I plan to merge very soon, so thanks again for your feedback!

joshdickson40:
Should be able to get on this either later today or tomorrow at the latest, will report back.

lminiero:
Merging.

PR #888, opened 2017-05-11
Proposal for new authkey mechanism to avoid session highjacking

lminiero:
This PR implements the [proposal](https://groups.google.com/forum/#!topic/meetecho-janus/D9FfyROM6gQ) I made on the group a couple of weeks ago on a new mechanism to protect sessions. To summarize the use case (for more details check the post), the problem I want to fix is people messing with your session if they know the ID. In face, if you know the session identifier, then you can use the session. This includes sending, for instance, a `"destroy"` to kick you out. While unlikely (they're 64bit identifiers, and so hard to just guess), it is a concern, especially because not all generate them randomly.

The idea I came up with is to add an optional property when creating a session. When that property is `true`, Janus returns a value you need to use in all requests addressed to the new session. If the value is missing or wrong, the request is rejected. This way, you're the only one that can operate on your session, as you're the only one to know the value, as simple as that. The default for the property would be `false` for backwards compatibility.

If you use `janus.js`, the feature is very trivial: just add `authkey: true` when creating the session,

```
// Create session
janus = new Janus(
	{
		server: server,
		authkey: true,		<---
		success: function() {
			[...]
```

anf the library will automatically extract the value from the response and use it in to all requests. From an API perspective, in case you use it directly, it's a very easy change too. Just add `"gen_auth_key": true` to a `"create"` request:

```
{
	"janus" : "create",
	"transaction" : "a1b2c3d4",
	"gen_auth_key" : true
}
```

and then read the `auth_key` property Janus sends back and store it:

```
{
	"janus" : "success",
	"transaction" : "a1b2c3d4",
	"data" : {
		"id" : 12345678,
		"auth_key" : 987654
	}
}
```

Then add it to all requests addressed to that specific session, e.g.:

```
{
	"janus" : "attach",
	"session_id" : 12345678,
	"transaction" : "e5f6g7h8",
	"plugin" : "janus.plugin.echotest",
	"auth_key" : 987654
}
```

For the long poll, just add it as an additional query string argument, e.g.:

```
GET --> http://server/janus/12345678?auth_key=987654
```

There has been some discussion on the post on whether this is efficient or not. Looks a pretty easy change to me, and I believe it can address the issue, so I implemented it in the meanwhile as it was a quick fix. Feedback on how to make this better would of course be welcome.


derrell:
My understanding is that both the session id and the auth key are private 64-bit values; neither is communicated beyond the two peers (janus server / janus client). What this is really doing, then, is just making the session id into a 128-bit value (composed of the existing session id and the new auth key), which makes it more difficult to guess. Is that a correct assessment? Or are session ids actually sometimes communicated beyond those two peers, while the auth key would not be? I don't think that's possible, as communicating the session id to another entity would not be useful to that other entity without also communicating the auth key.

lminiero:
That's a good point. No, session IDs are not communicated, as they're only relevant to you, to exchange information with Janus, and the same applies to handles. Even when talking with other users, the session ID is never passed to the other party, as there's no need for that (Janus bridges messages and media); plugins don't have any visibility of session and handles identifiers either. That said, they are easier to "leak", as they can appear, for instance, in web server logs as part of the path you contact (e.g., if using REST). Besides, there are deployments that don't ask for random IDs for sessions to Janus, but generate those themselves, including, for instance, by simply incrementing them for practical reasons. This makes them easier to "attack" and highjack, and associating a random value Janus generates for you makes it harder to do (and the 128-bit observation is probably good in that sense).

derrell:
Hmmm... In the case where the app is specifying some known sequence of session IDs, they would then have to keep track of this auth key. If they're going to have to keep track of something, wouldn't it be simpler for them to just keep track of a randomly-generated 64-bit session  id so that that their session id is no longer (easily) guessable? As you stated, for backward compatibility, the app will have to change to take advantage of the auth key (they'll have to request it), so there's little difference, it seems to me, to modifying the app to request and keep track of a random 64-bit value for the session id, and modifying the app to request and keep track of a random 64-bit value for the auth key in addition to their easily guessed session id. In other words, I think the only thing thing this PR is really gaining is the option for 128 bits instead of 64 bits, for those who are truly paranoid of guessers, at some (albeit not very large) expense in API complexity.

I don't oppose the PR; I'm just not convinced that it's really gaining much for the user.

derrell:
Well, except, maybe for the "leaking" aspect in log files. I guess that's a gain for the user.

uxmaster:
Actually, "leaking" through log files is still there
```
GET --> http://server/janus/12345678?auth_key=987654
```

lminiero:
Ouch, good catch too... well, not sure what to recommend then, as whatever we add would clash with the long poll GET requirements. Now that I think about it, the long poll request never gets to the core (it's handled by the plugin directly, which stores the latest events to shoot back), so I guess we could make the auth key unnecessary for that specific request.

joshdickson40:
Out of curiosity, what's the default behavior for this -- my understanding is that Janus is generating a random session id by default?

> Besides, there are deployments that don't ask for random IDs for sessions to Janus, but generate those themselves, including, for instance, by simply incrementing them for practical reasons.

I'd be interested to know why the solution for this isn't 'have those users change their logic' versus this new addition? I am not saying that is silly or wrong, I just don't understand the need for accommodating this via what's really a repetitive API field.

That being said, I would definitely support Janus using longer/standardized ids (such as adopting UUIDs). UUID is 128bit, so would be similar to the dual 64bit keys.

cmacq2:
The hypothetical attack scenario is that someone is somehow able to learn the necessary session & plugin handle identifiers and then able to interfere with the Janus session arbitrarily, or to trick the client (browser) into supplying the necessary information.

The current situation with the 'random' identifiers is not very good:

1. The function relies on `g_random_int()` which is a Mersenne Twister, i.e. more suited for shuffling through temporary playlists than long lasting 'secrets'.
2. Worse: while on Unices/Unix-likes (`ifdef G_OS_UNIX`) `/dev/urandom` is attempted, on non Unix platforms the seed is actually the current time + pid (`getpid()`). If reading `/dev/urandom` fails in the former case the same fallback is used (time + pid) but with an additional pid of the parent (`getppid()`). See: https://github.com/GNOME/glib/blob/master/glib/grand.c
(There's also a Windows specific fallback which seems less bad.)
3. Worse, the 64bit number is actually 53bits, with the upper 64-53 being masked out manually. See `util.c`, note the `FIXME` in `guint64 janus_random_uint64(void)`.

Additionally they are part of the URLs, which means they are 'leaked' by design. If you want to use these as secret tokens, i.e. not merely to disambiguate requests for lookup purposes, but to actually limit requests to the context of the relevant sessions, then you cannot have them be part of anything which may be automatically sent to Janus by the browser (client) or stored in some form that is exposed to 3rd parties. This rules out using any of:

- URLs
- Cookies
- Local Storage

It is basically just like defending against CSRF and one popular defense is to use a custom HTTP header (e.g. `X-CSRF-Token`) to send & receive such secrets. That approach is also generally well supported by JavaScript frameworks and libraries.

Additionally you need a way to validate that the tokens are not forged, expired or otherwise invalid. You can do that either by storing generated tokens for comparison by Janus later (with associated book keeping to clean out stale tokens and additional dependencies), or by constructing your token in such a way that you can prove only Janus could possibly have generated it. (Typically using an embedded message authentication code. JSON Web Token is a popular example. Needless to say, this does require rather better 'random' numbers for the tokens, though).

lminiero:
@cmacq2 thanks for the detailed analysys!

As to the random identifiers, unfortunately the 53-bit thing is pretty much a constraint we have to live with as that's the best precision JavaScript can get to. For what concerns random identifiers, I'm using the glib randomizers for convenience: I'd love something more truly "random", so if you guys can come up with something better that is as easy (and doesn't make Janus sit down everytime you generate one) I'm on board.

The headers stuff is probably out of discussion. It's very HTTP specific and the Janus API is supposed to be transport agnostic.

cmacq2:
Apart from all that, there is the added wrinkle that Janus seems to be intended as a "dumb" slave backend to a more sophisticated & trusted controller. If that is the intended usage scenario, you could argue that CSRF and other such tokens should be handled at the controller level.

In that case it would be better to adapt `janus.js` so developers have an easy way to inject some callbacks like this:

```
// injects X-CSRF-Token into a $.ajax() options argument.
const myRequestInterceptor = (url, settings) => {
  settings.headers = Object.assign(settings.headers || {}, {
    'X-CSRF-Token': lookupCSRFTokenFor(url, settings)
  });
  return settings;
};

// init, pass interceptor functions.
Janus.init({
  requestInterceptor: myRequestInterceptor, /* callback to augment $.ajax() options/arguments */
  requestSuccessInterceptor: myRequestSuccessInterceptor, /* $.ajax() success handler style callback, invoked before the rest of the 'real' success handler */
  requestErrorInterceptor: myRequestErrorInterceptor /* an $.ajax() error handler style callback, invoked before the rest of the 'real' error handler */
});
```

cmacq2:
@lminiero 

> As to the random identifiers, unfortunately the 53-bit thing is pretty much a constraint we have to live with as that's the best precision JavaScript can get to.

Well, yes and no. Any particular reason why the identifiers must be integer typed? After all, since they are meant to be opaque it *should* not matter whether or not it is possible to do meaningful integer operations on them.

So if you are willing to switch to a string type, you can have all the bits you want: JavaScript certainly does not limit strings at 53 bits max length! ;)

lminiero:
Changing session and handle IDs to strings would mean break 100% of the applications out there. It would mean a complete change to the APIs. Besides, those are keys to hash tables, and doing searches on ints means something way lighter than doing string hashes and compares each time we need to.

Not saying it won't happen, but it most definitely won't happen now.

lminiero:
Closing as it doesn't look like people really find this useful.

PR #893, opened 2017-05-16
Fix Overly Verbose Packet Queueing Message in Audio Bridge

### LINKED ISSUE #892: Quiet 'Too many packets' log warning
I get this message over and over again in my Janus logs:
```
[WARN] Too many packets in queue (12 > 12), removing older ones
```
I see this is generated from the audiobridge plugin, and the relevant section of code is commented thus:
```
```
Since it seems to happen so regularly, might we either reduce the log level, or limit the number of messages output?

There's also the math discrepancy that 12 isn't greater than 12... ;)

joshdickson40:
Fixes #892.

Discussion of change: https://groups.google.com/forum/#!topic/meetecho-janus/9ekR5Vvl9x8


lminiero:
This works for me! Later on I'll probably change this to align to the "SRTP replay" notifications, but for now this is enough. Thanks as usual for your help!

Issue #806, opened 2017-03-04
Warning "onaddstream is deprecated! Use peerConnection.ontrack instead." in janus.js


karth295:
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onaddstream

OS: Fedora
Browser: Firefox 51
janus-gateway: beb0b807aea52d30d697b48708db1ac391785666 (but glancing at source code, issue probably still exists at head)

lminiero:
Yeah we'll have to change that sooner or later.

lminiero:
Can you check if #942 works for you? It implements ontrack, and I'd like to merge ASAP.

lminiero:
Closing as we merged that PR.

PR #894, opened 2017-05-19
Support ontrack

joshdickson40:
This PR adds support for the [RTCPeerConnection.ontrack](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/ontrack) event in `janus.js` (as well as `janus.nojquery.js`, of course!) 

Currently, `janus.js`uses [RTCPeerConnection.onaddstream](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onaddstream) which has been removed from Web standards. Firefox has marked it deprecated (see #806).

As of right now, the only browser to implement `ontrack` is Firefox. Fortunately, `webrtc-adapter.js` is now supporting an `ontrack` shim as of [3.4.1](https://github.com/webrtc/adapter/releases/tag/v3.4.1) which makes it possible for Janus to cleanly move from `onaddstream` to `ontrack` and not deal with any of the particulars.

Aside from updating the adapter version and event name, the only change here is to deal with the fact that the event received by `ontrack` is different than the event received by `onaddstream` (particularly in that it receives an array of streams rather than a single stream.)

Feedback, testing, etc, welcome :)

### DIFF
--- a/html/audiobridgetest.html
+++ b/html/audiobridgetest.html
@@ -5,7 +5,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
 <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
 <title>Janus WebRTC Gateway: Audio Bridge Demo</title>
-<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/webrtc-adapter/3.1.5/adapter.min.js" ></script>
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/webrtc-adapter/3.4.3/adapter.min.js" ></script>
 <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/1.7.2/jquery.min.js" ></script>
 <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery.blockUI/2.70/jquery.blockUI.min.js" ></script>
 <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.0.2/js/bootstrap.min.js"></script>
--- a/html/devicetest.html
+++ b/html/devicetest.html
@@ -5,7 +5,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
 <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
 <title>Janus WebRTC Gateway: Echo Test</title>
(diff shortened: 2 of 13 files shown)


lminiero:
Thanks, this is most welcome! How come it returns an array of streams? Are there cases where you could end up with more than one in the callback?

joshdickson40:
See [example here](https://w3c.github.io/webrtc-pc/#h-advanced-peer-to-peer-example-with-warm-up) where they combine a video and audio stream in the same track.

```
// once remote track arrives, show it in the remote video element
pc.ontrack = function (evt) {
  if (evt.track.kind === "audio") {
      if (answerer) {
        audio = evt.transceiver;
        audio.setDirection("sendrecv");
        if (started && audioSendTrack) {
          audio.sender.replaceTrack(audioSendTrack);
        }
      }
    } else if (evt.track.kind === "video") {
      if (answerer) {
        video = evt.transceiver;
        video.setDirection("sendrecv");
            if (started && videoSendTrack) {
              video.sender.replaceTrack(videoSendTrack);
            }
          }
        }
        // don't set srcObject again if it is already set.
        if (!remoteView.srcObject)
          remoteView.srcObject = evt.streams[0];
    };
```

lminiero:
I'll give it a try as soon as I have some time to give my feedback. If others could do the same that would help merge it sooner. I'll also add this to the list of PRs to check on the group.

lminiero:
Tested and it looks fine to me, so I'll be brave and merge :smile: 
Thanks again for the fix!

lminiero:
@joshdickson40 I should have played with this more... Just found that it looks like this breaks Firefox, so I'm going to revert the change you made in janus[.nojquery].js. When we find a fix that works for both Chrome and Firefox we can restore it.

joshdickson40:
How does it fail? I tested this on Firefox 47 and 53 (current) before PRing and I'm re-testing it now without issue.

lminiero:
For us, just testing the echotest would result in a "no video available" box, on both Firefox 53 and 55. It happened on three different systems, Linux and Windows alike.

joshdickson40:
Got it, this must be a video thing because my audio testing is all passing. Let me try to replicate it this weekend and I can perhaps open a new PR once I have more information on what needs to change.

lminiero:
Yep, I think that only one of the two tracks is going through, and so even if the browser is receiving video (the stats confirm this) the stream only contains audio and so video is not rendered.

joshdickson40:
Okay, thanks Lorenzo. Sorry for the trouble!

lminiero:
No problem at all! It's actually my fault as I should have done some more tests. Thanks for your help and support on all this!

PR #903, opened 2017-05-24
Access-Control-Max-Age header and Janus.listDevices support for microphone-only input

agclark81:
Two contributions:

1 - A change to janus_http.c to set a "Access-Control-Max-Age" header on the responses to eliminate the need to send an OPTIONS request repeatedly to confirm that a cross-site POST is still allowed. This reduces overall network traffic since the CORS headers are static and aren't changing.

2 - A change to Janus.listDevices to accept a second parameter "config" that can take in custom parameters to pass into getUserMedia. While the default { audio: true, video: true } is fine whenever a webcam is attached, there are circumstances where you may only be interested in audio capture and not video, and the getUserMedia call will fail unless there is a webcam attached or video is set to false. This new parameter is optional, so there shouldn't be any impact to existing use.


transports/janus_http.c
```diff
@@ -1108,6 +1108,7 @@ int janus_http_handler(void *cls, struct MHD_Connection *connection, const char
 	if (!strcasecmp(method, "OPTIONS")) {
 		response = MHD_create_response_from_buffer(0, NULL, MHD_RESPMEM_PERSISTENT);
 		MHD_add_response_header(response, "Access-Control-Allow-Origin", "*");
+    MHD_add_response_header(response, "Access-Control-Max-Age", "86400");
```
lminiero: Broken indentation, and same for other lines you added. We use tabs, not spaces.

lminiero:
Both look like reasonable changes, thanks! I've just added a comment on the broken indentation in the C code.

transports/janus_http.c
agclark81: I'll note that for future changes. Would you like me to make that change, or were you going to make it?

lminiero: Please make that change yourself so that I can make a clean merge, thanks!

lminiero:
Thanks for the quick change! I'll give it a check tomorrow (about to leave the office now) and if it works I'll merge :+1:

agclark81:
I've added in an additional commit that hadn't been picked up earlier. This commit fixes an issue when a JSON array is passed into handleEvent, which can occur when max_poll_events is set to >1. If websockets aren't being used, it sets a timeout to call eventHandler in 200ms. If it's a JSON array, it then calls handle event for each item in the response, which in turn adds another 200ms-delayed call to eventHandler. Without this fix, you end up getting duplicated calls that pile up exponentially until there is sufficient backlog in the browser that the "gateway down" error is thrown.

lminiero:
Good catch, and thanks for fixing that too. Is this an issue you actually encountered, or something you evinced from the code? Asking as I don't recall ever getting this, but we may not have used the grouped events (can't recall what janus.js does by default on the top of my head).

agclark81:
This is something that we saw in the wild and is reproducible when max_poll_events is set to >1 and there are bundled events returned, such as from having multiple participants. It probably hasn't been seen by many others since janus.js defaults to max_poll_events=1. On larger webinars, where there is a lot of activity, setting max_poll_events to a value like 999 helps tremendously, since we've encountered the "gateway down" message when there is a firestorm of activity and it exhausts the browser's connection pool.

lminiero:
Makes a lot of sense, thanks for the clarification! I hope I'll be able to look into this before the day ends and merge :+1:

lminiero:
Looks good, merging, thanks again!

PR #904, opened 2017-05-25
Fix sdp parsing for audio in streaming plugin

chunmeng:
The curldata is reset after parsing video section and the audio sdp parsing will always return error.
Thus the streaming always missing audio.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -3301,12 +3301,26 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 	char vfmtp[2048];
 	char vcontrol[2048];
 	char uri[1024];
-	char transport[1024];
+	char vtransport[1024];
 	multiple_fds video_fds = {-1, -1};
-	int result;
-	result = janus_streaming_rtsp_parse_sdp(curldata->buffer, name, "video", &vpt,
-		transport, vrtpmap, vfmtp, vcontrol, &source->video_iface, &video_fds);
-	if(result != -1) {
+
+	int apt = -1;
+	char artpmap[2048];
+	char afmtp[2048];
+	char acontrol[2048];
+	char atransport[1024];
+	multiple_fds audio_fds = {-1, -1};
+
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for spotting this! I think I never tried RTSP streams with both audio and video and so never met this issue.

A couple of notes:

1. Just a nit: please use `/*   */` for comments, rather than `//`;
2. Have you tried if this breaks RTSP reconnections, if the server goes away and Janus tries to reconnect?
3. Will this introduce leaks, since we're using a new buffer we allocate here? It shouldn't be the case, as from a quick glance to the code I see you free it in case of success or error, but wanted to double check.

Pinging @fbertone as he's one of the most active Janus+RTSP users (and contributor of a lot of RTSP-related code), and so may have some additional feedback on this.

That said, please sign our [CLA](https://cla.conf.meetecho.com/janus/) so that we can merge this when ready.

fbertone:
Neither I have ever used it to stream both audio and video but it makes sense I guess.
I will not be able to test it before a couple of weeks

chunmeng:
1) Will do
2) Didn't test this case. Let me try and feedback later.
3) I also dislike having to be freeing in all possible path path. I'll probably do it another way - by parsing both video and audio first before proceeding with both setup. With this I only need to have one more transport[] for audio.

Signed.

lminiero:
Thanks for the clarification! I'll play with this when I get back to the office next week :+1:

chunmeng:
2) Reconnection test looks good.
[rtsp_reconnect_log.txt](https://github.com/meetecho/janus-gateway/files/1033388/rtsp_reconnect_log.txt)
3) Update the commit with re-arranged logic to avoid mem allocation.

lminiero:
Looks good, thanks! There's just a weird comment, but I'll fix it myself after merging so no problems.

I couldn't test the RTSP code for audio+video, as I don't have any RTSP server with both enabled to test it against right now, but I'm sure you made all the tests yourself to validate that. Checked with a video-only RTSP server, with reconnection too, and it still works so that's ok for me.

Merging!

PR #905, opened 2017-05-25
External token authentication and authorization

derrell:
The traditional Janus token authentication provides limited ability to control
access. It allows blocking access to plugins entirely, but no finer-grained
control than that, e.g., it is not possible to have a token that allows access
to particular rooms from the videoroom and/or particular streams from the
streaming plugin.

With this patch, external authentication is allowed. In the default case,
authentication is accomplished via an HTTP server. An example server
implementation is provided in the new directory auth-external, which
demonstrates how to allow certain requests while denying others, even to the
same plugin.

Although the default external authenticator sends its requests to an HTTP
server, other authenticators could be written. Only a single function pointer
need be changed to switch to a different authenticator.

The HTTP authenticator uses libcurl to talk with the authenticator server. If
libcurl is not available and external authentication is requested, a default
authenticator that always disallows requests is used.

The general->token_auth configuration key in janus.cfg may be set to a boolean
value as before ("yes", "no", "true", "false"). It may now additionally be set
to "external" to enable the external token authenticator.

One new configuration key is added to janus.cfg: general->token_auth_private.
For the default external authenticator (sending HTTP requests), this private
data is set to the URL at which the authenticator is listening.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+*~
+TAGS
+
 cmdline.c
 cmdline.h
 version.c
--- a/auth-external/.gitignore
+++ b/auth-external/.gitignore
@@ -1,0 +1,2 @@
+*~
+node_modules
--- a/auth-external/README
+++ b/auth-external/README
@@ -1,0 +1,12 @@
+###
+### To run this example authentication/authorization server...
+###
+
+# Install nodejs in whatever way your distribution provides it.
+apt-get install nodejs
+
+# Install the required sub-packages
+npm install .
+
+# Start it up!
+nodejs /authserver-example.js
--- a/auth-external/README
+++ b/auth-external/README
@@ -9,4 +9,4 @@ apt-get install nodejs
 npm install .
 
 # Start it up!
-nodejs /authserver-example.js
+nodejs ./authserver-example.js
--- a/auth-external/authserver-example.js
+++ b/auth-external/authserver-example.js
@@ -1,0 +1,142 @@
+//
+// Whhat port to listen on
+//
+var PORT = 3000;
+
+//
+// Configure (in this example, hard-coded) allowed tokens and their authorizations
+//
+var allowedTokens =
+{
+  // Token 'default' is only allowed the "standard" stuff, e.g., 
+  // not plugin-specific, or "trickle".
+  default : function(request)
+  {
+    var authenticated = allowStandard(request);
+console.log("DEFAULT: returning", authenticated);
+    return authenticated;
+  },
+
+  // Token 'echoonly' is only allowed access to the echo plugin (and the "standard" stuff)
+  echoonly : function(request)
+  {
+    var authenticated;
+
+    // Default to disallowing all requests
+    authenticated = allowStandard(request);
+
+    // Allow specific requests to the streaming plugin
+    if (request.plugin_package == "janus.plugin.echo")
+    {
+      authenticated = true;
+    }
+
+    return authenticated;
+  },
+
+  // Token 'restricted' is only allowed the following:
+  //   - Access to the streaming plugin's "list" and "start" requests;
+  //   - Access to the streaming plugin's "watch" request if the requested id is 999
+  //   - Complete access to the echo plugin
+  restricted : function(request)
+  {
+    var             authenticated;
+
+    // Default to disallowing all requests
+    authenticated = allowStandard(request);
+
+    // Allow specific requests to the streaming plugin
+    if (request.plugin_package == "janus.plugin.streaming" &&
+        request.janus == "message")
+    {
+      // Allow streaming video 999
+      if (request.body.request == "watch" && request.body.id == "999")
+      {
+        authenticated = true;
+      }
+
+      // Allow the "list" and "start" commands
+      if ( [ "list", "start" ].indexOf(request.body.request) != -1)
+      {
+        authenticated = true;
+      }
+    }
+
+    // Allow requests to the echo plugin
+    if (request.plugin_package == "janus.plugin.echo")
+    {
+      authenticated = true;
+    }
+
+    return authenticated;
+  },
+
+  // Token 'superuser' is granted unlimited access.
+  superuser : function(request)
+  {
+    return true;
+  }
+};
+
+// Prepare to create a web server
+var express = require('express');
+var bodyParser = require("body-parser");
+
+var app = express();
+
+app.use(bodyParser.json());
+
+// Authenticate when we receive a root-level POST request
+app.post(
+  "/",
+  function(req, res)
+  {
+    var             request = req.body;
+    var             authenticated = false;
+
+    console.log("authserver got request:\n" + JSON.stringify(request, null, "  "));
+
+    // Is this token recognized?
+    if (typeof allowedTokens[request.token] == "function")
+    {
+      // Yup. Authenticate/authorize based on the request and token
+      authenticated = allowedTokens[request.token](request);
+    }
+    console.log("authserver: " + (authenticated ? "allowed" : "not allowed"));
+
+    // Send back the result
+    res.writeHead(200, {'Content-Type': 'application/json'});
+    res.end(
+      JSON.stringify(
+        {
+          authenticated : authenticated
+        }));
+  });
+
+// Initialize immediately.
+var port = PORT;
+app.listen(port);
+console.log('Listening at http://localhost:' + port)
+
+
+//
+// Allow the standard requests that any valid token likely wants
+//
+function allowStandard(request)
+{
+  var authenticated = false;
+
+  // Allow trickle requests
+  if (request.janus == "trickle")
+  {
+    authenticated = true;
+  }
+
+  // Allow any request that's not to a specific plugin
+  if (! request.plugin_package)
+  {
+    authenticated = true;
+  }
+
+  return authenticated;
+}
--- a/auth-external/authserver-example.js
+++ b/auth-external/authserver-example.js
@@ -13,7 +13,6 @@ var allowedTokens =
   default : function(request)
   {
     var authenticated = allowStandard(request);
-console.log("DEFAULT: returning", authenticated);
     return authenticated;
   },
 
--- a/auth-external/package.json
+++ b/auth-external/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "authserver",
+  "version": "0.0.1",
+  "description": "Example token authentication server for Janus Gateway",
+  "main": "authserver-example.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "dependencies": {
+    "body-parser": "^1.17.2",
+    "express": "^4.15.3"
+  },
+  "author": "Derrell Lipman <derrell.lipman@unwireduniverse.com>",
+  "license": "GNU General Public License v3"
+}
--- a/auth.c
+++ b/auth.c
@@ -15,29 +15,64 @@
  * \ref core
  */
 
+#ifdef HAVE_LIBCURL
+#  include <curl/curl.h>
+#endif
 #include "auth.h"
 #include "debug.h"
 #include "mutex.h"
+#include "ice.h"
+#include "janus.h"
+#include "plugins/plugin.h"
 
 /* Hash table to contain the tokens to match */
 static GHashTable *tokens = NULL, *allowed_plugins = NULL;
 static gboolean auth_enabled = FALSE;
+static char auth_type = '-';
+static const void *auth_external_private_data = NULL;
 static janus_mutex mutex;
 
 static void janus_auth_free_token(char *token) {
 	g_free(token);
 }
 
+gboolean (*janus_auth_check_token_external)(const char *token, json_t *root) =
+#ifdef HAVE_LIBCURL
+        janus_auth_check_token_http;
+#else
+        janus_auth_check_token_false;
+#endif
+
+
 /* Setup */
-void janus_auth_init(gboolean enabled) {
-	if(enabled) {
-		JANUS_LOG(LOG_INFO, "Token based authentication enabled\n");
+void janus_auth_init(char type) {
+        auth_type = type;       /* save the requested authentication type */
+
+        switch(auth_type) {
+        case 'I' :              /* Internal (traditional) token authentication */
+		JANUS_LOG(LOG_INFO, "Internal token based authentication enabled\n");
 		tokens = g_hash_table_new_full(g_str_hash, g_str_equal, (GDestroyNotify)janus_auth_free_token, NULL);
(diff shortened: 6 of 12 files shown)


lminiero:
This looks like a tremendous work, thanks for this! Although I'm not sure we can merge this as it is. I most definitely don't want any nodejs code or sample server in here (it's not there even for the sample event handler). Anyway, I'm not rejecting the idea, which I believe is valid, so more on this below!

From how I see it, the existing token mechanism should already cover most of the requirements you have: the external authentication could "feed" the opaque tokens Janus validates. Of course, though, it would be up to you to keep them up to date, and you'd need a separate component to take care of these updates. Besides, I guess the main point here is also the ability to accept a request to a specific plugin, and possibly reject another request to the same plugin with the same token, something we currently can't do. The ability to have different authenticators, possibly tailored to the system, would be nice as well.

As such, I do see potential in this, especially in the flexibility it could provide, but that's not something I'd bake in the auth.c as you're doing now. One way to handle this could be to create a new type of plugins, pretty much as we did for event handlers. We could transform auth.c into an intermediary that can access different plugins: one of them could be something that works like the internal tokens we have now; another could be using curl as your does; others could be written to interface to other popular authentication systems; and most importantly, anybody could write their own authenticator, if they're using something custom.

In case that might happen, not sure how we should handle multiple authenticators enabled at the same time: e.g., a request comes for access to a plugin, who do you ask it to? Do we always make sure that only one can be used? Or maybe we could use different authenticators for different plugins. Anyway, those are implementation details, and probably thinking too much ahead.

Looking forward to your thoughts, and thanks again for working on this!

derrell:
Thanks for the review!

This is, actually, implemented as switchable/pluggable (via the janus_auth_check_token_external function pointer); it's simply not exposed to configuration files. The specific functions for the http authenticator could be easily moved to a separate file as the http pluggable authenticator. The traditional token authenticator could then be another pluggable one, so we start with two and people could additionally write their own. 

I thought about doing that, but my gut feeling is that the original internal token authenticator and the added external one likely solve most people's need. If you'd like for me to make that change, I'd be glad to.

I'm curious why you have such a strong feeling about the example nodejs code? It's really just user documentation. I suppose it could be moved to a documentation directory, but it seems useful to users to see, _somewhere_, how the the backend side would be implemented, no? Again, I'll be glad to remove it if it's a blocker (after all, **I** understand what the backend needs to do :-) );  I'd just like to understand what you're thinking is regarding example code as documentation.

Thanks again! Let me know your thoughts and I'll make the changes today. Maybe we can get the PR incorporated before the weekend. :-) :-) :-)

derrell:
If we wanted to allow for multiple authenticators in a chain we could define two more possible return codes from the authenticator. We might have these options:
- TRUE means authentication passes
- FALSE means it fails
- NEXT_OR_TRUE means try the next authenticator to see what it says; like TRUE if there is no next
- NEXT_OR_FALSE means try next if it exists, but FALSE if there is no next one.

Do you have any use cases in mind, where more than one authenticator might be desirable?

lminiero:
> This is, actually, implemented as switchable/pluggable

What I mean is a real modular approach, just as media, transport and event plugins work. As it is in your PR, the pluggability would need to be baked in the core, via different functions you can then set the pointer to. I'd much rather have this as a shared object that is completely independent of Janus itself, so that anybody can write their own, even outside the context of the official repo, and update an existing Janus instance. Not saying you'd need to do that, of course, I can definitely laid the foundation for this: I was more interested in understanding whether that made sense. From my perspective, this seems like a more flexible approach, and something that could be very easily extended by third party implementors.

> I'm curious why you have such a strong feeling about the example nodejs code?

Moving the code to a documentation folder wouldn't work for me either. Anything that is in the repo is blindly seen by users as "the way", and most of the times they don't go beyond that. I've specifically kept the node.js example of an event handler out of the repo for the same exact reason. Examples or reference applications can be shared out of band like on the group, or as separate repos (see @thehunmonkgroup's excellent event handler server, for instance).

> If we wanted to allow for multiple authenticators in a chain [..]

Thinking about this, that probably wouldn't be a good idea. Authentication requests are going to be synchronous, and chaining them could lead to very high response times, if not timeouts. Only one authenticator should be used for a request: what the policy is to decide which authenticator has authority on a specific request is something that can be discussed.

> Maybe we can get the PR incorporated before the weekend. :-) :-) :-)

Very unlikely to happen, sorry :smile: 
Unfortunately I don't have time to do any testing before next week, but anyway as anticipated I'd rather start preparing a new modular approach on which your improvements could be applied. I could prepare the new modular approach (which should be easy enough) early next week, and then write a skeleton module that you can extend with an ad-hoc PR that implements the curl-based hook to an external backend. I'll probably keep the existing auth-token mechanism part of the core as it is, as it receives directives from the Admin API, and as such couldn't be hooked to those plugins (who would each have a very different authentication mechanism, probably, so no shared properties to set). It would only be there for simpler use cases (just like the API secret), while anything more fine-grained will require an authenticator module to work.

Does this make sense?

derrell:
Yes, that makes sense, and is absolutely fine for me. I agree that the full modular approach is more general purpose. It seemed a bit overboard, but once implemented, it'll be a trivial matter to fit the http authenticator into that modular framework.

Note when you implement that modular approach that the requirement for an authenticator to do its job (including authorizing specific requests) is that the `root` object be passed to the authenticator in addition to the already-passed token so that it can inspect the entire request.

I see the following functions in the module interface:
- `void (*init)(gboolean enabled);`
  - called by _janus_auth_init_
  - for the default authenticator, this would create the hash table as is currently done
- `void (*deinit)(void);`
  - called by _janus_auth_deinit_
  - for the default authenticator, this would destroy the hash table
- `gboolean (*check_token)(const char *token, json_t *root);`
  - called by _janus_auth_check_token_
  - for the default authenticator, do a hash table lookup (ignoring the _root_ parameter)

Anything else? I think that's all that's needed

I'll eagerly await "early next week" and then add an http authenticator plugin that fits your modular interface.

Thanks!

derrell:
Even though admin requests to deal with tokens won't have any effect if a different authenticator is in use, maybe the hash tables should be initialized right in _janus_auth_init_ instead of in the default authenticator's `init` function, so that admin requests don't crash janus. That implies that the hash tables should be destroyed right in _janus_auth_deinit_ instead of in the default authenticator's `deinit` function.
Those two functions, for the default authenticator, may have no work to do.

lminiero:
> Note when you implement that modular approach that the requirement for an authenticator to do its job (including authorizing specific requests) is that the root object be passed to the authenticator in addition to the already-passed token so that it can inspect the entire request.

Would this be really needed? Can't this simply require some key properties from the request (to be defined) that provide the info the authenticator needs? e.g., session, handle, opaque IDs, token, plugin to subscribe to, maybe something else. I guess the motivation here is to be more fine-grained for what concerns plugin internals, but anything plugin specific would make the module not extensible enough, I believe, as plugins may change, and new plugins would not be covered easily. To be more specific, which requests do you see to be subject to this authentication scrutiny? It can't possibly be all requests, as that would drag the server down.

An alternative, or possibly enhancement to the original idea, might be providing plugins with a callback via the core to request an authentication property themselves for some commands: e.g., 'got a "create" request from this user, what should I do?', so that plugins can answer with a success or an error within the context of their own messaging. We already have something similar for when plugins send their own events to handlers, where the core acts as an intermediary towards handler plugins.

derrell:
Here's an example from the example nodejs auth server, for the token "restricted"
```
  // Token 'restricted' is only allowed the following:
  //   - Allow trickle request and requests without a plugin
  //   - Access to the streaming plugin's "list" and "start" requests;
  //   - Access to the streaming plugin's "watch" request if the requested id is 999
  //   - Complete access to the echo plugin
  restricted : function(request)
  {
    var             authenticated;

    // Allow "trickle" and requests without a plugin
    authenticated = allowStandard(request);

    // Allow specific requests to the streaming plugin
    if (request.plugin_package == "janus.plugin.streaming" &&
        request.janus == "message")
    {
      // Allow streaming video 999
      if (request.body.request == "watch" && request.body.id == "999")
      {
        authenticated = true;
      }

      // Allow the "list" and "start" commands
      if ( [ "list", "start" ].indexOf(request.body.request) != -1)
      {
        authenticated = true;
      }
    }

    // Allow requests to the echo plugin
    if (request.plugin_package == "janus.plugin.echo")
    {
      authenticated = true;
    }

    return authenticated;
  }
```

It's the job of the specific authenticator to deal with what to do with the request. In this case, I look at what plugin requested, and for certain plugin types, I dig into the message to see what's being requested. I wanted, here, to allow this particular token to access only stream 999 in the streaming plugin, but it can access anything in the echo plugin (and nothing in any other plugin).

I don't think we can guess what the authenticator might want to look at. There's no particular reason not to pass the message to the authenticator, and only the authenticator can know what it cares about. As to obsoleting an authenticator by changes to plugins, that's up to the user's authenticator if they write their own. The http authenticator itself doesn't have anything that is plugin-specific; only its server side, which is not part of janus, knows anything about what specifically it's looking for in the messages. There's no possibility of bitrot in the auth (or http auth authenticator) code here, as other code such as plugins change over time.

The callback mechanism wouldn't work with the http auth plugin. That would require rewriting it with a bidirectional tcp pipe, which makes writing the server side more onerous, and adds to the response time of an authentication request. I'd highly recommend against that option.

Cheers,

Derrell

lminiero:
> There's no particular reason not to pass the message to the authenticator

Well, I can think of one: if you're receiving 1000 requests, with external HTTP-based validation it's 1000 requests you make yourself to an external component! Really overkill, and will slow everything down because those requests will be synchronous... I'd much rather only enforce authentication on a selected subset of key requests. For instance, filtering trickle candidates seems useless. Makes more sense to only validate:

1. session creation or destruction request;
2. attaching to a handle (what auth tokens intercept right now), or detaching from one;
3. messages to a plugin (but in this case, I wouldn't have the core pass them to the authenticator: I'd let each plugin decide what they might want authenticated, so that they can reject themselves as part of the plugin messaging, and not a generic core error message).

Any other request is pretty much harmless, and forwarding everything would turn Janus into a snail.

lminiero:
Maybe one more thing one could want to authenticate is whether somebody can negotiate a PeerConnection: e.g., if we got an offer, maybe don't process it at all if this user can't do WebRTC. But considering that's what WebRTC is for this seems a bit useless too.

derrell:
I have considered, for the http response, that I could send back a list of requests that were always to be accepted (e.g., trickle), so that they needn't be sent to the server; and possibly a timeout on other requests, so multiple requests for the same thing within the timeout period would be resolved by the local cache.

It's an interesting question why some of these requests are passed to the authenticator at all. Why, for example, do trickle requests need to be authenticated? Eliminating only those would greatly cut down on the number of authentication requests. I could certainly (and probably should) filter those in the http authenticator ocde, and just return TRUE without ever passing them to the server.

Given that this is being generalized, however, and not specifically for the http authenticator, there may be use cases where the whole message, not just a subset of its parts that are "arbitrarily" chosen (well, maybe chosen via "educated guess" is a better description), is needed by the authenticator. 

What I envision needing for my own use is the plugin package name (e.g., "janus.plugin.echo"), the request.janus value (e.g., "message"), and the plugin data (is it always request.body?) . I'd probably be happy if only those things were passed to the the authenticator. I still don't agree, though, that extracting those fields is necessary work when some authenticators like the default one, won't need that information at all, and other authenticators may need more or less. I think it's best to just pass the `root` object to the authenticator and let it extract what it needs.

Note that even with 1000 requests, sent over http, the delay is likely to be entirely unnoticeable at the "human" speeds at which most of these requests occur. We're not talking video here; we're talking user-initiated actions (except for things like trickle which probably don't need to be authenticated at all, and in any case, could avoid the http request). Right?

Hopefully I'm convincing you. :-)

Derrell

lminiero:
I still don't like the idea of the whole message getting to the authenticator, because:

1. you're processing each request twice (the core already did that, and now the authenticator does it again);
2. for `message`, you need to know everything about the plugin signalling, which means this is not extensible or generic enough.

That's why I was suggesting, for `message`, a different approach:

1. core gets a `message` via Janus API;
2. the request gets to the plugin (e.g., videoroom) as usual;
3. plugin parses request as usual;
4. plugin finds request that might need some checks (e.g., create) and decides to involve the authenticator;
5. something like `gateway->authenticate_request(handle, something);` happens;
6. callback is handled by the core, which forwards the request to the authenticator responsible for the videor oom plugin, and invokes a method specifically targeted for plugin specific requests, e.g., `authenticator3->plugin_specific_request(session, handle, plugin, something);`
7. the authenticator does its thing on something that is plugin specific, but more "regulated";
8. TRUE/FALSE crawls back to plugin, that can then decide whether to accept the request or return an error.

The `something` above would still be a `json_t` of some sort, but much smaller in size and intent, we can "regulate" it somehow, or at the very least have each plugin use a known and documented set of things it can ask for.

lminiero:
If you really need something that validates each request, it's much easier to just put your own HTTP server (e.g., in nodejs) in front of Janus and proxy all requests. Each request gets to you first: you parse it, decide if it can go through, and if not respond with an error, otherwise you forward as is to Janus. You can even add whatever attribute you want there, for authentication purposes, and then strip them before forwarding to Janus.

Sounds like a simpler solution that wouldn't require any change to Janus itself.

derrell:
I was thinking about what you said:

> Maybe one more thing one could want to authenticate is whether somebody can negotiate a PeerConnection: e.g., if we got an offer, maybe don't process it at all if this user can't do WebRTC. But considering that's what WebRTC is for this seems a bit useless too.

This mechanism could be used to help prevent DOS attacks. At present, I think, someone could create many, many peer connections. Denying those at the authenticator level could help to prevent such attacks?

derrell:
This PR I created was very general, allowing for the authentication and authorization needs of nearly any app. I still think it's the best way to go in Janus (with the addition of the plugin architecture we discussed). For our own needs, though, I've now implemented a proxy in our code that fronts Janus, doing all of the authentication and authorization. Authorized requests are passed on to Janus; others return a Janus-like error without ever touching Janus.

I appreciate the discussion! 

Depending on your desires regarding this PR, we could close it, or leave it open for additional discussion as you implement the authentication plugin architecture.

Thanks!

Derrell

lminiero:
I appreciated the discussion too, and am looking forward to move it forward in the near future! Let's keep this open until we have a PR with the new approach to start from.

lminiero:
Closing this since, as you mentioned, one can achieve the same result via some kind of frontend. 

Unfortunately I didn't had time to investigate the alternative approach we discussed, and due to other priorities it will probably take some more time before that can happen. I personally still believe this to be quite orthogonal to Janus, so event processing plus an external frontend (or wrapper) still seems like the best way to take care of this, in a way that is transparent to Janus. Please feel free to write a new post on our Google group if you believe this deserves some more discussion. Thanks!

lminiero:
@derrell just FYI (in case you're not aware of it already), there's a new PR for authentication/authorization you may be interested in: #1159

PR #906, opened 2017-05-25
Include audio_port, data_port and video_port in info result for streaming plugin

Sean-Der:
Pretty simple commit, sometimes I need to know the ports after streaming room creation.


lminiero:
When we accepted the PR for random ports, I explicitly asked for this to be removed, as it's a security concern. If info returns the port, nothing prevents attackers to query it and start sending random stuff to that port. This is one of the reasons why I asked the random port option to be only possible via API, and not when configured statically, as you'd then have no way to retrieve the port you were assigned. The random ports are already returned to you when you create the mountpoint, so there's no real need for the info to include those too.

As such, we can't accept this PR, sorry!

Sean-Der:
Ugh I am an idiot sorry @lminiero I was asked to add this feature and didn't really think it through

Issue #900, opened 2017-05-23
Detect the audio level change from speaking to not speaking 


atyenoria:
https://github.com/meetecho/janus-gateway/pull/760
The functionality of audiolevel detection works well. I'm very thankful for this.
I have a question. Is it possible to detect the change from the state of speaking to the state of not speaking?
At now, I can get only the the change from the state of not speaking to the state of speaking.
This is very similar to https://github.com/otalk/hark .
Btw, for mobile app development, react-native-webrtc can't get the audio level ( oney/react-native-webrtc#276 ). So, There is no possible solution for this ( only for react-native developer? )
So, This feature is very good solution for detecting the speaker for both web and mobile at now.
How do you feel that?

lminiero:
Not sure exactly when we'll be able to work on this, but added to the TODO list :+1:

lminiero:
Sorry if this is taking a while, but we're quite busy these days... Just to clarify, are you talking of AudioBridge or VideoRoom for this events? I guess both, considering the mechanism is actually very similar?

Anyway, I've been thinking about how this could be done and I have a few doubts. Right now, we're sending several "talking" events in sequence, depending on the amount of packets we need to process before we can decide somebody's speaking. Following a new "talking"/"not talking" event, this would need to change, as we would only send a "talking" event the first time we detect the user started speaking, while we would not send others if our calculation still detects the user as speaking: we'd remain in the same state. On the other hand, if the calculation detects the user is not speaking, while he was speaking before, we should send something like "not talking".

The issue I see here with this new state-based notification mechanism, is in how we should handle new users joining: to make an example, what happens when you a new user joins *after* you send a "talking" event. In fact, this new user would not receive the "talking" event (we're not sending them on a regular basis anymore), but may only receive the "not talking" one later. We'll probably have to keep track of this talking state as part of the participant's properties, so that when somebody new joins we can notify about the talking status of all talking participants just as if the user started talking only now. The same would apply to both plugins, I guess, possibly with slightly different semantics.

Does this sound reasonable? Pinging @mirkobrankovic as he implemented the feature originally for additional feedback.

lminiero:
[...]

PR #910, opened 2017-05-31
Talking / stopped talking events, instead of talking-only repeats

lminiero:
Attempt to implement talking/stopped-talking events in both AudioBridge and VideoRoom plugins as discussed in #900. This replaces the recurring "talking" event we have in place right now.

This patch changes the behaviour so that you're only notified once when a user starts talking, and you're notified when the user stops talking as well. In both cases, the same configuration properties as before are used for the calculation. If you configured a window of 100 packets, it means we'll check that number of packets before deciding if a user is talking or not (using the average specified in the room settings), and after that we reset the count so that we have to wait for the next 100 packets. Considering we don't have a recurring "talking" event anymore, I guess we can reduce the default: e.g., 50 packets (~1s) is probably better than 100 (~2s).

For what concerns users joining after a "talking" event has been received, I modified the events we already send when you join, by adding a "talking" property to the info on existing participants. This should allow the client side to have a consistent talking state.

Tested this only briefly and it seems to work, but please give this a try yourself too to make sure it works as expected (and, more importantly, that it doesn't break anything).

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -268,7 +268,8 @@ record_file =	/path/to/recording.wav (where to save the recording)
 		{	// Participant #1
 			"id" : <unique numeric ID of the participant>,
 			"display" : "<display name of the participant, if any; optional>",
-			"muted" : <true|false, whether user is muted or not>
+			"muted" : <true|false, whether user is muted or not>,
+			"talking" : <true|false, whether user is talking or not (only if audio levels are used)>,
 		},
 		// Other participants
 	]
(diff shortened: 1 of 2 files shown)


mirkobrankovic:
Finnally improvement ;)
I did it quick and dirty ;p

mirkobrankovic:
Works perfect as far as i can see.
I tested state on joining, when talking and not talking:
talking

lminiero:
Thanks for the feedback! As soon as @atyenoria confirms this works for him as well, especially in the mobile apps he was working on, I'll merge then :+1:

atyenoria:
@lminiero @mirkobrankovic I checked it. It works well. But, I have one consideration. It seems that the event is only got from the others, not myself. I'm not sure that it is good to get the self detection. 
The use case for that is to check the audio device by myself in order to set the parameters of audio_active_packets = 50 and 
audio_level_average = 80.
For mobile and react-native developer, as I said, There is no solution for getting audio info from getUsermedia. 
Anyway, I think that this can be merged soon.
How do you feel that?

lminiero:
Yes, events are sent only to other participants, not yourself, because we
reused a helper method we use for other notifications too. Not sure if
notifying your own volume back would be of much use, though. I guess that
if it's only to make some tests, you'll have to join twice so that you can
get events from the other peer.

Il 01 giu 2017 11:13 PM, "Akinori Nakajima" <notifications@github.com> ha
scritto:

atyenoria:
I got it. Thanks. It seems to easier to test in client side.

mirkobrankovic:
Next stop... This could be reused to adjust audio level of speaker in audiobridge when someone is too loud , before mixer. Just my 3¢

lminiero:
May be useful, but not something I plan to do in this PR here.

lminiero:
Ok, since this seems to be working, I'll merge. We can discuss other possible enhancements on the group or a new issue. Thank you both for the quick feedback!

anildigital:
Any documentation for this?

PR #915, opened 2017-06-02
Added audio to screen sharing

giordanocardillo:
After some research I found out that audio can be sent along with ScreenSharing. You just need to ask for the stream then add the audio track to the screen share stream. Hope this can be useful.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1504,7 +1504,7 @@ function Janus(gatewayCallbacks) {
 						navigator.mediaDevices.getUserMedia(constraints)
 							.then(function(stream) { 
 								if(useAudio){
-									navigator.mediaDevices.getUserMedia({ audio: true })
+									navigator.mediaDevices.getUserMedia({ audio: true, video: false })
 									.then(function (audioStream) {
 										stream.addTrack(audioStream.getAudioTracks()[0]);
 										gsmCallback(null, stream);
--- a/html/janus.js
+++ b/html/janus.js
@@ -1498,11 +1498,21 @@ function Janus(gatewayCallbacks) {
 							streamsDone(handleId, jsep, media, callbacks, stream);
 						}
 					};
-					function getScreenMedia(constraints, gsmCallback) {
+					function getScreenMedia(constraints, gsmCallback, useAudio) {
 						Janus.log("Adding media constraint (screen capture)");
 						Janus.debug(constraints);
 						navigator.mediaDevices.getUserMedia(constraints)
-							.then(function(stream) { gsmCallback(null, stream); })
+							.then(function(stream) { 
+								if(useAudio){
+									navigator.mediaDevices.getUserMedia({ audio: true })
+									.then(function (audioStream) {
+										stream.addTrack(audioStream.getAudioTracks()[0]);
+										gsmCallback(null, stream);
+									})
+								} else {
+									gsmCallback(null, stream);
+								} 
+							})
 							.catch(function(error) { pluginHandle.consentDialog(false); gsmCallback(error); });
 					};
 					if(adapter.browserDetails.browser === 'chrome') {
@@ -1593,7 +1603,7 @@ function Janus(gatewayCallbacks) {
 								callbacks.error(error);
 							} else {
 								constraints = {
-									audio: isAudioSendEnabled(media),
+									audio: false,
 									video: {
 										mandatory: {
 											chromeMediaSource: 'desktop',
@@ -1609,7 +1619,7 @@ function Janus(gatewayCallbacks) {
 									}
 								};
 								constraints.video.mandatory.chromeMediaSourceId = event.data.sourceId;
-								getScreenMedia(constraints, callback);
+								getScreenMedia(constraints, callback, isAudioSendEnabled(media));
 							}
 						} else if (event.data.type == 'janusGetScreenPending') {
 							window.clearTimeout(event.data.id);
--- a/html/janus.nojquery.js
+++ b/html/janus.nojquery.js
@@ -1556,7 +1556,7 @@ function Janus(gatewayCallbacks) {
 						navigator.mediaDevices.getUserMedia(constraints)
 							.then(function(stream) { 
 								if(useAudio){
-									navigator.mediaDevices.getUserMedia({ audio: true })
+									navigator.mediaDevices.getUserMedia({ audio: true, video: false })
 									.then(function (audioStream) {
 										stream.addTrack(audioStream.getAudioTracks()[0]);
 										gsmCallback(null, stream);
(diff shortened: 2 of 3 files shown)


lminiero:
Thanks for this! Although there's already a PR for that, #838, are they substantially different? Does this work with both Chrome and Firefox? Any potential regression we should be wary of?

From a quick glance, it looks like you simply capture audio and video differently, and then you add the audio track to the video capture. This may be different from what @chadfurman wanted to accomplish in the PR I mentioned, because in his case we wanted to capture system audio as well, and not the mic audio as your patch would do. In this case, I guess we should introduce new flags for the audio capture as well, just as we do for audio: e.g., if audio is `true` or anything we do already, capture the mic; if it's `system` or something like that, try and get the system audio instead.

Not sure what the best approach is here. I think we can start easy, make sure your patch works everywhere and doesn't break anything, and in case merge that. Then, expand #838 to add the new audio property to try and get system audio in case (which would require a change to the screensharing plugin, though, and AFAIK doesn't work on Firefox).

giordanocardillo:
On chrome when you request screen media from the navigator api if you include audio: true, it causes an error, because "screen media is without audio" so with this patch, passing sendAudio: true to the screen sharing, the audio track will be added to the screen stream. The kind of audio you provide is defined at browser level, I mean, if you select "stereo mixing" it sends pc audio, if you select microphone, it send microphone audio

lminiero:
Thanks for the clarification. We'll make some tests to make sure it doesn't break anything on neither Chrome nor Firefox, and in case merge. Before we can do that, though, please sign our [CLA](https://cla.conf.meetecho.com/janus/).

lminiero:
PS: not sure what you mean by "stereo mixing". I most definitely am not prompted for anything like this when doing a getUserMedia on Chrome (Fedora 25).

giordanocardillo:
Sorry, I was giving a Windows based example, with "stereo mixing" I mean whichever record input you use to e cord pc audio, since the browser can use the default system audio input. BTW I signed the CLA

lminiero:
Just a quick post to reassure you this is still on my TODO list: I'm very busy these days, and won't be able to look into this until next week.

As a quick note, my guess is that this "stereo mixing" you mention is specific to your setup, where maybe this "Stereo mixing" is something that needs to be enabled in the OS mixer to be detected as a capture device. I'm pretty sure it's not there by default pretty much everywhere, and it most definitely is not there on Linux. The only way to capture system audio, AFAIK, is by still passing through the screensharing plugin, although, as @chadfurman pointed out, this needs a change in the plugin to make that possible, and IIRC is only possible for Chrome anyway.

Anyway, this is orthogonal to this patch, as it's more likely people will want to share their mic along the screensharing. That said, we'll have to make sure we leave the door open to "capture the system audio" for the future in the janus.js API.

chadfurman:
You don't need to update the plugin just to pass in the audio flag -- just
pass in "audio" and "tab" along with "screen" in the sources array.  This
will require an update to janus.js, though, just not the chrome extension.

Yes, only possible for chrome, and you won't get a stereo mix -- you'll
just get the system output.  It won't include your microphone. You'll need
to maintain a separate channel for the audio feed from the microphone.
Play with the audio constraints in janus.js w.r.t. screensharing and the
chrome extension's callback if you want to be able to choose between
desktop audio or input device selection (via a stereo mix feed)

___________

chads.website
LinkedIn <https://linkedin.com/in/chadfurman>
Github <https://github.com/chadfurman>
Twitter <https://twitter.com/chadfurman>
___________

chadfurman:
You don't need to update the plugin just to pass in the audio flag -- just
pass in "audio" and "tab" along with "screen" in the sources array.  This
will require an update to janus.js, though, just not the chrome extension.

Yes, only possible for chrome AFAIK, and you won't get a stereo mix --
you'll just get the system output.  It won't include your microphone.
You'll need to maintain a separate channel for the audio feed from the
microphone.  Play with the audio constraints in janus.js w.r.t.
screensharing and the chrome extension's callback if you want to be able to
choose between desktop audio or input device selection (i.e. a stereo mix
feed)


___________

chads.website
LinkedIn <https://linkedin.com/in/chadfurman>
Github <https://github.com/chadfurman>
Twitter <https://twitter.com/chadfurman>
___________

lminiero:
Oh I thought that's what we got out of the discussion started in #838: can't find anything there so not sure if this happened on the group or somewhere else. Wasn't there a need to update the plugin to add another property that should have been returned in the callback?

chadfurman:
So that property, in theory, holds the value of the "audio checkbox" from
the captureDesktop UI (in Chrome) that only appears once you pass in the
"audio" source.  Interesting fact, I added it in and it was consistently
undefined.  Not sure if it's just not implemented or if I did it wrong, but
yeah it's not required to get audio working.  Only to tell if they checked
the box in the UI.

I ended up wrapping the UI in another UI, so the user sees two popups
instead of just one -- this let me control the audio source.

___________

chads.website
LinkedIn <https://linkedin.com/in/chadfurman>
Github <https://github.com/chadfurman>
Twitter <https://twitter.com/chadfurman>
___________

giordanocardillo:
The audio i intended to put along with the screen recording was only what the system has checked as default input. I didn't know there is a chrome api to pass along the tab audio. I think that if there is such a feature, one can add two audio streams to the streamed mediaElement, so the client gets audio from tab and microphone. Only problem I can see in this is the handling of two audio streams though.

chadfurman:
It's only tab audio in linux/mac -- on a PC it's full system audio, but
only on desktop share not on application share

___________

chads.website
LinkedIn <https://linkedin.com/in/chadfurman>
Github <https://github.com/chadfurman>
Twitter <https://twitter.com/chadfurman>
___________

lminiero:
@giordanocardillo works fine, but you need to remove the `muted="muted"` line from the remote video element, though, as otherwise the screensharing demo never plays audio: https://github.com/giordanocardillo/janus-gateway/blob/1f44763c73655294595c47b43d799390cd0cee45/html/screensharingtest.js#L462

As soon as you do that change I'll merge, thanks!

lminiero:
Thinking about it, it's a change so tiny that I can take care of this myself... thanks again for this effort, merging!

giordanocardillo:
Thanks you!

lminiero:
@giordanocardillo seems we still can't find your signature in our [CLA](https://cla.conf.meetecho.com/janus/), could you take care of that? Thanks!

giordanocardillo:
I submitted again the CLA

chadfurman:
@lminiero does this mean that https://github.com/meetecho/janus-gateway/pull/838 is no longer needed?

lminiero:
@chadfurman not sure, as these two PRs tried to address a different issue. This PR here adds mic audio to the screensharing stream, while #838 aimed at adding system audio instead. Apparently, as Giustino said, in some cases system audio might appear as an option when prompted for audio here, but this didn't happen in none of our tests so I don't think that's a general feature we can rely on.

PR #917, opened 2017-06-04
bower: external webrtc-adapter instead of bundled one

ancorgs:
@lminiero, on [d5a5868331d2](https://github.com/meetecho/janus-gateway/commit/d5a5868331d2e24be2e54b1f93ce74e2ac135397) you removed the bundled adapter.js and started to use the externalized one. This adapt your bower.json to that new reality (with the very same version of adapter you are using now in the demos). I tested this with Jangouts (maybe the only project relying on Bower nowadays to use Janus) and it works. It will of course break nothing for projects not using Bower.

So thanks for merging 😉

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "janus-gateway",
-  "version": "0.1.0",
+  "version": "0.2.3",
   "homepage": "https://github.com/meetecho/janus-gateway",
   "authors": [
     "Lorenzo Miniero <lorenzo@meetecho.com>",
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! Considering there's nothing I have to check, I can merge right away :+1:

ancorgs:
Wow! that was fast!

lminiero:
There was nothing I had to validate myself, so no need to wait for tomorrow when I'd be back at the office.

PR #918, opened 2017-06-04
Added changes to pause & resume recorder

AjayChoudary:
As @lminiero suggested in [discussion](https://groups.google.com/forum/#!topic/meetecho-janus/L7hwyBmrPts) , i created the pull request to pause & resume the active recording by manipulating RTP Time-stamps.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -506,7 +506,7 @@ typedef struct janus_videoroom_publisher {
 	gint64 fir_latest;	/* Time of latest sent FIR (to avoid flooding) */
 	gint fir_seq;		/* FIR sequence number */
 	gboolean recording_active;	/* Whether this publisher has to be recorded or not */
-	gboolean recording_pause;
+	gboolean recording_paused;  /* Whether this publisher recording has been paused or not */
 	gchar *recording_base;	/* Base name for the recording (e.g., /path/to/filename, will generate /path/to/filename-audio.mjr and/or /path/to/filename-video.mjr */
 	janus_recorder *arc;	/* The Janus recorder instance for this publisher's audio, if enabled */
 	janus_recorder *vrc;	/* The Janus recorder instance for this publisher's video, if enabled */
@@ -680,6 +680,8 @@ static void janus_videoroom_message_free(janus_videoroom_message *msg) {
 #define JANUS_VIDEOROOM_ERROR_NOT_PUBLISHED		435
 #define JANUS_VIDEOROOM_ERROR_ID_EXISTS			436
 #define JANUS_VIDEOROOM_ERROR_INVALID_SDP		437
+#define JANUS_VIDEOROOM_ERROR_NO_RECORDER		438
+#define JANUS_VIDEOROOM_ERROR_INVALID_RECORDER	439
 
 
 static guint32 janus_videoroom_rtp_forwarder_add_helper(janus_videoroom_publisher *p,
@@ -1176,6 +1178,7 @@ json_t *janus_videoroom_query_session(janus_plugin_session *handle) {
 						json_object_set_new(recording, "video", json_string(participant->vrc->filename));
 					if(participant->drc && participant->drc->filename)
 						json_object_set_new(recording, "data", json_string(participant->drc->filename));
+					json_object_set_new(recording, "state", json_string(participant->recording_paused ?"Paused":"Active"));
 					json_object_set_new(info, "recording", recording);
 				}
 				g_clear_pointer(&participant, janus_videoroom_publisher_dereference);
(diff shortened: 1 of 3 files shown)


lminiero:
Thanks for working on this! From a quick glance there are some things that need to be fixed, though:

1. There's a new "keyframe" request that is not needed. I see you already added an automatic FIR/PLI request when recording is resumed.
2. Pausing/resuming recordings don't check if the recording object exists, which should cause an error if not.
3. The ts-related changes in the recorder are not what I meant by keeping the RTP header consistent. That should not be done in the recorder code, but outside. Look at the calls to `janus_rtp_header_update` to see how plugins handle that for features like source switching: you'll see you'll need a context for the recorder, in case, and that's the plugin's responsibility, as the recorder should just worry about dumping packets without messing with them (and risking to break them).

AjayChoudary:
@lminiero below are my comments
1. This "keyframe" is very handy for application servers, which are using rtp-forword. Instead of blindly requesting keyframe for every 10 seconds in janus, we can trigger keyframe request from App server whenever needed. 

2. Sorry I didn't get, which part you are referring to. Mostly i added NULL checks :)

3. I have checked `janus_rtp_header_update` it only depends on SSRC, but in our case SSRC will be same. And we don't need to store all those context details. 
Pause the recording is completely independent of plug-in(audio/video rooms or record&play).
I felt its good if we are doing recorder,  post-processor is even better place.
In my use case, I am pausing all the conference handles, but if any one wants to pause only one handle then they need this info in post-processor for synchronizing with other streams. 
At least for data channel we need to touch recorder.

lminiero:
Just a ping to tell you I haven't forgotten about this PR... I'm just awfully busy this period and can't look into it yet, sorry. That said, `janus_rtp_header_update` is not only used for switching: if you simulate an SSRC change, it will force the whole context to self update, and so make the sequence number and timestamp immediately consistent. We use something like this for viewers pausing/unpausing in the VideoRoom too (although there we only update the sequence number but not the timestamp).

record.h
```diff
@@ -48,6 +48,10 @@ typedef struct janus_recorder {
 	gint64 created;
 	/*! \brief Media this instance is recording */
 	janus_recorder_medium type;
+	/*! \brief Timestaps for pause and resume the recording */
```
lminiero: pausing and resuming

record.h
```diff
@@ -48,6 +48,10 @@ typedef struct janus_recorder {
...
+	gint64 paused_ts, offset_ts;
+	/*! \brief Whether recording paused or not */
```
lminiero: Whether the recording has been paused or not

record.h
```diff
@@ -77,6 +81,11 @@ janus_recorder *janus_recorder_create(const char *dir, const char *codec, const
  * @param[in] length The frame data length
  * @returns 0 in case of success, a negative integer otherwise */
 int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint length);
+/*! \brief Pause/Resume the recorder
+ * @param[in] recorder The janus_recorder instance
+ * @param[in] boolean Whether to pause or resume
+ * @returns 0 in case of success, a negative integer otherwise */
+int janus_recorder_pause(janus_recorder *recorder, gboolean pause);
```
lminiero: I'd rather have two separate methods, `janus_recorder_pause` and `janus_recorder_resume`, as from a semantic perspective a method called `janus_recorder_pause` that can be used to resume as well is counter-intuitive.

plugins/janus_videoroom.c
```diff
@@ -506,6 +506,7 @@ typedef struct janus_videoroom_publisher {
 	gint64 fir_latest;	/* Time of latest sent FIR (to avoid flooding) */
 	gint fir_seq;		/* FIR sequence number */
 	gboolean recording_active;	/* Whether this publisher has to be recorded or not */
+	gboolean recording_pause;
```
lminiero: This is nitpicking, I know, but `recording_paused` is a better name for this. Can you also add a comment on the side, as we do for the other properties?

plugins/janus_videoroom.c
```diff
@@ -2699,7 +2699,18 @@ static void janus_videoroom_recorder_create(janus_videoroom_publisher *participa
 		}
 	}
 }
-
+static void janus_videoroom_recorder_pause(janus_videoroom_publisher *participant, gboolean paused) {
```
lminiero: See below for the semantic names. In this case, to keep things compact, the name could be changed to `janus_videoroom_recorder_status` or something like this, in order to keep the boolean flag.

plugins/janus_videoroom.c
```diff
@@ -3349,6 +3362,28 @@ static void *janus_videoroom_handler(void *data) {
 						}
 					}
 				}
+				if(record_pause) {
```
lminiero: There is no check here on whether `arc`, `vrc` or `drc` exist. Trying to pause/resume a non existing recording should return an error, while in this case you're always succeeding no matter what.

lminiero: This is important, because you may set a recording to paused and then start it, and the fact that both succeeded may make you think that you started the recording as paused, and you can resume it later, which is not the case.

plugins/janus_videoroom.c
```diff
@@ -2964,6 +2975,7 @@ static void *janus_videoroom_handler(void *data) {
 				publisher->video_active = FALSE;
 				publisher->data_active = FALSE;
 				publisher->recording_active = FALSE;
+				publisher->recording_pause = FALSE;
```
lminiero: This should be reset in `hangup_media` as well. As a side note, it might be useful to have this property returned when doing `query_session` as well, as we have a section related to recording in there when querying publishers.

plugins/janus_videoroom.c
```diff
@@ -3398,6 +3433,22 @@ static void *janus_videoroom_handler(void *data) {
...
+					JANUS_LOG(LOG_VERB, "Video keyframe , sending FIR to %"SCNu64" (%s)\n",
+						participant->user_id, participant->display ? participant->display : "??");
+					gateway->relay_rtcp(participant->session->handle, 1, buf, 20);
+					/* Send a PLI too, just in case... */
+					memset(buf, 0, 12);
+					janus_rtcp_pli((char *)&buf, 12);
+					JANUS_LOG(LOG_VERB, "Video keyframe, sending PLI to %"SCNu64" (%s)\n",
+						participant->user_id, participant->display ? participant->display : "??");
+					gateway->relay_rtcp(participant->session->handle, 1, buf, 12);
+				}
```
lminiero: Why is this request not generating any event to confirm the success?

record.c
```diff
@@ -205,17 +210,43 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
...
+		janus_mutex_unlock_nodebug(&recorder->mutex);
+		return 0;
+	} else if(recorder->paused_ts) {
+		if(recorder->type == JANUS_RECORDER_DATA) {
+			recorder->offset_ts += janus_get_real_time() - recorder->paused_ts;
+		} else {
+			recorder->offset_ts = (uint32_t) recorder->offset_ts + (ntohl(rtp->timestamp) - (uint32_t)recorder->paused_ts);
+			JANUS_LOG(LOG_INFO, "RESUME TIME TS = %u OFFSET=%u \n", rtp->timestamp, recorder->paused_ts, recorder->offset_ts);
+		}
+		recorder->paused_ts = 0;
```
lminiero: What happens here when you pause/resume more than once? Not sure this would still work. Besides, there's no fix to the sequence numbers, which means they won't be sequential, with big holes in between. This will cause severe issues when trying to replay them via Record&Play, because the recipient will think packets have been missed, and NACK them. This is one of the reasons why I'm insisting on using the RTP context here too, as it's one of the things it takes care of for you.

lminiero:
Closing as this is based on a pull request we since closed (ICE restarts).

Issue #923, opened 2017-06-14
Avoid requesting  IPv6  from libhttpd when not needed. 


sdrsdr:
In janus_http.c  at the moment there are code paths that call MHD_start_daemon with MHD_USE_DUAL_STACK option that  mandates  MHD_USE_IPv6 but if the system is not ipv6 capable the http transport fails to create any sockets and gets disabled.  However there are code paths that select MHD_USE_IPv6 based on ipv6 variable that seems quite sane. Can we please extend this to all calls to MHD_start_daemon? Is it OK for me to create a pull request on this?

lminiero:
`MHD_USE_IPv6` does IPv6 only, so is not a good option in general. Where you see it used, is in code paths where the user requested binding to a specific interface or IP address, and so in that case if either is IPv6 only then it makes sense to use it.

On a more general note, we should only do `MHD_USE_DUAL_STACK` if IPv6 was enabled when Janus was started (via command line or config). I don't think we're passing that info to transports, though (the setting is more oriented to IPv6 support for ICE), or if we do expose that info via API. In case, that's what a change or PR should address.

sdrsdr:
I my  microhttpd.h I have `MHD_USE_DUAL_STACK = MHD_USE_IPv6 | 2048,` and having MHD_USE_IPv6 in flags causes libmicrohttpd to try to create socket of family AF_INET6 and this fails on my non ipv6 enabled system. Using this simple check I'm able to start janus with example configurations  on both ipv6 and  non ipv6 enabled machines.

PR #924, opened 2017-06-14
http transport ipv6 fixes for nonipv6 systems

sdrsdr:
A pull reqest about issue #923

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -33,6 +33,7 @@
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <netdb.h>
+#include <sys/stat.h>
 
 #include <microhttpd.h>
 
@@ -254,8 +255,11 @@ static struct MHD_Daemon *janus_http_create_daemon(gboolean admin, char *path,
 	static struct sockaddr_in addr;
 	struct sockaddr_in6 addr6;
 	gboolean ipv6 = FALSE;
+	unsigned ipv6_flags=0;
+	
 	if(ip && strstr(ip, ":"))
 		ipv6 = TRUE;
+		ipv6_flags=MHD_USE_IPv6;
 	if(ip || interface) {
 		gboolean found = FALSE;
 		struct ifaddrs *ifaddr = NULL, *ifa = NULL;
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for this PR, although a system-specific check on `/proc/net/if_inet6` is not my favourite way of handling that. My guess is that we should simply add a new setting in `janus.transport.http.cfg` that says something like `ipv6 = yes|no`, so that people can specify this themselves. Ideally we'd want to re-use the one in `janus.cfg`, but as I was saying we don't have any way to get that info to transports right now without a change in the internal transport plugins API. This would free us from system-specific stuff, and allow users to configure that themselves depending on their system. I can work on this new feature later.

sdrsdr:
What if we try to enable dual stack only if specifially configured to bind to 0:0:0:0 or the ipv6 equivalent. This way the "default" config will be safer and will diable possibly incompatible behavior?

lminiero:
The default behaviour is to bind to everything, and this would be incomplete without the possibility to bind to both ipv6 and ipv4 as it does now. That's why a simple on/off flag sounds like an easier way to do this, at least for me. We can make the default for ipv6 to be `false` so that ot doesn't break on systems like yours.

sdrsdr:
Well .. trying to bind to something not supported end in the whole transport totally unavailable ... I got my problem fixed, and if I've put ip=0.0.0.0 in my config I think I can go with "stock" version so feel free to solve the problem any way you like :) I'll be glad to test the solution you provide  :)

lminiero:
That's actually a good point: we probably don't need an ipv6 flag in the config, as we can already use the interface setting to do exactly that. The default binds to everything, setting `0.0.0.0` binds to IPv4 only. I'll make this more explicit in the config file comments, so that this shouldn't be an issue in the future.

lminiero:
The IPv6 equivalent is `::`, so setting that would only bind to IPv6.

sdrsdr:
just tried today to bind to 0.0.0.0 but this failed as it is treated as invalid IP i gess binding to 192... works fine but preents 127.0.0.1 to work ..

lminiero:
Darn, and here I hoped this would be easy! :smile:

sdrsdr:
Well I can track it down for you and fix it if you like ;)

lminiero:
Yep, was just thinking of adding a manual check for "0.0.0.0" and "::" to the code, to do the right thing without too many checks. Was this what you meant by tracking down?

sdrsdr:
yes

sdrsdr:
it depends how the code is written it might be easy to be done more cleanly

lminiero:
If you have a PR ready I'll gladly accept that! :grin: 
Otherwise I can take care of that myself, no worries.

sdrsdr:
btw are the config values trimmed? as this might be hard to spot

lminiero:
Yes, the config parser trims them.

sdrsdr:
I don't have anything ready but I'm on this codebase, I can fix this at some point

lminiero:
I'm thinking to add the check at the very beginning of this branch here: https://github.com/meetecho/janus-gateway/blob/master/transports/janus_http.c#L259

Basically, if it's "0.0.0.0", fill `addr` and set `addr.sin_addr` to `INADDR_ANY`; if it's "::", fill `addr6` and set `addr6.sin6_addr` to `in6addr_any`. This should be transparent to the rest of the code, as `addr` and `addr6` are what we then pass to libmicrohttpd.

I'll work on a commit shortly so that you can test this later.

sdrsdr:
great :)

lminiero:
This commit seems to fix it for me. Noticed that if you set `::` it does NOT bind to IPv6 only, but still does IPv4 and IPv6. Apparently that's due to how dual stack systems work.

sdrsdr:
Works here also :) Good work! 👏

PR #925, opened 2017-06-15
Integration of experimental VP9 SVC support in the VideoRoom plugin

lminiero:
This PR is the result of a collaboration with @agouaillard and @murillo128 (more on that in a detailed blog post to follow!) and implements a first experimental support of [VP9 SVC](https://webrtchacks.com/chrome-vp9-svc/) in the VideoRoom plugin. This basically means that the VideoRoom plugin can receive SVC video from publishers, and selectively forward all or part of that video to subscribers.

To showcase how this works, this PR also adds a new videoroom configuration and a new demo, `vp9svctest.html`, that you can join with a VP9-SVC-compatible browser. This means starting Chrome with the `--force-fieldtrials=WebRTC-SupportVP9SVC/EnabledByFlag_2SL3TL/` flag; without that, you'll be sending plain VP9, which will make this patch useless. Once you enable that mode, publishers will send multiple layers in the room, and receivers can decide which layer they want to receive: by default you'll get everything, but you can at any time downscale/upscale spatially (will change the resolution) and temporally (will change the framerate).

This is a first step, and so there may still be bugs, but it should already pretty much work. Besides, while at the moment the layer selection is manual (and done via `configure` requests on the viewer's handle), the idea is to make this more dynamic in the future, e.g., taking advantage of incoming REMB feedback or "slow link" events to automatically drop layers in case users seem not to have enough bandwidth.

Feedback welcome!


lminiero:
After the simulcast PR, merging this one as well. I don't think there's much else to be done, here, so any fix that may be needed will only come from some real testing out there.

PR #927, opened 2017-06-20
Give ICE 5 seconds before considering failed a definitive state

lminiero:
This PR tries to address the issue @mirkobrankovic highlighted on the [group](https://groups.google.com/forum/#!topic/meetecho-janus/WZWo_YcWqqE). Now, when we detect a `failed` state, we don't do the checks we did before right there, but we start a timer: if after 5 seconds ICE is still in the `failed` state, we check if there might be info we still are missing (e.g., last trickle not received yet), otherwise we notify a definitive failure. We also have a hard 15s timeout as well, though, so that if clients are not sending a "last trickle" notification we don't end up waiting forever.

Please test and provide feedback, as I'd like to merge soon. I tested this briefly (forced a failure by providing a broken ufrag) and it seems to work for me, but I'd like to make sure this doesn't introduce issues, race conditions, or other problems.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -928,20 +928,15 @@ gboolean janus_dtls_retry(gpointer stack) {
 		return FALSE;
 	janus_ice_stream *stream = component->stream;
 	if(!stream)
-		return FALSE;
+		goto stoptimer;
 	janus_ice_handle *handle = stream->handle;
 	if(!handle)
-		return FALSE;
+		goto stoptimer;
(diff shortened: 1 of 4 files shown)


mirkobrankovic:
Thanks for quick PR, it sounded to me easy but now I see it touches a lot of parts of code.
I just tested it and it works perfectly, now in LXD containers there is plenty of time for ICE/DTLS handshake.

I tested with boring first and wanted to revert to openssl to be sure it is ok, and both worked fine for me.

lminiero:
Thanks for the feedback! I'll wait a bit longer for some more, and then I'll merge.

lminiero:
Merging.

PR #928, opened 2017-06-20
fix for proper timestamps

tgabi333:
This is fix for internal timestamp handling.

Still what janus-admin gives back is wrong. Timestamps are not in a way they usable as a unix (or javascript) timestamp

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -28,13 +28,13 @@
 gint64 janus_get_monotonic_time(void) {
 	struct timespec ts;
 	clock_gettime (CLOCK_MONOTONIC, &ts);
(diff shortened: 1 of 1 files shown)


lminiero:
This is not a fix, it's breaking timestamps and breaks all plugins using the feature :slightly_smiling_face: 
We return the value in microseconds, not milliseconds, which explains why we multiply seconds by `1000'000` and divide nanoseconds by `1000`, before summing them. In fact, when we do checks we always compare it to `G_USEC_PER_SEC`, and this change would break those checks. If you need milliseconds, divide that value by 1000.

tgabi333:
You're right, i try to figure out how to get a real timestamp in janus-admin

PR #929, opened 2017-06-21
Added support for offerless INVITEs to the SIP plugin

lminiero:
As the title suggests, this adds support for offerless INVITEs. The support is only for *incoming* INVITEs, that is those the plugin *receives*. You're still required to provide an offer when calling yourself using the plugin.

The patch also fixes some other problems here and there, mostly related to some spurious POLLERR notifications that could occur from time to time, apparently caused by ICMP errors (mostly when the peer doesn't listen on the RTCP port).

Feedback welcome, as usual, especially if you use the SIP plugin a lot and this was something you were waiting for. I tested this briefly by generating offerless INVITE calls with SIPp (media included) and it seems to be working fine for me. More comprehensive testing would help confirm this works as expected, though.

As usual, I won't wait forever, and plan to merge soon, so the sooner you shout if something's wrong, the better!

### DIFF
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -206,12 +206,18 @@ $(document).ready(function() {
 										} else if(event === 'incomingcall') {
 											Janus.log("Incoming call from " + result["username"] + "!");
 											var doAudio = true, doVideo = true;
+											var offerlessInvite = false;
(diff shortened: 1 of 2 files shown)


lminiero:
Merging.

PR #930, opened 2017-06-26
Fixed some getStats related stuff for Firefox and Safari

lminiero:
It was about time we aligned to a more up-to-date usage of `getStats`, especially considering that getting the remote bitrate in some demos (e.g., EchoTest) wasn't working with Firefox anymore. This PR takes care of that, and has a shared management of that for Chrome, Firefox and Safari. I don't have an Edge to test against, so not even sure if they support 'getStats` at all: as such, for now `getBitrate` is disabled for Edge.

I'm planning to merge soon, so please test and let us know if this breaks your janus.js-based applications.

### DIFF
--- a/html/devicetest.js
+++ b/html/devicetest.js
@@ -364,7 +364,8 @@ $(document).ready(function() {
 										echotest.send({"message": { "bitrate": bitrate }});
 										return false;
 									});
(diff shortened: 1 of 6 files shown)


lminiero:
Merging.

PR #936, opened 2017-06-29
Early media support in the SIP plugin via 183+SDP

lminiero:
As the title suggests, this patch is supposed to implement early media support when an SDP is found in a `183 Session Progress` response to an INVITE. I took advantage of this to also add some new events to the SIP plugin, namely:

- you're now notified with a `ringing` event when a `180` arrives;
- I changed the answer to offerless INVITES to `accepting`, to avoid ambiguities with the actual `accepted` later on (before you'd get two `accepted`, one without SDP and one with); 
- if a `183` arrives and it contains an SDP, you get a `progress` event with an SDP attached (the demo handles it as a regular answer, but shows a toastr message clarifying it's early media);
- if a `200 OK` arrives after a `183`, you get an `accepted` event *without* SDP (you already received it before), meaning this is only there to confirm the call has been accepted.

Notice that, with early media, we only send the browser the SDP from the 183, *not* the 200, as we don't do renegotiations on the WebRTC side yet. This means that if the 183 SDP used sendonly media to play a message, and then the 200 OK turns it into sendrecv, this won't work, as we didn't get the microphone in the browser.

Tested this briefly and this seems to work, but please test and provide feedback to make sure this doesn't introduce any regression.


lminiero:
Seems to work fine, so I'll merge.

PR #942, opened 2017-07-04
Use tracks instead of streams in janus.js

lminiero:
This is an attempt to revive the effort @joshdickson40 made on #894, to support tracks instead of streams, as some methods are going to be deprecated. Namely, the changes in `janus[.nojquery].js` are:

1. changed `addstrem` to `addtrack`;
2. changed `onaddstrem` to `ontrack`;
3. removed `stream.stop()` and only left `track.stop()` (which we already had anyway).

These changes may NOT be transparent to the application, and definitely wasn't for some of the demos. In case the remote stream contains both audio and video, your `onremotestream` will be invoked twice. If all you do is passing the stream to the HTML5 video element, then you're fine: even if you don't set it the second time, the old stream autoupdates the HTML5 video element. If you're doing more than that, though (e.g., any setup or UI modification), you'll need to catch the fact the callback has already been invoked, and react accordingly (even just to return right away, in order not to setup some handlers twice).

To make a practical example, in some of the demos we check if the remote stream has video, and if not we put a big "no video" video placeholder. Since the first time the callback is invoked the stream only has an audio track, then that's always true. As such, when the callback is invoked the second time, we just check if the stream now contains a video track: if it does, we get rid of the placeholder and show the video instead. Just check the diffs in the demos to see how I did it for EchoTest, VideoCall, VideoRoom, etc.

If your question now is "why didn't you make it transparent for me, you bastard?", that wasn't possible unfortunately. The `ontrack` callback fires asynchronously, and we have no way to know in advance whether or not an audio or video track will really be available, or when. As such, that's up to you to do in the callback: should be trivial enough to do anyway (at least, it was in my demos).

Seems to work fine with Firefox and Chrome. Haven't tested with Edge or Safari. Pinging some of the folks that I know will definitely be interested in checking if this would break their apps, like @ancorgs @chadfurman and @thehunmonkgroup (I'm surely forgetting someone so feel free to mention others that may be interested).

Feedback welcome, as usual!


joshdickson40:
Thank you for picking this effort up, Lorenzo. Out of curiosity what was the issue in Firefox that was causing problems for some folks? I had attempted replication but I could not get the failure that was described.

lminiero:
I think it was related to the not-transparent impact on the demos described
above. The echo test demo is one of those that uses the no-video
placeholder.

Il 04 lug 2017 7:50 PM, "Joshua Dickson" <notifications@github.com> ha
scritto:

soulfly:
@lminiero could you create a room for VideoRoom plugin with H264 codec?
for now, there is a room 1234 but with VP8

so people will be able to play more with Safari and this PR

thanks

lminiero:
@soulfly sorry, not going to do anything to foster a wider adoption of H.264 over VP8. It's a question of principle, and you may think it's silly, but I don't like the move of big players to try and force H.264 (a royalty encumbered codec) on us. If you want to test this agains H.264 only clients, please install your own deployment and configure the room to be H.264 there.

soulfly:
@lminiero that's understandable, thanks. Will play with local installation then

lminiero:
Any feedback on this? I'd like to merge ASAP, but at the same time I'd love for this to work in apps that are not just my demos :wink:

chadfurman:
@lminiero I'll be sure to include this in the new version of the application.  We're building that bit out towards the latter end of next month.

MvEerd:
I've merged this and haven't noticed any issues on that service as a result

lminiero:
@MvEerd thanks for the feedback!

lminiero:
Since I haven't seen any objection, I'll probably merge tomorrow or the day after that (not doing it now as I'll have to align a couple of PRs after the merge and I don't have time at the moment).

lminiero:
Made a couple of last minute tweaks (the new VP9 SVC demo still didn't include the changes we made to other demos, and now the adapter we use is the latest, 5.0.1), merging now.

PR #943, opened 2017-07-05
Allow to override build date

bmwiedemann:
in order to allow for reproducible builds.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

Note: This variant will only work with GNU date. If BSD support is important, I can add this.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -164,8 +164,14 @@ dir_target = $(directory)-$(wildcard $(directory))
 dir_present = $(directory)-$(directory)
 dir_absent = $(directory)-
 
+if WITH_SOURCE_DATE_EPOCH
+build_date = $(shell LC_ALL=C date --utc --date="@$(SOURCE_DATE_EPOCH)")
+else
+build_date = $(shell date)
+endif
+
 version.c: FORCE | $(dir_target)
-	date | awk 'BEGIN {} {print "const char *janus_build_git_time = \""$$0"\";"} END {} ' >> version.c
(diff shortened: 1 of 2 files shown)


ancorgs:
It's worth noticing that this bug is desired for openSUSE and it's already being applied (via patch) in Debian, as you can see [here](https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=865554;filename=janus.diff.txt;msg=5). So it would be nice to have this "upstreamed" so distributions caring about reproducible builds can stop patching the files.

lminiero:
Will the default still be "use the current date", or is your change overriding that? Sorry for the possibly dumb question, but I'm not familiar with this mechanism.

bmwiedemann:
The default is still 'current date' as before.
Only if `$SOURCE_DATE_EPOCH` is set (e.g. happens when building packages for the openSUSE Tumbleweed distribution) it will use that date.

lminiero:
Thanks for the clarification! If you sign our [CLA](https://cla.conf.meetecho.com/janus/) I'll merge right away then :+1:

bmwiedemann:
signed the CLA.
Note: much of the patch was written by a Debian contributor (which is why I specified him as Author in the git commit header).
OTOH, it is below 10 lines and so much copy-paste from the reference code snippets that it is probably not copyrightable anyway.

lminiero:
Got it, thx: it's just that for all merges we still need a CLA signature anyway. Merging!

PR #944, opened 2017-07-05
Experimental support for VP8 simulcasting

lminiero:
After working on VP9 SVC in #925, I decided to have a look at VP8 simulcasting as well, as that's been around for a while longer, and already quite used. I found it much more challenging and frustrating than VP9 SVC (you can read about some of my doubts [here](https://groups.google.com/forum/#!topic/discuss-webrtc/czRbYBm1u2g)), but I managed to get something working. Notice that this is not complete, and is far from perfect, but seems to mostly work, so I decided to start sharing it with you guys, so that you could start playing with this too.

This patch tries to support the simulcasting support of both Chrome and Firefox, which implement it in different ways. From a `janus.js` perspective, you enable simulcasting by passing a `simulcasting: true` when doing a `createOffer` (no way to do it via `createAnswer` instead, maybe in the future). Depending on whether Chrome or Firefox is used, this triggers the right behaviour, respectively SDP munging and RID configuration. Both are WIP (especially the RID stuff and the related settings) but work fine for testing. Janus is able to then detect simulcasting has been activated on the PeerConnection, and notifies the plugin accordingly.

It is currently integrated only in the EchoTest/DeviceTest and VideoRoom plugins, and I updated the demos to use it. Both plugins are configured to expose some controls to choose which substream and/or temporal layer you want to receive: as a result, when incoming packets arrive, the plugin can use that information to choose which packets to discard and which to let through, fixing RTP headers and VP8 payload descriptors along the way. Once a change occurs, the user is notified, so that they can update the UI: in my demos, for instance, I use the first "change" event to unhide the controls, and update the buttons state (as a click may not result in a change right away). Apparently the temporal layer selector only works for Chrome (not sure Firefox implements temporal scalability), but the substream selector works for both. Of course, some substreams will only appear when the sender has enough bandwidth, so you'll have to play with the bandwidth know to trick the sender via REMB.

Notice that at this stage the changes are manual only: the plugin will only go from a substream to another if you send the related request, there's no automated mechanism. That's obviously planned for the future, but I want to be sure the basics work first. This includes realizing a substream you're receiving is not available anymore, e.g., because the browser decided to stop sending it (e.g., not enough bandwidth or excessive CPU): currently, this will result in a freeze in the video, so just click a lower layer and it should unfreeze. This is still not working as expected in all cases (e.g., Firefox seems to be sluggish when substream 0 is sent, and I see NACKs that shouldn't be there), but it's already usable in both demos.

That should be all. Please play with this, and maybe play with the RID configuration in `janus.js` as well to see if you can come up with better results. Feedback welcome!


lminiero:
Just FYI, this is how it looked like in one of my earler tests, with Chrome publishing in the VideoRoom and subscribing to itself with different tweaks :wink: 

VP8 simulcasting screenshot

fippo:
Firefox 55 has temporal scalability: https://bugzilla.mozilla.org/show_bug.cgi?id=1343143

lminiero:
There's something I'm doing wrong then, as changing temporal layers seems
to have no effect on my Firefox 56... One more thing to look into!

Il 05 lug 2017 8:01 PM, "Philipp Hancke" <notifications@github.com> ha
scritto:

plugins/janus_videoroom.c
```diff
@@ -1152,12 +1163,16 @@ json_t *janus_videoroom_query_session(janus_plugin_session *handle) {
...
 					json_t *recording = json_object();
 					if(participant->arc && participant->arc->filename)
 						json_object_set_new(recording, "audio", json_string(participant->arc->filename));
-					if(participant->vrc && participant->vrc->filename)
-						json_object_set_new(recording, "video", json_string(participant->vrc->filename));
+					if(participant->vrc[0] && participant->vrc[0]->filename)
+						json_object_set_new(recording, "video", json_string(participant->vrc[0]->filename));
+					if(participant->vrc[1] && participant->vrc[1]->filename)
+						json_object_set_new(recording, "video-sim1", json_string(participant->vrc[1]->filename));
+					if(participant->vrc[2] && participant->vrc[0]->filename)
```
MvEerd: Did you intend to put ```participant->vrc[2]->filename``` here instead?

lminiero: Yep, definitely a typo, thx for spotting this!

As a side note, we'll really have to decide what the best way to handle recordings and RTP forwarding will be, with Simulcasting enabled. Both are not an issue when we receive a single stream, but can be with three different substreams that you're not sure will all arrive. At the moment I'm simply forwarding the lowest substream when RTP forwarding (as it's the only one we know for sure is always sent) while for recording I'm saving each substream to a different file. That's definitely not my favourite approach, as while it's very useful for debugging (I can check what was sent for each of them), it much more problematic to handle when postprocessing (how do you "merge" them to be sure you always have the "best quality" at any moment?).

Probably I'll have to treat both as a "user with no trouble at all", and find a way to automatically drop or go up substreams as I notice them appearing/going away.

lminiero:
Made a lot of changes and fixes in these last few days, and the code works much better than it did when I first published this PR. If this effort interests you, I encourage you to start playing with it, as I plan to merge soon now so that this can be used in real scenarios as well. This will hopefully give us more info to make it work even better (as it's still not as good as I'd like it to be).

The demos (the ones in the PR, *NOT* the ones online... those are still outdated) now allow you to conditionally enable simulcasting, depending on whether you pass `?simulcast=true` as the query string. So `echotest.html` will be the regular EchoTest demo, while opening `echotest.html?simulcast=true` will use publish simulcast video in the demo. This is available for the VideoCall and VideoRoom demos as well (but not others, more on this below). The code to create and handle the simulcast UI (the buttons and notifications) has also been separated, so that it should be much easier to integrate in your own application, for instance.

I also implemented some automatic transitions depending on the available substreams. To make a simple example, when you try the EchoTest demo, the plugin automatically aims for the higher quality, and so assumes you want substream 2 and all the temporal layers. As such, it starts giving you what's available right away (substream 0), and then, as soon as substream 1 arrives, switches to that. Should substream 2 arrive, that's what you'd get then. If while you're receiving a substream, it becomes unavailable, the plugin drops to the substream below. This is done by a brutal check on timing: if we didn't receive data for a substream in 250ms, then we considering it gone for now (can probably be improved but seems to work). As soon as it becomes available again, we start sending it again. You can verify this dynamic behaviour by playing with the bandwidth indicator: setting it to 512kbps will eventually make substream 1 appear; 1.5mbps will get substream 2 working; going below again will drop some. This seems to work fine both when Chrome is simulcasting, and when Firefox is (although Firefox 56 seems to work much better than 54 in the simulcasting support). Of course, you can at any time set a cap yourself: if you explicitly set substream 1 in the buttons, the plugin will never give you substream 2 even if it's available. This will be very helpful for those scenarios where you're explicitly interested in getting a lower quality for some feeds (e.g., big video + many thumbnails).

Besides, the EchoTest, this dynamic behaviour is also available in the VideoCall and VideoRoom. Depending on how you configure the bandwidth for the sender, the available substreams for the peer (VideoCall) and viewers (VideoRoom) will change, and the mechanism described above will kick in. Notice that, in the VideoRoom plugin, simulcasting can obviously only be enabled for publishers: viewers will simply have the opportunity to switch between the available substreams, and it will work for them whether the browser supports simulcasting or not (everything is done for them within Janus). In the VideoCall, instead, it only works for the caller in Chrome, but not the callee: in fact, I tried enabling simulcasting in a createAnswer and, while it didn't fail, it resulted in no video being sent by the browser, so I can only assume it's not supported. I haven't tested with Firefox, but I don't see why it shouldn't work there.

I made some changes to the way recording simulcasted PeerConnections works too. For EchoTest and VideoCall, considering there's just a single peer, this was easy: whatever we would send the peer (or back), depending on the simulcast logic we would record. In the VideoRoom this is more problematic, though: we have different viewers, with different requirements and constraints, which means that any of them may possibly receive different stuff. As such, for now the VideoRoom plugin only records the base substream (substream 0) as that's the only one we have always available. RTP forwarding works differently, instead: rather than sending a single video stream out as we do now, e.g., by choosing the best quality possible, we send the three substreams to separate ports, and leave the logic (and the burden...) to the receiver application. It's this application that can decide what to drop and what to keep, and how to use it. As I'll explain below, one use case for this is the Streaming plugin.

As to the other plugins (those that involve video, so no AudioBridge, TextRoom or VoiceMail), as just introduced I added support for watching simulcasted streams in the Streaming plugin, while I only added a placeholder in the SIP and Record&Play plugin. By placeholder I mean that the plugins can detect simulcasting is being used by the peer, but they explicitly drop anything beyond substream 0. This is because of the nature of both plugins: in the SIP plugins we don't really have a feedback mechanism to allow the peer to choose what they should get, while in the Record&Play plugin we'd rather have a continuous recording that something that may have pauses in it (as always saving the best quality means detecting when higher substreams become unavailable, and so potential 250ms holes in there at times). In the Streaming plugin, instead, simulcasting works pretty much as the "switch" command already worked, with the difference that you can bind to three different ports in the *same* mountpoint: this is particularly useful, for instance, if you're RTP forwarding a simulcast VideoRoom publisher (for whom we'd send three streams out, one for each substream), as you can then have a mountpoint to restream them that way. Anyway, it's also useful if you're generating different qualities of the same stream yourself (e.g., via gstreamer or ffmpeg), as the simulcast mechanism could be used in that context too. This is an example of how you can configure a simulcast mountpoint (notice `videoport2` and `videoport3`):

```
[simulcast-sample]
type = rtp
id = 51
description = Simulcast test
audio = yes
video = yes
audioport = 6002
audiopt = 111
audiortpmap = opus/48000/2
videoport = 6004
videosimulcast = true
videoport2 = 6006
videoport3 = 6008
videopt = 100
videortpmap = VP8/90000
secret = adminpwd
```

and how you can RTP forward a VideoRoom publisher that is simulcasting to that mountpoint:

```
sfutest.send({message: {
    request: "rtp_forward",
    room: 1234,
    publisher_id: myid,
    host: "127.0.0.1",
    audio_port: 6002,
    video_port: 6004,
    video_port_2: 6006,
    video_port_3: 6008,
    secret: "adminpwd"
}, success: function(json) { console.log("Got response:", JSON.stringify(json)) }});
```

Opening that mountpoint in the Streaming demo will display the usual additional simulcast buttons, and the same "automatic transition" mechanism will try to kick in there too. Obviously, considering the source of the media is not known to the Streaming plugin and we can't send keyframe requests when we need to, the experience will be a bit different there (transitions may take a little longer), but still it should do what it's supposed to do for the most part.

That said, while I added and fixed a lot, there still is some stuff missing. We're not properly terminating RTCP for the simulcasted substreams yet, for instance, which means we don't collect stats for those. The same applies to NACKs and stuff like that. The automatic transition mechanism may be further improved by feedback we get from the recipient (RTCP losses, REMB), something I haven't done yet (this is still left to the controlling application, e.g., you're notified of losses, you ask for a lower substream yourself). Besides, I still haven't managed to get temporal scalability to work with Firefox: @fippo linked to an issue page explaining it should be available since 55, but my 56 definitely doesn't generate any temporal layer but 0 (verified with debugging); maybe I'm doing something wrong or there's some configuration I'm missing when setting the RIDs, something I'll ask about on the Mozilla group.

That's all for now, but should already give you something to play and experiment with. Although I'm flying to Prague tomorrow, where I'll be busy streaming the IETF meeting without much time to do anything else, I look forward to useful feedback you guys may give on this!

lminiero:
Well, since apparently nobody will look at this until they'll have to, I'll merge it :wink: 
Hopefully this will result in more people playing with this and providing feedback.

PR #954, opened 2017-07-14
Add libressl support

aduskett:
This fixes some preprocessor checks that determine whether
or not to use OpenSSL 1.1 or 1.0.x. Because LibreSSL returns 2.x
the macros think that OpenSSL 1.1 should be used. This patch
checks to see if there is an LIBRESSL_VERSION_NUMBER defined
as well.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -106,7 +106,7 @@ void *janus_dtls_sctp_setup_thread(void *data);
 #endif
 
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
 /*
  * DTLS locking stuff to make OpenSSL thread safe (not needed for 1.1.0)
  *
@@ -309,7 +309,7 @@ error:
 /* DTLS-SRTP initialization */
 gint janus_dtls_srtp_init(const char* server_pem, const char* server_key) {
 	const char *crypto_lib = NULL;
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
 	crypto_lib = "OpenSSL pre-1.1.0";
 	/* First of all make OpenSSL thread safe (see note above on issue #316) */
 	janus_dtls_locks = g_malloc0(sizeof(*janus_dtls_locks) * CRYPTO_num_locks());
@@ -328,7 +328,7 @@ gint janus_dtls_srtp_init(const char* server_pem, const char* server_key) {
 	JANUS_LOG(LOG_INFO, "Crypto: %s\n", crypto_lib);
 
 	/* Go on and create the DTLS context */
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
 	ssl_ctx = SSL_CTX_new(DTLSv1_method());
 #else
 	ssl_ctx = SSL_CTX_new(DTLS_method());
--- a/dtls.c
+++ b/dtls.c
@@ -309,8 +309,12 @@ error:
 /* DTLS-SRTP initialization */
 gint janus_dtls_srtp_init(const char* server_pem, const char* server_key) {
 	const char *crypto_lib = NULL;
-#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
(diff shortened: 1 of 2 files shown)


dtls.c
```diff
@@ -309,7 +309,7 @@ static int janus_dtls_load_keys(const char* server_pem, const char* server_key,
 /* DTLS-SRTP initialization */
 gint janus_dtls_srtp_init(const char* server_pem, const char* server_key) {
 	const char *crypto_lib = NULL;
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
 	crypto_lib = "OpenSSL pre-1.1.0";
```
lminiero: Couldn't this check also be used to change the `crypto_lib` string content? This is what we use as a display value in the Janus logs, and so may be helpful to make it explicit when LibreSSL is in use.

aduskett:
Hey! Thanks for the quick reply!

LibreSSL defines OPENSSL_VERSION_NUMBER as	0x20000000L.
However; LibreSSL is only API compatible with OpenSSL 1.0.x (see the issue?).
However, there is an LIBRESSL_VERSION_NUMBER which will return 0x2050500fL (latest version.)

Easy check to make for now. What do you think?

dtls.c
aduskett: It sure could!

aduskett:
Not sure.  Would it be possible to check for LIBRESSL_VERSION_NUMBER using configure.ac?
I am an expert with autotools.  

Running openssl version returns LibreSSL as well on my machine if that helps.

aduskett:
I added some small amount of logic in dtls.c to print out libressl if it's being built against it.  Good idea!

lminiero:
Looking around, it seems that one of the things LibreSSL did was removing some methods, like `RAND_egd`, so checking for those might help craft the summary. Anyway, we can worry about that after this is merged.

As a side note, can you sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge as soon as this has been tested? Thanks again!

dtls.c
```diff
@@ -309,8 +309,12 @@ static int janus_dtls_load_keys(const char* server_pem, const char* server_key,
 /* DTLS-SRTP initialization */
 gint janus_dtls_srtp_init(const char* server_pem, const char* server_key) {
 	const char *crypto_lib = NULL;
-#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if defined(LIBRESSL_VERSION_NUMBER)
```
lminiero: Won't this need an extra `#endif`? Not sure as I can't try compiling myself right now.

aduskett: There is.  It compiles fine. :)

#if OPENSSL_VERSION_NUMBER < 0x10100000L
#if defined(LIBRESSL_VERSION_NUMBER)
	crypt_lib = "LibreSSL"
#else
	crypto_lib = "OpenSSL pre-1.1.0";
#endif
	/* First of all make OpenSSL thread safe (see note above on issue #316) */
	janus_dtls_locks = g_malloc0(sizeof(*janus_dtls_locks) * CRYPTO_num_locks());
	int l=0;
	for(l = 0; l < CRYPTO_num_locks(); l++) {
		janus_mutex_init(&janus_dtls_locks[l]);
	}
	CRYPTO_THREADID_set_callback(janus_dtls_cb_openssl_threadid);
	CRYPTO_set_locking_callback(janus_dtls_cb_openssl_lock);
#else
	crypto_lib = "OpenSSL >= 1.1.0";
#endif

lminiero:
Note to self, for detection in configure.ac: https://gist.github.com/Sp1l/e3c709e870285c2cacc6

aduskett:
I signed the form as well.  Thanks so much for being so responsive!

lminiero:
Likewise! I'll try to have a quick run at the patch later or tomorrow, and in case everything seems fine I'll merge :+1:

aduskett:
Completely unrelated:

I wish there was a pjsip plugin. :(

lminiero:
> I wish there was a pjsip plugin. :(

I considered that as an alternative to the Sofia SIP plugin, but ended up working on a re-based plugin instead, which is available as another pull request. Working on a pjsip plugin as well would take too much energy we can devote on other activities, especially considering we already have SIP working in other ways and so may not be worth it. Of course, if somebody works on that we won't say no :smile:

lminiero:
Sorry for the lack of feedback, I've been abroad for the IETF and have just come back... I'll give this a go today.

lminiero:
Looks good, merging! I'll add a check in configure.ac for the summary, as discussed.

PR #955, opened 2017-07-16
infra: add support for building on travis-ci

fippo:
adds a .travis.yml which allows building on https://travis-ci.org/ on every pull request.

job output is here: https://travis-ci.org/fippo/janus-gateway/jobs/254195305
After logging in to https://travis-ci.org/ one can enable build on pushes or pull requests, see [their getting started guide](https://docs.travis-ci.com/user/getting-started/)

(yes, I really did not want to install dependencies locally... :-))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 dist: trusty
 sudo: false 
-language: node_js
+language: c
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+dist: trusty
+sudo: false 
+language: node_js
+
+env:
+  global:
+    - CPPFLAGS=-I${HOME}/srtp/include
+    - LDFLAGS=-L${HOME}/srtp/lib
+
+addons:
+  apt: 
+    packages:
+      - build-essential 
+      - libmicrohttpd-dev
+      - libjansson-dev
+      - libnice-dev
+      - libssl-dev
+      - libsrtp-dev
+      - libsofia-sip-ua-dev
+      - libglib2.0-dev
+      - libopus-dev
+      - libogg-dev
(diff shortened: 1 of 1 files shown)


lminiero:
Sorry for the lack of feedback, didn't have much time to look into this! I'll check if this is similar to what a colleague of mine already started doing some time ago, and look into this after the holidays :+1:

tgabi333:
I would add websocket support too (and cache downloadables)

emmanuelriou:
I would remove libsrtp-dev package as you install it manually in before_script

emmanuelriou:
build-essential, pkg-config, libtool and automake packages are also not necessary as already installed in trusty image.

lminiero:
@tgabi333 @emmanuelriou can you help tweak the Travis script per your suggestions, should I merge this? I'm not familiar with it, unfortunately, so until I study how it works I'll take all the help I can get :smile:

emmanuelriou:
hi @lminiero , I would just remove from the packages section:
- build-essential
- libsrtp-dev
- pkg-config
- libtool
- automake

as those packages are all by default already installed in trusty base image.

As @tgabi333  suggested, if you want also to test the build of all transport plugins, you might want to add the necessary packages for them, like websocket.

cmacq2:
@fippo 
@emmanuelriou 

Of course I only see this PR *after* I've submitted [my own](https://github.com/meetecho/janus-gateway/pull/1104)... Personally I would still prefer the Docker image based version for the reason that it is (re-)usable outside of Travis, notably also for local development.

lminiero:
Closing as this dates back to a couple of years ago, and there have been changes to the build process. I'll take a mental note that we may need to add a travis-ci build script in the future, but not sure when I'll be able to do that (we currently don't rely on that). Thanks @fippo!

PR #957, opened 2017-07-20
Avoid slow reverse-DNS requests caused by libwebsocket.

atoppi:
According to https://github.com/warmcat/libwebsockets/issues/537 and following my own experience, in some circumstances `lws_get_peer_addresses` can take several seconds to execute a reverse DNS request on a connected peer IP.  The effect is that sometimes a websocket connection with Janus takes several seconds before it is established.
This PR addresses the described issue by replacing `lws_get_peer_addresses` with `lws_get_peer_simple` that completely skips the RDNS request.
Some macros have been added too, in order to check the availability of the API at compile time (it should be available for lws > 2.0) and fallback to the previous implementation eventually.


lminiero:
Thanks! Waiting for some feedback before we merge this.

lminiero:
I see no objection, so merging.

PR #961, opened 2017-07-31
Honor trickle configuration when using ext. stream

longsleep:
The check for config.trickle was not done when an external stream is passed. This moves the check/set for config.trickle before the check for external provided stream, making it possible to disable ICE trickle when the external application provides the stream.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1374,6 +1374,7 @@ function Janus(gatewayCallbacks) {
 			}
 			return;
 		}
+		config.trickle = isTrickleEnabled(callbacks.trickle);
 		// Was a MediaStream object passed, or do we need to take care of that?
 		if(callbacks.stream !== null && callbacks.stream !== undefined) {
 			var stream = callbacks.stream;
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks for spotting this! Could you do the same fix in `janus.nojquery.js` as well? As soon as both files are updates and you sign our [CLA](https://cla.conf.meetecho.com/janus/) I'll merge this :+1:

longsleep:
Updated PR to change both files. Will look at the CLA - but this change is trivial so you might be able to pick it without CLA.

lminiero:
Let me know if you'd rather not sign and I'll incorporate myself.

longsleep:
> Let me know if you'd rather not sign and I'll incorporate myself.

Sure i will sign, just needs a while until i have management approval. I hope i will have it by tomorrow.

longsleep:
> Sure i will sign, just needs a while until i have management approval. I hope i will have it by tomorrow.

I just signed the CLA - please merge and close :)

lminiero:
Thanks, merging then!

PR #962, opened 2017-08-01
mutex: add define for PTHREAD_MUTEX_INITIALIZER

tmatth:
Used for initializing statically allocated mutexes.


lminiero:
Not sure what this is used for, in the code? We already have a `janus_mutex_init` = `pthread_mutex_init` for that.

lminiero:
Much clearer from the examples now, thanks, merging!

PR #971, opened 2017-08-10
Explicit mode non-interleaved

uasan:
Some IP cameras, by default send RTP in TCP connection (interleaved mode), this fix to UDP (non-interleaved mode)

https://www.ietf.org/mail-archive/web/mmusic/current/msg02030.html

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -3444,9 +3444,9 @@ static int janus_streaming_rtsp_parse_sdp(const char *buffer, const char *name, 
 	} while(port == -1);
 
 	if(IN_MULTICAST(ntohl(mcast))) {
-		g_snprintf(transport, 1024, "RTP/AVP;multicast;client_port=%d-%d", port, port+1);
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for this effort!

Before I merge this, I'd like to wait for some feedback, as I want to make sure it doesn't break it for anybody who's currently using the RTSP functionality on a regular basis (I personally don't). Can you advertise your PR on the [https://groups.google.com/forum/#!forum/meetecho-janus] to make people aware of this, so that they can test it in their setups?

lminiero:
PS: can you also sign our [CLA](https://cla.conf.meetecho.com/janus/)?

lminiero:
Thanks for signing! I just mentioned your PR [here](https://groups.google.com/forum/#!topic/meetecho-janus/_dgbhatsiV0), hoping that people using RTSP will test this. If we don't get any negative feedback, I'll merge in a few days.

uasan:
Ok

PR #973, opened 2017-08-11
Update README.md

mo8Zomo0:
Thank you for the great and easy to follow guide, I noticed some small things:

cmake, I added it to the aptitude line, not sure how it is with yum, possibly the same, but that should be fixed.
also not all use or have aptitude installed, apt-get might also do the trick.

I am not sure if people want additional installs under /usr or if generic mentioned path should better be /usr/local

The make instruction for janus differs from the others, it also lacked the sudo (also for the config)

Thank you for your work, and now lets try your GW!


lminiero:
cmake is not needed for Janus: you only need it if you want to compile libwebsockets yourself (which you might avoid if it's available in the repos). As such, I don't think we should add it to the explicit dependencies.

lminiero:
Closing for the reasons above. Thank anyway for contributing!

PR #984, opened 2017-08-30
Fixed inconsistent event naming for plugins event notifications

hijaq:
Most of plugins' event names in event notifications are named as `event`, but a few of them (probably because of copy and paste from the preceding code) have different names, like `status` or `videoroom`.

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -2486,7 +2486,7 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 							/* Also notify event handlers */
 							if(notify_events && gateway->events_is_enabled()) {
 								json_t *info = json_object();
-								json_object_set_new(info, "audiobridge", json_string(participant->talking ? "talking" : "stopped-talking"));
+								json_object_set_new(info, "event", json_string(participant->talking ? "talking" : "stopped-talking"));
 								json_object_set_new(info, "room", json_integer(participant->room->room_id));
 								json_object_set_new(info, "id", json_integer(participant->user_id));
 								gateway->notify_event(&janus_audiobridge_plugin, session->handle, info);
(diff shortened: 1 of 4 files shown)


lminiero:
Not sure I can accept this contribution, sorry. Inconsistencies are only relevant when they're in the same plugin: different plugins, instead, don't need to adhere to any specific convention, and they're free to implement any API they want. Besides, this PR would break all demos (and all user applications out there), as they expect events to be named a different way.

hijaq:
It's in the same plugin. For example, videoroom plugin:

`event` events are `created`, `edited`, `destroyed`, `published`, `unpublished`, `subscribed`, `unsubscribed` and all of them has the following structure:

```json
{
  "type": 64,
  "timestamp": 1483228800,
  "session_id": 2004798115,
  "handle_id": 3708519405,
  "event": {
    "plugin": "janus.plugin.videoroom",
    "data": {
      "event": "destroyed",
      "room": 1
    }
  }
}
```

and `videoroom` events (actually it's the same place in the code) are `talking` and `stopped-talking` with the following structure:

```json
{
  "type": 64,
  "timestamp": 1483228800,
  "session_id": 2004798115,
  "handle_id": 3708519405,
  "event": {
    "plugin": "janus.plugin.videoroom",
    "data": {
      "videoroom": "stopped-talking",
      "room": 1,
      "id": 12345
    }
  }
}
```

lminiero:
It's still an issue in terms of backwards compatibility. At the very least, both event names should be present for the time being, so that there is a transition process between a deprecated event and the real name. The changes could then be advertised on the group, so that people have time to adapt should they switch to master.

lminiero:
Closing as it currently breaks compatibility, and as discussed plugins don't really need to be aligned in terms of the APIs they expose. Please feel free to open a discussion on our Google group if you'd like to talk about this with the other users. Thanks!

PR #988, opened 2017-09-01
Websocket default values timout as HTTP

uasan:
Websocket default timout values, harmonized with HTTP default timout values

50s + 10s = 60s (default HTTP timeout)

### DIFF
--- a/conf/janus.transport.websockets.cfg.sample.in
+++ b/conf/janus.transport.websockets.cfg.sample.in
@@ -3,8 +3,8 @@
 [general]
 json = indented				; Whether the JSON messages should be indented (default),
 							; plain (no indentation) or compact (no indentation and no spaces)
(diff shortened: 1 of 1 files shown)


lminiero:
Not sure we need to change those values, as they're commented out and not even the defaults. That said, 50+10 are way to close to the 60s timeout and so would not be a good idea.

uasan:
> Not sure we need to change those values, as they're commented out and not even the defaults. 

I assure you, most will delete comment and will not change the values, so in comments you need to show the optimal values :)

lminiero:
Got it, but 50+10 is still way too high and close to the 60, considering there's RTT and random TCP/signalling delays to take into account. I'd make it 30/10 or 40/10.

uasan:
Ok

lminiero:
Closing as the changes are so minimal I can make them myself :smile: 
I'll set 30/10 in the example, just to keep it safe.

Issue #879, opened 2017-05-01
janus.js should not depend on JQuery


cmacq2:
The core `janus.js` library need not depend on JQuery.

Currently JQuery is used for two purposes: `$.ajax()` requests to Janus gateway and to check an element for a Chrome extension id/marker value. In modern and even quite a few dated browsers, `$.ajax()` can be replaced with [`fetch()`](https://developer.mozilla.org/en/docs/Web/API/Fetch_API) and the check for the marker could be rewritten quite straightforwardly with [`document.querySelector()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector)

Benefits are:

- One dependency less for the core `janus.js` library (and there are polyfills for folks who need backwards compatibility with browsers that lack a native  `fetch()` implementation). 
- With `fetch()` the code is typically quite a bit more compact, and it would yield easier to test code compared to nested JQuery callbacks resulting from nested `$.ajax()` calls.

Would you be interested in merging & carrying a PR that addresses this?

cmacq2:
Yeah that's almost but not quite what I'm after. ;) I'm proposing to use the native API of modern browsers, which also happens to give you a nicer async programming model (Promises instead of callbacks in callbacks). In fact, it's quite likely that in the browser you are using right now, XHR is actually a backwards compatibility, make-believe shim layered on top op `fetch()`.

cmacq2:
Also, what I somehow forgot to mention: neither `janus.js` nor `janus.nojquery.js` makes it easy to augment outgoing requests or intercept responses  (for example to work with custom HTTP headers). If the `gatewayCallbacks` parameter to the `Janus` constructor was extended with a couple of optional arguments/options, code like this could be the basic pattern

```
var outgoingRequestFilter = ... // passed in or just identity function to continue
var responseInterceptFilter = ... // passed in or just identity function to continue

function doFetch(url, options) {
  return new Promise(function(resolve) {
    resolve({url: url, options: options});
  })
  .then(outGoingRequestFilter || function(x) { return x; })
  .then(function(request) {
    if(request) {
      return fetch(request.url, request.options);
    }
  })
  .then(function(response) {
    if(response.ok) {
      return response;
    }
    else { // HTTP request succeeded, but HTTP status code indicates an error response
      return Promise.reject({
        status: response.status,
        statusText: response.statusText,
        response: response
      });
    }
  })
  .then(responseInterceptFilter || function(x) { return x; })
  .then(function(response) {
[...]

PR #991, opened 2017-09-04
Decouple `janus.js` from its dependencies.

cmacq2:
With this change client code may choose to override dependencies by providing a suitable shim/facade to map APIs:

 - isArray: check whether or not something is a JS array. You probably should use Array.isArray() here.
 - checkJanusExtension: check whether or not the Janus screen sharing extension for Chrome is installed.
   This is equivalent to checking whether or not an element with id=janus-extension-installed attribute exists.
 - webRTCAdapter: something equivalent to the adapter object from the webrtc-adapter library.
 - httpAPICall: a shim which performs a HTTP(S) API call to Janus.

Two implementations are provided for dependencies:
  - Janus.useOldDependencies() basically restores the previous behaviour by wrapping jQuery
  - Janus.useDefaultDependencies() provides a more minimalist alternative for modern browsers (it does not require jQuery).

Dependencies may be passed in by providing an optional additional argument to `Janus.init()` or including an optional `dependencies` property with the options object.

The code change is mostly straightforward except for httpAPICall which requires some further explanation:

== httpAPICall ==

This function accepts two parameters:

 - url: the HTTP URL to hit
 - options: a key/value property list containing:
    * verb: the HTTP method/verb/type of request, e.g. 'GET' or 'POST'
    * withCredentials: boolean, whether or not to include credentials/cookies etc. with the outgoing request
    * body: optional, object to be serialised to JSON and include as request body
    * success: a callback to invoke on successful completion of the request, taking a deserialised (parsed) response body (payload) parameter
    * error: a callback to invoke on error, taking a status text + error object/string as parameters
    * timeout: optional, a timeout in msec to impose on the outgoing request

The function must disable caching of the request, deal with content-type (assume application/json), and must take care to properly serialise/deserialising request/response body payloads as needed.


cmacq2:
This PR implements the outlined approach in #879

lminiero:
Thanks for what, even from a quick glance, looks like a huge effort! Since I'll be slow at reviewing this, please advertise this on the group as well, so that people can play with this (especially those interested in interacting with other frameworks) and we get more feedback on what might break in heterogeneous scenarios (in particular on HTTP calls).

A few notes:

1. I guess this makes `janus.nojquery.js` useless, but from the diff it looks like it's an empty file now, rather than being removed? Maybe just a display issue in gihtub.
2. I understand wanting to conditionally get rid of jQuery, but why involve the WebRTC adapter as well? I can't think of any sane person, wanting to remain sane, NOT using the adapter for their WebRTC application. Unless there are other approaches I'm not aware of that people sometimes prefer?
3. Stuff like `Janus.dependencies.webRTCAdapter.whatever` is way too verbose. Can't we get rid of at least the `dependencies.` part altogether, and have the dependency initialization simply define the methods we'll end up using? e.g., `Janus.httpCall`, rather than `Janus.dependencies.httpCall`, etc.
4. This breaks demos using `adapter` (e.g., `adapter.browserDetails.browser === "firefox"`) in case they're using something that is not the default adapter (which again does not make much sense to me).

Thanks!

cmacq2:
> Since I'll be slow at reviewing this, please advertise this on the group as well,

I'm not in "the group"...

Taking the other points one by one:

 1. I deleted the file, so this is probably Github.
 2. This is not to get rid of WebRTC adapter, but rather to let people pass it in explicitly if they want to. 
 3. Will do.
 4. I don't understand? If people do want the defaults they can either opt to explicitly pass in the value (i.e. manually import adapter themselves and pass it on to `Janus`) or they can simply do nothing and the code should fall back to trying the global `adapter` object instead... If people don't want the defaults, well then they presumably know what they are doing and are on their own.

Note that it is useful to let people pass in the `adapter` object explicitly, even if it is just boilerplate to pass in the default choice anyway. The *option* to do so provides for a much simplified way to integrate `janus.js` with build tooling like Webpack or Rollup in a project. Webpack/rollup and other tooling assumes modular code by default, and implicit assumptions about what global variables are available break this, and vice versa: Webpack/Rollup may generate code in which the adapter is *not* available as a global object. Now, each tool gives you the ability to work around it, but if you can simply pass in the dependencies for Janus explicitly that is at least half the job done right there because now you don't have to worry about those globals any more.

cmacq2:
Item 3 is now done: the deps are now bound to the `Janus` object instead.

lminiero:
Thanks for the clarifications! By group I meant the [Janus Google group](https://groups.google.com/forum/#!forum/meetecho-janus): most people only linger there, and don't follow GitHub issues and PRs, so I think it would be useful to make them aware of this effort. I'll write a post tomorrow to draw attention to your changes.

Thanks!

lminiero:
Just wrote a post on the Google group here: https://groups.google.com/forum/#!topic/meetecho-janus/sauFXAj3olk
Hopefully this will encourage more people to play with this and provide feedback.

html/janus.js
```diff
@@ -44,10 +44,109 @@ Janus.isExtensionEnabled = function() {
...
+
+				fetching = p.race([fetching, timeout]);
+			}
+			return fetching;
+		}
+	}
+};
+
+Janus.useOldDependencies = function (jQueryObj, adapterObj) {
+	var jq = jQueryObj || $;
```
thehunmonkgroup: I think
```javascript
var jq = jQueryObj || jQuery;
```
would be less brittle here? In the off case that someone was using other libs that expose $.

html/janus.js
```diff
@@ -44,10 +44,109 @@ Janus.isExtensionEnabled = function() {
...
+			}));
+		},
+	};
+};
+
 Janus.noop = function() {};
 
 // Initialization
-Janus.init = function(options) {
+Janus.init = function(options, deps) {
```
thehunmonkgroup: Since you're checking for ```options.dependencies```, what's the purpose of providing the extra arg?

html/janus.js
```diff
@@ -100,6 +199,13 @@ Janus.init = function(options) {
 			}
 		}
 		Janus.log("Initializing library");
+
+		var usedDependencies = deps || options.dependencies || Janus.useDefaultDependencies();
+		Janus.isArray = usedDependencies.isArray;
+		Janus.webRTCAdapter = usedDependencies.webRTCAdapter;
+		Janus.httpAPICall = usedDependencies.httpAPICall;
+		Janus.checkJanusExtendsion = usedDependencies.checkJanusExtendsion;
```
thehunmonkgroup: It seems a bit weird to hang these off the Janus object. I know the ```dependencies``` key was removed by request, just wondering if there's an even cleaner way to do this?

html/janus.js
cmacq2: I think that should be safe, will do.

html/janus.js
cmacq2: Either one is sufficient by itself, I think this is a remnant from various attempts I've played with while putting the PR together.

html/janus.js
cmacq2: Well, you could have people pass the deps to both `Janus.isExtensionEnabled` and `new Janus()`. In that case the function call argument (dependencies) could simply be captured & used, but it would break backwards compatibility of `Janus.isExtensionEnabled()`. So I don't think that's such a good idea.

html/janus.js
thehunmonkgroup: Then I'd recommend ripping out the extra arg, and leaving the ```dependencies``` key in ```options```

cmacq2:
>  it doesn't seem to address the issue of the global Janus object itself, or am I missing something?

That is correct. The only way to address that would be to wrap Janus in a true module as outlined [in this comment on the UMD PR](https://github.com/meetecho/janus-gateway/pull/792#issuecomment-326818438)

html/janus.js
```diff
@@ -37,13 +37,112 @@ Janus.isExtensionEnabled = function() {
 			// Older versions of Chrome don't support this extension-based approach, so lie
 			return true;
 		}
-		return ($('#janus-extension-installed').length > 0);
+		return Janus.checkJanusExtendsion();
```
lminiero: Typo? (s/Extendsion/Extension). There seems to be the same typo in all the code.

cmacq2:
Thanks for spotting that! Fixed the typo in `checkJanusExtendsion` -> `checkJanusExtension`.

cmacq2:
@lminiero @thehunmonkgroup any further thoughts on this?

thehunmonkgroup:
None from me, what I offered already is what I got.

lminiero:
I still didn't have much time to test this myself, sorry... I plan to do this in the next few days. I also bumped the post on the Google group, as nobody provided feedback there either.

doonfrs:
Hi;
I found the pull useful for two things:
1- Get rid of jquery as it does not seems so loved in SPAs nowadays.
2 - Making webRTCAdapter object configurable.

First change is the major, but I have question here, if we want to get rid from jquery why don't we use janus.noquery.js which already stable and doing its job ?
Because the propose would be maintaining one version, and minimizing dependencies, and janus.js is not heavily using jquery only some functions already covered in janus.nojquery.

I use janus.nojquery and it is stable.
What is the reason for switching between jquery or not as it is for internal usage only.

Thank you!

cmacq2:
> why don't we use janus.noquery.js which already stable and doing its job ?

As per the title of the PR: the purpose is to **decouple** Janus from its dependencies. Being able to get rid of JQuery that easily is merely a beneficial side effect of this. With this PR people can more easily integrate Janus with $framework or $library of their choice, because they get to inject the dependencies. Things this PR makes possible and even easy, which previously were either extremely hard or impossible:

 1. You can now simply use `janus.js` and intercept the API calls and redirect them into your own application logic. Just make sure to dispatch the passed callback at the appropriate moments and you are set. This lets you rewrite Janus API calls into e.g. Redux actions, or convert Janus API responses into payload of other Redux actions. The point isn't Redux here, the point is the flexibility this provides to integrate `janus.js` as-is within a project with more sophisticated needs than the Janus demo pages.
 2. Through your control over the actual API call itself, you can transparently augment and intercept HTTP headers, because you are in control over the final request and response formats. Injecting "Authorization" or working with CSRF tokens is now a lot simpler. Since it is in fact, recommended that people layer their own authentication on top of the Janus Gateway itself for anything beyond simple shared passwords/secrets this is actually a pretty big deal.
 3. It becomes a lot easier to integrate with build tooling which expects a proper module. This PR doesn't resolve the issue entirely, but certainly paves the way -- it is a lot less work to set this up now.

And of course you can do all of the above combined. Whereas the previous "no JQuery" version was merely a way to not require JQuery, it did not make it any easier to do any of the other things. Harder, in fact -- at least JQuery offers effective support for intercepting outgoing `$.ajax()` calls (but note that its support for intercepting responses may not work for item 2 because of the order in which callbacks/events are dispatched).

Finally and perhaps just as importantly:  the number of versions of `janus.js` is now reduced to one: `janus.js`. That's less code to maintain.

cmacq2:
Rebased to latest master.

@lminiero have you been able to test this some more yet?

lminiero:
@cmacq2 not yet but I plan to do it by today, as I'll be unavailable for the rest of a week due to an event abroad. I'll make some tests ASAP and come back to you!

One doubt I had thinking about it in the last few days that I can anticipate was on WebSockets, though: do you think we should "externalise" that dependency as well? Right now we use the basic WebSockets code all browsers support, but especially in the eventuality that janus.js will be integrated in other contexts, there may be WebSocket helpers there as well, or even different ways of doing WebSockets in general (see nodejs WebSockets vs. browser WebSockets). Does this make sense?

lminiero:
Just made a couple of quick tests and while the default dependencies seemed to be working nicely most of the times, I had some issues with fetch apparently.

While playing with the echotest, at a certain point I got a `<< network error >>: request timeout` on the long poll, which was clearly a false positive considering I was testing on localhost and Janus was still alive. All long polls were returned within 30s, so initially I thought there could be a strict 30000 somewhere in the code that could cause this, but the default timeout seems to be 60000 as before, so not sure what the cause was. Maybe a timeout related to a previous long poll is not cleared when you get the answer, and so it kicks in anyway later on?

Apart from this more serious issue, there were a couple of things I noticed and I thought I'd mention. One is that in some tests, when I suddenly shutting Janus while in a demo, I got several bootbox alerts one over the other, instead of just one; not sure if it's related to the above (e.g., more than one callback lingering in the background), but the behaviour should be to retry a few times, which you already do, and only notify the application when we finally give up. Besides, if I try to connect to a Janus that isn't there, I get an ugly `<< network error >>: TypeError: Failed to fetch`, instead of the more verbose `Lost connection to the gateway (is it down?)` we returned so far; not exactly sure we can consider this an issue, as your return value can give more context to what went wrong, but at any rate I'd try to return both here.

Finally, and this is more meant for when we'll merge this, can you update the text in `mainpage.dox` to document the new way one can initialize the Janus object? The general JavaScript section starts [here](https://github.com/meetecho/janus-gateway/blob/master/mainpage.dox#L102) and, as you can see, we have a [note](https://github.com/meetecho/janus-gateway/blob/master/mainpage.dox#L124) that would become outdated after we merge this. At this [line here](https://github.com/meetecho/janus-gateway/blob/master/mainpage.dox#L159) is where we introduce the Janus object creation, which is only a few lines, so aligning it to your changes would allow us to automatically update the documentation we have online as well: https://janus.conf.meetecho.com/docs/JS

Unfortunately I won't have any time to look into further updates before next Monday, but I'll ask one of my colleagues, @atoppi, to track this and provide feedback in the meanwhile.

Thanks!

cmacq2:
> One doubt I had thinking about it in the last few days that I can anticipate was on WebSockets, though: do you think we should "externalise" that dependency as well?

Ideally, yes. That would be just forwarding `new WebSocket()` to the dependencies, right?

cmacq2:
> While playing with the echotest, at a certain point I got a << network error >>: request timeout on the long poll, which was clearly a false positive considering I was testing on localhost and Janus was still alive. All long polls were returned within 30s, so initially I thought there could be a strict 30000 somewhere in the code that could cause this, but the default timeout seems to be 60000 as before, so not sure what the cause was. Maybe a timeout related to a previous long poll is not cleared when you get the answer, and so it kicks in anyway later on?

Rather more embarrassingly, my usage of `Promise.race()` did simply not work correctly. I've now fixed this.

cmacq2:
> Besides, if I try to connect to a Janus that isn't there, I get an ugly << network error >>: TypeError: Failed to fetch, instead of the more verbose Lost connection to the gateway (is it down?) we returned so far; not exactly sure we can consider this an issue, as your return value can give more context to what went wrong, but at any rate I'd try to return both here.

I think we should be careful here. From a code hygiene perspective, the fact that the demos simply use raw error objects and dump these in UI elements does not pass the smell test. ;)

More precisely, as I see it the choice is the following:

 * either `janus.js` should be intelligent enough to map errors to pretty user-facing error messages with the downside that calling code is forced to be dumbed down considerably (no descriptive error objects are possible anymore, after all) so things like failover to a backup Janus server becomes *very* hard to do...
* or we can enrich `janus.js` error reporting mechanism with more precise error objects, at the cost of having to update demo pages to do something sensible with that information.

Either option amounts to more work; and I feel it is somewhat more suitable in the scope of a separate (follow-up) PR.

cmacq2:
> Finally, and this is more meant for when we'll merge this, can you update the text in mainpage.dox to document the new way one can initialize the Janus object?

Sure, please remind me when this gets to merge-ready and trigger fingers start getting itchy. Just in case we decide to refactor things or externalise more deps like the `Websocket` constructor. ;)

cmacq2:
@lminiero please retest with latest version of the branch. I've cleaned up the fetch() based API call implementation somewhat.

lminiero:
Thanks! I'll test as soon as I'm back of the office (abroad for an event with limited connectivity).

lminiero:
Back at the office, will try to make some tests as soon as I have some free time from all the catching up.

lminiero:
As a quick note before I do some new tests, do you think it might make sense to pass an object, instead of individual parameters, to the "usedependencies" methods? This might make it more extensible in case in the future we end up needing more stuff to make configurable.

cmacq2:
Updated PR with proposed API change, and rebased on top of latest master.

lminiero:
Seems to be working nicely from a few quick tests, thanks! I'll ask my colleagues to test this as well, but I don't anticipate any issue that could prevent us from merging this very soon. Before that, could you make those few changes to the `mainpage.dox` file too, though? I only expect a few lines to need the change, specifically those related to the Janus object init process, and the fact there's no `no.jquery` file anymore.

Thanks again for contributing this!

cmacq2:
Added initial draft of documentation updates. Please review, I've been typing too much to be any good at proof reading it right now ;)

atoppi:
Looks good to me.

mainpage.dox
```diff
@@ -165,14 +173,16 @@
...
+ * - \c callback: a user provided function that is invoked when the initialization is complete.
+ * - \c dependencies: a user provided implementation of Janus library dependencies.
  *
  * Here's an example:
  *
  *
  \verbatim
 Janus.init({
    debug: true,
+   dependencies: Janus.useDefaultDependencies() // or: Janus.useOldDependencies() to get the behaviour of previous Janus versions
```
lminiero: Comma missing after `Janus.useDefaultDependencies()`

mainpage.dox
```diff
@@ -121,17 +121,25 @@
...
+ * depend on certain global variables. Two implementations are included in \c janus.js itself:
+ *
+ *  1. \c Janus.useDefaultDependencies which relies on native browser APIs,
+ *     which in turn require somewhat more modern browsers
+ *  2. \c Janus.useOldDependencies which uses jQuery (http://jquery.com/) instead,
+ *     and should provide equivalent behaviour to previous versions of \c janus.js
+ *
+ * By default \c Janus.useDefaultDependencies will be used, but you can override this
+ * when initialising the Janus library and pass a custom dependencies object instead.
+ * For details, refer to \ref js-dependencies the documentation on working with custom
```
lminiero: When you use `\ref anchor`, Doxygen inserts the section title inline. This means that, since in this case the title is "Working with custom janus.js dependencies", it's better to use something like:

    For details, refer to the documentation on \ref js-dependencies about...

as your current text at the moment is rendered like this when compiled:

    For details, refer to Working with custom janus.js dependencies the documentation

mainpage.dox
```diff
@@ -165,14 +173,16 @@
  *   - \c true or \c "all": all debuggers enabled (Janus.trace, Janus.debug, Janus.log, Janus.warn, Janus.error)
  *   - array (e.g., <code>["trace", "warn"]</code>): only enable selected debuggers (allowed tokens: trace, debug, log, warn, error)
  *   - \c false: disable all debuggers
- * - \c callback: a user provided function that is invoked when the initialization is complete. 
+ * - \c callback: a user provided function that is invoked when the initialization is complete.
```
lminiero: Just a nit: since callback is not the last item anymore, please use `;` instead of the dot.

mainpage.dox
```diff
@@ -574,6 +584,123 @@ janus.attach(
  *
  */
 
+/*!\page js-dependencies Working with custom \c janus.js dependencies
```
lminiero: Can you remove the `\c` from the section title? Just a nit as it simply renders poorly when compiled by Doxygen.

mainpage.dox
```diff
@@ -574,6 +584,123 @@ janus.attach(
...
+ * -# \c newWebSocket: a function which given WebSockets server and protocol arguments
+ * should return a new WebSocket (or something that acts like it)
+ * -# \c webRTCAdapter: an \c adapter object such as provided by the
+ * <a href="https://github.com/webrtc/adapter">webrtc-adapter</a> library
+ * -# \c isArray: a function which tests if a given argument is a JavaScript array
+ * -# \c checkJanusExtension: a function which tests if the Janus Screensharing extension
+ * for Chrome is installed/available
+ * -# \c httpAPICall: a function which given an url and options argument performs an
+ * HTTP API request to the Janus Gateway. This function is not as straightforward to implement,
+ * see \ref js-http-apicall the documentation on \c httpAPICall below for details.
```
lminiero: See above on the usage of \ref, I guess here you can replace the existing line:

    see \ref js-http-apicall the documentation on \c httpAPICall below...

with this:

    see the documentation on \ref js-http-apicall below

as the `js-http-apicall` already has an `httpAPICall` label which would be inserted inline.

mainpage.dox
```diff
@@ -574,6 +584,123 @@ janus.attach(
...
+ *
+ * Dependening on your needs you may not need to provide all these dependencies, e.g.
```
lminiero: s/Dependening/Depending

mainpage.dox
```diff
@@ -574,6 +584,123 @@ janus.attach(
...
+ *
+ * Two implementations of such a dependencies object are provided by \c janus.js:
+ *
+ *  1. \c Janus.useDefaultDependencies
+ *  2. \c Janus.useOldDependencies
+ *
+ * In turn, each of these implementations accept their dependencies as arguments or fallback on
+ * certain global variables. Below follows an overview:
+ *
+ * \section js-default-deps \c Janus.useDefaultDependencies
```
lminiero: Please remove the \c from here too.

mainpage.dox
```diff
@@ -574,6 +584,123 @@ janus.attach(
...
+ * The \c Janus.useOldDependencies method relies on:
+ *
+ * -# \c jQuery: the JQuery library (http://jquery.com/)
+ * -# \c WebSocket: support for the \c WebSocket API (https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API)
+ * -# \c adapter: \c adapter object from the <a href="https://github.com/webrtc/adapter">webrtc-adapter</a> library
+ *
+ * This function provides a simple upgrade path for existing applications which are heavily
+ * tied to jQuery (especially since previous versions of \c janus.js depended on it).
+ *
+ * \section js-http-apicall \c httpAPICall
```
lminiero: Please remove the \c from here too.

mainpage.dox
```diff
@@ -574,6 +584,123 @@ janus.attach(
...
+
+const setupDeps = () => Janus.useDefaultDependencies({
+	adapter,
+	// other dependencies elided
+});
+
+export const initialiseJanusLibrary = () => Janus.init({dependencies: setupDeps()});
+\endverbatim
+ *
+ * \section js-old-deps \c Janus.useOldDependencies
```
lminiero: Please remove the \c from here too.

cmacq2:
Updated to:

 1. Rebased the branch and made it two separate commits: the code change and the documentation update
 2. Took the advantage to rewrite the commit messages to better reflect the current state
 3. Fixed my usage of \ref throughout, various typos and incorporated your suggestions.
 4. Updated with a couple more references (e.g. "Redux" and "CSRF" now link to pages which explain them, just in case).

lminiero:
Thanks for fixing so quickly, fnally merging then!

lminiero:
Updated docs online too:

https://janus.conf.meetecho.com/docs/JS
https://janus.conf.meetecho.com/docs/js-dependencies

PR #993, opened 2017-09-05
Added text2pcap support for handles traffic

lminiero:
This PR adds a new feature that might be very useful when debugging RTP/RTCP traffic via tools like Wireshark. Specifically, it allows you, as an administrator (via Admin API) to turn on the dumping of all incoming and outgoing unencrypted RTP/RTCP packets for a handle to a text file, in [text2pcap](https://www.wireshark.org/docs/man-pages/text2pcap.html) format. This file can then be fed to the `text2pcap` tool Wireshark provides to convert it to a `.pcap` or `.pcapng` file, so that you can analyse the traffic via Wireshark or similar tools. This is a feature Firefox recently [added as well](https://blog.mozilla.org/webrtc/debugging-encrypted-rtp-is-more-fun-than-it-used-to-be/) to facilitate debugging of WebRTC traffic, bypassing the SRTP encryption.

Starting a dump via Admin API for a specific handle (IDs in the URL in the following example, they can be set in the request as well) should be simple enough, e.g.:

```
curl -X POST -H "Content-Type: application/json" -d '{"janus": "start_text2pcap", "folder": "/tmp", "filename": "my-test2pcap-dump.txt", "transaction": "123", "admin_secret": "janusoverlord"}' http://localhost:7088/admin/8412133783240844/2377476017639045
```

saves to `/tmp/my-test2pcap-dump.txt`. Stopping is even easier:

```
curl -X POST -H "Content-Type: application/json" -d '{"janus": "stop_text2pcap", "transaction": "123", "admin_secret": "janusoverlord"}' http://localhost:7088/admin/8412133783240844/2377476017639045
```

To convert the `/tmp/my-test2pcap-dump.txt` text file to a format Wireshark can handle, as anticipated, you can use the well known `text2pcap` tool:

```
text2pcap -D -n -l 1 -i 17 -u 1000,2000 -t '%H:%M:%S.' /tmp/my-test2pcap-dump.txt /tmp/my-test2pcap-dump.pcapng
```

Just open that file, decode as RTP, and RTP/RTCP packets will magically appear for your debugging pleasure.

That said, while I considered simply extending the existing `janus_recorder` mechanism to also allow for `text2pcap` output, I preferred using a separate piece of code (and helpers) for that. The reason is simple: I wanted to stress out that, although you can use it for that, as explained [here](https://webrtchacks.com/video_replay/), we don't consider this a way to record conversations and streams, but a tool for debugging. As such, while plugins can indeed make use of the feature if they want (e.g., the SIP plugin could use it to dump the SIP calls, or the Streaming plugin to dump RTSP traffic), this is currently only baked in the core, and specifically hooked to the Admin API so that Janus administrators can debug the handles the Janus instance is processing, if they want.

As usual, feedback welcome! Since this is an optional feature that shouldn't impact existing feature, I plan to merge soon.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,7 +29,7 @@ bin_PROGRAMS = janus
 
 headerdir = $(includedir)/janus
 header_HEADERS = apierror.h config.h log.h debug.h mutex.h record.h \
-	rtcp.h rtp.h sdp-utils.h ip-utils.h utils.h
+	rtcp.h rtp.h sdp-utils.h ip-utils.h utils.h text2pcap.h
 
 pluginsheaderdir = $(includedir)/janus/plugins
 pluginsheader_HEADERS = plugins/plugin.h
@@ -130,6 +130,8 @@ janus_SOURCES = \
 	utils.h \
 	version.c \
 	version.h \
+	text2pcap.c \
+	text2pcap.h \
 	plugins/plugin.c \
 	plugins/plugin.h \
 	transports/transport.h \
--- a/ice.c
+++ b/ice.c
@@ -1948,6 +1948,7 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 		JANUS_LOG(LOG_ERR, "No handle for stream %d??\n", stream_id);
 		return;
 	}
+	janus_session *session = (janus_session *)handle->session;
 	if(!component->dtls) {	/* Still waiting for the DTLS stack */
 		JANUS_LOG(LOG_WARN, "[%"SCNu64"] Still waiting for the DTLS stack for component %d in stream %d...\n", handle->handle_id, component_id, stream_id);
 		return;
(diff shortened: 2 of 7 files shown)


fippo:
typically you want not just rtp/rtcp but also ice + dtls. And an option to truncate the rtp data because full packets as text2pcap hexdump is a bit... verbose :-)

trankkilou:
So you are converting binary to text ?

lminiero:
> typically you want not just rtp/rtcp but also ice + dtls.

DTLS should be easy enough to add, but ICE won't be possible unfortunately, as libnice takes care of that transparently and doesn't expose that info to the application. But DTLS+ICE should be trivial to capture externally, with a simple tcpdump on the server: I guess the main advantage of this feature here is that it gives you access to the unencrypted RTP/RTCP traffic, something that with a plain tcpdump you wouldn't be able to do.

> And an option to truncate the rtp data because full packets as text2pcap hexdump is a bit... verbose :-)

Good point, thanks, I'll add that!

lminiero:
> So you are converting and storing binary data into text ? is it very efficient ?

I don't know about efficiency, but it's simple and effective. This isn't something I'd see as being kept on for all your streams all the time anyway, but only something you turn on for selected connections on demand whenever you have to debug something specific.

lminiero:
I don't see any objection, so merging.

Issue #913, opened 2017-06-01
Segmentation fault when wss and admin wss are enabled


tgabi333:
When i try to connect to swss it hangs up, janus process has quit with segmentation fault.

Config:
- only wss and admin swss are enabled

Actions:
- Try to connect to swss (not admin)

After admin_wss is turned off, it works.
Using libwebsockets 2.2.1

Logs: there is no relevant log event with 
Stack trace: https://pastebin.com/aHMfj8mY

lminiero:
Looks like a libwebsockets issue. Definitely not happening to me, but I'm on an older version. You should try to compile a debug version of the library, because the stacktrace is useless as it is.

tgabi333:
Tested libwebsockets: 1.5, 1.6, 1.7.9

sdrsdr:
same here libwebsockets-2.1.1

sdrsdr:
backtrace:
```
Thread 29 "sws thread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffc63c6700 (LWP 30973)]
0x00007fffc6bdc6c2 in lws_ssl_server_name_cb (ssl=0x7fffc0000ab0, ad=0x7fffc63c5810, arg=0x0) at /var/tmp/portage/net-libs/libwebsockets-2.1.1/work/libwebsockets-2.1.1/lib/ssl-server.c:171
171     /var/tmp/portage/net-libs/libwebsockets-2.1.1/work/libwebsockets-2.1.1/lib/ssl-server.c: No such file or directory.
(gdb) backtrace 
#0  0x00007fffc6bdc6c2 in lws_ssl_server_name_cb (ssl=0x7fffc0000ab0, ad=0x7fffc63c5810, arg=0x0) at /var/tmp/portage/net-libs/libwebsockets-2.1.1/work/libwebsockets-2.1.1/lib/ssl-server.c:171
#1  0x00007ffff6e48a10 in ssl_parse_clienthello_tlsext () from /usr/lib64/libssl.so.1.0.0
#2  0x00007ffff6e2e9d7 in ssl3_get_client_hello () from /usr/lib64/libssl.so.1.0.0
#3  0x00007ffff6e33377 in ssl3_accept () from /usr/lib64/libssl.so.1.0.0
#4  0x00007ffff6e42e1f in ssl23_accept () from /usr/lib64/libssl.so.1.0.0
[...]

PR #996, opened 2017-09-07
Fix for the SSL+WebSockets crashes (see #913)

lminiero:
See issue #913 for a motivation behind this fix. As suggested by @lws-team, this patch refactors the WebSockets plugin to use a single context and multiple vhosts, rather than multiple contexts as we did before. Since apparently vhosts are a more or less recent addition, this breaks what we called the "old API" compatibility (which IIRC was 1.5, so not really relevant anymore). Not sure how recent vhosts are, though, e.g., if we have to mandate a minimum version of libwebsockets in configure.ac now.

I tested this briefly but not much, so feedback welcome! Especially to make sure the WebSockets support still works fine in all your scenarios.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -353,39 +353,24 @@ AC_SUBST([MHD_CFLAGS])
 AC_SUBST([MHD_LIBS])
 AM_CONDITIONAL([ENABLE_REST], [test "x$enable_rest" = "xyes"])
 
-AM_CONDITIONAL([ENABLE_WEBSOCKETS_NEWAPI], false)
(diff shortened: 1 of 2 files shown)


lminiero:
We now check for `lws_create_vhost` in `configure.ac`, so no need to worry about libwebsockets versions. This mandates usage of libwebsockets >= 2.0 in the plugin, but I don't think this will be a problem for most users.

lminiero:
No objection received, and got confirmation from one of the affected users that it works, so merging.

PR #999, opened 2017-09-08
Fix NACK for bundle stream for listener

pitkonst:


lminiero:
Not sure about this fix, as it only moves a `janus_sdp_process` error check. If the cause was `janus_sdp_process` returning an error, how relevant is it for NACKs?

pitkonst:
Thanks for reply. I'll try to explain issue as I see it. In case of listener NACK processing is skipped due to check at `ice.c` (line 2118). It appears that `component->do_audio_nacks` and `component->do_video_nacks` are always false. SDP is ok and allows NACKs. `component->do_*_nacks` are set inside `janus_sdp_process`. In case of listener those value set correctly, but for a wrong stream. At `sdp.c` (line 137) `JANUS_ICE_HANDLE_WEBRTC_BUNDLE` flag is false, while it should be true due to browses SDP and `component->do_*_nacks` goes to video stream, instead of bundled audio stream.
It appears that root of a problem is in incorrect `JANUS_ICE_HANDLE_WEBRTC_BUNDLE` flag value inside `janus_sdp_process`, but it happens only when SDP answer comes from browser.
In this case the flag set at `janus.c` (line 1110) which goes right after `janus_sdp_process`. That is why moving block with flags set before call of `janus_sdp_process` solves the issue, at least in my case.

Sorry for verbose explonation. It could be better solution.

lminiero:
Thanks for your clarification: it makes sense, considering the NACKs activity depending on the negotiation in the SDP is a recent addition I made before the summer break, and so it may be buggy. I'll give a look and evaluate your patch.

lminiero:
Apologies for this very late reply, I've been awfully busy and abroad all week... I plan to make a couple of tests later today. I also advertised your PR in a [post on our group](https://groups.google.com/forum/#!topic/meetecho-janus/VP_8qcx0mm8) as they were experiencing issues that may have been related to the issue you're trying to fix, so their feedback will help.

As a side note, could you sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this as soon as we've evaluated it? Thanks!

trankkilou:
Hi,
Works for listeners as expected. But need the same for Publisher.
nice catch anyway 
br,
J-

pitkonst:
I've experienced same problems with Publisher, but do not dig inside yet. Just made some dirty overrides, to make NACK work again for Publisher. As for now it seems that Publisher problem involves simulcast-related commits.

pitkonst:
Done with CLA.

lminiero:
> As for now it seems that Publisher problem involves simulcast-related commits.

That's a good hint on what to look into next, thanks!

lminiero:
I think there's an easier way to fix the issue this PR addresses. Basically, we have some code to "fix" the streams when we detect bundle and/or rtcp-mux: https://github.com/meetecho/janus-gateway/blob/master/janus.c#L1127
This code simply "imports" some properties from the streams we need to remove (e.g., video and data) in the remaining stream (e.g., audio). We already have some imports there, but the NACK stuff is missing, so it should be enough to simply add the import there, rather than moving those pieces of code, which might cause some problems (I haven't delved in the implications yet).

I'll make some experiments on this, and then check what makes it problematic for publishers too.

lminiero:
Implemented the simpler change I mentioned here: https://github.com/meetecho/janus-gateway/commit/7708336de113d66927baaad71bb6e7451257af57
@pitkonst @trankkilou can you let me know if that solves the issue for you?
With respect to publishers, I verified we're currently ignoring lost packets on any simulcast layer that is not the base one: if you're not using simucast layer, that's not important at all (there's only one layer, and we handle nacks for those) and it should work as expected; if you are, it just means that we only send NACKs for packets lost on the base layer, and not on higher layers. Not sure if this is what you were experiencing.

trankkilou:
Unfortunately it doesn’t.

I restarted from a fresh install from today. (So my patch are gone)

I check the patch you made is well apply.

Nack are totally absent for listeners as for publisher at browser level.

In janus log, needs to up log level to « 6 » to notice the core can see NACKs but do nothing, so it is not a surprise I have nothing at browser level.

Regarding admin API: Nack audio are set to « false » and back video are set to « true »
Nacks counter increased only for listener.

So to sum : Nacks are seen by the core (for listener only) but not honored , slow link still missing for listener as publisher

J-
> Implemented the simpler change I mentioned here: 7708336 <https://github.com/meetecho/janus-gateway/commit/7708336de113d66927baaad71bb6e7451257af57>
> @pitkonst <https://github.com/pitkonst> @trankkilou <https://github.com/trankkilou> can you let me know if that solves the issue for you?
> With respect to publishers, I verified we're currently ignoring lost packets on any simulcast layer that is not the base one: if you're not using simucast layer, that's not important at all (there's only one layer, and we handle nacks for those) and it should work as expected; if you are, it just means that we only send NACKs for packets lost on the base layer, and not on higher layers. Not sure if this is what you were experiencing.
> 
> —

lminiero:
> Nacks are seen by the core (for listener only) but not honored

I think this is normal: there's packet loss on the publisher, so Janus doesn't have the packets the viewer is asking for (the packets we never got from the publisher), and cannot retransmit them.

Still not sure why NACKs are not being sent to the publisher when we fail to receive packets from him. Do you see lost packets in the Admin API?

trankkilou:
By "packets lost » you mean something available in the admin page (didn’t see nothing about it)  or a specific call I have to do to the admin API to get them ?

lminiero:
Admin API, just open it and you'll find it.

trankkilou:
Sorry I can’t see anything relative to packets loss or packets drop 
Here: https://janus.conf.meetecho.com/docs/admin <https://janus.conf.meetecho.com/docs/admin> or here:  http://www.meetecho.com/blog/understanding-the-janus-admin-api/

lminiero:
I obviously meant in the Admin API info you get on YOUR handles... the documentation is not up to date with all the stuff we have.

trankkilou:
Ok I think I got what you mean and I had to write down a quick script

The answer from admin api is here: https://pastebin.com/CbGqnKZ4 <https://pastebin.com/CbGqnKZ4>

The only thing relative I see is RTCP stats and video stats

"rtcp_stats": {
               "audio": {
                  "base": 48000,
                  "lsr": 88282189,
                  "lost": 2609,
                  "lost-by-remote": 0,
                  "jitter-local": 2,
                  "jitter-remote": 0
               },
               "video": {
                  "base": 90000,
                  "lsr": 88443484,
                  "lost": 0,
                  "lost-by-remote": 0,
                  "jitter-local": 0,
                  "jitter-remote": 0

trankkilou:
Side notes :

You can see only audio stats of tcp stats are populated even I simulated 5% of packets loss (mean audio and video)

If I comment out the right lines in ice.c (lines started @2126) , Nacks counter comes back but only for audio

So I think there is 2 issues here

J-

lminiero:
Are you using simulcasting or not?

trankkilou:
No you should have notice the codec: h264 -> impossible to have simulcasting

lminiero:
Have you tried VP8? Just trying to figure out if browsers handle NACKs differently for different codecs (although I guess it worked for you before for the same codec).

trankkilou:
No pb I can try vp8

I don’t know if the issue I raise where is from the beginning or not, never pay attention to such detail

Before I was just looking for the slow link event…

Will update

J-

trankkilou:
Update:

Same  in VP8

pitkonst:
Works for me, `slow_link` are in place too.

lminiero:
@pitkonst so it works for VideoRoom publishers too? Or more in general, you do see slow_link events associated with both uplink and downlink?

pitkonst:
It's not working for Publisher.

lminiero:
You mentioned that you had done some digging on this, and thought it could be related to simulcast stuff, but I checked and couldn't find anything wrong there. Do you have anything to share on that, that might me help me find out what's wrong? I'm working on putting up a small testbed to simulate losses myself, but it might take a while as I'm busy on other stuff too.

pitkonst:
I'm digging right now. Will update.

pitkonst:
Here is situation trace:
1. `ice.c L1985` `stream->video_ssrc_peer == packet_ssrc` and `video` set to 1.
2. `ice.c L2073` pass packet to plugin.
3.  `janus_videoroom.c L2638` changes `packet_ssrc` to `participant->video_ssrc` which set at `janus_videoroom.c L3265` to random number.
4. `ice.c L2118` always return because `header->ssrc` was set to different value at step 3.

Commenting `ice.c L2118-L2119` makes Publisher NACK work again.

trankkilou:
It exists many tool to simulate packets loss

On linux you can use « tc » (trafic control)

It is a little hard to use...

lminiero:
Yeah I know, but I usually test locally and I cannot simulate loss there. I'm preparing a separate testbed that I can test against.

trankkilou:
And it fixes for publisher also ! For video and audio !

On the publisher handle I have now using admin API


"rtcp_stats": {
               "audio": {
                  "base": 48000,
                  "lsr": 537337695,
                  "lost": 812,
                  "lost-by-remote": 0,
                  "jitter-local": 2,
                  "jitter-remote": 0
               },
               "video": {
                  "base": 90000,
                  "lsr": 537411404,
                  "lost": 714,
                  "lost-by-remote": 0,
                  "jitter-local": 6,
                  "jitter-remote": 0
               }
 In janus log

root@OMS-WEBRTC0:/opt/janus-gateway# tail -f /var/log/janus.log | grep -i nack
[Wed Sep 27 15:06:28 2017] [6959480221782399]  sent NACKs for 16 missing packets
[Wed Sep 27 15:06:33 2017] [6959480221782399]  sent NACKs for 9 missing packets
[Wed Sep 27 15:06:34 2017] [221835560932026]  retransmitted 1 packets due to NACK
[Wed Sep 27 15:06:38 2017] [6959480221782399]  sent NACKs for 14 missing packets
[Wed Sep 27 15:06:39 2017] [221835560932026]  retransmitted 1 packets due to NACK

Only with uplink loss simulated


Regarding browser console, both slow link are back

Conclusion: I think Piter pointed out the right direction

lminiero:
> Commenting ice.c L2118-L2119 makes Publisher NACK work again.

Thanks for spotting the issue! Commenting that line out won't work, though, as it will completely break stats and NACKs when simulcasting is involved. What we need to fix is just the check, and I can work on a patch to do that shortly.

lminiero:
Can you check if replacing `ntohl(header->ssrc)` with `packet_ssrc` in L2118 fixes it for you? Like this:

```
if(video && packet_ssrc != stream->video_ssrc_peer)
```

In fact, at L1984 we save the original value of the header SSRC in that variable, and so this should make the check work as expected.

lminiero:
@pitkonst @trankkilou this commit includes the change I suggested: https://github.com/meetecho/janus-gateway/commit/c4c63c9f37f65e37ccb8fb3d64c97a64b37619f7
Please let me know if that fixes it for you or not.

pitkonst:
It works. Of cause `packet_ssrc` is diclared inside block, so I've moved it from L1984 to L1979.
I think this is it. Thanks, Lorenzo.

lminiero:
Thanks for the feedback! I'll close this PR then, as we addressed the issue using two different commits, but 1000k thanks anyway for helping solve this for good :+1:

lminiero:
Next step will be generating slowlink events from lost packets (RTCP RR) rather than NACKs, as it would be a much more reliable way of figuring out problems. I'll add it to the TODO list.

pitkonst:
Actualy it was my primary target. Main part is working. I'll publish PR as soon, as it be ready.

lminiero:
Oh that would be really cool, looking forward to it, thanks!

trankkilou:
Hi lorenzo

I confirm it works thanks !

I tested also in h264, works also.

I notice the number of NACK in slow link event is much more less than what the core say.

Is it relative to what you said (RTCP RR vs NACK) ?

J-

lminiero:
You can get more NACKs for the same packet that has been lost, which is why it would be better to use RTCP RR to compute the absolute number of packets that have been lost in the last timed window.

trankkilou:
Ok make sens !

trankkilou:
@pitkonst can't wait to test it !

PR #1001, opened 2017-09-11
Combine videos

mvnrsa:
First attempt at a bash script that can combine the MJR files into one playable video.

### DIFF
--- a/postprocessing/janus-combine.sh
+++ b/postprocessing/janus-combine.sh
@@ -70,7 +70,7 @@ do
 done
 
 # Set variables saved to file during loop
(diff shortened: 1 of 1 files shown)


lminiero:
This is an interesting effort, but I see it more as a separate repo devoted to processing recordings, than to the main repo here. I think the basic janus-pp-rec is all we need (I actually thought about moving that to a separate repo as well), everything else is orthogonal to Janus. If you create a new repo for this, please let me know and I'll add it to the resources docs :+1:

mvnrsa:
Sure, no problem.  Will do.

mvnrsa:
https://github.com/mvnrsa/janus-combine-video

PR #1006, opened 2017-09-13
janus_streaming: g_async_queue_pop() will never return NULL

tmatth:
It blocks until the queue has data. The non-blocking version is
g_async_queue_try_pop().


lminiero:
Are you sure? IIRC it can still return NULL if it's interrupted (e.g., by a signal).

tmatth:
Only if you pass it a NULL queue, see:
https://github.com/GNOME/glib/blob/6b948d9613f2dcdec72828f3429ba01403063ff2/glib/gasyncqueue.c#L431
and https://github.com/GNOME/glib/blob/master/glib/gasyncqueue.c#L453

lminiero:
Closing as it only addresses a single plugin and not all: I'll keep note of this for future commits, so that this can be added to all in case. Thx!

PR #1024, opened 2017-10-10
Move SRTP specific definitions from rtp.h to a new header

lminiero:
As the title says, this PR moves all the SRTP related stuff we had in `rtp.h` to a new header, called `rtpsrtp.h`. The implementations are still in `rtp.c`.

The rationale for this comes from some issues we were told about, specifically with respect to plugins written as separate projects. Since SRTP definitions were in rtp.h, building plugins would require a dependendy on libsrtp as well, and one that had to be the same as the one Janus uses, due to the ifdefs we have in place. Since most plugins do not need SRTP at all, it made sense to split the SRTP stuff: the only stock plugin that uses SRTP is the SIP plugin, and we updated it to include the new header.

I took advantage of this change to also add a `typedef` to use `janus_rtp_header` instead of `rtp_header`, which is probably the only structure left that is still unprefixed. `rtp_header` is still there to ensure compatibility for a while longer, but I plan to get rid of it in the future, so if you're writing plugins stick to `janus_rtp_header` instead from now on.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,7 +29,7 @@ bin_PROGRAMS = janus
 
 headerdir = $(includedir)/janus
 header_HEADERS = apierror.h config.h log.h debug.h mutex.h record.h \
-	rtcp.h rtp.h sdp-utils.h ip-utils.h utils.h text2pcap.h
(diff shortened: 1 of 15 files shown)


tmatth:
LGTM

lminiero:
Thx for the feedback, merging then!

PR #1027, opened 2017-10-11
Static RTP forwarder for audiobridge

mikaelnousiainen:
There are some use cases for audiobridge, where the mixed audio needs to be forwarded always to another application, e.g. for local playback at the server or custom post-processing. This change allows adding a single statically configured RTP forwarder in the audiobridge plugin configuration file.

Note: A similar feature could be implemented in the videoroom plugin, but I don't have any use for it.

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -75,7 +75,9 @@ rtp_forward_always_on = true|false, whether silence should be forwarded when the
  * process the mixer audio stream. You can add new RTP forwarders with
  * the \c rtp_forward request; a \c stop_rtp_forward request removes an
  * existing RTP forwarder; \c listforwarders lists all the current RTP
- * forwarders on a specific AudioBridge room instance.
+ * forwarders on a specific AudioBridge room instance. As an alternative,
+ * you can configure a single static RTP forwarder in the plugin
+ * configuration file.
  *
  * \c create can be used to create a new audio room, and has to be
  * formatted as follows:
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -30,6 +30,11 @@ audiolevel_ext = yes|no (whether the ssrc-audio-level RTP extension must be
 	negotiated/used or not for new joins, default=yes)
 record = true|false (whether this room should be recorded, default=false)
 record_file =	/path/to/recording.wav (where to save the recording)
+rtp_forward_host = host address to forward RTP packets of mixed audio to
+rtp_forward_port = port to forward RTP packets of mixed audio to
+rtp_forward_ssrc = SSRC to use to use when streaming (optional: stream_id used if missing)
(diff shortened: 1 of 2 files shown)


lminiero:
Sounds like a useful thing to have, thanks! I don't see the same thing happening for the VideoRoom, though, as there can be many publishers in a room there, and each would need to have its own RTP forwarder sending so some different place, something you cannot formalize in a static way in a configuration file.

The only point I can raise before adding some comments inline is that the configuration doesn't seem to include an ID for the RTP forwarder: this means that there would be no way to programmatically (via API) get rid of a static forwarder, as the ID to remove would be unknown.

That said, could you sign our [CLA](https://cla.conf.meetecho.com/janus/) so that we can safely merge this when the time comes? Thanks again for this contribution!

mikaelnousiainen:
@lminiero Done, an ID can be specified. A random number will be used (just like before) if it's missing. In my use cases I don't ever need to remove the static forwarder, because it needs to be always active, this is good for making configuration complete.

lminiero:
@mikaelnousiainen thanks for the quick changes! I'll review early next week and come back with feedback ASAP. If everything works fine I'll merge right away :+1:

plugins/janus_audiobridge.c
```diff
@@ -875,16 +883,27 @@ static guint32 janus_audiobridge_rtp_forwarder_add_helper(janus_audiobridge_room
...
 	rf->always_on = always_on;
+
 	janus_mutex_lock(&room->rtp_mutex);
-	guint32 stream_id = janus_random_uint32();
-	while(g_hash_table_lookup(room->rtp_forwarders, GUINT_TO_POINTER(stream_id)) != NULL) {
-		stream_id = janus_random_uint32();
+
+	guint32 actual_stream_id;
+	if(stream_id > 0) {
+		actual_stream_id = stream_id;
```
lminiero: Should we actually check if the stream ID exists, and in case fail? As if it does, the code currently overwrites it with a random one, thus replacing whatever the user provided. Anyway, this is currently only allowed for the static RTP forwarder, and not for requests via the API, so we can safely ignore for the moment.

lminiero:
Tested and works fine, merging :+1: 
Thanks again for your contribution!

plugins/janus_audiobridge.c
mikaelnousiainen: It does the check, see the code below. If the ID exists, it will get a random ID. But as you mentioned, configuring a static RTP forwarder is always the first one. Thanks for merging!

lminiero: Yep, that was my point: using a random ID instead of what the user asked for is possibly a broken behaviour, and the request should fail instead. But since we don't currently allow users to specify a stream ID when adding a forwarder via API, it's a non-issue at the moment. I'll make sure this is addressed should it change in the future.

PR #1028, opened 2017-10-11
Make RTCPeerConnection a configurable dependency

sanbornhilland:
This change will be transparent to anyone who does not care
to customize the use of the RTCPeerConnection constructor.
However, it provides extra control to the options and constraints
being passed to the RTCPeerConnection constructor so that you
can do stuff like this:
```javascript
Janus.init({
  debug,
  callback,
  dependencies: Janus.useDefaultDependencies({
    newRTCPeerConnection(config, constraints) {
      // Add additional constraints
      constraints.optional.push({
        googCpuOveruseDetection: false,
      })

      return new RTCPeerConnection(config, constraints)
    }
  }),
})
```

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -48,11 +48,15 @@ Janus.useDefaultDependencies = function (deps) {
 	var f = (deps && deps.fetch) || fetch;
 	var p = (deps && deps.Promise) || Promise;
 	var socketCls = (deps && deps.WebSocket) || WebSocket;
+	var newRTCPeerConnection = (deps && deps.newRTCPeerConnection) ||
+		function(pc_config, pc_constraints) { return new RTCPeerConnection(pc_config, pc_constraints)}
+
 	return {
 		newWebSocket: function(server, proto) { return new socketCls(server, proto); },
 		isArray: function(arr) { return Array.isArray(arr); },
 		checkJanusExtension: function() { return document.querySelector('#janus-extension-installed') !== null; },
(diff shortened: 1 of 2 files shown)


lminiero:
Mh, if I understand the purpose of the PR, this isn't something I'd add to the init, but as an additional callback to the createOffer/createAnswer janus.js methods instead. This would allow you to provide different constraints for different PeerConnections, rather than sticking to the same for all PCs created in a session.

lminiero:
Maybe the additional arguments may simply be the list of constraints one wants to add, and to either merge with the ones we provide or replace them entirely, rather than a PC constructor overload. What do you think?

lminiero:
@sanbornhilland I ended up doing it in https://github.com/meetecho/janus-gateway/commit/30bbf335cfc6f9c4a45068c3298837cb6a3d24aa, which as anticipated puts the logic in createOffer/createAnswer instead. An example of usage:

```
echotest.createOffer(
	{
		// No media provided: by default, it's sendrecv for audio and video
		media: { data: true },	// Let's negotiate data channels as well
		rtcConstraints: [
			{ googCpuOveruseDetection: false },
			{ googImprovedWifiBwe: true }
		],
		// etc. etc.
```

so all you need to add is an array called `rtcConstraints` where put the additional constraints you want.

Tested and it seems to be working fine for me, so I'll close this PR. Thanks for your contribution and for noticing about the need for having something like this!

Issue #1030, opened 2017-10-12
Videoroom does not notify "joining" to participants


tugtugtug:
When a participant joins the room, no event is sent to the other participants, i.e. someone can join the room without others knowing and start listening to the conference, this is inconsistent with the other room plugins, and is also inconsistent with the leaving event.

1. publisher joins the room, no notification to participants
2. publisher leaves the room, "leaving" sent to participants

Much like the other plugins, the videoroom plugin should also offer a "joining" event to match the "leaving" event.

lminiero:
Not a bug. The plugin only advertizes active feeds (publishers sending media), not participants in general. Closing.

tugtugtug:
but the leaving event is published anyway even if the publisher doesn't publish

PR #1031, opened 2017-10-12
Fix issue #1030 by introducing a new flag to videoroom for joining events

tugtugtug:
With private_id and require_pvtid, one can actually create a videoroom
with proper identity attached to every listener. However, there is no
event delivered when a listening only participant joins the room. This
requires the admin to have to poll the participants list to see if any
participants have joined and may have already been listening.
Added a new flag notify_joining to create API of the videoroom to allow
enabling an extra event sent when a participant joins the room. This
allows proper management of listening only participants by the admin.

Verified that with notify_joining enabled, a joining event is delivered
to all current participants with the new participants id and display.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -78,7 +78,11 @@ playoutdelay_ext = yes|no (whether the playout-delay RTP extension must be
 	negotiated/used or not for new publishers, default=yes)
 record = true|false (whether this room should be recorded, default=false)
 rec_dir = <folder where recordings should be stored, when enabled>
-notify_joining = yes|no (whether to notify all participants when a new participant joins the room)
+notify_joining = true|false (optional, whether to notify all participants when a new
+            participant joins the room. The Videoroom plugin by design only notifies
+            new feeds (publishers), and enabling this may result extra notification
+            traffic. This flag is particularly useful when enabled with \c require_pvtid
+            for admin to manage listening only participants. default=false)
 \endverbatim
  *
  * Note that recording will work with all codecs except iSAC.
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -78,6 +78,7 @@ playoutdelay_ext = yes|no (whether the playout-delay RTP extension must be
 	negotiated/used or not for new publishers, default=yes)
 record = true|false (whether this room should be recorded, default=false)
 rec_dir = <folder where recordings should be stored, when enabled>
+notify_joining = yes|no (whether to notify all participants when a new participant joins the room)
 \endverbatim
  *
  * Note that recording will work with all codecs except iSAC.
@@ -240,7 +241,8 @@ static struct janus_json_parameter create_parameters[] = {
 	{"playoutdelay_ext", JANUS_JSON_BOOL, 0},
 	{"record", JANUS_JSON_BOOL, 0},
 	{"rec_dir", JSON_STRING, 0},
-	{"permanent", JANUS_JSON_BOOL, 0}
+	{"permanent", JANUS_JSON_BOOL, 0},
+	{"notify_joining", JANUS_JSON_BOOL, 0},
 };
 static struct janus_json_parameter edit_parameters[] = {
 	{"room", JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE},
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks for the contribution! This might be useful for scenarios like the one you described, but it's not properly documented, especially in the configuration file. It should clarify that by default (and by design) the VideoRoom only notifies about new feeds, and not participants, and that an optional (default=false) property is available for those that want to be notified whenever someone joins in general instead, warning that it might result in more verbose notifications traffic.

tugtugtug:
It's my pleasure! Updated the documentation as suggested

lminiero:
Thanks for the quick fixes! I'll need some time to test and review this though. Before we can merge, can you sign our [CLA](https://cla.conf.meetecho.com/janus/)?

plugins/janus_videoroom.c
```diff
@@ -952,6 +956,10 @@ int janus_videoroom_init(janus_callbacks *callback, const char *config_path) {
 			if(rec_dir && rec_dir->value) {
 				videoroom->rec_dir = g_strdup(rec_dir->value);
 			}
+			/* Default not notifying the joining event */
```
lminiero: "By default we don't notify about participants simply joining a room, but only about those that start publishing something"

plugins/janus_videoroom.c
```diff
@@ -1115,6 +1123,26 @@ static void janus_videoroom_notify_participants(janus_videoroom_participant *par
 	}
 }
 
+static void janus_videoroom_participant_joining(janus_videoroom_participant * p) {
```
lminiero: Remove extra space: `* p`

plugins/janus_videoroom.c
```diff
@@ -1570,6 +1599,8 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session
 		}
 		videoroom->videoorient_ext = videoorient_ext ? json_is_true(videoorient_ext) : TRUE;
 		videoroom->playoutdelay_ext = playoutdelay_ext ? json_is_true(playoutdelay_ext) : TRUE;
+		/* Default not notifying the joining event */
```
lminiero: Same text suggested above.

plugins/janus_videoroom.c
```diff
@@ -3343,6 +3374,9 @@ static void *janus_videoroom_handler(void *data) {
 				json_object_set_new(event, "id", json_integer(user_id));
 				json_object_set_new(event, "private_id", json_integer(publisher->pvt_id));
 				json_object_set_new(event, "publishers", list);
+				/* Notify all participants the publisher has joined the room */
```
lminiero: "We may need to notify all participants that a new user joined the room"

Alternate text since the current comment is broken, as it seems to indicate that: 1) you're notifying about a publisher, which isn't true (no she/he's publishing anything yet), and 2) that this notification will always happen (when that method actually checks the flag to see if it will do that).

plugins/janus_videoroom.c
```diff
@@ -1115,6 +1127,26 @@ static void janus_videoroom_notify_participants(janus_videoroom_participant *par
...
+			json_object_set_new(user, "display", json_string(p->display));
+		}
+		json_object_set_new(event, "videoroom", json_string("event"));
+		json_object_set_new(event, "room", json_integer(p->room->room_id));
+		json_object_set_new(event, "joining", user);
+		janus_mutex_lock(&p->room->participants_mutex);
+		janus_videoroom_notify_participants(p, event);
+		janus_mutex_unlock(&p->room->participants_mutex);
+		json_decref(event);
+		json_decref(user);
```
lminiero: This is not needed: `json_object_set_new` means `event` now owns `user`. Dereferencing `user` at this point is likely going to cause issues and maybe crashes. This might explain why your events were not working as expected for me: the content of "joining" was not what you added, but some Janus core API stuff that Jansson probably incorrectly added due to broken pointers.

lminiero: I can confirm that removing the `json_decref(user)` line fixes the issue for me.

tugtugtug:
Thanks @lminiero , give me a bit of time on the CLA, I'll get back to it tomorrow.

lminiero:
@tugtugtug any news?

tugtugtug:
@lminiero Sorry, got sidetracked by other things, should be okay now.

lminiero:
Thanks, merging then!

PR #1033, opened 2017-10-13
New media plugin to drive logic with Lua

lminiero:
This is a new plugin I've been working on for a few weeks, and that might be of interest to many of you. To keep it short (I'll go in more detail in the next paragraphs), this is a plugin just like EchoTest, VideoRoom, AudioBridge, etc., with the key difference that its C code is mainly made of "building blocks" or helper tools, while the actual logic is delegated to a Lua script you can write separately.

As you can guess, this should make it much easier to develop new Janus plugins, especially for those who have no experience with C development: in fact, all you'd need to do is implement your logic in Lua, tell the plugin you want to use your Lua script via configuration, and everything should "just work". As a proof of concept, I've provided two sample Lua scripts, `echotest.lua` and `videoroom.lua`, which as you can evince basically try to mimic the behaviour of the existing C plugins we already have. As a fun fact, I also recently used this plugin in a [Dangerous Demo at Astricon](https://www.youtube.com/watch?v=fOUiVBaZgnQ&feature=youtu.be&t=44m30s): if you want to check what that looked like, here's a link to the [lua script](https://gist.github.com/lminiero/9aeeda1be501fb636cad0c8057c6e076) I used in that context.

So, coming to some more details, and ripping off the doxygen documentation I started writing...

## What is this?

This is a plugin that implements a simple bridge to Lua scripts. While the plugin implements low level stuff like media manipulation, routing, recording, etc., all the logic is demanded to an external Lua script. This means that the C code exposes functions to the Lua script (e.g., to dictate what to do with media, whether recording should be done, sending PLIs, etc.), while Lua exposes functions to be notified by the C code about important events (e.g., new users, WebRTC state, incoming messages, etc.).

Considering the C code and the Lua script will need some sort of "contract" in order to be able to properly interact with each other, the interface (as in method names) must be consistent, but the logic in the Lua script can be completely customized, so that it fits whatever requirement one has (e.g., something like the EchoTest, or something like the VideoRoom).

## Lua interfaces

Every Lua script that wants to implement a Janus plugin must provide the following functions as callbacks (which unsurprisingly are basically the same things C plugins implement):

* `init()`: called when janus_lua.c is initialized;
* `destroy()`: called when janus_lua.c is deinitialized (Janus shutting down);
* `createSession()`: called when a new user attaches to the Janus Lua plugin;
* `destroySession()`: called when an attached user detaches from the Janus Lua plugin;
* `querySession()`: called when an Admin API query for a specific user gets to the Janus Lua plugin;
* `handleMessage()`: called when a user sends a message to the Janus Lua plugin;
* `setupMedia()`: called when a users's WebRTC PeerConnection goes up;
* `hangupMedia()`: called when a users's WebRTC PeerConnection goes down;
* `resumeScheduler()`: called by the C scheduler to resume coroutines.

While `init()`, `destroy()` and `resumeScheduler()` don't need any argument, all other functions expect at the very least a numeric session identifier, that will uniquely address a user in the plugin. Such a value is created dynamically by the C code, and so all the Lua script needs to do is track it as a unique session identifier when handling requests and pushing responses/events/actions towards the C code. Refer to the existing examples (e.g., `echotest.lua`) to see the exact signature for all the above callbacks (yeah I know, I'm lazy).

Notice that, along the above mentioned callbacks, Lua scripts can also implement functions like `incomingRtp()`, `incomingRtcp()` and `incomingData()` to handle those packets directly, instead of letting the C code worry about relaying/processing them. While it might make sense to handle incoming data channel messages with `incomingData()`, though, the performance impact of directly processing and manipulating RTP an RTCP packets is probably too high, and so their usage is currently discouraged.

## C interfaces

Just as the Lua script needs to expose callbacks that the C code can invoke, the C code exposes methods as Lua functions accessible from the Lua script. This includes means to push events, configure how media should be routed without handling each packet in Lua, sending RTCP feedback, start/stop recording and so on.

The following are the functions the C code exposes:

* `pushEvent()`: push an event to the user via Janus API;
* `notifyEvent()`: send an event to Event Handlers;
* `closePc()`: force the closure of a PeerConnection;
* `configureMedium()`: specify whether audio/video/data can be received/sent;
* `addRecipient()`: specify which user should receive a user's media;
* `removeRecipient()`: specify which user should not receive a user's media anymore;
* `setBitrate()`: specify the bitrate to force on a user via REMB feedback;
* `setPliFreq()`: specify how often the plugin should send a PLI to this user;
* `sendPli()`: send a PLI (keyframe request);
* `startRecording()`: start recording audio, video and or data for a user;
* `stopRecording()`: start recording audio, video and or data for a user;
* `pokeScheduler()`: notify the C code that there's a coroutine to resume.

As anticipated in the previous section, almost all these methods also expect the unique session identifier to address a specific user in the plugin. This is true for all the above methods expect `pokeScheduler()` which, together with Lua's `resumeScheduler()`, will be clearer in the next section.

## Lua/C coroutines scheduler

Lua is a single threaded environment. While it has a concept similar to threads called coroutines, these are not threads as known in C. In order to allow for an easy to implement asynchronous behaviour in Lua scripts, you can leverage a scheduler implemented in the C code.

More specifically, when the plugin starts a dedicated thread is devoted to the only purpose of acting as a scheduler for Lua coroutines. This means that, whenever this C scheduler is awaken, it will call the `resumeScheduler()` function in the Lua script, thus allowing the Lua script to execute one or more pending coroutines. The C scheduler only acts when triggered, which means it's up to the Lua script to tell it when to wake up: this is possible via the `pokeScheduler()` function, which does nothing more than sending a simple signal to the C scheduler to wake it up. As such, it's easy for the Lua script to implement asynchronous behaviour, e.g.:

1. Lua script needs to do something asynchronously;
2. Lua script creates coroutine, and takes note of it somewhere;
3. Lua script calls `pokeScheduler()`;
4. C code sends signal to the thread acting as a scheduler;
5. when the scheduling thread wakes up, it calls `resumeScheduler()`;
6. Lua script resumes the previously queued coroutine.

This simple mechanism is what the sample Lua scripts provided in this repo use, for instance, to handle incoming messages asynchronously, so you can refer to those to have an idea of how it can be used.

Note that you're obviously free to implement asynchronous behaviour any way you want, and you're not required to use this C scheduler. Anyway, you must implement a method called `resumeScheduler()` anyway, as the C code checks for its presence and fails if it's not there. If you don't need it, just create an empty function that does nothing and you'll be fine.

## Conclusions

I guess that's all, for now. Please let me know what you think of this!

Of course, before you get too excited and start basing all you have on this new plugin, remember this is obviously all still quite WIP, and there are many rough edges. I'm an absolute beginner with Lua, and so my scripts and code there very likely suck: I spent a lot of time writing a simple SDP parser, for instance, and while it kinda works it's been a pain. The existing samples, while they kinda work, are not feature pair with their C counterparts, and the error management is sketchy at best. Besides, there may be several improvements needed in the code that bridges Lua and C: the C scheduler may need some more work, for instance; besides, Lua states are not thread safe, which means we completely lock the Lua script for every request we handle, which means we'll have to be extra-careful to make sure this won't add perceivable delays.

That said, I believe this is a good starting point, and I encourage you to start playing with this if you're interested, as I'd definitely benefit from feedback and guidance to improve this effort. Just FYI, it's based on the reference counters branch, and not master.

Thanks!


chadfurman:
:tada:  Really exciting!

mirkobrankovic:
Good job !! :+1:

thehunmonkgroup:
Exciting! I've used Lua for years as an interface to FreeSWITCH, and really love working with it. I don't imagine I'll have much time to be writing Lua-based Janus plugins, but my chops are pretty good (on the scripting side, not the C API side), so if anyone has questions related to that, I may be able to help.

lminiero:
Nice! Considering your familiarity with Lua, I guess that even generic feedback on the approach we've taken, and if you feel there are ways some things could be done better, would definitely help!

deanelwood:
Great idea Lorenzo. Like others here I’ve had many applications built using Lua under FreeSWITCH which have worked out really well (simple code and pretty performant).

I can see this opening up some creative possibilities for a lot of developers that wouldn’t normally be comfortable with C.

Great job!

>> I spent a lot of time writing a simple SDP parser….<<

This has to be one of the most common sentences in VoIP circles over the last 6 years…. ;)

lminiero:
> This has to be one of the most common sentences in VoIP circles over the last 6 years…. ;)

Yeah, and after writing the SDP utils in C I really didn't think I would have needed to go through all that again... :smile:  I guess I could have added some more wrappers to the C methods we have to use those from Lua, but for some reason it didn't look as clean. Besides, SDP is basically strings manipulation, which in theory is easier to do in Lua. Hopefully the code I added will be improved along the road.

ivanovaleksey:
Thank you for your work @lminiero!

Isn't it a completely new approach for building plugins? :)

I am totally new to Lua. Does all that mean that one should have Lua installed in order to compile Janus? Do these *.lua files need to be compiled somehow?

> Lua states are not thread safe, which means we completely lock the Lua script for every request we handle
I am afraid I don't understand it. Could you please explain what does "completely lock the Lua script" mean?

chadfurman:
The *.lua scripts don't need to be compiled (they're scripts) -- however, I do think Janus will need to be restarted to reload the script.

https://github.com/meetecho/janus-gateway/pull/1033/files#diff-67e997bcfdac55191033d57a16d1408aR559  This mentions `lua-libs` as a dependency and references the README file for install instructions, however the readme file on the branch doesn't contain a mention of lua as far as I can tell -- @lminiero did I miss the install instructions for lua-libs?

lminiero:
> I am afraid I don't understand it. Could you please explain what does "completely lock the Lua script" mean?

It means we have a global mutex, and we lock it every time we need to invoke a Lua function, no matter who originates it (different users or the scheduler), and we unlock it when that function returns. You can't use the Lua state (the script) at the same time from different threads.

> the readme file on the branch doesn't contain a mention of lua as far as I can tell -- @lminiero did I miss the install instructions for lua-libs?

The README hasn't been updated yet.

chadfurman:
@ivanovaleksey Lua docs talk more about installation:  http://www.troubleshooters.com/codecorn/lua/lua_c_calls_lua.htm

The header files in their examples are the same as @lminiero's examples, and so the dependencies are probably the same :)

lminiero:
One thing I should point out is that what we check via pkgconfig in our
configure is the name the package has on fedora, my distro: I think one of
my colleagues told me the name to check on Ubuntu is different (not sure
how as I don't have my laptop with me) and so we'll have to update che
check to make sure the configure can actually always find your library when
installed. Not sure if other distros make this even harder or not.

Il 15 ott 2017 3:24 AM, "Chad Furman" <notifications@github.com> ha scritto:

@ivanovaleksey <https://github.com/ivanovaleksey> Lua docs talk more about
installation: http://www.troubleshooters.com/codecorn/lua/lua_c_calls_
lua.htm

The header files in their examples are the same as @lminiero
<https://github.com/lminiero>'s examples, and so the dependencies are
probably the same :)

—
You are receiving this because you were mentioned.
<https://github.com/meetecho/janus-gateway/pull/1033#issuecomment-336679076>,
or mute the thread

chadfurman:
I've been trying to get a Dockerfile together to compile this, and in the process I noticed this:

```
libsrtp version:           2.0.x
SSL/crypto library:        OpenSSL
DTLS set-timeout:          not available
DataChannels support:      yes
Recordings post-processor: yes
TURN REST API client:      yes
Doxygen documentation:     no
Transports:
    REST (HTTP/HTTPS):     no
    WebSockets:            yes (old API)
    RabbitMQ:              no
    MQTT:                  no
    Unix Sockets:          no
Plugins:
    Echo Test:             yes
    Streaming:             yes
    Video Call:            yes
    SIP Gateway:           yes
    Audio Bridge:          yes
    Video Room:            yes
    Voice Mail:            yes
    Record&Play:           yes
    Text Room:             yes
Event handlers:
    Sample event handler:  yes


```

It seems the Lua plugin isn't listed here -- not sure if it will matter?

chadfurman:
@lminiero I tried getting this to build in a docker container, and I ran into the Lua libs dependency error you mentioned (i.e. despite it already being installed; probably the package name is different?)

https://github.com/chadfurman/docker-janus/blob/master/Dockerfile#L140

lminiero:
@chadfurman what's the result of `pkg-config --list-all | grep lua`?

chadfurman:
@lminiero looks like it's not even listed in the container despite being installed from source on image build (on my Ubuntu host it's called lua50)

```
lotus@lappy:~/code/docker-janus$ docker run -t janus-lua pkg-config --list-all
gobject-2.0           GObject - GLib Type, Object, Parameter and Signal Library
gthread-2.0           GThread - Thread support for GLib
systemd               systemd - systemd System and Service Manager
sofia-sip-ua          sofia-sip-ua - Sofia-SIP - a RFC3261 compliant SIP User-Agent library
gmodule-2.0           GModule - Dynamic module loader for GLib
gmodule-export-2.0    GModule - Dynamic module loader for GLib
gio-unix-2.0          GIO unix specific APIs - unix specific headers for glib I/O library
libpcrecpp            libpcrecpp - PCRECPP - C++ wrapper for PCRE
openssl               OpenSSL - Secure Sockets Layer and cryptography libraries and tools
jansson               Jansson - Library for encoding, decoding and manipulating JSON data
libsrtp2              libsrtp2 - Library for SRTP (Secure Realtime Transport Protocol)
libwebsockets         libwebsockets - Websockets server and client library
libcrypto             OpenSSL-libcrypto - OpenSSL cryptography library
zlib                  zlib - zlib compression library
gio-2.0               GIO - glib I/O library
tic                   tic - ncurses 5.9 add-on library
libpcre               libpcre - PCRE - Perl compatible regular expressions C library with 8 bit character support
gmodule-no-export-2.0 GModule - Dynamic module loader for GLib
tinfo                 tinfo - ncurses 5.9 terminal interface library
opus                  Opus - Opus IETF audio codec (floating-point build)
nice                  libnice - ICE library
udev                  udev - udev
libcurl               libcurl - Library to transfer files with ftp, http, etc.
shared-mime-info      shared-mime-info - Freedesktop common MIME database
libpcreposix          libpcreposix - PCREPosix - Posix compatible interface to libpcre
ogg                   ogg - ogg is a library for manipulating ogg bitstreams
libssl                OpenSSL - Secure Sockets Layer and cryptography libraries
glib-2.0              GLib - C Utility Library

```

lminiero:
Looks like a setup issue then... what distro is this? If debian-based, have you tried installing the `liblua5.3-dev` package? Not sure what it's called on others.

chadfurman:
debian:jessie is the Dockerfile image base, which only has lua 5.2  --  I'm rebuilding the image with debian:stretch and will report back

chadfurman:
Debian Stretch seems to have a library incompatibility (though does have the liblua5.3-dev package)

```
checking whether to use /dev/console for error reporting... no
checking for pkg-config... yes
configure: creating ./config.status
config.status: creating libsrtp2.pc
config.status: creating Makefile
config.status: creating crypto/Makefile
config.status: creating doc/Makefile
config.status: creating crypto/include/config.h
gcc -DHAVE_CONFIG_H -Icrypto/include -I./include -I./crypto/include -fPIC  -fPIC -Wall -O4 -fexpensive-optimizations -funroll-loops -I/include  -c srtp/srtp.c -o srtp/srtp.o
In file included from srtp/srtp.c:52:0:
crypto/include/aes_gcm_ossl.h:58:20: error: field 'ctx' has incomplete type
     EVP_CIPHER_CTX ctx;
                    ^~~
Makefile:101: recipe for target 'srtp/srtp.o' failed
make: *** [srtp/srtp.o] Error 1
The command '/bin/sh -c export JANUS_WITH_POSTPROCESSING="${JANUS_WITH_POSTPROCESSING}"    && export JANUS_WITH_BORINGSSL="${JANUS_WITH_BORINGSSL}"    && export JANUS_WITH_DOCS="${JANUS_WITH_DOCS}"    && export JANUS_WITH_REST="${JANUS_WITH_REST}"    && export JANUS_WITH_DATACHANNELS="${JANUS_WITH_DATACHANNELS}"    && export JANUS_WITH_WEBSOCKETS="${JANUS_WITH_WEBSOCKETS}"    && export JANUS_WITH_MQTT="${JANUS_WITH_MQTT}"    && export JANUS_WITH_PFUNIX="${JANUS_WITH_PFUNIX}"    && export JANUS_WITH_RABBITMQ="${JANUS_WITH_RABBITMQ}"    && export JANUS_WITH_FREESWITCH_PATCH="${JANUS_WITH_FREESWITCH_PATCH}"    && export JANUS_BUILD_DEPS_DEV="${JANUS_BUILD_DEPS_DEV}"    && export JANUS_CONFIG_OPTIONS="${JANUS_CONFIG_OPTIONS}"    && if [ $JANUS_WITH_POSTPROCESSING = "1" ]; then export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --enable-post-processing"; fi     && if [ $JANUS_WITH_BORINGSSL = "1" ]; then export JANUS_BUILD_DEPS_DEV="$JANUS_BUILD_DEPS_DEV golang-go" && export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --enable-boringssl --enable-dtls-settimeout"; fi     && if [ $JANUS_WITH_DOCS = "1" ]; then export JANUS_BUILD_DEPS_DEV="$JANUS_BUILD_DEPS_DEV doxygen graphviz" && export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --enable-docs"; fi     && if [ $JANUS_WITH_REST = "1" ]; then export JANUS_BUILD_DEPS_DEV="$JANUS_BUILD_DEPS_DEV libmicrohttpd-dev"; else export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --disable-rest"; fi     && if [ $JANUS_WITH_DATACHANNELS = "0" ]; then export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --disable-data-channels"; fi     && if [ $JANUS_WITH_WEBSOCKETS = "0" ]; then export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --disable-websockets"; fi     && if [ $JANUS_WITH_MQTT = "0" ]; then export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --disable-mqtt"; fi     && if [ $JANUS_WITH_PFUNIX = "0" ]; then export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --disable-unix-sockets"; fi     && if [ $JANUS_WITH_RABBITMQ = "0" ]; then export JANUS_CONFIG_OPTIONS="$JANUS_CONFIG_OPTIONS --disable-rabbitmq"; fi     && /usr/sbin/groupadd -r janus && /usr/sbin/useradd -r -g janus janus     && DEBIAN_FRONTEND=noninteractive apt-get update     && DEBIAN_FRONTEND=noninteractive apt-cache search lua 5.3      && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-recommends install $JANUS_BUILD_DEPS_DEV ${JANUS_BUILD_DEPS_EXT}     && curl -fSL https://github.com/cisco/libsrtp/archive/v2.0.0.tar.gz -o ${BUILD_SRC}/v2.0.0.tar.gz     && tar xzf ${BUILD_SRC}/v2.0.0.tar.gz -C ${BUILD_SRC}     && cd ${BUILD_SRC}/libsrtp-2.0.0     && ./configure --prefix=/usr --enable-openssl     && make shared_library     && make install     && if [ $JANUS_WITH_BORINGSSL = "1" ]; then git clone https://boringssl.googlesource.com/boringssl ${BUILD_SRC}/boringssl     && cd ${BUILD_SRC}/boringssl     && sed -i s/" -Werror"//g CMakeLists.txt     && mkdir -p ${BUILD_SRC}/boringssl/build     && cd ${BUILD_SRC}/boringssl/build     && cmake -DCMAKE_CXX_FLAGS="-lrt" ..     && make     && mkdir -p /opt/boringssl     && cp -R ${BUILD_SRC}/boringssl/include /opt/boringssl/     && mkdir -p /opt/boringssl/lib     && cp ${BUILD_SRC}/boringssl/build/ssl/libssl.a /opt/boringssl/lib/     && cp ${BUILD_SRC}/boringssl/build/crypto/libcrypto.a /opt/boringssl/lib/     ; fi     && if [ $JANUS_WITH_DATACHANNELS = "1" ]; then git clone https://github.com/sctplab/usrsctp ${BUILD_SRC}/usrsctp     && cd ${BUILD_SRC}/usrsctp     && ./bootstrap     && ./configure --prefix=/usr     && make     && make install     ; fi     && if [ $JANUS_WITH_WEBSOCKETS = "1" ]; then git clone https://github.com/warmcat/libwebsockets.git ${BUILD_SRC}/libwebsockets     && cd ${BUILD_SRC}/libwebsockets     && mkdir ${BUILD_SRC}/libwebsockets/build     && cd ${BUILD_SRC}/libwebsockets/build     && cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_C_FLAGS="-fpic" ..     && make     && make install     ; fi     && if [ $JANUS_WITH_MQTT = "1" ]; then git clone https://github.com/eclipse/paho.mqtt.c.git ${BUILD_SRC}/paho.mqtt.c     && cd ${BUILD_SRC}/paho.mqtt.c     && make     && make install     ; fi     && if [ $JANUS_WITH_RABBITMQ = "1" ]; then git clone https://github.com/alanxz/rabbitmq-c ${BUILD_SRC}/rabbitmq-c     && cd ${BUILD_SRC}/rabbitmq-c     && git submodule init     && git submodule update     && autoreconf -i     && ./configure --prefix=/usr     && make     && make install     ; fi     && git clone https://github.com/meetecho/janus-gateway.git ${BUILD_SRC}/janus-gateway     && if [ $JANUS_WITH_FREESWITCH_PATCH = "1" ]; then curl -fSL https://raw.githubusercontent.com/krull/docker-misc/master/init_fs/tmp/janus_sip.c.patch -o ${BUILD_SRC}/janus-gateway/plugins/janus_sip.c.patch && cd ${BUILD_SRC}/janus-gateway/plugins && patch < janus_sip.c.patch; fi     && cd ${BUILD_SRC}/janus-gateway     && git checkout janus-lua     && ./autogen.sh     && ./configure ${JANUS_CONFIG_DEPS} $JANUS_CONFIG_OPTIONS     && make     && make install     && make config     && chown -R janus:janus /opt/janus     && cd ${BUILD_SRC}     && if [ $JANUS_WITH_BORINGSSL = "1" ]; then rm -rf boringssl; fi     && if [ $JANUS_WITH_DATACHANNELS = "1" ]; then rm -rf usrsctp; fi     && if [ $JANUS_WITH_WEBSOCKETS = "1" ]; then rm -rf libwebsockets; fi     && if [ $JANUS_WITH_MQTT = "1" ]; then rm -rf paho.mqtt.c; fi     && if [ $JANUS_WITH_RABBITMQ = "1" ]; then rm -rf rabbitmq-c; fi     && rm -rf         v2.0.0.tar.gz         libsrtp-2.0.0         janus-gateway     && DEBIAN_FRONTEND=noninteractive apt-get -y --auto-remove purge ${JANUS_BUILD_DEPS_EXT}     && DEBIAN_FRONTEND=noninteractive apt-get -y clean     && DEBIAN_FRONTEND=noninteractive apt-get -y autoclean     && DEBIAN_FRONTEND=noninteractive apt-get -y autoremove     && rm -rf /usr/share/locale/*     && rm -rf /var/cache/debconf/*-old     && rm -rf /usr/share/doc/*     && rm -rf /var/lib/apt/*' returned a non-zero code: 2

```

I've also pushed an update to the docker image to include Debian Stretch as well as the lua dependency: https://github.com/chadfurman/docker-janus/blob/master/Dockerfile#L42

lminiero:
Not familiar with Debian so can't help, sorry.

lminiero:
For people interested in this effort, I wrote a small tutorial on how to develop in Lua for Janus on our blog: http://www.meetecho.com/blog/tutorial-writing-a-janus-video-call-plugin-in-lua/

soulfly:
That's great news! Customising existing plugins in C is a huge pain for non-C developers :( So write plugins in Lua is a huge step forward 🥇

chadfurman:
Here is a docker container you can user to run this branch.  It's setup to checkout this branch automatically and to build all the lua deps needed for the echotest plugin.

The file isn't very well tested and might need you to EXPOSE some ports from the Dockerfile or to add some port mappings in the docker-compose.yaml if you go that route.

I just run `docker build -t janus-lua .` inside of the docker-janus folder, then I run `docker run -it janus-lua:latest`

chadfurman:
https://github.com/chadfurman/docker-janus   

Helps if I post the link.

lminiero:
This latest commit adds a new mechanism that was missing in the C/Lua interaction, that is a way for Lua to ask for a specific callback to be invoked after a provided amount of time. This should be quite helpful as the only mechanism we had so far for asynchronous behaviour was the C scheduler, but that triggers `resumeScheduler` ASAP. Besides, this new mechanism you to specify the function to invoke, an optional argument to pass (which MUST be a string) and the time to wait to do that. It should be particularly helpful when you're handling asynchronous behaviour that you want to inspect on a regular basis. Since it uses the GLib loop/context/source stuff, and there's the C/Lua context change to take into account, don't rely too much on it to be milliseconds precise...

The way it works is pretty simple:

```
function test(param)
	logger.print("Invoked test")
	if param == nil then
		logger.print("  -- No param")
	else
		logger.print("  -- Param: " .. param)
	end
end

-- This will cause an error (timeCallback needs three arguments)
timeCallback()
-- Invoke test() in 500 milliseconds
timeCallback("test", nil, 500)
-- Invoke test("ciccio") in 2 seconds
timeCallback("test", "ciccio", 2000)
```

I guess this might make the `pokeScheduler`/`resumeScheduler` mechanism reduntant, as it can be easily recreated using the timeCallback above (`timeCallback("resumeScheduler", nil, 0)`), but I'd rather keep them both in for the time being, if not because `pokeScheduler` is much less verbose.

Feedback welcome as usual.

plugins/janus_lua.c
```diff
@@ -234,6 +234,12 @@ typedef struct janus_lua_callback {
 	char *argument;
 } janus_lua_callback;
 
+/* helper function to sample the number of occupied slots into Lua stack */
```
lminiero: Helper (capitalized H)

plugins/janus_lua.c
```diff
@@ -234,6 +234,12 @@ typedef struct janus_lua_callback {
...
+static void janus_lua_stackdump(lua_State* l)
+{
```
lminiero: { on the same line as the function name (coding style of the project).

plugins/janus_lua.c
```diff
@@ -234,6 +234,12 @@ typedef struct janus_lua_callback {
...
+    int top = lua_gettop(l);
+    JANUS_LOG(LOG_WARN, "Total in lua stack %d\n", top);
```
lminiero: Is this temporary? How verbose/spammy is this for now? Make it a LOG_HUGE so that people can do LOG_WARN only if they want.

plugins/janus_lua.c
```diff
@@ -1246,6 +1250,7 @@ void janus_lua_create_session(janus_plugin_session *handle, int *error) {
 	lua_getglobal(t, "createSession");
 	lua_pushnumber(t, session->id);
 	lua_call(t, 1, 0);
+	lua_pop(lua_state,1);
```
lminiero: Space missing after the comma (coding style). May be the same in other lines, haven't checked.

Apart from that, why is `createSession` creating a thread, and not `init`, when both pass an argument and ignore the result? What's the rationale behind choosing whether to use the the state directly or creating a thread instead?

plugins/janus_lua.c
```diff
@@ -1346,9 +1353,17 @@ struct janus_plugin_result *janus_lua_handle_message(janus_plugin_session *handl
 	lua_getglobal(t, "handleMessage");
 	lua_pushnumber(t, session->id);
 	lua_pushstring(t, transaction);
-	lua_pushstring(t, message_text);
-	lua_pushstring(t, jsep_text);
-	lua_call(t, 4, 2);
+	uint8_t nargs = 2;
+	if(message_text != NULL) {
```
lminiero: This is going to break plugins if message is NULL but JSEP isn't, because they'll end up reading JSEP where they thing message is. Message must always be passed, even if NULL, because we have a function signature to respect: JSEP can be omitted instead.

marcelrouw:
Which lua module need I to install for json? On [luarocks](https://luarocks.org/search?q=json) I can find multiple json implementations
luarocks install json4lua
luarocks install ansicolors

I get the following error when running the lua echo test
`[ERR] [plugins/janus_lua.c:janus_lua_method_pushevent:376] JSON error: on line 1: control character 0xd near '"v=0'`

alexamirante:
@marcelrouw it's `luajson`

lminiero:
Small change to allow a plugin to override the plugin information (version, description, name, author, package name) in case they want to: in case they're not provided, the plugin returns the same info it returned before, that is the info of the Lua plugin itself (`janus.plugin.lua`, etc.). I added an example on how to do that in `echotest.lua`, which will show this when querying `/janus/info`:

      [..]
      },
      "janus.plugin.echolua": {
         "name": "Lua based EchoTest",
         "author": "Lorenzo Miniero",
         "description": "This is echotest.lua, a Lua based clone of janus.plugin.echotest",
         "version_string": "0.0.12",
         "version": 12
      },
      [..]

marcelrouw:
I'm currently writing a C function for playing a recording in Lua.
For getting `janus_lua_session` I need to call the `janus_lua_lookup_session`.
But how do I access `janus_plugin_session` from `janus_lua_extra.c` doesn't seems to be a reference to it present?

`janus_lua_session *janus_lua_lookup_session(janus_plugin_session *handle);`

lminiero:
@marcelrouw are you using the latest code? we've added this to `janus_lua_data.h` some time ago already: https://github.com/meetecho/janus-gateway/blob/6f59999761bb14bda6c92b00d5c9b03ce2fadc62/plugins/janus_lua_extra.c#L26

marcelrouw:
I have `janus_lua_data.h` included and I can call the function `janus_lua_lookup_session`.
But this function needs a parameter `janus_plugin_session *handle`. The problem is I can't find this parameter.

lminiero:
Check how the other functions in the existing code do the same. As explained in my introduction to the plugin, Lua script and C code use a unique integer identifier to address plugin sessions, which means what's what you'll have to use around: in C, there's a hashtable to recover the plugin session from that ID.

marcelrouw:
How can I get the parameter `janus_lua_plugin`? It's needed for the functions `notify_event` and `push_event`. In de scope of `janus_lua.c` it's of course available but in `janus_lua_extra.c` I can't access it.

chadfurman:
It looks like there might be need for a shared header file between all the lua C files?

marcelrouw:
@chadfurman there are already 2 header files available. I think that one of these can be used 
to achieve this.

```
/* Session definition and hashtable */
#include "janus_lua_data.h"
/* Extra/custom C hooks and code */
#include "janus_lua_extra.h"
```

lminiero:
There already is a shared header. You get the `janus_lua_plugin` instance by doing a lookup on the `lua_sessions' table, using the unique identifier the C and Lua code share as a way to address sessions. All the methods have an `id` parameter, that's what you use (and you can just see what they do as an example).

marcelrouw:
It this the right way to fetch the `janus_plugin`?
```
janus_lua_session *session = g_hash_table_lookup(lua_ids, GUINT_TO_POINTER(id));
janus_ice_handle *ice_handle = (janus_ice_handle *)session->handle->gateway_handle;
janus_plugin *plugin = (janus_plugin *)ice_handle->app;
```
It seems a bit far fetched. 
Also I run into an issue while compiling this code because `janus_ice_handle` is available in `ice.h`. After adding `#include "../ice.h"` in the `janus_lua_extra.c` I get the following error:
```
In file included from plugins/janus_lua_extra.c:26:
plugins/../ice.h:21:10: fatal error: 'agent.h' file not found
#include <agent.h>
```

lminiero:
@marcelrouw you don't have access to `janus_ice_handle` in plugins: that's core stuff. You only need the `janus_lua_session` instance, and the `handle` property to interact with the core. The `handle` you see in `janus_lua_session` (and other core-to-plugin interactions) is not the same handle used in the core.

This is all you need to do:

```
janus_lua_session *session = g_hash_table_lookup(lua_ids, GUINT_TO_POINTER(id));
```

so that you can use the handle to do whatever you need, e.g.:

```
janus_core->push_event(session->handle, &janus_lua_plugin, transaction, event, NULL);
```

marcelrouw:
How do I get access to `&janus_lua_plugin`? 
It's available in `janus_lua.c` but in `janus_lua_extra.c` I can't access it.
```
plugins/janus_lua_extra.c:121:44: error: use of undeclared identifier 'janus_lua_plugin'; did you mean 'janus_log_prefix'?
janus_core->push_event(session->handle, &janus_lua_plugin, "transaction", info, NULL);
```

lminiero:
Apologies, I meant `janus_lua_session`, not `janus_lua_plugin`. Simple example:

```
	janus_mutex_lock(&lua_sessions_mutex);
	janus_lua_session *session = g_hash_table_lookup(lua_ids, GUINT_TO_POINTER(id));
	janus_mutex_unlock(&lua_sessions_mutex);
```

Hashtable, struct definition and locks are all accessible in `janus_lua_extra.c` (just tried). If this is still unclear, I encourage you to study the stock plugins first, and then delve in how the C part of the Lua plugin works.

marcelrouw:
My current issue is not with getting the session. You put me on the right track some time a go with getting the session. `janus_lua.c` got some good examples for that. 

Currently I'm building a playout function for a recording based on the `janus_recordplay.c` plugin. I already got most of this working. But still have some things to improve, fix and test. One of those things is sending a `push_event`. Normally you would do this from the lua code. But in this case it needs to be send at the end of the playout thread (`g_thread_try_new`) and that makes it impossible send the `push_event` from the lua code. So I'm doing it from the C code.

My current issue is with `push_event` as the second parameter it needs `&janus_lua_plugin`. This parameter is not available in `janus_lua_extra.c` and I don't know how to get it. (I left out the mutex and refcount to keep the example small)

```
static int janus_lua_method_test(lua_State *lua_state) {
	guint32 id = lua_tonumber(lua_state, 1);
	janus_lua_session *session = g_hash_table_lookup(lua_ids, GUINT_TO_POINTER(id));

	json_t *info = json_object();
	json_object_set_new(info, "event", json_string("playing"));

	janus_core->push_event(session->handle, &janus_lua_plugin, "dummy", info, NULL);
}
```

If you put this piece of code in the `janus_lua_extra.c`. It won't compile and give the error `use of undeclared identifier 'janus_lua_plugin';`. 

```
plugins/janus_lua_extra.c:51:43: error: use of undeclared identifier 'janus_lua_plugin'; did you mean 'janus_log_prefix'?
        janus_core->push_event(session->handle, &janus_lua_plugin, "dummy", info, NULL);
```

(This piece of code actually does compile if you put it in `janus_lua.c`)

lminiero:
> My current issue is with push_event as the second parameter it needs &janus_lua_plugin.

Oh ok I think I got it now. Yes, in that case you might want to move the `janus_lua_plugin` [definition](https://github.com/meetecho/janus-gateway/blob/6f59999761bb14bda6c92b00d5c9b03ce2fadc62/plugins/janus_lua.c#L214) to the `janus_lua_data.h` header as other stuff, although it might take some work: not sure if just moving it will be enough, at the very least you'll have to get rid of the static bit.

A much easier way, though, might be relying on the Lua script for that. You do your stuff in the new C code, and delegate any notification to a new Lua function instead: any time you need to notify something, you call that Lua function from C with the right parameters, e.g.:

     notifyUser(id, content);

and have that Lua script take care of sending the `pushEvent()` for you. This is what we did in some private experiments and repositories, and it works well.

soulfly:
@lminiero a quick comment  - will videoroom.lua be a full copy of videoroom.c from features point of view? This is very powerful feature. It's hard to customize current C implementation for non-C developers, so this lua plugin will be a huge step forward

lminiero:
@soulfly no, it's not feature pair: only the basic SFU features are there. While most of the requests are the same, the Lua version doesn't load a configuration file, and doesn't expose as many configuration options either (e.g., those based on extensions). RTP forwarding is missing as well. That said, it should be easier to expand than the C version for those that want more "meat" in it.

chadfurman:
+1 for feature pairity ;) 

RTP forwarding will definitely be a nice addition in the future...

chadfurman:
we get lots of collisions using the random-suffix feature here.  Is the lua math.random() library ever initialized with os.time() ?  We are replacing random suffix with epoch timestamp suffixes (and protecting everything with an api secret and an external authentication layer)

lminiero:
I don't think we're initializing anything for the randomizer in the Lua scripts. If you let us know what to add and where we'll do that.

lminiero:
> in general you need to seed once for the whole app

So I guess something to invoke in the `init()` callback.

chadfurman:
sounds right.  Was not sure where it would go, myself. Thank you :)

lminiero:
Since we just merged the refcount branch, and this work here is basically done, I guess we can merge the Lua plugin too. Thank you all for the feedback!

Issue #1026, opened 2017-10-10
Janus mistakenly ignores trickle candidates if the candidates are received when cleaning up a previous ice session


tugtugtug:
Steps to reproduce:
1. have a client publishing video/audio to the videoroom.
2. unpublish
3. publish again in a short period of time.

What happens here is in janus.c we have the sdp being held up due to JANUS_ICE_HANDLE_WEBRTC_CLEANING being set, however the subsequent tickle ice candidates received are being send to the ice session that is cleaning up and effectively ignored. Since there is no error reported either way, we end up with a broken ice state.

I think we should clean up the pending candidates list as soon as we receive a new offer and allow caching the candidates from then on.

> If you want to open an issue, please make sure that:
> 
> 1. you actually found a bug: for generic questions, use the [meetecho-janus](http://groups.google.com/d/forum/meetecho-janus) group instead;
> 2. nobody opened the same bug already (do a search);
> 3. the issue wasn't already solved;
> 4. you don't paste a huge amount of text in the issue or comments: use a service like [pastebin](http://pastebin.com/) or similar;
> 5. you provide a GDB stacktrace if Janus crashed on you, and/or output from AddressSanitizer or Valgrind (more details available[here](http://janus.conf.meetecho.com/docs/debug.html)).
> 
> If so, just remove this text and share your findings!

lminiero:
I think this is part of the fixes @atoppi is working on in this branch here: https://github.com/meetecho/janus-gateway/tree/iceloop-sessions-mgmt (not in a PR yet)
If that's not working yet, please sync with him about possible solutions to this issue.

atoppi:
I'm wondering how to deal with this kind of scenario from a Janus API perspective.
I mean, the CLEANING flag could be possibly triggered in such cases:
1) DTLS handshake error
2) `janus_ice_webrtc_hangup`
3) `janus_ice_thread` leaving
The flag is cleared once `janus_ice_webrtc_free` is called, and this function is the one in charge of deleting pending trickles.
So I think that Janus should reject any trickles incoming while doing its internal cleanups and these cleanups should be as fast as possibile (that's why I'm removing the sleep calls).

tugtugtug:
I think it is a very common and useful case to allow quickly  unpublish and republish. We have this case to switch between desktop sharing and camera for example and is hitting this bug every single time. With your change to reject the newer jsep will simply make the client keep trying, which is not quite an ideal scenario for either the client or the server.
[...]

PR #1035, opened 2017-10-16
Better ICE/handle state management

atoppi:
**Update**: some changes have been ported from #1089 

This PR comes after many experiences collected by me and other people using Janus, and tries to address most of the common roots of these issues.
The objective is hardening the management of the handle state into the server, including the ICE-related threads and the plugin sessions.

What is changed into the Janus core:
1) The handle cleanup in `janus_ice_handles_check` is delayed if one of the following conditions is met:
a) `iceloop` is still running
b) `icethread` is still running
c) `send_thread` is still running
This should avoid every future issue about `heap-use-after-free`-like error, triggered by the Janus core that tries to re-use an already freed handle session. Every delay is printed as a WARN on the console.
The rationale here is delay a free until something is still using it, and let the core do its job without abruptly destroying the handle.
2) Removed some `sleep` calls to unhide state management errors. We should not rely on _magical_ sleep times to keep consistent the state.
3) The `janus_ice_dtls_alert` is now front-pushed into the packet queue. This should speed-up the media hangup process.
4) Any trickle/jsep received during the handle CLEANING state is rejected with a new error. This avoids strange behaviors like #1026.
5) The `hangup` response event has been made synchronous with the actual webrtc resources cleanup. This should help in case someone wants to quickly re-create a peer connection after having destroyed one (e.g. rapid unpublish/publish in videoroom).
6) **NEW** Use the active plugin sessions map in place of the old plugin sessions map

What is changed into the Janus plugins (~~only **Streaming** and **Videoroom**~~):
1) The active sessions map is used to validate every callback (besides the ones involving RTP) issued on the plugin and to validate the messages handled in a asynchronous fashion inside the plugin.
2) A destroy request on the plugin immediately deletes a session from the map, making fail every other subsequent request referring to that session. The cleanup is still executed in a lazy manner with a garbage-collector like approach. It is responsibility of the plugins to correctly handle valid sessions and not reuse them after they have been marked as destroyed.
3) Every plugin callback (not inherent to rtp/rtcp) is protected with `sessions_mutex` lock. This will decrease parallelism but highly increase the safety accessing to plugin sessions.
**BE VERY CAREFUL** because you can easily trigger a double locking by calling a lock protected method from inside another lock protected method (e.g. call `hangup_media` from inside `destroy_session`).
**THIS CAN BE DONE INDIRECTLY TOO** e.g. from inside a lock protected plugin callback, call a gateway callback that in turn calls lock protected plugins callback again. 
4)  **NEW** New method `hangup_media_internal` that is the previous `hangup_media`, while now `hangup_media` executes a lock protected `hangup_media_internal`. This is necessary to avoid double-locking issues (refer to 3) in `destroy_session`methods. Every plugin calls `hangup_media_internal` in `destroy_session`.

What is changed into the Streaming and Recordplay plugins:
1)  **NEW** There is no more a simulated `stop` message to trigger a hangup. Now the hangup is executed in a synchronous fashion.


lminiero:
Looks good to me, merging!

PR #1036, opened 2017-10-20
Use supplied WebRTC adapter throughout file

op:
Some calls were using adapter from the global namespace.


lminiero:
Good catch, thanks for spotting those! Could you sign our [CLA](https://cla.conf.meetecho.com/janus/) so I can merge this right away?

op:
Sorry for the delay! CLA is signed. Cheers!

lminiero:
Thanks, merging then!

PR #1037, opened 2017-10-24
fix libressl call in dtls.c

aduskett:
OpenSSL is reported even if compiled against libressl.
Fix this with a check to see if libressl is defined.

Also, add a ";" after LibreSSL.


lminiero:
Thanks for this fix! Hope this wasn't something I broke myself in recent commits... merging!

PR #1040, opened 2017-10-26
Allow multiple codecs in VideoRoom, and publishers to choose which one

lminiero:
This patch introduces a change in how the VideoRoom handles publishers.

As you know, currently when you create a VideoRoom you can specify which codecs must be forced on publishers: once you pick your choice (e.g., Opus for audio, VP9 for video), they'll be forced upon all publishers; if they support the codec, they'll be able to publish, or otherwise the unsupported media will be rejected (in an Opus/VP9 room Safari would fall back to being an audio-only publisher, for instance).

With this new pull request, this policy is loosened: when creating a room, you can actually choose more than one codec (up to three, for now) per media type, in order of preference, e.g:

    audiocodec = opus,pcmu
    videocodec = vp9,vp8,h264

that is, via a comma separated list of codecs. Specifying a single codec will have the VideoRoom plugin behave exactly as it did so far, so no need to worry about backwards compatibility.

This gives more flexibility to publishers, as they have a wider range of codecs available before they're rejected. In this specific example, preference is given to Opus and VP9: if they're available in the publisher's offer, the VideoRoom will answer with those; if VP9 isn't in the offer, the VideoRoom will check if VP8 is; if it isn't, it checks H.264; if that one is unavailable too, video is rejected.

Publishers also have an option to override the room-specified preference. For instance, if in the above case the publisher supports VP9 but actually wants to use VP8 instead, they can specify it via a `videocodec` property added when publishing:

    [..]
    videocodec: "vp8",
    [..]

The `audiocodec` property is available to override the audio codec preferences instead.

Of course, notice that loosening the constraints with respect to the codecs will also increase the chances of having publishers using codecs some viewers don't support. In this case, you'll have to pay extra attention to what viewer endpoints support, and make sure they subscribe to a subset of the media if some codecs are known not to be supported: an example that shows how Safari can subscribe to audio only if the publisher is VP8/VP9 is provided [here](https://github.com/meetecho/janus-gateway/blob/master/html/videoroomtest.js#L449).

I tested this briefly and it seems to be working as expected for me, in all the scenarios I checked. That said, if this effort interests you please do some checks yourselves as well, especially to make sure it doesn't break anything in your existing application. Feedback welcome!

### DIFF
--- a/conf/janus.plugin.videoroom.cfg.sample
+++ b/conf/janus.plugin.videoroom.cfg.sample
@@ -9,8 +9,10 @@
 ;              conference or 1 for a webinar)
 ; bitrate = <max video bitrate for senders> (e.g., 128000)
 ; fir_freq = <send a FIR to publishers every fir_freq seconds> (0=disable)
-; audiocodec = opus|isac32|isac16|pcmu|pcma (audio codec to force on publishers, default=opus)
-; videocodec = vp8|vp9|h264 (video codec to force on publishers, default=vp8)
+; audiocodec = opus|g722|pcmu|pcma|isac32|isac16 (audio codec(s) to force on publishers, default=opus
+;			can be a comma separated list in order of preference, e.g., opus,pcmu)
+; videocodec = vp8|vp9|h264 (video codec(s) to force on publishers, default=vp8
+;			can be a comma separated list in order of preference, e.g., vp9,vp8,h264)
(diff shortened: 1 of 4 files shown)


lminiero:
Other important point to highlight: flexible codecs means a higher chance of participants using different codecs, which means higher chance that a "switch" request will not be usable (switching from Joe to Bob will give you garbage if they're not using the same codecs).

lminiero:
Don't see any objection, so I'll merge.

PR #1044, opened 2017-10-30
Add optional exchange for RabbitMQ transport

pitkonst:
This allowes to set RabbitMQ exchange to send packets through. Exchange is starting point to build message delivery topologies. It could be heplful in multi application servers to single Janus server use cases.

### DIFF
--- a/transports/janus_rabbitmq.c
+++ b/transports/janus_rabbitmq.c
@@ -107,6 +107,9 @@ static gboolean rmq_janus_api_enabled = FALSE;
 static gboolean rmq_admin_api_enabled = FALSE;
 static gboolean notify_events = TRUE;
 
+/* FIXME: Should it be configable? */
+#define JANUS_RABBITMQ_EXCHANGE_TYPE "fanout"
+
 /* JSON serialization options */
 static size_t json_format = JSON_INDENT(3) | JSON_PRESERVE_ORDER;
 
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks for this contribution! Unfortunately I'm not familiar enough with RabbitMQ in general, but it's nice to see this opens the door to more than a single one-to-one relationship with applications servers. Can you explain a bit what this allows? Multiple AS connecting to a single Janus instance? Have you tested this too?

I'll try to make a review ASAP, but I'll be a bit busy this week, and the next I'll go abroad for the IETF. Could you also advertise this on the [meetecho-janus](https://groups.google.com/forum/#!forum/meetecho-janus) group so that more people are aware of this and can help testing?

Thanks again!

pitkonst:
Adding exchange makes possible to pass mesages from Janus to multiple queues. In this case AS sends message to Janus through input queue `to-janus`, creates additional queue `from-janus-[server-id]` for answers and when answer is ready all active servers recieves it through own `from-janus-*` queue. In this case server mast filter answers, based on transaction field. Event packets should be filtered too, based on handler id. But it is better use RabbitMQ event handler for events.

We tested it in multi AS - multi Janus setting (2 AS - 2 Janus), whitch is needed for scalability reasons.

lminiero:
Ah, ok, so an AS will also receive events related to stuff that another AS originated. I get your point about filtering on the transaction, but that's not always there, e.g., for most asynchronous notifications coming from plugins. I guess in that case the filtering must be more sophisticated, and you'd have to rely on session/handle identifiers for that (so that you can ignore events that come from couples you know you're not handling yourself).

transports/janus_rabbitmq.c
```diff
@@ -107,6 +107,9 @@ static gboolean rmq_janus_api_enabled = FALSE;
 static gboolean rmq_admin_api_enabled = FALSE;
 static gboolean notify_events = TRUE;
 
+/* FIXME: Should it be configable? */
```
lminiero: Typo: configurable

PR #1045, opened 2017-10-30
RabbitMQ event handler

pitkonst:
RabbitMQ event handler plugin. Makes sens when using with RabbitMQ transport. Basiclly this is combination of RabbitMQ transport and Simple event handler.

### DIFF
--- a/conf/janus.eventhandler.rabbitmqevh.cfg.sample
+++ b/conf/janus.eventhandler.rabbitmqevh.cfg.sample
@@ -1,0 +1,23 @@
+; This configures the RabbitMQ event handler.
+
+[general]
+enabled = no				; By default the module is not enabled
+events = all				; Comma separated list of the events mask you're interested
+							; in. Valid values are none, sessions, handles, jsep, webrtc,
+							; media, plugins, transports, core and all. By default we
+							; subscribe to everything (all)
+host = 192.168.99.100		; The address of the RabbitMQ server
+;port = 5672				; The port of the RabbitMQ server (5672 by default)
+;username = guest			; Username to use to authenticate, if needed
+;password = guest			; Password to use to authenticate, if needed
+;vhost = /					; Virtual host to specify when logging in, if needed
+exchange = janus-exchange
+route_key = janus-events	; Name of the queue for event messages
+
+;ssl_enable = no			; Whether ssl support must be enabled
+;ssl_verify_peer = yes		; Whether peer verification must be enabled
+;ssl_verify_hostname = yes	; Whether hostname verification must be enabled
+; certificates to use when SSL support is enabled, if needed
+;ssl_cacert = /path/to/cacert.pem
+;ssl_cert = /path/to/cert.pem
+;ssl_key = /path/to/key.pem
--- a/conf/janus.eventhandler.rabbitmqevh.cfg.sample
+++ b/conf/janus.eventhandler.rabbitmqevh.cfg.sample
@@ -6,6 +6,9 @@ events = all				; Comma separated list of the events mask you're interested
 							; in. Valid values are none, sessions, handles, jsep, webrtc,
 							; media, plugins, transports, core and all. By default we
 							; subscribe to everything (all)
+grouping = yes				; Whether events should be sent individually , or if it's ok
+							; to group them. The default is 'yes' to limit the number of
+							; messages
 host = 192.168.99.100		; The address of the RabbitMQ server
 ;port = 5672				; The port of the RabbitMQ server (5672 by default)
 ;username = guest			; Username to use to authenticate, if needed
--- a/conf/janus.eventhandler.rabbitmqevh.cfg.sample
+++ b/conf/janus.eventhandler.rabbitmqevh.cfg.sample
@@ -9,12 +9,12 @@ events = all				; Comma separated list of the events mask you're interested
 grouping = yes				; Whether events should be sent individually , or if it's ok
 							; to group them. The default is 'yes' to limit the number of
 							; messages
-host = 192.168.99.100		; The address of the RabbitMQ server
+host = localhost			; The address of the RabbitMQ server
 ;port = 5672				; The port of the RabbitMQ server (5672 by default)
 ;username = guest			; Username to use to authenticate, if needed
 ;password = guest			; Password to use to authenticate, if needed
 ;vhost = /					; Virtual host to specify when logging in, if needed
-exchange = janus-exchange
+;exchange = janus-exchange
 route_key = janus-events	; Name of the queue for event messages
 
 ;ssl_enable = no			; Whether ssl support must be enabled
--- a/configure.ac
+++ b/configure.ac
@@ -803,8 +803,8 @@ AM_COND_IF([ENABLE_SAMPLEEVH],
 	[echo "    Sample event handler:  yes"],
 	[echo "    Sample event handler:  no"])
 AM_COND_IF([ENABLE_RABBITMQEVH],
-  [echo "    RabbitMQ event handler:yes"],
-  [echo "    RabbitMQ event handler:no"])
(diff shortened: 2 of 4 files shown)


lminiero:
This is very exciting, thanks! The `\author` in the doxygen preamble should really have your name, though :slightly_smiling_face: 
As for your other PR, I'll try and review as soon as I have some time.

configure.ac
```diff
@@ -784,6 +802,9 @@ echo "Event handlers:"
 AM_COND_IF([ENABLE_SAMPLEEVH],
 	[echo "    Sample event handler:  yes"],
 	[echo "    Sample event handler:  no"])
+AM_COND_IF([ENABLE_RABBITMQEVH],
+  [echo "    RabbitMQ event handler:yes"],
+  [echo "    RabbitMQ event handler:no"])
```
lminiero: The indentation for these two echos seems wrong, maybe it's spaces instead of a tab?

configure.ac
```diff
@@ -804,4 +825,4 @@ AM_COND_IF([ENABLE_JAVASCRIPT_MODULES], [
 
 echo
 echo "If this configuration is ok for you, do a 'make' to start building Janus. A 'make install' will install Janus and its plugins to the specified prefix. Finally, a 'make configs' will install some sample configuration files too (something you'll only want to do the first time, though)."
-echo
+echo
```
lminiero: Any reason why you removed the newline? Something your IDE did automatically?

events/janus_rabbitmqevh.c
```diff
@@ -0,0 +1,724 @@
+/*! \file   janus_rabbitmqevh.c
+ * \author Lorenzo Miniero <lorenzo@meetecho.com>
```
lminiero: Should be:

    \author Piter Konstantinov <your mail>

as it's really your plugin!

events/janus_rabbitmqevh.c
```diff
@@ -0,0 +1,724 @@
...
+static void janus_rabbitmqevh_event_free(json_t *event) {
+	if(!event || event == &exit_event)
+		return;
+	json_decref(event);
+}
+
+/* JSON serialization options */
+static size_t json_format = JSON_INDENT(3) | JSON_PRESERVE_ORDER;
+
+/* FIXME: Should it be configable? */
```
lminiero: Same typo... damn copy/paste :wink:

events/janus_rabbitmqevh.c
```diff
@@ -0,0 +1,724 @@
...
+	item = janus_config_get_item_drilldown(config, "general", "password");
+	if(item && item->value)
+		password = g_strdup(item->value);
+	else
+		password = g_strdup("guest");
+
+	/* SSL config*/
+	item = janus_config_get_item_drilldown(config, "general", "ssl_enable");
+	if(!item || !item->value || !janus_is_true(item->value)) {
+		JANUS_LOG(LOG_INFO, "RabbitMQ SSL support disabled\n");
```
lminiero: Long lines should clarify that we're referring to the RabbitMQ *event handler*, otherwise this might be confused with long lines from the RabbitMQ transport instead. Spotted this but there may be others.

events/janus_rabbitmqevh.c
```diff
@@ -0,0 +1,724 @@
...
+		JANUS_LOG(LOG_FATAL, "Can't connect to RabbitMQ server: error declaring queue... %s, %s\n", amqp_error_string2(result.library_error), amqp_method_name(result.reply.id));
+		goto error;
+	}
+
+	/* Initialize the events queue */
+	events = g_async_queue_new_full((GDestroyNotify) janus_rabbitmqevh_event_free);
+	g_atomic_int_set(&initialized, 1);
+
+	GError *error = NULL;
+	handler_thread = g_thread_try_new("janus sampleevh handler", janus_rabbitmqevh_handler, NULL, &error);
```
lminiero: sed s/sampleevh/rabbitmqevh

events/janus_rabbitmqevh.c
```diff
@@ -0,0 +1,724 @@
...
+	if(error != NULL) {
+		g_atomic_int_set(&initialized, 0);
+		JANUS_LOG(LOG_FATAL, "Got error %d (%s) trying to launch the SampleEventHandler handler thread...\n", error->code, error->message ? error->message : "??");
```
lminiero: SampleEventHandler should be RabbitMQEventHandler.

events/janus_rabbitmqevh.c
```diff
@@ -0,0 +1,724 @@
...
+			int status = amqp_basic_publish(rmq_conn, rmq_channel, rmq_exchange, rmq_route_key, 0, 0, &props, message);
+			if(status != AMQP_STATUS_OK) {
+				JANUS_LOG(LOG_ERR, "Error publishing... %d, %s\n", status, amqp_error_string2(status));
+			}
+			free(event_text);
+			event_text = NULL;
+		}
+
+		/* Cleanup */
+		/* Done, let's unref the event */
```
lminiero: I guess that one of the two comments is redundant.

events/janus_rabbitmqevh.c
```diff
@@ -0,0 +1,724 @@
...
+						  "status": "started"
+					   }
+					}
+				*/
+				break;
+			default:
+				JANUS_LOG(LOG_WARN, "Unknown type of event '%d'\n", type);
+				break;
+		}
+
```
lminiero: No grouping management as done in the Sample event handler? It might greatly reduce the protocol overhead, when the queue contains more than one pocket as you read from it. The same code we have in the other handler should be re-usable exactly as it is.

pitkonst:
This is massive review. Thanks.
I thought grouping may add latency to events, but looked through code one more time and this realization is fine.

lminiero:
Grouping will actually reduce delay in this case, as if you don't group and have multiple events ready when the queue wakes up, you won't send the second available event right away (which you could), but only after you sent the first one; same thing for the last event in th queue, who will suffer from even more delay. It basically tries and make advantage of the fact that the queue will not immediately wake up for each new event due to concurrency.

Thanks for fixing everything so quickly! I'll give it a go shortly, and in case it works for me and I don't spot anything that might need fixing, I'll merge right away :+1:

conf/janus.eventhandler.rabbitmqevh.cfg.sample
```diff
@@ -0,0 +1,26 @@
...
+[general]
+enabled = no				; By default the module is not enabled
+events = all				; Comma separated list of the events mask you're interested
+							; in. Valid values are none, sessions, handles, jsep, webrtc,
+							; media, plugins, transports, core and all. By default we
+							; subscribe to everything (all)
+grouping = yes				; Whether events should be sent individually , or if it's ok
+							; to group them. The default is 'yes' to limit the number of
+							; messages
+host = 192.168.99.100		; The address of the RabbitMQ server
```
lminiero: Please put `localhost` here.

conf/janus.eventhandler.rabbitmqevh.cfg.sample
```diff
@@ -0,0 +1,26 @@
...
+;port = 5672				; The port of the RabbitMQ server (5672 by default)
+;username = guest			; Username to use to authenticate, if needed
+;password = guest			; Password to use to authenticate, if needed
+;vhost = /					; Virtual host to specify when logging in, if needed
+exchange = janus-exchange
```
lminiero: The `exchange` property should also be commented out by default, I guess?

lminiero:
Tested and works beautifully, amazing job! As soon as you can fix those two small nits in the sample configuration file, I'll merge :+1:

pitkonst:
I've overwatched this two. Thanks.

lminiero:
Thanks, merging!

PR #1047, opened 2017-10-31
Link quality estimation

pitkonst:
General goal of this PR is to have a metric of client’s internet connection quality to inform him and other participants. For this two metrics are added to handler stats: link quality and media link quality. Both metrics are calculated separately for uplink and downlink. Link quality estimation based on raw lost in channel. Link quality of 100 means no losses, 90 means 10% of packages were lost no matter were they restored with NACK or not. Media link quality is based on RR lost counter, it’s good for basic estimations of media quality (in terms of glitches and lags) for client.

Details

Client to Janus link quality is calculated as:
`(count_of_received_packets – count_of_retransmitted_packages) / expected_number_of_packages`
Received packages are already counted for sake of SR generation. But not quite correct. 
From RFC 1889: 
> cumulative number of packets lost: The total number of RTP data packets from source SSRC_n that have been lost since the beginning of reception. This number is defined to be the number of packets expected less the number of packets actually received, where the number of packets received includes any which are late or duplicates. Thus packets that arrive late are not counted as lost, and the loss may be negative if there are duplicates. The number of packets expected is defined to be the extended last sequence number received, as defined next, less the initial sequence number received. This may be calculated as shown in Appendix A.3.

To correct this [rtcp.c l466](https://github.com/pitkonst/janus-gateway/commit/a9a522fb5612d07ac3159fadf8f9239300b9cb2f#diff-c40fb57535974b0488f21216654ca253L466) moved up to count all received packages. Additionally retransmitted packages are counted at [rtcp.c l502](https://github.com/pitkonst/janus-gateway/commit/a9a522fb5612d07ac3159fadf8f9239300b9cb2f#diff-c40fb57535974b0488f21216654ca253R502).

Janus to client Janus link quality is calculated as:
`count_of_receiver_NACKs / count_of_packets_sent_to_receiver`
Sent packets are counted at [ice.c l3941](https://github.com/pitkonst/janus-gateway/commit/a9a522fb5612d07ac3159fadf8f9239300b9cb2f#diff-724b612068e82ba0ad6ed24da3a22e47R3941) individual for every stream. NACKs count is tricky. When we see NACK from receiver we must filter those for packet lost on uplink (client to Janus) path and those sent multiple times due to channel lag. For first one we check packer persistence in retransmit buffer. For second we use NACK resend check. Count done at [ice.c l2368](https://github.com/pitkonst/janus-gateway/commit/a9a522fb5612d07ac3159fadf8f9239300b9cb2f#diff-724b612068e82ba0ad6ed24da3a22e47R2368)

Media link qualities are calculated as:
`number_of_lost_packets / expected_number_of_packets`
where loss calculated based on SR and RR values.

Estimation filtering done for every estimated value at [rtcp.c l138](https://github.com/pitkonst/janus-gateway/commit/a9a522fb5612d07ac3159fadf8f9239300b9cb2f#diff-c40fb57535974b0488f21216654ca253R138). Filtering adds lag to estimation, but smoothed values showed accuracy ±1 at syntactic tests. Still necessity of filtering in general case is topic under discussion.

Testing

PR is in use for few weeks on our service and works quite well for connection problems detection. We made some syntactic test with different lose rates and lags and satisfied with accuracy.  Still it needs heavy testing for broad variety of use cases. 

Known problems

Channel lag more than 100ms affects Janus to client Janus link quality measurements because of false positive NACKs detection. We track root of this problem, but do not know how to solve it without heave changes in core code.

For now, there is no way to inform client about connection problem because in our case it is done on sever side. This PR do not replace current slow link mechanism. Comments and suggestions needed for this one.


lminiero:
Thanks for this effort! This is a really interesting improvement, although I'm wondering why you're taking NACKs into account at all, when you might just look at packets and losses. The fact we're currently relying on NACKs as a trigger for slowlink events, for instance, is actually not that reliable and I'd like to change that sooner or later (again, by figuring out lost packets in both directions instead). Unfortunately I will be awfully busy in the next few weeks (I'll be abroad ~20 days starting on the 5th) so there won't be much I can do myself to help, but I'm definitely interested in continuing discussing this effort.

pitkonst:
First version of this method used packet losses info as you said. And it worked quite well, without NACK. After NACK was fixed there was accuracy degradation. Test showed that Chrome calculates losses as difference of expected number of packets and received number of packets and count all NACKed packets as well. For example, in case of 10% packets drop rate link loses could be 0 because of NACK. I do not like idea of using on NACK as lost detection mechanism, but I do not really know more accurate way.

lminiero:
The reason I'm hesitant to use NACKs is that they might be disabled in a session. We already negotiate them dynamically, by checking if they're advertised in the SDP. This always results in NACKs being disabled for audio, currently (apparently the nack rtcp-fb attribute is never added for audio by browsers), while they may or may not be enabled for video depending on what the client sends in their offer/answer.

pitkonst:
I get it. Good point. For this cases values referred as media link quality could be used. As an alternative it is possible to switch between calculation algorithms for streams with NACK and without.

trankkilou:
Hi pitkonst

by 
count_of_received_packets – count_of_retransmitted_packages) / expected_number_of_packages
you mean
count_of_received_packets – count_of_retransmitted_packets) / expected_number_of_packets

or packets != packages ?

pitkonst:
Sorry for inaccuracy, packets == packages in this context.

trankkilou:
Hello Piter,

No problem it was just to be sure I have well understood.

It is a really good idea you had.

Did you make some test to know when NACK mechanism is not enough to balance packets losses ?

I guess @ link quality = 99 , I mean 1% packets loss, you should start having some issues and @95 NACK mechanism shouldn’t be enough :)

JJ-

pitkonst:
We did those tests. It occures loss rate, when video problems begines, dependes on RTT and direction. For example with Chrome as a client, video from Janus to client tolerate 5%-8% loss while, client to Janus video starts to skip frames with 2% loss. In both cases link detection mechanism allows to detect link problems before video issues comes up. This allows to warn clients about connection problems.

marcuspen:
Hi @pitkonst  and @lminiero !
We've been testing this branch out here at [Sohonet](https://www.sohonet.com/) as one of our applications uses Janus.
This work is very good and the new stats that are provided are quite useful. 
What's the status of this PR (other than the conflicts) and do you reckon it will be merged soon?

lminiero:
I haven't played with it yet, so this would need to be aligned first, for me to give an answer. Anyway, as I said in some previous responses too, this should use RR losses feedback and not NACKs, which as I explained cannot be relied upon (they're not used for audio, and may be disabled altogether). I'm planning to start using those for slowlink events too, so I wouldn't consider merging something that only uses NACKs.

marcuspen:
Cheers for the response @lminiero !
Fair enough. That makes perfect sense. I'm just having a look at the handle stats and I can't seem to see the RTCP report losses stat information or the total amount of packets sent since the last RTCP Report... just the `lost-by-remote` for audio and video.

Do we have these available to us in the handle stats? If so, then I can always extract the data from there and do the calculations until this PR is updated.

Also, could you confirm what the value `lost-by-remote` is in reference to? As in, is it packets or bytes? I only ask because I'm looking at the stats I have now, and my video `lost-by-remote` is much higher than my `out_stats.video_packets` which makes me think that `lost-by-remote` could be in bytes`...

lminiero:
We have both "lost" (Janus didn't receive some packets from peer) and "lost-by-remote" (peer didn't receive some packets from Janus) in the Admin API and events: https://github.com/meetecho/janus-gateway/blob/master/janus.c#L2355
The former we calculate ourselves, the latter is sent by the peer in RR messages. Both just count packets, not bytes.

pitkonst:
@marcuspen, @lminiero thanks for your comments on this PR.

I want to mention that there are two type of stats in this implementation: *-media-link-quality uses only RR and SR data, *-link-quality uses NACK counts. I do understand that using NACK data is not reliable in all cases, but we have to use it, because measurements made on base on RR only are not accurate enough for practical proposes.  
I'd love to see this PR merged, as we could use latest version of Janus in this case.
I suggest to implement switching so that correct type of measurements to be used in both cases (RR and SR for streams without NACK and NACKbased in case of presence of NACK). Would it be enough for consider merging this PR, or using NACK data is unacceptable in any case? Are there other problems with this PR?

lminiero:
@pitkonst thanks for the clarification! I guess the names of the properties are a bit misleading then, as I didn't understand the difference just by looking at them: I don't have any suggestion, though, so it's just for the sake of discussion.

If both pieces of information are always made available (that is, you can always choose which one to use when receiving data) then it's fine for me. I'll definitely have a look when it's aligned to master :+1:

pitkonst:
Thanks. The names are not quite good, and suggestions are always welcomed. I'll aling this with master ASAP.

marcuspen:
@pitkonst Is it possible to make the quality numbers a decimal? So that we can be more granular and say link quality is 99.5 for example?

pitkonst:
@marcuspen It is possible, but with limitations. Raw measurements are a bit noisy, and with basic filtering on 3 point it shows accuracy of ±1% (if you get lose of 98% it could actually be 97..99%). Inaccurate measurements but with almost no lag. We use this to warn users about connection problems and lag is critical in our case. It's possible to provide heavily filtered values with accuracy of 0.1% but with huge lag. This could be good for long time stats.

lminiero:
> This could be good for long time stats

I wouldn't worry about those here, that's what event handlers (and external processing) are for. Much more interesting to have something that can be used in real-time, here.

pitkonst:
PR is aligned now. I'm no sure i've done every thing right with github. Made some changes and fixes, works fine in tests.

lminiero:
@pitkonst thanks, I'll check that out as soon as I have some time!
The notification mentioned 250 commits now, though... any change you can rebase this, or re-publish as a new PR, so that when we merge we have a cleaner commit?

lminiero:
In fact I see a ton of unrelated changes in the diff... as if this included all the stuff we added recently, like renegotiations, etc.

pitkonst:
I'll better make new PR

Issue #1048, opened 2017-11-02
MQTT transport doesn't seem to support TLS connections to broker


oej:
It seems like the MQTT transport doesn't support connecting to a TLS enabled broker. We can connect to a TCP connection just fine, but with TLS enabled janus fails to connect.

We add a URL like "SSL://server.domain.tld:8883" and it fails miserably. I think the code may need SSL_options and some other stuff in order to support this kind of connection.

Solution: Either document that it's not supported, or document how it's support and fix the code :-)

This may be a feature request, but if it is, the bug is the documentation.

lminiero:
You're probably right: the MQTT was contributed so I'm less familiar with it than I am with the other plugins, but it makes sense that SSL options may be needed, as is the case for the RabbitMQ transport. I'll have a look at the library API as soon as I have some time.

lminiero:
Pinging @manifest, the MQTT transport plugin author, just in case he can nudge us in the right direction.

oej:
As always I would prefer to use TLS. And as always, I don't like that the URL starts with "SSL:" but that is a problem in the MQTT library I need to report elsewhere :-)

lminiero:
@oej can you check the above PR? I implemented it by simply looking at the docs but don't have time to test it unfortunately. Please let me know if that fixes it for you: there are new properties in the config to enable SSL for the connection.

lminiero:
Ping! :slightly_smiling_face:

manifest:
Hi there! In our case Janus communicates with MQTT broker in a safe internal network. In that case using SSL looks kind of overkill. For most cases that should be this way, I believe. From the first sight PR looks ok, I could check it more carefully later.

lminiero:
@manifest thx!

oej:
Wakeup. Will check.

@manifest "Safe internal network" is usually a bad assumption. Being able to run TLS makes it possible to do even more things.

manifest:
It required some more changes, so I've made another [PR](https://github.com/meetecho/janus-gateway/pull/1064). Note that to support ssl/tls we need to link `paho-mqtt3as` library. Should not be a big deal :-)

manifest:
There is [gist](https://gist.github.com/manifest/aa2cfb19b596669b710bbc0f8885d64f). It coud be used to set up development environment, build and start playing with the plugin.

lminiero:
[...]

PR #1049, opened 2017-11-02
SSL support in MQTT plugin

lminiero:
See #1048.

### DIFF
--- a/conf/janus.transport.mqtt.cfg.sample
+++ b/conf/janus.transport.mqtt.cfg.sample
@@ -4,7 +4,9 @@ enable = no							; Whether the support must be enabled
 json = indented						; Whether the JSON messages should be indented (default),
 									; plain (no indentation) or compact (no indentation and no spaces)
 
-url = tcp://localhost:1883			; The connection URL of the MQTT broker
+url = tcp://localhost:1883			; The connection URL of the MQTT broker: if you want
+									; to use SSL, make sure you type ssl:// instead of tcp://,
+									; and that you configure the SSL settings below
 ;client_id = guest					; Client identifier
 ;username = guest					; Username to use to authenticate, if needed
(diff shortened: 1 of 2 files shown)


oej:
Great hack and a very good start!

A few comment from reading the code:

1. "SSL" should in most text be "TLS". All versions of the very old SSL protocol are insecure today. We should not promote it. I would like to be able to use "mqtts" as a URL schema as in the standard, even though the library does not support it... ("ssl:" should be hidden...4ever)

2. To connect using TLS to a broker, only CACERT is needed. If the CA cert used is in the O/S CA bundle, it should not be needed
   - needs testing

3. In some cases a client cert is used, if so both client cert and key is needed. These should not be required.

4. One wants a setting to determine version of TLS and ciphers used, but that may come later.

Will hopefully be able to test a few things by the end of the week. Thank you!

lminiero:
The `ssl_` stuff you see are just prefixes for the config names and variables, and lazyness on my side as I just copied the exact same code we already had for the RabbitMQ transport. It's just meant as "secure context", no matter how it's then implemented by the library itself. The MQTT library itself refers to the secure context as `MQTTAsync_SSLOptions`, so we're basically just aligning to that.

Apart from this, I don't really know how the library works or what it requires. I think we don't mandate the presence of anything, when parsing the config, which means you should be able to only pass a subset of the info to the library, and then it's up to that to tell you whether it's enough or not.

As to the ciphers, it looks like the library [allows you](https://www.eclipse.org/paho/files/mqttdoc/MQTTAsync/html/struct_m_q_t_t_async___s_s_l_options.html#aa683926d52134077f27d6dc67bda13ab) to specify them when initializing, which means in the future it should be easy to add a related option in the settings.

Thanks for testing, please let me know if/when you believe it's good enough to merge as it is!

lminiero:
Closing as #1064 does this better.

PR #1054, opened 2017-11-07
options.error is supposed to return two strings

dasilva333:
all the calls to httpAPICall functions contain an error handler that returns two strings, I fixed options.error in once instance where it was returning a string and an object so that the error handler doesn't see [object object] in the final error string

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -96,7 +96,7 @@ Janus.useDefaultDependencies = function (deps) {
 				}
 			}).catch(function(error) {
 				if(typeof(options.error) === typeof(Janus.noop)) {
-					options.error(error.message || '<< internal error >>', error);
+					options.error(error.message || '<< internal error >>', error && error.error ? error.error.message : error);
 				}
 			});
 


lminiero:
Thanks for this contribution! I'll be abroad for a while so I won't be able to quickly review this: just pinging @cmacq2 to be sure that was the intended behaviour in the meanwhile.

lminiero:
@dasilva333 just checked and looks good, thanks! Still waiting for the CLA to be signed before I can merge though: please let me know when that's done and I'll go through :+1:

cmacq2:
The previous version was deliberate, to signal to client code what the actual error was.
See also the discussion here: https://github.com/meetecho/janus-gateway/pull/991 

The basic problem is, how can the client code respond appropriately to errors which range from:

- A HTTP 4xx error? (I.e. your request was bad...)
- A HTTP 5xx error? (I.e. the server is not feeling well today...)
- A timeout?
- Failure to parse JSON returned by Janus? I.e. a bug in Janus sending you broken JSON?

A simple, plain string 'message' is not great for signalling errors to code, and also the message is likely inappropriate if you have a need for proper i18n/l10n.

cmacq2:
That said, there is general 'merit' in revisiting error signalling in janus.js to make it all more consistent and update the demo pages accordingly. That would certainly help to improve code quality a lot from the perspective of using `janus.js` as a library.

lminiero:
@dasilva333 @cmacq2 what do you think should happen with the PR then? I'm ok with both leaving it as it is, and merging, depending on what you guys would be more comfortable with. New approaches would have to wait until I'm back, probably, as I'll be abroad for about 10 more days.

cmacq2:
@lminiero I would leave it as is. @dasilva333 are you interested in drafting a new PR which provides a comprehensive overhaul of error signalling within `janus.js`?

lminiero:
Finally back from the long tour abroad, so apologies for the slow feedback... closing the PR then, we can still use it to discuss refactoring error management if you want.

PR #1055, opened 2017-11-07
restrict STUN to IPv4 as the whole code around is not ready for IPv6

mzanetti:
I have tried to make this dependent on janus_ipv6_enabled and then later replace the inet_addr() with inet_pton(), but I ran into more issues with the STUN code and ipv6 support so I decided it's probably best to fix this by disabling the address resolution for v6 until someone familiar with the code can look at adding proper v6 support.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -750,10 +750,13 @@ int janus_ice_set_stun_server(gchar *stun_server, uint16_t stun_port) {
 		stun_port = 3478;
 	JANUS_LOG(LOG_INFO, "STUN server to use: %s:%u\n", stun_server, stun_port);
 	/* Resolve address to get an IP */
+	struct addrinfo hints;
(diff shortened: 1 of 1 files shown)


lminiero:
Currently abroad for some time, so I won't be able to quickly review this. Anyway, we shouldn't need any specific family related code, as that's what `ip-utils.c` should help on. Not exactly sure as to whether or not we were already using the helpers there, but I think that would be the way to go. Not sure if just dropping the check would be good, as that check also gathers what Janus considers to be the public IP, which is used in a bunch of things (e.g., SDP rewriting).

lminiero:

lminiero:
Closing as we're trying to do it properly in #1062 instead. Thanks!

mzanetti:
Nice, thanks a lot!

PR #1057, opened 2017-11-08
JANUSSID Cookie

cescp:
When using Janus in several instances behind a load balancer, all requests from a given client must go to the same instance, in order to avoid problems: if a client requests (POST) a session id from a Janus in instance A, but next request (GET with session_id), goes to instance B, it will get a response of non existing session.
The classic load balancer from AWS can stick the assignment of a given instance to a client, provided that a cookie is set. See: Configure Sticky Sessions for Your Classic Load Balancer (http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-sticky-sessions.html#enable-sticky-sessions-duration)
This pull request implements the following behaviour:
- at the beginning of the Janus session it sets the cookie JANUSSID
- at the end of the Janus session it unsets the cookie JANUSSID

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -1048,7 +1048,7 @@ void janus_http_session_created(void *transport, guint64 session_id) {
 
 	/* create cookie */
 	g_snprintf(cookie_value, sizeof(cookie_value), "%"G_GUINT64_FORMAT, (guint64)session_id);
-	cookie_max_age = 86400;	
+	cookie_max_age = 86400;
 }
 
 void janus_http_session_over(void *transport, guint64 session_id, gboolean timeout) {
@@ -1072,7 +1072,7 @@ void janus_http_session_over(void *transport, guint64 session_id, gboolean timeo
 
 	/* remove cookie */
 	g_snprintf(cookie_value, sizeof(cookie_value), "%"G_GUINT64_FORMAT, (guint64)session_id);
-	cookie_max_age = -1;	
+	cookie_max_age = -1;
 }
 
 /* Connection notifiers */
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -173,6 +173,10 @@ static char *admin_ws_path = NULL;
 /* Custom Access-Control-Allow-Origin value, if specified */
 static char *allow_origin = NULL;
 
+/* cookies */
+static char cookie_value[500];
+static int cookie_max_age = 0;
+
 /* REST and Admin/Monitor ACL list */
 GList *janus_http_access_list = NULL, *janus_http_admin_access_list = NULL;
 janus_mutex access_list_mutex;
@@ -1041,6 +1045,10 @@ void janus_http_session_created(void *transport, guint64 session_id) {
 	session->destroyed = 0;
 	g_hash_table_insert(sessions, janus_uint64_dup(session_id), session);
 	janus_mutex_unlock(&sessions_mutex);
+
+	/* create cookie */
+	g_snprintf(cookie_value, sizeof(cookie_value), "%"G_GUINT64_FORMAT, (guint64)session_id);
+	cookie_max_age = 86400;	
 }
 
 void janus_http_session_over(void *transport, guint64 session_id, gboolean timeout) {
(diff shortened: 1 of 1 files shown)


lminiero:
That looks helpful indeed, thanks for adding the feature! I'll be abroad for some time, so I won't be able to quickly review this, but hopefully I'll find some time to do that soon. Something I can already ask is to please capitalize the first letter in comments, e.g.:

    /* Create cookie */

as that's how our comments are styled in the code.

lminiero:
@cescp just had a look at the code but I don't think a static, and shared, value for the cookie works. First of all, IIUC it only works with a single session: as soon as a second session is created, whether the first one is done or not the cookie string is overwritten. Besides, this might easily result in concurrent access to that string from different threads, with possibly unpredictable results. Or am I misunderstanding something in your patch?

Thanks!

lminiero:
Ping :slightly_smiling_face:

cescp:
Sorry, I didnt't see your message.

I see your point. You are right: there could be conflicts if several
clients that share the cookie jar (e.g. from two tabs in the same
browser) request several Janus sessions.
As the server of the cookie is always the same (load balancer name)
and the name of the cookie has to be constant (parameter when
configuring the load balancer), the only thing that we can change is
the cookie value.

I propose the following solution:

- let's assume we have a load balancer (lb.example.org), configured
with cookie JANUSSERVER as sticky cookie, with two instances behind it
(with ip addresses: 1.2.3.4 and 5.6.7.8)

- when a client from browser A, tab 1 requests a janus session to
lb.example.org, the load balancer redirects the request to
e.g. 1.2.3.4. The Janus in this instance sets a cookie with name
JANUSSERVER and the own ip address as value: 

`lb.example.org JANUSSERVER 1.2.3.4`

- subsequent requests to the same janus session from browser A and tab
  1 will send this cookie to load balancer, and requests will be
  redirected to 1.2.3.4

- if a new janus session from browser A, but tab 2 is requested, as
  both tabs share the cookie jar, the cookie will be sent, load
  balancer will redirect the request to 1.2.3.4 and a new janus
  session will start. Again, the cookie will be set, but the value
  will be the same as before:

`lb.example.org JANUSSERVER 1.2.3.4`

  There would be no problem on overwriting it, as it has the same value
  as before.

- when janus session from browser A and tab 1 is closed, the cookie
  should not be unset. Instead, cookie would have a limited duration
  (e.g. one day).

- if a client from browser B, tab 1 requests a new janus session, another
  cookie will be set, but as it will be stored on a different cookie jar,
  there won't be any conflict.

With this behaviour, all requests from a given browser A (even from
different tabs) will go to the same instance. This is not optimal in
terms of balancing, but at least we are sure that all requests for a
given janus session go to the same Janus server, as we wanted.

On the contrary, requests from a different browser or a different
computer will be balanced among the instances, as they don't share the
cookies.

What do you think?

lminiero:
> With this behaviour, all requests from a given browser A (even from
> different tabs) will go to the same instance. This is not optimal in
> terms of balancing, but at least we are sure that all requests for a
> given janus session go to the same Janus server, as we wanted.

Actually, this is kind of an issue.

One approach to scale a Janus application is distributing the application-level context over Janus-level properties. To make a simple example, a simple way to scale a VideoRoom based conference is allocating different publishers on different Janus instances: you have room 1234, Bob publishes on Janus n.1 (where a 1234 room exists), Alice publishes on Janus n.2 (where a different 1234 room exists); all people interested in watching Bob will start a subscriber on Janus n.1, those interested in Alice will subscribe on Janus n.2. The two Janus instances know nothing of each other, and the two rooms are completely separated, but from an application perspective you can present them as the same room to the end user, transparently or not, and thus scale a Janus-based conference much more than you could with a single instance.

The approach you suggested would prevent that, as all users would end up in the same place. I don't have a solution or a clear idea on how this could be handled better, though. Can you open a discussion on our [Google group](https://groups.google.com/forum/#!forum/meetecho-janus) linking to your PR, so that other people interested in load balancing as you are can chime in and provide feedback?

Thanks!

lminiero:
Any feedback on the points above?

lminiero:
Without a clear solution to this issue I raised, I'm afraid we'll have to close this pull request. Please do propose a new PR in the future, if you believe you found one. As a side note, my guess is that your specific requirement might actually be addressed outside of Janus, e.g., as part of whatever frontend you place in front of it HTTP wise (e.g., nginx or httpd): AFAIK you can set cookies there too, so a frontend responsible for a specific instance could bridge that gap you're missing.

Thanks for the contribution!

Issue #732, opened 2017-01-05
WebSocket transport dead lock


vecio:
I'm using Janus video room plugin to develop a new website, and have found this issue for several times. e.g. the WebSocket won't accept any new client after some tests.

After tracing the code execution, I found the code block at 

```c
JANUS_LOG(LOG_VERB, "[%s-%p] WebSocket connection accepted\n", log_prefix, wsi);
if(ws_client == NULL) {
	JANUS_LOG(LOG_ERR, "[%s-%p] Invalid WebSocket client instance...\n", log_prefix, wsi);
	return -1;
}
/* Clean the old sessions list, in case this pointer was used before */
janus_mutex_lock(&old_wss_mutex);
```

I have some old clients exited without cleanup, e.g. no detach handle, etc.

Is this expected to happen when clients don't cleanup resources before disconnect?

lminiero:
PS: if you find it easy enough to replicate, you can try enabling the locking debug in Janus and try again until it happens. This way, you should be able to see exactly where in the code the double locking happens for you.

vecio:
Thanks. I just enabled the lock_debug, and will wait for this to happen again.

BTW: when will the new refcount branch will be considered stable?

lminiero:
As soon as it is! :wink: 
Jokes apart, there are people testing this and we try to do that whenever we can too. If you're interested in giving it a go in some of your application scenarios, it will definitely help move it forward faster!

vecio:
Will try it soon. And I just noticed another issue, if I start a WebSocket connection, close it, then open another connection soon, a timeout message will be sent to the new connection with old session id.

I'm not sure if this is a bug, because I can deal with it in client code.

lminiero:
Yeah, that's something we experimented too, and happens when a new connection arrives as soon as the previous one ends and the two share the same "pointer". In that case, Janus might be a little to slow to notify the timeout, use the pointer it had, and so the event will get to the new connection instead of the old one. Again, something that with the reference counter branch won't happen.

To solve that issue in the short term, we might simply add a map in the transport, so that we always feed the core with a different pointer no matter what it points to, as it would be mapped to the actual transport structure.

lminiero:
@vecio any update on the lok debugging?

vecio:
Not occurred any more, have been running the service for one month.

lminiero:
Ok, closing the issue then. Feel free to reopen should this reappear.

shrhoads:
[...]

PR #1060, opened 2017-11-10
LWS_MAX_SMP=1 in README LWS build instructions

mquander:
As per the conclusion in https://github.com/meetecho/janus-gateway/issues/732.

(As of the time of this writing I still experienced the deadlocking issue in Janus master.)


lminiero:
Good point, it should be made clearer. Merging :+1:

PR #1062, opened 2017-11-11
Fixed STUN check at startup in IPv6 network (fixes #1053)

### LINKED ISSUE #1053: STUN fails in IPv6 enabled networks
We have janus configured with this STUN server: numb.viagenie.ca

Starting Janus with this config in IPv4 only network works fine, but starting it in IPv6 enabled networks fails with this log: 
```
STUN server to use: numb.viagenie.ca:3478
  >> 2600:3c03::f03c:91ff:fe93:39a4:3478
[FATAL] [ice.c:janus_ice_set_stun_server:799] Error sending STUN BINDING test
[FATAL] [janus.c:main:3774] Invalid STUN address numb.viagenie.ca:0
```

I have tracked it down to the combination of
https://github.com/meetecho/janus-gateway/blob/master/ice.c#L756
and
https://github.com/meetecho/janus-gateway/blob/master/ice.c#L791

The first of those resolves the STUN server name, and apparently does so preferring IPv6 addresses over IPv4 ones, the second line however, uses inet_addr() which according to its man page only supports IPv4.

lminiero:
See discussion in #1053 and the attempt @mzanetti made in #1055. This patch should fix the issue as it tries to use and get the right address: tested IPv6 binding request with `blues.viagenie.ca`.
I expect more issues along the road, as we currently use that public address to also override stuff in the SDP (e.g., candidates), so we should also keep track of what kind of address our public address is, but that can wait.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -768,7 +768,7 @@ int janus_ice_set_stun_server(gchar *stun_server, uint16_t stun_port) {
 		return -1;
 	}
 	janus_stun_port = stun_port;
-	JANUS_LOG(LOG_VERB, "  >> %s:%u\n", janus_stun_server, janus_stun_port);
+	JANUS_LOG(LOG_INFO, "  >> %s:%u (%s)\n", janus_stun_server, janus_stun_port, addr.family == AF_INET ? "IPv4" : "IPv6");
 	/* Test the STUN server */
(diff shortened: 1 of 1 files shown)


lminiero:
Merging.

PR #1063, opened 2017-11-11
Experiments to improve performance of the ICE send thread

lminiero:
See discussion in this [post](https://groups.google.com/forum/#!topic/meetecho-janus/8_XGmAnOv-4) for rationale.

Basically, it looks like g_list_last and g_list_append are dragging performance down, sometimes, and this is a quick and rough experiment to address that. Since Glib GList is not a circular list, we "fix" it by keeping a pointer to the last item in the list ourselves. Anytime we need to append, we create a new item ourselves and add it to the last item we know, updating the next/prev links accordingly.

I'm abroad and will be for a while, so this is just an experiment, and definitely untested, as I don't have time for that and won't for 2-3 weeks. I don't know if this messes up the receiving/not receiving notifications, or the retransmissions buffer, both things this patch addresses. Please let me know if you notice improvements, and most importantly if this doesn't break anything.


emmanuelriou:
Thanks Lorenzo,

I will take some time to test it sometimes this week or next week!

Cheers,

Emmanuel.

mquander:
You might already know this, but the [glib documentation says](https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#glib-Doubly-Linked-Lists.description):

> If you want fast access to both the start and the end of the list, and/or the number of items in the list, use a [GQueue](https://developer.gnome.org/glib/stable/glib-Double-ended-Queues.html) instead.

That might be a more concise and less error-prone approach (we don't have to worry about bugs where we change the list but forget to update our pointer) than keeping track of the last item ourselves, with the same resulting performance.

lminiero:
@mquander good point! (and again apologies for the awfully late reply... back home for good now!) Not sure why I never considered GQueue for that: maybe I was under the (wrong) assumption you could only push to tail and pop from head.

@emmanuelriou if you can confirm you see improvements with this patch, I'll update it so that we use GQueue instead.

emmanuelriou:
Sorry @lminiero , was sick last week, couldn't do the tests as expected
I've just done a few small tests, enough to say that yes it is definitely better without g_list_last/g_list_append, better in terms of time spent in ice_send_thread spotted by perf.
So yes it is definitely an improvement, you can go ahead with GQueue stuff ;-)

lminiero:
Just made the change to use GQueue. Looking at how a GQueue is defined:

```
struct GQueue {
  GList *head;
  GList *tail;
  guint  length;
};
```

it is indeed what we were doing manually (keeping track of both head and tail) but with a much cleaner and consistent API. Please let me know if this still results in a better experience, and if there are any issues we need to fix before I can merge.

ice.c
```diff
@@ -1430,16 +1427,13 @@ void janus_ice_component_free(GHashTable *components, janus_ice_component *compo
 	}
 	if(component->retransmit_buffer != NULL) {
 		janus_rtp_packet *p = NULL;
-		GList *first = g_list_first(component->retransmit_buffer);
-		while(first != NULL) {
-			p = (janus_rtp_packet *)first->data;
-			first->data = NULL;
-			component->retransmit_buffer = g_list_delete_link(component->retransmit_buffer, first);
+		while(g_queue_get_length(component->retransmit_buffer) > 0) {
```
emmanuelriou: Is it not enough to do
while ((p = (janus_rtp_packet *)g_queue_pop_head(component->retransmit_buffer)) != NULL)
instead of recomputing the length of the queue first?

lminiero: I think I derived this from the code we use for hashtables, were if you iterate on an empty table you get an error. Shouldn't be an issue to use the approach you suggest instead.

Anyway, there's no recomputing: the length of the queue is one of the properties of the GQueue object, which means we're accessing a value that already exists, not traversing the whole queue to see how may items are there.

emmanuelriou: Yes you're right, my mistake, there is no length computing. So it is just the cost of a function call, which is not needed, so yes it is better to remove the call.
No other remarks then :-)

lminiero: Ack, I'll try and update this as soon as I can (a bit busy today). If you can provide some feedback on whether or not the patch as it is gives you improvements with no regression (I especially want to make sure we're not breaking anything with this change), that would definitely help speed up the merging progress! Thanks!

emmanuelriou: Weird, I think I had not spotted this first time:
This time, I get a lot of:
GLib-CRITICAL **: g_queue_peek_tail: assertion 'queue != NULL' failed
as soon as packets are forwarded.

Seems that g_queues have never been allocated.

It was a very simple test, with a videoroom-like plugin, one sender, and one receiver.

lminiero: We create those queues as soon as we start receiving packets, but that will never happen for a sendonly stream, so we should add a check for its existence before processing it. Will fix shortly.

lminiero: Fixed (hopefully).

emmanuelriou: OK great, works good now, no more glib critical errors.
Performance is good also, ice_send_thread is not spotted as a bottleneck by perf anymore

lminiero: Cool, thanks for confirming! Unless there is any objection, I'll make the change you suggested on the length thing and merge, then :+1:

emmanuelriou: Sounds good :-) Thanks for developing this!

lminiero:
Merging, thanks for the feedback!

PR #1064, opened 2017-11-11
Add SSL/TLS support for MQTT plugin

manifest:
PR for the [issue](https://github.com/meetecho/janus-gateway/issues/1048).


lminiero:
@oej did you have a chance to test this PR too? Thx!

oej:
In order to use TLS there should be no requirement to specify a CAcert. The normal trust store of the system should be used - but I haven't tested if the library actually use this. This requires some investigation.

As I reflected in the other PR, all versions of the SSL protocol are insecure. Even though our library use "ssl" as a prefix, I think we should promote the prefix "TLS" for all settings related to actually using TLS - since we never support SSL. It's a step forward and require some code, but a "good thing".

Also, the MQTT scheme for TLS enabled MQTT seems to be "mqtts:" in most implementations, even though the paho library for some odd reason use "ssl:".

Will try to compile and test, since I need to steal this code for my MQTT event handler implementation. ;-)

oej:
First test failed. Got no clear error message, so I have to dig down a bit.

lminiero:
Any update? Is this still not working?

lminiero:
@oej did you find out eventually what's causing it to fail? I'd like to understand if this can be merged or not. Thx!

oej:
I merged the same code to my MQTT event branch and it failed. Haven't worked for a while because of the holidays but will get on it very soon. I want TLS as we merge my event support.

lminiero:
Thx, I'll wait for your feedback then!

saket424:
I attempted to apply this PR from @manifest and it worked fine. I was able to successfully pub/sub with an tls-enabled broker protected with a username and password. I agree with @oej thoughts on this topic and await an improved PR from him that just relies on the normal trust store ohe system an no special certificates

lminiero:
Thanks for the feedback @saket424. If this kinda works already, any objection to merge this as it is, and then incorporate any fix @oej or others might have later on? Or is there any good reason not to?

saket424:
I second that Lorenzo. Let us get it merged in now and @oej can improve upon it when he has something better. BTW, I am interested in  the admin mqtt interface and will be exploring that.

saket424:
Any progress with merging this PR?

lminiero:
I'm waiting for feedback to decide if it can be merged or not.

lminiero:
Considering this works already for basic use cases requiring encryption (thanks @saket424 for the feedback!), I'll merge this as it is, for now. We can always improve the behaviour later on.

L3w1s-L1u:
I managed to enable TLS for MQTT by cheating janus on compile time that libpaho-mqtt3as.so.1.3.4 is what she wants to link with `-lpaho-mqtt3a` :
```
    sudo ln -s  /usr/local/lib/libpaho-mqtt3as.so.1.3.4   /usr/lib/libpaho-mqtt3a.so
```
I think Janus should be able to dynamically link TLS version of libpaho-mqtt by reading the config file and finding out that TLS has been enabled by user. However the behavior is not what I expected. Or is there any configure option or environment variable available for this purpose?  Neither Janus nor paho.mqtt.c's docs gives any info about this. Don't know where it goes wrong ...

cameronelliott:
> In order to use TLS there should be no requirement to specify a CAcert. The normal trust store of the system should be used - but I haven't tested if the library actually use this. This requires some investigation.
> 
> As I reflected in the other PR, all versions of the SSL protocol are insecure. Even though our library use "ssl" as a prefix, I think we should promote the prefix "TLS" for all settings related to actually using TLS - since we never support SSL. It's a step forward and require some code, but a "good thing".
> 
> Also, the MQTT scheme for TLS enabled MQTT seems to be "mqtts:" in most implementations, even though the paho library for some odd reason use "ssl:".
> 
> Will try to compile and test, since I need to steal this code for my MQTT event handler implementation. ;-)

I believe the requirement to create a CACERT has been fixed

PR #1067, opened 2017-11-12
Implement websocket LWS_CALLBACK_GET_THREAD_ID callback

RouquinBlanc:
Following [the discussion](https://groups.google.com/forum/#!topic/meetecho-janus/HsFaEXBz4Cg) on Meetecho's forum about problems using websocket transport on mac osx, implementing the LWS_CALLBACK_GET_THREAD_ID callback in protocols[0] will allow calls to `lws_callback_on_writable` to notice that the current thread is not the lws service thread, and force the ongoing `poll` (if any) to end right away instead of waiting until timeout (currently the service poll timeout is 50ms). This should speed up response time in some cases, and does not seem very harmful.

This is beneficial only when using libwebsockets@master where this new feature is implemented, but compatible with versions as far as `1.5-stable` (didn't look farther).

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -872,6 +872,8 @@ static int janus_websockets_callback_http(
 				return -1;
 			}
 			break;
+		case LWS_CALLBACK_GET_THREAD_ID:
+			return (uint64_t)pthread_self();
 		default:
 			break;
 	}


lminiero:
Thanks for this! I don't have time to test it myself now, though, as I'm abroad for a conference and busy almost all day. Can you confirm it works as expected with previous versions of libwebsockets on Linux too? As a side note, is the casting to `uint64_t` needed or fine? Shouldn't it be `pthread_t` instead, or does libwebsockets cast everything to `uint64_t` for internal comparison anyway, as it expects a 64-bit pointer?

RouquinBlanc:
I will test on linux and let you know.

The cast directly to `int` (the return type) or no cast at all was giving me a compilation error: `cast from pointer to smaller type 'int' loses information`. But casting into an uint64 an then letting the auto conversion to int worked fine... I don't have a way to test that on a 32bit platform though, but it should be fine; to be tested as well

RouquinBlanc:
Did a quick test on linux with lws 2.0.3, no problem.

lminiero:
Thx, I'll do a quick check myself in the next few days and merge if fine :+1:

lminiero:
Pinging @atoppi so that he can check too.

lminiero:
Apologies for this very late feedback, as I was abroad... Looks good, merging, thanks!

PR #1068, opened 2017-11-12
brew installation of libcurl is now called just `curl`

RouquinBlanc:
Helping the mac user to deploy using brew. Note that with [the websocket issue](https://groups.google.com/forum/#!topic/meetecho-janus/HsFaEXBz4Cg), you would need libwebsockets from master to use the libwebsockets transport on mac osx. Waiting for feedbacks on that before updating the doc on that.


lminiero:
Thanks, merging!

PR #1071, opened 2017-11-16
Fix SSL library detection

aduskett:
Currently, searching for RAND_egd doesn't seem to work properly against
OpenSSL 1.1, as it always returns false.

Instead, searching for the function tls_config_set_ca_mem in libtls seems to
be the preferred way to determine if LibreSSL is being compiled against, as
this function does not exist in OpenSSL.

Openntpd uses this callout in their configure.ac. Because Openntpd is
maintained by the OpenBSD foundation, and they also maintain LibreSSL, it would
be a good idea to follow their example.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -225,10 +225,9 @@ AS_IF([test "x$enable_dtls_settimeout" = "xyes"],
       ])
 AM_CONDITIONAL([ENABLE_DTLS_SETTIMEOUT], [test "x$enable_dtls_settimeout" = "xyes"])
 
-AC_CHECK_LIB([crypto],
-             [RAND_egd],
-             [AM_CONDITIONAL([LIBRESSL_DETECTED], false)],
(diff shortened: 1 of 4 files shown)


lminiero:
Thanks for the fix! I'm abroad for another week more or less, so won't have much time to review or check if it doesn't break anything in my setup. I'll try to do that as well. Pinging @atoppi to ask him if he can do a quick check in his setup and confirm everything still works. Thanks!

atoppi:
Janus is compiling and executing fine in my setup (`openssl 1.0.2g`).
Tried a quick test: after installing `openssl 1.1.0` from sources, Janus compiles correctly, but the process does not find openssl `.so` files at startup (they have been put in `/usr/local/lib`).
I'm not sure this is linked to the PR though.

lminiero:
Thanks, merging!

PR #1073, opened 2017-11-16
Add openssl pre 1.1 api macro

aduskett:
Instead of repeating the check for OPENSSL_VERSION_NUMBER < 0x10100000L or
defined(LIBRESSL_VERSION_NUMBER) over and over, it's cleaner to add the
checks for LIBRESSL_VERSION_NUMBER in dtls-bio.h and set a new define
JANUS_USE_OPENSSL_PRE_1_1_API if LibreSSL or OpenSSL < 1.1 are being
compiled against.

This will ease the transition to OpenSSL 1.0.2/1.1 support in the future
and will keep the code cleaner and easier to manage.

### DIFF
--- a/dtls-bio.c
+++ b/dtls-bio.c
@@ -36,7 +36,7 @@ int janus_dtls_bio_filter_new(BIO *h);
 int janus_dtls_bio_filter_free(BIO *data);
 
 /* Filter initialization */
-#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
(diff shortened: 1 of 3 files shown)


lminiero:
Didn't you add this as part of the other LibreSSL related PR already?

aduskett:
I didn't actually! I did a PR to fix the print statement because I was dumb and screwed up the logic in that c file, but this cleans up the if defines a bit more. :)

lminiero:
Makes sense, thanks, merging!

PR #1082, opened 2017-11-22
fixing media_hangup deadlock

suranapranay:
Not setting this flag and returning leads to eventual deadlock while sessions_mutex in janus.c is acquired but not released indefinitely by the "janus_audiobridge_hangup_media"

janus_audiobridge.c
while(participant->working)
2741     g_usleep(5000);

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -2613,7 +2613,7 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 			JANUS_LOG(LOG_ERR, "[Opus] Ops! got an error accessing the RTP payload\n");
 			g_free(pkt->data);
 			g_free(pkt);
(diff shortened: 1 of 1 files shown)


lminiero:
Have you really encountered issues with accessing the payload of a packet? Sounds weird. Maybe a padding only packet? Anyway, we should indeed clear the flag in that case, so good catch!

That said, please sign our [CLA](https://cla.conf.meetecho.com/janus/) at the end of this page so that we can merge.

suranapranay:
Signed it.
Yes I saw that in one of my logs.
Sometimes I was seeing Null room id so I had put in a nullcheck for it, by copying this check. I was seeing my check hit a lot more than this one. That is another bug I'm trying to figure out.

lminiero:
Apologies for this late reply, just came back home after a long period abroad. Thanks again, merging!

PR #1085, opened 2017-11-25
Update README for certificates

oej:
- 2048 bit key in the example
- Pointer to letsencrypt.org


lminiero:
Thanks for the suggested text, merging!

PR #1089, opened 2017-11-27
Refcount improvements

atoppi:
This PR tries to improve the `refcount` branch in many ways.
Some of the changes are ported from #1035 .

Changelog ins random order:
- reject sdps and trickles directed to handles in CLEANING state. This should avoid strange behaviors when quickly destroying/creating a peer connection on the same handle.
- use the active plugin sessions map in place of the old plugin sessions map. This is a pretty straightforward change, Janus should check the validity of a given plugin session using a live map of active plugin sessions and not storing the deleted ones.
- Check `g_main_loop_is_running` only before doing `g_main_loop_quit`. The rationale here is not using the glib loop status to control the flow in Janus logic, but only to manage the glib loop.
-  Add some missing `janus_plugin_session_is_alive` checks. This should avoid some potential crashes.
-  Front-push the DTLS alert packet onto the packet queue. This should help getting a faster and cleaner peer connection teardown.
- `janus_ice_thread` joins `send_thread` in case the latter has been created and a `detach` has been requested. This should avoid a crash happening when the `send_thread` uses contexts and resources being freed by `janus_ice_thread`.
- `janus_ice_thread` creation is moved to the bottom of `janus_ice_setup_local`, in order to create the iceloop thread when all the resources and the contexts that the thread needs are correctly initialized. As a consequence the `sleep` call has been removed from `janus_ice_thread`, because the loop can start without any other waiting.
- Remove the wait for `g_main_loop_is_running` in `janus_ice_webrtc_hangup`. I think this is not the correct approach to handle a hangup request received too early (e.g. when the glib loop still have to pass in running state). This issue will be discussed and addressed later.

### DIFF
--- a/apierror.h
+++ b/apierror.h
@@ -63,6 +63,8 @@
 #define JANUS_ERROR_UNEXPECTED_ANSWER			469
 /*! \brief The auth token the request refers to doesn't exist */
 #define JANUS_ERROR_TOKEN_NOT_FOUND				470
+/*! \brief The current request cannot be handled because of not compatible WebRTC state */
+#define JANUS_ERROR_WEBRTC_STATE				471
 
 
 /*! \brief Helper method to get a string representation of an API error code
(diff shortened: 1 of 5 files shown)


lminiero:
Closing as tracking new changes in #1097.

PR #1090, opened 2017-11-28
Add a "configure" request to SIP plugins to alter sending audio/video RTP to remote peer

saket424:
Add a "configure" request to SIP plugins so as to allow temporarily disable/enable sending of audio/video rtp to remote peer without the latency associated with hold/unhold signalling

### DIFF
--- a/plugins/janus_nosip.c
+++ b/plugins/janus_nosip.c
@@ -152,6 +152,10 @@ static struct janus_json_parameter recording_parameters[] = {
 	{"peer_video", JANUS_JSON_BOOL, 0},
 	{"filename", JSON_STRING, 0}
 };
+static struct janus_json_parameter configure_parameters[] = {
+        {"audio", JANUS_JSON_BOOL, 0},
+        {"video", JANUS_JSON_BOOL, 0}
+};
 
 /* Useful stuff */
 static volatile gint initialized = 0, stopping = 0;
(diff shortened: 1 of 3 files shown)


lminiero:
Thanks! Notice that, since you're re-using `audio_send` and `video_send`, a call to "hold" or "unhold" will potentially undo whatever you configured manually. If that's not what you want, you may want to add new properties that can override whatever `audio_send` and `video_send` are dynamically set to in other parts of the code.

That said, please sign the [CLA](https://cla.conf.meetecho.com/janus/) at the end of that page so that we can merge this when ready.

saket424:
I will sign the CLA soon, after I get approvals from the legal team from the company I consult for.

My intended use of configure "audio/video/data" is to have a consistent interface in the plugin to alter a peer connection dynamically in videoroom,streaming,sip,..  -- aka a unilateral brute force mechanism without the necessity to signal that intent to the remote peer/gateway that you are doing so for a variety of reasons which may include saving bandwidth, avoiding roundtrip signalling latency overhead in real-time sensitive applications etc.

So the expectation is you either use the clean sip standards based hold/unhold re-invite method of notifying the remote peer of your actions or this brute-force unliateral configure method but not both. So the questions of inconsistency/undoing should not arise because an API user will pick one strategy or the other and stick with it. do you agree with this set of assumptions?

lminiero:
Well, you cannot guarantee one won't use both. Besides, once we'll get proper renegotiation working, it may be a remote re-INVITE that updates the media direction and those two properties, so something outside of the user's control. That's why I highlighted the fact those properties `configure` now can modify can actually be overwritten by other events: if this is something you'd rather avoid, and still allow a `configure` directive to act, then some new properties (e.g., `audio_send_cfg`) that are also checked when relaying packets would be needed.

saket424:
I had not considered the remote end re-negotiations aspects and the inconsistencies that it may cause. Will have to rethink the applicability of this patch. Adding all new cfg properties in sip file to specify which strategy takes precedence etc sounds messy. let us table this PR for the moment since hold/unhold already exists

PR #1092, opened 2017-11-28
Log error codes for SCTP-related errors

mquander:
Self-explanatory. All the usrsctp library calls return -1 and set `errno` on error. See the [usrsctp documentation](https://github.com/sctplab/usrsctp/blob/master/Manual.md) for details.


lminiero:
Thanks for this, definitely helpful! I may add a call to `strerror` to those later on just to make them even clearer. Merging!

Issue #1091, opened 2017-11-28
Handle partial messages in data channels


lgrahl:
I've had a glance at your code and it seems you're not handling the `MSG_EOR` flag which means you will treat incoming partial SCTP messages as whole messages, thus violating the message-oriented principle of SCTP (and data channels). This is a widespread issue in many implementations we're currently trying to clean up.

If you need help, you can look at the [relevant section in RAWRTC's code](https://github.com/rawrtc/rawrtc/blob/master/src/librawrtc/sctp_transport.c#L1406) or at [Mozilla's implementation](https://dxr.mozilla.org/comm-central/source/mozilla/netwerk/sctp/datachannel/DataChannel.cpp#1737). I'm the author of RAWRTC and I've also worked on Mozilla's data channel code, so feel free to ping me if you have any questions.

Tracked (and resolved) for Firefox in [bug 979417](https://bugzilla.mozilla.org/show_bug.cgi?id=979417). For Chromium, this is tracked in [issue 7774](https://bugs.chromium.org/p/webrtc/issues/detail?id=7774).

lminiero:
Thanks for the heads up! I heavily based the SCTP code on one of the examples provided as part of the `usrsctp` repo, so if it wasn't there, it makes sense we're not handling that flag: we definitely didn't add it ourselves, as my knowledge of the library is frankly limited.

What would this entail? Simply fill in a buffer until `MSG_EOR` is set, and only then pass the message to the core in case?

lgrahl:
Yeah, I think many people based it on that example. That would explain why there are so many implementations that don't handle the flag.

> Simply fill in a buffer until `MSG_EOR` is set, and only then pass the message to the core in case?

Exactly that. Be aware that there is a corner case for unreliable messages being aborted. See [my code](https://github.com/rawrtc/rawrtc/blob/master/src/librawrtc/sctp_transport.c#L823) that handles this event (which needs to be registered).

lminiero:
Created a simple PR for the buffering part. I'll work on the corner case you mentioned later on, thanks for the pointer!

lminiero:
As a side note, do you know of any way to test this?

lgrahl:
There isn't an easy way to test this but with my master thesis starting soon, I'll definitely write a test for that case. Ping me in a few months time in case you don't hear from me.

lminiero:
[...]

PR #1093, opened 2017-11-29
Handle MSG_EOR in datachannels

lminiero:
See #1091.


lminiero:
Well, actually we currently support a single data channel per PeerConnection: that's a limitation we added in the very early days, when I first implemented the support for DC, and then we never removed it, as it would need some core-level refactoring as well (most importantly, making applications aware of multiple channels, rather than the global "data" one they see now).

lgrahl:
Oh, okay, I wasn't aware of that.

lminiero:
Merging, then: I'll work on the edge case you mentioned later on, as I don't have much time for that right now, unfortunately, and I don't want to keep this frozen too long. Thanks again!

PR #1097, opened 2017-12-04
Refcount improvements 2

atoppi:
This is a subset and a refinement of #1089, after #1035 has been merged into `master`.
Respect to the original #1089 the joining mechanism for `janus_ice_thread` with `send_thread` has been changed from a `g_thread_join` to a way less elegant `while(!condition)` and `sleep` loop. This is because `g_thread_join` consumes a reference to the passed-in thread, and I think it is better to handle the reference from inside the `send_thread`, mainly because `janus_ice_thread` and `send_thread` do have a separate and different life cycle:
- `janus_ice_thread` can be short termed and can be destroyed and recreated when needed.
- `janus_ice_send_thread` is created once, just after the first `janus_ice_thread`.
Using a check on `handle->send_thread` we are sure that in case of `detach` the `send_thread` terminates before `janus_ice_webrtc_free` is called.


lminiero:
Looks good to me, merging!

PR #1103, opened 2017-12-08
Instruct ESLint to ignore Janus JS module contents

cmacq2:
Instruct ESLint to ignore Janus JS module contents when linting your project.

This change makes the generated JS modules easier to consume in your project, now it really should be just a drop-in.


lminiero:
Aren't all the changes you made just comments? Or am I missing something?

cmacq2:
Yes but the first comment: `/* eslint-disable */` is what does the trick. ;)

lminiero:
Thanks for the clarification, merging then!

PR #1104, opened 2017-12-08
Provide support for CI, using Docker & Travis CI

cmacq2:
The purpose of this PR is to make it a *lot* easier to run CI on the Janus project. Incidentally it also makes it a lot easier to build/test various different build environments/configurations for Janus.

With this PR:
 - A developer can use predefined Docker images to build Janus and do basic sanity checking.
 - A Travis CI configuration is added, if you opt in to letting Travis run on your (fork of this) repository you get automatic CI (consisting of running the Docker images) for free.
 - In theory you no longer need to have all the development dependencies installed to work on Janus, instead you could just use docker.

The PR consists of two commits, the first of which adds a couple of predefined docker files to build the Docker images from, and the second adds the Travis CI integration.

A couple of points worth noting:

 1. Currently the docker files refer to 'my' Docker Hub repository, ideally there would be a Meetecho one to refer to instead. I'd suggest that if this PR is merged, a follow up commit changes this once the images have been synced to a new Meetecho repository.
 2. I'd like to propose that Meetecho also enable Travis CI on the Janus repository once the PR is merged.
 3. The docker images are by no means intended as the 'definitive' judgement on what build configurations the project should support, instead they are meant to get the ball rolling and provide a starting point for discussion.
 4. Yes, these images are pretty large. Although it is not too bad for an occasional download and nothing which CI shouldn't be able to handle, if you are actively developing & pushing the images you probably do want to be on a cheap + fast broadband link.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+
+sudo: required
+
+node_js:
+    - node
+
+notifications:
+    email:
+        on_success: change
+        on_failure: change
+
+script: ./docker/travis-ci.sh
+
+env:
+ - image=cmacq2/janus-build-env:js-modules
+ - image=cmacq2/janus-build-env:http-w-openssl
+
--- a/docker/.dockerignore
+++ b/docker/.dockerignore
@@ -1,0 +1,4 @@
+*.docker
+travis-ci.sh
+build-docker-images.sh
+.dockerignore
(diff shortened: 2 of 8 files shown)


lminiero:
Sorry, I'm not going to add docker stuff to the repo. This has been proposed other times in the past, and I've always replied it's orthogonal to the project itself. I believe this belongs to a different repo, as the projects [listed here](https://janus.conf.meetecho.com/docs/resources#docker).

lminiero:
Closing: please let me know if a separate project will be created for this, and I'll be glad to add to the [resources](https://janus.conf.meetecho.com/docs/resources)!

PR #1110, opened 2017-12-15
Handle RTCP for all remote SSRCs, including video simulcast

lminiero:
As the title says, the plan for this PR is to add support for RTCP feedback not only for the regular audio and video, but also for the simulcast SSRC streams, if available.

At the time of writing, we're able to process incoming RTCP feedback addressing those SSRCs, but we don't generate RTCP RR packets for those additional SSRCs yet, which means the work is quite obviously not complete. That said, I'm publishing anyway as all the required changes core-wise have been done already, so it's a matter of actually preparing those packets now and sending them. This will need changes to the way we invoke `janus_rtcp_fix_ssrc`, that is currently hardcoded to the old way of doing things: I'll rewrite that part next week.


lminiero:
Tested (especially RR and NACK) and works for me for the additional SSRCs. Merging.

PR #1113, opened 2017-12-21
Simplify internals by always assuming single stream/component

lminiero:
I thought about this for a while, and then eventually went on with it.

Just to give you some context, so far Janus was flexibly in terms of ICE streams and components. In case the client didn't support bundle, we would create different streams (and so allocate different ports, DTLS contexts, etc.) per media; for the same reason, just in case the client didn't support rtcp-mux, we would create different components (again, different ports, DTLS contexts, etc.) for RTP and RTCP too. If it turned up the client did indeed support them, we would get rid of the extra streams and components, and fall back to the single one to use.

We already had a "force bundle and/or force rtcp-mux" option when launching Janus, and for Chrome and Firefox we would force it from janus.js anyway. Now that we know that Safari and Edge do support both too, it really makes no sense to keep this optional support around: it made the code too complex for no reason at all. Besides, the upcoming multistream support would have made this optional support impossible to handle (we can't possibly allocate 10 ports for every new PeerConnection). As such, I just went on and got rid of all the configurable bundle/rtcp-mux stuff: now we just assume they're supported by default, and greatly simplified the code as you can see from the diff.

Please test and let me know if something broke: I'm planning to merge soon, so if you can make additional tests with Edge and Safari too just to confirm they still work as expected, it would help too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,11 +350,6 @@ or on the command line:
 	-T, --ice-tcp                 Whether to enable ICE-TCP or not (warning: only
                                   works with ICE Lite)
                                   (default=off)
-	-U, --bundle                  Whether to force BUNDLE or not (whether audio,
-                                  video and data will always be bundled)
-                                  (default=off)
-	-u, --rtcp-mux                Whether to force rtcp-mux or not (whether RTP
-                                  and RTCP will always be muxed)  (default=off)
 	-q, --max-nack-queue=number   Maximum size of the NACK queue (in ms) per user
                                   for retransmissions
 	-t, --no-media-timer=number   Time (in s) that should pass with no media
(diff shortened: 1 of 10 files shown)


mquander:
The echo test audio and video seems to work for me with this branch on Windows 10, default configs, latest Edge, running Janus locally under WSL. Is there any other specific thing I should test?

lminiero:
I'm pretty sure this confirms that it works with Edge, but if you could
test something where Janus offers as well (in the echo test the SDP offer
always comes from the user) that would be great! Streaming demo or
Videoroom demo with a couple people in would suffice.

Thanks for the feedback, and have some great holidays!

Il 23 dic 2017 10:59 PM, "Marshall Quander" <notifications@github.com> ha
scritto:

The echo test audio and video seems to work for me with this branch on
Windows 10, default configs, latest Edge, running Janus locally under WSL.
Is there any other specific thing I should test?

—
You are receiving this because you authored the thread.
<https://github.com/meetecho/janus-gateway/pull/1113#issuecomment-353750548>,
or mute the thread

lminiero:
I didn't see any objection and @mquander confirmed it still works with Edge, so I'll merge.

PR #1114, opened 2017-12-21
Recursive mkdir when a full path has been passed as recording filename.

atoppi:
This PR enables Janus to create the full path specified for a recording file.
As of now the API `janus_recorder_create` accepts a `dir` and a `filename`, then builds the path only if `dir` has been passed and opens the file `/dir/filename` or simply `filename`.
With the changes proposed here, the new logic is:
- **with** `dir` and **with** `filename` <del> check that `filename` is not a path too, and then preserve previous behavior</del> warn if `dir` and `filename` are paths  then preserve previous behavior
- **without** `dir` and **with** `filename` split `filename` in `parent` and `base`, then continue with previous behavior using `parent` as `dir` and `base` as `filename`
- any other case should be identical to previous implementation (e.g. wiithout `filename` a kind of random file name is generated)

The splitting of `filename` has been obtained with the methods `dirname` and `basename` of `libgen.h`. These methods should be available in all Unix-like systems.

### DIFF
--- a/record.c
+++ b/record.c
@@ -108,18 +108,17 @@ janus_recorder *janus_recorder_create(const char *dir, const char *codec, const 
 		const char *filename_parent = dirname(copy_for_parent);
 		/* Get filename base file */
 		const char *filename_base = basename(copy_for_base);
-		/* If "filename" is a actual file name and not a path, then dirname returns "." and basename "filename" */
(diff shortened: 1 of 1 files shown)


lminiero:
Looks good to me, merging!

PR #1118, opened 2017-12-26
 Enable transport wide cc for publisher streams on video room plugin

murillo128:
cc/ @lminiero

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -324,24 +324,40 @@ uint janus_get_max_nack_queue(void) {
 	return max_nack_queue;
 }
 /* Helper to clean old NACK packets in the buffer when they exceed the queue time limit */
-static void janus_cleanup_nack_buffer(gint64 now, janus_ice_stream *stream) {
+static void janus_cleanup_nack_buffer(gint64 now, janus_ice_stream *stream, gboolean audio, gboolean video) {
 	if(stream && stream->rtp_component) {
 		janus_ice_component *component = stream->rtp_component;
 		janus_mutex_lock(&component->mutex);
-		if(component->retransmit_buffer) {
-			janus_rtp_packet *p = (janus_rtp_packet *)g_queue_peek_head(component->retransmit_buffer);
-			while(p && (now - p->created >= (gint64)max_nack_queue*1000)) {
+		if(audio && component->audio_retransmit_buffer) {
+			janus_rtp_packet *p = (janus_rtp_packet *)g_queue_peek_head(component->audio_retransmit_buffer);
+			while(p && (!now || (now - p->created >= (gint64)max_nack_queue*1000))) {
 				/* Packet is too old, get rid of it */
-				g_queue_pop_head(component->retransmit_buffer);
+				g_queue_pop_head(component->audio_retransmit_buffer);
 				/* Remove from hashtable too */
 				janus_rtp_header *header = (janus_rtp_header *)p->data;
 				guint16 seq = ntohs(header->seq_number);
-				g_hash_table_remove(component->retransmit_seqs, GUINT_TO_POINTER(seq));
+				g_hash_table_remove(component->audio_retransmit_seqs, GUINT_TO_POINTER(seq));
 				/* Free the packet */
 				g_free(p->data);
 				p->data = NULL;
 				g_free(p);
-				p = (janus_rtp_packet *)g_queue_peek_head(component->retransmit_buffer);
+				p = (janus_rtp_packet *)g_queue_peek_head(component->audio_retransmit_buffer);
+			}
+		}
+		if(video && component->video_retransmit_buffer) {
+			janus_rtp_packet *p = (janus_rtp_packet *)g_queue_peek_head(component->video_retransmit_buffer);
+			while(p && (!now || (now - p->created >= (gint64)max_nack_queue*1000))) {
+				/* Packet is too old, get rid of it */
+				g_queue_pop_head(component->video_retransmit_buffer);
+				/* Remove from hashtable too */
+				janus_rtp_header *header = (janus_rtp_header *)p->data;
+				guint16 seq = ntohs(header->seq_number);
+				g_hash_table_remove(component->video_retransmit_seqs, GUINT_TO_POINTER(seq));
+				/* Free the packet */
+				g_free(p->data);
+				p->data = NULL;
+				g_free(p);
+				p = (janus_rtp_packet *)g_queue_peek_head(component->video_retransmit_buffer);
 			}
 		}
 		janus_mutex_unlock(&component->mutex);
@@ -1365,6 +1381,10 @@ void janus_ice_stream_free(GHashTable *streams, janus_ice_stream *stream) {
 	stream->audio_payload_types = NULL;
 	g_list_free(stream->video_payload_types);
 	stream->video_payload_types = NULL;
+	g_free(stream->audio_codec);
+	stream->audio_codec = NULL;
+	g_free(stream->video_codec);
+	stream->video_codec = NULL;
 	g_free(stream->audio_rtcp_ctx);
 	stream->audio_rtcp_ctx = NULL;
 	g_free(stream->video_rtcp_ctx[0]);
@@ -1414,20 +1434,35 @@ void janus_ice_component_free(GHashTable *components, janus_ice_component *compo
 		janus_dtls_srtp_destroy(component->dtls);
 		component->dtls = NULL;
 	}
-	if(component->retransmit_buffer != NULL) {
+	if(component->audio_retransmit_buffer != NULL) {
 		janus_rtp_packet *p = NULL;
-		while((p = (janus_rtp_packet *)g_queue_pop_head(component->retransmit_buffer)) != NULL) {
+		while((p = (janus_rtp_packet *)g_queue_pop_head(component->audio_retransmit_buffer)) != NULL) {
 			/* Remove from hashtable too */
 			janus_rtp_header *header = (janus_rtp_header *)p->data;
 			guint16 seq = ntohs(header->seq_number);
-			g_hash_table_remove(component->retransmit_seqs, GUINT_TO_POINTER(seq));
+			g_hash_table_remove(component->audio_retransmit_seqs, GUINT_TO_POINTER(seq));
 			/* Free the packet */
 			g_free(p->data);
 			p->data = NULL;
 			g_free(p);
 		}
-		g_queue_free(component->retransmit_buffer);
-		g_hash_table_destroy(component->retransmit_seqs);
+		g_queue_free(component->audio_retransmit_buffer);
+		g_hash_table_destroy(component->audio_retransmit_seqs);
+	}
+	if(component->video_retransmit_buffer != NULL) {
+		janus_rtp_packet *p = NULL;
+		while((p = (janus_rtp_packet *)g_queue_pop_head(component->video_retransmit_buffer)) != NULL) {
+			/* Remove from hashtable too */
+			janus_rtp_header *header = (janus_rtp_header *)p->data;
+			guint16 seq = ntohs(header->seq_number);
+			g_hash_table_remove(component->video_retransmit_seqs, GUINT_TO_POINTER(seq));
+			/* Free the packet */
+			g_free(p->data);
+			p->data = NULL;
+			g_free(p);
+		}
+		g_queue_free(component->video_retransmit_buffer);
+		g_hash_table_destroy(component->video_retransmit_seqs);
 	}
 	if(component->candidates != NULL) {
 		GSList *i = NULL, *candidates = component->candidates;
@@ -2092,7 +2127,6 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 			if(res != srtp_err_status_ok) {
 				if(res != srtp_err_status_replay_fail && res != srtp_err_status_replay_old) {
 					/* Only print the error if it's not a 'replay fail' or 'replay old' (which is probably just the result of us NACKing a packet) */
-					janus_rtp_header *header = (janus_rtp_header *)buf;
 					guint32 timestamp = ntohl(header->timestamp);
 					guint16 seq = ntohs(header->seq_number);
 					JANUS_LOG(LOG_ERR, "[%"SCNu64"]     SRTP unprotect error: %s (len=%d-->%d, ts=%"SCNu32", seq=%"SCNu16")\n", handle->handle_id, janus_srtp_error_str(res), len, buflen, timestamp, seq);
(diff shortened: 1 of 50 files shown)


lminiero:
Thanks for this! I'm still away for the holidays for a few more days, so I won't be able to do a deep review until then. For the time being, I'll add some quick notes to the code (mostly related to nits, though, e.g., aligning to the code style).

ice.c
```diff
@@ -1245,6 +1245,8 @@ void janus_ice_stream_free(janus_ice_stream *stream) {
 	stream->video_rtcp_ctx[1] = NULL;
 	g_free(stream->video_rtcp_ctx[2]);
 	stream->video_rtcp_ctx[2] = NULL;
+	g_slist_free(stream->transport_wide_received_seq_nums);
```
lminiero: This is a list of allocated elements (pointers to `janus_rtcp_transport_wide_cc_stats` objects), but the allocated values are never freed. You probably want a [g_slist_free_full](https://developer.gnome.org/glib/stable/glib-Singly-Linked-Lists.html#g-slist-free-full) here, that uses `g_free` as the free function.

ice.c
```diff
@@ -2023,6 +2025,44 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 							stream->video_is_keyframe = &janus_h264_is_keyframe;
 					}
 				}
+				/* If we need to handle transport wide cc*/
+				if (stream->do_transport_wide_cc)
```
lminiero: Please put the `{` on the same line as the condition (just a nit to match the coding style of the rest of the code). There are a few other occurences in the code too.

ice.c
```diff
@@ -2023,6 +2025,44 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
...
+						gettimeofday(&now,0);
+						
+						/* Create <seq num, time> pair */
+						janus_rtcp_transport_wide_cc_stats* stats = g_malloc0(sizeof(janus_rtcp_transport_wide_cc_stats));
+						if(stats == NULL) {
+							JANUS_LOG(LOG_FATAL, "Memory error!\n");
+							return;
+						}
+						/* Check if we have a sequence wrap */
+						if (transport_seq_num<0x0FFF && (stream->transport_wide_cc_last_seq_num&0xFFFF)>0xF000)
```
lminiero: Can this check be confused by out of sequence packets, and/or retransmissions (we don't do RF4588 yet), thus increasing the cycles more times than it should?

ice.c
```diff
@@ -3183,6 +3227,78 @@ void *janus_ice_send_thread(void *data) {
...
+				char rtcpbuf[1300];
+				
+				/* Lock session */
+				janus_mutex_lock(&handle->stream->mutex);
+				
+				/* Order packet list */
+				GSList* sorted = g_slist_sort(handle->stream->transport_wide_received_seq_nums,rtcp_transport_wide_cc_stats_comparator);
+				
+				/* Create full stats queue */
+				GQueue* packets = g_queue_new();
```
lminiero: Wouldn't it be better to just create the queue once, as a property of the stream, instead of allocating one everytime we need to send the feedback?

ice.c
```diff
@@ -3183,6 +3227,78 @@ void *janus_ice_send_thread(void *data) {
...
+				
+				/* For all packets */
+				for (GSList *it = sorted; it; it = it->next)
```
lminiero: Just thinking out loud here, so please correct me if I'm wrong, but rather than iterating on a possibly long list for every feedback message (something we've removed for other features we had, as from code profiling it's where we took most of the time) a possible alternative might be working on the RTCP message packet by packet: i.e., the RTCP message is part of the stream struct, and every time we get a new RTP packet, we update the RTCP message itself as you do here rather than iterating, and when it's time to send it, we just send what we prepared over time. This would save us the need to allocate and keep a list, and iterating over it at specified instants: we can just reset the RTCP message when we're done, and always work on the same one. Does this make sense?

rtcp.c
```diff
@@ -1044,3 +1044,345 @@ int janus_rtcp_nacks(char *packet, int len, GSList *nacks) {
 	rtcp->length = htons(words);
 	return words*4+4;
 }
+
+
+inline void set1(guint8 *data,size_t i,guint8 val)
```
lminiero: Just a nit (as this is code specific to rtcp.c and so is never visible outside of here) but please prefix the internal methods as well (e.g., `janus_rtcp_set1` or whatever prefix makes more sense), as it makes for better readability and generated documentation in general. As a side note, please add a space after the commas in signatures as well.

rtcp.c
```diff
@@ -1044,3 +1044,345 @@ int janus_rtcp_nacks(char *packet, int len, GSList *nacks) {
...
+}
+inline void set4(guint8 *data,size_t i,guint32 val)
+{
+	data[i+3] = (guint8)(val);
+	data[i+2] = (guint8)(val>>8);
+	data[i+1] = (guint8)(val>>16);
+	data[i]   = (guint8)(val>>24);
+}
+
+enum PacketStatus
```
lminiero: Just a nit again (sorry about that, I know how annoying I am! :smile: ), but please don't use camelcase in the code: something like `janus_rtp_packet` would match the code better, and same for the values (e.g., `not_received`).

ice.c
lminiero: UPDATE: seeing how you pass the queue to create the message, we might use that as the part we update packet by packet, instead of storing a list to update the queue: I guess this would still do the trick and help make it more efficient by removing the traversal (or at least A traversal).

rtcp.c
```diff
@@ -1044,3 +1044,345 @@ int janus_rtcp_nacks(char *packet, int len, GSList *nacks) {
...
+	
+	/* Next byte */
+	len += 8;
+	
+	/* Initial time in us */
+	guint64 timestamp = 0;
+	
+	/* Store delta array */
+	GQueue* deltas = g_queue_new();
+	GQueue* statuses = g_queue_new();
```
lminiero: Still not sure about this additional allocations done for each RTCP feedback message, but I'll have to dig in the code deeper to understand if we can avoid those.

rtcp.c
```diff
@@ -1044,3 +1044,345 @@ int janus_rtcp_nacks(char *packet, int len, GSList *nacks) {
...
+	while (len%4)
+		/* Add padding */
+		data[len++] = 0;
+	
+	/* Set RTCP Len */
+	rtcp->length = htons((len/4)-1);
+	
+	/* Done */
+	return len;
+}
```
lminiero: Missing new line, possibly something your text editor did?

rtp.c
```diff
@@ -222,6 +222,19 @@ int janus_rtp_header_extension_parse_rtp_stream_id(char *buf, int len, int id,
 	return 0;
 }
 
+int janus_rtp_header_extension_parse_transport_wide_cc(char *buf, int len, int id,
+	uint16_t* transSeqNum) {
+	uint32_t bytes = 0;
+	if(janus_rtp_header_extension_find(buf, len, id, NULL, &bytes, NULL) < 0)
+		return -1;
+	//  0                   1                   2                   3
```
lminiero: Just a nit: please use `/*` to `*/` instead of `//`.

lminiero:
@murillo128 FYI, I just merged the `bundle-rtcpmux-only` branch in master, which means you can use master as a basis for this effort now. Thanks!

lminiero:
@murillo128 just tried to build but it fails for me with a ton of undefined references to `set1`, `set2`, `set3` and `janus_push_bits`

```
janus-rtcp.o: In function `janus_rtcp_transport_wide_cc_feedback':
/home/lminiero/workspace/code/soleil/github/pulls/twcc/rtcp.c:1122: undefined reference to `set2'
/home/lminiero/workspace/code/soleil/github/pulls/twcc/rtcp.c:1123: undefined reference to `set2'
/home/lminiero/workspace/code/soleil/github/pulls/twcc/rtcp.c:1126: undefined reference to `set1'
/home/lminiero/workspace/code/soleil/github/pulls/twcc/rtcp.c:1160: undefined reference to `set3'
/home/lminiero/workspace/code/soleil/github/pulls/twcc/rtcp.c:1201: undefined reference to `janus_push_bits'
[..]
collect2: error: ld returned 1 exit status
```

Apparently, the cause is the `inline` directive for those methods. I fixed the `set*` methods by defining them as `static inline` instead of just `inline` (as they're not used outside of `rtcp.c`anyway). For `janus_push_bits` instead I had to remove the `inline` entirely: to avoid the `no previous prototype` warning, I then just left the signature in `utils.h` and moved the implementation to `utils.c` instead.

Not sure why the `inline` stuff doesn't work: it may be related to my compiler (`6.4.1`) or something else (compile options?) but don't know what. Any idea on the cause? I guess making those functions inline is supposed to increase their performance, so it sucks if we can't use it. At any rate, with these changes it compiles for me, so I'll try and make some tests and come back with more feedback besides what I already wrote a few days ago.

As a side note (not breaking), I also got a:

    ice.c:3007:6: warning: no previous prototype for ‘rtcp_transport_wide_cc_stats_comparator’ [-Wmissing-prototypes]

which can be fixed by making that method `static` in `ice.c`.

lminiero:
PS: once you move `janus_push_bits` to `utils.c` and just leave the definition in `rtcp.h`, you can define the implementation `inline` apparently. The compiler doesn't seem to complain: I'll have to check if it will crash at runtime when trying to access the method.

lminiero:
FYI, patch to get to where I am now available here: https://pastebin.com/eSGHvqRU

lminiero:
Made a few tests and I took some notes: sharing them before I forget, so that we can discuss those, especially the last point.

1. First of all, it looks like REMB do still work even when transport wide cc is enabled: to me this is good news, as it means that, although we allow the client to compute its own bandwidth, we can still put a cap on it from the server side if we want/need. That's cool!

2. It looks like Chrome only puts the extension in video packets, but then puts the `rtcp-fb` `transport-cc` attribute in the audio m-line too? I expected both to be present in both media streams, as this is a transport wide thing: now I understand that when you clarified the "transport wide" purpose you probably meant with respect to simulcast streams, and how the related sequence number increases sequentially across those multiple streams, and not audio/video.

3. Partly related to that (and only a memo) we should probably get rid of the `rtcp-fb` `transport-cc` attribute when generating an answer and the transport wide extension has NOT been enabled: I made some tests with and without the extension, and the attribute is there in all cases. That said, I see you have a note for that, saying: "It is safe to add transport-wide rtcp feedback mesage here, won't be used unless the header extension is negotiated"

4. Even enabling the extension, the CPU usage seems pretty much the same as when it's not used: I guess this makes the considerations I made about the lists and allocated queues in my previous review less relevant. I still think that part can be made more efficient, but this is something that can definitely wait, and may only be needed after this is merged.

5. Disabling the bitrate mechanism in the videoroom (`bitrate=0` in the configuration or setting `No limit` in the demo bandwidth control) and enabling the transport-wide setting for the default room, I see it working: I captured a pcap dump via our integrated text2pcap support and saw both the extension being sent by the browser (id=5) and the regular feedback being sent by Janus (type=205, fmt=15). That said, it looks like the bandwidth decisions taken by the browser are a bit "weird": I've done all testing on my local machine, but the bandwidth rarely got to 1mbps, and was often dropping to then get up again, and so on. At first I thought this could be because the default demo room sends a PLI on a regular basis (every 10 seconds), which could be seen by the browser as indication of a problem: even disabling that, though, the behaviour was pretty much the same, actually maybe even worse. The same kinda happened when simulcasting as well. It's definitely not a throughput issue, as if I disable the extension, and just force a bitrate of say 2mbps, it works fine. Do you know what may be causing this? Is the same happening to you too? I guess in general the existing RR feedback might contribute to that to, but since it's all local it shouldn't happen (unless we're doing something wrong when preparing and sending the messages); besides, if that were the case, I guess it would affect the case of extension-disabled-high-REMB-value too. Attaching a snapshot of the `bitsSentPerSecond` graph from webrtc-internals in the two scenarios to give you an idea:

Extension enabled, bitrate=0, periodic PLI (10s)
transport-wide-cc-bits

Extension enabled, bitrate=0, no periodic PLI
transport-wide-cc-bits_nopli

Thanks again for this very valuable effort, looking forward to see it evolve and be merged soon!

murillo128:
> Partly related to that (and only a memo) we should probably get rid of the rtcp-fb transport-cc attribute when generating an answer and the transport wide extension has NOT been enabled: I made some tests with and without the extension, and the attribute is there in all cases. That said, I see you have a note for that, saying: "It is safe to add transport-wide rtcp feedback mesage here, won't be used unless the header extension is negotiated"

The problem is that from what I see the extensions are set by the janus apps, and the rtcp-fb attributes are set by janus core, so we would need to signal from the app to the core when it has to be enabled or not. Adding the rtcp-fb transport-cc is harmless, as it only signals that you are able to receive them, not if they should be used (which is controlled by the extension), so IMHO it should be easier and safe to include it always.

> That said, it looks like the bandwidth decisions taken by the browser are a bit "weird": I've done all testing on my local machine, but the bandwidth rarely got to 1mbps, and was often dropping to then get up again, and so on. At first I thought this could be because the default demo room sends a PLI on a regular basis (every 10 seconds), which could be seen by the browser as indication of a problem: even disabling that, though, the behaviour was pretty much the same, actually maybe even worse. The same kinda happened when simulcasting as well. It's definitely not a throughput issue, as if I disable the extension, and just force a bitrate of say 2mbps, it works fine. Do you know what may be causing this? Is the same happening to you too? I guess in general the existing RR feedback might contribute to that to, but since it's all local it shouldn't happen (unless we're doing something wrong when preparing and sending the messages); besides, if that were the case, I guess it would affect the case of extension-disabled-high-REMB-value too. 

I will make more tests to see if I can provide more insights about that

ice.c
murillo128: The problem are out of order packets, as you need to provide the difference with previous one. So if you  receive packet n, misses n+1, and receives n+2, then you will calculate the delta for n+2 since n. But if you receive later n+1, the delta for n+2 would be incorrect.

lminiero: Ack, thx for clarifying!

murillo128:
Checked my implementation of transport wide cc and this should be the graphic:

image

How can I enable the pcap dump to ensure that the rtcp packet is well formed?

lminiero:
An easy way to do that with the VideoRoom demo is adding the following snippet:

```
Janus.httpAPICall("http://localhost:7088/admin/" + janus.getSessionId() + "/" + sfutest.getId(), {
	verb: 'POST',
	withCredentials: false,
	body: {"janus": "start_text2pcap", "folder": "/tmp", "filename": "my-test2pcap-dump.txt", "transaction": "123", "admin_secret": "janusoverlord"},
	success: function(json) {
		Janus.vdebug(json);
	},
		Janus.error(textStatus + ": " + errorThrown);	// FIXME
	}
});
```

after the handle has been created, e.g., [here](https://github.com/meetecho/janus-gateway/blob/master/html/videoroomtest.js#L98). This will save the file called `my-test2pcap-dump.txt` to `/tmp`, so feel free to make this more dynamic in case you have other publishers in the room and want to save to different pcaps. You can then convert to pcap/pcapng with `text2pcap`:

    text2pcap -D -n -l 1 -i 17 -u 1000,2000 -t '%H:%M:%S.' /tmp/my-test2pcap-dump.txt /tmp/my-test2pcap-dump.pcapng

Notice that this requires the Admin API to be enabled, which is `admin_http = yes` in `janus.transport.http.cfg` (as we're sending an HTTP request to `http://localhost:7088/admin/` in the snippet).

murillo128:
And this is how it looks when I limit the bndwth to 800kbps with vmware custom network settings:

image

I will post the same graphics with janus now.

How can I disable the periodic I frame request?

murillo128:
This are with janus, no limit:

image

Something is not working correctly

lminiero:
Yep, this seems in line with what I was getting.

murillo128:
I got it working properly on my local setup, could you give it a try?

lminiero:
Will try and come back to you in a few minutes, thx!

lminiero:
Mh, looks like I'm still getting some sort of erratic behaviour, the average bitrate seems to be ~800/900kbps:
transport-wide-cc-bits_pt2

murillo128:
Could you double check? this is mine now:

image

lminiero:
> Could you double check?

I tried a couple of times yesterday, with and without the periodic PLI, and I always got the same result. I also double checked I was on the right code base and got your latest commit, and did a `make clean` before installing again, but nothing changed. I'll try from a clean branch and let you know.

lminiero:
@murillo128 started from scratch and now the results are indeed in line with yours!

transport-wide-cc-bits_pt3

I think the issue came from the fact that the branch I had was still based on the old `bundle-rtcpmux-only`, and so merging your new changes messed something in the code. Starting from master fixed it instead. As a side note, could you rebase your PR to master, as we eventually merged `bundle-rtcpmux-only` since you first published it? It may be as simple as doing it from the github PR page, but not sure about that.

That said, it looks like it's still not working as expected when simulcasting:

transport-wide-cc-bits_pt3_simulcast

It starts ramping up very quickly and then drops to ~800 again: watching the feed (which without an additional participant you can do via `newRemoteFeed(myid, "ciccio", "opus", "vp8")` on the JavaScript console) I do indeed only see the first two substreams available, while the higher definition is not available. Any thought on what may be different here, apart from the fact that the transport sequence numbers are "scattered" across multiple streams instead of just one?

lminiero:
@murillo128 thx! One more step that seems needed is hitting the Edit button here, and change the base to master too: I didn't want to do it myself as I don't know as it risks messing up your commits.

murillo128:
Not sure if I have done it properly..

lminiero:
Yep, that should be it! :+1:

lminiero:
Considering it works fine for regular usage, and that only simulcast streams seem to still not work as expected, I think we can merge already: we can sort the simulcast problem out later on, while we better integrate this in other plugins too for instance. Thanks again for your contribution!

PR #1121, opened 2018-01-03
plugins: fix async queue pop usage

tmatth:
Refs #1006


lminiero:
As you said in the original PR, it only returns NULL when the queue you pass to the method is NULL: we do set `messages` to NULL when deinitializing plugins, but before doing that we also join the message processing threads, which means it will never happen. Looks safe enough to me, so merging, thanks! (although not sure what we gained by removing that single check :smile: )

PR #1130, opened 2018-01-08
Npm es6 post processing

RouquinBlanc:
Hi Lorenzo,

This is a proposal to automatic build an es2015 compliant version of Janus automatically when using npm to manage dependencies.

I started using webpack and npm with es6 for managing/bundling the front-end code.

Ideally, we would like to have janus as an npm dependency, and then just import it like this:
```js
import adapter from 'webrtc-adapter';
import Janus from 'janus-gateway/html/janus';
```

Nothing that original, and I've seen many discussions on different pull requests.
So far, correct me if I'm wrong:

- we can generate as es6 compliant version of janus.js during compilation; but that's a bit too complicated when you're *only* interested into that one janus.es.js file
- there is a [janus-gateway](https://www.npmjs.com/package/janus-gateway) npm package available, but it is outdated (0.2.3) and does not provide all the different flavors of janus, including es6
- there is an npm directory in this repo, which uses rollup to generate an es6 compliant version of janus, but at the moment, 1) it is broken as rollup latest version require more configuration, and 2) it does not add anything when using npm, in fact it is not even called during install process.

This PR open the discussion about automatically generating the different flavors of Janus from `janus.js` during the npm installation. I'm not sure if this is the best way to go, but at least it solves the problem I explained earlier.

Right now, if I just:
```bash
npm install --save RouquinBlanc/janus-gateway#npm_es6
```

The resulting package installed contains both `janus.js` and `janus.es.js`, allowing me to import janus properly in our code.

This is mainly for discussion purpose; I'm not really an npm expert, but this seems much more acceptable than what we currently do, which is generating the es6 version of janus (manually, as we use an old version of janus), then copying the resulting file at the right location...

Note that this generation phase could happen in a separate git repo:

- Use janus-gateway untouched as a dependency
- Then generate the different flavors on the fly

This approach could allow to support several backward versions of janus, but that's the only advantage compared to integrating this directly here, in the main repo, so on the long term it's much less interesting...

Thanks & Best regards

### DIFF
--- a/npm/rollup.config.js
+++ b/npm/rollup.config.js
@@ -5,7 +5,9 @@ export default {
     name: 'Janus',
     input: 'module.js',
     output: {
-        strict: false
+        strict: false,
+        format: 'es',
+        file: '../html/janus.es.js',
     },
     plugins: [
         replace({
--- a/package.json
+++ b/package.json
@@ -5,7 +5,9 @@
   "main": "html/janus.js",
   "files": [
     "html/janus.js",
-    "html/janus.es.js"
+    "html/janus.es.js",
+    "npm/rollup.config.js",
+    "npm/module.js"
   ],
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -5,8 +5,7 @@
   "main": "html/janus.js",
   "files": [
     "html/janus.js",
-    "html/janus.nojquery.js",
-    "html/adapter.js"
+    "html/janus.es.js"
   ],
   "repository": {
     "type": "git",
@@ -16,6 +15,13 @@
   "license": "GPL-3.0",
   "bugs": {
     "url": "https://github.com/meetecho/janus-gateway/issues"
+  },
+  "scripts": {
+    "postinstall": "cd npm && rollup -c rollup.config.js -f es -o ../html/janus.es.js"
+  },
+  "devDependencies": {
+    "rollup": "^0.50.0",
+    "rollup-plugin-replace": "^2.0.0"
   },
   "homepage": "https://github.com/meetecho/janus-gateway#readme"
 }
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "scripts": {
     "postinstall": "cd npm && rollup -c rollup.config.js -f es -o ../html/janus.es.js"
   },
-  "devDependencies": {
+  "dependencies": {
     "rollup": "^0.50.0",
     "rollup-plugin-replace": "^2.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -6,8 +6,7 @@
   "files": [
     "html/janus.js",
     "html/janus.es.js",
-    "npm/rollup.config.js",
-    "npm/module.js"
+    "npm/*.js"
   ],
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "url": "https://github.com/meetecho/janus-gateway/issues"
   },
   "scripts": {
-    "postinstall": "cd npm && rollup -c rollup.config.js -f es -o ../html/janus.es.js"
+    "postinstall": "cd npm && rollup -c rollup.config.js"
   },
   "dependencies": {
     "rollup": "^0.50.0",


lminiero:
No idea what this commit does as I'm a JavaScript idiot and don't use this, but isn't this mandating the .es. thing? We have three different flavors of JS as targets when building janus.js:

```
  --enable-javascript-es-module
                          Generate an ECMAScript style module from janus.js
  --enable-javascript-umd-module
                          Generate an UMD style module from janus.js
  --enable-javascript-iife-module
                          Generate an IIFE style wrapper around janus.js
  --enable-javascript-common-js-module
                          Generate an CommonJS style module from janus.js
```

RouquinBlanc:
The flags you mention generate alternate flavors of janus.js by building the project. It means you have to:

- clone the repo,
- install all dependencies
- configure with the right flags
- make
- copy the generated es module in your project

While when you are coding a pure client (and as such don't need any backend), you only need the different flavors of janus.js, and that's it.

Without those changes, if you `npm install` janus-gateway repo, it will only retrieve the html/janus.js file as a dependency of your project. The changes I propose will, during the installation phase (more exactly in the postinstall phase, so right after installing the copied files), generate the different flavors of JS and put them in html/janus[.flavor].js automatically. Well, it only does the es module flavor right now, I wanted to discuss with you and possibly other people using the npm installation before adding the other flavors.

atoppi:
@RouquinBlanc so IIUC you are proposing a way to install the janus module via npm without using the `makefile`. I think this is a good idea to help not c-friendly devs getting access to js code.
I have some remarks though:
1) the `postinstall` script is always executed, even using the `makefile`
2) in general we need a configuration to avoid previous point and enable other kinds of modules

RouquinBlanc:
in point 1) do you mean that you would like to have only one script to be executed both from npm and make? This looks like a good idea indeed. I will have to think about it; the post install script is using `rollup`, which is a pure node tool, installed on the fly as a dependency while running the `npm install`. I will need to look at how the files are generated in the makefile version, and see if we can put them in common without too much pain!

And yes, for point 2) it would remain optional when building via makefile, I don't know if it is possible in npm to specify an option when installing, I will search.

atoppi:
@RouquinBlanc yes I was thinking about always demanding to `npm` scripts, so the bundles will be built by the `npm` tools, making easy to invoke only the bundle generation part.
As of now, the bundles are built with the following (extracted from `makefile`)
```
cd npm && npm install && touch node_modules/rollup/bin/rollup
cd npm && npm run rollup -- --o $(patsubst npm/%,%,$@) --f $*
```
So we are invoking `npm install` and then `npm run rollup` with specified flavor parameters (note that parameters passing is permitted with `--` syntax).
A simple approach would be to create a brand new script in `package.json` that handles the `install` and the `rollup`, or we could even overload the `install`.

RouquinBlanc:
Yes, indeed!

tgabi333:
i think the best approach would be to move out js codes into a shiny new repo. 

- The bound between JS codes and janus only is the API (ws or http transports).
- On npm the js library is outdated
- a new repo could handle all the flavors js codes
- a new repo's output could be (free) CDN delivered
- all the demos and tests should be there (which would be a great start for test automations)
- new js repo could be semantically versioned

lminiero:
It's not that simple. Often we add new features to plugins, that are somehow reflected in the demos. If this feature is in a pull request, how do you handle the demo part if it's another demo? Another PR there? Keep them up-to-date somehow? Looks overly complicated and something I for sure wouldn't want to bother with.

I've often been an advocate of putting the complex stuff (like nodejs stuff and the like) somewhere else, but the basic demos are here to stay. For anything more complex than that, I agree that a more "free" environment like another repo that doesn't need to be constrained by the structure we have here may be a better option.

RouquinBlanc:
Unfortunately I'm under water right now and cannot spend time on it... But
I really think we would benefit for an easy way to include a recent version
of the Janus js code from npm, any react project does that at the moment!


Le mer. 4 avr. 2018 à 16:20, Lorenzo Miniero <notifications@github.com> a
écrit :

lminiero:
Closing this for the time being, then. Feel free to propose a new pull request in the future, should you have some time to work on it. Thanks again for the discussion!

PR #1133, opened 2018-01-15
SRTP support in Streaming mountpoints and RTP forwarders

lminiero:
As the title says, this patch adds SRTP support to both RTP-based Streaming mountpoints and RTP forwarders (VideoRoom only, AudioBridge will follow). This is particularly useful in those scenarios where you might need to relay RTP externally, but don't like the idea of doing that without encryption.

Configuring the feature in both cases is quite simple, and only requires two properties:
1. how many bits to use for authentication tags (32 or 80)
2. the key to use, as a SDES-compliant bade64 encoded string

Of course, for Streaming mountpoints the key will refer to what you'll use to decrypt the stream, while for RTP forwarders to the key to use to encrypt. This means that, if you plan to pass media from Janus to Janus, you'll configure both mountpoint and forwarder with the same exact info.

For streaming mountpoints, the two properties are called, in both configuration file and API calls, `srtpsuite` (for tags size) and `srtpcrypto` (for the key), e.g.:

```
[srtp-sample]
type = rtp
id = 11
description = SRTP sample
audio = yes
video = yes
audioport = 5102
audiopt = 111
audiortpmap = opus/48000/2
videoport = 5104
videopt = 100
videortpmap = VP8/90000
srtpsuite = 32
srtpcrypto = WbTBosdVUZqEb6Htqhn+m3z7wUh4RJVR8nE15GbN
secret = adminpwd
```

In the VideoRoom, you can create an SRTP forwarder exactly like you create RTP forwarders, with the difference that you specify `srtp_suite` and `srtp_crypto` too: the meanings are the same as the ones introduced above (notice the underscores are there only to comply with the syntax of other VideoRoom API properties). Here's a simple example to create one from JavaScript:

```
sfutest.send({message:
   {
       request: "rtp_forward",
       room: 1234,
       publisher_id: myid,
       host: "127.0.0.1",
       audio_port: 5102,
       video_port: 5104,
       srtp_suite: 32,
       srtp_crypto: "WbTBosdVUZqEb6Htqhn+m3z7wUh4RJVR8nE15GbN",
       secret: "adminpwd"
    },
    success: function(json) {
       console.log("Got response:", JSON.stringify(json))
    }
});
```

Tested briefly and seems to be working as expected. Notice, though, that with SRTP you'll have to be extra careful in not involving streams that may cause SRTP decryption/encryption errors, e.g., re-sending packets that may have the same SSRC/seq/ts.

## Next steps
This preliminary effort is basically done, but there are a couple of things I'd like to add before merging:
1. add the same SRTP support to RTP forwarding in the AudioBridge too
2. most importantly, allow for a way to SRTP encrypt once to multiple destinations.
In fact, the way it works now it assumes that if you want to SRTP forward to Server1 and Server2, you'll have to create two different RTP forwarders: even if the keys are the same and so are the packets, this will though result in two different SRTP encrypts. It would be much better to only encrypt once, and then relay this already encrypted packet to multiple destinations.

I'll have to think about the best way to formalize this considering the existing RTP forwarder constraints.

### DIFF
--- a/conf/janus.plugin.audiobridge.cfg.sample
+++ b/conf/janus.plugin.audiobridge.cfg.sample
@@ -15,11 +15,14 @@
 ;     The following lines are only needed if you want the mixed audio
 ;     to be automatically forwarded via plain RTP to an external component
 ;     (e.g., an ffmpeg script, or a gstreamer pipeline) for processing
+;     By default plain RTP is used, SRTP must be configured if needed
 ; rtp_forward_id = numeric RTP forwarder ID for referencing it via API (optional: random ID used if missing)
 ; rtp_forward_host = host address to forward RTP packets of mixed audio to
 ; rtp_forward_port = port to forward RTP packets of mixed audio to
 ; rtp_forward_ssrc = SSRC to use to use when streaming (optional: stream_id used if missing)
 ; rtp_forward_ptype = payload type to use when streaming (optional: 100 used if missing)
+; rtp_forward_srtp_suite = length of authentication tag (32 or 80)
+; rtp_forward_srtp_crypto = key to use as crypto (base64 encoded key as in SDES)
 ; rtp_forward_always_on = true|false, whether silence should be forwarded when the room is empty (optional: false used if missing)
 
 [general]
--- a/conf/janus.plugin.streaming.cfg.sample.in
+++ b/conf/janus.plugin.streaming.cfg.sample.in
@@ -40,6 +40,15 @@
 ; databuffermsg = yes|no (whether the plugin should store the latest
 ;		message and send it immediately for new viewers)
 ;
+; In case you want to use SRTP for your RTP-based mountpoint, you'll need
+; to configure the SRTP-related properties as well, namely the suite to
+; use for hashing (32 or 80) and the crypto information for decrypting
+; the stream (as a base64 encoded string the way SDES does it). Notice
+; that with SRTP involved you'll have to pay extra attention to what you
+; feed the mountpoint, as you may risk getting SRTP decrypt errors:
+; srtpsuite = 32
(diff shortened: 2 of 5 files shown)


lminiero:
The AudioBridge plugin now supports SRTP forwarding too. Time to tackle optimizations now...

lminiero:
This new commit implements the optimization I mentioned before. Nothing changes API wise, but internally we optimize the usage of SRTP context. Specifically, when creating a new RTP forwarder for a publisher that needs SRTP, we check if we already have one with the same:
1. media type (e.g., audio or video)
2. target SSRC
3. target payload tye
If we don't, we create a new SSRC context and associate it with the new RTP forwarder; if we do, we pick the existing SSRC context and associate it with the new RTP forwarder too. When forwarding incoming packets from the publisher, we can then ensure we only encrypt once per SRTP context, no matter how many recipients share it.

Tested it briefly and it seems to be working as expected. At this point, I consider the patch ready, and will just wait for feedback to decide whether there's anything that needs fixing or we can merge.

lminiero:
Merging.

saket424:
I tested this feature using ffmpeg and the streaming plugin. Works!

[srtpsample]
type = rtp
id = 18
description = srtp test
audio = no
video = yes
videoport = 2424
videopt = 96
videortpmap = H264/90000
videofmtp = packetization-mode=1
videobufferkf = yes
srtpsuite = 32
srtpcrypto = WbTBosdVUZqEb6Htqhn+m3z7wUh4RJVR8nE15GbN

ffmpeg -i rtsp://184.72.239.149/vod/mp4:BigBuckBunny_175k.mov -an -c:v copy -flags global_header -bsf dump_extra -f rtp -srtp_out_suite SRTP_AES128_CM_HMAC_SHA1_32 -srtp_out_params "WbTBosdVUZqEb6Htqhn+m3z7wUh4RJVR8nE15GbN" srtp://127.0.0.1:2424

PR #1134, opened 2018-01-19
Support for RFC4588 retransmissions (rtx/90000)

lminiero:
As the title says, this PR adds support for [RFC4588](https://tools.ietf.org/html/rfc4588) retransmissions, that is the RTP retransmission payload format that Chrome negotiates by default.

You may wander why this was done, considering we support NACKs and do retransmissions already. In theory, it wasn't strictly needed, as Firefox doesn't support them and, when refused, Chrome does plain retransmissions as well (blindly re-sending the previously encrypted SRTP packet). Anyway, just re-sending the SRTP packet as it was originally prepared isn't the proper way: there were reasons for standardizing such a format (you can read more in the RFC), and besides support for it is mandated in the [RTP usage draft](https://tools.ietf.org/html/draft-ietf-rtcweb-rtp-usage-26#section-6.1) for WebRTC, so I went on and did it. Eventually, although not trivial, it turned out to be less work than I expected, and it fits nicely in the way the core works.

Nothing changes in how you use Janus as a user or developer. This is negotiated automatically via SDP at the core level, so if the browser supports it, Janus will use it: Janus will also try and negotiate it by default, and fall back to the old way of doing retransmissions if the browser doesn't support it instead.

Tested briefly and seems to work. I encourage you to play with this PR to make sure you don't notice any regression in your application, as I plan to merge relatively soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ or on the command line:
 	-T, --ice-tcp                 Whether to enable ICE-TCP or not (warning: only
                                   works with ICE Lite)
                                   (default=off)
-	-R, --no-rfc-4588             Whether to disable RFC4588 retransmissions
+	-R, --rfc-4588                Whether to enable RFC4588 retransmissions
                                   support or not  (default=off)
 	-q, --max-nack-queue=number   Maximum size of the NACK queue (in ms) per user
                                   for retransmissions
--- a/README.md
+++ b/README.md
@@ -348,6 +348,8 @@ or on the command line:
 	-T, --ice-tcp                 Whether to enable ICE-TCP or not (warning: only
                                   works with ICE Lite)
                                   (default=off)
+	-R, --no-rfc-4588             Whether to disable RFC4588 retransmissions
+                                  support or not  (default=off)
 	-q, --max-nack-queue=number   Maximum size of the NACK queue (in ms) per user
                                   for retransmissions
 	-t, --no-media-timer=number   Time (in s) that should pass with no media
--- a/conf/janus.cfg.sample.in
+++ b/conf/janus.cfg.sample.in
@@ -67,7 +67,8 @@ cert_key = @certdir@/mycert.key
 
 
 ; Media-related stuff: you can configure whether if you want
-; to enable IPv6 support (still WIP, so handle with care), the maximum size
+; to enable IPv6 support (still WIP, so handle with care), if RFC4588
+; support should be negotiated or not (they are by default), the maximum size
 ; of the NACK queue (in milliseconds, defaults to 500ms) for retransmissions, the
 ; range of ports to use for RTP and RTCP (by default, no range is envisaged), the
 ; starting MTU for DTLS (1472 by default, it adapts automatically), and
@@ -77,6 +78,7 @@ cert_key = @certdir@/mycert.key
 [media]
 ;ipv6 = true
 ;max_nack_queue = 500
+;rfc_4588 = no
 ;rtp_port_range = 20000-40000
 ;dtls_mtu = 1200
 ;no_media_timer = 1
--- a/conf/janus.cfg.sample.in
+++ b/conf/janus.cfg.sample.in
@@ -68,7 +68,7 @@ cert_key = @certdir@/mycert.key
 
 ; Media-related stuff: you can configure whether if you want
 ; to enable IPv6 support (still WIP, so handle with care), if RFC4588
-; support should be negotiated or not (they are by default), the maximum size
+; support should be negotiated or not (off by default), the maximum size
 ; of the NACK queue (in milliseconds, defaults to 500ms) for retransmissions, the
 ; range of ports to use for RTP and RTCP (by default, no range is envisaged), the
 ; starting MTU for DTLS (1472 by default, it adapts automatically), and
@@ -78,7 +78,7 @@ cert_key = @certdir@/mycert.key
 [media]
 ;ipv6 = true
 ;max_nack_queue = 500
-;rfc_4588 = no
+;rfc_4588 = yes
 ;rtp_port_range = 20000-40000
 ;dtls_mtu = 1200
 ;no_media_timer = 1
--- a/html/janus.js
+++ b/html/janus.js
@@ -2409,7 +2409,7 @@ function Janus(gatewayCallbacks) {
 		// (based on https://gist.github.com/ggarber/a19b4c33510028b9c657)
 		var lines = sdp.split("\r\n");
 		var video = false;
-		var ssrc = [ -1 ], ssrc_fid = -1;
+		var ssrc = [ -1 ], ssrc_fid = [ -1 ];
 		var cname = null, msid = null, mslabel = null, label = null;
 		var insertAt = -1;
 		for(var i=0; i<lines.length; i++) {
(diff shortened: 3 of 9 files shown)


lminiero:
One thing I should clarify is that, with this patch, plugins might receive the same packet twice, something that would never happen before. In fact, with the old approach browsers would simply send us again the same SRTP packet: if we had already received it (e.g., retransmission happened twice, or we got the packet after we NACKed it and so got it twice), decryption at the second attempt would fail and the duplicate packet would be dropped. In this case, decryption will NOT fail, as with RFC4588 there's a different SSRC/sequence number space used just for that, and we only "rebuild" the original packet after decrypting with SRTP. We do know it's a retransmitted packet, at this stage, but we don't know if we already received the same packet before.

This different SSRC/sequence number space is actually a good thing for different reasons (e.g., avoiding to confuse the RTCP stats with packets that we don't know if they're retransmitted or simply very delayed), but, again, might mean your plugin can get the same packet twice. There are a few cases where this might cause problems, e.g., recordings: for that, I plan to update the post-processor to skip duplicates when re-ordering packets. For other scenarios (e.g., RTP forwarders, legacy interaction with SIP, etc.) we may have to just assume the recipient is smart enough to remove duplicates.

In upcoming commits I'll work on a way to detect these duplicates when receiving packets in the first place. e.g., using the NACK mechanism we have in place where we're already tracking sequence numbers we did and did not receive. As a side note, for the sake of backwards compatibility I'll also add a flag (command line and config) to turn this new retransmission mechanism off for those who don't want it for some reason.

lminiero:
I've reconsidered the default value, in light of the best backwards compatibility. Now, by default, RFC4588 retransmissions are NOT negotiated: if you want them, you need to enable them explicitly, either by setting `rfc_r488=yes` in `janus.cfg` or by passing `-R` on the command line.

trankkilou:
Hi Lorenzo,

I add your PR and activate it. 

I ve test RTP streaming H264/ OPUS from FFMPEG + Chrome and it seems to work

But how to check it is use when I got NACK ?

lminiero:
> But how to check it is use when I got NACK ?

You can capture traffic with Wireshark, and check if you ever see packets with the rtx SSRCs (either Janus' or the peer's).

trankkilou:
Do I have to decode the trafic as « RTP » ?

I can see RTP sequence number but no hole inside even with NACK, no rtf SSRCs

Can you give me the right Wireshark filter ?

lminiero:
> Do I have to decode the trafic as « RTP » ?

Yep.

trankkilou:
I can’t help if I don’t know what to looking at

I have a filter on rtp type (rtp.P_type-100) -> to match my video packets

Janus said I got 200 NAck @ 18:11:10
[Thu Jan 25 18:11:04 2018] [8326289250338868] Retransmitted 6 packets due to NACK (video stream #0)
[Thu Jan 25 18:11:10 2018] [8326289250338868] Retransmitted 199 packets due to NACK (video stream #0)

I have check every packets 1 by 1 between 18:09 and 18:12 and I see nothing obvious. No hole in seq number (each are incremented by one)


J--

amnonbb:
Tried to test it with ffmpeg stream and I got blink video.
In chrome://webrtc-internals/ i see a lot packet lost.
The same result with rfc_4588 = no option.
When I back to latest master branch the stream is ok.

lminiero:
@amnonbb apologies for the late reply, I was abroad for an event. Can you elaborate on your testing scenario? It's unclear to me what you tried to simulate: was ffmpeg introducing losses on purpose? Nothing changes in the regular streaming with this patch: only how we do retransmissions.

amnonbb:
I test actually my production working scenario where i send rtp stream with ffmpeg.
On client i see a lot of losses
https://pastebin.com/d13KqkXh
https://pastebin.com/7xhRipPW
Now it's maybe really a lot of losses on client, but when i back to git master all losses magically disappeared.

lminiero:
Is this also happening with this branch when you disable RFC4588 support? Or were your tests with it already disabled, which is the default? What about when it's enabled? Again, it's weird that you see the losses here and not in master: the delivery of packets has not been touched.

amnonbb:
The same result with rfc_4588 = yes/no option.
Today i tried to do rtp forward from videoroom to stream plugins.
I got the same problem.
https://pastebin.com/yrahRgmg
https://pastebin.com/mz4xRk27
i'll try today to exclude internet problem and deploy this branch on local network.

amnonbb:
I did test (ffmpeg stream and videoroom forward w/o rfc_4588 = option)
 in local network and got the same result - a lot of losses. :(
I tried to look into log level 7 but did not find something to understand - who and why decided that client got packetloss

lminiero:
Looks like some aligns I did with master recently broke something, as the EchoTest doesn't work either for instance... Good chances are this is why it's not working as expected for you. I'll look into it.

lminiero:
Fixed.

amnonbb:
Look like issue with a lot losses is gone.

lminiero:
> Look like issue with a lot losses is gone.

Yeah so I guess your issue was actually caused by the other problem I fixed. Basically we were not overwriting the SSRC because it looked like the `retransmission` flag was set when it really wasn't (in a recent change there were some `g_malloc0` changed to `g_malloc`, so we have to initialize ALL properties and we weren't doing it here).

amnonbb:
I see throw admin api in handle info flags:
"rfc4588-rtx": true
on video mountpoints - true
but on all audio mountpoints the flags is
"rfc4588-rtx": false
Does it mean retransmission not supported or it's going old way?

lminiero:
We only do retransmissions (old way or not) when the negotiation uses NACKs for the media stream. Right now, browsers only negotiate them for video, which means for audio there are none.

amnonbb:
When i restart janus with live srtp stream (to srtp mountpoint) 
i get this error
https://pastebin.com/6NeXvAsP
To resolve it i must to restart stream.

lminiero:
@amnonbb this is unrelated, as it's part of the new SRTP support in the Streaming plugin.

lminiero:
Now that we added the last missing piece (detection of RTP duplicates) we can merge.

trankkilou:
Ok I have understood how to check this is working.
I started filtering on ssrc id, as the ssrc id for retransmitting is different but it is easier to filter on "rtp type" (97)
so the filter in Wireshark is: rtp.p_type == 97
OFC you need to decode as RTP packets first.
After what it is simple to check in real time with janus log. It matchs perfectly
- "Sent NACKs for X missing packets (video stream #0)" means janus server ask for retransmit X packets
- "Retransmitted 1 packets due to NACK (video stream #0)" means Janus server is retransmitting X packets 
I tested both in H264 and VP8 and wifi (natural packet dropping :) )
Honestly, from a user perspective, I didn't notice difference with old mechanism. So I am a little disappointed :(

lminiero:
> I started filtering on ssrc id, as the ssrc id for retransmitting is different but it is easier to filter on "rtp type" (97)
> so the filter in Wireshark is: rtp.p_type == 97

It's not always 97, you have to see what's negotiated in the SDP.

> Honestly, from a user perspective, I didn't notice difference with old mechanism. So I am a little disappointed :(

Functionally it's exactly the same thing. Anyway, this new approach does not mess with RTCP statistics (before, retransmissions could be detected as delayed packets instead of retransmissions, meaning stuff like RTT could be wildly miscaldulated), which means it helps browsers and Janus do the right estimations when looking at stats to decide how to react to network conditions. So you may not notice anything (as a user you're not supposed to, it's low level stuff), but it's better.

lminiero:
> But for janus, what can he do « more » (more than retransmission )regarding « network conditions » ?

At the moment, nothing, but it does mean more precise RTCP info in both Admin API and Event Handlers.

Issue #756, opened 2017-01-25
No check after malloc in videoroom plugin


chadfurman:
https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_videoroom.c#L2982

If there's a memory error, this will segfault

lminiero:
I had this discussion with a user on this some time ago. I had checks on the return value from allocations all over the place, and he suggested most were unneeded, as if it really returns NULL it's an out of memory error, which means the application will crash anyway. As such, in some new code I didn't add any controls over that. Do you believe we should have those checks instead?

chadfurman:
I was taught by my professor that if you don't check the return value of malloc, the world will end.

If you _always_ check the return value of malloc, the program... might not... crash.  It just will stop behaving as expected....

lminiero:
That's what I knew too and what I argued, but his argumentations convinced me. Searching around it looks like it's still an active discussion among other developers as well, e.g., those mentioning overcommits and stuff like that as a rationale behind the fact you may not need to check those values at all.

Anyway, I guess I agree checks probably don't harm, so I'll slowly add them back in when I have some time to check the whole code (help on that is appreciated of course!). Besides, I'll have to make sure the if(not) contents are coherent in branches like the reference counter: if there we need an unref, for instance, that in master isn't there, and it just merges cleanly, I'll also have to look into all the changes there.

lminiero:
In order to avoid problems like those, I'll only restore those checks in the reference counters branch, and leave master as it is. This way, when reference counters are merged eventually, we'll have that sorted out too.

chadfurman:
That makes sense. I just merged in reference counters to our codebase this week. Still trying to make heads or tails of it. Looks like I need to rethink how I destroy all my data structures. I did get the conflicts resolved. Regardless, this would be a Google groups discussion separate from this ticket haha

lminiero:
> Still trying to make heads or tails of it. Looks like I need to rethink how I destroy all my data structures.

Yeah, it is indeed a big change in how memory is managed, which is one of the reasons it's taking a while before this is merged. I want to be sure we're not introducing leaks, rather than solving them, and making sure we ref/unref properly in all cases. For this testing from multiple parts can indeed help, so thanks for giving it a try.

That said, I added the `g_malloc0` checks. I'll probably have to do the same for stuff like `g_strdup()` and `json_object()`, among others, but I'll do that some other time.

uxmaster:
[...]

PR #1137, opened 2018-01-22
Memory allocation improvements

douglascaetano:
Hi! I've made some general fixes and improvements to Janus code. The changes were the following:

- wrong usage of getsockname();
- unneeded or redundant use of g_malloc0() and memset() functions, either because they were used together, or the allocated memory is for a struct that has all members set after the allocation, or a memcpy() is done after the allocation;
- unneeded memory allocations checking, as GLib's g_malloc() and others abort the program on failure;
- some unneeded run-time checking of pointers to decide which of g_malloc() and g_realloc() functions should be used - g_realloc(NULL, size) is equivalent to g_malloc(size);
- allocation bug on transports/janus_http.c where calloc() were substituted incorrectly with a g_malloc0();
- a possible memory leak at janus_text2pcap_create();
- an incorrect comment header for janus_plugin_result_destroy(); and
- removed castings from memory allocations.

Please refer to the commits' comments for more details. Thanks!

### DIFF
--- a/config.c
+++ b/config.c
@@ -188,10 +188,6 @@ error:
 
 janus_config *janus_config_create(const char *name) {
 	janus_config *jc = g_malloc0(sizeof(janus_config));
-	if(jc == NULL) {
-		JANUS_LOG(LOG_FATAL, "Memory error!\n");
-		return NULL;
-	}
 	if(name != NULL) {
 		jc->name = g_strdup(name);
 	}
@@ -258,10 +254,6 @@ janus_config_category *janus_config_add_category(janus_config *config, const cha
 		return c;
 	}
 	c = g_malloc0(sizeof(janus_config_category));
-	if(c == NULL) {
-		JANUS_LOG(LOG_FATAL, "Memory error!\n");
-		return NULL;
-	}
 	c->name = g_strdup(category);
 	config->categories = g_list_append(config->categories, c);
 	return c;
@@ -293,10 +285,6 @@ janus_config_item *janus_config_add_item(janus_config *config, const char *categ
 	if(item == NULL) {
 		/* Create it */
 		item = g_malloc0(sizeof(janus_config_item));
-		if(item == NULL) {
-			JANUS_LOG(LOG_FATAL, "Memory error!\n");
-			return NULL;
-		}
 		item->name = g_strdup(name);
 		item->value = g_strdup(value);
 		if(c != NULL) {
--- a/dtls.c
+++ b/dtls.c
@@ -433,10 +433,6 @@ janus_dtls_srtp *janus_dtls_srtp_create(void *ice_component, janus_dtls_role rol
 		return NULL;
 	}
 	janus_dtls_srtp *dtls = g_malloc0(sizeof(janus_dtls_srtp));
-	if(dtls == NULL) {
-		JANUS_LOG(LOG_FATAL, "Memory error!\n");
-		return NULL;
-	}
 	/* Create SSL context, at last */
 	dtls->srtp_valid = 0;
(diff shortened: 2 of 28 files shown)


lminiero:
Thanks for this, it's hugely appreciated! The unneeded checks for the return value of the`g_malloc` stuff is something I started to get rid of already and then never completed (there was some related discussion in #756 and other older issues), so thanks for taking care of that. I'm more hesitant when it comes to replacing some `g_malloc0` with `g_malloc`, mostly because, while it's true that in those cases we may be setting all properties, in the future we might add new properties we don't initialize right away, and that would lead to trouble... is the extra load of the implied memset there really releant? If so, I guess this means we'd just have to be more careful when making changes.

Due to the size of the patch, I'll need some time to review this, also because I'll be a bit busy these days. Please let me know if you add anything to the patch in the meanwhile, and thanks again!

ivanovaleksey:
Maybe it would be better to split single PR into multiple? 
It would make it easier to review and discuss changes and you can merge them separately. 
Since commits are already nicely separated from each other it wouldn't hard to create multuple PRs.

douglascaetano:
@lminiero, my personal rule is to never use `calloc()` (or `g_malloc0()` in this case) when not needed. I think that structure initialization must be done carefully always and should not simply rely on this zero initialization. There are other good arguments on this side in the wild. As for the performance, I did some simple tests here on `g_malloc()` vs `g_malloc0()`:

    ./g_malloc  49.28s user 0.00s system 99% cpu 49.280 total
    ./g_malloc  49.08s user 0.00s system 99% cpu 49.085 total
    ./g_malloc  49.10s user 0.00s system 99% cpu 49.109 total

    ./g_malloc0  84.60s user 0.00s system 99% cpu 1:24.61 total
    ./g_malloc0  84.66s user 0.00s system 99% cpu 1:24.66 total
    ./g_malloc0  84.64s user 0.00s system 99% cpu 1:24.65 total

The program allocates and frees 1500 bytes of memory in a loop for 1 billion times. By these tests, `g_malloc0()` is more than 70% slower than `g_malloc()`. The code: https://gist.github.com/douglascaetano/d4cc5715e887263de944e104d1b311a3

I think that the substitution should be made at least on hot paths and in the obvious places (e.g. small structures like the `janus_*_message` ones).

Anyway, I made the substitution only on places where I felt confident to do so. I left several other places intact because I had yet not enough knowledge of those parts of the code.

@ivanovaleksey, I could do that, no problems, if you guys think this is easier.

lminiero:
Thanks for the clarification, and it does make sense indeed.

Separating the PRs per type would indeed help me review the code, thanks! Of course, there's no need for a PR per commit: it could be a PR for all `g_malloc0` replacements, a PR for unneeded checks, a PR for typos/fixes, or whatever differentiation you think makes sense (and is easier for you, I don't want to make it more trouble for you...). Thanks again for this huge effort!

douglascaetano:
@lminiero, I tried to separate commits, but the best I could do was this, because the commits here depend on each other (and refer to more or less the same modifications). It would be a bit time consuming to make commits independent and, in the end, redundant work, because they would have to be modified later to merge again. I hope this helps you at least a bit. :)

lminiero:
@douglascaetano just did a review and it looks good to me! Please let me know if you planned to kee on working on this PR for some other fix, or if this can be merged as is :+1:

douglascaetano:
No plans for now. Do you have any suggestions on things to improve in this matter? Otherwise I think it's good to merge as is.

lminiero:
Nope, I was just wondering if you were still looking for other things to fix of the similar kind... merging then, thanks again!

PR #1138, opened 2018-01-23
Memory allocation bugs

douglascaetano:
Hi!

As asked on PR #1137, I detached some commits and made this new PR. The following commits fix some problems with memory usage and allocation. Please refer to each commits message for more details.

Thanks!

### DIFF
--- a/text2pcap.c
+++ b/text2pcap.c
@@ -31,6 +31,7 @@
  * \ref core
  */
 
+#include <errno.h>
 #include <sys/time.h>
  
 #include "text2pcap.h"
--- a/text2pcap.c
+++ b/text2pcap.c
@@ -80,8 +80,6 @@ janus_text2pcap *janus_text2pcap_create(const char *dir, const char *filename, i
 
 	/* Try opening the file now */
 	f = fopen(fname, "ab");
-	if (f == NULL)
-		f = fopen(fname, "wb");
 	if (f == NULL) {
 		JANUS_LOG(LOG_ERR, "fopen(%s) error: %d\n", fname, errno);
 		g_free(fname);
--- a/text2pcap.c
+++ b/text2pcap.c
@@ -50,51 +50,51 @@ const char *janus_text2pcap_packet_string(janus_text2pcap_packet type) {
 }
 
 janus_text2pcap *janus_text2pcap_create(const char *dir, const char *filename, int truncate) {
+	janus_text2pcap *tp;
+	char newname[1024];
+	char *fname;
+	FILE *f;
+
 	if(truncate < 0)
 		return NULL;
-	/* Create the text2pcap instance */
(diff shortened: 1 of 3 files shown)


text2pcap.c
```diff
@@ -50,57 +50,60 @@ const char *janus_text2pcap_packet_string(janus_text2pcap_packet type) {
 }
 
 janus_text2pcap *janus_text2pcap_create(const char *dir, const char *filename, int truncate) {
+	janus_text2pcap *tp;
+	char newname[1024];
+	char *fname;
+	FILE *f;
```
lminiero: All of the above pointers should be initialized to NULL.

lminiero:
@douglascaetano now that we merged #1137 I'll wait for the tweaks to this patch (I see there now is a conflict though?) to merge this too.

text2pcap.c
douglascaetano: Actually, there is no need for it. If you follow the code, you'll see:

- `*tp` will be initialized with the allocation right before being used for the first time;
- `*f` will be initialized in both branches of the first `if`, also before the first use; and
- `*fname` will be initialized whenever `*f` is not set to `NULL` in the previous assignments, and when `*f` is in fact `NULL`, the code that uses `*fname` will never be reached.

But I admit the code is a bit confusing, so a refactoring is on the way. :)

douglascaetano:
@lminiero, do you mind if I do a rebase and a forced push so git history is kept cleaner?

lminiero:
> @lminiero, do you mind if I do a rebase and a forced push so git history is kept cleaner?

No problem at all, thx!

douglascaetano:
Done! I took off some complexity on that function. Hope it's easier to understand now! :grinning:

douglascaetano:
By the way, why do you try `fopen()` twice? AFAIK, if an error occurs with "a" mode, the same would occur for "w" mode, wouldn't?

lminiero:
> By the way, why do you try fopen() twice? AFAIK, if an error occurs with "a" mode, the same would occur for "w" mode, wouldn't?

Not sure, but the plan was to try appending first, and if that failed create the file from scratch. Not sure if trying to append to a non-existing file creates a new file nevertheless, in which case the second fopen is indeed useless.

douglascaetano:
I see. I asked in case you had some trouble before. Indeed, according to the [`fopen()` man page](http://man7.org/linux/man-pages/man3/fopen.3.html), the file is created if it does not exist in either `w` or `a` modes, so it is safe in this regard. If there are no objections, can I add a commit for this?

lminiero:
Sure, makes sense to add it as part of the fixes, thx!

douglascaetano:
Done! :blush:

lminiero:
Thx, merging then!

PR #1141, opened 2018-01-24
Add full-trickle support to Janus local candidates

lminiero:
As those more familiar with ICE and WebRTC may know, Janus does half-trickling: this means that it accepts [trickle](https://webrtchacks.com/trickle-ice/) candidates from peers, but always puts its own candidates in the SDP instead. This means that we need to wait for all Janus candidates to have been gathered, before Janus can send its SDP offer or answer. 

In general, this is not an issue at all. Janus is not like a browser on a desktop, which may have a ton of interfaces, STUN and TURN configured, and a lot of candidates to gather and shoot around. Most of the times, Janus is on a public address with maybe 1-2 interfaces, which means the gathering process is typically very quick. As a result, there's no impact at all for doing half-trickling, which greatly simplifies things. That said, if you're in a situation where you'd benefit from full trickling instead, then this patch adds support for it.

From a Janus perspective, you enable it (it's disabled by default) either by setting `full_trickle=yes` in `janus.cfg`, or by passing `-f` on the command line. When enabled, Janus will send "trickle" candidates as events, and in the same format as users currently send their own trickle candidates to Janus instead:

```
{
   "janus": "trickle",
   "session_id": 103912560900294,
   "sender": 1030472959325354,
   "candidate": {
      "sdpMLineIndex": 0,
      "candidate": "candidate:1 1 udp 2013266431 192.168.1.80 35584 typ host"
   }
}
```

The last candidate is notified with the usual `completed:true` just as the [documentation](https://janus.conf.meetecho.com/docs/rest) explains users do:

```
{
   "janus": "trickle",
   "session_id": 103912560900294,
   "sender": 1030472959325354,
   "candidate": {
      "completed": true
   }
}
```

We already modified `janus.js` to handle incoming trickle candidates automatically, so if you want to use full-trickling, you just need to make sure you use the updated version. If you're not using `janus.js` but are handling the WebRTC part yourself, you'll need to make sure you're prepared to handle incoming trickle candidates and enforce them the way regular WebRTC applications do.

As usual, I've tested briefly and it seems to be working fine, but of course feedback is more than welcome! Nothing should change if you don't care about full-trickling and want to keep on doing things as Janus always did (half-trickling is still the default), but I encourage you to test anyway, as some of the code there changed and so I want to be 100% sure there are no regressions (don't say I didn't warn you in case...).

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,8 @@ or on the command line:
                                   (experimental)  (default=off)
 	-l, --libnice-debug           Whether to enable libnice debugging or not
                                   (default=off)
+	-f, --full-trickle            Do full-trickle instead of half-trickle
+                                  (default=off)
 	-I, --ice-lite                Whether to enable the ICE Lite mode or not
                                   (default=off)
 	-T, --ice-tcp                 Whether to enable ICE-TCP or not (warning: only
--- a/conf/janus.cfg.sample.in
+++ b/conf/janus.cfg.sample.in
@@ -90,7 +90,10 @@ cert_key = @certdir@/mycert.key
 ; browsers, but in the gathering of relay candidates by Janus itself,
 ; e.g., if you want to limit the ports used by a Janus instance on a
 ; private machine. Furthermore, you can choose whether Janus should be
-; configured to work in ICE-Lite mode (by default it doesn't). Finally,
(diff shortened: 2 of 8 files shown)


lminiero:

PR #1142, opened 2018-01-24
Stop using an unlimited thread pool for all requests

lminiero:
The way the core works right now, we have a thread pool for processing requests, no matter the transport that originated them. It is a Glib thread pool with an "unlimited" size, meaning we don't put any boundary on it: each new request that we receive (create, destroy, trickle, message, Admin API messages, etc.) is handled by one of the threads in the pool, as a task. While this is in general not much of an issue, it can be if you have a lot of users generating traffic, as each user may for instance send a ton of trickle messages, and in case more threads are needed they'll be spawned (and re-used by Glib when possible), which might be overkill for small requests that might be better handled more synchronously. Limiting the size of the thread pool is not an option, as it might cause scenarios where all threads are kept busy by longer operations, causing unacceptable delays in serving quick requests.

This patch changes this approach, and envisages a "master" thread for handling requests instead, pretty much as we do in plugins. Whenever we get a new request from a transport, we add it to a queue; this queue is accessed by the above mentioned thread; then the thread checks what kind of request it is, and only spawns a thread from the pool as before if it is a "message" for a plugin, and otherwise handles it directly. The reason for this differentiation is that message for plugins are the only ones that can actually take more time, and so the only ones that really deserve some asynchronous management: all others are meant to be much quicker, and requests that can be handled in a more synchronous way and in sequence by the master thread.

I tested this briefly and it seems to be working as expected, with no noticeable delay or regression. That said, if you have any feedback on the above (e.g., if you think I'm making no sense) please chime in and let's discuss if there are other approaches that might work better instead. I of course also encourage you to test this as soon as possible, since if this works fine for most I plan to merge soon considering it might provide considerable improvements in performance (that's why I started working on it in the first place).

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -363,7 +363,9 @@ static janus_transport_callbacks janus_handler_transport =
 		.events_is_enabled = janus_events_is_enabled,
 		.notify_event = janus_transport_notify_event,
 	};
-GThreadPool *tasks = NULL;
+static GAsyncQueue *requests = NULL;
+static janus_request exit_message;
+static GThreadPool *tasks = NULL;
 void janus_transport_task(gpointer data, gpointer user_data);
 ///@}
 
@@ -606,7 +608,7 @@ janus_request *janus_request_new(janus_transport *transport, void *instance, voi
 }
 
 void janus_request_destroy(janus_request *request) {
-	if(request == NULL)
+	if(request == NULL || request == &exit_message)
 		return;
 	request->transport = NULL;
 	request->instance = NULL;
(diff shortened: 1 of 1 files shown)


uxmaster:
Did you measure improvements in performance?

lminiero:
> Did you measure improvements in performance?

Didn't do any test, I just wrote this... which is one of the reasons why I've shared this with the community, hoping you guys will help us assess whether the patch was worth the effort or not. I can't do everything myself :wink:

atoppi:
Looks good to me.

lminiero:
Merging.

Globik:
Any chance for the janus gateway working properly in multiprocess environment?

Globik:
The kore itself therotically can be a part of a task under the main programm based on a web-framework. No need for a transport layer.

lminiero:
No plan to do multi process, multi thread is fine.

Il 04 mar 2018 18:09, "Globik" <notifications@github.com> ha scritto:

PR #1146, opened 2018-01-30
readd set stereo sdp

chadfurman:
This was missing in the new version of Janus.js with all the fancy simulcasting and stuff.  Brought it back :)


chadfurman:
Closing PR because this is not as configurable as it should be, breaks things in the audio bridge, and never was actually in the official Janus.js to begin with.

PR #1148, opened 2018-01-31
disabling md5, RC4, SSL 2/3, and compression

chadfurman:
Removing weak ciphers in dtls.c

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -84,7 +84,7 @@ static void janus_dtls_notify_state_change(janus_dtls_srtp *dtls) {
 
 
 /* DTLS stuff */
-#define DTLS_CIPHERS	"ALL:NULL:eNULL:aNULL"
+#define DTLS_CIPHERS	"HIGH:!aNULL:!MD5:!RC4"
 /* Duration for the self-generated certs: 1 year */
 #define DTLS_AUTOCERT_DURATION	60*60*24*365
 
@@ -486,7 +486,8 @@ janus_dtls_srtp *janus_dtls_srtp_create(void *ice_component, janus_dtls_role rol
 		janus_dtls_srtp_destroy(dtls);
 		return NULL;
 	}
-	SSL_set_options(dtls->ssl, SSL_OP_SINGLE_ECDH_USE);
+	const long flags = SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3 | SSL_OP_NO_COMPRESSION | SSL_OP_SINGLE_ECDH_USE;
+	SSL_set_options(dtls->ssl, flags);
 	SSL_set_tmp_ecdh(dtls->ssl, ecdh);
 	EC_KEY_free(ecdh);
 #ifdef HAVE_DTLS_SETTIMEOUT


lminiero:
Have you verified whether this works with all browsers? The DTLS setup for the handshake in WebRTC is different from the TLS handshake transport protocols like HTTPS and WSS do, and is much more constrained in terms of what it can and will not accept. Considering the certificates in WebRTC are all self signed, this seems a bit overkill.

oej:
If any browser or implementation use  RC4, SSL 2/3 and compression for webrtc I would be very, very disappointed and it would be a story to tell our grandchildren. I think this is very safe. The use of MD5 for certificate checksums is still out there for old-school TLS/HTTPS, but I doubt if any implementation use that for dtls

lminiero:
Thanks for the feedback @oej! I'm admittedly quite ignorant when it comes to these aspects, which is why I am always careful when anything touches the DTLS code, that is at the very foundation of WebRTC PeerConnections. I'll have a look at how browsers do the setup of ciphers for DTLS, to see if we can try and get closer to what they do and expect.

lminiero:
Apparently the libwebrtc code does this:

```
  // Select list of available ciphers. Note that !SHA256 and !SHA384 only
  // remove HMAC-SHA256 and HMAC-SHA384 cipher suites, not GCM cipher suites
  // with SHA256 or SHA384 as the handshake hash.
  // This matches the list of SSLClientSocketOpenSSL in Chromium.
  SSL_CTX_set_cipher_list(
      ctx, "DEFAULT:!NULL:!aNULL:!SHA256:!SHA384:!aECDH:!AESGCM+AES256:!aPSK");
```

for ciphers, and this:

```
SSL_set_options(ssl_, SSL_OP_SINGLE_ECDH_USE);
```

for options, which is what we had already. I guess this makes the options additions redundant?

chadfurman:
The following ciphers are considered weak or deprecated, according to my auditing team:
```
TLS_DHE_RSA_WITH_CAMELLIA_256_CBC_SHA (0x0088)
 TLS_DHE_DSS_WITH_CAMELLIA_256_CBC_SHA (0x0087)
 TLS_ECDH_anon_WITH_AES_256_CBC_SHA (0xc019)
 TLS_DH_anon_WITH_AES_256_CBC_SHA (0x003a)
 TLS_DH_anon_WITH_CAMELLIA_256_CBC_SHA (0x0089)
 TLS_RSA_WITH_CAMELLIA_256_CBC_SHA (0x0084)
 TLS_DHE_RSA_WITH_SEED_CBC_SHA (0x009a)
 TLS_DHE_DSS_WITH_SEED_CBC_SHA (0x0099)
 TLS_DHE_RSA_WITH_CAMELLIA_128_CBC_SHA (0x0045)
 TLS_DHE_DSS_WITH_CAMELLIA_128_CBC_SHA (0x0044)
 TLS_ECDH_anon_WITH_AES_128_CBC_SHA (0xc018)
 TLS_DH_anon_WITH_AES_128_CBC_SHA (0x0034)
 TLS_DH_anon_WITH_SEED_CBC_SHA (0x009b)
 TLS_DH_anon_WITH_CAMELLIA_128_CBC_SHA (0x0046)

chadfurman:
The flags I chose came from here: https://stackoverflow.com/a/32597934/293492

lminiero:
Anybody that can test this against Edge and/or Safari?

lminiero:
Works fine in my tests, so merging. Thanks again for your contribution @chadfurman!

PR #1152, opened 2018-02-06
Modernize Janus

ibc:
Yeah modernize Janus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
-Janus WebRTC Gateway
-====================
+Janus WebRTC Gateway to the Future
+==================================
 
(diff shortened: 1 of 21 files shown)


lminiero:
These are way too many changes for you to having made them manually! :D

mirkobrankovic:
Seems legit to me ;)

dwoz:
If it were not for all the missing context after this patch is applied I'd be all for it. Maybe modify the regex to not include the entire line. :|

lminiero:
Since the future is too far away and this isn't Stargate, I'll have to reject this PR...

ibc:
braveheart_3 2

PR #1153, opened 2018-02-07
RTP compensation for clock skew and mountpoint collision detection

atoppi:
This PR contains two very sensible features:

1) The **detection on a RTP streaming mountpoint of a collision with a different SSRC**. This means that if a RTP mountpoint is being fed by a live RTP stream, and another stream with a different SSRC tries to feed the same mountpoint (e.g. same UDP port), Janus will now discard the packets incoming from the new feed and will notice the user with a warning. This feature can be enabled using a new attribute in the mountpoint configuration: `collision` (expressed in milliseconds). In case of collision (more than one SSRC hitting the same port), the plugin will discard incoming RTP packets with a new SSRC unless this many milliseconds passed from the last packet with the old SSRC, which would then switch the current SSRC to the new one. While this feature could be intended like a very basic hijacking protection, it is more designed to protect Janus mountpoints from streamers misconfiguration or human errors at runtime (e.g. launch twice the same streamer).

2) A **counter-measure for RTP feeders' clock skew**. 
We noticed that very long (> 4 hours) recordings of a mountpoint in the streaming plugin occasionally showed a **lip-sync** issue. It come up that using different devices to feed the audio stream and video stream of a mountpoint is the cause of this lip-sync error. Specifically the differences among the sources clock generate a clock skew problem on Janus. This means e.g. that the video source RTP is arriving to Janus faster than the nominal rate (that depends upon video format), while the audio source RTP is arriving slower than the nominal rate, and Janus do its best to relay and record these packets exactly with the same timing it received them. This is always true for setups that involve different clock sources to stream audio and video related to same mountpoint, including for example setups with one board/laptop with two USB devices (camera and audio card).
The issue is present for online viewers too (the browsers). While the playout buffers management and Sender Reports coming from Janus permit the browsers to adjust the lip-sync, eventually the issue shows up after some hours. 
The proposed solution is a routine acting at the mountpoint entrance that compares the received RTP timestamp with an expected local RTP timestamp, in order to evaluate the current offset and compensate with incremental adjustments. The implementation is largely inspired by the pseudo-code found in [Colin Perkins' book, Chapter 6, par. Compensation for clock skew](https://books.google.it/books?id=OM7YJAy9_m8C&printsec=frontcover&hl=it&source=gbs_ge_summary_r&cad=0#v=onepage&q&f=false). The only difference is that we do not have a receiving buffer, so instead of working on a buffer we have to react _online_ in case of a compensation is needed. Specifically:
- If the sender is faster than the nominal rate, we drop N incoming packets and adjust subsequent RTP timestamps and sequence number taking them _back in time_. Online viewers will detect a little spike in inter-frame arrival time (should not be a problem), while offline postprocessors like `janus-pp-rec` will not be affected by this correction.
- If the sender is slower than the nominal rate, we simulate a loss, performing a RTP timestamp and sequence number jump _ahead in the future_. Online viewers will detect a packet loss (unless the code is changed to burst sending dummy packets), while postprocessors should insert missing packets (like `janus-pp-rec` already does).

By doing this for audio and video packets, we are sure that the audio and video streams are synchronized with the local clock (that should be synchronized in turn with NTP).
Our results are encouraging: we tested the algorithm with very long streaming sessions (>12h) using devices that showed the skew issue, and the lip-sync is perfect both for recordings and online viewers.
The feature is managed in the mountpoint configuration with the boolean attributes `audioskew` and `videoskew`. As of now, the operational thresholds are hardcoded to 40ms.

Sorry for the long post, I did my best to summarize the details.
Any question for further details or feedback is very very welcome.

### DIFF
--- a/conf/janus.plugin.streaming.cfg.sample.in
+++ b/conf/janus.plugin.streaming.cfg.sample.in
@@ -25,6 +25,8 @@
 ; audioiface = network interface or IP address to bind to, if any (binds to all otherwise)
 ; audiopt = <audio RTP payload type> (e.g., 111)
 ; audiortpmap = RTP map of the audio codec (e.g., opus/48000/2)
+; audioskew = yes|no (whether the plugin should perform skew
+;		analisys and compensation on incoming audio RTP stream, EXPERIMENTAL)
 ; videoport = local port for receiving video frames
 ; videomcast = multicast group port for receiving video frames, if any
 ; videoiface = network interface or IP address to bind to, if any (binds to all otherwise)
@@ -35,6 +37,11 @@
 ; videosimulcast = yes|no (do|don't enable video simulcasting)
 ; videoport2 = second local port for receiving video frames (only for rtp, and simulcasting)
 ; videoport3 = third local port for receiving video frames (only for rtp, and simulcasting)
+; videoskew = yes|no (whether the plugin should perform skew
+;		analisys and compensation on incoming video RTP stream, EXPERIMENTAL)
+; collision = in case of collision (more than one SSRC hitting the same port), the plugin
+;		will discard incoming RTP packets with a new SSRC unless this many milliseconds
+;		passed, which would then change the current SSRC (0=disabled)
 ; dataport = local port for receiving data messages to relay
 ; dataiface = network interface or IP address to bind to, if any (binds to all otherwise)
 ; databuffermsg = yes|no (whether the plugin should store the latest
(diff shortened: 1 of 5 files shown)


atyenoria:
Doest this algorithm also be used in the Video-Room and Audio-Bridge plugin recoding ?

atoppi:
@atyenoria while this PR is only affecting the streaming plugin, the code added in `rtp.c` is totally general, and can be applied to every RTP stream with a Janus RTP context, including the RTP coming from WebRTC participants. We have still to decide if this will be extended to other scenarios.

lminiero:
@atyenoria as Alessandro explained, this work started because of something weird we noticed in the Streaming plugin itself. Most of the times, the skew there is caused by different timings in the components that generate the media, which is more of a problem with the Streaming plugin as here media is captured by something that is not a browser. With a browser originating media, I don't expect there to be issues like this.

atyenoria:
Thanks! I got it.

PR #1155, opened 2018-02-11
fixed parse width/height from sps

PeterXu:
shown in code.

### DIFF
--- a/postprocessing/pp-h264.c
+++ b/postprocessing/pp-h264.c
@@ -181,9 +181,9 @@ static void janus_pp_h264_parse_sps(char *buffer, int *width, int *height) {
 
 	/* We skipped what we didn't care about and got what we wanted, compute width/height */
 	if(width)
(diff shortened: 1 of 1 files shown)


lminiero:
Is this really needed? Can you elaborate on what it fixes? I never saw problems in extracting the correct resolution of a keyframe with the existing code, and I use the same code successfully in some live RTP servers as well (so not only postprocessing).

lminiero:
@PeterXu ping :wink:

PeterXu:
For some cases, frame_crop_bottom_offset or frame_crop_top_offset may not be 0.
In my testing, H264 640x360 from Chrome(version: 63) will not be processed correctly.

lminiero:
Thanks for the clarification, I'll make some tests and check if it still works as expected in my other scenarios too. In the meanwhile, could you sign our CLA at the end of [this page](https://cla.conf.meetecho.com/janus/) so that I can merge after I've done my checks?

PeterXu:
I have signed it and expect your checking.

lminiero:
Looks like it's still working, merging then!

PR #1158, opened 2018-02-19
use jquery.fn.one instead of started toggle

graingert:


lminiero:
Nice, I didn't know about this feature of jQuery. Could you add the same to the other demos as well? We also need you to sign our CLA at the end of [this page](https://cla.conf.meetecho.com/janus/) before we can merge.

Thanks!

lminiero:
Looked like an easy enough change, so I did it myself. Thanks for bringing this to my attention!

PR #1164, opened 2018-02-27
config: remove wrong backslash in fmtp sdp

fippo:
it is wrong and Chrome does not parse it.

Came up on discuss-webrtc. @lminiero: shall we have @ibc review this?

### DIFF
--- a/conf/janus.plugin.streaming.cfg.sample.in
+++ b/conf/janus.plugin.streaming.cfg.sample.in
@@ -126,7 +126,7 @@ secret = adminpwd
 ;videoport = 8004
 ;videopt = 126
 ;videortpmap = H264/90000
-;videofmtp = profile-level-id=42e01f\;packetization-mode=1
+;videofmtp = profile-level-id=42e01f;packetization-mode=1
 
 ;
 ; This is a sample configuration for Opus/VP8 multicast streams


lminiero:
The backslash you see there is just needed to escape the `;` when parsing the cfg file, it does not appear in the SDP. If the mountpoint is created via API it must not be passed instead.

lminiero:
Ops thought you were serious for a second! I shouldn't review things so early in the morning :smile:

fippo:
but it seems to have appeared in the discuss-webrtc discussion?

lminiero:
Mh, you're right, just saw this on the group: it definitely works as expected with my Chrome though. I'm more inclined to think it's a misconfiguration the guy made. If you remove the slash, everything after the semicolon disappears as it's the character that is used for comments in our INI syntax. This means that only the profile remains, while the packetization mode goes away.

ibc:
Y U USING MICROSOFT WINDOWS INI FORMAT FOR CONFIG FILES?
SO ; MUST BE ESCAPED BECAUSE OTHERWISE IT STARS A COMMENT?
Y U NOT JSON OR YAML?
CAN I CONFIGURE JANUS THE WEBRTC GATEWAY WITH EXCEL?

lminiero:
Just tried and this is what I get in the SDP in Chrome:

    a=fmtp:126 profile-level-id=42e01f;packetization-mode=1

for this mountpoint configuration

```
[h264-sample]
type = rtp
id = 10
description = H.264 live stream coming from gstreamer
audio = no
video = yes
videoport = 8004
videopt = 126
videortpmap = H264/90000
;videortpmap = X-H264UC/90000
videofmtp = profile-level-id=42e01f\;packetization-mode=1
```

lminiero:
@ibc #1116 :smile:

lminiero:
> CAN I CONFIGURE JANUS THE WEBRTC GATEWAY WITH EXCEL?

WIP...

ibc:
hide the pain harold 27022018094434

ibc:
man distracted by laptop 27022018094145

lminiero:
:rofl:

fippo:
actually I was serious -- it had been soooooo long since I touched ini files that I had forgotten about the need to escape semicolons. Maybe this can be run through standard?

ibc:
`pc.setRemoteDescription(sdp.ini)`

fippo:
clearly web^Wrtc devs want INI support. Lets take this to the WG in a month!

lminiero:
Or maybe I should just get rid of INI files altogether and join the new millenium! :smile:

PR #1165, opened 2018-02-28
sipre: update sipreg_register to re 0.5.7

alfredh:
This patch updates the sipre plugin to use the latest api of libre 0.5.7

in 0.5.7 the sipreg api was changed:

https://github.com/creytiv/re/releases/tag/v0.5.7

please merge to master.

### DIFF
--- a/plugins/janus_sipre.c
+++ b/plugins/janus_sipre.c
@@ -4021,7 +4021,7 @@ void janus_sipre_mqueue_handler(int id, void *data, void *arg) {
 			/* Send the REGISTER */
 			int err = sipreg_register(&session->stack.reg, session->stack.sipstack,
 				session->account.proxy,
-				session->account.identity, session->account.identity, session->stack.expires,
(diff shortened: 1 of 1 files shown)


lminiero:
Won't this break compatibility with previous releases, though? I think it might be better to make this conditional to the version detected in configure. I'll check if that can be done easily.

alfredh:
yes it will break compatibility with older versions of libre.

my apologies for that :)

Alfred

lminiero:
@alfredh I noticed your headers don't include any versioning info as a define of some sort. Any chance this could be added in the future? It would allow for a much easier management of change in the APIs directly in the code with macros, rather than having to resort on configure.ac tweaks to detect exact versions and add defines there.

lminiero:
For the time being, handled this as a configure check. Thansk for the heads up!

alfredh:
great idea, we can add version info in the header files ..

Thanks for merging

PR #1166, opened 2018-02-28
Add support for SRTP with AES-GCM

lminiero:
As the title says, this PR adds support for [SRTP with AES-GCM](https://tools.ietf.org/html/rfc7714), which is something that Chrome is testing behind a flag: `chrome://flags/#enable-webrtc-srtp-aes-gcm`

The support is only in the core, for the moment, which means it only applies to the SRTP setup for WebRTC PeerConnections. The next step is doing the same for the SIP, SIPre and NoSIP plugins, which also allow for SRTP to be negotiated, even though via SDES and not DTLS-SRTP.

Feedback welcome!


lminiero:
@alfredh FYI, just added the same to the plugins that can do SRTP on the "legacy" side. Please let me know if you need any help for testing/validating against your code and, most importantly, if you find out anything we're not doing properly!

lminiero:
Merging.

PR #1167, opened 2018-02-28
Typo fixed. "ge" replaced with "be"

alienpavlov:


lminiero:
Rather than forcing you to sign our CLA for such a tiny change, I applied it myself. Thanks for spotting it out!

t291436520:
Hello：
         I have a question to be consulted about when sending DTMF in Janus. There is a createDTMFSender method that can not find the method at debug. What do you need to update? 
                                                                                                Thanks。

PR #1168, opened 2018-03-01
Link quality stats

pitkonst:
Clean version of [1047](https://github.com/meetecho/janus-gateway/pull/1047)


lminiero:
Thanks, the diff looks much cleaner now! And most importantly, I can check what's new again :smile: 
I'll get back to you with some feedback as soon as possible.

ice.c
```diff
@@ -3929,6 +3942,9 @@ void *janus_ice_send_thread(void *data) {
 									stream->video_first_rtp_ts[0] = timestamp;
 								}
 							}
+							// update sent packets counter
```
lminiero: /* Update sent packets counter */

As a convention, we don't use `//` for comments, and sentences in comments always start capitalized.

ice.c
```diff
@@ -4197,4 +4213,4 @@ void janus_ice_dtls_handshake_done(janus_ice_handle *handle, janus_ice_component
 		json_object_set_new(info, "connection", json_string("webrtcup"));
 		janus_events_notify_handlers(JANUS_EVENT_TYPE_WEBRTC, session->session_id, handle->handle_id, handle->opaque_id, info);
 	}
-}
+}
```
lminiero: Missing newline here, maybe your editor removed it?

janus.c
```diff
@@ -4337,4 +4345,4 @@ gint main(int argc, char *argv[])
 	JANUS_PRINT("Bye!\n");
 
 	exit(0);
-}
+}
```
lminiero: Same here.

rtcp.c
```diff
@@ -133,6 +133,47 @@ static void janus_rtcp_incoming_sr(janus_rtcp_context *ctx, janus_rtcp_sr *sr) {
...
 }
 
+/* Link quality estimate filter coefficient */
+#define LINK_QUALITY_FILTER_K 3.0
+
+static double janus_rtcp_link_quality_filter(double last, double in) {
+	if (last == 0) {
+		return in;
+	}
+	else {
```
lminiero: } else {

As a code convention, we put the `else` on the same line as the closing backet.

rtcp.c
```diff
@@ -133,6 +133,47 @@ static void janus_rtcp_incoming_sr(janus_rtcp_context *ctx, janus_rtcp_sr *sr) {
...
+		return (1.0 - 1.0/LINK_QUALITY_FILTER_K) * last + (1.0/LINK_QUALITY_FILTER_K) * in;
+	}
+}
+
+/* Update link quality stats based on RR */
+static void janus_rtcp_rr_update_stats(rtcp_context *ctx, report_block rb) {
```
lminiero: This should be `janus_report_block`, as we recently typedeffed all the RTCP structs to have the `janus_` prefix: the version without may be deprecated and removed in the future.

rtcp.c
```diff
@@ -1366,4 +1453,4 @@ int janus_rtcp_transport_wide_cc_feedback(char *packet, size_t size, guint32 ssr
 
 	/* Done */
 	return len;
-}
+}
```
lminiero: See above about missing newline.

rtcp.h
```diff
@@ -257,6 +257,26 @@ typedef struct rtcp_context
 	uint32_t expected;
 	uint32_t expected_prior;
 	uint32_t lost, lost_remote;
+
+	uint32_t retransmited;
+	uint32_t retransmited_prior;
```
lminiero: Just a typo: missing t in "retransmitted"?

rtcp.c
```diff
@@ -133,6 +133,47 @@ static void janus_rtcp_incoming_sr(janus_rtcp_context *ctx, janus_rtcp_sr *sr) {
...
+		return in;
+	}
+	else {
+		return (1.0 - 1.0/LINK_QUALITY_FILTER_K) * last + (1.0/LINK_QUALITY_FILTER_K) * in;
+	}
+}
+
+/* Update link quality stats based on RR */
+static void janus_rtcp_rr_update_stats(rtcp_context *ctx, report_block rb) {
+	uint32_t ts = janus_get_monotonic_time();
```
lminiero: `janus_get_monotonic_time()` returns a `int64_t`, not a `uint32_t`.

rtcp.h
```diff
@@ -257,6 +257,26 @@ typedef struct rtcp_context
...
+
+	/* Inbound RR process */
+	uint32_t rr_last_ts;
+	uint32_t rr_last_ehsnr;
+	uint32_t rr_last_lost;
+	uint32_t rr_last_nack_count;
```
lminiero: If all these were originated by a `janus_get_monotonic_time()`, see previous comment on `int64_t`.

rtcp.h
```diff
@@ -443,4 +477,4 @@ int janus_rtcp_nacks(char *packet, int len, GSList *nacks);
  * @returns The message data length in bytes, if successful, -1 on errors */
 int janus_rtcp_transport_wide_cc_feedback(char *packet, size_t len, guint32 ssrc, guint32 media, guint8 feedback_packet_count, GQueue *transport_wide_cc_stats);
 
-#endif
+#endif
```
lminiero: Missing newline.

pitkonst:
@lminiero thanks for review. Fixed.

lminiero:
Thx, will check later or tomorrow!

lminiero:
@pitkonst I'm not sure I understand how this works. For an easy test, I'm using the EchoTest demo locally, and at a certain point I hit the "disable video" button, which basically instructs the plugin not to dump and not relay the incoming packets; a few seconds later I push the "enable video" button and media starts flowing again, and video is recovered quickly (thanks to a PLI to resend a keyframe) and works.

Anyway, I'm getting strange results in the quality indicators you added. Apparently, they're either 100 or 0, no middle ground. I thought this could be casting problems and loss of detail from `double` to `uint32_t`, but printing the `double` values themselves seems to confirm those are indeed the values that are returned. After the interruption, at times I got a:

	[WARN] In link quality=-nan, media link quality=-nan

instead of

	[WARN] In link quality=100.000000, media link quality=100.000000

both coming from this modified log:

	JANUS_LOG(LOG_WARN, "In link quality=%f, media link quality=%f\n",
		ctx->in_link_quality, ctx->in_media_link_quality);

which is weird for incoming media (which is unaffected, it's the plugin that drops the frames). For outgoing video, the log was consistently:

	[WARN] Out link quality=-nan, media link quality=-nan

and the stats in the Admin API never recovered:

		[..]
		"out-link-quality": 0,
		"out-media-link-quality": 0
	}

even though the video was working as expected and no loss. Am I misunderstanding how this works, or is something not working as expected?

lminiero:
PS: one thing I noticed is that the lost by remote value was very high, ~5000, which doesn't seem right for the short time I paused the outgoing video. I'll have to check if the code we have to parse that from RR packets is broken, as I guess that might explain the `0` value (lost may be way larger than what we actually sent).

pitkonst:
In most case we do tests with VideoRoom plugin. I did tests as you described.
Disabling video lead to interruption media stream, which is an extreme case of connection problem, and should give 0 quality value. Disabling of out stream do not affects input stream, so input stats should be always 100, as it was in my test. There was a bug, with NaN value and out stats did not recover with enabling data flow, now it is fixed. And there is strange behavior with RR packets on disabled stream.

To see different stats values we simulate connection problems (packet drop and lag) in given direction.

lminiero:
Thanks, I can confirm the patch you applied fixed it for me! Checking the Admin API using the "autorefresh" property of the demo page (which means an Admin API poll every 5 seconds or so), when I interrupted the stream both `out-link-quality` and `out-media-link-quality` dropped to 0; resuming the stream, `out-link-quality` went up to 100 right away, while `out-media-link-quality` only started increasing progressively (34, 70, 87, 94, up to 100).

Thanks for the quick fix! I'll make some more tests in other plugins too, e.g., VideoRoom, and in some more different network conditions than my local network. If you haven't done that already, please advertise your effort on our [Google group](https://groups.google.com/forum/#!forum/meetecho-janus) as well, so that people there are aware of this and can make some tests too.

lminiero:
That said, if the next tests I'll make (either tomorrow or early next week) go as expected, I'll definitely merge soon, so thanks again for this!

tgabi333:
Great contribution, love to see in master, thanks @pitkonst

lminiero:
I don't see any objection, so I'll merge: we can track any issue in master itself. Thanks again @pitkonst!

PR #1183, opened 2018-03-19
Janus.js  'prepareWebrtc' function optimisations 

soulfly:
1) Drop pre Firefox 38 support. According to this stat https://en.wikipedia.org/wiki/Template:Firefox_usage_share it's only 0.02%
2) Added support for Chrome 59 getUserMedia constraints format
3) Optimized a case where 'media.video' is an object, so we do not calculate height+width in this case.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1772,69 +1772,71 @@ function Janus(gatewayCallbacks) {
 				if(simulcast && !jsep && (media.video === undefined || media.video === false))
 					media.video = "hires";
 				if(media.video && media.video != 'screen' && media.video != 'window') {
-					var width = 0;
-					var height = 0, maxHeight = 0;
-					if(media.video === 'lowres') {
-						// Small resolution, 4:3
-						height = 240;
-						maxHeight = 240;
-						width = 320;
-					} else if(media.video === 'lowres-16:9') {
-						// Small resolution, 16:9
-						height = 180;
-						maxHeight = 180;
-						width = 320;
-					} else if(media.video === 'hires' || media.video === 'hires-16:9' ) {
-						// High resolution is only 16:9
-						height = 720;
-						maxHeight = 720;
-						width = 1280;
-					} else if(media.video === 'stdres') {
-						// Normal resolution, 4:3
-						height = 480;
-						maxHeight = 480;
-						width  = 640;
-					} else if(media.video === 'stdres-16:9') {
-						// Normal resolution, 16:9
-						height = 360;
-						maxHeight = 360;
-						width = 640;
-					} else {
-						Janus.log("Default video setting is stdres 4:3");
-						height = 480;
-						maxHeight = 480;
-						width = 640;
-					}
-					Janus.log("Adding media constraint:", media.video);
+					if(typeof media.video === 'object') {
+						videoSupport = media.video;
+					}else{
+						var width = 0;
+						var height = 0, maxHeight = 0;
+						if(media.video === 'lowres') {
+							// Small resolution, 4:3
+							height = 240;
+							maxHeight = 240;
+							width = 320;
+						} else if(media.video === 'lowres-16:9') {
+							// Small resolution, 16:9
+							height = 180;
+							maxHeight = 180;
+							width = 320;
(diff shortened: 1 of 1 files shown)


lminiero:
@soulfly thanks for your contributions as usual! I'm abroad this week so I'll be quite slow to get back to you, sorry... I'll try to have a look ASAP.

tgabi333:
Just asking: shouldn't it be the responsibility of webrtc-adapter and use the format according to specs?

lminiero:
Not when you're playing with constraints, I think.

tgabi333:
webrtc-adapter already contains shims for getUserMedia constraints, all the codes creating constraint should stick to spec's format, aren't they?

html/janus.js
```diff
@@ -1772,80 +1772,71 @@ function Janus(gatewayCallbacks) {
...
-						var firefoxVer = parseInt(window.navigator.userAgent.match(/Firefox\/(.*)/)[1], 10);
-						if(firefoxVer < 38) {
-							videoSupport = {
-								'require': ['height', 'width'],
-								'height': {'max': maxHeight, 'min': height},
-								'width':  {'max': width,  'min': width}
-							};
+					if(typeof media.video === 'object') {
+						videoSupport = media.video;
+					}else{
```
lminiero: Missing spaces here (should be `} else {`)

html/janus.js
```diff
@@ -1772,80 +1772,71 @@ function Janus(gatewayCallbacks) {
...
+						Janus.log("Adding media constraint:", media.video);
+
+						if(Janus.webRTCAdapter.browserDetails.browser == 'firefox') {
 							videoSupport = {
 								'height': {'ideal': height},
 								'width':  {'ideal': width}
 							};
+						} else {
+							var chromeVer = Janus.webRTCAdapter.browserDetails.version;
+					    if(chromeVer >= 59) {
```
lminiero: Broken indentation?

html/janus.js
```diff
@@ -1772,80 +1772,71 @@ function Janus(gatewayCallbacks) {
...
+								videoSupport = {
+									'height': {'ideal': height},
+									'width':  {'ideal': width}
+								};
+							}else{
```
lminiero: Missing spaces.

soulfly:
If we are talking about `chromeVer >= 59` check then it probably can be handled via adapter.js. I did some checks and there is some code mentioning 61 version https://github.com/webrtc/adapter/blob/master/src/js/chrome/getusermedia.js#L70 which is a bit different from what I did with 59 ver.

So probably we can remove this one check for Chrome version in this PR,
let me double check that on different Chrome versions

soulfly:
What I learnt while investigating adapter.js code:

1) If you are running >= **Chrome61** then all constraints go as is, w/o any pre-processing
Here is where it goes 
https://github.com/webrtc/adapter/blob/master/src/js/chrome/getusermedia.js#L71
https://github.com/webrtc/adapter/blob/master/src/js/chrome/getusermedia.js#L216

2) This is an old spec format:

```
var videoSupport = {
   'mandatory': {
     'maxHeight': maxHeight,
     'minHeight': height,
     'maxWidth':  width,
     'minWidth':  width
  },
  'optional': []
};
```

The old spec was the only one supported by Chrome until Chrome 59 (June 2017) which added support for the new one:

```
var videoSupport = {
  'height': {'ideal': height},
  'width':  {'ideal': width}
}
```

Older versions of Chrome can also use the new spec through adapter.js

====

So what I can say is that we can remove a check I introduced ( `if(chromeVer >= 59)` ) and completely switch to the new spec format. All Chrome browsers >= 61 will use it as is and all older versions will use it via adapter.js.

So it's clear now why adapter.js uses constraints for Chrome >= 61 as is. It's because newer version of Chrome supports both formats: the old one and the new one. 

Hope it's clear now

You can read some more info here https://addpipe.com/blog/getusermedia-video-constraints/

Will modify my PR soon.

lminiero:
@soulfly thanks, I'll check ASAP and merge if it looks fine :+1:

lminiero:
Looks good, merging!

PR #1185, opened 2018-03-22
Add MQTT event handler

oej:
This pull request adds an MQTT event plugin. We have used this in a running system for a while and done some magic using Node Red to handle the events. 

This module does not yet support TLS connections, at the point of merging the other MQTT plugin did not support it. When we have a working MQTT/TLS implementation I'll merge some code for that too.


lminiero:
Thanks a lot for this effort! I'll try to come back with a review ASAP (as you know the IETF can be quite time consuming :smile: ).

One quick note I have is related to the new types and functions I see you added to events.h and event.c: could you make sure they're prefixed with `janus_`?

Besides, not sure if you're aware of the feature I added in #1160, which is a way to allow you to tweak event handlers via Admin API: we have some example in both HTTP and RMQ (both limited to simple tweaks at the moment), so if there's anything that makes sense for your new handler as well, it might be cool to have something there too. If you already did that feel free to ignore! (haven't looked at the code yet)

lminiero:
Just got back to the office, so I'll make a review later today. Thanks again for working on this!

events/janus_mqttevh.c
```diff
@@ -0,0 +1,897 @@
...
+
+                .get_api_compatibility = janus_mqttevh_get_api_compatibility,
+                .get_version = janus_mqttevh_get_version,
+                .get_version_string = janus_mqttevh_get_version_string,
+                .get_description = janus_mqttevh_get_description,
+                .get_name = janus_mqttevh_get_name,
+                .get_author = janus_mqttevh_get_author,
+                .get_package = janus_mqttevh_get_package,
+
+                .incoming_event = janus_mqttevh_incoming_event,
```
lminiero: As anticipated in a previous comment, we now also support a `handle_request` callback for "poking" the handler via Admin API. Anyway, feel free to disregard that, as I can add it myself once we merge. If you probe a handler that doesn't implement it, the Admin API simply returns an error so nothing breaks anyway.

events/janus_mqttevh.c
```diff
@@ -0,0 +1,897 @@
...
+
+	msg.payload = payload;
+	msg.payloadlen = strlen(payload);
+	msg.qos = ctx->publish.qos;
+	msg.retained = retain;
+
+	/* TODO: The payload if generated by json_dumps needs to be freed
+	free(payload);
+	payload = (char *) NULL;
+	*/
```
lminiero: Looking at the code, isn't it always the case now? I guess it's commented now because `MQTTAsync_sendMessage` hasn't happened yet. Thinking about it, the current MQTT transport doesn't free that either, which makes me think that maybe paho-c frees the payload itself when done with the `MQTTAsync_message` instance? Either that, or MQTT has been leaking so far and we didn't notice :slightly_smiling_face:

lminiero:
@oej any feedback on the comments above? Really looking forward to merging this, thanks again for your contribution!

lminiero:
@oej any update on this effort? I know you're very busy on other activities, so if you want I can work on a new plugin that's based on your contribution, fix what needs fixing and push a first version myself. Thanks!

oej:
I'll be happy to tag along and review a new effort. My module has been working for a long time now, so the basics are right even though the rest of Janus has moved forward :-)

lminiero:
Ack, I'll work on one as soon as I can and mention you when it's in a ready enough state :+1:

lminiero:
@oej I just pushed a new version of the plugin in #1349: please let me know if that works for you. Notice that there are a couple of important changes you should be aware of. Thanks again for working on this! I'll close this PR in the meanwhile.

PR #1189, opened 2018-03-26
WIP: include janus mjr metadata into postprocessed video

xadhoom:
Opening this PR for discussion, if ok, I'll complete and clean up it before merge.

Problem:
under normal circumstances, the joins into a recorded room are not simultaneous, so each stream will start and end at different times.

When you need to postprocess files (play them back in a synced way, pad with blank video at beginning and/or end of the stream), is difficult to get when each stream starts, because janus-pp-rec output is not formatted (ie machine readable) and you don't know when each stream ends.

So you have to parse janus-pp-rec output (hoping for no log format changes between releases) to get start and using generated video metadata (using for example ffprobe) to get the duration of each file.

Thankfully, ffprobe output can be formatted in json/xml.

So why not add .mjr header into the generated .mp4/webm/whatever metadata? Doing that, and thanks to ffproble parsable output, is possibile to get all informations in one pass.

The tag choosen is the `comment`, which seems to be common between all formats.

In this PR I've done only for h264, just to experiment a bit.

If it is interesting, I'll be happy to complete it for the other containers.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -17,15 +17,15 @@
  * G.711 (mu-law or a-law) frames. In case the recording contains text
  * frames as received via data channels, instead, a .srt file will be
  * generated with the text content and the related timing information.
- * 
+ *
  * Using the utility is quite simple. Just pass, as arguments to the tool,
  * the path to the .mjr source file you want to post-process, and the
  * path to the destination file, e.g.:
- * 
+ *
 \verbatim
 ./janus-pp-rec /path/to/source.mjr /path/to/destination.[opus|wav|webm|h264|srt]
-\endverbatim 
- * 
+\endverbatim
+ *
  * An attempt to specify an output format that is not compliant with the
  * recording content (e.g., a .webm for H.264 frames) will result in an
  * error since, again, no transcoding is involved.
@@ -44,7 +44,7 @@
  * the frames in a valid container. Any further post-processing (e.g.,
  * muxing audio and video belonging to the same media session in a single
  * .webm file) is up to third-party applications.
- * 
+ *
  * \ingroup postprocessing
  * \ref postprocessing
  */
@@ -116,7 +116,7 @@ int main(int argc, char *argv[])
 			post_reset_trigger = val;
 		JANUS_LOG(LOG_INFO, "Post reset trigger: %d\n", post_reset_trigger);
 	}
-	
+
 	/* Evaluate arguments */
 	if(argc != 3) {
 		JANUS_LOG(LOG_INFO, "Usage: %s source.mjr destination.[opus|wav|webm|mp4|srt]\n", argv[0]);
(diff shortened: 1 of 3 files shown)


tgabi333:
Relying on publishers mediaState/hangup/webrtcState events is more accurate imo.

i saw a lot of processing issues resulting falsey length of video files (see https://github.com/meetecho/janus-gateway/commit/33dcf724826a917a35c1b53345e793868960f72a) with various browser and networking issues.

lminiero:
I wasn't aware of this possibility. If it really is common to all formats, and we're sure it won't break compatibility or playout, I'm definitely ok with this :+1: . As a side note, it might be easier to add another option to `janus-pp-rec` to just print the .mjr JSON header to stdout: we already have a `--header` option which parses that header, so another option to just print that and nothing else should be trivial enough.

xadhoom:
Well all containers except for .srt support tags, not sure about opus, need to check.

I'll work a bit on this next week and also add add the option to dump the json header as-is.

lminiero:
> As a side note, it might be easier to add another option to janus-pp-rec to just print the .mjr JSON header to stdout: we already have a --header option which parses that header, so another option to just print that and nothing else should be trivial enough

FYI, just added a `--json` option to the postprocessor which does exactly what I wrote a few months ago here. As a side note, is this effort still alive, or should we close this?

lminiero:
I ended up working on this myself, as it wasn't too hard, in https://github.com/meetecho/janus-gateway/commit/7ce60372269a0c185abaedea4e7d89afd4e8c1fa
Now the post-processor passes the stringified version of the .mjr JSON header to all the containers: if those support writing it down somewhere, they do, otherwise it's ignored. You can also override what's written in the file by setting the `JANUS_PPREC_METADATA` env variable (useful if you want something different to be put in the files).

It currently works when postprocessing to webm/mp4 (using the `av_dict_set` method from your patch) and opus (where we use the `DESCRIPTION` tag); it isn't when postprocessing to wav or srt, as neither have any way of adding comments or something alike, apparently.

Example of metadata in a webm file, checked using `mediainfo`:

```
[..]
Writing application                      : Lavf58.12.100
Writing library                          : Lavf58.12.100
Comment                                  : {"t": "v", "c": "vp8", "s": 1466784279305521, "u": 1466784279922784}
[..]
```

Same example, but when postprocessing with a `JANUS_PPREC_METADATA="Ciccio"` env variable:

```
[..]
Writing application                      : Lavf58.12.100
Writing library                          : Lavf58.12.100
Comment                                  : Ciccio
[..]
```

Example of metadata in an opus file, checked using `opusinfo`:

```
[..]
Encoded with Janus post-processor 0.5.0
User comments section follows...
	DESCRIPTION={"t": "a", "c": "opus", "s": 1500971682510127, "u": 1500971682641222}
[..]
```

Same example, but when postprocessing with a `JANUS_PPREC_METADATA="Ciccio"` env variable:

```
[..]
Encoded with Janus post-processor 0.5.0
User comments section follows...
	DESCRIPTION=Ciccio
[..]
```

I'm sure there may be other tags or properties this info may be written to, and that others may come up with other information to add besides what we're putting in right now. Anyway, this is a start, and seems to fix your original requirement, so I'll close this PR which is now outdated. Thanks for suggesting the idea and contributing that first patch!

PR #1195, opened 2018-04-01
Adds dynamic properties (PINs, Secrets, etc) for Streaming Plugin

RobertLowe:
- Adds dynamic `edit` method for streaming plugin like in: e2fff24cd07253a831faaad17bb551702a997435

👍 

aww-yeah

Thx @lminiero at this rate my c skills won't be rust and dust 😄

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -118,7 +118,7 @@ rtspiface = network interface IP address or device name to listen on when receiv
  * (invalid JSON, invalid request) which will always result in a
  * synchronous error response even for asynchronous requests.
  *
- * \c list , \c create , \c destroy , \c recording , \c enable and
+ * \c list , \c create , \c destroy , \c recording , \c edit , \c enable and
  * \c disable are synchronous requests, which means you'll
  * get a response directly within the context of the transaction. \c list
  * lists all the available streams; \c create allows you to create a new
@@ -128,6 +128,7 @@ rtspiface = network interface IP address or device name to listen on when receiv
  * recorded while it's broadcasted; \c enable and \c disable respectively
  * enable and disable a mountpoint, that is decide whether or not a
  * mountpoint should be available to users without destroying it.
+ * \c edit allows you to dynamically edit some mountpoint properties (e.g., the PIN);
  *
  * The \c watch , \c start , \c pause , \c switch and \c stop requests
  * instead are all asynchronous, which means you'll get a notification
@@ -152,6 +153,7 @@ rtspiface = network interface IP address or device name to listen on when receiv
  * \ingroup plugins
  * \ref plugins
  */
+
 
 #include "plugin.h"
 
@@ -252,6 +254,14 @@ static struct janus_json_parameter watch_parameters[] = {
 };
 static struct janus_json_parameter adminkey_parameters[] = {
 	{"admin_key", JSON_STRING, JANUS_JSON_PARAM_REQUIRED}
+};
+static struct janus_json_parameter edit_parameters[] = {
+	{"id", JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE},
+	{"new_description", JSON_STRING, 0},
+	{"new_secret", JSON_STRING, 0},
+	{"new_pin", JSON_STRING, 0},
+	{"new_is_private", JANUS_JSON_BOOL, 0},
+	{"permanent", JANUS_JSON_BOOL, 0}
 };
 static struct janus_json_parameter create_parameters[] = {
 	{"type", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
@@ -718,7 +728,7 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 		if(!notify_events && callback->events_is_enabled()) {
 			JANUS_LOG(LOG_WARN, "Notification of events to handlers disabled for %s\n", JANUS_STREAMING_NAME);
 		}
-		/* Iterate on all rooms */
+		/* Iterate on all mountpoints */
 		GList *cl = janus_config_get_categories(config);
 		while(cl != NULL) {
 			janus_config_category *cat = (janus_config_category *)cl->data;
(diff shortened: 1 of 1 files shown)


plugins/janus_streaming.c
```diff
@@ -118,12 +118,13 @@ rtspiface = network interface IP address or device name to listen on when receiv
...
  *
- * \c list , \c create , \c destroy , \c recording , \c enable and
+ * \c list , \c create , \c destroy , \c recording , \c edit , \c enable and
  * \c disable are synchronous requests, which means you'll
  * get a response directly within the context of the transaction. \c list
  * lists all the available streams; \c create allows you to create a new
  * mountpoint dynamically, as an alternative to using the configuration
- * file; \c destroy removes a mountpoint and destroys it; \c recording
+ * file; + * dynamically, as an alternative to using the configuration file; 
+ * \c edit allows you to dynamically edit some mountpoint properties (e.g., the PIN);
```
lminiero: Is the line before this a leftover/typo?

plugins/janus_streaming.c
```diff
@@ -148,6 +149,37 @@ rtspiface = network interface IP address or device name to listen on when receiv
...
+ *
+ * A successful destruction procedure will result in an \c edited response:
+ *
+\verbatim
+{
+	"mountpoint" : "edited",
+	"id" : <unique numeric ID>,
+	"permanent" : <true|false>
+}
+\endverbatim
```
lminiero: It's a bit weird to see documentation for a single method, considering all the others are missing... (`Actual API docs: TBD.`) I guess you can leave this out until the whole API is documented.

plugins/janus_streaming.c
```diff
@@ -718,7 +758,7 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 		if(!notify_events && callback->events_is_enabled()) {
 			JANUS_LOG(LOG_WARN, "Notification of events to handlers disabled for %s\n", JANUS_STREAMING_NAME);
 		}
-		/* Iterate on all rooms */
+		/* Iterate on all mountpoints */
```
lminiero: Damn copy/paste! :smile:

plugins/janus_streaming.c
```diff
@@ -2222,7 +2371,8 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session
 			/* The category to remove is the mountpoint name */
 			janus_config_remove_category(config, mp->name);
 			/* Save modified configuration */
-			janus_config_save(config, config_folder, JANUS_STREAMING_PACKAGE);
+			if(janus_config_save(config, config_folder, JANUS_STREAMING_PACKAGE) < 0)
+				save = FALSE;	/* This will notify the user the mountpoint is not permanent */
```
lminiero: This looks like a good suggestion, it might be worth adding this to other plugins as well (and to the other potentially permanent requests in this plugin).

RobertLowe:
@lminiero I did test it, it was working for me, but a second review would be great! 😄

plugins/janus_streaming.c
RobertLowe: Yea, I wasn't sure, I figure you'd notice and make the call! 👍

plugins/janus_streaming.c
RobertLowe: Damn, I missed another!

plugins/janus_streaming.c
RobertLowe: Totally, but my thoughts we're we might as well guess this started!

plugins/janus_streaming.c
RobertLowe: Yea, missed a newline!

lminiero: There's an extra `+ *` too, though.

plugins/janus_streaming.c
lminiero: I'd rather see this happen in a different PR that adds a documentation for everything: having a single thing in would look weird in the generated docs.

plugins/janus_streaming.c
```diff
@@ -2155,6 +2195,115 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session
...
+			g_free(old_pin);
+		}
+		if(save) {
+			/* This change is permanent: save to the configuration file too
+			 * FIXME: We should check if anything fails... */
+			JANUS_LOG(LOG_VERB, "Modifying streaming mountpoint %"SCNu64" permanently in config file\n", mp->id);
+			janus_mutex_lock(&config_mutex);
+			char cat[BUFSIZ];
+			/* The mountpoint ID is the category */
+			g_snprintf(cat, BUFSIZ, "%"SCNu64, mp->id);
```
lminiero: By testing I found this mistake here. The Streaming plugin works differently from AudioBridge/VideoRoom/TextRoom, and the name of the category is _not_ the ID, but its name. Trying to edit an existing mountpoint (actually leaving everything untouched):

    streaming.send({message: {request: "edit", id: 3, secret: "adminpwd", permanent: true}})

I got this new invalid mountpoint instance in my config file:

    [3]
    description = mu-law file source (music)
    secret = adminpwd

while the actual mountpoint instance I wanted to modify (`file-ondemand-sample`, which has ID 3) was left untouched.

plugins/janus_streaming.c
```diff
@@ -2155,6 +2195,115 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session
...
+			janus_config_remove_category(config, cat);
+			/* Now write the mountpoint details again */
+			janus_config_add_category(config, cat);
+			janus_config_add_item(config, cat, "description", mp->description);
+			if(mp->is_private)
+				janus_config_add_item(config, cat, "is_private", "yes");
+			if(mp->secret)
+				janus_config_add_item(config, cat, "secret", mp->secret);
+			if(mp->pin)
+				janus_config_add_item(config, cat, "pin", mp->pin);
```
lminiero: This is broken too: you're removing the category with `janus_config_remove_category`, but only adding a small set of properties when writing it again (description, is_private, secret, pin), leaving out all the ton of properties the different mountpoints can have (see what the "create" code does for instance).

lminiero:
Added a couple of additional notes. Unfortunately it's not mergeable as it is now, as it breaks mountpoints.

plugins/janus_streaming.c
RobertLowe: Ah, I think I see, so we should use `mp->name` in some cases and I guess `mp->id` in others? Sorry, I didn't check configuration saving as it wasn't something I use, I'll look into it, any suggestions further would be helpful.

plugins/janus_streaming.c
RobertLowe: Ah, thanks, will patch.

RobertLowe:
@lminiero wow, what a challenge 😄 tested saving seems to work fine for me now, let me know! Other requests handled. Commit amend'd 😟 to keep clean.  👍

lminiero:
Thanks for the quick fixes! I'll give it a go later and merge if it works fine :+1:

lminiero:
Looks good, merging!

lminiero:
@RobertLowe just FYI, while working on some documentation for the plugin I noticed that your `edited` response is the only one that uses a different name than `streaming` (`"mountpoint":"edited"`). I'm changing this to align it to the rest of the code.

RobertLowe:
@lminiero Thanks for the heads up, that's great, wasn't aware how the other plugins reacted!

RobertLowe:
@lminiero just a side thought, the inter-plugin terminology could be more standard, for instance there's `mountpoints` and `rooms` and `streaming` variables but when I think of the REST API they're all `rooms` essentially or `channels`.  Food for thought. Also noticed `configure` and `edit`, `configure` is for the handle and `edit` for the `room/mountpoint`. Maybe something like `handle` and `edit` to be distinct?

Issue #979, opened 2017-08-21
Reconnect and resume transport


iclems:
Today, whenever the connection is lost for some time, the Janus instance error is called and the session becomes disconnected. 

At this stage, there's no clean way to resume a session, but it's not killed and cleaned up automatically either. 

We could consider implementing directly at janus.js level a reconnect logic and monitoring of the network status. It feels complicated to implement that outside of janus.js and too much to recreate a complete Janus instance every time the connection comes back online.

lminiero:
Sessions are cleaned up after about 60s: that's what the timeout mechanism is for. That said, reclaiming a session might be an interesting feature: I thought about this some time ago, but never worked on it as I remember finding some issues in the logic I wanted to follow. We can definitely discuss how this could be implemented in upcoming versions.

iclems:
Sure that'd be interesting. On my side, I implemented a logic similar to what Firebase does with its realtime database: watch for network status and API disconnects, and automatically schedule reconnects if the network is up with randomized increasing times or instantly at the time the network comes back up. It's pretty efficient and helps with scalability. I'm wondering if this wouldn't be better placed directly within Janus rather than outside. As an app using Janus, you'd expect it to manage reconnecting automatically I suspect?

iclems:
More importantly, there's no easy way to clean up sessions locally today from what I understand. Right now, my auto reconnect mechanism checks if there's any connected session and if not, automatically reconnects. But since the previous sessions have failed and set their connected = false, at this stage, calling session.destroy doesn't work so you're stuck with ghost disconnected sessions locally.

lminiero:
Have you checked the Admin API already? https://janus.conf.meetecho.com/docs/admin
If you query a session, it will tell you if it exists, when it was last refreshed, etc.

iclems:
Sure, but that's not meant to be used directly by end users right? My point is that sessions on the client side are a mess if you keep the window running and have to face network disconnects and reconnect automatically.

lminiero:
[...]

PR #1199, opened 2018-04-05
Claim Session Feature

geigev:
This is my attempt at implementing "session claims" to solve the issue discussed  in https://github.com/meetecho/janus-gateway/issues/979 and in this post: https://groups.google.com/forum/#!topic/meetecho-janus/5ZkzlZiRfU8. 

A new transport can claim a session by making a `claim` request that contains the correct `secret_session_id`. This id must be specified when the session is created, or no users will be authorized to make `claim` requests.

The `--reclaim_session_timeout` flag sets a grace period where a session can be reclaimed by another transport before the session is cleaned up. The default value is `0`, so the timeout must be set in the janus config file, or with the flag, to create a grace period.

I am not sure if the way I implemented `secret_session_id` makes sense. I only added it because a post mentioned the potential of "stealing" sessions, but I'm not sure why a client would know another clients session id.

This is my first Janus PR, so proceed with caution. I figured that I would push it since it helped me solve an issue I was having with websocket disconnects and another user was having similar issues: https://groups.google.com/forum/#!topic/meetecho-janus/bhyK0ldvd9Q

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -515,7 +515,7 @@ static gpointer janus_sessions_watchdog(gpointer user_data) {
 }
 
 
-janus_session *janus_session_create(guint64 session_id) {
+janus_session *janus_session_create(guint64 session_id, guint64 secret_session_id) {
 	janus_session *session = NULL;
 	if(session_id == 0) {
 		while(session_id == 0) {
@@ -528,9 +528,13 @@ janus_session *janus_session_create(guint64 session_id) {
 			}
 		}
 	}
+	if(secret_session_id == 0) {
+		JANUS_LOG(LOG_INFO, "No secret ID specified for session %"SCNu64"\n", session_id);
+	}
 	session = (janus_session *)g_malloc(sizeof(janus_session));
 	JANUS_LOG(LOG_INFO, "Creating new session: %"SCNu64"; %p\n", session_id, session);
 	session->session_id = session_id;
+	session->secret_session_id = secret_session_id;
 	janus_refcount_init(&session->ref, janus_session_free);
 	session->source = NULL;
 	g_atomic_int_set(&session->destroyed, 0);
@@ -774,10 +778,13 @@ int janus_process_incoming_request(janus_request *request) {
 	char error_cause[100];
 	json_t *root = request->message;
 	/* Ok, let's start with the ids */
-	guint64 session_id = 0, handle_id = 0;
+	guint64 session_id = 0, handle_id = 0, secret_session_id = 0;
 	json_t *s = json_object_get(root, "session_id");
 	if(s && json_is_integer(s))
 		session_id = json_integer_value(s);
+	json_t *ss = json_object_get(root, "secret_session_id");
+	if(ss && json_is_integer(ss))
+		secret_session_id = json_integer_value(s);
 	json_t *h = json_object_get(root, "handle_id");
 	if(h && json_is_integer(h))
 		handle_id = json_integer_value(h);
@@ -830,8 +837,12 @@ int janus_process_incoming_request(janus_request *request) {
 				goto jsondone;
 			}
 		}
+		json_t *secret_id = json_object_get(root, "secret_id");
+		if(secret_id != NULL) {
+			secret_session_id = json_integer_value(id);
+		}
 		/* Handle it */
-		session = janus_session_create(session_id);
+		session = janus_session_create(session_id, secret_session_id);
 		if(session == NULL) {
 			ret = janus_process_error(request, session_id, transaction_text, JANUS_ERROR_UNKNOWN, "Memory error");
 			goto jsondone;
@@ -1025,11 +1036,12 @@ int janus_process_incoming_request(janus_request *request) {
 		/* Send the success reply */
 		ret = janus_process_success(request, reply);
 	} else if(!strcasecmp(message_text, "claim")) {
-		if(reclaim_session_timeout == 0) {
-			ret = janus_process_error(request, session_id, transaction_text, JANUS_ERROR_UNAUTHORIZED, "Claiming sessions is not enabled");
+		janus_mutex_lock(&session->mutex);
+		if(session->secret_session_id == 0 || secret_session_id != session->secret_session_id) {
+			janus_mutex_unlock(&session->mutex);
+			ret = janus_process_error(request, session_id, transaction_text, JANUS_ERROR_UNAUTHORIZED, "Unauthorized to claim this session");
 			goto jsondone;
 		}
-		janus_mutex_lock(&session->mutex);
 		if(session->source != NULL) {
 			/* Give old tranport a timeout -- is this the right thing to do? */
 			session->source->transport->session_over(session->source->instance, session->session_id, TRUE);
--- a/janus.c
+++ b/janus.c
@@ -197,6 +197,9 @@ static json_t *janus_create_message(const char *status, uint64_t session_id, con
 #define DEFAULT_SESSION_TIMEOUT		60
 static uint session_timeout = DEFAULT_SESSION_TIMEOUT;
 
+#define DEFAULT_RECLAIM_SESSION_TIMEOUT		0
+static uint reclaim_session_timeout = DEFAULT_RECLAIM_SESSION_TIMEOUT;
+
 
 /* Information */
 static json_t *janus_info(const char *transaction) {
@@ -218,6 +221,7 @@ static json_t *janus_info(const char *transaction) {
 	json_object_set_new(info, "data_channels", json_false());
 #endif
 	json_object_set_new(info, "session-timeout", json_integer(session_timeout));
+	json_object_set_new(info, "reclaim-session-timeout", json_integer(reclaim_session_timeout));
 	json_object_set_new(info, "server-name", json_string(server_name ? server_name : JANUS_SERVER_NAME));
 	json_object_set_new(info, "local-ip", json_string(local_ip));
 	if(public_ip != NULL)
@@ -461,8 +465,10 @@ static gboolean janus_check_sessions(gpointer user_data) {
 				continue;
 			}
 			gint64 now = janus_get_monotonic_time();
-			if (now - session->last_activity >= (gint64)session_timeout * G_USEC_PER_SEC &&
-					!g_atomic_int_compare_and_exchange(&session->timeout, 0, 1)) {
+			if ((now - session->last_activity >= (gint64)session_timeout * G_USEC_PER_SEC &&
+					!g_atomic_int_compare_and_exchange(&session->timeout, 0, 1)) ||
+					((g_atomic_int_get(&session->transport_gone) && now - session->last_activity >= (gint64)reclaim_session_timeout * G_USEC_PER_SEC) &&
+							!g_atomic_int_compare_and_exchange(&session->timeout, 0, 1))) {
 				JANUS_LOG(LOG_INFO, "Timeout expired for session %"SCNu64"...\n", session->session_id);
 				/* Mark the session as over, we'll deal with it later */
 				janus_session_handles_clear(session);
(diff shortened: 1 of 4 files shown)


lminiero:
Thanks for working on this! It seems very much in line with what I planned to do, sooner or later, as discussed in other posts, e.g., https://groups.google.com/d/topic/meetecho-janus/5ZkzlZiRfU8/discussion

I'll review later to let you know if I have any comments, especially considering we merged the refcount branch yesterday and so there might be things to be wary of. One thing I can anticipate right away is that  I guess you can get rid of the session secret stuff: I tried adding something like this in the past, in #888, and nobody felt it was useful, so you can just drop that part.

geigev:
>  I tried adding something like this in the past, in #888, and nobody felt it was useful, so you can just drop that part

Gotcha, makes sense. I removed the secret stuff.

janus.c
```diff
@@ -1017,6 +1024,28 @@ int janus_process_incoming_request(janus_request *request) {
 		json_t *reply = janus_create_message("success", session_id, transaction_text);
 		/* Send the success reply */
 		ret = janus_process_success(request, reply);
+	} else if(!strcasecmp(message_text, "claim")) {
+		janus_mutex_lock(&session->mutex);
+		if(session->source != NULL) {
+			/* Give old tranport a timeout -- is this the right thing to do? */
+			session->source->transport->session_over(session->source->instance, session->session_id, TRUE);
+			janus_request_destroy(session->source);
```
lminiero: I'll have to check if this can cause issues with reference counters.

janus.c
```diff
@@ -1017,6 +1024,28 @@ int janus_process_incoming_request(janus_request *request) {
 		json_t *reply = janus_create_message("success", session_id, transaction_text);
 		/* Send the success reply */
 		ret = janus_process_success(request, reply);
+	} else if(!strcasecmp(message_text, "claim")) {
+		janus_mutex_lock(&session->mutex);
+		if(session->source != NULL) {
+			/* Give old tranport a timeout -- is this the right thing to do? */
+			session->source->transport->session_over(session->source->instance, session->session_id, TRUE);
```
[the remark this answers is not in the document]
geigev: Thought that approach may have been a bit naive -- I'll take a look at the transports.

geigev:
Thanks for the feedback. I added a `session_reclaimed` callback based on your comments. I have not tested all the transports yet; might take me a couple days to get around to it as I have never used the pfunix or the mqtt transport before.

transports/transport.h
```diff
@@ -221,6 +221,11 @@ struct janus_transport {
 	 * @param[in] session_id The session ID that was closed (if the transport cares)
 	 * @param[in] timeout Whether the cause for the session closure is a timeout (this may interest transport plugins more) */
 	void (* const session_over)(janus_transport_session *transport, guint64 session_id, gboolean timeout);
+	/*! \brief Method to notify the transport plugin that a session it owned was claimed by another transport
+	 * \note A transport plugin should close the connection as a result of such an event
+	 * @param[in] transport Pointer to the transport session instance
+	 * @param[in] session_id The session ID that was closed (if the transport cares) */
+	void (* const session_claimed)(janus_transport_session *transport, guint64 session_id);
```
lminiero: I think `session_claimed` should have more information. Right now, this only tells the transport a session has been claimed. Anyway, it doesn't tell the transport whether the session is still handled by the same transport or not, and if so which one. I'm not exactly sure what we should add here so no need to do anything right now (I need to think about this), just wanted to raise the point to have this discussion.

janus.c
```diff
@@ -1027,8 +1027,8 @@ int janus_process_incoming_request(janus_request *request) {
 	} else if(!strcasecmp(message_text, "claim")) {
 		janus_mutex_lock(&session->mutex);
 		if(session->source != NULL) {
-			/* Give old transport a timeout -- is this the right thing to do? */
-			session->source->transport->session_over(session->source->instance, session->session_id, TRUE);
+			/* Notify transport that the session has been claimed */
+			session->source->transport->session_claimed(session->source->instance, session->session_id);
```
lminiero: This is partly related to the point above. My feeling is that both `session_over` and `session_claimed` here make sense, even though it might be ambiguous and confusing (what if the transport really gets rid of the session in the first callback, and involves the core? it shouldn't though). Besides, `session_over` might close the previous transport, which is not idead.

At any rate, in `session_claimed` I wouldn't put the *old* transport, but the *new* one, to tell the transport plugin which trasport instance is tied to the session *now*. If we want to avoid `session_over` closing transports in that case, we might change its signature as well to avoid it when reclaiming is involved. Does this make sense?

transports/janus_http.c
```diff
@@ -1062,6 +1064,11 @@ void janus_http_session_over(janus_transport_session *transport, guint64 session
 	janus_mutex_unlock(&sessions_mutex);
 }
 
+void janus_http_session_claimed(janus_transport_session *transport, guint64 session_id) {
+	/* Session has been claimed -- is there anything to do here? */
+	JANUS_LOG(LOG_WARN, "Session has been claimed: (%"SCNu64"), has been claimed\n", session_id);
```
lminiero: This is where I'd update the `sessions` hashmap, for instance: whether the session existed in the map or not, a reclaim here would tell the transport "look, this session exists, and it's handled by this new transport". If it existed, it would sever the relationship between the old transport and the session ID; in both cases, long poll requests for the sessions wouldn't fail, as we would have a mapping there.

transports/janus_pfunix.c
```diff
@@ -441,6 +443,22 @@ void janus_pfunix_session_over(janus_transport_session *transport, guint64 sessi
 	janus_mutex_unlock(&clients_mutex);
 }
 
+void janus_pfunix_session_claimed(janus_transport_session *transport, guint64 session_id) {
+	/* Someone else has claimed the session: close the connection */
```
lminiero: Apart from the fact that the transport instance here would be the new one, and not the old, why close the previous transport connection just because a session has been claimed by another transport? I see legitimate cases where a different transport might claim a session and leave the old transport alive (e.g., it's still handling other sessions). I'd only use the reclaim stuff for handing sessions over.

I guess this was motivated by how session timeouts currently do close connections in `session_over`, which is why I think we should envisage the claim scenario as different.

transports/janus_websockets.c
```diff
@@ -834,6 +836,21 @@ void janus_websockets_session_over(janus_transport_session *transport, guint64 s
 	janus_mutex_unlock(&transport->mutex);
 }
 
+void janus_websockets_session_claimed(janus_transport_session *transport, guint64 session_id) {
+	/* Someone else has claimed the session: close the connection */
```
lminiero: See above, I wouldn't tear down the transport.

transports/transport.h
```diff
@@ -221,6 +221,11 @@ struct janus_transport {
 	 * @param[in] session_id The session ID that was closed (if the transport cares)
 	 * @param[in] timeout Whether the cause for the session closure is a timeout (this may interest transport plugins more) */
 	void (* const session_over)(janus_transport_session *transport, guint64 session_id, gboolean timeout);
+	/*! \brief Method to notify the transport plugin that a session it owned was claimed by another transport
+	 * \note A transport plugin should close the connection as a result of such an event
+	 * @param[in] transport Pointer to the transport session instance
```
lminiero: We should clarify here that we're referring to the new transport instance.

lminiero:
PS: it shows a couple of comments as outdated (which is not the case), and hides them, so please expand them manually in case as they're important IMHO.

janus.c
geigev: I think I understand. Calling both of them seems a bit strange to me since the session isn't actually over, rather, it has just been reclaimed: a bit ambiguous as you stated. My initial thought is to pass both the old and new transport to `session_claimed` and handle all the relevant logic in that callback, but maybe I'm missing something and there is a good reason to call `session_over` as well?

lminiero: I think it's a good idea to keep them both and expand `session_over` to fix the ambiguity, mostly for the case where the old and new transports are of a different nature (e.g., you reclaimed an HTTP transport from WebSockets). In this case, a new boolean flag in `session_over` called something like `claimed` might tell the first transport plugin that yes, this is over for this transport, but only because someone claimed it, and so for this specific case the old transport should not be disconnected (for `reclaim=FALSE` the behaviour might remain the one we have). Then, a `session_reclaimed` on the second transport plugin (which might be the same as the first if the claim is done, e.g., from WebSockets to WebSockets) would tell it that the session should now be handled. In case of HTTP, this would mean:

1. session_over(old_transport, sessionId, claim=TRUE) --> remove sessionId from map
2. session_reclaimed(new_transport, sessionId) --> map sessionId to new transport

This is probably a cleaner way of handling the scenario. Adding both old and new transport to `session_reclaimed` could cause problems when switching transport protocol, as in that case the transport plugin would be passed a pointer that belongs to another plugin.

geigev:
Just pushed some updates based on Lorenzo's helpful feedback and suggestions.
This is not ready yet, and I need to test it. Just wanted to gather my thoughts here while things were fresh in my head.

The initial PR did not properly consider the various possibilities for "claims" and the unique logic necessary for certain transport plugins. I believe this PR is now cognizant of the following:
- Potential that a transport instance may be handling multiple sessions
- Claim may emanate from a different transport plugin (websocket->http, pfunix-> rabbit etc.)
- Claim may emanate from the same transport plugin (http->http)

‌As suggested by Lorenzo, the `session_over` callback now includes a boolean called `claimed`. It is also passed its old transport instance (like before). This allows transport plugins to take the necessary steps to deal with "claim" events. Most transport plugins should simply return immediately if `claimed` is true. The http transport is unique and removes the sessionId from its hash table

A new callback, `session_claimed`, was also added. It is passed its new transport instance. This notifies the transport plugin that it has claimed a session. For most transport plugins, no steps are required, but the http transport must create an entry in it hashtable for the session that it claimed. I think this is required because of the way long polling works. Other transports can simply return because the core should handle all scenarios. Initially I was concerned that the `gateway->transport_gone` callback would be problematic: transports holding a reference to session claimed by someone else will try clean up the session. I believe the `if(session->source && session->source->instance == transport)` check stops this from occurring.

lminiero:
Excellent, thanks again for making all these fixes so quickly! I'll give it a test later today and let you know how it goes :+1:

lminiero:
Didn't have time yesterday, will try today, sorry about that... in the meanwhile, could you fill in the form at the end of [this page](https://cla.conf.meetecho.com/janus/) to sign our CLA? This is needed to get this merged eventually. Thanks!

geigev:
@lminiero I applied the patch and submitted the CLA form. Thanks for testing it and all your help!

I've done some testing for cross-transport session claims ("session hot potato") and everything seems solid. I'd still like to run a test that has an actual videoroom running and isn't simply passing around an empty session. I'll try to do this today or tomorrow.

lminiero:
@geigev any update on those last few tests you were planning? If you can confirm those are ok, I'm ok with merging! :+1:

geigev:
@lminiero It's been working well for me the last few days in my actual application. Feel free to merge :+1:

anisha-agrawal:
@lminiero @geigev I wanted to use the reclaim session feature to handle cases where websocket connection is lost on poor network, or network loss, network switch etc. In such cases we can either reconnect via the reclaim session feature, or we can do a complete re-initialization ( rejoin room by creating a new session, attaching to videoroom plugin and then joining room with same participant id and publishing and subscribing media). In the first option when we reclaim the session, I dont get the events which were missed when the websocket was disconnected, which results in inconsistent state at client side. I am hence unsure if I am approaching the problem correctly, I have 2 questions - 
1. Why was auto reconnection not made part of the client sdk and left upto the sdk integrator to deal with - here we have to handle consecutive keepalive ack misses as well as websocket errors to understand websocket connection loss and that reconnection is needed.
2. What is the right approach to do reconnection

PR #1205, opened 2018-04-12
Automatic changes of layers/substreams when simulcast/SVC are involved (VideoRoom)

lminiero:
As the title suggests, this is an attempt to implement automatic changes on spatial/temporal layers and substreams when a publisher is using Simulcast or VP9-SVC in the VideoRoom plugin. Notice it changes nothing if you're NOT doing Simulcast or VP9-SVC.

As you may know, at the moment these changes are manual, done via explicit `configure` requests sent on a subscriber's handle, e.g., as done [here](https://github.com/meetecho/janus-gateway/blob/master/html/videoroomtest.js#L644) in the VideoRoom demo with ad-hoc buttons when Simulcast is enabled. Obviously, the idea behind the manual triggers was not that people should choose it manually, but that changes could be triggered programmatically depending on some events at an application level: e.g., you get too many slow link events, you pick a lower layer to see if things get better. This patch adds an option to try and make this automatic instead, using REMB feedback from the subscriber.

Basically, the idea is that for each publisher we compute the bitrate for all available layers and substreams: this way we know how much bandwidth they each consume. This info is also made available in Admin API requests for the publisher handle, in the plugin specific information, if you want to have a look at it. Then, when we get an incoming REMB feedback from subscribers ("I can only get up to 500kbps"), we can use that info to decide the best substream/layer to send back automatically: the idea is to start from the top, and go down until we find a stream with a bitrate that SHOULD fit in the REMB constraints.

This was conceived as an optional, but dynamic, feature. It's optional because I still want to allow developers to have more control on the feature themselves, if they want (e.g., the above mentioned example of slow link events triggering changes, and other policies to try and ramp up again). Enabling it is as simple as specifying an `autochange:true` property either when creating a new subscriber (do automatic changes since the beginning) or in subsequent `configure` requests (switch to automatic now). Notice that when you send a request to explicitly switch to a specific substream or layer (like when you click one of the related existing buttons in the demo), the autochange mode is disabled, so if you want to restore it again you'll need a new configure, e.g.:

    feeds[index].send({message: { request: "configure", autochange: true }});

That said, it is far from perfect at the moment: while computing the publishers' bitrates seems fine, for what concerns subscribers, at least in my local tests, I seem to get ambiguous feedback from REMB at times (which may indeed be caused by the fact I was doing all my tests on the same machine, and so CPU usage and BWE may play a role here), which can result in choosing suboptimal streams when better options are available. At times I didn't get REMB feedback at all. I'll have to make some more tests with multiple machines.

Nevertheless, I wanted to share it already so that people could play with it and hopefully give feedback to improve how it currently works.


lminiero:
Closing this as REMB are a no go, and I'll have to think of a different way to do this properly.

PR #1209, opened 2018-04-17
RTP clock skew detection in videoroom plugin.

atoppi:
This PR introduces a clock skew compensation algorithm for publishers' media in videoroom plugin.
The rationale behind the PR is the same described in #1153, the only difference being that RTP is now coming from a WebRTC source and the compensation acts on every publisher inside a videoroom.
The same functions from #1153 are re-used. We added an extra RTP context for participants to keep track of RTP timestamps and sequence numbers.
Beware that the current implementation does not support skew detection in simulcasting. Once proved that this PR works well, adding the support to simulcasting should be quite a easy task though.


lminiero:
Closing as too much changed. We can revisit this again later on.

PR #1212, opened 2018-04-19
added FullHD, 4K streams support for getUserMedia via 'fhdres' and '4kres' presets

soulfly:
I want to use FullHD cameras with Janus, so this pull requests address it.
Also 4K support is added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ Makefile.in
 
 .deps
 .dirstamp
+
+# OS X
+.DS_Store
--- a/html/janus.js
+++ b/html/janus.js
@@ -1872,11 +1872,21 @@ function Janus(gatewayCallbacks) {
 							height = 180;
 							maxHeight = 180;
 							width = 320;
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -53,3 +53,6 @@ Makefile.in
 
 .deps
 .dirstamp
+
+# OS X
+.DS_Store
```
lminiero: What's this?

soulfly: I'm developing on macOS. This is kind of standard file generated almost in each macOS folder, something like desktop.ini in Microsoft Windows. https://en.wikipedia.org/wiki/.DS_Store
It's normal to add it to gitignore

lminiero:
Looks good! Can you filling in the form in our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge? (although I thought you had already done that) Thanks!

lminiero:
Thanks, merging then! :+1:

PR #1215, opened 2018-04-24
Opus FEC implementation proposal

mtbramo:
It's an implementation of 'Forward Error Correction' featue to improve  poor audio quality due to packets dropped or lost.
We use Opus encoder capability to embed redundant data and decoder one to rebuild lost packet from FEC data obtain in next packet it receives. We also use 'Packet lost concealment' feature by passing null information to decoder in case of multiple consecutive sequences losts.

Tested on 0.3.0 release, audio quality improvement verified (subjective).

### DIFF
--- a/conf/janus.plugin.audiobridge.cfg.sample
+++ b/conf/janus.plugin.audiobridge.cfg.sample
@@ -30,6 +30,7 @@
 								; if this key is provided in the request
 ;events = no					; Whether events should be sent to event
 								; handlers (default is yes)
+use_fec = false                 ; enable or disable FEC feature (opus)
 
 [1234]
 description = Demo Room
--- a/conf/janus.plugin.audiobridge.cfg.sample
+++ b/conf/janus.plugin.audiobridge.cfg.sample
@@ -30,7 +30,7 @@
 								; if this key is provided in the request
 ;events = no					; Whether events should be sent to event
 								; handlers (default is yes)
-use_fec = false                 ; enable or disable FEC feature (opus)
+use_fec = false					; enable or disable FEC feature (opus)
 
 [1234]
 description = Demo Room
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -635,6 +635,9 @@ rtp_forward_always_on = true|false, whether silence should be forwarded when the
 #define JANUS_AUDIOBRIDGE_AUTHOR			"Meetecho s.r.l."
 #define JANUS_AUDIOBRIDGE_PACKAGE			"janus.plugin.audiobridge"
 
+#define MIN_SEQUENTIAL 						2
+#define MAX_MISORDER						50
+
 /* Plugin methods */
 janus_plugin *create(void);
 int janus_audiobridge_init(janus_callbacks *callback, const char *config_path);
@@ -827,6 +830,7 @@ typedef struct janus_audiobridge_room {
 static GHashTable *rooms;
 static janus_mutex rooms_mutex = JANUS_MUTEX_INITIALIZER;
 static char *admin_key = NULL;
+static gboolean use_fec = FALSE;
 
 typedef struct janus_audiobridge_session {
 	janus_plugin_session *handle;
@@ -868,6 +872,9 @@ typedef struct janus_audiobridge_participant {
 	/* Opus stuff */
 	OpusEncoder *encoder;		/* Opus encoder instance */
 	OpusDecoder *decoder;		/* Opus decoder instance */
+	uint16_t expected_seq;          /* Expected sequence number */
+	uint16_t probation; 		/* Used to determine new ssrc validity */
+	uint32_t last_timestamp;	/* Last in seq timestamp */
 	gboolean reset;				/* Whether or not the Opus context must be reset, without re-joining the room */
 	GThread *thread;			/* Encoding thread for this participant */
 	janus_recorder *arc;		/* The Janus recorder instance for this user's audio, if enabled */
@@ -1137,7 +1144,6 @@ typedef struct wav_header {
 /* Opus settings */		
 #define	BUFFER_SAMPLES	8000
 #define	OPUS_SAMPLES	160
-#define USE_FEC			0
 #define DEFAULT_COMPLEXITY	4
 
 
@@ -1321,6 +1327,10 @@ int janus_audiobridge_init(janus_callbacks *callback, const char *config_path) {
 		if(!notify_events && callback->events_is_enabled()) {
 			JANUS_LOG(LOG_WARN, "Notification of events to handlers disabled for %s\n", JANUS_AUDIOBRIDGE_NAME);
 		}
+		janus_config_item *fec = janus_config_get_item_drilldown(config, "general", "use_fec");
+		use_fec = fec && fec->value && janus_is_true(fec->value);
+		JANUS_LOG(LOG_VERB, "FEC: %s\n", use_fec ? "enable" : "disable");
+
 		/* Iterate on all rooms */
 		GList *cl = janus_config_get_categories(config);
 		while(cl != NULL) {
@@ -2757,12 +2767,12 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 			int error = 0;
 			OpusDecoder *decoder = opus_decoder_create(participant->room->sampling_rate, 1, &error);
 			if(error != OPUS_OK) {
-				JANUS_LOG(LOG_ERR, "Error resetting Opus decoder...\n");
+				JANUS_LOG(LOG_ERR, "(participant %"SCNu64") Error resetting Opus decoder...\n", participant->user_id);
 			} else {
 				if(participant->decoder)
 					opus_decoder_destroy(participant->decoder);
 				participant->decoder = decoder;
-				JANUS_LOG(LOG_VERB, "Opus decoder reset\n");
+				JANUS_LOG(LOG_VERB, "(participant %"SCNu64") Opus decoder reset\n", participant->user_id);
 			}
 			participant->reset = FALSE;
 		}
@@ -2776,6 +2786,29 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 		/* We might check the audio level extension to see if this is silence */
 		pkt->silence = FALSE;
 		pkt->length = 0;
+
+		//First check if probation period
+		if(participant->probation == MIN_SEQUENTIAL){
+			participant->probation--;
+			participant->expected_seq = pkt->seq_number + 1;
+			JANUS_LOG(LOG_WARN, "(participant %"SCNu64") Probation started with ssrc = %"SCNu32", seq = %"SCNu16" \n", participant->user_id, ntohl(rtp->ssrc), pkt->seq_number);
+			g_free(pkt->data);
+			g_free(pkt);
+			return;
+		}
+		else if(participant->probation != 0){
+			//Decrease probation
+			participant->probation--;
+			//TODO: Reset probation if sequence number is incorrect and DSSRC also; must have a correct sequence
+			if(!participant->probation){
+				//Probation is ended
+			JANUS_LOG(LOG_WARN, "(participant %"SCNu64") Probation ended with ssrc = %"SCNu32", seq = %"SCNu16" \n", participant->user_id, ntohl(rtp->ssrc), pkt->seq_number);
+			}
+			participant->expected_seq = pkt->seq_number + 1;
+			g_free(pkt->data);
+			g_free(pkt);
+			return;
+		}
 
 		if(participant->extmap_id > 0) {
 			/* Check the audio levels, in case we need to notify participants about who's talking */
@@ -2831,16 +2864,93 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 		int plen = 0;
 		const unsigned char *payload = (const unsigned char *)janus_rtp_payload(buf, len, &plen);
 		if(!payload) {
-			JANUS_LOG(LOG_ERR, "[Opus] Ops! got an error accessing the RTP payload\n");
+			JANUS_LOG(LOG_ERR, "(participant %"SCNu64")[Opus] Ops! got an error accessing the RTP payload\n", participant->user_id);
 			g_free(pkt->data);
 			g_free(pkt);
 		  	participant->working = FALSE;
 			return;
 		}
-		pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, BUFFER_SAMPLES, USE_FEC);
+
+
+		//Check sequence number received, verify if relevant to expected one.
+		if(pkt->seq_number == participant->expected_seq){
+			/* regular decode */
+            //output_samples = max_frame_size;
+            pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, BUFFER_SAMPLES, 0);
+
+            //update last_timestamp
+            participant->last_timestamp = pkt->timestamp;
+
+            //Increment according to previous seq_number
+			participant->expected_seq = pkt->seq_number + 1;
+		}
+		else if(pkt->seq_number > participant->expected_seq){
+			//Sequence(s) losts
+			uint16_t gap = pkt->seq_number - participant->expected_seq;
+			JANUS_LOG(LOG_VERB, "(participant %"SCNu64"),   %"SCNu16" sequence(s) lost, sequence = %"SCNu16",  expected seq = %"SCNu16" \n", participant->user_id, gap, pkt->seq_number, participant->expected_seq);
+
+			//Use FEC if sequence lost < DEFAULT_PREBUFFERING
+			uint16_t start_lost_seq = participant->expected_seq;
+			if(use_fec && gap < DEFAULT_PREBUFFERING){
+				for(uint8_t i = 1; i <= gap ; i++){
+					int32_t output_samples;
+					janus_audiobridge_rtp_relay_packet *lost_pkt = g_malloc(sizeof(janus_audiobridge_rtp_relay_packet));
+					lost_pkt->data = g_malloc0(BUFFER_SAMPLES*sizeof(opus_int16));
+					lost_pkt->ssrc = 0;
+					lost_pkt->timestamp = participant->last_timestamp + (i * 960);
+					lost_pkt->seq_number = start_lost_seq++;
+					lost_pkt->silence = FALSE;
+					lost_pkt->length = 0;
+
+					JANUS_LOG(LOG_VERB, "(participant %"SCNu64"), Prepare lost packet with seq = %"SCNu16", timestamp = %"SCNu32" \n", participant->user_id, lost_pkt->seq_number, lost_pkt->timestamp);
+					if(i == gap){
+						/* attempt to decode with in-band FEC from next packet */
+                		opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+                		JANUS_LOG(LOG_VERB, "(participant %"SCNu64",   FEC(1) output_samples =  %"SCNu32" \n", participant->user_id, output_samples);
+                		lost_pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)lost_pkt->data, output_samples, 1);
+                	}
+                	else{
+                		opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+                		JANUS_LOG(LOG_VERB, "(participant %"SCNu64"),   FEC(NULL) output_samples =  %"SCNu32" \n", participant->user_id, output_samples);
+                    	lost_pkt->length = opus_decode(participant->decoder, NULL, plen, (opus_int16 *)lost_pkt->data, output_samples, 1);
+                	}
+
+					if(lost_pkt->length < 0) {
+						JANUS_LOG(LOG_ERR, "(participant %"SCNu64")[Opus] Ops! got an error decoding the Opus frame: %"SCNu16" (%s)\n", participant->user_id, lost_pkt->length, opus_strerror(lost_pkt->length));
+						g_free(lost_pkt->data);
+						g_free(lost_pkt);
+						return;
+					}
+   					/* Enqueue the decoded frame */
+					janus_mutex_lock(&participant->qmutex);
+					/* Insert packets sorting by sequence number */
+					participant->inbuf = g_list_insert_sorted(participant->inbuf, lost_pkt, &janus_audiobridge_rtp_sort);
+					janus_mutex_unlock(&participant->qmutex);
+				}
+			}
+			//then regular decode
+			pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, BUFFER_SAMPLES, 0); //NO FEC
+			//Increment according to previous seq_number
+			participant->expected_seq = pkt->seq_number + 1;
+		}else{
+			//In late sequence or sequence wrapped
+			//Check first if sequence wrapped
+			if((participant->expected_seq - pkt->seq_number) > MAX_MISORDER){
+				JANUS_LOG(LOG_VERB, "(participant %"SCNu64"),   SN WRAPPED seq =  %"SCNu16", expected_seq =  %"SCNu16" \n", participant->user_id, pkt->seq_number, participant->expected_seq);
+				participant->expected_seq = pkt->seq_number + 1;
+			}
+			else{
+				JANUS_LOG(LOG_WARN, "(participant %"SCNu64"),   IN LATE SN seq =  %"SCNu16", expected_seq =  %"SCNu16" \n", participant->user_id, pkt->seq_number, participant->expected_seq);
+			}
+			g_free(pkt->data);
+			g_free(pkt);
+			return;
+		}
+
+
 		participant->working = FALSE;
 		if(pkt->length < 0) {
-			JANUS_LOG(LOG_ERR, "[Opus] Ops! got an error decoding the Opus frame: %d (%s)\n", pkt->length, opus_strerror(pkt->length));
+			JANUS_LOG(LOG_ERR, "(participant %"SCNu64")[Opus] Ops! got an error decoding the Opus frame: %"SCNu16" (%s)\n", participant->user_id, pkt->length, opus_strerror(pkt->length));
 			g_free(pkt->data);
 			g_free(pkt);
 			return;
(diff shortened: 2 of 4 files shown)


lminiero:
Thanks! I'll try and do a review ASAP :+1: 

In the meanwhile, could you sign our CLA by filling the form [at the end of this page](https://cla.conf.meetecho.com/janus/)? This is needed before we can merge.

lminiero:
Can you fix the indentation before I review? It's quite hard to follow the code, especially in the `opus_decode` changes... Thx!

mtbramo:
Sorry, I have update the indentation. Hope that it will be OK!

lminiero:
It still seems a bit broken to me in some parts, but it is more readable now; I'll address those nits as part of the review when I look into it :+1:

conf/janus.plugin.audiobridge.cfg.sample
```diff
@@ -30,6 +30,7 @@
 								; if this key is provided in the request
 ;events = no					; Whether events should be sent to event
 								; handlers (default is yes)
+use_fec = false                 ; enable or disable FEC feature (opus)
```
lminiero: Please use tabs and not spaces before the comment (as the other lines).

plugins/janus_audiobridge.c
```diff
@@ -868,6 +872,9 @@ typedef struct janus_audiobridge_participant {
 	/* Opus stuff */
 	OpusEncoder *encoder;		/* Opus encoder instance */
 	OpusDecoder *decoder;		/* Opus decoder instance */
+	uint16_t expected_seq;      /* Expected sequence number */
+	uint16_t probation; 		/* Used to determine new ssrc validity */
+	uint32_t last_timestamp;	/* Last in seq timestamp */
```
lminiero: Same here: tabs, not spaces (definitely an issue for `expected_seq` at the moment).

plugins/janus_audiobridge.c
```diff
@@ -1321,6 +1327,10 @@ int janus_audiobridge_init(janus_callbacks *callback, const char *config_path) {
 		if(!notify_events && callback->events_is_enabled()) {
 			JANUS_LOG(LOG_WARN, "Notification of events to handlers disabled for %s\n", JANUS_AUDIOBRIDGE_NAME);
 		}
+		janus_config_item *fec = janus_config_get_item_drilldown(config, "general", "use_fec");
+		use_fec = fec && fec->value && janus_is_true(fec->value);
+		JANUS_LOG(LOG_VERB, "FEC: %s\n", use_fec ? "enable" : "disable");
```
lminiero: Why doing a global setting for FEC? Wouldn't it be better to have a per room, or even better a per-participant, FEC property? Ideally, this could be done depending on the SDP negotiation: if the user offers FEC in the SDP, then we enable it for them, otherwise we don't. What do you think about that?

As a side note, I guess the log line there is temporary and just for debug purposes?

plugins/janus_audiobridge.c
```diff
@@ -2757,12 +2767,12 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 			int error = 0;
 			OpusDecoder *decoder = opus_decoder_create(participant->room->sampling_rate, 1, &error);
 			if(error != OPUS_OK) {
-				JANUS_LOG(LOG_ERR, "Error resetting Opus decoder...\n");
+				JANUS_LOG(LOG_ERR, "(participant %"SCNu64") Error resetting Opus decoder...\n", participant->user_id);
```
lminiero: Unless they're just there for temporary debugging purposes, please remove all those `(participant %"SCNu64")` prefixes you added. I guess in some places it might be useful to have, but this format is not in line with how we prefix such things normally. I can take care of that as part of a different commit.

plugins/janus_audiobridge.c
```diff
@@ -2757,12 +2767,12 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
 			} else {
 				if(participant->decoder)
 					opus_decoder_destroy(participant->decoder);
 				participant->decoder = decoder;
-				JANUS_LOG(LOG_VERB, "Opus decoder reset\n");
+				JANUS_LOG(LOG_VERB, "(participant %"SCNu64") Opus decoder reset\n", participant->user_id);
```
lminiero: See above.

plugins/janus_audiobridge.c
```diff
@@ -2777,6 +2787,28 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 		pkt->silence = FALSE;
 		pkt->length = 0;
 
+		//First check if probation period
```
lminiero: Please change all comments that use `//` to `/*   */` instead. Just writing this here, please fix all occurrences.

plugins/janus_audiobridge.c
```diff
@@ -2777,6 +2787,28 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+		if(participant->probation == MIN_SEQUENTIAL) {
+			participant->probation--;
+			participant->expected_seq = pkt->seq_number + 1;
+			JANUS_LOG(LOG_WARN, "(participant %"SCNu64") Probation started with ssrc = %"SCNu32", seq = %"SCNu16" \n", participant->user_id, ntohl(rtp->ssrc), pkt->seq_number);
```
lminiero: The `pkt` instance contains the SSRC as well, no need to `ntohl(rtp->ssrc)` again.

plugins/janus_audiobridge.c
```diff
@@ -2777,6 +2787,28 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+			g_free(pkt->data);
+			g_free(pkt);
+			return;
+		} else if(participant->probation != 0) {
+			//Decrease probation
+			participant->probation--;
+			//TODO: Reset probation if sequence number is incorrect and DSSRC also; must have a correct sequence
+			if(!participant->probation){
+				//Probation is ended
+                JANUS_LOG(LOG_WARN, "(participant %"SCNu64") Probation ended with ssrc = %"SCNu32", seq = %"SCNu16" \n", participant->user_id, ntohl(rtp->ssrc), pkt->seq_number);
```
lminiero: Broken indentation (use tabs, not spaces).

plugins/janus_audiobridge.c
```diff
@@ -2831,16 +2863,90 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
-			JANUS_LOG(LOG_ERR, "[Opus] Ops! got an error accessing the RTP payload\n");
+			JANUS_LOG(LOG_ERR, "(participant %"SCNu64")[Opus] Ops! got an error accessing the RTP payload\n", participant->user_id);
 			g_free(pkt->data);
 			g_free(pkt);
 		  	participant->working = FALSE;
 			return;
 		}
-		pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, BUFFER_SAMPLES, USE_FEC);
+
+
```
lminiero: Please remove one of the extra new lines here.

plugins/janus_audiobridge.c
```diff
@@ -2831,16 +2863,90 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+		//Check sequence number received, verify if relevant to expected one.
+		if(pkt->seq_number == participant->expected_seq) {
+			/* regular decode */
+            //output_samples = max_frame_size;
+            pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, BUFFER_SAMPLES, 0);
+
+            //update last_timestamp
+            participant->last_timestamp = pkt->timestamp;
+
+            //Increment according to previous seq_number
```
lminiero: All the above lines have broken indentation (use tabs, not spaces).

plugins/janus_audiobridge.c
```diff
@@ -2831,16 +2863,90 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+			participant->expected_seq = pkt->seq_number + 1;
+		} else if(pkt->seq_number > participant->expected_seq) {
+			//Sequence(s) losts
+			uint16_t gap = pkt->seq_number - participant->expected_seq;
+			JANUS_LOG(LOG_VERB, "(participant %"SCNu64"),   %"SCNu16" sequence(s) lost, sequence = %"SCNu16",  expected seq = %"SCNu16" \n", participant->user_id, gap, pkt->seq_number, participant->expected_seq);
+
+			//Use FEC if sequence lost < DEFAULT_PREBUFFERING
+			uint16_t start_lost_seq = participant->expected_seq;
+			if(use_fec && gap < DEFAULT_PREBUFFERING) {
+                for(uint8_t i = 1; i <= gap ; i++) {
```
lminiero: Broken indentation (use tabs, not spaces).

plugins/janus_audiobridge.c
```diff
@@ -2831,16 +2863,90 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+						/* attempt to decode with in-band FEC from next packet */
+                		opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+                		JANUS_LOG(LOG_VERB, "(participant %"SCNu64",   FEC(1) output_samples =  %"SCNu32" \n", participant->user_id, output_samples);
+                		lost_pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)lost_pkt->data, output_samples, 1);
+                	} else {
+                		opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+                		JANUS_LOG(LOG_VERB, "(participant %"SCNu64"),   FEC(NULL) output_samples =  %"SCNu32" \n", participant->user_id, output_samples);
+                    	lost_pkt->length = opus_decode(participant->decoder, NULL, plen, (opus_int16 *)lost_pkt->data, output_samples, 1);
+                	}
+
```
lminiero: All the above lines have broken indentation (use tabs, not spaces).

plugins/janus_audiobridge.c
```diff
@@ -2831,16 +2863,90 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+					lost_pkt->length = 0;
+
+					JANUS_LOG(LOG_VERB, "(participant %"SCNu64"), Prepare lost packet with seq = %"SCNu16", timestamp = %"SCNu32" \n", participant->user_id, lost_pkt->seq_number, lost_pkt->timestamp);
+					if(i == gap) {
+						/* attempt to decode with in-band FEC from next packet */
+                		opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+                		JANUS_LOG(LOG_VERB, "(participant %"SCNu64",   FEC(1) output_samples =  %"SCNu32" \n", participant->user_id, output_samples);
+                		lost_pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)lost_pkt->data, output_samples, 1);
+                	} else {
+                		opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
```
lminiero: What's the purpose of this call, here? Isn't this the non-FEC branch?

plugins/janus_audiobridge.c
```diff
@@ -2831,16 +2863,90 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+				participant->expected_seq = pkt->seq_number + 1;
+			} else {
+				JANUS_LOG(LOG_WARN, "(participant %"SCNu64"),   IN LATE SN seq =  %"SCNu16", expected_seq =  %"SCNu16" \n", participant->user_id, pkt->seq_number, participant->expected_seq);
+			}
+			g_free(pkt->data);
+			g_free(pkt);
+			return;
+		}
+
+
```
lminiero: Please remove extra new-line.

plugins/janus_audiobridge.c
```diff
@@ -2851,25 +2957,27 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 		participant->inbuf = g_list_insert_sorted(participant->inbuf, pkt, &janus_audiobridge_rtp_sort);
 		if(participant->prebuffering) {
 			/* Still pre-buffering: do we have enough packets now? */
-			if(g_list_length(participant->inbuf) == DEFAULT_PREBUFFERING) {
+			if(g_list_length(participant->inbuf) > DEFAULT_PREBUFFERING) { //If we have packets lost during prebuffering ... best to test > instead of == to stop prebuffering
```
lminiero: Please put the comment in a new line below, inside the if.

plugins/janus_audiobridge.c
```diff
@@ -2851,25 +2957,27 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
-				JANUS_LOG(LOG_VERB, "Prebuffering done! Finally adding the user to the mix\n");
+				JANUS_LOG(LOG_VERB, "(participant %"SCNu64") Prebuffering done! Finally adding the user to the mix\n", participant->user_id);
 			} else {
-				JANUS_LOG(LOG_VERB, "Still prebuffering (got %d packets), not adding the user to the mix yet\n", g_list_length(participant->inbuf));
+				JANUS_LOG(LOG_VERB, "(participant %"SCNu64") Still prebuffering (got %d packets), not adding the user to the mix yet\n", participant->user_id, g_list_length(participant->inbuf));
 			}
 		} else {
 			/* Make sure we're not queueing too many packets: if so, get rid of the older ones */
-			if(g_list_length(participant->inbuf) >= DEFAULT_PREBUFFERING*2) {
+			if(g_list_length(participant->inbuf) >= DEFAULT_PREBUFFERING*3) {
```
lminiero: Why change the multiplier here?

plugins/janus_audiobridge.c
```diff
@@ -2851,25 +2957,27 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
 				gint64 now = janus_get_monotonic_time();
 				if(now - participant->last_drop > 5*G_USEC_PER_SEC) {
-					JANUS_LOG(LOG_VERB, "Too many packets in queue (%d > %d), removing older ones\n",
-						g_list_length(participant->inbuf), DEFAULT_PREBUFFERING*2);
+					JANUS_LOG(LOG_VERB, "(participant %"SCNu64") Too many packets in queue (%d > %d), removing older ones\n", participant->user_id,
+						g_list_length(participant->inbuf), DEFAULT_PREBUFFERING*3);
```
lminiero: And here too?

plugins/janus_audiobridge.c
```diff
@@ -2851,25 +2957,27 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
 					participant->last_drop = now;
 				}
 				while(g_list_length(participant->inbuf) > DEFAULT_PREBUFFERING*2) {
```
lminiero: The multiplier hasn't been changed here, for instance, so not sure if this will cause issues (haven't checked other parts of the code).

conf/janus.plugin.audiobridge.cfg.sample
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: I'm agree that the best solution is to only use SDP negotiation information and according to the value associate to FEC usage perform operation. 
I will do it in a second step if you want.

plugins/janus_audiobridge.c
mtbramo: remove this information but I can tell you that in some traces it's important to have participant info.

plugins/janus_audiobridge.c
mtbramo: same

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: SSRC value (in pkt) is previously set to 0 value.

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: No, here it's the case of audio loss concealment with repair technique associate to the codec (silence substitution ...)

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: We have actually some trouble associate to the size of inbuf (packet dropped); I have return the multiplier like before.

plugins/janus_audiobridge.c
mtbramo: done

plugins/janus_audiobridge.c
mtbramo: No, because I would like to create an "hysteresis" and remove a complete sequences (and  also reduce the end to end delay) one time than stay with maximum buffer size and remove time to time one packet.

lminiero:
@mtbramo sorry for the lack of feedback, I'm currently abroad so I'll do another review as soon as I come back :+1:

plugins/janus_audiobridge.c
```diff
@@ -2777,6 +2787,28 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 		pkt->silence = FALSE;
 		pkt->length = 0;
 
+		/* First check if probation period */
```
lminiero: What's the purpose of this probation, here? It looks like you're discarding some packets without ever decoding or mixing them. This seems to be happening whether FEC is in use or not, and I don't understand why.

plugins/janus_audiobridge.c
```diff
@@ -2777,6 +2787,28 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+			participant->probation--;
+			participant->expected_seq = pkt->seq_number + 1;
+			JANUS_LOG(LOG_WARN, "Probation started with ssrc = %"SCNu32", seq = %"SCNu16" \n", ntohl(rtp->ssrc), pkt->seq_number);
+			g_free(pkt->data);
+			g_free(pkt);
+			return;
+		} else if(participant->probation != 0) {
+			/* Decrease probation */
+			participant->probation--;
+			/* TODO: Reset probation if sequence number is incorrect and DSSRC also; must have a correct sequence */
```
lminiero: Is that DSSRC a typo? Anyway, it's not clear what this other branch is doing either.

plugins/janus_audiobridge.c
```diff
@@ -2837,10 +2869,81 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+					participant->inbuf = g_list_insert_sorted(participant->inbuf, lost_pkt, &janus_audiobridge_rtp_sort);
+					janus_mutex_unlock(&participant->qmutex);
+				}
+			}
+			/* then regular decode (NO FEC) */
+			pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, BUFFER_SAMPLES, 0);
+			/* Increment according to previous seq_number */
+			participant->expected_seq = pkt->seq_number + 1;
+		} else {
+			/* In late sequence or sequence wrapped */
```
lminiero: Does this mean you drop RTP packets when the sequence number resets from 65535 to 0? That doesn't sound right...

plugins/janus_audiobridge.c
```diff
@@ -2866,10 +2969,12 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 						g_list_length(participant->inbuf), DEFAULT_PREBUFFERING*2);
 					participant->last_drop = now;
 				}
-				while(g_list_length(participant->inbuf) > DEFAULT_PREBUFFERING*2) {
+				while(g_list_length(participant->inbuf) > DEFAULT_PREBUFFERING) {
```
lminiero: Why removing the *2 here?

plugins/janus_audiobridge.c
```diff
@@ -2866,10 +2969,12 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
 					/* Remove this packet: it's too old */
 					GList *first = g_list_first(participant->inbuf);
 					janus_audiobridge_rtp_relay_packet *pkt = (janus_audiobridge_rtp_relay_packet *)first->data;
+					JANUS_LOG(LOG_ERR, "list length = %d, Remove sequence = %d\n",
+						g_list_length(participant->inbuf), pkt->seq_number);
```
lminiero: Why the LOG_ERR? This might happen a lot, and flood the logs.

plugins/janus_audiobridge.c
```diff
@@ -3931,8 +4039,8 @@ static void *janus_audiobridge_handler(void *data) {
 			answer->s_name = g_strdup(s_name);
 			/* Add a fmtp attribute */
 			janus_sdp_attribute *a = janus_sdp_attribute_create("fmtp",
-				"%d maxplaybackrate=%"SCNu32"; stereo=0; sprop-stereo=0; useinbandfec=0\r\n",
-					participant->opus_pt, participant->room->sampling_rate);
+				"%d maxplaybackrate=%"SCNu32"; stereo=0; sprop-stereo=0; useinbandfec=%s\r\n",
+					participant->opus_pt, participant->room->sampling_rate, use_fec ? "1" : "0");
```
lminiero: Why a string? It can be %d with `use_fec ? 1 : 0`.

plugins/janus_audiobridge.c
mtbramo: As you could see, I use expected sequence to implement FEC. So, you must be sure to start with a normal sequence (2 or 3 consecutive seq number) coming from one unique source. You must even go far from what I done here to be sure to get packet coming from only one distant source and also to detect a change in distant ssrc. It’s not important to drop 2 packets but more important is to start decoding correctly. It’s defense mechanism.

plugins/janus_audiobridge.c
mtbramo: No

plugins/janus_audiobridge.c
mtbramo: No,it’s not te case. I have checked sequence wrapped and doesn’t see any problem.

plugins/janus_audiobridge.c
mtbramo: In normal situation, it doesn’t happens a lot. Now, if you want to remove it or change it ...
Today, it help to detect audio quality problem.

plugins/janus_audiobridge.c
mtbramo: Ok.

Thank you for your review of the code. I just would like to know if you have tested the FEC and verified that it improve audio quality. And I would like to know if you have found some interest in the implementation.

I’m on holiday since 14 may, so sorry for the lack of feedback during that time.

plugins/janus_audiobridge.c
lminiero: The problem is that this is not done only for FEC, but even when FEC is disabled, which I don't think is the right way to go. Anyway, we can discuss this later on.

plugins/janus_audiobridge.c
lminiero: Then what does DSSRC stand for? :smile:

plugins/janus_audiobridge.c
lminiero: But the only way out of that if branch is a `return`...

plugins/janus_audiobridge.c
lminiero: I'd rather change the verbosity level.

plugins/janus_audiobridge.c
lminiero: I had no time to test anything yet, sorry. It's definitely of interest, but of course so many changes in something that currently works require a deep review, which is why I'm nitpicking pretty much everything... Thanks again and enjoy your vacation! I'll be abroad for a conference on the 14th, so I'll get back to you a couple of days later probably.

lminiero:
@mtbramo I hope you enjoyed your vacation!

I did a few tests with your patch, and it does seem to make audio quality better in case of packet loss, even considerable packet loss :+1: 
I used the [comcast](https://github.com/tylertreat/comcast) tool to simulate different levels of packet loss (e.g., 10% or 30%) and with FEC enabled it always sounded decent, while under the same conditions the code in master would result in garbled audio. This seems to confirm the good work you did here!

One thing I noticed is that this can cause some issues with the post-processor, but this isn't something you need to worry about. Specifically, and just FYI, if you record an individual contribution by a participant to an AudioBridge (which is something you can do via a `configure` with `record:true`), in case there are many losses as simulated above the resulting audio file does NOT recover gracefully as the mixer does. This may be because the post-processor never decodes anything (it just puts the Opus packets in an .opus format file), and when a packet loss is found it inserts silence frames to compensate. I guess that, since the media players don't know FEC is in the packets, they don't use the redundant information, and so the quality suffers as a result. I'll have to check if .ogg/.opus as a format has a way to say "use FEC", but I don't think so: if not, I guess there's nothing we can do about that. Anyway, these are things I'll worry about after we merge, as it's only collateral.

That said, I think this is almost ready to be merged. I'd ask you to only take care of a few things to finish:

1. sign our [CLA](https://cla.conf.meetecho.com/janus/) (this is needed for all contributions, there's a small form to fill in at the end of the page);
2. turn the global FEC property in something that can be evinced from the SDP negotiation (per participant FEC);
3. disable the probation period code if FEC is not enabled (see discussion in the code comments), unless you think there's a good reason for keeping it;
4. remove the log line, or change its verbosity level, related to the "list length" thing (we discussed this inline as well);
5. change the `useinbandfec` printf format to `%d` with `use_fec ? 1 : 0` (see discussion above).

This should be all trivial changes: especially 3, 4 and 5: I guess 2. may require a bit more of an effort but shouldn't be hard either, looking at how you organized the code already.

As soon as that's done, we'll be good to go and merge! Thanks again for this very useful effort :+1:

mtbramo:
Lorenzo, thank you for your feedback.

I will take your different point in account and perform the merge as soon as possible.
Thanks.

mtbramo:
Lorenzo, sorry for the late, I have to wait for an official response from company legal authority to let me contribute to Janus source code. 
I have made change concerning point 1,2,4 and 5. Concerning point 3, I let you see if you want that we removed completely the probation period; This probation period is just to initialize correctly the parameter expected_seq (when we get 2 consecutive packets). I doesn't understand also that we always start with a sequence number of 1, we should get here the remote starting seq number.

lminiero:
@mtbramo I see you already signed the CLA, though, is there any other obstacle to actually contribute this? As to the sequence number, starting from 1 is as a valid choice as starting randomly: we do it just to keep things simple.

mtbramo:
No there is no more obstacle to contribute. But let me know if I have other things to do how to proceed because I'm not an expert on GIT.

lminiero:
Oh, sorry, I meant in response to this:

> I have to wait for an official response from company legal authority to let me contribute to Janus source code

I guess you meant you _had_ to wait for that response, which you now got, apologies for the misunderstanding!

For the other changes, I haven't done a review of your latest changes yet, I'll do that ASAP. I do see theres's a conflicting file, though, so that should be resolved or we can't merge. If you're unfamiliar with how git works with that, it might be easier to try and apply your patch on a brand new pull request. Whatever works best for you!

lminiero:
Closing as replaced by #1275

PR #1216, opened 2018-04-25
[add repeat join->leave]:After the once "attach",thne publisher/subscriber can repeat join - > leave - > join...,

MinorUncle:
The problem with the videoroom plug-in is that if call "attach"->"join"->"leave"->"join", there will receive a "Already in as a publisher on this handle" error, but this is a necessary function if publisher/subscriber need to switch room and do not need to call attach/detach each time.This commit adds this function, which can repeat attach->join->leave->join...


MinorUncle:
This submission has some bug, but this demand is necessary.

lminiero:
Actually that's how it's supposed to work right now: that is, once a handle has been assigned a role and a user, it's forever, and it's not meant to be re-used. I didn't check your code, but in general trying to re-use the session will result in undefined behaviour.

lminiero:
PS: in case you want to submit a new PR in the future, please provide commit messages in English.

PR #1219, opened 2018-04-29
Allow only secure protocols/ciphers in janus_http

agclark81:
Set MHD_OPTION_HTTPS_PRIORITIES to "PFS:-VERS-TLS1.0:-VERS-TLS1.1:-3DES-CBC:-ARCFOUR-128" to disable insecure protocols and ciphers (e.g., SSL 3.0, TLS 1.0), since browsers that support WebRTC all support secure protocols/ciphers. Enables any Janus HTTPS endpoint to pass network scans and PCI validations by disabling insecure protocols and ciphers.


lminiero:
Can this cause issues with some clients? The ciphers endpoints use for WebRTC use a different stack than the one used for HTTPS, so I wouldn't want things to stop working for, e.g., mobile apps because of this.

agclark81:
All of the reference browsers (Chrome, Safari, Firefox, Edge) that support WebRTC all support TLS 1.2 and secure/non-weak ciphers, but I'm happy to re-submit this as a configuration parameter so that it's opt-in if that would be preferable. The default cipher string could be sent if the parameter isn't configured so that there is no behavior change for existing installations.

lminiero:
A configuration parameter would help, thanks! Typically it's not just browsers that talk to Janus, so having flexibility to override those defaults would be helpful.

lminiero:
Wondering if we should have the same in the WebSockets library too. Checking the [documentation](https://libwebsockets.org/lws-api-doc-master/html/structlws__context__creation__info.html) it looks like this can be done setting the `ssl_cipher_list` property, which I guess works pretty much the same way as `MHD_OPTION_HTTPS_PRIORITIES` does.

lminiero:
@agclark81 did you have time to make that configurable? As soon as that's ready, I think this is good to merge.

lminiero:
@agclark81 since it was a simple enough change, I made it configurable myself. I mentioned you in the commit description, since this was indeed your contribution. Closing this, thanks!

PR #1220, opened 2018-04-30
video delay when using simulcast.

hissinger:
we used the simulcast function to create a broadcast server with janus(janus.plugin.streaming).
we found that the video delay increases every time substream was changed. maybe it seems to be a problem with timestamp computation when changing substream.
we increased the timestamp by about 10 seconds, and the video jitter buffer of client (chrome) is initialized. As a result, no video delay occurs.
here is the debug log for chrome.
> [37426:43671:0430/180805.114427:WARNING:frame_buffer2.cc(227)] A frame about to be decoded is out of the configured delay bounds (11138 > 10000). Resetting the video jitter buffer.

and when changing the substream, the frame is discarded. here is the debug log for chrome.

> [37426: 35339: 0430 / 180412.274080: WARNING: frame_buffer2.cc (311)] Frame with (picture_id: spatial_id) (1000000000000000943: 0) has invalid frame references, dropping frame.

this causes the video to stop for several seconds. this also seems to be a problem that occurs when vp8 header information is initialized when substream is changed.


lminiero:
Does it work as expected when handling simulcast in the VideoRoom plugin, that is without the Streaming plugin step? If it does, it's likely that we're simply missing some of the code to handle switches there, since I added support to that plugin in a latter phase and we didn't test it much.

lminiero:
PS: your changes would break things even more, I'm afraid... I'll have a look when I get back, currently abroad for a conference.

lminiero:
@hissinger any feedback on my VideoRoom vs. Streaming question?

hissinger:
@lminiero first, when we explain the broadcast platform.

caster(chrome) -> janus(VideoRoom plugin) -> janus(Streaming plugin) -> watcher(chrome)

we use rtp_forward between videoroom and streaming. we do not use simulcast of VideoRoom without Streaming plugin.

and since only janus_streaming.c has been modified, it will not be a problem to deal with simulcast in VideoRoom plugin without Streaming plugin. 😀

lminiero:
I understand your scenario. What I'm asking is if you're getting the same delay when watching using the VideoRoom directly, or if it only happens when adding the Streaming plugin as an intermediary. If it works fine with the VideoRoom, it's clear that we're missing some steps in that other plugin, and that would give me hints on what to fix. I don't like the idea of messing with the timestamp as you did, and would like to look for a better solution instead.

hissinger:
there is no video delay in the new watcher when video delay occurs during simulcast.
and since we are changing the substream in the Streaming plugin, it does not seem to be a problem if VideoRoom works well. In the VideoRoom, you should also check if the video delay occurs and accumulates while changing the substream.

hissinger:
Increasing delay in chrome seems to be caused by an increase in the jitter value due to the timestamp recalculated by clockgettime() while changing the substream.
If the substream changes, I think it should be considered a new stream because ssrc changes.

lminiero:
Found a bug in the Streaming simulcast support and fixed it, let me know if it helps.

hissinger:
I did a test with the latest source. The result is that the video delay still occurs. The video has a delay of about 1 second, and the audio has no delay. 😂

lminiero:
Can't replicate. You still haven't told me if the delay is there when just watching with the VideoRoom.

hissinger:
we can not test with the VideoRoom plugin only. we are testing in the following environment.

caster(chrome) -> janus(VideoRoom plugin) -> janus(Streaming plugin) -> watcher(chrome)

are you testing with only the VideoRoom plugin?  i suggest you to repeat the test several times because video delay occurs only when changing substream.

lminiero:
I'm testing with both, and it works fine for me. That's why I'm asking you to test with VideoRoom only as well, since you're getting the delay in the Streaming viewers. I understand that's your environment, but I don't think it's really that problematic to test a VideoRoom-only simulcast, especially considering the default demo has that if you add `?simulcast=true` to the web page url.

hissinger:
I tested it with videoroomtest, but it was still delayed. The googTargetDelayMs value now exceeds 1k. But strangely, when I raised the bandwidth of caster to 2mb rather than 1mb, it seems that delay occurred. This is not certain.

Also, every time you change the substream, the video stoped and was corrupted very shortly.

andrew-khv:
Hello. Hissinger, you used to write that you've increased jitter buffer in Chrome. Can you help me? How did you do it?

PR #1222, opened 2018-05-03
Only log SCTP message bodies at >= HUGE level

mquander:
VERB log level is very useful for diagnosing a lot of things, but if you are using a data-traffic intensive application then SCTP message bodies are going to take up the vast majority of your VERB-level log. Unless there is some really specific use case for why it's good to log them, moving them to HUGE level seems reasonable.

(In addition, I imagine that data channel message bodies are likely to be more private or more security-relevant than a lot of the other metadata in the log, so that's another reason you might want to be more conservative about logging them. That's certainly true for our product.)


lminiero:
Make sense, thanks for taking care of this, merging!

PR #1225, opened 2018-05-04
Clear flag if datachannel is rejected by setting the port to 0.

fancycode:
The `JANUS_ICE_HANDLE_WEBRTC_DATA_CHANNELS` flag must be cleared if a datachannel gets rejected by setting the port to `0`. Also fixes logging of unsupported protocols.


lminiero:
Good catch, thanks for your contribution! Merging :+1:

Issue #1213, opened 2018-04-19
dtls: memcpy with negative size


tmatth:
Found with stress testing + ASAN:

```
==12==ERROR: AddressSanitizer: negative-size-param: (size=-2)
    #0 0x7f062e40e9a1 in __asan_memcpy (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x8c9a1)
    #1 0x7f062d42ebf5  (/lib/x86_64-linux-gnu/libcrypto.so.1.0.0+0x111bf5)
    #2 0x7f062d42db9b in BIO_write (/lib/x86_64-linux-gnu/libcrypto.so.1.0.0+0x110b9b)
    #3 0x42b7c7 in janus_dtls_bio_filter_write /tmp/janus-gateway/dtls-bio.c:133
    #4 0x7f062d42db9b in BIO_write (/lib/x86_64-linux-gnu/libcrypto.so.1.0.0+0x11
https://github.com/meetecho/janus-gateway0b9b)
    #5 0x7f062d786bf1  (/lib/x86_64-linux-gnu/libssl.so.1.0.0+0x25bf1)
    #6 0x7f062d7995f3  (/lib/x86_64-linux-gnu/libssl.so.1.0.0+0x385f3)
    #7 0x7f062d784a21  (/lib/x86_64-linux-gnu/libssl.so.1.0.0+0x23a21)
==12==AddressSanitizer: while reporting a bug found another one. Ignoring.
    #8 0x42983b in janus_dtls_srtp_send_alert /tmp/janus-gateway/dtls.c:899
    #9 0x45809c in janus_ice_send_thread /tmp/janus-gateway/ice.c:3193
    #10 0x7f062dc47bb4  (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x70bb4)
    #11 0x7f062c4ac6b9 in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76b9)
    #12 0x7f062c1e241c in clone (/lib/x86_64-linux-gnu/libc.so.6+0x10741c)

0x6290006cc211 is located 17 bytes inside of 16578-byte region [0x6290006cc200,0x6290006d02c2)
allocated by thread T122975 (iceloop 6107218) here:
    #0 0x7f062e41a602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
[...]

PR #1226, opened 2018-05-04
Remove unused "srtp_mutex".

fancycode:


lminiero:
Not really sure what this was for: I think it was probably from a long time ago, when we could still end up trying to SRTP encrypt from different threads, which is something that will never happen in the current code. Thanks for spotting this and fixing it!

fancycode:
You're right about SSL/SRTP.

#1213 looks similar, but it's `janus_ice_thread` racing with `janus_ice_send_thread`, where my crash was with SCTP (and BoringSSL instead of OpenSSL). But yes, using a SSL context from different threads will lead to strange errors.

lminiero:
That shouldn't be the case, though, as we're explicitly handling the thread safeness for OpenSSL in dtls.c: https://github.com/meetecho/janus-gateway/blob/master/dtls.c#L108

fancycode:
Hmm, according to the OpenSSL documentation it's *not* safe to use the same SSL context from different threads: https://www.openssl.org/docs/man1.0.2/crypto/threads.html

> Note that OpenSSL is not completely thread-safe, and unfortunately not all global resources have the necessary locks. Further, the thread-safety does not extend to things like multiple threads using the same SSL object at the same time.

lminiero:
Which is why we use the `CRYPTO_THREADID_set_callback` stuff, which AFAIK is supposed to take care of exactly that. We did have crashes due to race conditions in the past (the commented block I linked to in my previous comment has more info) that this approach fixed.

lminiero:
This might be one more reason to start working on using the libnice loop thread for outgoing media too, in order to get rid of the send thread we have now. I have been thinking about this these last few days, but still haven't investigated how complicated or messy it would be with the glib event loop. I already worked on something like this when prototyping my own ICE stack (something I never got to submit here, actually) but that used libevent2.

fancycode:
> Although I guess you mean that this may not be enough, as some structures may not have locks in place, and so would never be locked even in that case?

Yes, e.g. `SSL_write` / `SSL_shutdown` don't do any locking:
https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/ssl/ssl_lib.c#L1043
https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/ssl/ssl_lib.c#L1058

> This might be one more reason to start working on using the libnice loop thread for outgoing media too, in order to get rid of the send thread we have now. 

Sounds like a good idea.

PR #1228, opened 2018-05-08
add helper func to reduce videoroom code

mtdxc:
this pr don't add any new function, but for reduce redundancy code in videoroom plugin, I have test for build in mac and ubuntu. 

add helper func to reduce videoroom code
> janus_videoroom_codecstr
> janus_videoroom_reqfir

move janus_codec to public for common use in feature
janus_audiocodec -> janus_videoroom_audiocodec
janus_videocodec -> janus_videoroom_videocodec

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1252,137 +1252,6 @@ static GAsyncQueue *messages = NULL;
 static janus_videoroom_message exit_message;
 
 
-/* Payload types we'll offer internally */
-#define OPUS_PT		111
-#define ISAC32_PT	104
-#define ISAC16_PT	103
-#define PCMU_PT		0
-#define PCMA_PT		8
-#define G722_PT		9
-#define VP8_PT		96
-#define VP9_PT		101
-#define H264_PT		107
-
-typedef enum janus_videoroom_audiocodec {
-	JANUS_VIDEOROOM_NOAUDIO,	/* Publishers won't do any audio */
-	JANUS_VIDEOROOM_OPUS,		/* Publishers will have to/may use OPUS */
-	JANUS_VIDEOROOM_PCMU,		/* Publishers will have to/may use PCMU 8K */
-	JANUS_VIDEOROOM_PCMA,		/* Publishers will have to/may use PCMA 8K */
-	JANUS_VIDEOROOM_G722,		/* Publishers will have to/may use G.722 */
-	JANUS_VIDEOROOM_ISAC_32K,	/* Publishers will have to/may use ISAC 32K */
-	JANUS_VIDEOROOM_ISAC_16K	/* Publishers will have to/may use ISAC 16K */
-} janus_videoroom_audiocodec;
-static const char *janus_videoroom_audiocodec_name(janus_videoroom_audiocodec acodec) {
-	switch(acodec) {
-		case JANUS_VIDEOROOM_NOAUDIO:
-			return "none";
-		case JANUS_VIDEOROOM_OPUS:
-			return "opus";
-		case JANUS_VIDEOROOM_PCMU:
-			return "pcmu";
-		case JANUS_VIDEOROOM_PCMA:
-			return "pcma";
-		case JANUS_VIDEOROOM_G722:
-			return "g722";
-		case JANUS_VIDEOROOM_ISAC_32K:
-			return "isac32";
-		case JANUS_VIDEOROOM_ISAC_16K:
-			return "isac16";
-		default:
-			/* Shouldn't happen */
-			return "opus";
-	}
-}
-static janus_videoroom_audiocodec janus_videoroom_audiocodec_from_name(const char *name) {
-	if(name == NULL)
-		return JANUS_VIDEOROOM_NOAUDIO;
-	else if(!strcasecmp(name, "opus"))
-		return JANUS_VIDEOROOM_OPUS;
-	else if(!strcasecmp(name, "isac32"))
-		return JANUS_VIDEOROOM_ISAC_32K;
-	else if(!strcasecmp(name, "isac16"))
-		return JANUS_VIDEOROOM_ISAC_16K;
-	else if(!strcasecmp(name, "pcmu"))
-		return JANUS_VIDEOROOM_PCMU;
-	else if(!strcasecmp(name, "pcma"))
-		return JANUS_VIDEOROOM_PCMA;
-	else if(!strcasecmp(name, "g722"))
-		return JANUS_VIDEOROOM_G722;
-	JANUS_LOG(LOG_WARN, "Unsupported audio codec '%s'\n", name);
-	return JANUS_VIDEOROOM_NOAUDIO;
-}
-static int janus_videoroom_audiocodec_pt(janus_videoroom_audiocodec acodec) {
-	switch(acodec) {
-		case JANUS_VIDEOROOM_NOAUDIO:
-			return -1;
-		case JANUS_VIDEOROOM_OPUS:
-			return OPUS_PT;
-		case JANUS_VIDEOROOM_ISAC_32K:
-			return ISAC32_PT;
-		case JANUS_VIDEOROOM_ISAC_16K:
-			return ISAC16_PT;
-		case JANUS_VIDEOROOM_PCMU:
-			return PCMU_PT;
-		case JANUS_VIDEOROOM_PCMA:
-			return PCMA_PT;
-		case JANUS_VIDEOROOM_G722:
-			return G722_PT;
-		default:
-			/* Shouldn't happen */
-			return OPUS_PT;
-	}
-}
-
-typedef enum janus_videoroom_videocodec {
-	JANUS_VIDEOROOM_NOVIDEO,	/* Publishers won't do any audio */
-	JANUS_VIDEOROOM_VP8,		/* Publishers will have to/may use VP8 */
-	JANUS_VIDEOROOM_VP9,		/* Publishers will have to/may use VP9 */
-	JANUS_VIDEOROOM_H264		/* Publishers will have to/may use H264 */
-} janus_videoroom_videocodec;
-static const char *janus_videoroom_videocodec_name(janus_videoroom_videocodec vcodec) {
-	switch(vcodec) {
-		case JANUS_VIDEOROOM_NOVIDEO:
-			return "none";
-		case JANUS_VIDEOROOM_VP8:
-			return "vp8";
-		case JANUS_VIDEOROOM_VP9:
-			return "vp9";
-		case JANUS_VIDEOROOM_H264:
-			return "h264";
-		default:
-			/* Shouldn't happen */
-			return "vp8";
-	}
-}
-static janus_videoroom_videocodec janus_videoroom_videocodec_from_name(const char *name) {
-	if(name == NULL)
-		return JANUS_VIDEOROOM_NOVIDEO;
-	else if(!strcasecmp(name, "vp8"))
-		return JANUS_VIDEOROOM_VP8;
-	else if(!strcasecmp(name, "vp9"))
-		return JANUS_VIDEOROOM_VP9;
-	else if(!strcasecmp(name, "h264"))
-		return JANUS_VIDEOROOM_H264;
-	JANUS_LOG(LOG_WARN, "Unsupported video codec '%s'\n", name);
-	return JANUS_VIDEOROOM_NOVIDEO;
-}
-static int janus_videoroom_videocodec_pt(janus_videoroom_videocodec vcodec) {
-	switch(vcodec) {
-		case JANUS_VIDEOROOM_NOVIDEO:
-			return -1;
-		case JANUS_VIDEOROOM_VP8:
-			return VP8_PT;
-		case JANUS_VIDEOROOM_VP9:
-			return VP9_PT;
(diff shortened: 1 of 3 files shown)


mtdxc:
I hope it will be merge back to master and less copy and paste in the code

lminiero:
I'll need some time to review, I'll try and come back to you when I have some time. Thx!

plugins/janus_videoroom.c
```diff
@@ -1431,6 +1300,7 @@ typedef struct janus_videoroom_session {
 	janus_mutex mutex;
 	janus_refcount ref;
 } janus_videoroom_session;
+// janus_plugin_session -> janus_videoroom_session
```
lminiero: I guess this comment here is a leftover? Please remove this. As a side note, per our code style comments should always be in the form of `/* */` and not `//`.

plugins/janus_videoroom.c
```diff
@@ -1671,7 +1541,47 @@ static void janus_videoroom_message_free(janus_videoroom_message *msg) {
 	g_free(msg);
 }
 
+void janus_videoroom_codecstr(janus_videoroom * videoroom, char* audio_codecs, char* video_codecs, int str_len, const char* split)
```
lminiero: Just nitpicking here, please bear with me! Per code style, for pointers the convention is `type *variable`, which is not matched in the signature above. I noticed the same thing in a few other places as well, so please fix those too.

plugins/janus_videoroom.c
```diff
@@ -1671,7 +1541,47 @@ static void janus_videoroom_message_free(janus_videoroom_message *msg) {
...
+{
+	if(audio_codecs){
```
lminiero: Nitpicking again... :smile: 
Please leave a space before the curly brackets: `if(audio_codecs) {` (same for a few other points where I've seen the same thing, e.g., for video_codecs below).

plugins/janus_videoroom.c
```diff
@@ -1671,7 +1541,47 @@ static void janus_videoroom_message_free(janus_videoroom_message *msg) {
...
+		}
+		if (videoroom->vcodec[2] != JANUS_VIDEOCODEC_NONE) {
+			g_strlcat(video_codecs, split, str_len);
+			g_strlcat(video_codecs, janus_videocodec_name(videoroom->vcodec[2]), str_len);
+		}
+	}
+}
 
+void janus_videoroom_reqfir(janus_videoroom_publisher* publisher, const char* reason){
+	/* Send a FIR to the new RTP forward publisher */
```
lminiero: If this is a generic, and shared, way of triggering PLI/FIR messages (as suggested by the `reason` above), then I guess the comment above is broken?

plugins/janus_videoroom.c
```diff
@@ -1706,15 +1616,13 @@ static guint32 janus_videoroom_rtp_forwarder_add_helper(janus_videoroom_publishe
 	if(!is_data && srtp_suite > 0 && srtp_crypto != NULL) {
 		/* First of all, let's check if there's already an RTP forwarder with
 		 * the same SRTP context: make sure SSRC and pt are the same too */
-		char media[10];
-		memset(media, 0, sizeof(media));
+		char media[10] = {0};
```
lminiero: Does this set all the bytes in the char array to 0?

plugins/janus_videoroom.c
```diff
@@ -1772,6 +1680,7 @@ static guint32 janus_videoroom_rtp_forwarder_add_helper(janus_videoroom_publishe
 	forward->ssrc = ssrc;
 	forward->substream = substream;
 	forward->is_data = is_data;
+	// host to ip
```
lminiero: Please remove this comment as well (which again I assume is a leftover while studying the code?)

plugins/janus_videoroom.c
```diff
@@ -2830,26 +2697,7 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session
...
-		g_snprintf(video_codecs, sizeof(video_codecs), "%s", janus_videoroom_videocodec_name(videoroom->vcodec[0]));
-		if(videoroom->vcodec[1] != JANUS_VIDEOROOM_NOVIDEO) {
-			g_strlcat(video_codecs, "|", sizeof(video_codecs));
-			g_strlcat(video_codecs, janus_videoroom_videocodec_name(videoroom->vcodec[1]), sizeof(video_codecs));
-		}
-		if(videoroom->vcodec[2] != JANUS_VIDEOROOM_NOVIDEO) {
-			g_strlcat(video_codecs, "|", sizeof(video_codecs));
-			g_strlcat(video_codecs, janus_videoroom_videocodec_name(videoroom->vcodec[2]), sizeof(video_codecs));
-		}
+		janus_videoroom_codecstr(videoroom, audio_codecs, video_codecs, 100, "|");
```
lminiero: I'd put `sizeof(video_codecs)` or `sizeof(audio_codecs)` there instead of 100, just in case in the future we change the size of those arrays and forget to update it in this call too.

plugins/janus_videoroom.c
```diff
@@ -2883,29 +2731,10 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session
...
-			g_snprintf(audio_codecs, sizeof(audio_codecs), "%s", janus_videoroom_audiocodec_name(videoroom->acodec[0]));
-			if(videoroom->acodec[1] != JANUS_VIDEOROOM_NOAUDIO) {
-				g_strlcat(audio_codecs, ",", sizeof(audio_codecs));
-				g_strlcat(audio_codecs, janus_videoroom_audiocodec_name(videoroom->acodec[1]), sizeof(audio_codecs));
-			}
-			if(videoroom->acodec[2] != JANUS_VIDEOROOM_NOAUDIO) {
-				g_strlcat(audio_codecs, ",", sizeof(audio_codecs));
-				g_strlcat(audio_codecs, janus_videoroom_audiocodec_name(videoroom->acodec[2]), sizeof(audio_codecs));
-			}
+			janus_videoroom_codecstr(videoroom, audio_codecs, video_codecs, 100, ",");
```
lminiero: Is this needed, since we just did the same thing a few lines above?

plugins/janus_videoroom.c
```diff
@@ -3056,28 +2886,9 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session
...
-			g_snprintf(video_codecs, sizeof(video_codecs), "%s", janus_videoroom_videocodec_name(videoroom->vcodec[0]));
-			if(videoroom->vcodec[1] != JANUS_VIDEOROOM_NOVIDEO) {
-				g_strlcat(video_codecs, ",", sizeof(video_codecs));
-				g_strlcat(video_codecs, janus_videoroom_videocodec_name(videoroom->vcodec[1]), sizeof(video_codecs));
-			}
-			if(videoroom->vcodec[2] != JANUS_VIDEOROOM_NOVIDEO) {
-				g_strlcat(video_codecs, ",", sizeof(video_codecs));
-				g_strlcat(video_codecs, janus_videoroom_videocodec_name(videoroom->vcodec[2]), sizeof(video_codecs));
-			}
+			janus_videoroom_codecstr(videoroom, audio_codecs, video_codecs, 100, ",");
```
lminiero: See above for the fixed size.

plugins/janus_videoroom.c
```diff
@@ -3911,28 +3693,15 @@ void janus_videoroom_setup_media(janus_plugin_session *handle) {
...
-			while (!g_atomic_int_get(&videoroom->destroyed) && g_hash_table_iter_next(&iter, NULL, &value)) {
-				janus_videoroom_publisher *p = value;
-				if(p == participant) {
-					continue;	/* Skip the new publisher itself */
-				}
-				JANUS_LOG(LOG_VERB, "Notifying participant %"SCNu64" (%s)\n", p->user_id, p->display ? p->display : "??");
-				int ret = gateway->push_event(p->session->handle, &janus_videoroom_plugin, NULL, pub, NULL);
-				JANUS_LOG(LOG_VERB, "  >> %d (%s)\n", ret, janus_get_api_error(ret));
-			}
+			janus_videoroom_notify_participants(participant, pub);
```
lminiero: Why remove the mutex lock around this call? The method doesn't lock it itself.

plugins/janus_videoroom.c
```diff
@@ -4188,27 +3948,19 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
...
-					participant->fir_latest = now;
-					char rtcpbuf[24];
-					janus_rtcp_fir((char *)&rtcpbuf, 20, &participant->fir_seq);
-					JANUS_LOG(LOG_VERB, "Sending FIR to %"SCNu64" (%s)\n", participant->user_id, participant->display ? participant->display : "??");
-					gateway->relay_rtcp(handle, video, rtcpbuf, 20);
-					/* Send a PLI too, just in case... */
-					janus_rtcp_pli((char *)&rtcpbuf, 12);
-					JANUS_LOG(LOG_VERB, "Sending PLI to %"SCNu64" (%s)\n", participant->user_id, participant->display ? participant->display : "??");
-					gateway->relay_rtcp(handle, video, rtcpbuf, 12);
+					janus_videoroom_reqfir(participant, "");
```
lminiero: Why an empty string, instead of something like "Regular keyframe request"?

rtp.c
```diff
@@ -601,3 +601,108 @@ const char *janus_srtp_error_str(int error) {
...
+			return VP8_PT;
+		case JANUS_VIDEOCODEC_VP9:
+			return VP9_PT;
+		case JANUS_VIDEOCODEC_H264:
+			return H264_PT;
+		default:
+			/* Shouldn't happen */
+			return VP8_PT;
+	}
+}
```
lminiero: Missing newline here.

rtp.h
```diff
@@ -195,4 +195,38 @@ int janus_rtp_skew_compensate_audio(janus_rtp_header *header, janus_rtp_switchin
...
+/* Payload types we'll offer internally */
+#define OPUS_PT		111
+#define ISAC32_PT	104
+#define ISAC16_PT	103
+#define PCMU_PT		0
+#define PCMA_PT		8
+#define G722_PT		9
+#define VP8_PT		96
+#define VP9_PT		101
+#define H264_PT		107
```
lminiero: These payload types here should be statically defined in `rtp.c` and not here, I think, especially considering they're not used anywhere else in the code (nor should they I think). If you see any reason for having them defined in the header, instead, then they should be prefixed (e.g., `JANUS_OPUS_PT` or `JANUS_AUDIO_OPUS_PT`).

plugins/janus_videoroom.c
mtdxc: yes it worked in c++

mtdxc:
sorry about my bad codestyle, I have changed the code and commit again

plugins/janus_videoroom.c
RSATom: C99 Standard 6.7.8.21
> If there are fewer initializers in a brace-enclosed list than there are elements or members of an aggregate, or fewer characters in a string literal used to initialize an array of known size than there are elements in the array, the remainder of the aggregate shall be initialized implicitly the same as objects that have static storage duration.

So yes, it's valid.

lminiero: Thanks for clarifying!

plugins/janus_videoroom.c
```diff
@@ -1671,7 +1540,47 @@ static void janus_videoroom_message_free(janus_videoroom_message *msg) {
...
+			g_strlcat(video_codecs, janus_videocodec_name(videoroom->vcodec[1]), str_len);
+		}
+		if (videoroom->vcodec[2] != JANUS_VIDEOCODEC_NONE) {
+			g_strlcat(video_codecs, split, str_len);
+			g_strlcat(video_codecs, janus_videocodec_name(videoroom->vcodec[2]), str_len);
+		}
+	}
+}
 
+static void janus_videoroom_reqfir(janus_videoroom_publisher *publisher, const char *reason){
```
lminiero: Can you add a space towards the end? `const char *reason) {`

plugins/janus_videoroom.c
```diff
@@ -1671,7 +1540,47 @@ static void janus_videoroom_message_free(janus_videoroom_message *msg) {
...
+	JANUS_LOG(LOG_VERB, "%s sending FIR to %"SCNu64" (%s)\n", reason, publisher->user_id, publisher->display ? publisher->display : "??");
+	gateway->relay_rtcp(publisher->session->handle, 1, buf, 20);
+	/* Send a PLI too, just in case... */
+	janus_rtcp_pli((char *)&buf, 12);
+	JANUS_LOG(LOG_VERB, "%s sending PLI to %"SCNu64" (%s)\n", reason, publisher->user_id, publisher->display ? publisher->display : "??");
+	gateway->relay_rtcp(publisher->session->handle, 1, buf, 12);
+	/* Update the time of when we last sent a keyframe request */
+	publisher->fir_latest = janus_get_monotonic_time();
+}
 /* Error codes */
```
lminiero: Please add a newline after the method and before the Error codes comment.

lminiero:
Looks good! Thanks again for your contribution, merging :+1:

PR #1231, opened 2018-05-12
Only use a single thread per-PeerConnection instead of two

lminiero:
As you may or may not know, when a PeerConnection is created in Janus internally we use two different threads: one is devoted to the libnice GMainContext/GMainLoop (the Glib event loop that notifies us about changes in ICE state and incoming packets), while another thread is spawned for outgoing traffic, the so-called "send thread". This second thread is also responsible for sending RTCP feedback on a regular basis, sending stats to event handlers, and a few other things.

These past few days I've studied how to get rid of the additional thread, and have a single one do all the work instead. More specifically, I decided to try and see if I could take advantage of the existing Glib event loop libnice was already using, which is what I ended up doing in this pull request. I found a useful tutorial on [how to create new custom GSources](https://developer.gnome.org/gnome-devel-demos/stable/custom-gsource.c.html.en), which was quite helpful as it uses GAsyncQueue, which we were already using for the "send thread". This allowed me to simulate a data source via this asynchronous queue, so that the event loop would pick packets when available from there, and send them from the loop thread instead. I also created more "traditional" sources for the recurring RTCP and stats as well.

I had this basically ready for a few days already, but it was not working properly, with weird delays in the outgoing traffic part. I exchanged some ideas with the libnice developers, and [Olivier was really helpful](https://lists.freedesktop.org/archives/nice/2018-May/001428.html), as he explained what I probably missing in my code, which was spot-on.

I have tested this briefly and this seems to be working exactly as intended. From a few measurements it *seems* to use slightly less CPU than before, which if confirmed would be great, but I haven't done any true investigation yet to be fair: at any rate, managing to get rid of a thread per connection is still a huge achievement and improvement anyway, as it will save a lot of resources; besides, it already simplified parts of the code, allowing us to get rid of some locks here and there. I'm sure that before merging I'll be able to find a few other parts that might be improved further.

Since this is basically done, I'd ask you to test this and provide feedback ASAP. In case you find anything weird or a regression, please let me know now and not after I merge (which if I don't get any objection will happen soon).

EDIT: one more reason this is interesting is that, in the future, we might even think of spawning a limited number of threads to host multiple libnice agents, all sharing the same loop. More on that later.


RSATom:
> I found a useful tutorial on how to create new custom GSources, which was quite helpful as it uses GAsyncQueue, which we were already using for the "send thread".

I've tried to follow mentioned tutorial and found it's incomplete, or maybe even wrong. It doesn't contain any way to wake up waiting thread without any timeout if some other thread push item to GAsyncQueue. Finally I've found how to solve it - use eventfd concept. Maybe you are already using it, but I didn't find it in sources (but of course maybe I've missed something).

You could look at my implementation at:
https://github.com/RSATom/janus-gstreamer-plugin/blob/master/plugins/gstreamer/QueueSource.h
https://github.com/RSATom/janus-gstreamer-plugin/blob/master/plugins/gstreamer/QueueSource.cpp
It's C++, but I think it could be ported to C with ease I think.

One possible issue - I didn't test this code on high load yet...

fancycode:
Could this or something similar also be used to get rid of the SCTP thread? This thread is also sending data as described in https://github.com/meetecho/janus-gateway/pull/1226#issuecomment-386616741

lminiero:
I'll have to check as it's been a while since the last time I worked on that code.

lminiero:
The sooner we get feedback on this effort to merge it, the sooner I can look into that too :)

ruddell:
Tested with ~250 viewers with the streaming plugin and didn't notice any issues.

lminiero:
> Tested with ~250 viewers with the streaming plugin and didn't notice any issues.

Excellent, thanks for the feedback! We'll make a couple of other tests ourselves and hopefully merge very soon :+1:

RSATom:
@ruddell, just curious, what hardware did you use for that test, and what cpu load did you get?

RSATom:
@lminiero, am I correct, with 250 viewers Janus will use at least 250 threads? i.e. at least one thread for one viewer?

lminiero:
@RSATom "at least" makes it look like a bad thing: we currently (in master) use 250*2=500 threads for 250 viewers :wink:

lminiero:
And that was the whole purpose of this effort: cutting down on the resources used per handle.

lminiero:
Well, one thread per participant is not that much, actually: I know some other popular tool uses many more, while others are multiprocess and purely event based. It's a matter of design.

This is an important first step in making everything more efficient in our codebase (well, we already greatly cut in the number of threads in other contexts, but this is the first take on media). Since this PR moves the management of all media for a user to an event loop, it means that in the future we might "aggregate" more users on the same event loop without having to change anything much in the code.

Whether or not this is something we'll want to do it depends: an RTSP server doesn't have to worry about encryption, for instance, and SRTP does take its toll; put too many in the same thread and you risk not being able to keep up with the delivery. But I like the idea that we can do it, after we merge this PR: then it will be a matter of making some tests to evaluate how well the different approaches perform.

lminiero:
I did some tests locally and I didn't notice many differences CPU wise. Anyway, even if the CPU is a bit higher, using half of the threads is still a huge advantage, and so are the other options it opens for us. We can improve this along the road with more testing: in the meanwhile, I'm going to bump the version number (I just tagged v0.4.0) and merge this.

Thanks you all for the feedback!

PR #1232, opened 2018-05-13
choxx/janus gateway

choxx:
Added background threads to reconnect to RTSP cameras & STUN server binding if they aren't available at startup using `pthread`. The thread will exit once it gets successful connection reply.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -22,6 +22,7 @@
 #include <fcntl.h>
 #include <stun/usages/bind.h>
 #include <nice/debug.h>
+#include <pthread.h>
 
 #include "janus.h"
 #include "debug.h"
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks for your contribution, but I'm not really sure I understand this pull request. We already have reconnections attempts for RTSP, optionally even in the case when the RTSP server is not immediately reachable: https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_streaming.c#L110
Besides, this patch also includes retries for STUN/TURN, which as already discussed in the past is something we don't want in the code.

lminiero:
Closing, thanks for the contribution anyway!

PR #1236, opened 2018-05-15
expose janus.js extension config via dependency API

s-ol:


lminiero:
Sorry for this late reply, I've been away for a few days... I'm afraid a change we did today may have introduced a conflict in your patch, could you check it?

lminiero:
Thanks, I'll test it later today and merge in case I see no issue :+1:

lminiero:
Looks good, merging!

PR #1240, opened 2018-05-18
Keep ICE component alive while SCTP thread is running.

fancycode:
This should fix the following error (line numbers might be a bit off due to debugging code added):
```
==1543==ERROR: AddressSanitizer: heap-use-after-free on address 0x61700182428c at pc 0x000000521f2e bp 0x7fa259c4fcf0 sp 0x7fa259c4fce8
READ of size 4 at 0x61700182428c thread T4044 (sctp 6012426324)
    #0 0x521f2d in janus_dtls_fd_bridge /path/to/janus-gateway/dtls.c:1043:4
    #1 0x531a8b in janus_dtls_send_sctp_data /path/to/janus-gateway/dtls.c:1082:3
    #2 0x62af1a in janus_sctp_thread /path/to/janus-gateway/sctp.c:1310:4
    #3 0x7fa396faff04  (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x6df04)
    #4 0x7fa395ed5183 in start_thread /build/eglibc-ripdx6/eglibc-2.19/nptl/pthread_create.c:312
    #5 0x7fa3957e303c in clone /build/eglibc-ripdx6/eglibc-2.19/misc/../sysdeps/unix/sysv/linux/x86_64/clone.S:111

0x61700182428c is located 12 bytes inside of 736-byte region [0x617001824280,0x617001824560)
freed by thread T3886 (iceloop 6012426) here:
    #0 0x4cca82 in __interceptor_free /b/build/slave/linux_upload_clang/build/src/third_party/llvm/compiler-rt/lib/asan/asan_malloc_linux.cc:78:3
    #1 0x557c26 in janus_ice_component_free /path/to/janus-gateway/ice.c:1321:2
    #2 0x5558b7 in janus_ice_component_destroy /path/to/janus-gateway/ice.c:1228:2
    #3 0x554e8b in janus_ice_stream_destroy /path/to/janus-gateway/ice.c:1147:3
    #4 0x553fa2 in janus_ice_webrtc_free /path/to/janus-gateway/ice.c:1099:3
    #5 0x55994d in janus_ice_thread /path/to/janus-gateway/ice.c:2589:2
    #6 0x7fa396faff04  (/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x6df04)
```

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1225,6 +1225,7 @@ void janus_ice_component_destroy(janus_ice_component *component) {
 		janus_refcount_decrease(&stream->ref);
 		component->stream = NULL;
 	}
+	janus_dtls_srtp_destroy(component->dtls);
 	janus_refcount_decrease(&component->ref);
 }
 
--- a/sctp.c
+++ b/sctp.c
@@ -261,7 +261,7 @@ janus_sctp_association *janus_sctp_association_create(void *dtls, uint64_t handl
 	g_snprintf(tname, sizeof(tname), "sctp %"SCNu64, sctp->handle_id);
 	void *component = ((janus_dtls_srtp *)dtls)->component;
 	if (component) {
-		janus_refcount_increase(&((janus_ice_component*)component)->ref);
+		janus_refcount_increase(&((janus_ice_component *)component)->ref);
 	}
 	janus_refcount_increase(&sctp->ref);
 	sctp->thread = g_thread_try_new(tname, &janus_sctp_thread, sctp, &error);
@@ -270,7 +270,7 @@ janus_sctp_association *janus_sctp_association_create(void *dtls, uint64_t handl
 		janus_mutex_unlock(&sctp->mutex);
 		JANUS_LOG(LOG_ERR, "[%"SCNu64"] Got error %d (%s) trying to launch the SCTP thread...\n", handle_id, error->code, error->message ? error->message : "??");
 		if (component) {
-			janus_refcount_decrease(&((janus_ice_component*)component)->ref);
+			janus_refcount_decrease(&((janus_ice_component *)component)->ref);
 		}
 		janus_refcount_decrease(&sctp->ref);	/* This is for the failed thread */
 		janus_refcount_decrease(&sctp->ref);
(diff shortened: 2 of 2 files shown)


ice.c
```diff
@@ -1225,6 +1225,7 @@ void janus_ice_component_destroy(janus_ice_component *component) {
 		janus_refcount_decrease(&stream->ref);
 		component->stream = NULL;
 	}
+	janus_dtls_srtp_destroy(component->dtls);
```
lminiero: Won't this cause problems? We normally do this in `janus_ice_component_free`, where we also set the pointer to NULL, though. Since you're not doing it here, the references may have gone to 0 when we're about to free the component, and the further call to `janus_dtls_srtp_destroy` might crash it.

sctp.c
```diff
@@ -259,12 +259,19 @@ janus_sctp_association *janus_sctp_association_create(void *dtls, uint64_t handl
 	GError *error = NULL;
 	char tname[16];
 	g_snprintf(tname, sizeof(tname), "sctp %"SCNu64, sctp->handle_id);
+	void *component = ((janus_dtls_srtp *)dtls)->component;
+	if (component) {
+		janus_refcount_increase(&((janus_ice_component*)component)->ref);
```
lminiero: Just a nit for code style: please separate the type and the asterisk with a space (in that other line too).

sctp.c
```diff
@@ -1338,6 +1347,9 @@ void *janus_sctp_thread(void *data) {
 	}
 	JANUS_LOG(LOG_VERB, "[%"SCNu64"] Leaving SCTP association thread\n", sctp->handle_id);
 	janus_refcount_decrease(&sctp->ref);
+	if (component) {
```
lminiero: I guess that the reference count protects us here, but shouldn't you check again if `dtls->component` is NULL here, since there may have been a long while in between?

ice.c
fancycode: `janus_ice_component_free` is only called when the the refcount of the component reaches zero - which now only happens when the sctp thread ends. By calling `janus_dtls_srtp_destroy` from `janus_ice_component_destroy`, the refcount of `dtls->sctp` is decremented and will eventually reach zero - causing the thread to end and to release the component.

Also `janus_dtls_srtp_destroy` can be called multiple times, it atomically checks and sets `dtls->destroyed`.

sctp.c
fancycode: The component is refcounted, so it should not get cleared. If it would and we would check here again, it would leak a reference.

ice.c
lminiero: :+1:

sctp.c
fancycode: Done

sctp.c
lminiero: Good point :+1:

lminiero:
Thanks for the quick clarifications, looks good to me then. Merging!

PR #1241, opened 2018-05-18
Only build package to relay if queue exists.

fancycode:
Fixes memory leaks for cases where the handle exists but doesn't have a packet queue.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -4005,7 +4005,7 @@ void *janus_ice_send_thread(void *data) {
 }
 
 void janus_ice_relay_rtp(janus_ice_handle *handle, int video, char *buf, int len) {
-	if(!handle || buf == NULL || len < 1)
+	if(!handle || handle->queued_packets == NULL || buf == NULL || len < 1)
 		return;
 	if((!video && !janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_HAS_AUDIO))
 			|| (video && !janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_HAS_VIDEO)))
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for the patch! Just to nitpick, with this change though you're checking if the queue esists a few lines before actually using it: if the queue is removed while you allocate the packet, you directly try and put in queue to something that doesn't exist. Wouldn't it be easier to just deallocate the just created packet in an `else` to the existing `if(handle->queued_packets != NULL)` check?

fancycode:
I assumed the queue could not be deleted while the package is built - otherwise there is some locking missing as it could deleted right now between the `if(handle->queued_packets)` and the `g_async_queue_push(...)`.

lminiero:
The relay methods don't lock anything for performance reason, and the GAsyncQueue methods are mutex-protected only while you call them.

fancycode:
> The relay methods don't lock anything for performance reason, and the GAsyncQueue methods are mutex-protected only while you call them.

Yeah, sure. However my PR doesn't change the semantics of the function. If `queued_package` could get reset while the packet is being build before my changes, there already was a race condition (and still is).

The solution is to either introduce locks (which obviously is bad here), or otherwise make sure the functions can't get called for handles that are about to be destroyed. But this should be done in a separate PR imho.

lminiero:
Yep, I agree! And I realize the same could happen with the code we have in place now. Still, I fill it's still less likely to cause issues when an action is done immediately after the if, rather than if some other ops are in the middle (if not for timing reasons). If you feel it doesn't make that much of a difference at all, then no objection from me to merge.

fancycode:
It's a race condition, so it could happen any time and gets more likely under load. I re-added the check before calling `g_async_queue_push` and added a `TODO` marker about the race.

lminiero:
Thx! If I can add a last nitpicky note, please use `/* */` instead of `//` for comments :blush:

fancycode:
Oops :smile: Done and squashed.

lminiero:
Sorry, I know I'm annoying! :smile: 
Looks good to me then, merging :+1:

fancycode:
> Sorry, I know I'm annoying! :smile:

No worries, I'm all for having a consistent style! Maybe would be a good idea to enforce this through automated CI checks?

lminiero:
Nah, I'm not that picky... besides, wouldn't it be quite hard to enforce? We already have an `.editorconfig` file in the repo to help with some other, more basic, properties.

PR #1242, opened 2018-05-18
Free any queued packets before the queue is destroyed.

fancycode:
Also moved the common code to a separate function.


lminiero:
Looks good, merging, thanks!

PR #1243, opened 2018-05-18
Join SCTP thread before releasing resources.

fancycode:


lminiero:
If that's ok for you, I'll have a deeper look at this on Monday (it's almost end-of-day here). The diff is small and simple, but I want to be sure about the potential implications, e.g., deadlocks that may or may not occur, and I don't have time to look into this right now.

In the meanwhile, thanks again for the great help! 4 PRs in the same day are, if not a record, damn close :smile:

lminiero:
Looked at the code and I'm not sure what benefit the patch would provide: `janus_sctp_association_free` is only going to be called when there are no more references to the SCTP stack, and one is held by the thread itself. This means that, when `janus_sctp_association_free` is called, the thread loop is already over, and on its way out, so there's no risk of it going to try and use the stack after it's been destroyed. Am I missing something?

lminiero:
Maybe you meant to join the thread in `janus_sctp_association_destroy` instead?

fancycode:
You're right, `janus_sctp_association_destroy` should be changed instead - I updated the PR.

lminiero:
Thx! I'll make some tests later today and merge if it's fine :+1:

lminiero:
Looks good, merging!

mquander:
I wonder if this bug might have been related to the [usrsctp address reuse issue](https://groups.google.com/forum/#!searchin/meetecho-janus/sctp%7Csort:date/meetecho-janus/8MDjukdI4eA/iGv1mQKxAwAJ) I experienced a few weeks ago. It's not obvious to me at a glance why a race between setup and this part of destruction would cause that issue, though. usrsctp is [supposed to be](https://github.com/sctplab/usrsctp/issues/37) internally threadsafe.

PR #1244, opened 2018-05-18
Use JSEP trickle parameter when handling answers

ruddell:
As [explained in the docs (below)](https://github.com/meetecho/janus-gateway/blob/master/mainpage.dox#L1325-L1331), adding `trickle: false` to the JSEP should disable trickling for that handle.  This was working up until #1113 where the code to set the flags for answers [was removed](https://github.com/meetecho/janus-gateway/pull/1113/files#diff-e26c229e0800bdaf6e4e995fba5d3c4dL1121) (it still works [for offers](https://github.com/meetecho/janus-gateway/pull/1113/files#diff-724b612068e82ba0ad6ed24da3a22e47R2746)).

> Please notice that, if for any reason you don't want to use the trickling of ICE candidates from your application (which means you'll include them all in the SDP OFFER or ANSWER, which is usually not recommended), you'll have to add an additional <code>"trickle" : false</code> attribute to the "jsep" object, to explicitly tell Janus you won't send any trickle candidate (by default Janus will always assume support for trickle).

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -1204,7 +1204,11 @@ int janus_process_incoming_request(janus_request *request) {
 				}
 				if(!offer) {
 					/* Set remote candidates now (we received an answer) */
-					janus_flags_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_TRICKLE);
+					if(do_trickle) {
+						janus_flags_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_TRICKLE);	
+					} else {	
+						janus_flags_clear(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_TRICKLE);	
+					}
 					janus_request_ice_handle_answer(handle, audio, video, data, jsep_sdp);
 				} else {
 					/* Check if transport wide CC is supported */


lminiero:
IIRC we changed it for a reason, but I may be wrong... I'll have to check if restoring it as your patch does will cause any regression. At any rate, the `trickle:false` does not mean that Janus should not trickle, but is just a notification that the browser won't; Janus is free to do what it wants, and that's dictated by how you configured it (that is, whether `full_trickle=yes` or not). I think it's probably redudant anyway, because when trickle is not used by the client, typically you have a `end-of-candidates` attribute in the SDP to notify there will be no others.

ruddell:
Janus doesn't seem to respect the `end-of-candidates` in the offer/answer, but that would probably be the preferred solution.  I guess that's what [this comment is referring to](https://github.com/meetecho/janus-gateway/blob/45e8fff2239b7e41dfb1e4c0708a667f3de812ec/sdp.c#L565-L569).  

This change would only affect developers that manually add the `trickle: false` to the JSEP object, otherwise it [defaults to the current behavior](https://github.com/ruddell/janus-gateway/blob/8053c61becc83b9b446bb6d5c307ae0bd728e9b6/janus.c#L1093-L1095).

I'm not too invested in the outcome of this PR, I updated my code to use `full_trickle` and don't have issues anymore.  I can update the docs to point out the `trickle: false` option only works for offers if you prefer

lminiero:
> Janus doesn't seem to respect the end-of-candidates in the offer/answer, but that would probably be the preferred solution. I guess that's what this comment is referring to.

Yep, even though I wouldn't handle the `end-of-candidates` there. Both Chrome and Firefox put it at the end of each m-line, while per the (not complete yet) specification you should only have it at the end of the SDP (which is where we put it ourselves, for instance). Edge for instance would ignore other candidates after the first `end-of-candidates` it sees. I'd probably put it somewhere after the whole SDP has been handled.

> This change would only affect developers that manually add the trickle: false to the JSEP object, otherwise it defaults to the current behavior.
> 
> I'm not too invested in the outcome of this PR, I updated my code to use full_trickle and don't have issues anymore. I can update the docs to point out the trickle: false option only works for offers if you prefer

No need, I believe your patch does the right thing. I'll make a couple of tests just to be sure and merge in case :+1: 
Thanks again for spotting this inconsistency out!

lminiero:
Looks good, merging!

PR #1245, opened 2018-05-22
Use atomic API to get/set flags.

fancycode:
This fixes races where flags are get/set from different threads.


lminiero:
That's a good idea! But I'm not comfortable with the change from `uint32_t` to `guint`. That's an alias for `unsigned int`, which AFAIK is not guaranteed to be 32 bit.

fancycode:
In practice `guint` should be 32 bit on all systems, however I changed it to use `gsize` and the `g_atomic_pointer_*` API, so it's (at least) 32 bit now.

lminiero:
I think `guint` is only 32-bit on 64-bit machines. There probably aren't many 32-bit machines/distros around anymore, but you can install Janus on those. As to gsize, I thought the `g_atomic_pointer_*` methods always expected a pointer, but the [documentation](https://developer.gnome.org/glib/stable/glib-Atomic-Operations.html) seems to suggest it works on anything the same size as a pointer, which includes gsize. This does require twice the space now in most cases (gpointer is 64-bit, while uint32_t is 32) but I don't see any way out of it unfortunately...

fancycode:
Well, you could always switch to the GCC atomic builtins (if available in the current compiler) which can use `uint32_t`: https://gcc.gnu.org/onlinedocs/gcc-4.8.2/gcc/_005f_005fatomic-Builtins.html

Unfortunately those are not exposed through glib.

lminiero:
Yeah, sticking to glib makes things easier, especially considering it avoids potential compatibility issues (older compilers, Mac OS, other stuff). I guess that for now we can use the pointer/gsize stuff, and then maybe revisit this later on: after all, this in theory is transparent to applications, which are supposed to work on `janus_flags` directly.

I'll make a couple of tests to check everything is still fine and in case merge. Thanks again for the contributions!

PR #1248, opened 2018-05-23
Fix additional size for packets that will be protected by libsrtp.

fancycode:
Not enough additional data was reserved in some cases, causing read-buffer-overflow when sending out the encrypted packets.

`srtp_protect` writes up to `SRTP_MAX_TRAILER_LEN` additional bytes:
https://github.com/cisco/libsrtp/blob/master/include/srtp.h#L398

`srtp_protect_rtcp` writes up to `SRTP_MAX_TRAILER_LEN+4` additional bytes:
https://github.com/cisco/libsrtp/blob/master/include/srtp.h#L1304

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -2516,7 +2516,7 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 							retransmits_cnt++;
 							/* Enqueue it */
 							janus_ice_queued_packet *pkt = g_malloc(sizeof(janus_ice_queued_packet));
-							pkt->data = g_malloc(p->length+SRTP_MAX_TRAILER_LEN);
+							pkt->data = g_malloc(p->length+SRTP_MAX_TAG_LEN);
 							memcpy(pkt->data, p->data, p->length);
 							pkt->length = p->length;
 							pkt->type = video ? JANUS_ICE_PACKET_VIDEO : JANUS_ICE_PACKET_AUDIO;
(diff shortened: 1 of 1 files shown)


lminiero:
This seems excessive. `SRTP_MAX_TRAILER_LEN` is defined as:

    #define SRTP_MAX_TRAILER_LEN (SRTP_MAX_TAG_LEN + SRTP_MAX_MKI_LEN)

where

    #define SRTP_MAX_TAG_LEN 16
    #define SRTP_MAX_MKI_LEN 128

but we don't use MKI.

fancycode:
`SRTP_MAX_TRAILER_LEN` would be the future-proof version but changed back to `SRTP_MAX_TAG_LEN`. Still needs the additional 4 bytes for RTCP though.

ice.c
```diff
@@ -2516,7 +2516,7 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 							retransmits_cnt++;
 							/* Enqueue it */
 							janus_ice_queued_packet *pkt = g_malloc(sizeof(janus_ice_queued_packet));
-							pkt->data = g_malloc(p->length);
+							pkt->data = g_malloc(p->length+SRTP_MAX_TAG_LEN);
```
lminiero: This actually should only happen if we're using RFC4588, as in that case we encrypt again: if not, we're sending an already encrypted packet, and so `srtp_protect` is not done.

lminiero: Thinking about it, it's really not that important... I can add that later on, and the additional checks you added are important enough to merge them ASAP. Thanks!

PR #1249, opened 2018-05-23
Fix locking issues in videoroom plugin.

fancycode:
Updates for changes in 6e9eae3b10cc620dab0ce14d6adcb3a3468e6505 which now always locks the room before processing the request.


lminiero:
Argh, you're right, we have a ugly deadlock now... merging, thanks for spotting this!

PR #1250, opened 2018-05-23
Remove SCTP thread.

fancycode:
Also change SCTP socket to be non-blocking, thus removing the need for a separate thread to run `usrsctp_connect`.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -116,12 +116,6 @@ gchar *janus_dtls_get_local_fingerprint(void) {
 }
 
 
-#ifdef HAVE_SCTP
-/* Helper thread to create a SCTP association that will use this DTLS stack */
-void *janus_dtls_sctp_setup_thread(void *data);
-#endif
-
-
 #if JANUS_USE_OPENSSL_PRE_1_1_API
 /*
  * DTLS locking stuff to make OpenSSL thread safe (not needed for 1.1.0)
@@ -598,26 +592,12 @@ int janus_dtls_srtp_create_sctp(janus_dtls_srtp *dtls) {
 		return -4;
 	if(janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_ALERT))
 		return -5;
-	dtls->sctp = janus_sctp_association_create(dtls, handle->handle_id, 5000);
+	dtls->sctp = janus_sctp_association_create(dtls, handle, 5000);
 	if(dtls->sctp == NULL) {
 		JANUS_LOG(LOG_ERR, "[%"SCNu64"] Error creating SCTP association...\n", handle->handle_id);
 		return -6;
 	}
-	/* We need to start it in a thread, since it has blocking accept/connect stuff */
-	janus_refcount_increase(&dtls->ref);
-	janus_refcount_increase(&dtls->sctp->ref);
-	GError *error = NULL;
-	char tname[16];
-	g_snprintf(tname, sizeof(tname), "sctpinit %"SCNu64, handle->handle_id);
-	g_thread_try_new(tname, janus_dtls_sctp_setup_thread, dtls, &error);
-	if(error != NULL) {
-		/* Something went wrong... */
-		janus_refcount_decrease(&dtls->ref);
-		janus_refcount_decrease(&dtls->sctp->ref);
-		JANUS_LOG(LOG_ERR, "[%"SCNu64"] Got error %d (%s) trying to launch the DTLS-SCTP thread...\n", handle->handle_id, error->code, error->message ? error->message : "??");
-		return -7;
-	}
-	return 0;
+	return janus_sctp_association_setup(dtls->sctp);
 #else
 	/* Support for datachannels hasn't been built in */
 	return -1;
@@ -937,7 +917,6 @@ void janus_dtls_srtp_destroy(janus_dtls_srtp *dtls) {
 	/* Destroy the SCTP association if this is a DataChannel */
 	if(dtls->sctp != NULL) {
 		janus_sctp_association_destroy(dtls->sctp);
-		janus_refcount_decrease(&dtls->sctp->ref);
 		dtls->sctp = NULL;
 	}
 #endif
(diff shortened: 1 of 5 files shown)


dtls.c
```diff
@@ -598,26 +592,12 @@ int janus_dtls_srtp_create_sctp(janus_dtls_srtp *dtls) {
...
-	g_thread_try_new(tname, janus_dtls_sctp_setup_thread, dtls, &error);
-	if(error != NULL) {
-		/* Something went wrong... */
-		janus_refcount_decrease(&dtls->ref);
-		janus_refcount_decrease(&dtls->sctp->ref);
-		JANUS_LOG(LOG_ERR, "[%"SCNu64"] Got error %d (%s) trying to launch the DTLS-SCTP thread...\n", handle->handle_id, error->code, error->message ? error->message : "??");
-		return -7;
-	}
-	return 0;
+	return janus_sctp_association_setup(dtls->sctp);
```
lminiero: I'm wondering if we should merge `janus_sctp_association_setup` in `janus_sctp_association_create`, since there's no separate thread needed to call that method anymore.

ice.c
```diff
@@ -4016,6 +4031,23 @@ void janus_ice_relay_data(janus_ice_handle *handle, char *buf, int len) {
 }
 #endif
 
+void janus_ice_relay_sctp(janus_ice_handle *handle, char *buffer, int length) {
+#ifdef HAVE_SCTP
+	if(!handle || handle->queued_packets == NULL || buffer == NULL || length < 1)
+		return;
+	/* Queue this packet */
+	janus_ice_queued_packet *pkt = g_malloc(sizeof(janus_ice_queued_packet));
+	pkt->data = g_malloc(length);
```
lminiero: I'm not sure there's a need to allocate anything: IIRC the buffer you get from the SCTP stack is already allocated, meaning we can just "steal" it,

    pkt->data = buffer;

On the other end, this is what we were doing in `janus_sctp_message_create` too, so better keep it as you did it for now and possibly revisit it later on.

sctp.c
```diff
@@ -133,8 +126,8 @@ static void janus_sctp_association_free(const janus_refcount *sctp_ref) {
 	sctp = NULL;
 }
 
-janus_sctp_association *janus_sctp_association_create(void *dtls, uint64_t handle_id, uint16_t udp_port) {
-	if(dtls == NULL || udp_port == 0)
+janus_sctp_association *janus_sctp_association_create(void *dtls, struct janus_ice_handle* ice, uint16_t udp_port) {
```
lminiero: There's no need for the `struct` part, and please invert the whitespace and the asterisk (code style).
As a side note, since we're explicitly referring to the handle type, it's probably worth to stop referring to `dtls` as a pointer to void and use the proper type as well. This would allow us to also add a reference to `dtls` as well here as we do for the handle: we had a reference before, but with the thread gone it's now missing, meaning we should add it here.

sctp.c
```diff
@@ -174,12 +168,18 @@ janus_sctp_association *janus_sctp_association_create(void *dtls, uint64_t handl
...
 	usrsctp_sysctl_set_sctp_ecn_enable(0);
 	if((sock = usrsctp_socket(AF_CONN, SOCK_STREAM, IPPROTO_SCTP, janus_sctp_incoming_data, NULL, 0, (void *)sctp)) == NULL) {
-		JANUS_LOG(LOG_ERR, "[%"SCNu64"] Error creating usrsctp socket... (%d)\n", handle_id, errno);
+		JANUS_LOG(LOG_ERR, "[%"SCNu64"] Error creating usrsctp socket... (%d)\n", sctp->handle_id, errno);
+		janus_refcount_decrease(&sctp->ref);
+		return NULL;
+	}
+	/* Make the socket non-blocking. Connect, close, shutdown etc will not block
+	 * the thread waiting for the socket operation to complete. */
+	if (usrsctp_set_non_blocking(sock, 1) < 0) {
```
lminiero: Let's hope this won't introduce some race conditions! :smile:

sctp.c
```diff
@@ -295,55 +271,60 @@ int janus_sctp_association_setup(janus_sctp_association *sctp) {
 #ifdef HAVE_SCONN_LEN
 	sconn.sconn_len = sizeof(struct sockaddr_conn);
 #endif
-	if(usrsctp_connect(sock, (struct sockaddr *)&sconn, sizeof(struct sockaddr_conn)) < 0) {
+	int res = usrsctp_connect(sctp->sock, (struct sockaddr *)&sconn, sizeof(struct sockaddr_conn));
+	if(res < 0 && errno != EINPROGRESS) {
 		JANUS_LOG(LOG_ERR, "[%"SCNu64"] Error connecting to SCTP server at port %"SCNu16" (%d)\n", sctp->handle_id, sctp->remote_port, errno);
 		return -1;
 	}
 	JANUS_LOG(LOG_VERB, "[%"SCNu64"] Connected to the DataChannel peer\n", sctp->handle_id);
```
lminiero: But if we're ignoring `EINGPROGRESS`, doesn't this mean we're giving the channel for connected, even if it might very well not be?

sctp.h
```diff
@@ -94,6 +96,7 @@ typedef struct janus_sctp_channel {
 typedef struct janus_sctp_association {
 	/*! \brief Opaque pointer to the DTLS instance related to this SCTP association */
 	void *dtls;
+	struct janus_ice_handle* ice;
```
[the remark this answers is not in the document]
lminiero: PS: there's the inverted asterisk/whitespace thing too.

sctp.h
```diff
@@ -74,6 +74,8 @@ void janus_sctp_deinit(void);
 #define DATA_CHANNEL_FLAGS_SEND_RSP 0x00000002
 #define DATA_CHANNEL_FLAGS_SEND_ACK 0x00000004
 
+struct janus_ice_handle;
```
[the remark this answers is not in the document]
fancycode: By using forward declaration here instead of including `ice.h`, all files only including `sctp.h` but not `ice.h` don't need to get recompiled if `ice.h` changes. This helps speeding up compilation and avoiding unnecessary recompilation.

dtls.c
fancycode: Done

ice.c
fancycode: I would do that in a separate PR if necessary.

sctp.c
fancycode: Done (`dtls` already was refcounted by casting it).

sctp.c
fancycode: Right, but that could have been the case before, too (e.g. if the setup thread did not run yet or was not finished). The code also tracks the state of the different channels in the SCTP association.

fancycode:
All comments addressed, thanks for your feedback.

mquander:
@fancycode you are awesome, thank you for doing all this data-channel related work.

lminiero:
Apologies, I meant to have a further look at this last week and then work got in the way... I'll make some tests today and get back to you later :+1:

lminiero:
Made a few tests, introducing network latency as well, and it seems to work just nicely: thanks for the great work, merging!

PR #1251, opened 2018-05-23
bi-directional data channel for streaming-plugin

joukhaha:
Hi all,
I'd like to have access to bidirectional data channel on the streaming plugin. I know this has been discussed on the forums on a few occasions, but I wasn't able to figure out a (dead) simple way to achieve that based on the available pointers.

As a result I tried to modify the plugin to support it directly. Due to the existence of the one-way channel, the modifications needed to janus_streaming.c seem to be quite small. The attached PR simply implements a  few line janus_streaming_incoming_data callback which relays the received data back to the udp client. The return address is picked from the address from which an udp client has sent the previous data-channel message, and therefore at present some data must first be sent by the udp-client in order for the return message to find its target. After this the plugin relays the datachannel to both directions.

Do you think including something like this to the official janus_streaming plugin would be feasible, and do you find any severe issues with the implementation?

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -688,6 +688,9 @@ struct janus_plugin_result *janus_streaming_handle_message(janus_plugin_session 
 void janus_streaming_setup_media(janus_plugin_session *handle);
 void janus_streaming_incoming_rtp(janus_plugin_session *handle, int video, char *buf, int len);
 void janus_streaming_incoming_rtcp(janus_plugin_session *handle, int video, char *buf, int len);
+
+void janus_streaming_incoming_data(janus_plugin_session *handle, char *buf, int len);
+
 void janus_streaming_hangup_media(janus_plugin_session *handle);
 void janus_streaming_destroy_session(janus_plugin_session *handle, int *error);
 json_t *janus_streaming_query_session(janus_plugin_session *handle);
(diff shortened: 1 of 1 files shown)


lminiero:
Apologies for the late feedback, I've been busy on some other activities and missed this.

Sorry, but I don't think we're going to merge this. The Streaming plugin is meant to be a one-direction broadcasting tool, and breaking this pattern for data would raise the question about why not audio/video too, and I don't want to get there. Besides, there's a very loose relationship between a Streaming mountpoint and its source(s): different sources can send stuff at different points in time, and at no time the Streaming plugin binds a local server socket to a specific client (we never do any connect). As such, it's very unclear as to what should happen when a data channel messages comes in: who do you send it to? Even the "last client who sent stuff" you implemented means 1. keeping track of that somehow, and 2. potential leaks of information (see RTP bleed for more info on why).

Thanks for your contribution anyway!

joukhaha:
thanks for the feedback. 

I do agree that bi-directional data might not fully conform with the basic philosophy of the Streaming plugin (albeit at first I was fully convinced that the plugin actually must do bidirectional data as anything else seemed somehow unlogical). At least in my case, however, it would be very useful to have a built-in and webrtc based method to send e.g. commands to the streaming source. Integrating such a possibility to the streaming plugin would seem to be the most straightforward way to do it, but as you say, it might require a bit more planning to do it in a well defined fashion. I don't see an obvious problem in either keeping track of the last active client nor how the bi-directional rtp channel would be any more vulnerable than the single-directional channel, but considering my know-how on the topic this certainly does not mean that there couldn't be any issues.

If you consider that this should not be implemented in the Streaming plugin, which combination of existing plugins would you suggest to use to achieve a bidirectional data channel? Or can you suggest a direction to follow for making this 'properly' within the streaming plugin framework?

lminiero:
> If you consider that this should not be implemented in the Streaming plugin, which combination of existing plugins would you suggest to use to achieve a bidirectional data channel? Or can you suggest a direction to follow for making this 'properly' within the streaming plugin framework?

The VideoRoom plugin allows you to forward incoming datachannel messages from publishers to one or more destinations, via UDP. The only limitation is that, since UDP is being used, messages larger than the MTU will be simply dropped and lost. As an alternative, the TextRoom plugin has a way to forward incoming messages to a remote backend via HTTP instead.

Issue #1229, opened 2018-05-09
Extend dependencies in Janus.init


phsultan:
Hi !

I'm initializing Janus with a specific version of `adapter.js` the following way:
```
Janus.init({
    dependencies: {
      webRTCAdapter: MyWebRTCAdaper
    },
    debug: "all",
    callback: function() {console.log('ok');}
})
```

Doing so will not create the following objects in Janus:
```
    Janus.isArray = usedDependencies.isArray;
    Janus.httpAPICall = usedDependencies.httpAPICall;
    Janus.checkJanusExtension = usedDependencies.checkJanusExtension;
    Janus.newWebSocket = usedDependencies.newWebSocket;
```

And leads to an error when creating a new Janus instance. To reproduce the error, the code below can be used:
```
Janus.init({
	debug: "all",
	dependencies: {
		webRTCAdapter: {} // Empty object used for testing, the same error is thrown with a valid adapter.js instance
	},
	callback: function() {
		console.log('Janus init ok');
		var serverWsURL = 'wss://a-valid-secured-websocket-server-name:8989';
		var janusInstance = new Janus({
			server: serverWsURL,
			success: function() {
				console.log('Instance created');
			},
			error: function(error) {
				Janus.error(error);
			},
			destroyed: function() {
				console.log('Instance destroyed');
			}
		});
	}
});
```

Error thrown: `TypeError: Janus.isArray is not a function`.

As a workaround, I can build the `dependencies` object with all the necessary objects, but instead I modified `Janus.js` in order to accept to extend the dependencies using the `deps` argument in `useDefaultDependencies()`. The diff is simple, although I can make a PR if needed:
```
diff --git a/src/Janus.js b/src/Janus.js
index ccf4635..4c33ccc 100644
--- a/src/Janus.js
+++ b/src/Janus.js
@@ -33,7 +33,7 @@ Janus.useDefaultDependencies = function (deps) {
                newWebSocket: function(server, proto) { return new socketCls(server, proto); },
                isArray: function(arr) { return Array.isArray(arr); },
                checkJanusExtension: function() { return document.querySelector('#janus-extension-installed') !== null; },
-               webRTCAdapter: (deps && deps.adapter) || adapter,
+               webRTCAdapter: (deps && deps.webRTCAdapter) || adapter,
                httpAPICall: function(url, options) {
                        var fetchOptions = {
                                method: options.verb,
@@ -102,7 +102,7 @@ Janus.useOldDependencies = function (deps) {
[...]

PR #1255, opened 2018-05-28
Use defaultDependencies according to documentation.

phsultan:
Documentation specifies how to properly extend Janus dependencies
using the defaultDependencies() function.
This fixes https://github.com/meetecho/janus-gateway/issues/1229
which was actually not an issue, and rather me not reading the
docs !


lminiero:
Thanks for the fix, merging!

PR #1256, opened 2018-05-28
fix memory leak

zgjzzhw:
fix memory leak when events enabled

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -701,9 +701,8 @@ static int janus_request_check_secret(janus_request *request, guint64 session_id
 			}
 		}
 		/* We consider a request authorized if either the proper API secret or a valid token has been provided */
-		if(!secret_authorized && !token_authorized) {
-			return janus_process_error(request, session_id, transaction_text, JANUS_ERROR_UNAUTHORIZED, NULL);
-		}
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for your contribution! Can you please sign our [CLA](https://cla.conf.meetecho.com/janus/) by filling in the form in that page?

Looking at your changes, instead of using `json_decref` it might be easier to just use `json_object_set_new` which, as per the documentation, steals the reference automatically. Anyway, I'll have to check if I used ``json_object_set` on purpose or not at the time.

lminiero:
Looks like we don't have to worry about those references being needed by the caller, since we do unref those objects when an event is pushed but handlers are disabled or unavailable: https://github.com/meetecho/janus-gateway/blob/master/events.c#L73

As such, changing those `json_object_set` to `json_object_set_new` should indeed fix the leaks. As soon as you make this change and sign the CLA I'll merge, thx!

lminiero:
I spotted another leak (`JANUS_EVENT_TYPE_EXTERNAL` events), so I changed those `json_object_set` to `json_object_set_new` myself and pushed the changes: I clarified in the commit that you found the leaks. Thanks for the help!

PR #1264, opened 2018-06-01
New media plugin to drive logic with JavaScript (via Duktape)

lminiero:
This is basically the same thing as #1033, but for JavaScript instead of Lua, so most of the concepts (if not all of them) are pretty much the same: I based it on the same code, which means that the functions implemented in C and the callbacks the scripts must implement are almost exactly the same as well.

## The engine
I used [Duktape](http://duktape.org/), a lightweight and embeddable Javascript engine, to implement the plugin, as suggested by @saghul several months ago (my fault for always coming back to his great ideas so much time later :smile: ). I only started to seriously look into it a couple of days ago, which means that probably many things might be done differently, or better: nevertheless, this is a first experimental attempt, so I welcome any feedback, especially from those that are familiar with Duktape.

I had to struggle with Duktape a little, but I think the end result is already good enough. I found out that many useful features one gives for granted when using JavaScript, e.g. in nodejs, are actually NOT available out of the box in Duktape, for the sake of code portability and to keep the codebase as lightweight as possible: for this attempt, it was the case for instance of missing `console.log` support and the ability to import modules. Luckily enough, Duktape has many extras available, so I added both `duktape_console.c/.h` and `duk_module_duktape.c/.h` to do what I needed. The module stuff is actually more problematic than that (more on this below), but for what I needed in the sample it did the trick.

## Samples
As I did with the Lua module, I created some sample code to test the plugin functionality. This time, though, I only created an `echotest.js` logic script, plus a `janus-sdp.js` module for SDP manipulation. The former I needed to test if everything was working as expected; the latter because I needed some SDP function, and because I wanted to test the module import functionality. Both seem to work as expected. The `echotest.js` sample is interesting also because it shows how I used `console.log` and a custom function for prefixing text, and most importantly how I created a `Duktape.modSearch` function to implement the module import (more on this below).

If you're looking for more, I can already anticipate I'm not going to write a `videoroom.js` script as I did for Lua, though: if you're interested in doing this yourself, maybe using `videoroom.lua` as a reference, I'd love to hear if and how it works. Of course, please let me know about any other attempt at writing a script using this plugin!

## Modules and imports
If you expect your npm modules to work out of the box, and to have freedom to import whatever you want easily, you're probably in for a shock... as anticipated, Duktape was conceived to be lightweight and only have what's strictly needed: anything else is a plus, and must be added. The [documentation](https://github.com/svaarala/duktape-wiki/blob/master/HowtoModules.md) provides useful info on how you can implement such a functionality, and this is what I did (see `janus-sdp.js` in the samples). I had to also implement my own native function for reading a file, as that's missing from the out-of-the-box Duktape too... there may be easier and more obvious ways to do the same thing, but I don't know Duktape well enough to know how.

What you need to do to be able to load modules via our Duktape scripting, is write your own `Duktape.modSearch` function in JavaScript: this is basically the nuts and bolts of what to do when a specific module is loaded. If you look at my `echotest.js` sample, you'll see this:

```
var folder = getModulesFolder();
console.log('Modules folder:', folder);

Duktape.modSearch = function(id) {
	console.log('Loading module:', id);
	// We read the file from the folder the core returned
	var res = readFile(folder + '/' + id + '.js');
	if(typeof res === 'string') {
		console.log('Module loaded');
		return res;
	}
	throw new Error('Module not found: ' + id);
}
```

We could look at `id` to react differently according to the plugin we're trying to load. Since we only load a single module there (`janus-sdp`), we do it the easy way: we know where the file is (`getModulesFolder()` is a native function that returns a folder configured in `janus.plugin.duktape.cfg`), so we read the content of the script into a string (using `readFile()`, another native function I implemented in the plugin code) and we return that. `Duktape.modSearch` then loads it automatically for us.

This definitely works for simple modules like the one I've written, as when I then do this:

```
var sdpUtils = require("janus-sdp");
```

my module works exactly as expected. Again, not sure what happens if you try to do more complex things, or if there's any better way to do the same thing: there might be well known mechanisms to integrate with node modules for instance. I'm pretty much a noobie with all this, so if you have any hint on how to make it better, it would be more than welcome!

## Duktape Licensing
One last note on licensing. Duktape is very easy to integrate because you just copy the required source files, and you compile them along your code. To integrate it into the plugin, I created a subfolder called `duktape-deps` and copied the Duktape source files there. Duktape is licensed under the MIT, but to make the distinction clear from the GPLv3 license of the rest of the Janus project, I copied the License and Authors file of the Duktape project to the `duktape-deps` subfolder: this should make things clearer.

If you have any reason to believe this may cause problems or confusion, please let me know: one alternative might be requiring interested users to download the Duktape source themselves, and then add some configure setting to pass the path to the source so that we can refer to that, but I'd rather avoid these acrobatics if possible.

## That's all, folks!
Please test and, most importantly, provide feedback and suggestions!


RobertLowe:
Can I has v8? 😈 (more over curious why the selection, I have no clue)

Lotta sexy features there and here. 😄

lminiero:
I honestly don't know... you may want to check the Duktape homepage for info on what they're compatible with, in terms of scripting versions.

lminiero:
Unless you meant why Duktape? It was suggested and I just went with it: I didn't do much of a search for alternatives.

RobertLowe:
@lminiero I was referring to the V8 engine (what runs Chrome & Node), it's in c++ however, so many not the best for embedded systems? Although other than memory requirements I'm hard press to think where node/v8 can't run, I could be ignorant my knowledge of embedded systems is low.

---

Duktape does seem to have some module support:

https://github.com/svaarala/duktape/tree/master/extras/module-duktape
https://github.com/svaarala/duktape/tree/master/extras/module-node

Which is great! I'm not sure extras included by default? Outside that I'm not sure what APIs are provided (aka what's on `global`).

Have you considered another approach to interfacing with javascript: a native node module that binds with janus APIs? Outside in, rather than inside out. The advantage there is full support of node/V8 APIs, debugging ( chrome://inspect )?

lminiero:
> Duktape does seem to have some module support:

Yep, that's what I discussed in the PR description above.

> Have you considered another approach to interfacing with javascript: a native node module that binds with janus APIs? Outside in, rather than inside out. The advantage there is full support of node/V8 APIs, debugging ( chrome://inspect )?

Not really, no, and to be honest I'm not planning anything in that direction. Anyway, nothing prevents plugins to be just stubs that then interact with other components however you want (e.g., socket based as well, for instance), so this is definitely something that could be implemented as a third party plugin (as in, not by me :smile: ).

saghul:
Amazing work @lminiero ! I love how we randomly talk and suddenly I check GitHub and there is a PR from you with an implementation :-) Kudos!

lminiero:
@saghul unfortunately I only follow up on your cool ideas a ton of months later... :smile:

lminiero:
Merging this as it's basically ready as it is. We can improve on it along the road.

RobertLowe:
Anyone looking for what functions are available: https://github.com/meetecho/janus-gateway/blob/709fce4e695ddae64dbde3ff2123851068633976/plugins/janus_duktape.c#L1252-L1294 👍

lminiero:
I'm pushing the updated documentation online, so it should appear soon. Anyway, Some functions may indeed not be documented there yet (e.g., `getModulesFolder` and `readFile` that I only added in a later moment).

lminiero:
@RobertLowe documentation pushed now: https://janus.conf.meetecho.com/docs/duktape

RobertLowe:
Awesome @lminiero, can you put any thoughts on how this interops with the existing http/websocket APIs, I imagine some interesting use cases there without needing to add http/socket functionality to duktape. Simplest use case I can think of would be authentication

lminiero:
The best you can do is relay in hmac-signed tokens: https://janus.conf.meetecho.com/docs/auth.html#signed
To use it in a Duktape-based plugin script, you'd have to validate those tokens accordingly. You can check the VideoRoom code to see how that's currently done in C.

lminiero:
That was related to plugin-specific authentication mechanisms. No need to do any interop with the Janus API, instead: those details are abstracted from plugins, and this includes Duktape scripts as well. You only need to react to the callbacks accordingly (e.g., track sessions, handle incoming requests and providing responses/events, etc.)

PR #1265, opened 2018-06-03
Reduce keepalive time to less then 30 seconds

Joe-Palmer:
We have encountered environments where the WebSocket connection is closed after 30 seconds of inactivity. The current Janus keepalive is set to 30 seconds so we have a race condition and we often see the Jauns WebSocket closed in these environments. I have changed this to 25 seconds as this is the default that Socket IO uses (pingInterval):

https://socket.io/docs/server-api/

This change solves the issue for us so I'm proposing it for everyone in case others come across similar environments.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -691,7 +691,7 @@ function Janus(gatewayCallbacks) {
 	function keepAlive() {
 		if(server === null || !websockets || !connected)
 			return;
-		wsKeepaliveTimeoutId = setTimeout(keepAlive, 30000);
+		wsKeepaliveTimeoutId = setTimeout(keepAlive, 25000);
 		var request = { "janus": "keepalive", "session_id": sessionId, "transaction": Janus.randomString(12) };
 		if(token !== null && token !== undefined)
 			request["token"] = token;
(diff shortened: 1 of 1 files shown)


lminiero:
I'd rather not hardcode any change this way. I was already planning for a way to customise longpoll and keep-alive timers via additional properties when creating a session, so that's how I'd like to see something like these 25s enforced instead.

lminiero:
Besides, you can enable the transport-level WS ping/pong mechanism on the server side anyway.

Joe-Palmer:
That makes sense. I figured that as it is already hardcoded, it would be better to hardcode it to a safer value. Although allowing it to be configurable would be best.

Is there any documentation on enabling the transport-level WS ping/pong mechanism?

lminiero:
Only in the cfg file I think: https://github.com/meetecho/janus-gateway/blob/master/conf/janus.transport.websockets.cfg.sample.in#L6

Joe-Palmer:
Great, thanks 👍

lminiero:
Just did the change I was talking about: you can configure the longpoll timeout (e.g., `longPollTimeout: 30000`) and the keepalive period (e.g., `keepAlivePeriod: 5000`) when creating a session. I also changed the keepalive default if you do nothing to `25000`, to address your proposal.

Thanks for the contribution and the discussion, closing this!

PR #1270, opened 2018-06-11
New Nanomsg transport plugin

lminiero:
Pretty self-explainatory: this is a new transport plugin that allows you to communicate with Janus using [Nanomsg](https://nanomsg.org/).

Notice that the only communication pattern this plugin implements is `PAIR`, that is the one-to-one scalability protocol: https://nanomsg.org/v1.1.4/nn_pair.html
This means that you can only have a single client talking to Janus at any given time. This was done because basically I found no way to identify "clients" using the other patterns instead. This is the same limitation the RabbitMQ plugin has, so not a big deal I think: in the future we might think of some kind of RSVP mechanism, where a Pub/Sub approach is used to create an ad-hoc pair for a new client.

The plugin is quite easy to configure. You choose an address, specify if you'll bind to this address (local address) or if you're connecting to it (remote address), and you're done. Tested briefly with a nodejs application using the `nanomsg` npm module and it seems to be working as expected.

As usual, I welcome your feedback, especially if you're quite familiar with Nanomsg and have suggestions to provide before we merge.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -264,6 +264,16 @@ conf_DATA += conf/janus.transport.pfunix.cfg.sample
 EXTRA_DIST += conf/janus.transport.pfunix.cfg.sample
 endif
 
+if ENABLE_NANOMSG
+transport_LTLIBRARIES += transports/libjanus_nanomsg.la
+transports_libjanus_nanomsg_la_SOURCES = transports/janus_nanomsg.c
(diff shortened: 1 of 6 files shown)


mirkobrankovic:
Great work as usual :)
I can suggest Phoenix Pub/Sub as a really good one :D

lminiero:
Merging as I didn't see any objection. Maybe when people start using it we can iron out any bug that may appear.

PR #1271, opened 2018-06-11
systemd unix domain socket support

atom-bomb:
these small changes allow the unix socket transport to use sockets managed by systemd, which provides a useful set of tools for managing the unix socket resources and may improve system startup performance.

here is the relevant systemd documentation:
https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html

here is a nice blog post on the topic:
http://0pointer.de/blog/projects/socket-activation.html

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -258,7 +258,7 @@ if ENABLE_PFUNIX
 transport_LTLIBRARIES += transports/libjanus_pfunix.la
 transports_libjanus_pfunix_la_SOURCES = transports/janus_pfunix.c
 transports_libjanus_pfunix_la_CFLAGS = $(transports_cflags)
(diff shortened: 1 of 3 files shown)


atom-bomb:
ok, thanks for the quick feedback. i've disabled the feature by default. please check my work.. i've tested it but i'm not accustomed to working with autotools.

also, for the sake of a useful commit and comment trail, when compiled with systemd socket support the unix command transport will still work as before when it is not launched by systemd.socket. of course it's safer to disable everything by default.

lminiero:
Thx for the quick fix! I'll have a look later today and merge if it's all fine :+1:

lminiero:
Looks good, merging!

PR #1275, opened 2018-06-13
FEC Implementation

mtbramo:
Lorenzo, I have made a merge with meetecho/janus-gateway/master.
Regards,

 Eric.


plugins/janus_audiobridge.c
```diff
@@ -2781,6 +2787,28 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 		pkt->silence = FALSE;
 		pkt->length = 0;
 
+		/* First check if probation period */
```
lminiero: This probation period is still being done even when FEC is disabled, can you add the check to only do that if FEC is in use? I can do that after the merge otherwise.

plugins/janus_audiobridge.c
```diff
@@ -2846,10 +2874,79 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 			g_free(pkt);
 			return;
 		}
-		pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, BUFFER_SAMPLES, USE_FEC);
-		g_atomic_int_set(&participant->decoding, 0);
+
+		/* Check sequence number received, verify if relevant to expected one. */
```
lminiero: Same here, this should all be done only if FEC is in use, I guess: unless there's a good reason to have it done for all cases?

plugins/janus_audiobridge.c
```diff
@@ -3227,7 +3330,7 @@ static void *janus_audiobridge_handler(void *data) {
 					opus_encoder_ctl(participant->encoder, OPUS_SET_MAX_BANDWIDTH(OPUS_BANDWIDTH_WIDEBAND));
 				}
 				/* FIXME This settings should be configurable */
```
lminiero: I think we can remove the FIXME comment above now.

plugins/janus_audiobridge.c
```diff
@@ -3640,7 +3743,7 @@ static void *janus_audiobridge_handler(void *data) {
 					opus_encoder_ctl(new_encoder, OPUS_SET_MAX_BANDWIDTH(OPUS_BANDWIDTH_WIDEBAND));
 				}
 				/* FIXME This settings should be configurable */
```
lminiero: As above.

plugins/janus_audiobridge.c
```diff
@@ -3905,6 +4008,9 @@ static void *janus_audiobridge_handler(void *data) {
 			if(participant->opus_pt < 0) {
 				/* TODO Handle this case */
 				JANUS_LOG(LOG_ERR, "Offer doesn't contain Opus..?\n");
+			} else {
+				participant->fec = janus_sdp_attribute_fec_enable(offer, participant->opus_pt);
```
lminiero: There are a couple of issues here:

1. `participant->fec` here is set to `TRUE` only when we're working on the answer, which is *after* we handled the `join` most of the times; this means that when we set up the encoder and decoder, `participant->fec` is still `FALSE`. A check on whether or not fec is enabled should be done in the `join` handler code instead, and then used for building the SDP answer as well later on.
2. I personally don't like the idea of a new ad-hoc method in the SDP utils for this FEC thing: the semantics used here are very specific to the Opus codec, and not general to other codecs as well. The `janus_sdp_attribute_*` stuff should be used here instead, without touching the utils. I'll add a comment later showing you how you can detect if FEC is being negotiated by the other party, and how to add the `;useinbandfec=1` part to your answer.

EDIT: actually, looking at the code below, we're already adding the `useinbandfec` part, so what's missing is just the detection of whether or not the user wants to do FEC. This is actually quite easy, and doesn't require any change to SDP utils at all: a simple `strstr(msg->jsep_sdp, "useinbandfec=1")` in the `join` code should do the trick.

sdp-utils.c
```diff
@@ -641,6 +641,38 @@ int janus_sdp_get_codec_pt(janus_sdp *sdp, const char *codec) {
 	return -1;
 }
 
+gboolean janus_sdp_attribute_fec_enable(janus_sdp *sdp, const int codec_pt) {
```
lminiero: See discussion made above: this whole new method is not needed, we're already configuring FEC when creating the attribute using the SDP utils:

	janus_sdp_attribute *a = janus_sdp_attribute_create("fmtp",
		"%d maxplaybackrate=%"SCNu32"; stereo=0; sprop-stereo=0; useinbandfec=%d\r\n",
			participant->opus_pt, participant->room->sampling_rate, participant->fec ? 1 : 0);

sdp-utils.h
```diff
@@ -228,6 +228,12 @@ janus_sdp *janus_sdp_parse(const char *sdp, char *error, size_t errlen);
  * @returns 0 in case of success, a negative integer otherwise */
 int janus_sdp_remove_payload_type(janus_sdp *sdp, int pt);
 
+/*! \brief Helper method to check from attribute status of FEC associate to  specific codec
```
lminiero: See above: please remove the new method.

plugins/janus_audiobridge.c
mtbramo: 1. I doesn't understand, I'm looking at the SDP offer (janus_sdp *offer ....), then we will prepare the answer according from the fec information that we get from offer. You get opus pt from offer too
2. I can use if you want a strstr() and remove all utils change

sdp-utils.h
mtbramo: Ok

sdp-utils.c
mtbramo: Ok

plugins/janus_audiobridge.c
mtbramo: Ok

plugins/janus_audiobridge.c
mtbramo: Whatever you use FEC or not, we can remove probation.

plugins/janus_audiobridge.c
mtbramo: Same previous reply

plugins/janus_audiobridge.c
lminiero: I'm not against this probation, since you explained it was actually useful in some cases. I thought this was mostly useful when doing FEC, though: is this not the case?

plugins/janus_audiobridge.c
lminiero: Yes, what I meant is that the new utils change to detect FEC information is overkill (a strstr is probably enough), and actually wouldn't belong to the SDP utils as too codec specific.

mtbramo:
Lorenzo, I have made new change to take your remarks. 
**Note**: I haven't tested then.

lminiero:
Thanks for the quick fixes! I'll make some tests later and in case it all works fine I'll merge :+1: 
I know I said it many times already, but thanks again for this effort!

lminiero:
The code doesn't compile as it is:

```
plugins/janus_audiobridge.c: In function ‘janus_audiobridge_incoming_rtp’:
plugins/janus_audiobridge.c:2947:16: error: ‘janus_audiobridge_participant {aka struct janus_audiobridge_participant}’ has no member named ‘working’; did you mean ‘talking’?
   participant->working = FALSE;
```

We recently changed the way we "lock" encoders and decoders, due to some crashes that were occurring when cleaning up under heavy load: https://github.com/meetecho/janus-gateway/commit/2a4cabc0c77fc8dc4a24f6e80c1f2eecd39a79f9


		if(!g_atomic_int_compare_and_exchange(&participant->decoding, 0, 1)) {
			/* This means we're cleaning up, so don't try to decode */
			g_free(pkt->data);
			g_free(pkt);
			return;
		}

One more thing I only now noticed is that you're not calling `g_atomic_int_set(&participant->decoding, 0);` in the error branches of your new code. This patch solves both issues:

```
diff --git a/plugins/janus_audiobridge.c b/plugins/janus_audiobridge.c
index 86eee778..33929dc1 100644
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -2934,6 +2934,7 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 		} else {
 			/* In late sequence or sequence wrapped */
 			/* Check first if sequence wrapped */
+			g_atomic_int_set(&participant->decoding, 0);
 			if((participant->expected_seq - pkt->seq_number) > MAX_MISORDER){
 				JANUS_LOG(LOG_HUGE, "SN WRAPPED seq =  %"SCNu16", expected_seq =  %"SCNu16" \n", pkt->seq_number, participant->expected_seq);
 				participant->expected_seq = pkt->seq_number + 1;
@@ -2944,7 +2945,7 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
 			g_free(pkt);
 			return;
 		}
-		participant->working = FALSE;
+		g_atomic_int_set(&participant->decoding, 0);
 		if(pkt->length < 0) {
 			JANUS_LOG(LOG_ERR, "[Opus] Ops! got an error decoding the Opus frame: %"SCNu16" (%s)\n", pkt->length, opus_strerror(pkt->length));
 			g_free(pkt->data);
```

lminiero:
Found another missing call to `g_atomic_int_set(&participant->decoding, 0);`
Anyway, no need for you to apply these fixes: I just tested and it with these fixes it seems to work fine, so I'll just merge and make them myself. Merging!

mtbramo:
Thanks also, sorry for the missing call to g_atomic_int_set(&participant->decoding, 0);

plugins/janus_audiobridge.c
```diff
@@ -2846,10 +2874,79 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, int video, cha
...
+		if(pkt->seq_number == participant->expected_seq) {
+			/* regular decode */
+			pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, BUFFER_SAMPLES, 0);
+
+			/* update last_timestamp */
+			participant->last_timestamp = pkt->timestamp;
+
+			/* Increment according to previous seq_number */
+			participant->expected_seq = pkt->seq_number + 1;
+		} else if(pkt->seq_number > participant->expected_seq) {
```
atoppi: Leaving this as a reminder for the next days.
This comparison is not dealing with uint16 wraparound, leading to unexpected behavior in case of misordered packets occurring near the uint16 upper limit (e.g. expected seq 2, seq number 65533). In particular the expected seq number will be wrongly updated a few lines below.

PR #1286, opened 2018-06-21
Add support for H.264 simulcasting

lminiero:
Read [this update](http://webrtcbydralex.com/index.php/2018/06/21/h-264-finally-a-first-class-citizen-in-webrtc-stacks/) earler today, which explains how H.264 simulcast should now be possible in Chrome, although with some limitations when compared to VP8 simulcast (e.g., no temporal scalability). Since we support VP8 simulcasting already, and it didn't look too much work, I tweaked the code to allow it for H.264 as well in all of the plugins that support it today. I changed the UI in the demos too to hide the temporal layer selection buttons if it's not VP8.

Unfortunately my Chrome Canary still doesn't seem to ship this new functionality, so I couldn't test it (Chrome would still always send a single H.264 stream even with simulcast enabled). I decided to start and publish this anyway just in case someone does have a version that support it: in case you do, please let me know if this works as expected, and in case I'll merge. If not, I'll wait until my Chrome updates to do some more testing myself.

### DIFF
--- a/html/devicetest.js
+++ b/html/devicetest.js
@@ -285,7 +285,7 @@ $(document).ready(function() {
 									if((substream !== null && substream !== undefined) || (temporal !== null && temporal !== undefined)) {
 										if(!simulcastStarted) {
 											simulcastStarted = true;
-											addSimulcastButtons();
+											addSimulcastButtons(temporal !== null && temporal !== undefined);
 										}
 										// We just received notice that there's been a switch, update the buttons
 										updateSimulcastButtons(substream, temporal);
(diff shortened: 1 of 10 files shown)


lminiero:
If you just want one of the temporal layers to be what everyone gets, no, you'll have to only feed that to the plugin yourself, otherwise each viewer can choose what to stick to. Notice there's no temporal scalability support for h264 simulcast, as it's not part of the new chrome patch.

tezet:
Thank you for publishing this change! I can confirm it is working as expected. I was testing it with my WebRTC.org based Android app as a source and Chrome as a receiving end. There were some changes required in WebRTC in order to make use of the new Simulcast adapter on Android, I am not sure if this is the case for Chrome, but definitely it is required to pass following field trial flag: _WebRTC-H264Simulcast/Enabled/_

lminiero:
Oh, thanks for testing and confirming it works! I still see no flag to enable it in my Chrome, so I couldn't test it myself.

PR #1295, opened 2018-07-04
Adds dukcurl for HTTP/HTTPS support to duktape, and libcheck for unit testing

RobertLowe:
This is a first pass and up for review. I thought tests would be a good case here to balance out my skill level, this was mostly gluing in dukcurl. It needs a follow up commit to hook into extras, examples in Janus and assorted packaging.

@lminiero be gentle but thorough, this PR is to open discussion on PoF here.

I feel this can open a lot of possibilities, the tests work, but I haven't verified use inside Janus

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -581,3 +581,9 @@ configs:
 
 clean-local:
 	-rm -f docs/doxygen_sqlite3.db
+
+
+SUBDIRS += tests
+
+check_duktape_extra_curl:
+	$(MAKE) -C tests check_duktape_extra_curl
--- a/configure.ac
+++ b/configure.ac
@@ -218,6 +218,7 @@ AC_ARG_ENABLE([systemd-sockets],
               [],
               [enable_systemd_sockets=no])
 
+
 PKG_CHECK_MODULES([JANUS],
                   [
                     glib-2.0 >= $glib_version
@@ -743,6 +744,21 @@ PKG_CHECK_MODULES([EVENTS],
                   ])
 
 
+PKG_CHECK_MODULES([GLIB], [glib-2.0 >= 2.50.3])
+
+PKG_CHECK_MODULES([CHECK], [
+  glib-2.0 >= $glib_version
+  check >= 0.9.6
+  libmicrohttpd
+])
+
+
+
+AC_HEADER_STDC
+AC_CHECK_HEADERS([stdlib.h])
+AC_FUNC_MALLOC
+
+
 ##
 # JavaScript modules
 ##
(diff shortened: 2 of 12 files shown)


lminiero:
Thanks for experimenting with this! I can already anticipate this is not something we'd merge, though: this is one of the things I'd expect developers to add themselves, e.g., to `janus_duktape_extra.c`, for their own requirements, pretty much as you did in this effort.

RobertLowe:
Just for my own learning, reasons for not merging? is there a different way you'd like to see this packaged? 

Would you anticipate anything to be merged for duktape, it's a pretty limiting VM :(

Thanks!

RobertLowe:
Perhaps there is a larger issue here then for both JS & Lua, both cannot defer to external resources. This at least in some form would enable that for JS.

I completely understand feature creeping and not wanting to maintain third party code, but presently, adding modules like dukcurl aren't drop-in solutions they require developers to modify C source, perhaps there could be better loading mechanisms for LUA & JS which can stay out of core janus? Do you have any thoughts there?

lminiero:
Actually this is not that much a of problem in Lua: when you import external modules, e.g., as installed via `luarocks`, you have access to a lot of stuff, even related to networking. I did use an HTTP stack for some demos that way, and even an MQTT stack IIRC. It's duktape that seems to be much more limiting in that regard, mostly because of how it doesn't provide anything out of the box.

Anyway, it's exactly as you said: we don't want too much stuff baked in there, and I definitely don't want to mantain it. Duktape is not really user-friendly as an engine, and it has always required people messing with the C code to add native features: that's what the `janus_duktape_extra.c` was conceived for.

Not sure how to make this easier in terms of loading mechanisms: I guess one way might be via further shared objects these plugins load themselves (e.g., you create your HTTP stack, make it available as a module, and the duktape plugin loads it at startup in order to register the new functions from there). Anyway, no idea if this can be done in Duktape or not (considering everything is embedded in the project itself) and the possible implications.

That said, I've been thinking of a third module to delegate logic to a "remote" component with IPC, e.g., via nanomsg: this way, this new plugin would have the same C stuff the Lua and Duktape plugins have, but all API calls would be sent to and received from an external application. Didn't have the time to work on it yet, but it's in the back of my mind. This would allow you to write your application logic in whatever language/framework you want (e.g., nodejs as well), as long as it can communicate with the plugin, and leave you complete freedom to do what you like and use what's available there. Native features in Janus (e.g., if you wanted the code to do mixing, which none of these plugins do) would still require working in C and implement the hooks in the plugin, of course, but it should be easier to implement and extend in terms of pure application logic.

RobertLowe:
Thank you for the solid response. I'll investigate LUA more.

An IPC would be awesome. I look forward to your future experiments.

Closing the PR.

PR #1297, opened 2018-07-06
A new ICE stack: jice (Janus ICE)

lminiero:
Hi all,

while this effort is quite immature at this stage, I got to a point where it kinda works in my test setup, and so  I decided to start publishing it anyway. To make it short, it's a first attempt at implementing our own ICE stack, as an alternative to libnice: ideally, it could completely replace libnice in the future, but to be honest it's going to be hard to be feature-pair, and it's going to take a lot of time, which means that in case we go on with jice (which is not sure yet) it's much more likely they'll both stay in for a while.

If you're willing to play with it (knowing it's VERY experimental), we could really use your help and feedback! We plan to start doing some tests next week ourselves, but the more info we have the better. To use jice instead of libnice, all you need to do is pass `-j` or `--jice` on the command line, or set `ice_stack = jice` in `janus.cfg`. Anyway, notice that it has severe limitations when compared to libnice at the moment...

## What is there
At the moment, jice only does ICE Lite, and has no support at all for IPv6, TCP, TURN or other fancy stuff. Even ICE Lite is actually half baked in, really, and so may or may not always work. Anyway, it should be enough to do some tests in most publicly reachable environments, e.g., in a LAN.

While ICE Full is not supported yet, the code can make a test against a STUN server, and there's even a method to detect the kind of NAT Janus is behind, if you pass it two STUN servers: anyway, the method is not invoked anywhere at the moment, we might add something to configure it later on.

## What is NOT there
A lot... the whole ICE process is just sketched at the moment: e.g., priority management, active connectivity checks, proper pairing, nominations, etc.; basically, whatever makes for a complete ICE implementation. We have code almost ready for that, but not really, we'll need to add some logic, and that in part applies to the existing ICE Lite support as well.

As anticipated, IPv6 isn't supported either: that will come sooner or later.

While STUN client support is there (although mostly not used), TURN isn't supported at all. To be honest, that has a very low priority for us at the moment, and if we it turns out jice is promising in terms of performance, TURN is the very last thing we'll work on eventually.

## How can you help?
Just test your applications against a Janus with the jice stack enabled! Hopefully things will work, probably the most often won't. If you enable ICE debugging, the console will show A LOT of jice-related info, so that might help in case you want to share your results.

## That's all, folks
Looking forward to your feedback!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,7 +28,7 @@ CLEANFILES = $(NULL)
 bin_PROGRAMS = janus
 
 headerdir = $(includedir)/janus
-header_HEADERS = apierror.h config.h log.h debug.h mutex.h record.h \
+header_HEADERS = apierror.h config.h log.h debug.h mutex.h record.h jice.h \
 	rtcp.h rtp.h rtpsrtp.h sdp-utils.h ip-utils.h utils.h refcount.h text2pcap.h
 
 pluginsheaderdir = $(includedir)/janus/plugins
@@ -118,6 +118,8 @@ janus_SOURCES = \
 	ice.h \
 	janus.c \
 	janus.h \
+	jice.c \
+	jice.h \
 	log.c \
 	log.h \
 	mutex.h \
--- a/README.md
+++ b/README.md
@@ -342,6 +342,8 @@ or on the command line:
 	-C, --config=filename         Configuration file to use
 	-F, --configs-folder=path     Configuration files folder (default=./conf)
 	-c, --cert-pem=filename       DTLS certificate
+	-j, --jice                    Use jice instead of libnice for ICE
+                                  (default=off)
 	-k, --cert-key=filename       DTLS certificate key
 	-K, --cert-pwd=text           DTLS certificate key passphrase (if needed)
 	-S, --stun-server=filename    STUN server(:port) to use, if needed (e.g.,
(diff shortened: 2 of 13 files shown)


sarumjanuch:
Hi, @lminiero, just out of human curiosity, what is wrong with libnice? Is there any rationale doc that can be read? I mean something like this https://wiki.asterisk.org/wiki/display/AST/RTP+engine+replacement, which will help a third-party person to understand, what is wrong inside.

lminiero:
Added IPv6 support and it seems to be working nicely!

ocrete:
@lminiero There is someone who volunteered to work on removing the global lock from libnice, I think we have a good plan now (see the discussion at https://gitlab.freedesktop.org/libnice/libnice/merge_requests/12 for details). And we actually uncovered some more potential bugs while doing this review, so we should hopefully end up with an even better result.

lminiero:
@ocrete that's excellent news, thanks for sharing! I'm currently on vacation so I missed those notifications :+1:
Looking forward to test the changes as soon as they'll be available!

lminiero:
Note to those interested: there's a new pull request on the libnice repo, which is aiming at doing things the proper way, https://gitlab.freedesktop.org/libnice/libnice/merge_requests/13

PR #1304, opened 2018-07-19
Support screensharing with externally-set constraints

s-ol:
before this commit, attempting to use Janus with externally-set
screensharing properties using

```
  media: {
    video: {
      mandatory: {
        ...
      },
    },
  },
```

would fail if no other video device was connected since the check in
[L2095](https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L2095)
caused Janus to error out before attempting to create the stream.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -2092,7 +2092,7 @@ function Janus(gatewayCallbacks) {
 					var audioExist = devices.some(function(device) {
 						return device.kind === 'audioinput';
 					}),
-					videoExist = devices.some(function(device) {
+					videoExist = isScreenSendEnabled(media) || devices.some(function(device) {
 						return device.kind === 'videoinput';
 					});
 
@@ -3069,6 +3069,22 @@ function Janus(gatewayCallbacks) {
 		if(media.videoRecv === undefined || media.videoRecv === null)
 			return true;	// Default
 		return (media.videoRecv === true);
+	}
+
+	function isScreenSendEnabled(media) {
+		Janus.debug("isScreenSendEnabled:", media);
+		if (media === undefined || media === null)
+			return false;
+		if (typeof media.video !== 'object' || typeof media.video.mandatory !== 'object')
+			return false;
+		const constraints = media.video.mandatory;
+		if (constraints.chromeMediaSource)
+			return constraints.chromeMediaSource === 'desktop' || constraints.chromeMediaSource === 'screen';
+		else if (constraints.mozMediaSource)
+			return constraints.mozMediaSource === 'window' || constraints.mozMediaSource === 'screen';
+		else if (constraints.mediaSource)
+			return constraints.mediaSource === 'window' || constraints.mediaSource === 'screen';
+		return false;
 	}
 
 	function isDataEnabled(media) {


lminiero:
Thanks! I'll review when I get back to the office next week.

s-ol:
@lminiero you are right, the comment doenst make sense anymore. Try to trace my example above: when you specify the mandatory constraints yourself, it will fall through all the Janus-internal screensharing handling until there. Then it will fail because there is no video device set, because janus doesnt realize that the getUserMedia call will ask for screen-media - thats why I added the new function for detecting those calls and ignoring the 'no video' test in those cases.

lminiero:
But that's what I don't get. Your change will make `videoExist=true`, which will lead to this line here: https://github.com/risetechnologies/janus-gateway/blob/a3c0f8d417a462fecb36f87c5e123776670cdc83/html/janus.js#L2129
that is a `navigator.mediaDevices.getUserMedia(gumConstraints)` call that, since we want to screenshare, will fail (it's not a real device). Shouldn't your custom constraints end up in the `getScreen` or `getScreenMedia` calls above instead?

Apologies if I'm missing something obvious, I'm still jetlagged :grin:

lminiero:
Or are you actually trying to _prevent_ the code to get in the `getUserMedia` branch? That is, you configure screensharing, add your constraints, it hits the right place (`getScreen`), but the current code then gets in the `getUserMedia` branch as well due to a broken check your patch fixes?

s-ol:
@lminiero mh, the point of this PR is to enable janus to recognize when the application has **already taken care of the constraint generation** and to then just pass the constraints to `getUserMedia`.

Without this PR this doesn't always work because in the situation where a user has no cameras attached janus.js aborts a request prematurely because it thinks it will fail, since it cannot tell the difference between a screen-capture and a camera-capture constraint. This is what the PR fixes by disabling the explicit check for video availability when it recognizes that the constraints are targetting screensharing.

lminiero:
After a couple of weeks of relaxing vacation, I think I understand your change more. Basically, with your change, a desire to screenshare will not go through our screensharing code anymore when constraints are provided, but will fall down to the regular `getUserMedia` (which is what we do behind the curtains anyway, after the extension has been involved) and be handled there, while previously it would be blocked.

This seems dangerous, though. First of all, in the screensharing code we have several fallbacks, depending on the browser and the browser version. Some will be deprecated soon, and others will be added later on (e.g., as soon as `getDisplayMedia()` becomes available), but still Firefox and Chrome do some things differently at the moment.

To me, it looks like a better fix to the issue would be making the "video is an object and not a string" part aware that the custom constraints may be for screensharing, and not just camera video as it is now: e.g., with an object property that specified the target, that we would remove before passing it to the `getUserMedia`. This would allow `janus.js` to involve the new custom constraints in the same code we already have, without letting a screenshare be created by an area in the code which was not conceived for that (and which would likely fail on some of the browsers, depending on what's done externally).

What do you think about that?

s-ol:
>After a couple of weeks of relaxing vacation, I think I understand your change more. Basically, with your change, a desire to screenshare will not go through our screensharing code anymore when constraints are provided, but will fall down to the regular getUserMedia (which is what we do behind the curtains anyway, after the extension has been involved) and be handled there, while previously it would be blocked.

correct.

>This seems dangerous, though. First of all, in the screensharing code we have several fallbacks, depending on the browser and the browser version. Some will be deprecated soon, and others will be added later on (e.g., as soon as getDisplayMedia() becomes available), but still Firefox and Chrome do some things differently at the moment.

the point would be that an application that is aware of these issues, and cares to handle them explicitly outside of janus.js is enabled to do so.

> To me, it looks like a better fix to the issue would be making the "video is an object and not a string" part aware that the custom constraints may be for screensharing, and not just camera video as it is now: e.g., with an object property that specified the target, that we would remove before passing it to the getUserMedia. This would allow janus.js to involve the new custom constraints in the same code we already have, without letting a screenshare be created by an area in the code which was not conceived for that (and which would likely fail on some of the browsers, depending on what's done externally).

I cannot quite follow your proposal here, the `making the "video is an object and not a string" part aware that the custom constraints may be for screensharing` part sounds like the change I made, no?

lminiero:
> the point would be that an application that is aware of these issues, and cares to handle them explicitly outside of janus.js is enabled to do so.

But that's the problem, if a user is clueless (and most are) they'll likely end up in a branch that breaks their application if they do that. While I understand your requirement, and how this helps you satisfy it, that's the reason why I think it's "dangerous" in its current form. If more control is what you need, it looks like it would be easier to externalise the MediaStream capture process, and then pass it manually to the `createOffer` or `createAnswer`, rather than have janus.js do it for you.

> I cannot quite follow your proposal here, the making the "video is an object and not a string" part aware that the custom constraints may be for screensharing part sounds like the change I made, no?

No, what I meant is that right now, when video is an object, this happens:

	if(media.video && media.video != 'screen' && media.video != 'window') {
		if(typeof media.video === 'object') {
			videoSupport = media.video;
		}

and the whole screensharing branch we have is skipped, as it fails the first check. Your patch skips the whole screensharing branch as well, it just allows it to get to `getUserMedia` if it's a screen capture, which is not the same thing. I'd like the screensharing branch to still be accessed, even in case it's an object, if we can figure out the object represents constraints for screensharing and not a camera.

For how I see it, the problem probably comes from a poor choice in how I allowed custom constraints at the time, that is in overloading the same property (e.g., `video`) with different meanings (boolean, a string, an object). At the moment, I see two ways of incorporating your requirement in janus.js:

1. we extend the object you pass in the constraints with a custom property of `video`, so that we can use it to discriminate the real target (e.g., type: "screen"); the custom property is stripped before passing it to the JSEP APIs;
2. we allow for a different property, e.g., `videoConstraints`, that can contain the custom constraints; this way, if you do `video: true, videoConstraints: { .. }` we know it's for a camera, while `video: "screen", videoConstraints: { .. }` gets us in the screensharing branch and gives us access to the custom constraints; of course, for backwards compatibility setting `video = object` would still work, but would imply camera only.

Does this make sense? I'm not asking you to do this, of course, this is for the sake of discussion, and to make sure we end up with something that fixes your requirement and at the same time ensuring we don't introduce ambiguities in the library behaviour. If you think this is a sane approach, I can try and come up with a PR of my own to see if it works for you.

lminiero:
As a side note, you're probably interested in this commit I just pushed: https://github.com/meetecho/janus-gateway/commit/8455727571e88d29b935351eb1d5863b37dc70b4
I don't know if the new `getDisplayMedia` API has any support for constraints, though.

lminiero:
Thinking about it, there's no reason to overly complicate things as of now... merging your solution, and we can revisit it later on if need arises. Thanks again for your contribution!

PR #1305, opened 2018-07-20
Fixed segmentation fault in pp-h264

thomas-apz:
I fixed on a segmentation fault in janus-pp-rec when using h264.

### DIFF
--- a/postprocessing/pp-h264.c
+++ b/postprocessing/pp-h264.c
@@ -391,6 +391,8 @@ int janus_pp_h264_process(FILE *file, janus_pp_frame_packet *list, int *working)
 					tot -= psize;
 				}
 				/* Done, we'll wait for the next video data to write the frame */
+				if(tmp->next == NULL || tmp->next->ts > tmp->ts)
+					break;
 				tmp = tmp->next;
 				continue;
 			} else if((fragment == 28) || (fragment == 29)) {	/* FIXME true fr FU-A, not FU-B */


lminiero:
Thanks, but I don't understand what this fixes? Without the break, you get to a `tmp = tmp->next` (so now `tmp=NULL`)  followed by a `continue`. This is in a while loop that checks `while(*working && tmp != NULL) {`, and so we would break anyway. What am I missing?

thomas-apz:
Hi Lorenzo, 

Without this break the continue goes directly to the while(1) and the
next instruction is if(tmp->drop) (where I stumbled on a SEGV). 

Am I wrong ? 

Since I added the fix, I encountered no SEGV. 

Thomas 

On 20.07.2018 14:34, Lorenzo Miniero wrote:

> Thanks, but I don't understand what this fixes? Without the break, you get to a tmp = tmp->next (so now tmp=NULL) followed by a continue. This is in a while loop that checks while(*working && tmp != NULL) {, and so we would break anyway. What am I missing? 
> 
> --
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].

 

Links:
------
[1]
https://github.com/meetecho/janus-gateway/pull/1305#issuecomment-406587537
[2]

lminiero:
Ooh, you're right, I missed the inner while... I shouldn't look at code so early in the morning :smile: 
Can you fill in the form in our [CLA](https://cla.conf.meetecho.com/janus/) so that we can merge this?

thomas-apz:
It's done. 

On 20.07.2018 14:51, Lorenzo Miniero wrote:

> Ooh, you're right, I missed the inner while... I shouldn't look at code so early in the morning 😄
> Can you fill in the form in our CLA [1] so that we can merge this? 
> 
> --
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub [2], or mute the thread [3].

 

Links:
------
[1] https://cla.conf.meetecho.com/janus/
[2]
https://github.com/meetecho/janus-gateway/pull/1305#issuecomment-406591450
[3]

lminiero:
Thanks, merging then :+1:

tgabi333:
:+1:

PR #1321, opened 2018-07-27
ICE candidate parser - ICE foundation parsing too short

michelepra:
https://tools.ietf.org/html/rfc5245#section-15.1
rfc for candidate reporting that foundation can be 1*32ice-char but janus parser limit to 31 char


lminiero:
Good catch, thanks! Can you fill the form in our [CLA](https://cla.conf.meetecho.com/janus/) so that we can merge this?

michelepra:
CLA filled

lminiero:
Thanks, merging then!

PR #1322, opened 2018-07-27
add padding for sdes cname

garry81:
sdes chunk should end with null terminator

see: https://tools.ietf.org/html/rfc1889
6.4 SDES: Source description RTCP packet
 Items are contiguous, i.e., items are not individually padded to a
32-bit boundary. Text is not null terminated because some multi-octet
encodings include null octets. The list of items in each chunk is
terminated by one or more null octets, the first of which is
interpreted as an item type of zero to denote the end of the list,
and the remainder as needed to pad until the next 32-bit boundary. A
chunk with zero items (four null octets) is valid but useless.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3259,7 +3259,7 @@ static gboolean janus_ice_outgoing_rtcp_handle(gpointer user_data) {
 	if(stream && stream->component && stream->component->out_stats.audio.packets > 0) {
 		/* Create a SR/SDES compound */
 		int srlen = 28;
-		int sdeslen = 20;
+		int sdeslen = 24;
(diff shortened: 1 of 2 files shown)


garry81:
this will fix webrtc warning
`W/sdes.cc: (line 106): Unexpected end of packet while reading chunk #1. Expected to find text of size`

lminiero:
@garry81 why close this? This was actually a good catch! Sorry if I forgot to answer, I planned to do this at the office later. If that works for you, I can make the change manually: otherwise, feel free to reopen this. We'll need you to sign the CLA before this is merged though.

PR #1323, opened 2018-07-30
add null terminator to sdes cname item

garry81:
sdes item should have one or more null octets as terminator. this will
fix webrtc warning:
Unexpected end of packet while reading chunk #1. Expected to find text of size 10

see: https://tools.ietf.org/html/rfc1889
6.4 SDES: Source description RTCP packet
 Items are contiguous, i.e., items are not individually padded to a
32-bit boundary. Text is not null terminated because some multi-octet
encodings include null octets. The list of items in each chunk is
terminated by one or more null octets, the first of which is
interpreted as an item type of zero to denote the end of the list,
and the remainder as needed to pad until the next 32-bit boundary. A
chunk with zero items (four null octets) is valid but useless.

### DIFF
--- a/rtcp.c
+++ b/rtcp.c
@@ -992,9 +992,9 @@ int janus_rtcp_sdes_cname(char *packet, int len, const char *cname, int cnamelen
 	rtcp->type = RTCP_SDES;
 	rtcp->rc = 1;
 	int plen = 8;	/* Header + chunk + item header */
-	plen += cnamelen+2;
-	if((cnamelen+2)%4)	/* Account for padding */
-		plen += 4;
+	plen += cnamelen+3; /* cname item header(2) + cnamelen + terminator(1) */
+	/* calculate padding length. assume that plen is shorter than 65535 */
+	plen = (plen + 3) & 0xFFFC;
 	if(len < plen) {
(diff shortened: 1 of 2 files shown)


lminiero:
Does this replace #1322? Or did you close that by mistake?

garry81:
Yes, this replace #1322

lminiero:
Thanks! I'll make a couple of tests and, if this works, I'll merge :+1:

rtcp.c
```diff
@@ -992,9 +992,9 @@ int janus_rtcp_sdes_cname(char *packet, int len, const char *cname, int cnamelen
 	rtcp->type = RTCP_SDES;
 	rtcp->rc = 1;
 	int plen = 8;	/* Header + chunk + item header */
-	plen += cnamelen+2;
-	if((cnamelen+2)%4)	/* Account for padding */
-		plen += 4;
+	plen += cnamelen+3; /* cname item header(2) + cnamelen + terminator(1) */
+	/* calculate padding length. assume that plen is shorter than 65535 */
+	plen = (plen + 3) & 0xFFFC;
```
lminiero: Why are you adding `+3` again here? Didn't you already add it to plen before?

lminiero: Nevermind that, just tested and verified it works as expexted.

lminiero:
Thanks, merging!

PR #1324, opened 2018-07-30
Add VP9-SVC support to the Streaming plugin

lminiero:
As the title explains, this patch adds support for VP9-SVC mountpoints to the Streaming plugin. This complements the same support we added to the VideoRoom in #925, as now you can RTP forward a publisher doing VP9-SVC to a Streaming mountpoint, for instance, and it will work as expected. Unlike for VideoRoom publishers, you don't need to enable any flag in Chrome to take advantage of this on the Streaming side: viewers are able to decode VP9 SVC by default.

Feedback welcome! And merging soon in case I don't see objections.

### DIFF
--- a/conf/janus.plugin.streaming.cfg.sample.in
+++ b/conf/janus.plugin.streaming.cfg.sample.in
@@ -41,6 +41,7 @@
 ; videoport3 = third local port for receiving video frames (only for rtp, and simulcasting)
 ; videoskew = yes|no (whether the plugin should perform skew
 ;		analisys and compensation on incoming video RTP stream, EXPERIMENTAL)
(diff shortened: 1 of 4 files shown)


saket424:
Can you include the relevant VP9-SVC mount point snippet that you used in streaming.cfg so I can give it a whirl?

Thanks

lminiero:
I don't have any snippet ready, but apart from forcing VP9 as video codec this is the property you need to add: https://github.com/meetecho/janus-gateway/blob/a212a831ded760e603a7622eebc2d037c776e7e5/conf/janus.plugin.streaming.cfg.sample.in#L44

PR #1325, opened 2018-07-31
fixed the problem that RTP timestamp is skewed when changing stream

hissinger:
hi.

It has been a long time since the last PR. I issued PR that the delay of the video would increase if the stream changed every time. but I accidentally erased the last PR.

I think there is a problem with the timestamp calculation that causes the delay of the video when changing the stream.  If the timestamp is computed as `janus_get_monotonic_time()`, I think there is an error inevitably.  So the last time I added 1000000 to the timestamp to initialize the jitter buffer.

it would be enough to compute the RTP timestamp offset without using `janus_get_monotonic_time()`. With the new code change, there was no accumulation of image delay in the test.


lminiero:
When switching source, we do indeed update sequence numbers and timestamps to make it look like the same stream to the viewer (that's what simulcast is supposed to do anyway), and we do make use of `janus_get_monotonic_time()` to compute the new timestamp according to what we got before, as done here:

https://github.com/meetecho/janus-gateway/blob/master/rtp.c#L454

As you can see, we know the previous timestamp, we know how much passed between the previous packet and this one, and so we calculate the new timestamp based on that difference. Previously we statically increased by, I think, 4500, not caring at all about how much time passed, and the result was obviously often terrible.

I don't think adding `1000000` helps. It might help resetting the jitter buffer, as you say, but it will mess all other RTCP statistics we send after that, because they won't match anymore: as such, I can't merge this. If you think there's a problem in how the new timestamp is calculated, please fix it in `janus_rtp_header_update` instead, or in other places you think may cause this issue.

hissinger:
I agree that adding `1000000` is not a good solution as you said. So I also delete the code from this PR.

And, Is it because the delay of the video is gradually increased every time the stream of the image is changed because the new timestamp is calculated based on `janus_get_monotonic_time()`? I think the difference in the timestamp between the packets originally sent by the source is different.

lminiero:
`janus_get_monotonic_time()` is only used when there's an SSRC change. In that case, since the timestamps are different and we want them to be in the same "space" and make them continuous, we have to figure out how to "translate" the new timestamps to what they were before. The way we do it is calculate how much time passed before the previous (old SSRC) packet, and the (new SSRC) packet we just received. We do that with `janus_get_monotonic_time()` because we also keep track of when we sent the last one, using the very same method. This way, if we know the new packet arrived 50ms after the previous one, we can compute the new timestamp from the 90000Hz and the 50ms interval, which in this case gives us `+4500`. This way, the timestamps are supposed to increase exactly as if the stream never changed.

Not sure what may be causing the desync in your case. You may want to play with that part, e.g., the way `time_diff` is calculated in that part. Another option is printing out `time_diff` when it's calculated as you use the simulcasting feature, so that when you see a delay you can see if there's a bug there.

hissinger:
I have already understood that you used `janus_get_monotonic_time()` to calculate the timestamp. However, if you measure the time with `janus_get_monotonic_time()`, it is time to arrive at the media server(janus), so it is the time including the delay in the network.

Therefore, when the video stream is changed, the RTP timestamp of the video received by the viewer may be later than the timestamp of the original source. If the stream is changed many times, the difference will be bigger, and the difference from the timestamp of the audio RTP will also be bigger, so it is considered that the audio and video are not synchronized.

hissinger:
As you know, the transmission delay in the network can vary over time. If the delay is large at the time the video stream changes, the resulting timestamp difference can not be recovered.

PR #1333, opened 2018-08-09
url-encoding token and apisecret parameters

agclark81:
url-encoding token and apisecret parameters in long-poll request since both parameters can contain special characters (e.g., = at the end of an hmac signature) which, if left unencoded, will cause the token verification to fail and the session to expire if there is no other activity occurring


lminiero:
Thanks! I'll have a look as soon as I get back to the office, which should happen in a couple of days.

lminiero:
Looks good, merging!

PR #1337, opened 2018-08-16
Videocall: decline incoming call fix

Lestrangge:
This fixes the critical issue with declining the incoming call.
With add we quickly get the 'hangingup' property of the handle being set to 1 and then 2, so we can only decline a call once, after that we continue having a 'peer', so, we continue recieving a call

### DIFF
--- a/plugins/janus_videocall.c
+++ b/plugins/janus_videocall.c
@@ -1004,7 +1004,7 @@ void janus_videocall_hangup_media(janus_plugin_session *handle) {
 	}
 	if(g_atomic_int_get(&session->destroyed))
 		return;
-	if(g_atomic_int_add(&session->hangingup, 1))
+	if(g_atomic_int_and(&session->hangingup, 1))
 		return;
 	/* Get rid of the recorders, if available */
 	janus_mutex_lock(&session->rec_mutex);


lminiero:
Thanks for the contribution! I'm still out of the office so can't do a review right now. I'll have to look into this better, as we use this kind of atomic base construct a lot, and so want to be 100% sure we're doing something wrong before changing it.

lminiero:
Ok, if I understood the change correctly, [g_atomic_int_and](https://developer.gnome.org/glib/stable/glib-Atomic-Operations.html#g-atomic-int-and) behaves differently: by passing `1`, it means that if the current value is different from 0 we always return, and since the result of the AND is 1, that's what `hangingup` stays at. With the `g_atomic_int_add` we have, instead, if `hangingup` was `1`, now it becomes `2`.

Anyway, I'm not sure I understand why this should fix the issue. In both cases, if `hangingup` is greater than zero the method causes `janus_videocall_hangup_media` to immediately return, as the return value is non-zero all the same (1 or 2 it doesn't matter).

Am I missing something obvious? (which might be, I've just returned from vacation :slightly_smiling_face: )

lminiero:
Actually, there is a considerable difference, which might break things. If `hangingup` is 0, `g_atomic_int_and` will allow the method to continue, but will leave `hangingup` to 0 (0 AND 1 = 0), which means that if we go through `janus_videocall_hangup_media` it will go through again, which is wrong (we have that value exactly to ensure we only pass through that once per call). With `g_atomic_int_add`, it goes through and `hangingup` is now 1, which is what we want.

If you're experiencing issues when calls are being declined, I don't think the issue is there, but somewhere else. Can you elaborate on the scenario that breaks for you, maybe with some steps to replicate?

lminiero:
Maybe we should use [g_atomic_int_compare_and_exchange](https://developer.gnome.org/glib/stable/glib-Atomic-Operations.html#g-atomic-int-compare-and-exchange) instead?

Lestrangge:
> If you're experiencing issues when calls are being declined, I don't think the issue is there, but somewhere else. Can you elaborate on the scenario that breaks for you, maybe with some steps to replicate?

Its really easy to reproduce the bug
Auth two users, say, Alice & Bob
Alice calls Bob
Bob declines
Alice calls Bob
Bob declines
Bob is said to still be in a call with Alice

I'll try to work out why does ' g_atomic_int_and' help after finishing the current task)

Lestrangge:
ooops
Didnt mean to close it .-.

lminiero:
> Its really easy to reproduce the bug

Thanks for the steps, I'll have a look as well.

Lestrangge:
I see, hangingup is now never set to 1 and is always 0, so we always perform the whole function. Trying to find another solution

lminiero:
I've already done some tests with `g_atomic_int_compare_and_exchange` and it solves your steps for me. I'll push in a few minutes.

lminiero:
The above commit should fix it for you: please let me know if that's not the case. Closing this PR as it's not needed anymore.

PR #1339, opened 2018-08-21
Duplication fix

Lestrangge:
Mistype fix)


lminiero:

lminiero:
Looks like we have the same problem in the Lua version, which is where I copied this from... damn copy/paste :wink:  I'll fix it there as well.

PR #1340, opened 2018-08-22
Janus.js reconnect() fix

Lestrangge:
When changing the network or loosing connection to one and trying to perform more than one api call, unhandled exception is recieved (cant take index of of undefined) as server is set to undefined. Forcely retrying all servers looks like a better idea (also helps establishing connection to one account from different devices as we forcely retry all servers)

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -449,8 +449,6 @@ function Janus(gatewayCallbacks) {
 		callbacks.success = (typeof callbacks.success == "function") ? callbacks.success : Janus.noop;
 		callbacks.error = (typeof callbacks.error == "function") ? callbacks.error : Janus.noop;
 		callbacks["reconnect"] = true;
-		server=null;
-		serversIndex=0;
 		createSession(callbacks);
 	};
 	this.getSessionId = function() { return sessionId; };
@@ -753,7 +751,7 @@ function Janus(gatewayCallbacks) {
 			wsHandlers = {
 				'error': function() {
 					Janus.error("Error connecting to the Janus WebSockets server... " + server);
-					if (Janus.isArray(servers)) {
+					if (Janus.isArray(servers) && !callbacks["reconnect"]) {
 						serversIndex++;
 						if (serversIndex == servers.length) {
 							// We tried all the servers the user gave us and they all failed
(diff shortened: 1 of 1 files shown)


lminiero:
Forcely retrying all servers is not a good idea, as the purpose of that method is reconnecting to the same server where your session was taking place. If you end up somewhere else that will fail.

lminiero:
> Oh i see, multiple janus instances are possible and this solution does not work in this case. I only use it to fallback from wss to https so...)

But you don't need the reconnect feature to fallback. Just pass an array of servers instead of a single server when creating a Janus object, and it will automatically fallback in case the previous one doesn't work (see the `server` documentation [here](https://janus.conf.meetecho.com/docs/JS)).

Lestrangge:
I see. I use reconnect feature when the handle looses connecction
I mean, you use reconnect func once, and the server id get grown by one. And not getting lower on reconnect error

lminiero:
> I use reconnect feature when the handle looses connecction

Ah ok, got it. Yes, in that case the fallback I mentioned would not work for you, as it only works the first time. For your specific use case, I guess you can add an optional property, that you can intercept via the callback object: e.g.

    janus.reconnect({ server: "wss://janus.example2.com/ws", ... });

or

    janus.reconnect({ server: [ "https://janus.example2.com/janus", "wss.//janus.example2.com/ws" ], ... });

meaning you can then check for the value of `callbacks["server"]` in `reconnect, and do something else than what we do right now. Anyway, changing server would mean you'd have to reset the `sessionId` and all the handles as well, as it's very likely you'll end up on a separate server, and so all those properties would now be invalid.

Lestrangge:
Yeah, i see. Actually added reducing the server id on error, works allright for my case

lminiero:
> Actually added reducing the server id on error, works allright for my case

Not sure what you mean by this? What did fix your issue?

Lestrangge:
Ive updated the pr
This works pretty good, now we can only reconnect to server we were connected to

html/janus.js
```diff
@@ -835,9 +835,9 @@ function Janus(gatewayCallbacks) {
 			},
 			error: function(textStatus, errorThrown) {
 				Janus.error(textStatus + ":", errorThrown);	// FIXME
-				if(Janus.isArray(servers)) {
+				if(Janus.isArray(servers) && !callbacks["reconnect"]) {
 					serversIndex++;
-					if(serversIndex == servers.length) {
+					if(serversIndex == servers.length ) {
```
lminiero: You added an extra whitespace by mistake, here.

lminiero:
I just spotted a small typo, if you can fix that I'll merge :+1:

Lestrangge:
Done ^-^

lminiero:
Thanks, merging then!

PR #1341, opened 2018-08-28
fixes for nosip_plugin

DmitryYudin:
Please, see the commits comment for details.

### DIFF
--- a/plugins/janus_nosip.c
+++ b/plugins/janus_nosip.c
@@ -1340,7 +1340,7 @@ static void *janus_nosip_handler(void *data) {
 					/* Clean up SRTP stuff from before first, in case it's still needed */
 					janus_nosip_srtp_cleanup(session);
 				}
-				janus_nosip_sdp_process(session, parsed_sdp, !offer, FALSE, &changed);
+				janus_nosip_sdp_process(session, parsed_sdp, !offer, sdp_update, &changed);
 				/* Check if offer has neither audio nor video, fail */
 				if(!session->media.has_audio && !session->media.has_video) {
 					JANUS_LOG(LOG_ERR, "No audio and no video being negotiated\n");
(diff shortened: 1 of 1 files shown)


plugins/janus_nosip.c
```diff
@@ -1186,6 +1186,11 @@ static void *janus_nosip_handler(void *data) {
 			const char *msg_sdp_type = json_string_value(json_object_get(generate ? msg->jsep : root, "type"));
 			const char *msg_sdp = json_string_value(json_object_get(generate ? msg->jsep : root, "sdp"));
 			gboolean sdp_update = json_is_true(json_object_get(generate ? msg->jsep : root, "update"));
+			if(!generate) {
+				if(session->media.ready) {
```
lminiero: Why two layers of if when they can be together?

plugins/janus_nosip.c
```diff
@@ -1988,7 +1993,7 @@ static void *janus_nosip_relay_thread(void *data) {
 			!g_atomic_int_get(&session->destroyed) && !g_atomic_int_get(&session->hangingup)) {
 		if(session->media.updated) {
 			/* Apparently there was a session update */
-			if(have_server_ip && (inet_aton(session->media.remote_ip, &server_addr.sin_addr) == 0)) {
+			if(have_server_ip && (inet_aton(session->media.remote_ip, &server_addr.sin_addr) != 0)) {
```
lminiero: Aaah, good catch: I checked the SIP and SIPre plugins and they're already correct, so I guess it was either a copy/paste error at the time, or a failure to fix this one as well when we fixed those.

plugins/janus_nosip.c
DmitryYudin: single line - ok)

DmitryYudin: force updated

lminiero:
Thanks, merging then!

PR #1342, opened 2018-08-29
relay REMBs of the lowest peer bandwidth estimation

adam-briggs:
it looks like the janus streaming plugin has been sending the streaming source REMBs of the highest incoming bandwidth estimation rather than the lowest.
in my limited testing, this change seems to work.


lminiero:
Ah, good catch, what a silly typo!

By the way, did you change your Github handle? It shows you as a first time contributor, but I remember you using `atom-bomb` as username. If so, could you sign the CLA again? Thanks!

adam-briggs:
oh sry and thanks for the reminder. i have another github account for my day job. i have now signed the CLA again.

lminiero:
Thanks, merging then!

PR #1343, opened 2018-08-29
fix (nosip): open local ports according to audio/video presence

DmitryYudin:
1. No reasons to reopen local ports (i.e. close then open).
2. Since has_[audio|video] flags are sticky there are no events on which we should close (already opened) local ports.
3. (fix) Need to open appropriate local port if media presence signaled

Context:
Initial offer does not contain a presence of video stream, but second does. We set has_video=FALSE. On renegotiation we receive an answer and set has_video=TRUE, but local port remains closed

Further consideration:
1. The same issue 1/2/3 may be also addressed to other yesSIP plugins
2. Please, consider to move pipefd construction from allocate_local_port() to janus_nosip_relay_thread() bootstrap

### DIFF
--- a/plugins/janus_nosip.c
+++ b/plugins/janus_nosip.c
@@ -581,7 +581,7 @@ static void janus_nosip_srtp_cleanup(janus_nosip_session *session) {
 void janus_nosip_sdp_process(janus_nosip_session *session, janus_sdp *sdp, gboolean answer, gboolean update, gboolean *changed);
 char *janus_nosip_sdp_manipulate(janus_nosip_session *session, janus_sdp *sdp, gboolean answer);
 /* Media */
-static int janus_nosip_allocate_local_ports(janus_nosip_session *session);
+static int janus_nosip_allocate_local_ports(janus_nosip_session *session, gboolean update);
 static void *janus_nosip_relay_thread(void *data);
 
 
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for this additional fix! I'll need some time to evaluate and review, I'll get back to you as soon as I have a clearer idea of the implications.

plugins/janus_nosip.c
```diff
@@ -1762,41 +1761,21 @@ char *janus_nosip_sdp_manipulate(janus_nosip_session *session, janus_sdp *sdp, g
...
-	session->media.local_video_rtcp_port = 0;
-	session->media.video_ssrc = 0;
-	if(session->media.pipefd[0] > 0) {
-		close(session->media.pipefd[0]);
-		session->media.pipefd[0] = -1;
-	}
-	if(session->media.pipefd[1] > 0) {
-		close(session->media.pipefd[1]);
-		session->media.pipefd[1] = -1;
+	if(!update) {
```
lminiero: I understand the usage of the new property, but why is the `!update` case only resetting `pipefd`, and not the audio and video RTP/RTCP stuff as well as we did before? Since this is the case of the call just starting, we need to be sure we're starting fresh. While we close those sockets and reset them when the relay thread closes, it's safer to do it here as well.

plugins/janus_nosip.c
```diff
@@ -1899,7 +1878,13 @@ static int janus_nosip_allocate_local_ports(janus_nosip_session *session) {
 		}
 	}
 	/* We need this to quickly interrupt the poll when it's time to update a session or wrap up */
-	pipe(session->media.pipefd);
+	if(!update) {
+		pipe(session->media.pipefd);
+	}
+	/* Let relay_thread to accommodate the changes */
+	if(update) {
+		session->media.updated = TRUE;
```
lminiero: This is currently already done at the end of `janus_nosip_sdp_process`, which also uses `pipefd` to wake the relay thread. We do it there because it's when the peer sends us an SDP, and so we use the potentially new port to reconnect the sockets.

My feeling is that you only need this here as well if the local SDP is an answer, as in that case there wouldn't be another SDP in response to trigger `janus_nosip_sdp_process`. As such, the `if(update)` should become something like if(update && answer)`, and you'll need to add the pipe wakeup here too. Does this make sense in your scenario?

plugins/janus_nosip.c
DmitryYudin: As far as I understand current design the `!update` case is granted to be the first call to `_allocate` routine(). Please, correct me if I wrong. The idea was only to create `pipefd` once.
Even in your case, sockets and pipe remain opened if `relay_thread` has not been created for some reasons. Since you manage session life-time with reference counter, it is also might be safer to release these resources on session destroy.

lminiero: You can have multiple calls on the same handle in sequence, though. When you close a call, and then start a new one on the same handle (so not a renegotiation), you'll again hit the allocate method with a `!update`. Since you may have used audio and video in the previous call, we need to reset those sockets as well, as I don't want them to be reused if it's a different call.

plugins/janus_nosip.c
DmitryYudin: You right, the `janus_nosip_sdp_process` wakeup thread on next message.
I think just a
`if(update) {
    wakeup()
}`
would be the correct behaviour. Please, confirm.

DmitryYudin:
Force pushed. 
Please, review.

DmitryYudin:
Oh sorry, for that.
I'm a newcomer for github.

pushed

lminiero:
Just found some time to play with this and it looks like it's working as expected, thanks! I modified the `nosiptest.js` code to not send video in the caller's `createOffer`, and only added video subsequently with this command:

	caller.createOffer(
		{
			media: { addVideo: true },
			success: function(jsep) {
				Janus.debug("[caller] Got SDP!");
				Janus.debug(jsep);
				var body = {
					request: "generate"
				};
				caller.send({message: body, jsep: jsep});
			},
			error: function(error) {
				Janus.error("WebRTC error:", error);
			}
		});

and I could see video appearing both locally and remotely.

I also found a small issue, though, namely an error connecting the sockets at a certain point:

```
[4423498386410921] Negotiation update, checking what changed...
[ERR] [plugins/janus_nosip.c:janus_nosip_connect_sockets:1946] [NoSIP-0x619000080280] Couldn't connect video RTP? (192.168.1.80:32708)
[ERR] [plugins/janus_nosip.c:janus_nosip_connect_sockets:1947] [NoSIP-0x619000080280]   -- 9 (Bad file descriptor)
[ERR] [plugins/janus_nosip.c:janus_nosip_connect_sockets:1953] [NoSIP-0x619000080280] Couldn't connect video RTCP? (192.168.1.80:32709)
[ERR] [plugins/janus_nosip.c:janus_nosip_connect_sockets:1954] [NoSIP-0x619000080280]   -- 9 (Bad file descriptor)
```

In this case, it was the callee not being able to connect the sockets to the caller's ports. Not sure if this was fatal or not, but I don't think so (I could still see and hear). I think this is caused by the fact that the `session->media.updated = TRUE` part is done both when receiving a remote SDP and when handling an update coming from the user: as a test, I changed line 1696 to this:

	if(answer && update && changed && *changed) {

so that when processing remote SDPs we only reconnect the sockets if it's answer, and the error disappeared: media also seemed fine, but I cannot say for sure as the default NoSIP demo only gives the perspective of the caller and not the one of the callee.

Could you let me know if you're getting the same error, and if the fix I suggest keeps it working in your application as well?

DmitryYudin:
In my specific environment I do not observe "Couldn't connect video RTP?" message as well as any side-effect on "if(answer .. )" fix.

DmitryYudin:
Should I amend "if(answer .. )" to PR or you'd like to arrange for stand alone commit?

lminiero:
No need for any further commit, I'll merge as it is and just take the `if(answer)` thing into account in case it turns out to be needed when people start using it. Thanks again for your contribution!

PR #1345, opened 2018-09-01
use dtls timeout value only when DTLSv1_get_timeout is succeed

garry81:
timeout_value is 0 even when DTLSv1_get_timeout is failed. it makes to perform DTLS retransmission too often. It's better to check return value of DTLSv1_get_timeout to avoid this problem.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -1076,7 +1076,10 @@ gboolean janus_dtls_retry(gpointer stack) {
 		goto stoptimer;
 	}
 	struct timeval timeout = {0};
-	DTLSv1_get_timeout(dtls->ssl, &timeout);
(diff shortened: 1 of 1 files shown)


lminiero:
What are the reasons that may cause `DTLSv1_get_timeout` to fail?

garry81:
not sure. it shouldn't be failed after handshake with timeout value but it does. :( Might be a problem of boringssl.
Anyway, when a device tries to connect to janus on the connection with long latency(RTT was 300ms), janus emitted DTLS timeout msg more than 10 times in a second. after I checked return value of `DTLSv1_get_timeout`, timeout was triggered as expected.(100ms, 200ms, 400ms, ...)

lminiero:
Got it, thanks for the clarification. Merging then!

PR #1348, opened 2018-09-04
ES6 -> ES5 replace

alienpavlov:
replaced **const** to **var**. my IDE complains on it because it configured to ES5 support only


lminiero:
Ack, merging :+1:

PR #1349, opened 2018-09-05
New MQTT event handler

lminiero:
This is a small refactoring of the original effort made by @oej in #1185, and adds a new MQTT-based event handler to Janus. I tried to address my review comments to the original PR, and aligned that version to where the handlers are now. This also includes a couple of minor changes to the other handlers (fixes I spotted while working on this) and revisited the helper methods in `events.c/.h`.

I tested this briefly with a very dumb nodejs based subscriber:

```
var mqtt = require('mqtt')
var client  = mqtt.connect('mqtt://localhost:1883')

client.on('connect', function () {
	var events = [ "sessions", "handles", "external",
		"jsep", "webrtc", "media", "plugins", "transports", "core" ];
	for(var i in events) {
		client.subscribe('/janus/events/' + events[i], function (err) {
			if(err)
				console.log("Error subscribing:", err);
		});
	}
	console.log("Connected");
});

client.on('message', function(topic, message) {
	console.log("[" + topic + "] ", message.toString());
});
```

and it seems to work, but I'd love some feedback as I didn't test much more than that.

A few notes that may be relevant, especially to @oej as it may break his current setup:

1. the original patch had the wrong name for a few event types: e.g., `session` instead of `sessions`, `handle` instead of `handles`, etc. I Fixed this in the current version, but this means that recipients configured to work with the old version may need to fix the subscriptions they make (e.g., `/janus/events/sessions` rather than `/janus/events/session`).
2. It looks like from time to time (this doesn't seem consistent) the first core event fails to deliver:
```
[WARN] FAILURE: MQTT EVH message propably not sent to topic /janus/events/core on tcp://localhost:1883. Result -3
[WARN] Can't publish to MQTT topic: /janus/events, return code: -3
````
I guess this may be related to the `Improve the MQTT connect message (the first one)` TODO @oej had as a memo in his version?

3. The MQTT event handler doesn't support grouping of events: I guess this was removed in the original version because it is configured to be able to send events on different queues, rather than a single one (as HTTP and RMQ do), which would make the concept of grouping much harder to implement.
4. I added the new "tweak" functionality to this module as well, but the only thing it allows you to do for now is choosing which events to push; not sure what else is dynamic and can be changed, but that's for future enhancements (e.g., change the server to send events to).

Feedback welcome!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -311,6 +311,16 @@ conf_DATA += conf/janus.eventhandler.rabbitmqevh.cfg.sample
 EXTRA_DIST += conf/janus.eventhandler.rabbitmqevh.cfg.sample
 endif
 
+if ENABLE_MQTTEVH
+event_LTLIBRARIES += events/libjanus_mqttevh.la
+events_libjanus_mqttevh_la_SOURCES = events/janus_mqttevh.c
+events_libjanus_mqttevh_la_CFLAGS = $(events_cflags)
+events_libjanus_mqttevh_la_LDFLAGS = $(events_ldflags) -lpaho-mqtt3as
+events_libjanus_mqttevh_la_LIBADD = $(events_libadd)
+conf_DATA += conf/janus.eventhandler.mqttevh.cfg.sample
+EXTRA_DIST += conf/janus.eventhandler.mqttevh.cfg.sample
+endif
+
 ##
 # Plugins
 ##
--- a/README.md
+++ b/README.md
@@ -33,9 +33,9 @@ you are interested in WebSockets support for the Janus API)
 * [cmake](http://www.cmake.org/) (only needed if you are interested in
 WebSockets and/or BoringSSL support, as they make use of it)
 * [rabbitmq-c](https://github.com/alanxz/rabbitmq-c) (only needed if
-you are interested in RabbitMQ support for the Janus API)
+you are interested in RabbitMQ support for the Janus API or events)
 * [paho.mqtt.c](https://eclipse.org/paho/clients/c) (only needed if
-you are interested in MQTT support for the Janus API)
+you are interested in MQTT support for the Janus API or events)
 * [nanomsg](https://nanomsg.org/) (only needed if
 you are interested in Nanomsg support for the Janus API)
 * [libcurl](https://curl.haxx.se/libcurl/) (only needed if you are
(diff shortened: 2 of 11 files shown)


oej:
Comments:
1. OK
2. Will check. There's probably a timing issue with the paho threads. We need to wait until some event has been fired.
3. Grouping was considered agains the model I used for the topics. 
4. MQTT has SRV records that well, hrrm, should be implemented for failover. Agree that some sort of failover mechanism would be good. Pushing events to different servers based on a filter seems very advance, but may be a cool feature (TM).

The general idea is to publish EVERYTHING and be able to filter on your subscriptions of topics or subtopics with wildcards in the receiving end. Filtering on publishing seems like something that may be codable and interesting, but a bit against the publish/subscribe model.

lminiero:
> Filtering on publishing seems like something that may be codable and interesting, but a bit against the publish/subscribe model.

Actually this is already part of the core, now: when you configure what you want to receive as a plugin, this updates a mask that is a property of the struct itself, and the core will not pass you the event at all if you don't want it. This helps when you don't care, for instance, for frequent events like RTCP stats.

oej:
Oh, I thought of more deep filtering. The category of events setup is something I'm all behind. Sorry for the confusion.

lminiero:
Yep, I agree deep filtering (e.g., for the same event type, send some but
not others) is best left up to the recipient (although I'm planning ways to
throttle at least RTCP stats depending on runtime API requests).

Il mer 5 set 2018, 15:05 Olle E. Johansson <notifications@github.com> ha
scritto:

oej:
One may want either the on-going during-session updates or a summary after each session. I would suggest two different categories of events for these.

lminiero:
If you mean RTCP, we already added them for events, after someone asked for
this. Not sure how they're enabled as I'm on the phone and AFK, but IIRC
you always get a summary when the PC is closed now, even when you set the
period for the stats to 0.

Il mer 5 set 2018, 15:39 Olle E. Johansson <notifications@github.com> ha
scritto:

lminiero:
@oej any objection to merge this as it is? AFAICT, apart from the event names I mentioned and a couple of fixes from my original review, it should pretty much work as your version did already. We can then start adding the rest along the road.

If you'd rather test it first (in case you didn't already) no problem, and I'll wait for the green light :+1:

oej:
It's not like we have to bother with backwards compatibility with me and possible unknown testers of my code. Go ahead and merge and we'll take it from there. Thank you very much for your efforts!

oej:
And I'll test in a while.

lminiero:
Well, I'd still like it not to break your environment just because you upgraded! :smile: After all, this was your creation and you've been using it for some time already. Anyway, I'll merge then and we'll resume from there. Thanks again for working on this!

PR #1350, opened 2018-09-06
fix (janus): unneeded mutex_unlock() on handle_create() fail

DmitryYudin:


lminiero:
Mh, looks like there is another one in `janus_ice_handle_attach_plugin`:
https://github.com/meetecho/janus-gateway/blob/master/ice.c#L1003

I'll have to check if this is a leftover, or if it's a problem: it might be that we are actually NOT locking the mutex while we should. Thanks for spotting this, I'll have a look at the code.

lminiero:
The mutex is used directly within `janus_session_handles_find`, `janus_session_handles_insert`, `janus_session_handles_remove` and `janus_session_handles_clear`, so those extra calls are indeed unneeded. I'll merge your fix and make the other one myself. Thanks for spotting this!

PR #1353, opened 2018-09-07
ice: satinary check

DmitryYudin:
just to make it clear for the reader how the `handle->stream->component` and `component` are interconnected


lminiero:
That cannot happen. While in the past we allowed for multiple components (we supported non-bundle), now all PeerConnections have exactly 1 stream and 1 component. I don't think this check is needed.

DmitryYudin:
If so, it migth be better to remove `component` from argument list due to no use.

lminiero:
That may be, but this will happen in a later refactoring, if and when I'll start looking into unified plan. Closing.

PR #1356, opened 2018-09-11
Handle when retransmission buffer is NULL

zane-hpcnt:
This produced `GLib-CRITICAL **: g_hash_table_lookup: assertion 'hash_table != NULL' failed` messages


lminiero:
Thanks for your contribution! I'll have a look at the code and get back to you, as the NACKs part is delicate and I want to make sure we don't introduce regressions with this.

lminiero:
Ok, just had a look at the code and I can definitely see what you're fixing here. The `audio_retransmit_seqs` and `video_retransmit_seqs` might be NULL, although that should only happen when NACKs are disabled or `max_nack_queue` is 0: in that case, incoming NACKs would try to look for packets to retransmit in a non-existing hashtable, hence the assertion. I'm wondering if we should add a check before that part to avoid processing NACKs at all when they're disabled for the session, but for now your fix seems to do the trick.

I see you signed the CLA already, so merging, thanks again for the help!

PR #1357, opened 2018-09-11
clear nacked packet properly in case of rfc4588

garry81:


lminiero:
Thanks for your contribution! I'll get back to you after I've studied the code a bit better: that code is responsible for the generation of NACKs, and I just fixed a broken part last week, so I want to be 100% sure we're not breaking anything now.

garry81:
`cur_seq->seq` is inserted to hash table but `new_seqn` is used as key in `janus_ice_nacked_packet_cleanup`
I just changed to pass `cur_seq->seq` to `janus_ice_nacked_packet_cleanup` to remove proper item

lminiero:
You're right, that piece of code is even in the `else` of the `if(cur_seq->seq == new_seqn) {`, which means `new_seqn` is most definitely not the sequence number we're handling and should remove later... good catch, thanks for spotting this! Merging :+1:

PR #1358, opened 2018-09-11
codec detection during videocall accept

codebot:
I'm a bit afraid that I don't fully understand everything that is happening, but I believe that #1286 may have inadvertently changed the logic for detecting codecs during call acceptance.

Maybe it's just how I am using Janus (wrapped on a server) but without this change, I am unable to use the `set` API to start recording. Video calls start as expected, but since `has_audio` and `has_video` are both `FALSE`, the `set` handler does not create a recorder. With this change, recording starts as expected upon receipt of a `set` message containing `record: true`.

Thanks for your time. Cheers :tada:


lminiero:
Yeah, definitely looks like a copy/paste error, thanks for spotting this! I see you've signed the CLA already, so merging right away :+1:

PR #1359, opened 2018-09-11
Simulcast support for RTP forwarder and recordings

lminiero:
While simulcast support has been around for a while, there were a couple of areas were it was still a bit sketchy. In particular, for instance, simulcast publishers in the VideoRoom could not be recorder properly, nor could you forward the "best" stream to a single endpoint either. The way we handled this so far was to always record the base substream when simulcast was available (thus meaning the recording would be quite low resolution, even if good video was available), and for RTP forwarding we had substream support, meaning that you could relay the individual substreams to three different ports.

For recordings the limitation was obvious, for RTP forwarding it isn't, as that part is already helpful as it is (being able to relay the individual substreams to separate ports merges well with the simulcast support in the Streaming plugin). Anyway, there are cases where you don't want the individual substreams to be relayed, but want the best current video stream to be forwarded to a separate endpoint, thus having a forwarder act pretty much as a regular viewer (assuming of course that the recipient can receive and handle varying resolutions).

This patch addresses exactly this limitation. More specifically, it moves the processing of simulcast RTP packets to a dedicated core function, adding a support structure that is similar to the already existing "RTP switching context" in concept. This allowed us to update the plugins that were using simulcasting before to use the dedicated function instead: first of all, this had the obvious advantage of cleaning up the code and avoiding redundances, but also meant it would ease the integration of simulcast functionality in other parts of the code as well. This is exactly what we did for recordings and RTP forwarders in the VideoRoom, which are now both aware of simulcast, and can handle it if needed.

Please notice that, while for recordings this support is automatic (if a publisher is simulcasting, recording his contribution will automatically try and get the best video available), for RTP forwarders it isn't, and on purpose. In fact, as I explained, we already had some kind of simulcast support there (forwarding the individual streams), which is a feature that we don't want to lose since it can be incredibly helpful in some scenarios. As such, that remains the default behaviour: if you want to RTP forward simulcast streams as a single stream to a remote endpoint, you'll have to add `simulcast: true` to your request, e.g.:

	sfutest.send({
		message: {
			request: "rtp_forward",
			room: 1234,
			publisher_id: myid,
			secret: "adminpwd",
			host: "127.0.0.1",
			simulcast: true,            <--- this parameter here
			video_port: 17004
		}, success: function(result) {
			console.log("Video forwarder:", result["rtp_stream"]["video_stream_id"]);
		}
	});

When it's there, the code will automatically try to forward the best stream available to the recipient, ensuring that RTP headers make it look like a consistent flow just as it happens for WebRTC viewers.

I've tested this while working on the feature and it seems to be working as expected. I'm not aware of any regression, but I didn't look hard. Since I'm planning to merge soon, feedback is welcome as usual!

### DIFF
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -62,6 +62,8 @@ var recordingId = null;
 var selectedRecording = null;
 var selectedRecordingInfo = null;
 
+var doSimulcast = (getQueryStringValue("simulcast") === "yes" || getQueryStringValue("simulcast") === "true");
+
 
 $(document).ready(function() {
 	// Initialize the library (all console debuggers enabled)
@@ -419,7 +421,11 @@ function startRecording() {
 		
 		recordplay.createOffer(
 			{
-				// By default, it's sendrecv for audio and video...
+				// By default, it's sendrecv for audio and video... no datachannels
+				// If you want to test simulcasting (Chrome and Firefox only), then
+				// pass a ?simulcast=true when opening this demo page: it will turn
+				// the following 'simulcast' property to pass to janus.js to true
+				simulcast: doSimulcast,
 				success: function(jsep) {
 					Janus.debug("Got SDP!");
 					Janus.debug(jsep);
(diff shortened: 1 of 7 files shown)


lminiero:
Merging.

Issue #1362, opened 2018-09-13
Videoroom crashes when a publisher leaves and require_pvtid is set to true


tugtugtug:
The video room has the owner subscription list when require_pvtid is set to true. There are a couple of bugs in the current implementation.
1. the owner subscription list isn't refcounted (i.e. can be accessed unsafely).
2. the owner subscription list isn't cleaned up when the publisher leaves and the publisher removes the last reference of the subscribers. i.e. the owner subscription list could be accessed after the subscribers are freed.

I think to properly fix this, we should use refcounted subscription list for the owner (as commented in code). However, it seems to have less value maintaining a subscriber instance if the publisher has already left.

Along with this major issue, I found a few synchronisation issues.
1. private_ids are accessed without protection in a couple of places (should be protected by room mutex). As a side note, I found the mutexes are getting unmanageable in this module and is asking for some review and proper documentation.
2. videoroom_mutex is unlocked twice in the "kick" handler.
3. owner could be accessed without the room lock or holding a reference when a new subscriber is added.
4. owner subscription list is accessed without the subscribers_mutex in janus_videoroom_hangup_media_internal.
5. Possible double free of the subscriber and publisher if the publisher frees the subscriber list first in janus_videoroom_hangup_media_internal, and subscriber races to be freed (subscriber->feed check to have publisher without any synchronisation, and then publisher removes the subscriber).  

I have a patch for addressing the issues above against master/0a5f379a (without doing refcounting for owner subscription).

https://pastebin.com/1F0jD6CS

lminiero:
Thanks for your detailed analysis! The double unlocks are probably some leftovers from changes to the code that we missed, so good catch. I'll have to look at the patch and the possible implications: I'll probably only be able to do that early next week, as tomorrow I'll be out of the office. Can you share your fix as a pull request so that it's easier to review and discuss changes, if needed?

tugtugtug:
No hurry, next week is fine.
One thing I'm still puzzled is how does it make sure the access to subscriber->feed is safe. e.g. if the publisher leaves, and calls janus_videoroom_hangup_media_internal to clean up the subscribers. What makes the g_pointer_clear of the feed synchronise properly with the other subscriber message handlers.

lminiero:
Closing as we merged your PR.

tugtugtug:
Thank you @lminiero !

PR #1363, opened 2018-09-13
Fix videoroom crash when require_pvtid is true (issue #1362).

tugtugtug:
- When a publisher leaves the room, make sure the subscribers are
  also synchronously removed from their owners if any.
- Protected private_ids accesses in a couple of cases.
- Removed redudent unlock of videoroom_mutex in the "kick" handler.
- Added synchronisation of access to the owner when a new subscriber
  is added.
- Added synchronisation when the owner subscription list is accessed
  in janus_videoroom_hangup_media_internal.
- Fixed a possible double free of the subscriber and publisher if
  the publisher frees the subscriber list first in
  janus_videoroom_hangup_media_internal, while the subscriber races to
  be freed in the same function, in which case the subscriber->feed
  is not cleared when checked and then publisher removes the subscriber's
  reference.

Change-Id: I02dc5168ea32d1430b67e9ee374f7f7d2bd9a55d

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1494,6 +1494,12 @@ static void janus_videoroom_publisher_dereference(janus_videoroom_publisher *p) 
 	janus_refcount_decrease(&p->ref);
 }
 
+static void janus_videoroom_publisher_dereference_by_subscriber(janus_videoroom_publisher *p) {
+	/* This is used by g_pointer_clear and g_hash_table_new_full so that NULL is only possible if that was inserted into the hash table. */
+	janus_refcount_decrease(&p->session->ref);
+	janus_refcount_decrease(&p->ref);
+}
+
 static void janus_videoroom_publisher_dereference_nodebug(janus_videoroom_publisher *p) {
 	janus_refcount_decrease_nodebug(&p->ref);
 }
@@ -3620,7 +3626,6 @@ struct janus_plugin_result *janus_videoroom_handle_message(janus_plugin_session 
 		response = json_object();
 		json_object_set_new(response, "videoroom", json_string("success"));
 		/* Done */
-		janus_mutex_unlock(&videoroom->mutex);
 		janus_refcount_decrease(&videoroom->ref);
 		goto plugin_response;
 	} else if(!strcasecmp(request_text, "listparticipants")) {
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! I'll review later today and get back to you :+1:

plugins/janus_videoroom.c
```diff
@@ -4369,6 +4374,34 @@ void janus_videoroom_hangup_media(janus_plugin_session *handle) {
 	janus_mutex_unlock(&sessions_mutex);
 }
 
+static void janus_videoroom_hangup_subscriber(janus_videoroom_subscriber * s) {
+	/* already hangup */
```
lminiero: Just a nit, but can you please make sure all comments start with a capital letter?

plugins/janus_videoroom.c
```diff
@@ -4369,6 +4374,34 @@ void janus_videoroom_hangup_media(janus_plugin_session *handle) {
...
+	if (!s->feed) {
+		return;
+	}
+	/* check if the owner needs to be cleaned up */
+	if(s->pvt_id > 0 && s->room != NULL) {
+		janus_mutex_lock(&s->room->mutex);
+		janus_videoroom_publisher *owner = g_hash_table_lookup(s->room->private_ids, GUINT_TO_POINTER(s->pvt_id));
+		if(owner != NULL) {
+			janus_mutex_lock(&owner->subscribers_mutex);
```
lminiero: Why lock inside the method just for this, when we could lock the whole method when we call it? (see below for related comment). From a quick glance it doesn't look we have anything that would deadlock because of that, even though I haven't checked in detail yet.

plugins/janus_videoroom.c
```diff
@@ -4408,21 +4441,17 @@ static void janus_videoroom_hangup_media_internal(janus_plugin_session *handle)
 		participant->remb_latest = 0;
 		participant->fir_latest = 0;
 		participant->fir_seq = 0;
-		while(participant->subscribers) {
-			janus_videoroom_subscriber *s = (janus_videoroom_subscriber *)participant->subscribers->data;
+		GSList *subscribers = participant->subscribers;
+		participant->subscribers = NULL;
+		janus_mutex_unlock(&participant->subscribers_mutex);
```
lminiero: See note above: this unlock was *after* the while before. Although I guess it's pretty much the same as here we're working on a copy of the subscribers list, and the method itself removes from the *actual* list using the mutex?

plugins/janus_videoroom.c
```diff
@@ -4431,16 +4460,10 @@ static void janus_videoroom_hangup_media_internal(janus_plugin_session *handle)
 		if(subscriber) {
 			subscriber->paused = TRUE;
 			janus_videoroom_publisher *publisher = subscriber->feed;
+			/* It is safe to use feed as the only other place sets feed to NULL
+			   is in this function and accessing to this function is synchronized
+			   by sesssions_mutex */
```
lminiero: Typo (sesssions).

lminiero:
Just tested and it seems to be working fine, so I'll merge and fix the small nits myself (comments, typo). Thanks again for your contribution!

PR #1364, opened 2018-09-14
Fix makedistcheck

tmatth:


lminiero:
Thanks for the fix, merging!

PR #1366, opened 2018-09-14
ice: typo (flag cleared multiple times)

DmitryYudin:


lminiero:
Good catch... merging, thanks!

PR #1367, opened 2018-09-14
nosip: resource leak if no answer received

DmitryYudin:


lminiero:
I'll try to review later today, thanks for your contribution!

plugins/janus_nosip.c
```diff
@@ -583,6 +583,7 @@ char *janus_nosip_sdp_manipulate(janus_nosip_session *session, janus_sdp *sdp, g
 /* Media */
 static int janus_nosip_allocate_local_ports(janus_nosip_session *session, gboolean update);
 static void *janus_nosip_relay_thread(void *data);
+static void janus_nosip_relay_thread_close_internal(janus_nosip_session *session);
```
lminiero: I don't think the name of the method is correct, considering it is not related to the the relay thread, but is more of a cleanup of the media session: maybe something like `janus_nosip_media_cleanup`?

DmitryYudin: done

lminiero:
Looks good, merging then! Thanks again for your contribution: I think I'll try and integrate it in the SIP and SIPre plugins as well, considering they work pretty much the same way on that.

PR #1368, opened 2018-09-14
nosip: better port choice (fixed number of attempts replaced with a full scan within a given range)

DmitryYudin:
Test case:
min_port = 10000
max_port = 60000
Ports from 10000 to 59995 are opened on a first session start and remained open
Port slider set to 10000 (to start discovering from 10000 as a worst case)

It takes 30-40ms to run to port 5996 for Ubuntu laptop with core-i5@2.5GHz.
It takes ~150ms if we close rtp fd on unsuccessful bind as it was originally done.
One can implement look-up table (as an array of bit-masks) to pick up free port for 0ms.

### DIFF
--- a/plugins/janus_nosip.c
+++ b/plugins/janus_nosip.c
@@ -263,8 +263,11 @@ static gboolean notify_events = TRUE;
 static janus_callbacks *gateway = NULL;
 
 static char *local_ip = NULL;
-static uint16_t rtp_range_min = 10000;
-static uint16_t rtp_range_max = 60000;
+#define DEFAULT_RTP_RANGE_MIN 10000
+#define DEFAULT_RTP_RANGE_MAX 60000
+static uint16_t rtp_range_min = DEFAULT_RTP_RANGE_MIN;
+static uint16_t rtp_range_max = DEFAULT_RTP_RANGE_MAX;
+static uint16_t rtp_range_slider = DEFAULT_RTP_RANGE_MIN;
 
 static GThread *handler_thread;
 static void *janus_nosip_handler(void *data);
@@ -658,8 +661,15 @@ int janus_nosip_init(janus_callbacks *callback, const char *config_path) {
 				rtp_range_min = rtp_range_max;
 				rtp_range_max = temp_port;
 			}
+			if(rtp_range_min % 2)
+				rtp_range_min++;	/* Pick an even port for RTP */
+			if(rtp_range_min > rtp_range_max) {
+				JANUS_LOG(LOG_WARN, "Wrong port range: %u -- %u\n", rtp_range_min, rtp_range_max);
+				rtp_range_max = rtp_range_min;
+			}
 			if(rtp_range_max == 0)
 				rtp_range_max = 65535;
+			rtp_range_slider = rtp_range_min;
 			JANUS_LOG(LOG_VERB, "NoSIP RTP/RTCP port range: %u -- %u\n", rtp_range_min, rtp_range_max);
 		}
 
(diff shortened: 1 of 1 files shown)


plugins/janus_nosip.c
```diff
@@ -657,8 +660,15 @@ int janus_nosip_init(janus_callbacks *callback, const char *config_path) {
 				rtp_range_min = rtp_range_max;
 				rtp_range_max = temp_port;
 			}
+			if(rtp_range_min % 2)
+				rtp_range_min++;	/* Pick an even port for RTP */
+			if(rtp_range_min > rtp_range_max) {
+				JANUS_LOG(LOG_VERB, "Wrong port range: %u -- %u\n", rtp_range_min, rtp_range_max);
+				return -1;
```
lminiero: I wouldn't return an error here. I'd swap min and max so that the range is now valid, and turn the `LOG_VERB` to a `LOG_WARN` to clarify the new range we're enforcing.

plugins/janus_nosip.c
```diff
@@ -1760,6 +1770,72 @@ char *janus_nosip_sdp_manipulate(janus_nosip_session *session, janus_sdp *sdp, g
...
+		} else if(janus_nosip_bind_socket(rtcp_fd, rtcp_port)) {
+			close(rtp_fd);
+			rtp_fd = -1;
+			/* rtcp_fd still unbind, reuse it */
+		} else {
+			fds[0] = rtp_fd;
+			fds[1] = rtcp_fd;
+			ports[0] = rtp_port;
+			ports[1] = rtcp_port;
+			rtp_range_slider = rtp_port_next; 					/* Write global slider */
```
lminiero: I still haven't understood what this slider is for, but you're modifying it in a non-thread safe way.

plugins/janus_nosip.c
```diff
@@ -1798,107 +1874,86 @@ static int janus_nosip_allocate_local_ports(janus_nosip_session *session, gboole
...
 		}
+
+		static unsigned session_count = 0;
+		static int fds[DEFAULT_RTP_RANGE_MAX - DEFAULT_RTP_RANGE_MIN + 1] = {0, };
+		if(session_count == 0) {
+			unsigned n_ports_locked = 0;
+			for(uint16_t port = rtp_range_min; port < rtp_range_max - 4; port++) {
+				int fd = socket(AF_INET, SOCK_DGRAM, 0);
+				if(fd == -1) {
+					// https://askubuntu.com/questions/162229/how-do-i-increase-the-open-files-limit-for-a-non-root-user
```
lminiero: Please remove that comment.

plugins/janus_nosip.c
```diff
@@ -1798,107 +1874,86 @@ static int janus_nosip_allocate_local_ports(janus_nosip_session *session, gboole
...
+				if(janus_nosip_bind_socket(fd, port)) {
+					continue;
+				}
+
+				fds[port - DEFAULT_RTP_RANGE_MIN] = fd;
+				n_ports_locked++;
+			}
+			JANUS_LOG(LOG_ERR, "-------------------------------------------\n");
+			JANUS_LOG(LOG_ERR, "number ports locked %u of %u\n", n_ports_locked, rtp_range_max - rtp_range_min);
+			JANUS_LOG(LOG_ERR, "-------------------------------------------\n");
```
lminiero: I assume these log lines are there for your debugging purposes? If not, please remove them.

plugins/janus_nosip.c
```diff
@@ -1798,107 +1874,86 @@ static int janus_nosip_allocate_local_ports(janus_nosip_session *session, gboole
...
+		}
+		session_count++;
 	}
+	uint64_t t_start = janus_get_monotonic_time();
 	/* Start */
-	int attempts = 100;	/* FIXME Don't retry forever */
-	if(session->media.has_audio) {
+	if(session->media.has_audio && 
+		(session->media.local_audio_rtp_port == 0 || session->media.local_audio_rtcp_port == 0)) {
```
lminiero: Add one more tab of indentation here (coding style nit).

plugins/janus_nosip.c
```diff
@@ -1798,107 +1874,86 @@ static int janus_nosip_allocate_local_ports(janus_nosip_session *session, gboole
...
+		if(session->media.audio_rtp_fd != -1) {
+			JANUS_LOG(LOG_WARN, "Audio RTP unbind socket detected, closing ...\n");
+			close(session->media.audio_rtp_fd);
+			session->media.audio_rtp_fd = -1;
+		}
+		if(session->media.audio_rtcp_fd != -1) {
+			JANUS_LOG(LOG_WARN, "Audio RTCP unbind socket detected, closing ...\n");
```
lminiero: What are these two log lines for, and why are they LOG_WARN? This means they'll be VERY verbose, and might confuse users if they're not really warnings.

plugins/janus_nosip.c
```diff
@@ -1798,107 +1874,86 @@ static int janus_nosip_allocate_local_ports(janus_nosip_session *session, gboole
...
-	if(session->media.has_video) {
+	if(session->media.has_video && 
+		(session->media.local_video_rtp_port == 0 || session->media.local_video_rtcp_port == 0)) {
+		if(session->media.video_rtp_fd != -1) {
+			JANUS_LOG(LOG_WARN, "Video RTP unbind socket detected, closing ...\n");
+			close(session->media.video_rtp_fd);
+			session->media.video_rtp_fd = -1;
+		}
+		if(session->media.video_rtcp_fd != -1) {
+			JANUS_LOG(LOG_WARN, "Video RTCP unbind socket detected, closing ...\n");
```
lminiero: See above.

plugins/janus_nosip.c
```diff
@@ -1798,107 +1874,86 @@ static int janus_nosip_allocate_local_ports(janus_nosip_session *session, gboole
...
+		JANUS_LOG(LOG_VERB, "Video RTCP listener bound to port %d\n", ports[1]);
+		session->media.video_rtp_fd = fds[0];
+		session->media.video_rtcp_fd = fds[1];
+		session->media.local_video_rtp_port = ports[0];
+		session->media.local_video_rtcp_port = ports[1];
 	}
+
+	uint64_t t_end = janus_get_monotonic_time();
+	printf("port allocation took %d ms, port_next=%d\n", (uint32_t)(1000*(t_end - t_start)/G_USEC_PER_SEC), rtp_range_slider);
+	fflush(stdout);
```
lminiero: Please don't use printf or fflush directly, but use our logging primitives instead. Anyway, I guess you only added those to measure the times in your tests? If so, please remember to remove them (and the t_start/t_end variables) before we decide to merge.

plugins/janus_nosip.c
DmitryYudin: The second commit is not for merging, just for testing.

plugins/janus_nosip.c
DmitryYudin: left?

lminiero: Nope, to the right (deeper indentation for the second line of the function call).

plugins/janus_nosip.c
DmitryYudin: By design we do not have any sockets unbind. Something went wrong if this happened, but we can work normally.

plugins/janus_nosip.c
DmitryYudin: This is just a helper which points to a last position available. There is no need to synchronize access across different threads since the only drawback is that will have to check a couple more port candidates.

plugins/janus_nosip.c
DmitryYudin: Ok.
This only can happen if min=max=odd_number

lminiero:
Could you rebase without the comments/log lines that you added for debugging, so that we can see what would be merged eventually? As soon as you do that, I'll make a couple of tests and merge if I see no weird behaviour :+1:

lminiero:
I still see a couple of small things to fix, but we're talking about nits (mostly coding style) so something I can do myself. Thanks again for your contribution, merging then!

PR #1369, opened 2018-09-14
ice: add "selected-pair" to "webrtcup" message

DmitryYudin:
Don't you mind if I add connection description to "webrtcup" message?
Should I wrap it into "data" section:
```
"date" {
    "selected-pair": 192.168.... <->
}
```
I'll update docs if you ok.


lminiero:
I'm not sure we need this. You can get the selected pair from the stats on the client side already.

lminiero:
Closing as I don't think this is needed, the WebRTC APIs on the client side have stats that expose that information as well.

PR #1371, opened 2018-09-17
streaming plugin: use even RTP port for RTSP

nu774:
RFC3550 says RTP SHOULD use an even/odd ports pair (even for RTP);
Darwin Streaming Server returns 400 when odd RTP port is given in
SETUP.


lminiero:
Sounds reasonable, thanks for the contribution! Can you fill in the form in our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

lminiero:
Thanks, merging then!

PR #1373, opened 2018-09-18
mqevh: minor memleak

DmitryYudin:


events/janus_rabbitmqevh.c
```diff
@@ -377,7 +365,20 @@ int janus_rabbitmqevh_init(const char *config_path) {
...
+	return 0;
+
+error:
+	/* If we got here, something went wrong */
+	success = FALSE;
+	if(route_key)
+		g_free((char *)route_key);
+	if(exchange)
+		g_free((char *)exchange);
+	goto cleanup_and_return;
```
lminiero: Why put `cleanup_and_return` *before* `error`, and then go backwards? It's probably better to put `error` first and let it flow in `cleanup_and_return` instead; successful runs would have a `goto cleanup_and_return` instead. This way, everything always goes in the same direction, which is more intuitive. This is how we do it in a few other plugins as well, I believe.

lminiero: PS: besides, something like `done` is probably a better name than `cleanup_and_return`, after the reordering.

DmitryYudin: As for reordering, I think the successful branch should have less goto's. I'll do reorder since you have another opinion.

events/janus_rabbitmqevh.c
```diff
@@ -377,7 +374,11 @@ int janus_rabbitmqevh_init(const char *config_path) {
 		g_free((char *)ssl_key_file);
```
lminiero: Are you sure we can free vhost, credentials and certificate stuff in the "success" case? I'm not super familiar with the library, so I don't know if it makes its own internal copy of those strings: I do know we were freeing the host before, so that should be fine. I'll make a couple of tests to see if it works, even though my RabbitMQ server is not configured with credentials or SSL support.

lminiero:
I configured credentials (even though the server doesn't need them) and it looks good, no issue. Merging!

events/janus_rabbitmqevh.c
DmitryYudin: I did not run, but I checked the call sequence.
The pointer falls into openssl and does not resides anywhere within rabbit:
'ssl_key_file' ->  amqp_ssl_socket_set_key() -> SSL_CTX_use_PrivateKey_file()

PR #1375, opened 2018-09-19
rtcp: division by zero

DmitryYudin:
rtcp(160): sanitary, did't catch
other: catched


lminiero:
Good catch (no pun intended!), thanks, merging! :+1:

pcgpcgpcg:
- - 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1376, opened 2018-09-20
Remove redundant inc/dec `dtls->ref`

zane-hpcnt:
dlts->ref is increased at `janus_dtls_srtp_create`,
and is decreased at `janus_dtls_srtp_destroy`

So those lines seem to be not needed.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1315,7 +1315,6 @@ void janus_ice_component_free(const janus_refcount *component_ref) {
 	}
 	if(component->dtls != NULL) {
 		janus_dtls_srtp_destroy(component->dtls);
-		janus_refcount_decrease(&component->dtls->ref);
 		component->dtls = NULL;
 	}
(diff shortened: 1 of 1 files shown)


lminiero:
I'm not sure they were redundant. IIRC we added those after some crashes with DTLS code, especially during the PeerConnection shutdown process.

zane-hpcnt:
What I found by git blame on that is just commit when refcount was first added, would you recall any issue or discussion reported about the crash?

lminiero:
I don't remember if it was discussed here, but I do remember working on a DTLS-related crash with a colleague, that was also related to refcounting. Unless there is a good reason to remove those calls apart from a potential redundancy, I'd leave it as it is.

lminiero:
Closing for now for the reasons said above, I'll keep this in mind though. Thanks again for bringing this to our attention!

PR #1379, opened 2018-09-24
Introduce OpenSSL BIO that directly writes to the libnice agent.

fancycode:
Instead of passing all data through the BIO filter (and thus copying it), the new BIO directly sends all packets using the libnice agent.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -410,8 +410,8 @@ gint janus_dtls_srtp_init(const char *server_pem, const char *server_key, const 
 	JANUS_LOG(LOG_INFO, "Fingerprint of our certificate: %s\n", local_fingerprint);
 	SSL_CTX_set_cipher_list(ssl_ctx, DTLS_CIPHERS);
 
-	if(janus_dtls_bio_filter_init() < 0) {
-		JANUS_LOG(LOG_FATAL, "Error initializing BIO filter\n");
+	if(janus_dtls_bio_agent_init() < 0) {
+		JANUS_LOG(LOG_FATAL, "Error initializing BIO agent\n");
 		return -8;
 	}
 
@@ -441,7 +441,6 @@ static void janus_dtls_srtp_free(const janus_refcount *dtls_ref) {
 	/* BIOs are destroyed by SSL_free */
 	dtls->read_bio = NULL;
 	dtls->write_bio = NULL;
-	dtls->filter_bio = NULL;
 	if(dtls->srtp_valid) {
 		if(dtls->srtp_in) {
 			srtp_dealloc(dtls->srtp_in);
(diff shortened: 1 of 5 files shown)


lminiero:
The BIO is needed, it does DTLS fragmentation. Besides, we now only do libnice operations from the loop thread, something that I'm not sure your patch would ensure?

fancycode:
From what I understand from the filter, it was used to keep track of the number of bytes that were written to it per invocation to make sure the same number of bytes was written to the agent later in `janus_dtls_fd_bridge`. The same is achieved with the new BIO that directly passes any bytes written to it to the agent. Fragmentation is handled through the MTU which has not changed compared to the previous filtering BIO.

Regarding threading: all locations calling OpenSSL APIs that potentially write to the underlying BIO (`SSL_do_handshake`, `SSL_read` (if still performing the handshake), `SSL_write` and `DTLSv1_handle_timeout`) where previously calling `janus_dtls_fd_bridge` (which then sent to the libnice agent). This is now done internally through the new BIO and therefore happens on the same thread as before.

lminiero:
Maybe what I'm remembering is the "black box" BIO, that when accessed would return the whole buffer rather than the individual pieces a custom BIO has access to. I'll have to double check.

fancycode:
> Maybe what I'm remembering is the "black box" BIO, that when accessed would return the whole buffer rather than the individual pieces a custom BIO has access to.

That was probably the memory BIO you were using before (wrapped in the filter). This doesn't keep track of the individual writes but might combine then when reading - your filter made sure to read the individual blocks back. My new BIO will pass any writes directly through to the libnice agent without storing anything inbetween.

lminiero:
> That was probably the memory BIO you were using before (wrapped in the filter).

No, it was actually the BIO *before* we started using the filter. Using a regular BIO would hide the individual writes, and just get us a larger buffer to send. Our filter would instead intercept the cut down messages (according to the MTU) and keep them as separate buffers in the list. Anyway, your agent does the same thing, as it also intercepts the smaller chunks as it should: I verified it yesterday.

I'll do a deeper review and some more tests later (it's a very delicate part of the PeerConnection process and so I want to be extra sure it's fine!), and if I don't see any issue I'll merge :+1:

dtls.c
```diff
@@ -969,54 +951,6 @@ int janus_dtls_verify_callback(int preverify_ok, X509_STORE_CTX *ctx) {
...
-	}
-	int pending = BIO_ctrl_pending(dtls->filter_bio);
-	while(pending > 0) {
-		JANUS_LOG(LOG_HUGE, "[%"SCNu64"] >> Going to send DTLS data: %d bytes\n", handle->handle_id, pending);
-		char outgoing[pending];
-		int out = BIO_read(dtls->write_bio, outgoing, sizeof(outgoing));
-		JANUS_LOG(LOG_HUGE, "[%"SCNu64"] >> >> Read %d bytes from the write_BIO...\n", handle->handle_id, out);
-		if(out > 1500) {
-			/* FIXME Just a warning for now, this will need to be solved with proper fragmentation */
-			JANUS_LOG(LOG_WARN, "[%"SCNu64"] The DTLS stack is trying to send a packet of %d bytes, this may be larger than the MTU and get dropped!\n", handle->handle_id, out);
```
lminiero: I've noticed that you've removed this warning where you now send the packet, could you add it again? Maybe right after the "negative size" check, as that's where we're checking the size already anyway.

lminiero:
Thanks! Reviewed again this morning and it looks good to me. Merging then :+1:

dtls.c
fancycode: The handle is not known yet at the top of the new function, so I re-added the warning just before the call to `nice_agent_send`. Shouldn't the check be against the configured `mtu` anyway? If somebody configures the MTU to be 2000 bytes, the warning should not be shown imho.

Also if you want to avoid sending packets larger than 1500 bytes, it should work to return `min(mtu, 1500)` from the `BIO_CTRL_DGRAM_QUERY_MTU` in `janus_dtls_bio_agent_ctrl`.

lminiero: I don't really want to prevent that: MTU is only a problem when it is. In case Janus is used entirely in a local network it might not be a problem at all. But it's a good point that we should add the warning when setting the MTU in the first place.

PR #1382, opened 2018-09-25
UI tweaks to make that silly iOS thing happy

lminiero:
Pinging @thehunmonkgroup as I've followed his excellent [blog post](https://webrtchacks.com/guide-to-safari-webrtc/) on all the things you need to do to make sure WebRTC works as expected on Safari and iOS, and so a quick review would be very welcome!

As the title explains, these are a few changes to both `janus.js` and the JavaScript of the demos to ensure they work on iOS: we tested it briefly on the iPhone of a colleague and they seem to work fine (for EchoTest and VideoRoom, at least), but more feedback would help, especially considering I tend to stay very far away from Apple products. The main changes we did were:

1. make sure `pagehide` is used instead of `beforeonload` (which we use to destroy the session when the page is closed) on iOS;
2. add `playsinline` to all audio and video elements;
3. only use `setSinkId` if supported (something we needed for the Device Test demo).

This doesn't cover *all* the suggestions Chad made in his post, but it should be enough for what we use in the demo. Just to quickly recap his tips:

1. `beforeunload` --> `pagehide` :heavy_check_mark: 
2. `playsinline :heavy_check_mark:
3. autoplay rules and the need for user gestures --> not a problem for us (we have a Start button on all demo pages) :heavy_check_mark:
4. no low/limited video resolutions --> not a problem for us, I guess (unless you ask for "lowres" in janus.js? not sure what happens when you do that)
5. new getUserMedia() request kills existing stream track --> I guess this breaks our renegotiation code in janus.js for iOS: to be honest, I have no intention of doing anything on that, too much work, they should fix it instead
6. media device IDs change on page reload --> don't care
7. speaker selection not supported --> added check to device test demo :heavy_check_mark:
8. no VP8 support --> we know, Apple doesn't like standards
9. other esoteric bugs --> can't care less

Feedback welcome, as I'll probably merge soon!

### DIFF
--- a/html/audiobridgetest.js
+++ b/html/audiobridgetest.js
@@ -284,7 +284,7 @@ $(document).ready(function() {
 									var addButtons = false;
 									if($('#roomaudio').length === 0) {
 										addButtons = true;
-										$('#mixedaudio').append('<audio class="rounded centered" id="roomaudio" width="100%" height="100%" autoplay playsinline/>');
+										$('#mixedaudio').append('<audio class="rounded centered" id="roomaudio" width="100%" height="100%" autoplay/>');
 									}
 									Janus.attachMediaStream($('#roomaudio').get(0), stream);
 									if(!addButtons)
--- a/html/audiobridgetest.js
+++ b/html/audiobridgetest.js
@@ -284,7 +284,7 @@ $(document).ready(function() {
 									var addButtons = false;
 									if($('#roomaudio').length === 0) {
 										addButtons = true;
-										$('#mixedaudio').append('<audio class="rounded centered" id="roomaudio" width="100%" height="100%" autoplay/>');
+										$('#mixedaudio').append('<audio class="rounded centered" id="roomaudio" width="100%" height="100%" autoplay playsinline/>');
 									}
 									Janus.attachMediaStream($('#roomaudio').get(0), stream);
 									if(!addButtons)
--- a/html/voicemailtest.js
+++ b/html/voicemailtest.js
@@ -144,7 +144,7 @@ $(document).ready(function() {
 													$('#listen').click(function() {
 														$('#rec').remove();
 														$('#done').parent().append(
-															'<audio id="rec" style="width:100%;height:100%;" autoplay playsinline controls preload="auto">' +
+															'<audio id="rec" style="width:100%;height:100%;" autoplay controls preload="auto">' +
 																'<source id="opusrec" src="' + msg["recording"] + '" type="audio/ogg""></source>' +
 																'Your browser doesn\'t support the playout of Opus files' +
 															'</audio>'
--- a/html/voicemailtest.js
+++ b/html/voicemailtest.js
@@ -144,7 +144,7 @@ $(document).ready(function() {
 													$('#listen').click(function() {
 														$('#rec').remove();
 														$('#done').parent().append(
-															'<audio id="rec" style="width:100%;height:100%;" autoplay controls preload="auto">' +
+															'<audio id="rec" style="width:100%;height:100%;" autoplay playsinline controls preload="auto">' +
 																'<source id="opusrec" src="' + msg["recording"] + '" type="audio/ogg""></source>' +
 																'Your browser doesn\'t support the playout of Opus files' +
 															'</audio>'
--- a/html/devicetest.js
+++ b/html/devicetest.js
@@ -83,9 +83,10 @@ function initDevices(devices) {
 			$('#audio-device').append(option);
 		} else if(device.kind === 'videoinput') {
 			$('#video-device').append(option);
-		} else if(device.kind === 'audiooutput') {
+		} else if(device.kind === 'audiooutput' && setSinkId in HTMLMediaElement.prototype) {
 			// Apparently only available from Chrome 49 on?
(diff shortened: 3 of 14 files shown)


html/audiobridgetest.js
```diff
@@ -284,7 +284,7 @@ $(document).ready(function() {
 									var addButtons = false;
 									if($('#roomaudio').length === 0) {
 										addButtons = true;
-										$('#mixedaudio').append('<audio class="rounded centered" id="roomaudio" width="100%" height="100%" autoplay/>');
+										$('#mixedaudio').append('<audio class="rounded centered" id="roomaudio" width="100%" height="100%" autoplay playsinline/>');
```
thehunmonkgroup: Is this an audio-only tag? If so, I don't think playsinline is needed, it's only for video elements.

html/voicemailtest.js
```diff
@@ -144,7 +144,7 @@ $(document).ready(function() {
 													$('#listen').click(function() {
 														$('#rec').remove();
 														$('#done').parent().append(
-															'<audio id="rec" style="width:100%;height:100%;" autoplay controls preload="auto">' +
+															'<audio id="rec" style="width:100%;height:100%;" autoplay playsinline controls preload="auto">' +
```
thehunmonkgroup: Same as previous comment, playsinline not needed here.

lminiero:
> Having a user gesture alone does not solve autoplay issues -- in this case you would need to call the .play() method on the element in question, which would be allowed if it's invoked in response to a user gesture

Good point, and Chrome (in my M71 Canary) has started implementing the same stupid logic as well: https://developers.google.com/web/updates/2017/09/autoplay-policy-changes

> However, if you're always asking for mic/camera permissions before playing audio/video, this would also satisfy the autoplay rules.

That's an even stupider choice by browser implementors, as it cuts out datachannel-only and recv-only (e.g., webinar) applications. Unfortunately they use the same "permission" for host candidates as well (which they're trying to patch with mDNS candidates).

> I haven't looked at that code, but do you even need the old stream after the renegotiation?

In case we're switching devices we close it before accessing the new one so it shouldn't be an issue, you're right. It is, though, if you choose the same device, isn't it? Or if you're sending the same device on two different PeerConnections? Not sure what the bug is in Safari there.

html/audiobridgetest.js
lminiero: Ack, thx for the clarification! Just removed it.

thehunmonkgroup:
>  It is, though, if you choose the same device, isn't it? Or if you're sending the same device on two different PeerConnections? 

The bug is at the gUM() level, not the peer connection level. If you are no longer using the stream from the first gUM(), then just kill all the tracks on that stream and use the stream from the new gUM() request, and you *should* be fine (unless there's some other dragon there I haven't hit).

> That's an even stupider choice by browser implementors, as it cuts out datachannel-only and recv-only (e.g., webinar) applications.

They're battling the autoplay BS from advertisers, who of course could use WebRTC to get around the current audio/video element restrictions if the WebRTC permissions were too lax. I don't think it's a stupid choice as much as a hard choice. Also, it doesn't *exclude* the datachannel/recv-only cases, it forces another step of user interaction before permitting, which is probably sensible at least on mobile. My god, I'm defending browser vendors, what's happening?? ;)

lminiero:
> The bug is at the gUM() level, not the peer connection level. If you are no longer using the stream from the first gUM(), then just kill all the tracks on that stream and use the stream from the new gUM() request, and you _should_ be fine (unless there's some other dragon there I haven't hit).
> 

We may have a problem when only changing a single device, e.g., to replace the video track while keeping audio as it is. In that case, the first thing we do is removing the old video track from the local stream, then do the gUM, and finally add the video track to the local stream. I believe the gUM, though, asks you for both audio and video anyway, which I suspect would cause the issue you mentioned.

> My god, I'm defending browser vendors, what's happening?? ;)

We're getting older! :smile:

thehunmonkgroup:
If I recall correctly from my testing, a video only gUM() request will not affect audio on a previous gUM() request with audio tracks, it will only affect the video track. So I'm guessing in that case you could pretty easily fix the issue by only requesting video in the 2nd gUM(), then swapping out the video track in the 1st gUM().

lminiero:
It's the "pretty easily" part we may disagree on :smile: 
I'll have a look to see how hard it would be to change that code.

lminiero:
These two commits should fix that, by internally using a couple of new properties called `keepAudio` and `keepVideo`. Made a few tests with Chrome to check what the gUM would do and it seems to do the trick, but I don't have access to an iOS device to see if it really works. Besides, I don't have a demo ready either, as I triggered the renegotiations using the JS console directly, e.g.:

```
echotest.createOffer(
	{
		// Keep video, do a new gUM for audio
		media: { replaceAudio: true, video: true },
		success: function(jsep) {
			Janus.debug(jsep);
			echotest.send({message: {audio: true, video: true}, "jsep": jsep});
		},
		error: function(error) {
		}
	});
```

I guess this pretty much concludes my work here, pending some more feedback. Thanks again @thehunmonkgroup for these tips!

lminiero:
Ack, merging then! In case it turns out this isn't enough we can work on that later.

PR #1384, opened 2018-09-26
Fix deadlock in duktape plugin when calling pushevent from coroutines(async)

gabriel-munteanu:
How deadlock is happening:

1. janus_duktape.c:2282 janus_duktape_scheduler -> 2292 janus_mutex_lock(&duktape_mutex);	duk_get_global_string(duktape_ctx, "resumeScheduler");  call pushevent() from js
2. janus_duktape.c:486 janus_duktape_method_pushevent -> 570 janus_core->push_event
3. janus.c:2689 janus_plugin_push_event -> 2743 plugin->get_package()
4. janus_duktape.c:1675 -> 1679 janus_mutex_lock(&duktape_mutex);

With this fix I also fixed the type allowed for the 4th parameter of pushevent. It can be null and is not string :).

### DIFF
--- a/plugins/janus_duktape.c
+++ b/plugins/janus_duktape.c
@@ -371,7 +371,8 @@ static void *janus_duktape_async_event_helper(void *data) {
 		janus_core->push_event(asev->session->handle, &janus_duktape_plugin, asev->transaction, asev->event, asev->jsep);
 	}
 	json_decref(asev->event);
-	json_decref(asev->jsep);
+	if(asev->jsep)
+		json_decref(asev->jsep);
 	g_free(asev->transaction);
 	janus_refcount_decrease(&asev->session->ref);
 	g_free(asev);
@@ -500,7 +501,7 @@ static duk_ret_t janus_duktape_method_pushevent(duk_context *ctx) {
 			janus_duktape_type_string(DUK_TYPE_STRING), janus_duktape_type_string(duk_get_type(ctx, 2)));
 		return duk_throw(ctx);
 	}
-	if(duk_get_type(ctx, 3) != DUK_TYPE_STRING && duk_get_type(ctx, 3) != DUK_TYPE_UNDEFINED) {
+	if(duk_get_type(ctx, 3) != DUK_TYPE_STRING && duk_get_type(ctx, 3) != DUK_TYPE_UNDEFINED && duk_get_type(ctx, 3) != DUK_TYPE_NULL) {
 		duk_push_error_object(ctx, DUK_RET_TYPE_ERROR, "Invalid argument (expected %s, got %s)\n",
 			janus_duktape_type_string(DUK_TYPE_STRING), janus_duktape_type_string(duk_get_type(ctx, 3)));
 		return duk_throw(ctx);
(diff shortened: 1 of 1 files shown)


gabriel-munteanu:
I translated for myself the videoroom plugin from Lua to JS and when I started to test I saw that when I was joining the room nothing is happening and after that the duktape was frozen. I started Janus with gdb, set a breakpoint janus_plugin_push_event and saw that when entering  plugin->get_package(),  I couldn't step after janus_mutex_lock(&duktape_mutex); .

Another solution to this problem is to call js function getPackage after the js plugin is loaded and store that string to be used after directly from C code.
I don't see any case where getPackage would need to return different results.

lminiero:
> 3\. janus.c:2689 janus_plugin_push_event -> 2743 plugin->get_package()

Ok, I think here lies the problem. I'm pretty sure this is something I can replicate with the Lua plugin as well (I'll try in a minute). The `get_name`, `get_package`, etc., should only be invoked once (typically at startup), and not every time they're called, as those values are supposed to be immutable. Once we call them we can store the value and return that, which would fix the issue as there would be no mutex used anymore. I'll make some tests and keep you posted.

lminiero:
@gabriel-munteanu just pushed the fix I was mentioning. Please let me know if it solves the deadlock you were receiving without the need for your patch (I verified it fixes the issue in the Lua plugin).

gabriel-munteanu:
Yes, that commit solves that deadlock. 

I will come back later with another pull request for parameter restriction in janus_duktape_method_pushevent (2nd and 4th can be null an that is a different type than string).

lminiero:
> Yes, that commit solves that deadlock.

Thanks for confirming!

> I will come back later with another pull request for parameter restriction in janus_duktape_method_pushevent (2nd and 4th can be null an that is a different type than string).

Any plan on when you'll submit? I was planning to tag a new version of Janus today (0.4.4), but I can wait for this other fix if you think it's important.

lminiero:
Closing this PR in the meanwhile.

gabriel-munteanu:
It is important but I'm a little bit busy with wedding preparations :D.
Please make you the following fixes in janus_duktape.c
L493: add this to condition && duk_get_type(ctx, 1) != DUK_TYPE_NULL
L503: add this to condition && duk_get_type(ctx, 3) != DUK_TYPE_NULL

You need this this because otherwise you will get: Duktape error: Error: Invalid argument (expected DUK_TYPE_STRING, got DUK_TYPE_NULL) .
duk_get_string() will know how to handle a null value.

You can check in the videoroom.lua that sometimes you call pushEvent with some parameters set to nil.

lminiero:
> It is important but I'm a little bit busy with wedding preparations :D.

Wow, congratulations! Then what are you still doing here?! Get out! :smile: 

> Please make you the following fixes in janus_duktape.c
> L493: add this to condition && duk_get_type(ctx, 1) != DUK_TYPE_NULL
> L503: add this to condition && duk_get_type(ctx, 3) != DUK_TYPE_NULL

Thx, will do :+1: I'll refer to the changes you already made in this patch.

lminiero:
Just applied your fix, thanks again!

PR #1385, opened 2018-09-27
Generate last chunk of transport wide cc fb msg correctly

kangsik:
* The current implementation generates malformed last packet chunk
* While statements check whether each queue is empty after the last status is popped, resulting in 1 less byte chunk generated.

### DIFF
--- a/rtcp.c
+++ b/rtcp.c
@@ -1413,14 +1413,13 @@ int janus_rtcp_transport_wide_cc_feedback(char *packet, size_t size, guint32 ssr
 			/* Write chunk */
 			word = janus_push_bits(word, 1, 1);
 			word = janus_push_bits(word, 1, 1);
-			/* Get fist status */
-			janus_rtp_packet_status status = (janus_rtp_packet_status) GPOINTER_TO_UINT(g_queue_pop_head (statuses));
-			/* Write rest */
-			while(!g_queue_is_empty(statuses)) {
+			/* Write rest statuses  */
+			unsigned int i = 0;
+			for (i=0;i<statuses_len;i++) {
+				/* Get each status */
+				janus_rtp_packet_status status = (janus_rtp_packet_status) GPOINTER_TO_UINT(g_queue_pop_head (statuses));
 				/* Write */
 				word = janus_push_bits(word, 2, (guint8)status);
-				/* Next */
-				status = (janus_rtp_packet_status) GPOINTER_TO_UINT(g_queue_pop_head (statuses));
 			}
 			/* Write pending */
 			word = janus_push_bits(word, 14-statuses_len*2, 0);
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for the contribution!

Ok, I think I understand the issue: right now, if the only element in the queue is the one we already popped before the while, the pop succeeds, but we never do the `janus_push_bits` because we don't get in the while.

But I don't understand why you're switching from a while to a for loop. Can't you solve that issue simply removing that first pop before the while, and in the while switch the `janus_push_bits` and pop there are now so that we iterate a "pop and add" until the queue is empty?

lminiero:
PS: could you sign and fill the form in our [CLA](https://cla.conf.meetecho.com/janus/) so that we can merge this when ready?

kangsik:
1. I signed the CLA!
2. I thought the for loop is more concise and straightforward  :)

rtcp.c
```diff
@@ -1413,14 +1413,13 @@ int janus_rtcp_transport_wide_cc_feedback(char *packet, size_t size, guint32 ssr
 			/* Write chunk */
 			word = janus_push_bits(word, 1, 1);
 			word = janus_push_bits(word, 1, 1);
-			/* Get fist status */
-			janus_rtp_packet_status status = (janus_rtp_packet_status) GPOINTER_TO_UINT(g_queue_pop_head (statuses));
-			/* Write rest */
-			while(!g_queue_is_empty(statuses)) {
+			/* Write rest statuses  */
```
lminiero: Looks like there's an extra space at the end of the comment: can you change the text to "Write all the statuses" too?

rtcp.c
```diff
@@ -1432,14 +1431,13 @@ int janus_rtcp_transport_wide_cc_feedback(char *packet, size_t size, guint32 ssr
 			/* Write chunck */
 			word = janus_push_bits(word, 1, 1);
 			word = janus_push_bits(word, 1, 0);
-			/* Get fist status */
-			janus_rtp_packet_status status = (janus_rtp_packet_status) GPOINTER_TO_UINT(g_queue_pop_head (statuses));
-			/* Write rest */
-			while(!g_queue_is_empty(statuses)) {
+			/* Write rest statuses  */
```
lminiero: Same here.

lminiero:
As a side note, what drove you to work on the fix? Did you notice that the estimated bandwidth was not growing as it should, or did it come from a static analysis of the code?

kangsik:
We enabled transport-wide-cc extension to let the publishers get more exact bandwidth estimation. 
Without transport-wide-cc, publishers get significantly higher  bandwidth estimations than realistic values. The estimated bandwidth values also fluctuated much less after enabling transport-wide-cc.

But.. enabling transport-wide-cc crashed our janus instances several times. So I came up with this  PR and #1386  :)

rtcp.c
```diff
@@ -1432,14 +1431,13 @@ int janus_rtcp_transport_wide_cc_feedback(char *packet, size_t size, guint32 ssr
...
 			word = janus_push_bits(word, 1, 0);
-			/* Get fist status */
-			janus_rtp_packet_status status = (janus_rtp_packet_status) GPOINTER_TO_UINT(g_queue_pop_head (statuses));
-			/* Write rest */
-			while(!g_queue_is_empty(statuses)) {
+			/* Write all the statuses */
+			unsigned int i = 0;
+			for (i=0;i<statuses_len;i++) {
+				/* Get each status */
+			        janus_rtp_packet_status status = (janus_rtp_packet_status) GPOINTER_TO_UINT(g_queue_pop_head (statuses));
```
lminiero: Looks like this line has broken indentation (spaces instead of tabs).

lminiero:
I did a bit of testing for both patches and they look fine, great work! As soon as you can fix that couple of nits I mentioned, I'll merge :+1:

PR #1386, opened 2018-09-27
Limit packet counts per single transport wide cc FB message

kangsik:
* This prevents potential stack corruption made by janus_rtcp_transport_wide_cc_feedback()
(janus_rtcp_transport_wide_cc_feedback() doesn't handle packet buffer overflow)
* A packet can take up to 2.125 byte(16+1 bit) in the transport wide cc FB message
* Considering buffer size and the MTU, 400 seems to be good limit for packet counts per single FB  message

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3404,13 +3404,40 @@ static gboolean janus_ice_outgoing_rtcp_handle(gpointer user_data) {
 		/* Free and reset stats list */
 		g_slist_free(stream->transport_wide_received_seq_nums);
 		stream->transport_wide_received_seq_nums = NULL;
-		/* Get feedback packet count and increase it for next one */
-		guint8 feedback_packet_count = stream->transport_wide_cc_feedback_count++;
-		/* Create rtcp packet */
-		int len = janus_rtcp_transport_wide_cc_feedback(rtcpbuf, size,
-			stream->video_ssrc, stream->video_ssrc_peer[0], feedback_packet_count, packets);
-		/* Enqueue it, we'll send it later */
-		janus_ice_relay_rtcp_internal(handle, 1, rtcpbuf, len, FALSE);
+
+		guint packets_len = 0;
+		while ((packets_len = g_queue_get_length(packets)) > 0) {
+			GQueue *packets_to_process;
+			// if packets_len > 400, transport_wide_cc packet could be overflow the buffer
+			if (packets_len > 400) {
+				// split the queue into two
+				GList *new_head = g_queue_peek_nth_link(packets, 400);
+				GList *new_tail = new_head->prev;
+				new_head->prev = NULL;
+				new_tail->next = NULL;
+				packets_to_process = g_queue_new();
+				packets_to_process->head = packets->head;
+				packets_to_process->tail = new_tail;
+				packets_to_process->length = 400;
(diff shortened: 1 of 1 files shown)


ice.c
```diff
@@ -3404,13 +3404,40 @@ static gboolean janus_ice_outgoing_rtcp_handle(gpointer user_data) {
...
-		/* Create rtcp packet */
-		int len = janus_rtcp_transport_wide_cc_feedback(rtcpbuf, size,
-			stream->video_ssrc, stream->video_ssrc_peer[0], feedback_packet_count, packets);
-		/* Enqueue it, we'll send it later */
-		janus_ice_relay_rtcp_internal(handle, 1, rtcpbuf, len, FALSE);
+
+		guint packets_len = 0;
+		while ((packets_len = g_queue_get_length(packets)) > 0) {
+			GQueue *packets_to_process;
+			// if packets_len > 400, transport_wide_cc packet could be overflow the buffer
```
lminiero: Please use `/*  */` for comments and not `//` (here and in a few of the following lines).

ice.c
```diff
@@ -3404,13 +3404,40 @@ static gboolean janus_ice_outgoing_rtcp_handle(gpointer user_data) {
...
+				packets->length = packets_len - 400;
+			} else {
+				packets_to_process = packets;
+			}
+
+			/* Get feedback packet count and increase it for next one */
+			guint8 feedback_packet_count = stream->transport_wide_cc_feedback_count++;
+			/* Create rtcp packet */
+			int len = janus_rtcp_transport_wide_cc_feedback(rtcpbuf, size,
+					stream->video_ssrc, stream->video_ssrc_peer[0], feedback_packet_count, packets_to_process);
```
lminiero: Just an indentation nit: there should just be a single tab of difference here, not two.

lminiero:
I'm wondering if, rather than splitting packets, we should send feedback more frequently instead. At the moment, we send this feedback once per second: when, if you pointed out, we need to split, we send 2 or more packets together, while if we made that feedback more frequent (500ms? 300ms?) we'd reduce the risk of overflowing, and time the same feedback in a less bursty way. Probably the best approach is doing both: more frequent feedback, and in case we still have to split, then so be it.

What is your feeling on this? Any thoughts on what you think would be a reasonable tradeoff for sending? (not sure how frequent Chrome sends its own). For now I'll merge this as it is, as it's still helpful and an important fix. Later on we can think about this further improvement.

Thanks again, merging!

kangsik:
@lminiero Actually, our servers already have modification which sends feedbacks every 200 ms.  
I thought that modification would be enough to prevent stack corruption, but it turned out that sometimes we receive more than 1000 packets in 200 ms.. and the janus crashes. 

ps. https://tools.ietf.org/html/draft-holmer-rmcat-transport-wide-cc-extensions-01 say we should send feedback per every video frame.

lminiero:
> Actually, our servers already have modification which sends feedbacks every 200 ms.

Got it, I'll take note of that then.

> ps. https://tools.ietf.org/html/draft-holmer-rmcat-transport-wide-cc-extensions-01 say we should send feedback per every video frame.

Never going to happen :wink:

PR #1397, opened 2018-10-05
[FEEDBACK] Redefine mutexes to use GMutex instead of pthread_mutex_t

lminiero:
This is more of a request for feedback than an actual pull request, as the code change is really marginal: actual question towards the end, after a brief technical explanation of what I changed.

I've been wondering if there was anything we could do to make locks in principle more efficient, e.g., toying with the idea of doing our own thing based on atomic operations rather than using mutexes directly. While I've discarded that idea (apparently busy waits are useful in some cases and quite harmful in others), I still wanted to better understand if there were options, and found out that, on Linux, [futexes](https://en.wikipedia.org/wiki/Futex) are supposed to be much more efficient as a native solution.

Rather than rewriting the mutex code myself to use those (especially considering we'd need to still support other systems Janus can be built on which might not support those), I checked how Glib implements mutexes, and found out that [they do use futex](https://bugzilla.gnome.org/show_bug.cgi?id=731986), if available on the system, and fallback to pthread mutexes otherwise. As such, I simply redefined our `janus_mutex` and `janus_condition` to use the [GMutex](developer.gnome.org/glib/stable/glib-Threads.html) methods instead. They were pretty much a 1-1 replacement, apart for a missing `PTHREAD_MUTEX_INITIALIZER` static initialization, that I set to `{ 0 }`, which should be fine as according to the docs:

    Its placement in static storage ensures that it will be initialised to all-zeros, which is appropriate.

My request for feedback comes from a couple of major doubts. From [what I've read](https://stackoverflow.com/questions/6364314/why-is-a-pthread-mutex-considered-slower-than-a-futex), on Linux the pthread mutex implementation also makes use of futex where available, only involving system calls when there are contentions (which should be pretty much what GMutex does in its "native" implementation): as such, is it really worth to switch just for the apparent overhead pthread has? I've personally not done any comparative testing of the two different approaches yet: is there any info or personal experience you can share on whether or not you think there are actual benefits on Linux, or if it's all just theoretical?

Thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -76,6 +76,12 @@ AC_ARG_ENABLE([dtls-settimeout],
               [],
               [enable_dtls_settimeout=no])
 
+AC_ARG_ENABLE([pthread-mutex],
+              [AS_HELP_STRING([--enable-pthread-mutex],
+                              [Use pthread_mutex instead of GMutex (see #1397)])],
+              [],
+              [enable_pthread_mutex=no])
+
 AC_ARG_ENABLE([turn-rest-api],
               [AS_HELP_STRING([--disable-turn-rest-api],
                               [Disable TURN REST API client (via libcurl)])],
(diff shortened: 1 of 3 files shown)


atoppi:
Tested Janus `master` on 
```
Ubuntu 18.04 x64
GNU C library libc6 2.27
gcc 7.3.0
glib 2.56
```
Scenario
```
1 Echotest session
2 Streaming sessions
2 Videoroom sessions
some polls to admin api
```
`strace` only showed `futex` calls.

In addition to this, recent `gperf` benchmarks for **streaming plugin** did not show evidence of high lock contention.

lminiero:
Pthreads still have a lot overhead, though, that with this approach we should at least be able to overcome. Anyway, the right approach is probably to try and reduce the need for locks in the first place.

Since I think this might be useful nevertheless, I'll make it optional: a new configure flag that allows you to choose if you want to use pthread mutexes (the default) or GMutex instead. This should allow people (and us) to experiment and test by themselves on whatever master is.

lminiero:
This commit adds a new `configure` flag, `--enable-pthread-mutex`, to force the usage of pthread mutexes (what we've been using so far) instead of GMutex, which is now the default. Notice that if you change a setting and you want to recompile, you'll have to either touch `mutex.h`, or do a `make clean`, as the configure script doesn't do it by itself.

lminiero:
Merging as this is now configurable.

PR #1398, opened 2018-10-08
add endian define/include to pp-rtp.h

Piasy:
without that definition, janus-pp-rec will use big endian in macOS, which will fail to convert h264 video into mp4.


lminiero:
Good catch, not sure why it wasn't there since it is in the core `rtp.h`. Merging, thanks!

PR #1399, opened 2018-10-09
Refactored handle loop (and thread) as persistent

lminiero:
This patch refactors some internal core handle mechanism, specifically with respect to the event loop.

As you may know, we recently unified incoming and outgoing media traffic over the same event loop, run by a dedicated thread: incoming traffic was already handled via a loop (thanks to libnice), and we simply added outgoing traffic as a custom source to reuse the same loop, thus cutting the number of threads in half. In fact, this allowed us to have a single thread per PeerConnection, rather than two as before. You can refer to #1231 for more information on that effort.

Some race conditions and issues we've debugged lately have convinced us to move one step further. Specifically, at the moment Janus plugins can have their callbacks, for the same plugin session, invoked by different threads: in fact, while incoming media (RTP/RTCP/data) will always be triggered by the loop thread, other callbacks (e.g., `create_session`, `hangup_media`, others) often are not. This can cause issues when, for instance, the core is notifying a plugin that a PeerConnection has been torn down, but traffic is still flowing for a little while via another thread.

This patch fixes this problem by extending what the existing loop has allowed so far, and by making it handle-centric rather than PeerConnection-centric. Specifically, the loop and the thread are now started right away, as soon as the handle is attached, and not only when a PeerConnection is created as before: this way, we can "abuse" the event loop for some more handle-related events. Besides, the loop and its thread are now persistent, meaning they're only torn down when the related handle is detached, and again not when the PeerConnection is closed which is what has happened so far. This has the small disadvantage that now a thread exists for a handle even when no (or before any) PeerConnection is up an running, but it shouldn't be much of a problem: we're still talking of a single thread per handle, and most of them will always have a PeerConnection anyway.

This allowed us to channel a couple of relevant events, namely `hangup_media` and `destroy_session`, via the event loop, meaning that now both those callbacks will be received, for a specific plugin session, from the same thread responsible for media, thus preventing ugly race conditions that might affect the media delivery. We thought about other callbacks too, but came up to the conclusion they're not needed: in fact, `create_session` precedes any media-related consideration, and `setup_media` is triggered by the loop already (it's invoked as a consequence of a selected-pair notification from libnice).

We tested this briefly and plan to do some more checks to assess its stability and performance. Considering we plan to merge this soon, and that it is an important change, I definitely encourage you to give this a try yourselves as well, just to be double sure we're not breaking anything for you.


fancycode:
I like the idea. However as a result, plugins must not block while processing the events to avoid blocking the media processing running on the same thread, right?

atoppi:
@fancycode The only callbacks blocking the media loop are `hangup` and `detach`. Your plugin should not do any other blocking call during the execution. However `hangup` and `detach` are both meant to stop a Peer Connection, so even a bad programmed callback might be a false problem.

lminiero:
Yep, as @atoppi only `hangup_media` and `destroy_session` are channeled through this loop, and they both imply media is gone. As such, there's no media processing to worry about.

peffis:
A use-after-free error was spotted on this branch in audiobridge plugin. 

Libasan log: https://pastebin.com/E45tanQh
What happens is that the mixer thread is still accessing the data under a participant @ https://github.com/meetecho/janus-gateway/blob/handle-loop/plugins/janus_audiobridge.c#L4328 when the participant struct had already been freed in https://github.com/meetecho/janus-gateway/blob/handle-loop/plugins/janus_audiobridge.c#L945

As you know I have spotted similar issues also on other branches before so it is not clear if this can be put to blame on this PR. The PR clears a few similar errors up though so perhaps the problem above can be looked into within the scope of the PR or as a new issue.

atoppi:
@peffis definitely not strictly related to this PR, but something that needs to be fixed though.

lminiero:
Yep, I agree that this is a separate issue, and one that would need to be solved with a better usage of reference counters in that plugin.

lminiero:
@peffis I just merged this fix I applied to master, which should solve your problem: https://github.com/meetecho/janus-gateway/commit/ad94a0905acdb58caa9b8fcd185bea0fd6504873

lminiero:
We did some testing this week and it seems to indeed fix many of the issues we were experiencing, so I'll merge.

PR #1400, opened 2018-10-09
Print video rotation

Piasy:


lminiero:
Extensions don't have a fixed ID, they're negotiated. The fact that Chrome hardcodes some values, doesn't mean others will use the same as well. As such, I'm afraid I can't merge this.

Piasy:
Could we add an option to specify CVO ID?

Piasy:
Okey, I'll add an option for that, and move it out from h264 file :)

postprocessing/janus-pp-rec.c
```diff
@@ -131,6 +132,12 @@ int main(int argc, char *argv[])
 			ignore_first_packets = val;
 		JANUS_LOG(LOG_INFO, "Ignoring first packets: %d\n", ignore_first_packets);
 	}
+	if(g_getenv("JANUS_RTP_HEADER_EXT_ID_CVO") != NULL) {
```
lminiero: The convention we followed for other env variables, is to use the prefix `JANUS_PPREC_`. As such, I'd rename it to something like `JANUS_PPREC_EXT_CVO` or `JANUS_PPREC_EXTID_CVO`, which might come in handy if we decide to parse other extensions in the future.

postprocessing/janus-pp-rec.c
```diff
@@ -665,6 +672,55 @@ int main(int argc, char *argv[])
 	JANUS_LOG(LOG_INFO, "Counted %"SCNu32" frame packets\n", count);
 
 	if(video) {
+		if (rtp_header_ext_id_cvo != -1) {
+			/* Parse RTP header extension now */
+			int rotation = 0, got_rotation = 0;
+			janus_pp_frame_packet *tmp = list;
+			while(tmp) {
```
lminiero: Why are you parsing in a separate while? In case there are many packets, that might take a long time, which is unneeded if we consider we already go through all the packets in other parts of the code. My suggestion would be making the CVO extraction a separate function with signature in `pp-rtp.h` and implementation in `janus-pp-rec.c`, and then invoke that function from the webm and mp4 code: this would allow to avoid repetitions (or only having it for mp4, as it was initially), and integrate it with the approach we have in use for keyframes for instance (printing when we receive a keyframe while traversing the document). Does this make sense?

Piasy:
@lminiero Updated.

After we point `rtp` into `prebuffer`, we shouldn't change it's content if we still need access to `rtp`, otherwise, its field will be broken.

So I use another `prebuffer2`.

lminiero:
@Piasy but the extension is still parsed where it was before, not when going through the video... this makes the method signature in the header useless. It should be in the `*_preprocess` method of the individual video processors. Besides, your RTP extension code seems to assume it will always be the first extension in the list, which might not be the case: the code we have to look for extensions in Janus is better suited for this.

Rather than ask for another change, I'll try to prepare a small branch that looks for the CVO using the Janus-based code, and integrate it in webm and mp4 processors, so that you can test it and check if it works for your case. I'll keep you posted.

lminiero:
@Piasy as a side note, do you have an .mjr with video orientation extension in it I can use for testing?

Piasy:
@lminiero I didn't understand you want to move the `janus_pp_parse_rtp_header_ext_cvo` calls to `*_preprocess` functions.

My code parse all RTH extensions, [there is a while clause](https://github.com/meetecho/janus-gateway/pull/1400/files#diff-8b5cd0be1437a5068ae1394e6b551a72R105).

I have such mjr files, I'll attach it here.

[Archive.zip](https://github.com/meetecho/janus-gateway/files/2480894/Archive.zip)

It contains 3 mjr recorded from native Android WebRTC client, native iOS WebRTC client, and browser Android client, which has 270, 90, 0 clockwise rotation.

Piasy:
@lminiero But why can't we parse RTP header extension at `janus-pp-rec.c`? My revised code doesn't need a separate loop, and we also don't need add code twice for h264/webm.

lminiero:
@Piasy it's ok to parse there, but it should be available to the individual processors. For instance, this is what a `--parse` flag on one of your recordings returns in my patch:

```
This is a video recording:
  -- Codec:   h264
  -- Created: 1539047403915266
  -- Written: 1539047404340191
SSRC detected: 484166380
Counted 716 RTP packets
Counted 716 frame packets
The video is rotated
Video rotation: 270 degrees
  -- 640x480 (fps [2970,3060] ~ 30)
Parsing and reordering completed, bye!
```

In the main code, I just count how many times the video rotated: in this case, there was just a single rotation (270), so even if the extension was there more than once, I summarize and say "The video is rotated" to clarify it's not a 0 degrees video. I take note of this rotation, and print it again in the h264 portion of the code.

As an example of other extensions, I also added support for the audio level extension, which I integrated in the audio processors. In both cases, there's not much we can do with the values we extract, but they're useful to print anyway.

I'll cleanup the code and publish it as a PR shortly, so that you can check if it works fine for your use case.

lminiero:
I just published the code: looking forward to your feedback!

lminiero:
Closing as I merged the other PR :+1:

PR #1402, opened 2018-10-10
store transport seq num before dropping packets

garry81:
when we drop duplicate packets, its transport seq num is not stored.
so, these packets are reported as missing. we should store transport
seq number before dropping packets.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -2137,6 +2137,34 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 					buf += 2;
 					header = (janus_rtp_header *)buf;
 				}
+				/* Check if we need to handle transport wide cc */
+				if(stream->do_transport_wide_cc) {
+					guint16 transport_seq_num;
+					/* Get transport wide seq num */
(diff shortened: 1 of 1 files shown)


lminiero:
But if it's duplicate, didn't we store that sequence number already? It looks like this change is adding the same sequence number to the list more than once for duplicate packets. Am I missing something?

garry81:
even though rtp data is duplicated due to multiple nack, packet has new transport seq number for each retransmission packet.

lminiero:
Ah, got it now, makes sense, thanks! I'll do a quick review and merge if it's fine :+1:

lminiero:
As a side note, does this only apply to rtx via RFC4588, or to plain retransmissions as well? In that case, doesn't the browser just re-send whatever they sent before, TWCC extension included?

garry81:
I just ran quick test w/o RFC 4588 on Chrome, retransmission packet has new transport seq number as same as w/ RFC 4588

lminiero:
Looks good, thanks again for this fix, merging then!

PR #1403, opened 2018-10-10
Fix HTMLMediaElement.srcObject for older Chrome (< 52)

vahidov:
According to (https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/srcObject#Browser_compatibility) srcObject in HTMLMediaElement works in chrome after 52

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -293,7 +293,7 @@ Janus.init = function(options) {
 		Janus.attachMediaStream = function(element, stream) {
 			if(Janus.webRTCAdapter.browserDetails.browser === 'chrome') {
 				var chromever = Janus.webRTCAdapter.browserDetails.version;
-				if(chromever >= 43) {
+				if(chromever >= 52) {
 					element.srcObject = stream;
 				} else if(typeof element.src !== 'undefined') {
 					element.src = URL.createObjectURL(stream);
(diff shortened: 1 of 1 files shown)


lminiero:
Since Janus has been around for a few years already, if we had that value in I'm pretty sure it worked since the versions we wrote there :wink:

vahidov:
@lminiero Sure no ;-)

first of all - for chrome <=54 you need enable max-bundle on bundlePolicy (in this reason your have broken videoroom demo) - https://groups.google.com/forum/#!topic/meetecho-janus/AUSqJRegOfE.

Second - srcObject. On vp8 codec not working in chrome 49 on winXp. Using commited code - let it works. Tested on several computers.

lminiero:
Those versions of Chrome are so old that I really don't care, so that's fine by me. Can you please fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so that this can be merged?

vahidov:
done

lminiero:
Thanks, merging then!

dada-amater:
> Those versions of Chrome are so old that I really don't care, so that's fine by me. Can you please fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so that this can be merged?

Chrome 49 is specific. It is the last version of chrome for Windows XP and Vista. It is the only way for users with such old system to enjoy modern webrtc. In our case it makes 1.4%.

PR #1407, opened 2018-10-11
Fixed some compatibility issues in streaming plugin.

cnzjy:
Some rtsp server doesn't support GET_PARAMETER command, I used OPTIONS instead of it.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -5198,7 +5198,7 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 		const char *timeout = strstr(curldata->buffer, ";timeout=");
 		if(timeout != NULL) {
 			/* There's a timeout to take into account: take note of the
-			 * value for sending GET_PARAMETER keepalives later on */
+			 * value for sending OPTIONS keepalives later on */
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for your contribution!

I'm not super-familiar with RTSP devices, so I don't know if this will have an impact. Have you checked if OPTIONS is supported by all devices, and that it does indeed keep the session alive in all cases as GET_PARAMETER does? I don't want to end up in a place where devices with which it was working fine, now don't because OPTIONS doesn't work for them.

cnzjy:
An OPTIONS request returns the request types the server will accept, this is a basic command in rtsp. Refer to this article:
https://en.wikipedia.org/wiki/Real_Time_Streaming_Protocol
But the GET_PARAMETER request is an optional command. Some rtsp server such as ffserver (ffmpeg project) doesn't support this command.
I found VLC use OPTIONS request to keep the session live (I captured packets using Wireshark)  , so I commit this modification.

lminiero:
Sounds good, thanks for the clarification! Can you fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge then?

cnzjy:
I have signed it.

lminiero:
Thanks, merging then!

PR #1408, opened 2018-10-11
Allow for predefined number of threads/loops to handle all media

lminiero:
This is related to #1399, of which it is basically an extension.

As explained in that PR, each handle currently has its own thread, and its own event loop. This event loop is responsible of all incoming and outgoing media packets for that handle (ICE, DTLS, RTP, RTCP, datachannels), and for a few events. As such, the thread model for Janus handles is, by default, a thread per handle.

This is not bad, but since we now have a proper event loop to handle all that, we started wondering if, as an option, it might make sense to also allow for multiple of those event loops to be "hosted" on the same thread: considering the event-based nature of media and events, this might be an interesting road to persue, if not to evaluate its performances, e.g., in terms of the impact on context switching, resources, and additional operations the same thread would have to perform when sharing loops.

This is what this patch aims to do. It adds a new property that you can configure in `janus.cfg`, called `event_loops`: by default, this is `0`, meaning the feature is disabled, and everything works exactly as before (each handle has its own thread/loop, which is what Janus will keep on doing by default for a long time). Setting this value to a numeric value, instead, will change the pattern, and create, at startup, the specified amount of threads/event loops that all handles will share: when this feature is enabled, new handles will be assigned (at the moment via a simple round-robin) to one of the existing thread/loops, rather than creating a new one. At the same time, when the handle is detached we don't destroy the loop/thread, but we simple remove the user from the loop/thread it had been assigned to.

We did a very simple and brief functional test, which seemed to confirm the feature seems to work as expected: we'll need to do some more testing to assess its performance, and whether or not it's worth the effort. An interesting question, for instance, will be how many static event loops an interested user should create at startup, and what the dependency would be to the host machine characteristics (e.g., in terms of how many CPU cores there are). If this is a feature that interests you, please test and provide feedback!

### DIFF
--- a/conf/janus.cfg.sample.in
+++ b/conf/janus.cfg.sample.in
@@ -54,8 +54,8 @@ admin_secret = janusoverlord	; String that all Janus requests must contain
 ;
 ;reclaim_session_timeout = 0		; How long (in seconds) we should wait for a
 ;							janus session to be reclaimed after the transport
-;							is gone. After the transport is gone, a session 
-;							times out when no request is received for 
+;							is gone. After the transport is gone, a session
+;							times out when no request is received for
 ;							reclaim_session_timeout seconds (default=0s).
 ;							Setting this to 0 will disable the timeout
 ;							mechanism, and sessions will be destroyed immediately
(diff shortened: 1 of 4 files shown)


lminiero:
We did some testing of this patch as well, and found out how in some scenarios it can really improve performances. Considering it's disabled by default, and that we (and you) can play with it some more to identify some recommended settings, I'll merge.

PR #1413, opened 2018-10-16
Support for a couple of RTP extensions in the postprocessor

lminiero:
This patch is actually a revisitation of the original effort started in #1400, and so most of the credits should go to @Piasy.

It adds two new environment variables that can be passed when launching the post-processor:

1. `JANUS_PPREC_AUDIOLEVELEXT`: ID of the audio level RTP extension;
2. `JANUS_PPREC_VIDEOORIENTEXT`: ID of the video orientation RTP extension.

Manually passing these identifiers is needed as they're advertised via SDP, and so this info is lost when we save recordings to the .mjr format.

When an RTP extension ID is specified, the postprocessor also looks for their values in the RTP packets it goes through, and stores them in the structure it uses for each of the packets. This way, the info is available to the individual processors (e.g., webm, h264, opus, etc.) in case later on we find a way to use them properly for more than just printed information.

The audio level extension values are only printed when debugging is at least 5. The video orientation extension values, instead, are interpreted by the postprocessor to give a summary, that can be especially useful when just parsing recordings with `--parse`. Specifically, if the video is not "straight" (orientation=0), the postprocessor will tell you if the video is rotated, and if the video rotated more than once, the postprocessor will tell you that as well. The information is not used in any way by the postprocessor: it's just printed so that you're aware of that.

Again, the original idea came to @Piasy, and so most of the credits go to him. I'm just publishing this as a separate effort due to some differences in the way I wanted the code to look like: since I had already asked him for multiple revisions, it didn't seem fair to keep on asking for changes, and it seemed easier to do the changes myself. I tried to reuse the same code we use for parsing extensions in Janus, which should make it feel more consistent.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -17,15 +17,15 @@
  * G.711 (mu-law or a-law) frames. In case the recording contains text
  * frames as received via data channels, instead, a .srt file will be
  * generated with the text content and the related timing information.
- * 
+ *
  * Using the utility is quite simple. Just pass, as arguments to the tool,
  * the path to the .mjr source file you want to post-process, and the
  * path to the destination file, e.g.:
- * 
+ *
 \verbatim
 ./janus-pp-rec /path/to/source.mjr /path/to/destination.[opus|wav|webm|h264|srt]
-\endverbatim 
- * 
+\endverbatim
+ *
  * An attempt to specify an output format that is not compliant with the
  * recording content (e.g., a .webm for H.264 frames) will result in an
  * error since, again, no transcoding is involved.
@@ -45,7 +45,7 @@
  * the frames in a valid container. Any further post-processing (e.g.,
  * muxing audio and video belonging to the same media session in a single
  * .webm file) is up to third-party applications.
- * 
+ *
  * \ingroup postprocessing
  * \ref postprocessing
  */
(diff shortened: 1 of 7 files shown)


postprocessing/janus-pp-rec.c
```diff
@@ -476,19 +499,30 @@ int main(int argc, char *argv[])
...
 		JANUS_LOG(LOG_VERB, "  -- RTP packet (ssrc=%"SCNu32", pt=%"SCNu16", ext=%"SCNu16", seq=%"SCNu16", ts=%"SCNu32")\n",
 				ntohl(rtp->ssrc), rtp->type, rtp->extension, ntohs(rtp->seq_number), ntohl(rtp->timestamp));
 		if(rtp->csrccount) {
 			JANUS_LOG(LOG_VERB, "  -- -- Skipping CSRC list\n");
 			skip += rtp->csrccount*4;
 		}
+		audiolevel = -1;
+		rotation = -1;
 		if(rtp->extension) {
 			janus_pp_rtp_header_extension *ext = (janus_pp_rtp_header_extension *)(prebuffer+12);
```
Piasy: If this packet has CSRC field, then the extension is not at `prebuffer+12`.

Piasy:
Thanks for your work! It works well in my use case.

Besides, my origin PR contains several other small changes, I can submit separate PRs if you don't mind. :)

lminiero:
If they were generic fixes, a different pull request would be indeed
helpful. Thanks for the note on CSRC, I'll make the fix after lunch!

Il mar 16 ott 2018, 13:28 Piasy <notifications@github.com> ha scritto:

postprocessing/janus-pp-rec.c
lminiero: Actually there's no real need for that: this is just to print how many extensions there are, but we pass the start of the buffer to `janus_pp_rtp_header_extension_find`, which finds the actual extension we look for skipping CSRC as well. Anyway, it's true that we need to fix the check if we want to print the correct number there.

lminiero: Oh but we need that to update the skip value as well, so yeah, it was an important fix: thanks for spotting that!

lminiero:
> Thanks for your work! It works well in my use case.

Since I merged your point on extensions, I'll merge. Thanks again for your original effort, and looking forward to more fixes if you have others!

PR #1414, opened 2018-10-16
utils: constify read-only parameters

tmatth:


lminiero:
Won't this require casting to `const char *` when calling these methods?

tmatth:
@lminiero not in this case, only if going the other way (if passing a const char * parameter to a function expecting a char * as a parameter, you'd get warnings about discarding const).

lminiero:
Thanks for the clarification: looks good, thanks, merging!

PR #1427, opened 2018-10-26
Add support for dumping to .pcap directly

lminiero:
This patch extends the support we first added in #993.

As you may already know, in Janus you can currently dump the unencrypted traffic for a handle (RTP/RTCP) via a `start_text2pcap` request sent via Admin API: this results in each incoming or outgoing packet to be serialized to a text string with some metadata, and saved to a text file, which needs to subsequently be passed to `text2pcap` in order to get a `.pcap` or `.pcapng` file you can look at, e.g., via Wireshark. While this works and has been useful to me in several cases, it is also a resource hungry process: the text serialization in particular means that it takes quite a toll on the process, and CPU usage is not low.

As such, I started looking into ways to save to a `.pcap` file directly and natively, rather than going through the text file and the `text2pcap` translation, which is what this patch provides. I didn't want to add any additional dependency (in this case to libpcap), so this process is done manually in the core: the pcap headers are created accordingly, and each time we save a packet we create the ethernet/ip/udp layers we need to encapsulate the payload to dump. This seems to work nicely, from the few tests I made, and more importantly it's incredibly more lightweight: my guess is that it has pretty much the same impact as our .mjr recordings, so quite neglectible.

This doesn't replace the text2pcap support, it's an alternative, as there may be more things we might add to that in the future (e.g., comments to packets and other stuff that pcapng supports and basic pcap doesn't): as such, I partly reused the Admin API commands. If you use `start_text2pcap` as a command, you'll use the old approach, while if you use the `start_pcap` command, you'll use the new approach instead: all the other properties (e.g., the `folder`, `filename` and `truncate` attributes) are exactly the same; of course, use a different extension for the filename part depending on which approach you'll use (typically `.txt` for text2pcap and `.pcap` for pcap).

As an example, you can use this snippet on the JavaScript console when running the EchoTest demo and it should save all traffic:

```
Janus.httpAPICall("http://localhost:7088/admin/" + janus.getSessionId() + "/" + echotest.getId(), {
	verb: 'POST',
	withCredentials: false,
	body: {
		janus: "start_pcap",
		folder: "/tmp",
		filename: "my-test2pcap-dump.pcap",
		transaction: "123",
		admin_secret: "janusoverlord"
	},
	success: function(json) {
		Janus.vdebug(json);
	},
		Janus.error(textStatus + ": " + errorThrown);
	}
});
```

I'm planning to merge really soon, so feedback is welcome!

### DIFF
--- a/html/admin.html
+++ b/html/admin.html
@@ -135,11 +135,15 @@
 						<div id="info" class="col-md-8">
 							<h4>Handle Info <i id="update-handle" class="fa fa-refresh" title="Refresh handle info" style="cursor: pointer;"></i></h4>
 							<div id="options" class="hide">
-								<label class="checkbox-inline">
-									<input id="autorefresh"type="checkbox" value="">Autorefresh
+								<label class="checkbox-inline" title="Autorefresh this info every 5s">
+									<input id="autorefresh" type="checkbox" value="" title="Autorefresh this info every 5s">Autorefresh
 								</label>
-								<label class="checkbox-inline">
-									<input id="prettify"type="checkbox" value="">Prettify
+								<label class="checkbox-inline" title="Show information as HTML">
+									<input id="prettify" type="checkbox" value="" title="Show information as HTML">Prettify
+								</label>
+								<label class="checkbox-inline" title="Start of stop capturing traffic to .pcap">
+									<input id="capture" type="checkbox" value="" title="Start of stop capturing traffic to .pcap">
+									<span id="capturetext">Start capture</span>
 								</label>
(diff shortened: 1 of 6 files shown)


lminiero:
Also added a new control in the Admin API web demo to tweak pcap/text2pcap captures (to make @atoppi happy as he's too lazy to use curl! :laughing: )

atoppi:
LVGTM

lminiero:
Made some more tests, so I'll merge. I'll publish a blog post shortly that will provide some more information.

Issue #1300, opened 2018-07-11
RTP and RTCP timestamp synchronization problem for single stream


chadisid:
I am extracting rtp and rtcp packets of both audio and video stream from videoroom plugin. For example, When I extract rtp and rtcp packets  of same audio stream. The timestamp of rtp packet start with 0 always and RTP timestamp inside the RTCP packet is random value. Both are not synchronized. When I was using older janus 0.2.6 it was synchronized.  This enabled me to find  NTP time of each RTP packet. Now its not the case with 0.4.2. This issue started from following update. 
https://github.com/meetecho/janus-gateway/commit/77bcaa7f0de6cf681ea6fbdb4e55e4b02dfddd0b#diff-724b612068e82ba0ad6ed24da3a22e47R2117

In that janus_rtp_header_update where the timestamp is updated from original random timestamp to 0 in case of audio and video for RTP packet. But the RTP timestamp in the RTCP packet of the stream is not updated before sending to the plugin. So when I try to relay the RTP and RTCP packet of same stream one start with 0 at RTP packet and other start based on the random number in RTCP. Will RTCP update in ice.c will work or will it break any other things?. 
```
if(video) {
       	rtcp_header *r = (rtcp_header *)buf;
       	if (r->type == RTCP_SR) {
     		rtcp_sr *sr = (rtcp_sr*)r;
       		uint64_t rtp_ts = (unsigned long)ntohl(sr->si.rtp_ts);
       		uint32_t ntp_frac = (unsigned long)ntohl(sr->si.ntp_ts_lsw);
       		uint32_t ntp_sec =  (unsigned long)ntohl(sr->si.ntp_ts_msw);
                unsigned long long time_secs, time_usecs,time_secs_1, time_usecs_1;
                const unsigned long long OFFSET = 2208988800ULL;
                const unsigned long long NTP_SCALE = 4294967295ULL;
                time_secs =  ntp_sec - OFFSET;
                time_usecs = (ntp_frac + 1) * 1000000UL / NTP_SCALE;
                time_secs_1 =  ntp_sec - OFFSET;
                time_usecs_1 = (ntp_frac + 1) * 1000000UL / NTP_SCALE;
                struct timeval t;
                t.tv_sec = (time_t) time_secs;
                t.tv_usec = (suseconds_t) time_usecs;
                time_t nowtime;
                struct tm *nowtm;
                nowtime = t.tv_sec;
                nowtm = localtime((const time_t *)&nowtime);
[...]

PR #1428, opened 2018-10-26
Fix SSRCs and RTP timestamps in incoming RTCP reports

atoppi:
To have some more context refer to #1300.
This PR aims at fixing SSRC and RTP timestamp fields in the Sender Reports (SR) and Receiver Reports (RR) that the core is passing to a plugin, thus letting a renegotiation not disrupting the relationship between RTP and RTCP inside a plugin.
We have tried to accomplish something similar to what the Janus core already does for the RTP traffic. In order to keep coherent the flow of the RTCP traffic toward a plugin, the idea is to make use of the RTP context tied to that RTCP stream and to to re-use the local and remote SSRCs that Janus store for an active stream.
So for every incoming RTCP packet, we invoke `janus_rtcp_fix_report_data` just before passing the packet to a plugin. This function:
- parses the packet
- for every SR/RR it finds
- checks if the received SSRC is the one expected from the `janus_rtp_switching_context` (in negative case it drops the **entire** compound packet)
- updates the SSRCs according to the `janus_ice_stream` context
- updates the RTP timestamps according to the `janus_rtp_switching_context`

This seems to work quite well, however these changes do come with some caveats:
1) Some **dropping logic** has been added.
An entire RTCP packet will be dropped if the SSRC is not the one that either the Janus context or RTP context knows. This commonly means that while renegotiating, a compound packet containing a Goodbye RTCP message for the old SSRC and some packets containing a RTCP SR/RR message for the new SSRC are going to be dropped because in the former case the Janus context has already been updated (and so `old_ssrc != janus_context->current_ssrc`) while in the latter case we must wait some RTP traffic in order to update the RTP context (so `new_ssrc != rtp_context->current_ssrc`).
However I don't think this will be a big issue, because browsers tend to send RTCP feedback quite frequently, and I can't imagine a scenario where a lost RTCP will break something.
2) This patch **does not cover all RTCP messages** and **all SR/RR fields**.
We have tried to address to users that want to parse SR/RR in their custom plugins. So we restricted the focus to those report messages and even for those, some fields have still to be fixed (like cumulative lost packets, packet and octects count and things like these). However we don't think this is a priority as of now, because SSRC and timestamp should be enough for the majority of user cases.
3) The RTP timestamp in the SR/RR **may lose reliability** inside a plugin.
By using the RTP context, we are relying on the accuracy of the `time_diff` [estimate](https://github.com/meetecho/janus-gateway/blob/80f0e273a845d8ff2dcf2d586adc2d25c92d097f/rtp.c#L499). This means that every error occurring on this estimate will be reflected on the RTCP SR directed to a plugin after a renegotiation.
We're internally discussing on how this evaluation could be improved, but do not expect something too soon.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -2596,11 +2596,12 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 							video = 0;
 						} else if(rtcp_ssrc == stream->video_ssrc) {
 							video = 1;
-						} else {
+						} else if(janus_rtcp_has_fir(buf, len) || janus_rtcp_has_pli(buf, len) || janus_rtcp_get_remb(buf, len)) {
 							/* Mh, no SR or RR? Try checking if there's any FIR, PLI or REMB */
-							if(janus_rtcp_has_fir(buf, len) || janus_rtcp_has_pli(buf, len) || janus_rtcp_get_remb(buf, len)) {
-								video = 1;
-							}
+							video = 1;
+						} else {
+							JANUS_LOG(LOG_WARN,"[%"SCNu64"] Dropping RTCP packet with unknown SSRC (%"SCNu32")\n", handle->handle_id, rtcp_ssrc);
+							return;
 						}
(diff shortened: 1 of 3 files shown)


atoppi:
Leaving an [output](https://pastebin.com/vKSyWV5i) from a plugin perspective to better explain what is happening here.

PR #1431, opened 2018-10-29
Disconnect websocket on timeout gateway message (JS)

nevcos:
Janus Gateway sometimes sends a timeout message which is not properly handled by the existing Javascript Janus code (v0.4.5):

```
{
   "janus": "timeout",
   "session_id": 112005209173740
}
```

When this timeout message comes:
1. A warn is shown - `Unknown message/event timeout on session 112005209173740`
2. The connection keeps open and the Janus instance/session becomes unusable (returns server errors every time informing that the session no longer exists).

This PR code handles the timeout message, closing the websocket connection with the following code/message - `3504 / Gateway timeout`

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -701,10 +701,12 @@ function Janus(gatewayCallbacks) {
 				Janus.debug("No provided notification callback");
 			}
 		} else if(json["janus"] === "timeout") {
-		    Janus.error("Timeout on session " + sessionId);
-		    Janus.debug(json);
-		    ws.close(3504, "Gateway timeout");
-		    return;
+			Janus.error("Timeout on session " + sessionId);
+			Janus.debug(json);
+			if (websockets) {
+				ws.close(3504, "Gateway timeout");
+			}
+			return;
 		} else {
 			Janus.warn("Unknown message/event  '" + json["janus"] + "' on session " + sessionId);
 			Janus.debug(json);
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for the contribution! Can you sign and fill the form in our [CLA](https://cla.conf.meetecho.com/janus) so that this can be merged after a review?

nevcos:
@lminiero Done, thanks!

html/janus.js
```diff
@@ -700,6 +700,11 @@ function Janus(gatewayCallbacks) {
 				// Send to generic callback (?)
 				Janus.debug("No provided notification callback");
 			}
+		} else if(json["janus"] === "timeout") {
+		    Janus.error("Timeout on session " + sessionId);
+		    Janus.debug(json);
+		    ws.close(3504, "Gateway timeout");
```
lminiero: This seems to assume this callback will always be triggered by a websocket, which is not the case. The same callback is used for HTTP as well, which would break here.

lminiero:
> @lminiero Done, thanks!

Are you sure? I don't see your signature in the list... did you also fill in the form at the end of the page?

nevcos:
@lminiero I'm sorry, I've miss the end of the page :blush:. Now signed :+1:

lminiero:
@nevcos can you address the comment I added in the review? I don't see other obstacles to this getting merged. Thanks!

html/janus.js
nevcos: Good point :+1: will fix it

lminiero:
Thanks for the quick fix: merging then!

nevcos:
@lminiero: It was a pleasure 👍!
Thanks you for this awesome project 🏆

PR #1432, opened 2018-10-30
sdp-utils: minor doc corrections

tmatth:


lminiero:

PR #1433, opened 2018-10-31
sdp-utils: use enum type instead of defines

tmatth:
No functional change.


lminiero:
Probably cleaner, thanks :+1: 
Merging!

PR #1441, opened 2018-11-17
prometheus plugin

DmitryYudin:
There is initial version which only expose events counter to http://$host:$port/metrics.

The plugin depends on prometheus-cpp static library which can be built as:
```
git clone --depth 1 -b v0.6.0 --recurse-submodules https://github.com/jupp0r/prometheus-cpp && cd $(basename "$_") && \
mkdir -p .build && cd "$_" && cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DENABLE_PUSH=OFF -DENABLE_TESTING=OFF .. && \
make -j8 && \
make install
```

Hope, you'll find it useful)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -323,6 +323,17 @@ conf_DATA += conf/janus.eventhandler.mqttevh.cfg.sample
 EXTRA_DIST += conf/janus.eventhandler.mqttevh.cfg.sample
 endif
 
+if ENABLE_PROMETHEUSEVH
+event_LTLIBRARIES += events/libjanus_prometheusevh.la
+events_libjanus_prometheusevh_la_SOURCES = events/janus_prometheusevh.cc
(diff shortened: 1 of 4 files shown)


lminiero:
Thanks for working on this! To be honest, I'm not familiar with Prometheus, so I'll need some time to review this.

lminiero:
I've given this some thought, and I believe this would be better served as a separate project, e.g., as Mozilla did for their SQLite Event Handler: https://github.com/mozilla/janus-eventhandler-sqlite
The reasoning behind this is that the configure check seems overly complicated, we're involving C++ in a C project, and I know nothing about Prometheus: considering I'd have to maintain this after it's merged, I'd rather this lives elsewhere where it can be taken care of.

Thanks for your contribution anyway! Please let me know if you'll publish this somewhere an ad-hoc repository, and I'll make sure it's listed in the [resources](https://janus.conf.meetecho.com/docs/resources) page where interested people can find it.

PR #1443, opened 2018-11-21
Fix segfault when specifying a config file with no slashes

ColonelThirtyTwo:
Ex. when running `janus -C janus.config`.

Also fix some const-correctness with get_filename


ColonelThirtyTwo:
Nevermind, looks like its already fixed in master

PR #1446, opened 2018-11-23
Initialize Last Will and Testament properties for mqttevh

0nkery:
Hi!

Added initialization for MQTT's Last Will and Testament options. There was some default
values but they weren't used anywhere. I tested it manually and it works.

But it comes that LWT is enabled by default now. Should I add some flag to enable/disable it?

### DIFF
--- a/conf/janus.eventhandler.mqttevh.jcfg.sample
+++ b/conf/janus.eventhandler.mqttevh.jcfg.sample
@@ -28,10 +28,11 @@ general: {
 	#addevent = true				# Whether we should add the event type to the base topic
 
 	# Initial message sent to status topic
-	#initial_status = "{\"event\": \"connected\", \"eventhandler\": \"janus.eventhandler.mqttevh\"}"
+	#connect_status = "{\"event\": \"connected\", \"eventhandler\": \"janus.eventhandler.mqttevh\"}"
+	# Message sent after disconnect or as LWT
+	#disconnect_status = "{\"event\": \"disconnected\"}"
 
 	#will_enabled = false							# Whether to enable LWT (default: false)
-	#will_content = "{\"event\": \"disconnected\"}"	# LWT content
 	#will_retain = 1								# Whether LWT should be retained (default: 1)
 	#will_qos = 0									# QoS for LWT (default: 0)
 
--- a/conf/janus.eventhandler.mqttevh.jcfg.sample
+++ b/conf/janus.eventhandler.mqttevh.jcfg.sample
@@ -27,9 +27,9 @@ general: {
 	#topic = "/janus/events"		# Base topic (default: /janus/events)
 	#addevent = true				# Whether we should add the event type to the base topic
 
-	#will_enabled = no								# Whether to enable LWT (default: no)
+	#will_enabled = false							# Whether to enable LWT (default: false)
 	#will_content = "{\"event\": \"disconnect\"}"	# LWT content
-	#will_retain = yes								# Whether LWT should be retained (default: yes)
+	#will_retain = 1								# Whether LWT should be retained (default: 1)
 	#will_qos = 0									# QoS for LWT (default: 0)
 
 	# Additional parameters if "mqtts://" schema is used
--- a/conf/janus.eventhandler.mqttevh.jcfg.sample
+++ b/conf/janus.eventhandler.mqttevh.jcfg.sample
@@ -16,15 +16,16 @@ general: {
 							# plain (no indentation) or compact (no indentation and no spaces)
 
 	url = "tcp://localhost:1883"	# The URL of the MQTT server. Only tcp supported at this time.
-	client_id = "janus.example.com"	# Janus client id. You have to configure a unique ID.
+	client_id = "janus.example.com"	# Janus client id. You have to configure a unique ID (default: guest).
 	#keep_alive_interval = 20		# Keep connection for N seconds (default: 30)
 	#cleansession = 0				# Clean session flag (default: off)
-	#mqtt_qos = 1					# Default MQTT QoS for published events
+	#retain = 0						# Default MQTT retain flag for published events
+	#qos = 1						# Default MQTT QoS for published events
 	#disconnect_timeout = 100		# Seconds to wait before destroying client
 	#username = "guest"				# Username for authentication (default: no authentication)
 	#password = "guest"				# Password for authentication (default: no authentication)
 	#topic = "/janus/events"		# Base topic (default: /janus/events)
-	#addtype = true					# Whether we should add the event type to the base topic
+	#addevent = true				# Whether we should add the event type to the base topic
 
 	#will_enabled = no								# Whether to enable LWT (default: no)
 	#will_content = "{\"event\": \"disconnect\"}"	# LWT content
--- a/conf/janus.eventhandler.mqttevh.jcfg.sample
+++ b/conf/janus.eventhandler.mqttevh.jcfg.sample
@@ -27,8 +27,11 @@ general: {
 	#topic = "/janus/events"		# Base topic (default: /janus/events)
 	#addevent = true				# Whether we should add the event type to the base topic
 
+	# Initial message sent to status topic
+	#initial_status = "{\"event\": \"connected\", \"eventhandler\": \"janus.eventhandler.mqttevh\"}"
+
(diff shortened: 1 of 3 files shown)


lminiero:
Yes, it should be optional and disabled by default, please, for backwards compatibility. Anyway, I just noticed that I forgot to convert the MQTT evh configuration file to the new jcfg format, I'll have to fix that.

lminiero:
Thanks for the quick fix! Please see the changes I've made to the configuration file, though, which is now in the new libconfig format.

0nkery:
Okay, I'll merge these changes later.

And there is also some issue with retained LWT that should be considered. In case of abnormal disconnect when LWT is fired and retained, some subscriber will not be available to tell if Janus is up or not if this subscriber will crash later on. So proposal is to add another one message which will be fired right after mqttevh start. I'll add it later too.

lminiero:
> And there is also some issue with retained LWT that should be considered. In case of abnormal disconnect when LWT is fired and retained, some subscriber will not be available to tell if Janus is up or not if this subscriber will crash later on. So proposal is to add another one message which will be fired right after mqttevh start. I'll add it later too.

I think that's not needed. We can't be responsible for subscribers dying as well. It feels like your suggested approach would overcomplicate the plugin behaviour.

conf/janus.eventhandler.mqttevh.jcfg.sample
```diff
@@ -16,16 +16,21 @@ general: {
 							# plain (no indentation) or compact (no indentation and no spaces)
 
 	url = "tcp://localhost:1883"	# The URL of the MQTT server. Only tcp supported at this time.
-	client_id = "janus.example.com"	# Janus client id. You have to configure a unique ID.
+	client_id = "janus.example.com"	# Janus client id. You have to configure a unique ID (default: guest).
 	#keep_alive_interval = 20		# Keep connection for N seconds (default: 30)
 	#cleansession = 0				# Clean session flag (default: off)
-	#mqtt_qos = 1					# Default MQTT QoS for published events
+	#retain = 0						# Default MQTT retain flag for published events
+	#qos = 1						# Default MQTT QoS for published events
```
lminiero: Why have you changed the name of that property? That's a breaking change for people using the plugin in production right now.

conf/janus.eventhandler.mqttevh.jcfg.sample
```diff
@@ -16,16 +16,21 @@ general: {
...
 	#disconnect_timeout = 100		# Seconds to wait before destroying client
 	#username = "guest"				# Username for authentication (default: no authentication)
 	#password = "guest"				# Password for authentication (default: no authentication)
 	#topic = "/janus/events"		# Base topic (default: /janus/events)
-	#addtype = true					# Whether we should add the event type to the base topic
+	#addevent = true				# Whether we should add the event type to the base topic
```
lminiero: Same here: you can't just change the name of properties.

conf/janus.eventhandler.mqttevh.jcfg.sample
lminiero: Apologies, my mistake... just noticed that it's the name in the config that's wrong, and that it's actually "qos" in the code :+1:

conf/janus.eventhandler.mqttevh.jcfg.sample
lminiero: My fault here too, please disregard!

conf/janus.eventhandler.mqttevh.jcfg.sample
0nkery: Yep, I just adjusted config properties with real property names in code.

lminiero: Thanks for catching those mistakes! (which I should have catched myself in the first place)

0nkery:
Glad to help. :)

So preferable way of doing this is to send such a message by myself in my own plugin logic? Then the next question arises: how to make this one message retained while leaving all the others non-retained?

BTW, this plugin already sends kind-of-initial message and my change requires only to check if LWT is retained and make this initial message retained or not.

If we decide that this change is not going to be implemented, then we should maybe remove `will_retain` config or add some kind of warning about possible implications of using `will_retain`.

manifest:
It may make sense for people who will decide to use MQTT transport plugin without the MQTT event handler, but first things first, we definitely need it here and can add support of LWT to transport plugin any time later.

0nkery:
@lminiero Hi, could you review last changes, please?

lminiero:
Sure, I'll do it later today.

manifest:
@0nkery Sending `{"event": "disconnect"}` on normal connection close is missed as far I can see. It may also be useful to move those two status messages to configuration to allow users of the event handler customize their format.

events/janus_mqttevh.c
```diff
@@ -369,7 +384,7 @@ static void janus_mqttevh_client_reconnect_success(void *context, MQTTAsync_succ
 	}
 	json_t *info = json_object();
 	json_object_set_new(info, "event", json_string("connected"));
```
lminiero: Shouldn't the two lines above be removed, since we're not using this statically? (`janus_mqttevh_send_message` is commented out)

events/janus_mqttevh.c
```diff
@@ -522,7 +550,7 @@ static int janus_mqttevh_init(const char *config_path) {
 	}
 	/* Read configuration */
 	char filename[255];
-	g_snprintf(filename, 255, "%s/%s.cfg", config_path, JANUS_MQTTEVH_PACKAGE);
+	g_snprintf(filename, 255, "%s/%s.jcfg", config_path, JANUS_MQTTEVH_PACKAGE);
```
lminiero: You're missing the fallback to `.cfg` that the other plugins have. This is actually my fault, but since you're fixing this part as well, the fallback should be added here.

events/janus_mqttevh.c
```diff
@@ -633,6 +663,41 @@ static int janus_mqttevh_init(const char *config_path) {
...
+	} else {
+		ctx->publish.connect_status = g_strdup(DEFAULT_CONNECT_STATUS);
+	}
+
+	disconnect_status_item = janus_config_get(config, config_general, janus_config_type_item, "disconnect_status");
+	if(disconnect_status_item && disconnect_status_item->value) {
+		ctx->publish.disconnect_status = g_strdup(disconnect_status_item->value);
+	} else {
+		ctx->publish.disconnect_status = g_strdup(DEFAULT_DISCONNECT_STATUS);
+	}
```
lminiero: But does this mean that the connect/disconnect stuff is ALWAYS sent, now? I'm not sure I'm ok with this: as the last will, this should be configurable and optional. The Janus core already sends a message when starting up, and one when shutting down.

events/janus_mqttevh.c
0nkery: Yep, it was there and I decided to not touch this part for now. I'll be happy to remove dead code. :)

events/janus_mqttevh.c
0nkery: Okay, no problem.

lminiero:
PS: did you change your Github username? If so, could you sign the [CLA](http://cla.conf.meetecho.com/janus/) again, as it's tied to the username that made the contribution? Thanks!

events/janus_mqttevh.c
0nkery: Connection stuff was there before. Disconnection stuff was added by me to keep things simmetrical - consider the case where Janus gracefully terminates and retained status message still indicates that Janus is online which is not true.

There is no way to retain such events coming from Janus Core, I guess.

0nkery:
I didn't sign any CLA before but I'll do.

0nkery:
Done that, Lorenzo. :)

events/janus_mqttevh.c
0nkery: Added fallback and made config parsing more robust.

events/janus_mqttevh.c
lminiero: Got it, thanks for the clarification.

lminiero:
Looks good, thanks, merging then! :+1:

lminiero:
Fixed a couple of typos I found when updating my local instance (an unused variable, and a broken printf format).

0nkery:
Thanks!

PR #1450, opened 2018-11-29
Don’t put rtx packets in retransmission buffer

glenn-hpcnt:
If rtx packet(RFC4588) put in buffer, accidentally rtx packet sequence will be conflicted by another rtp sequence(If packets got same sequence)

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3949,7 +3949,7 @@ static gboolean janus_ice_outgoing_traffic_handle(janus_ice_handle *handle, janu
 				}
 				/* Before encrypting, check if we need to copy the unencrypted payload (e.g., for rtx/90000) */
 				janus_rtp_packet *p = NULL;
-				if(max_nack_queue > 0 && pkt->type == JANUS_ICE_PACKET_VIDEO && component->do_video_nacks &&
+				if(max_nack_queue > 0 && !pkt->retransmission && pkt->type == JANUS_ICE_PACKET_VIDEO && component->do_video_nacks &&
 						janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_RFC4588_RTX)) {
 					/* Save the packet for retransmissions that may be needed later: start by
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for this: I thought we had some checks for that already, so let me review and I'll get back to you later today :+1:

lminiero:
It looks like we can just use the `pkt->retransmission` property for this check. We should also add a similar check a few lines before (~3952), that is here:

	/* Before encrypting, check if we need to copy the unencrypted payload (e.g., for rtx/90000) */
	janus_rtp_packet *p = NULL;
	if(max_nack_queue > 0 && pkt->type == JANUS_ICE_PACKET_VIDEO && component->do_video_nacks &&
			janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_RFC4588_RTX)) {
		/* Save the packet for retransmissions that may be needed later: start by
		 * making room for two more bytes to store the original sequence number */

which should become something like this:

	/* Before encrypting, check if we need to copy the unencrypted payload (e.g., for rtx/90000) */
	janus_rtp_packet *p = NULL;
	if(max_nack_queue > 0 && !pkt->retranmission && pkt->type == JANUS_ICE_PACKET_VIDEO && component->do_video_nacks &&
			janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_RFC4588_RTX)) {
		/* Save the packet for retransmissions that may be needed later: start by
		 * making room for two more bytes to store the original sequence number */

while I'd simplify your patch at ~4050, turning it simply into:

	if(max_nack_queue > 0 && !pkt->retransmission) {

Does this make sense, and do you see this working? If so, I can make the changes myself, since it would just be a matter of adding a couple of `pkt->retransmission` here and there.

glenn-hpcnt:
I think it would be better. so I'm added !pkt->retransmission in if statement

lminiero:
Looks good, thanks, merging!

PR #1451, opened 2018-11-29
Nanomsg transport libconfig migration

mickiebyrd:
Nanomsg transport seems to have been missed in the conversion to libconfig. The .jcfg configuration file is never tried.


mickiebyrd:
Sure

lminiero:
Thanks, merging then!

PR #1452, opened 2018-12-01
ice: factor duplicate code into loop

tmatth:
No functional change


lminiero:
I'm making a lot of changes to the core internals in the unified-plan branch, so considering it changes a lot of the properties you touch, I won't merge this, sorry.

tmatth:
@lminiero np, thanks for the feedback.

PR #1454, opened 2018-12-02
ice: increase stream refcount while nacked cleanup callback is alive

### LINKED ISSUE #1453: Segfault in  janus_ice_nacked_packet_cleanup
This is hard to reproduce, but using commit c0e9577b I hit the following segfault:

```
janus.log:ASAN:DEADLYSIGNAL
janus.log-=================================================================
janus.log-==13==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000320 (pc 0x557b4c81dec1 bp 0x7f6399eed4e0 sp 0x7f6399eed2f0 T50)
janus.log-==13==The signal is caused by a READ memory access.
janus.log-==13==Hint: address points to the zero page.
janus.log-#0 0x557b4c81dec0 in janus_ice_nacked_packet_cleanup /tmp/janus-gateway/ice.c:330
janus.log-#1 0x7f63b0caac72  (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x4cc72)
janus.log-#2 0x7f63b0caa1f4 in g_main_context_dispatch (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x4c1f4)
janus.log-#3 0x7f63b0caa5bf  (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x4c5bf)
janus.log-#4 0x7f63b0caa8d1 in g_main_loop_run (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x4c8d1)
janus.log-#5 0x557b4c842ae9 in janus_ice_handle_thread /tmp/janus-gateway/ice.c:1056
janus.log-#6 0x7f63b0cd2104  (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x74104)
janus.log-#7 0x7f63af7f26da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)
janus.log-#8 0x7f63af51b88e in __clone (/lib/x86_64-linux-gnu/libc.so.6+0x12188e)
janus.log-
janus.log-AddressSanitizer can not provide additional info.
janus.log-SUMMARY: AddressSanitizer: SEGV /tmp/janus-gateway/ice.c:330 in janus_ice_nacked_packet_cleanup
janus.log-Thread T50 (hloop 618052118) created by T9 here:
janus.log-#0 0x7f63b1656d2f in __interceptor_pthread_create (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x37d2f)
janus.log-#1 0x7f63b0cefe2f  (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x91e2f)
janus.log-
janus.log-Thread T9 created by T0 here:
janus.log-#0 0x7f63b1656d2f in __interceptor_pthread_create (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x37d2f)
janus.log-#1 0x7f63b0cefe2f  (/usr/lib/x86_64-linux-gnu/libglib-2.0.so.0+0x91e2f)
janus.log-
```

Is it possible that this timeout:
https://github.com/meetecho/janus-gateway/blob/c0e9577b06862c7d90461eda24cd2c0257c30cad/ice.c#L330

is being invoked due the timeout created here:
https://github.com/meetecho/janus-gateway/blob/c0e9577b06862c7d90461eda24cd2c0257c30cad/ice.c#L2496

after this block is invoked?
https://github.com/meetecho/janus-gateway/blob/c0e9577b06862c7d90461eda24cd2c0257c30cad/ice.c#L1400

tmatth:
Fixes #1453

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -329,6 +329,7 @@ static gboolean janus_ice_nacked_packet_cleanup(gpointer user_data) {
 		pkt->handle->handle_id, pkt->seq_number, pkt->handle->stream->video_ssrc_peer[pkt->vindex], pkt->vindex);
 	g_hash_table_remove(pkt->handle->stream->rtx_nacked[pkt->vindex], GUINT_TO_POINTER(pkt->seq_number));
 
+	janus_refcount_decrease(&pkt->handle->stream->ref);
 	return G_SOURCE_REMOVE;
 }
 
@@ -2494,6 +2495,7 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 								np->seq_number = cur_seq->seq;
 								np->vindex = vindex;
 								GSource *timeout_source = g_timeout_source_new_seconds(5);
+								janus_refcount_increase(&stream->ref);
 								g_source_set_callback(timeout_source, janus_ice_nacked_packet_cleanup, np, (GDestroyNotify)g_free);
 								g_source_attach(timeout_source, handle->mainctx);
 								g_source_unref(timeout_source);


lminiero:
This will result in a leak, if we quit the loop (handle being closed) and there was such an event pending: since it will never be fired, stream will never be unreffed, which will probably cause other properties never to be deallocated (possibly even handle itself).

tmatth:
@lminiero good point, would it make sense to do something like (assuming it's possible):
* on g_main_context teardown, look to see if there are still registered nacked_packet_cleanup callbacks
* decref the Ice_stream for each?

@atoppi any suggestions?

In testing, I'm at 0 dangling references, but I hadn't considered the case where the main context is gone without having fired the callback.

lminiero:
> * on  quitting the loop, look to see if there are still registered nacked_packet_cleanup callbacks

I'm not sure it's possible with Glib, IIRC I looked and didn't find anything to inspect the internal queue for events to dispatch. My feeling is that, if it's accessing the stream property in `janus_ice_nacked_packet_cleanup` that can cause the issue, we can just check if that is not NULL before using it. Since stream is set to NULL only in `janus_ice_webrtc_free`, and that method is only be called by the loop itself (`janus_ice_outgoing_traffic_finalize` and `janus_ice_handle_thread`), we'll never incur in race conditions there.

atoppi:
For future reference: indeed looks like there is a [method](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#g-main-context-pending) to check for pending sources on a glib context.

As for the issue and the PR: the issue #1453 is probably caused by a hangup served before a nack cleanup task [here](https://github.com/meetecho/janus-gateway/blob/master/ice.c#L3739). I agree with @lminiero, in general we should check for `handle->stream` pointer in the cleanup task.
A good idea might be removing the cleanup task source before invoking `janus_ice_webrtc_free` like [we already do for others sources](https://github.com/meetecho/janus-gateway/blob/master/ice.c#L3724), adding these lines
```
if(handle->nacked_cleanup_source) {
	g_source_destroy(handle->nacked_cleanup_source);
	g_source_unref(handle->nacked_cleanup_source);
	handle->nacked_cleanup_source = NULL;
}
```
so not letting the timeout trigger.

tmatth:
Thanks @atoppi and @lminiero for the quick feedback, I've opened a new PR based on your suggestions: https://github.com/meetecho/janus-gateway/pull/1456

PR #1458, opened 2018-12-04
RFC: Use timestamp as session version

tmatth:
This is recommended by https://tools.ietf.org/html/rfc4566#section-5.2


lminiero:
I'm not sure that's a good idea. Maybe for the initial version, but in updates it needs to be incremental AFAIK. I think browsers start from 1 as well.

PR #1464, opened 2018-12-08
Replace enable with enabled

manifest:
MQTT and RabbitMQ transport plugins mistakenly have `enable` property in their configuration while other plugins have `enabled`. It seems as a good time to change those properties' names during migration to `jcfg`.

### DIFF
--- a/conf/janus.transport.mqtt.jcfg.sample
+++ b/conf/janus.transport.mqtt.jcfg.sample
@@ -1,6 +1,6 @@
 # Configuration of the MQTT additional transport for the Janus API.
 general: {
-	enable = false						# Whether the support must be enabled
+	enabled = false						# Whether the support must be enabled
 	json = "indented"						# Whether the JSON messages should be indented (default),
 										# plain (no indentation) or compact (no indentation and no spaces)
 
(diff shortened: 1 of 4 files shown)


lminiero:
I'm hesitant to accept it, as it means a breaking change for those using these plugins in production right now. At the very least, the plugins should accept the old syntax as well (maybe with a warning printed when found), even though the new configuration files only list the new.

manifest:
> It means a breaking change for those using these plugins in production right now
Right. I see now that there is a callback on obsolete configuration format.

> At the very least, the plugins should accept the old syntax as well (maybe with a warning printed when found), even though the new configuration files only list the new
Fair enough.

lminiero:
> Right. I see now that there is a callback on obsolete configuration format.

I mostly meant something like this:

    janus_config_item *ssl = janus_config_get(config, config_general, janus_config_type_item, "ssl_enabled");
    if(ssl == NULL) {
    	/* Try legacy property */
    	ssl = janus_config_get(config, config_general, janus_config_type_item, "ssl_enable");
    	if(ssl && ssl->value) {
    		JANUS_LOG(LOG_WARN, "Found deprecated 'ssl_enable' property, please update it to 'ssl_enabled' instead\n");
    	}
    }

This way we start using the versions with "enabled" in, but we still allow configurations with the old "enable" versions to work (with a warning on the console), at least for a little while.

lminiero:
Thanks, merging then!

PR #1465, opened 2018-12-10
Fix code execution order

0nkery:
SIlly me did things in the wrong order - basically sending `disconnected` message AFTER actual disconnect.


lminiero:
Silly me as well, I should have spotted this when reviewing :slightly_smiling_face: 
Merging, thanks for the fix!

PR #1470, opened 2018-12-13
rtcp: fix get_remb bugs

fippo:


rtcp.c
```diff
@@ -942,14 +942,15 @@ uint32_t janus_rtcp_get_remb(char *packet, int len) {
...
 		return 0;
 	janus_rtcp_header *rtcp = (janus_rtcp_header *)packet;
-	if(rtcp->version != 2)
-		return 0;
 	/* Get REMB bitrate, if any */
 	int total = len;
+	int offset = 0;
 	while(rtcp) {
+                if(rtcp->version != 2)
+                        break;
```
lminiero: Looks like there still is some weird indentation in the above two lines, could you fix that?

fippo: done... I think. Now it looks good in less and the patch

lminiero:
Looks good! I'll apply the same changes to other methods where we iterate on compound packets in `rtcp.c`. I'm sure there will be more to fix, but it's a start. Thanks!

PR #1474, opened 2018-12-14
Improve Makefile.am and configure.ac to better support clang compiler.

atoppi:
This PR aims at enhancing Janus build-chain in order to better support the `clang` compiler.

**What has been added**
When in use, the `clang` compiler is detected  while issuing `configure`. This distinction is necessary because some specific `CFLAGS` that are used for `gcc` have a different name in `clang`.
<del>e.g. `-rdynamic` in `gcc`  becomes `-Wl,--export-dynamic` in `clang`.</del>
(Now `-Wl,--export-dynamic` is used as a common flag)
e.g. `-Wno-override-init` becomes `-Wno-initializer-overrides` in `clang`.

**What has been improved in general**
The `CFLAGS` are not appended in Makefile anymore but in the `configure.ac`<del>but the Macro `AX_APPEND_COMPILE_FLAGS` has been added to `configure.ac`. The macro checks for every flag availability and then appends it to `CFLAGS`.</del>
<del>**IMPORTANT**: To use this new macro I had to install the package `autoconf-archive`.</del>
The obsolete Macro `AC_GNU_SOURCE` has been replaced by `AC_USE_SYSTEM_EXTENSIONS`.

**How to compile Janus with `clang`**
First install `clang`. In Ubuntu this means `apt install clang`. 
Just set `clang` as your compiler while issuing the usual commands.
```
./autogen.sh
./configure CC=clang
sudo make clean
make -j$(nproc)
sudo make install
```
I have added a <del>"`using clang compiler`"</del> "`using compiler`" print in the configuration summary to double-check if the script has recognized the compiler.

**Why so many warnings?**
When compiling Janus with `clang` expect A LOT of warnings about memory alignment. This compiler seems to be more sensible to pointer casting respect to `gcc`. However many of those warns (especially the ones coming from the ICE stack `char*` buffers castings) should be bogus warnings, because the compiler ignores that those pointers contains very large `char[]` arrays and those arrays, at least on my x86_64 machine, are  16-bytes aligned.
The other warnings about the printf format can also be ignored.
**EDIT:** the cast alignment warnings have now been suppressed (really too much noise).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,23 +1,8 @@
 ACLOCAL_AMFLAGS = -I m4
 
-# FIXME: These flags should be produced in configure.ac using AX_APPEND_COMPILE_FLAGS
-AM_CFLAGS = \
-	-Wall -Wstrict-prototypes -Wmissing-prototypes -Wmissing-declarations \
-	-Wunused -fstrict-aliasing -Wextra -Wformat=2 -Winit-self -Winline \
-	-Wpacked -Wpointer-arith -Wmissing-format-attribute -Wmissing-noreturn \
-	-Wnested-externs -Wstrict-prototypes -Wunsafe-loop-optimizations \
-	-Wwrite-strings -Wno-missing-field-initializers -Wno-unused-parameter \
-	-Wcast-align -Wformat-nonliteral -Wformat-security -Wswitch-default \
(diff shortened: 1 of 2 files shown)


lminiero:
Looks good, merging!

PR #1487, opened 2019-01-02
Missing sendDtmf success callback call (JS)

nevcos:
The `sendDtmf` receives a `callbacks` object but `callbacks.success` is never called.

Given that `insertDTMF` doesn't return a `Promise`, the fix just calls the success callback in the last line.


lminiero:
Thanks for catching that, merging!

PR #1489, opened 2019-01-07
Rtcp fixes

atoppi:
This PR follows @fippo work on #1470 and fixes all the bugs found out while fuzzing the RTCP parsing methods in Janus. 
We added a bunch of new methods to check the RTCP packets before trying to access their fields, namely:
- `janus_rtcp_check_len` : validate minimum packet size and actual packet size against the header length field.
-  `janus_rtcp_check_rr`: validate the packet size against the report block count field.
-  `janus_rtcp_check_sr`:  validate minimum packet size for RTCP SR and validate the packet size against the report block count field.
-  `janus_rtcp_check_fci`: validate RTCP feedback messages containing a defined size FCI payload. E.g. Generic NACKs (4 bytes FCI), FIR (8 bytes FCI), PLI (0 bytes FCI). Also AFB is checked with a 0 bytes FCI because actual FCI has variable length in AFB.
- `janus_rtcp_check_remb`: check RTCP AFB REMB message minimum size and validate against the numssrc field.

Also we have extended the methods `janus_is_rtp` and `janus_is_rtcp` to check the packet length, because they were used before SRTP decrypting and so before the checks libsrtp does.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -286,12 +286,14 @@ static gboolean janus_is_dtls(gchar *buf) {
 	return ((*buf >= 20) && (*buf <= 64));
 }
 
-static gboolean janus_is_rtp(gchar *buf) {
+static gboolean janus_is_rtp(gchar *buf, guint len) {
+	if (len < 2) return FALSE;
 	janus_rtp_header *header = (janus_rtp_header *)buf;
 	return ((header->type < 64) || (header->type >= 96));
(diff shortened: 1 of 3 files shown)


lminiero:
Looks good, merging! And thanks to @fippo for kickstarting all this :1st_place_medal:

PR #1494, opened 2019-01-17
Added support for VP9 simulcasting

lminiero:
Since it was announced that [Chrome Canary now supports simulcasting for VP9 too](https://twitter.com/juberti/status/1085764367113572353), I tweaked the Janus code to allow you to use it. Luckily enough, I didn't need to change much: in fact, we already had code to parse temporal scalability info in VP9 payload descriptors (for SVC), so I just had to hook that up to the simulcast code, and loosen up some checks we had in the plugins that do simulcasting.

From a quick check (I only tested with the EchoTest plugin, by passing the `?simulcast=true&vcodec=vp9` query string to the demo), it seems to be working, with a small caveat: I do see temporal layers, and I can select them, but I never see SSRC changes, which seems to suggest the other substreams are not being sent. The resolution in the only SSRC I get is quite high, so that seems to confirm it. I'll check if it is a bug in Chrome.

Planning to merge soon, so if you have objections please let me know.

### DIFF
--- a/html/echotest.js
+++ b/html/echotest.js
@@ -59,6 +59,8 @@ var audioenabled = false;
 var videoenabled = false;
 
 var doSimulcast = (getQueryStringValue("simulcast") === "yes" || getQueryStringValue("simulcast") === "true");
+var acodec = (getQueryStringValue("acodec") !== "" ? getQueryStringValue("acodec") : null);
+var vcodec = (getQueryStringValue("vcodec") !== "" ? getQueryStringValue("vcodec") : null);
 var simulcastStarted = false;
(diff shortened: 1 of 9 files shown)


lminiero:
Looks like it was a miscommunication, and that simulcasting is still not there for VP9: the fact that temporal layers appear seems to be a "bug", and a side effect of simulcasting code in the SDP. Anyway, I'll keep this PR active in case VP9 ends up getting simulcast support, for the time being.

lminiero:
Closing as it may never happen. In case, we can restart from here.

Issue #1498, opened 2019-01-25
Streaming mount point info lacks properties if no secret is set


Musashi178:
If requesting the info for a RTP streaming mount point where no secret has been set, no "admin" information will be returned like audio UDP or video UDP port. 

The only workaround for this is to always use a mount point secret, but there might be situations where I do not want that.

lminiero:
Closing as we merged your fix.

PR #1499, opened 2019-01-25
enables extended mount point info by default if no secret is assigned

Musashi178:
This solves #1498 by setting admin to true if no secret has been set for the streaming mount point.


lminiero:
Thanks for spotting that! Can you fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/), so that I can merge this?

Musashi178:
Done

lminiero:
Thanks, merging then!

PR #1501, opened 2019-01-29
 remove unnecessary line cause session delete in advance

richard1219:


lminiero:
Actually the code is correct: `janus_session_find` adds a reference if the session is found, and so that's the reference we decrease if the session already exists. Can you elaborate more what issue you think this is fixing? Anything I'm missing, e.g., did you experience an issue yourself because of this?

richard1219:
> Actually the code is correct: `janus_session_find` adds a reference if the session is found, and so that's the reference we decrease if the session already exists. Can you elaborate more what issue you think this is fixing? Anything I'm missing, e.g., did you experience an issue yourself because of this?

There is a "janus_refcount_decrease(&session->ref)" in "goto jsondone"(line:1485) already, so, a second "decrease" here is not necessary.

ice.c
```diff
@@ -1208,6 +1207,7 @@ gint janus_ice_handle_destroy(void *core_session, janus_ice_handle *handle) {
 				g_main_loop_quit(handle->mainloop);
 			}
 		}
+        janus_refcount_decrease(&handle->ref);
```
lminiero: Indentation is broken (please use tabs, not spaces).

lminiero:
> There is a "janus_refcount_decrease(&session->ref)" in "goto jsondone"(line:1485) already, so, a second "decrease" here is not necessary.

Ah good catch! Yep, you're right indeed. I've added a note on a small nit in your other commit.

lminiero:
@richard1219 apart from the nit, can you also fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/), so that I can merge this after it's done?

lminiero:
@richard1219 any update on the fix and the CLA signature? I'd like to merge soon, as it the last missing fix to include in the next version tag. If you can't work on that, please let me know and I'll apply your fixes myself, no worries (it's just a couple of lines so copy/pasting is no big deal).

lminiero:
Considering it's been almost 10 days since you last answered, I'll assume you're busy and can't take care of this. As such, I'll incorporate your fixes myself and close this PR, and then publish the new tagged version. Thanks for the help!

PR #1502, opened 2019-01-30
Call ID in SIP events. Support for custom Call ID.

zayim:
Hi,

regarding the [question on Google group](https://groups.google.com/forum/#!topic/meetecho-janus/7d-c7lKGE4E), here is the implementation of Call IDs in the SIP (and SIPre) plugin events. I also implemented support for providing user-defined Call ID to be used across the call.

Backwards compatibility is intact, `call_id` parameter is not mandatory, if left out, behaviour is as before (random `call_id` is generated), and in SIP (and SIPre) plugin events, there is just new parameter.

I have already signed CLA.

Thank you.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -3240,6 +3240,7 @@ error:
 			json_object_set_new(event, "sip", json_string("event"));
 			json_object_set_new(event, "error_code", json_integer(error_code));
 			json_object_set_new(event, "error", json_string(error_cause));
+			json_object_set_new(event, "call_id", json_string(session->callid));
 			int ret = gateway->push_event(msg->handle, &janus_sip_plugin, msg->transaction, event, NULL);
 			JANUS_LOG(LOG_VERB, "  >> Pushing event: %d (%s)\n", ret, janus_get_api_error(ret));
 			json_decref(event);
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -157,6 +157,7 @@
 \verbatim
 {
 	"request" : "call",
+	"call_id" : "<user-defined value of Call-ID SIP header used in all SIP requests throughout the call; optional>",
 	"uri" : "<SIP URI to call; mandatory>",
 	"headers" : "<array of key/value objects, to specify custom headers to add to the SIP INVITE; optional>",
 	"srtp" : "<whether to mandate (sdes_mandatory) or offer (sdes_optional) SRTP support; optional>",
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks! I'm currently abroad, so I'll need a few days to review. I'll get back to you as soon as I can have a look.

plugins/janus_sip.c
```diff
@@ -2493,9 +2499,16 @@ static void *janus_sip_handler(void *data) {
...
 			session->status = janus_sip_call_status_inviting;
-			/* Create a random call-id */
-			char callid[24];
-			janus_sip_random_string(24, (char *)&callid);
+			char *callid;
+			json_t *request_callid = json_object_get(root, "call_id");
+			/* Take call-id from request, if it exists */
+			if (request_callid) {
+				callid = g_strdup(json_string_value(request_callid));
+			/* If call-id does not exist in request, create a random one */
```
lminiero: The comment is in the wrong place, please put it inside the `else` and properly indented (same in other plugin, I guess).

plugins/janus_sipre.c
```diff
@@ -2065,9 +2065,16 @@ static void *janus_sipre_handler(void *data) {
...
+			char *callid;
+			json_t *request_callid = json_object_get(root, "call_id");
+			/* Take call-id from request, if it exists */
+			if (request_callid) {
+				callid = g_strdup(json_string_value(request_callid));
+			/* If call-id does not exist in request, create a random one */
+			} else {
+				callid = g_malloc0(24);
+				janus_sipre_random_string(24, callid);
+			}
```
lminiero: See note above.

lminiero:
One thing that popped to mind though is that there is no check to see if the Call ID already exists: in both plugins we have a `callids` hashtable, so that will cause problems in case of conflicts. In fairness, we didn't have the check before either, but since we were generating random strings it was less likely to be an issue: with user provided values, the check should be enforced I believe.

zayim:
Thanks for the comments! I pushed fixes for memory leak and comment indentation.

Regarding possibly duplicated `callid`, what do you think behavior should be, reject the call with SIP 400 Bad Request, or just generate random `callid` instead of user provided? Or something else, maybe?

lminiero:
> Thanks for the comments! I pushed fixes for memory leak and comment indentation.

Thanks for the quick fixes!

> Regarding possibly duplicated `callid`, what do you think behavior should be, reject the call with SIP 400 Bad Request, or just generate random `callid` instead of user provided? Or something else, maybe?

Mh I was only thinking about validating the callid provided by users when originating a new call, but I hadn't thought about incoming calls with a callid that matches one we're handling already. Let's skip this for now, as I'll have to think about it: I'll merge this contribution and we can worry about the rest later.

Thanks again!

PR #1503, opened 2019-01-31
Fix infinite loop when an HTTP connection breaks

jalaziz:
The glib `g-cond-wait-until` method does not return ETIMEDOUT when a
timeout condition occurs. Instead, it returns FALSE. Expecting the
error code resulted in an infinite loop when the HTTP connection is
invalidated before a response is generated.

See https://developer.gnome.org/glib/stable/glib-Threads.html#g-cond-wait-until.

I don't believe this patch is complete yet. The wait condition is normally signaled in `janus_http_send_message`. "Invalid HTTP" errors are also handled in `janus_http_send_message`. However, in the case of an error, the wait condition is not signaled to complete. Should the condition also be signaled (with an empty response) to allow any waiters to cleanup before the timeout?

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -1484,7 +1484,7 @@ parsingdone:
 	janus_mutex_lock(&msg->wait_mutex);
 	while(!msg->got_response) {
 		int res = janus_condition_wait_until(&msg->wait_cond, &msg->wait_mutex, wakeup);
-		if(msg->got_response || res == ETIMEDOUT)
+		if(msg->got_response || !res)
 			break;
 	}
 	janus_mutex_unlock(&msg->wait_mutex);
@@ -1737,7 +1737,7 @@ parsingdone:
 	janus_mutex_lock(&msg->wait_mutex);
 	while(!msg->got_response) {
 		int res = janus_condition_wait_until(&msg->wait_cond, &msg->wait_mutex, wakeup);
-		if(msg->got_response || res == ETIMEDOUT)
+		if(msg->got_response || !res)
 			break;
 	}
 	janus_mutex_unlock(&msg->wait_mutex);


lminiero:
Mh that's weird, I think we tested scenarios where plugins would take too long to respond and IIRC it worked as expected. That said, this might have been before using conditions for that, a change that I think we did indeed make after that. I'll have a look at the patch as soon as I get back to the office, thanks!

jalaziz:
Thanks @lminiero!

We tested a different scenario where we client couldn't connect to the RTP port cause our client to disconnect. When the client disconnected, we noticed microhttpd threads piling up, each consuming almost all the CPU.

`gstack` revealed the following stack on all the threads:
```Thread 1 (Thread 0x7f39517fa700 (LWP 1274)):
#0  0x00007ffc1f1d2ca0 in clock_gettime ()
#1  0x00007f39b31b2306 in clock_gettime () from target:/lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f39b382fea7 in g_cond_wait_until () from target:/lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007f39b0126ff3 in janus_http_handler (cls=<optimized out>, connection=<optimized out>, url=<optimized out>, method=<optimized out>, version=<optimized out>, upload_data=0x0, upload_data_size=0x7f39517f9710, ptr=0x7f3970002588) at transports/janus_http.c:1486
#4  0x00007f39b00ff2b1 in ?? () from target:/lib/x86_64-linux-gnu/libmicrohttpd.so.12
#5  0x00007f39b0101028 in ?? () from target:/lib/x86_64-linux-gnu/libmicrohttpd.so.12
#6  0x00007f39b0102d2e in ?? () from target:/lib/x86_64-linux-gnu/libmicrohttpd.so.12
#7  0x00007f39b01069b2 in ?? () from target:/lib/x86_64-linux-gnu/libmicrohttpd.so.12
#8  0x00007f39b3275fa3 in start_thread () from target:/lib/x86_64-linux-gnu/libpthread.so.0
#9  0x00007f39b31a47ef in clone () from target:/lib/x86_64-linux-gnu/libc.so.6```

Tracing things to `janus_http.c:1486` and `g_cond_wait_until` revealed the issue with the error code. To be clear, the pthread-based wait returns the error code, but the glib wait does not on timeout.

Still, it seems odd to wait for the timeout in the case of an invalidated HTTP connection. However, before I fix that, I wanted to get your insights on what the correct behavior should be.

lminiero:
> Still, it seems odd to wait for the timeout in the case of an invalidated HTTP connection. However, before I fix that, I wanted to get your insights on what the correct behavior should be.

I think that changing the behaviour is something that can be addressed in a separate effort. The first thing to do is definitely to merge your patch as it fixes a broken management in the condition management: this is definitely an issue that was introduced when I switched from native thread primitives to the glib wrapper ones, so my fault for not catching the different return codes there.

Can you fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

jalaziz:
> Out of curiosity: how did you manage to use gstack/pstack ? Which distro/architecture are you running?

We're running CoreOS and Janus is running inside a Docker container. I used CoreOS's `toolbox` feature to run a privileged container, installed `procps` and `gdb`, then ran `gstack` against the thread PIDs under the Janus process tree (found with `top -H`).

lminiero:
> @lminiero signed.

Thanks, merging then! :+1:

lminiero:
Mh, now that I think about it, since we're using `janus_condition_wait_until`, this may now break in case people are NOT using GMutex (they can be disabled via configure), as our method wraps the pthread-based one in that case, which returns the value we were checking before. Probably I'll have to add a new define (e.g., `JANUS_CONDITION_FAILURE`) that is set to `FALSE` or `ETIMEDOUT` depending on what we're wrapping, and use that in the code.

jalaziz:
@lminiero I'm not sure about that. If GMutex is not being used, it uses `janus_condition_timedwait` no? That indeed returns ETIMEDOUT.

lminiero:
Yep, I edited my comment shortly after writing that, as we do check for the define in the HTTP code. Sorry for the noise!

tmatth:
Oh nice, I'm pretty sure we've hit this sporadically but couldn't identify the cause (although we suspected it was related to the http transport) and it was only being triggered when bots were trying known http attacks against us. Great fix.

PR #1505, opened 2019-01-31
To send public(nat_1_1_mapping) ip in sdp connection info when janus behind nat.

DileepNunna:


lminiero:
Sorry but this patch cannot be accepted. It forces the Janus public IP in the SDP utils, that should be unaware of core details since they're used in plugins too. See my question in the issue you opened.

PR #1506, opened 2019-02-01
Added sdpSemantics flag plan-b value until the Unified Plan SDP is used

oscarvadillog:
Yesterday **Unified Plan** SDP Phase 3 was relased:
[https://webrtc.org/web-apis/chrome/unified-plan/](https://webrtc.org/web-apis/chrome/unified-plan/)

Some demos stopped working in Janus with the latest stable version of Chrome, such as the videocall

With this PR we force M65+ browsers to  use **"plan-b"** semantics until the Unified Plan SDP is ready on Janus.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1604,6 +1604,9 @@ function Janus(gatewayCallbacks) {
 				// This is Edge, enable BUNDLE explicitly
 				pc_config.bundlePolicy = "max-bundle";
 			}
+			if(Janus.webRTCAdapter.browserDetails.browser === "chrome" && Janus.webRTCAdapter.browserDetails.version >= 65) {
+				pc_config.sdpSemantics = "plan-b";
+			}
 			Janus.log("Creating PeerConnection");
 			Janus.debug(pc_constraints);
 			config.pc = new RTCPeerConnection(pc_config, pc_constraints);


lminiero:
I'm not sure I like the change: we should see what else we need to fix in janus.js to get it working, as we're doing in #1459.

lminiero:
Besides, we have code in janus.js right now that for Chrome >= 72 uses new stuff like replaceTrack when renegotiating and replacing audio or video streams, and your patch would break those.

oscarvadillog:
I understand, it works well for me because I'm using version v0.6.0 of janus.js applying this patch.

In any case, wouldn't it have been better to add replaceTrack in the unified-plan branch? 
Have you thought about when to merge the branch? It's still too early. Isn't it?

The videocall demo doesn't work right now with Chrome >= 72 and master branch / old tags.
Personally, I think I will still not use replaceTrack, I prefer to support Chrome 72 with v0.6.0 release.

Thanks Lorenzo!

lminiero:
> In any case, wouldn't it have been better to add replaceTrack in the unified-plan branch?

It *is* there already, but that's irrelevant. The unified plan branch refactors the whole core to support multistream, the janus.js changes are related to that. The change I applied in master simply uses transceivers for the single audio/video stream master supports in order to avoid issues during renegotiations, as otherwise trying to replace a track actually adds a new one and media stops working. My guess is that the issue you refer to (which AFAIK only impacts VideoCall, though, not other plugins) can be fixed without touching the core in a similar way.

> Have you thought about when to merge the branch? It's still too early. Isn't it?

If nobody tries it and provides feedback I can't really say: one day I'll just merge ad we'll find out :wink:

lminiero:
Just verified that with #1459 the VideoCall works, even with the master version of janus.js, meaning it's something else we need to fix, and not the semantics: possibly something in the SDP. I'll try to have a look although I probably won't be able to come up with a fix now, since I'm abroad for work.

lminiero:
Mh, I think I may have found out where the problem is. In the old version of Janus, we were using `audio`,` video` and `data` as mid values, so what also ended up in the bundle attribute: the new version of Janus, and browsers too, use numeric IDs instead, like `0`, `1`, `2` and so on.

This is not a problem in terms of SDP: in fact, we adapt the mid to the offers we receive, and so do browsers. It *is* a problem with Unified Plan browsers, though, as they negotiate an RTP extension to carry the mid, namely `urn:ietf:params:rtp-hdrext:sdes:mid`. As such, in the VideoCall scenario with Janus master we end up with mids `0`, `1` and `2` on one side (the caller), and with `audio`, `video` and `data` on the other (the callee), since in the first case mids are chosen by the browser, in the other by Janus (which originates an offer towards the callee). Each browser receives RTP packets with the right SSRC, but the packets also contain the RTP extension with the original mid as set by the other party: since the browser doesn't recognize the mid (e.g., they got `video` but expected `0` or `1`) the packet is discarded.

That said, fixing the mids in Janus master is a no-go, since there are a lot of assumptions on just there being audio or video: this has already been fixed in the `unified-plan` branch. Rewriting the RTP extension is too problematic as well. What we can do, though, is rejecting the mid RTP extension entirely: this way, only SSRC will be used, and that should do the trick.

I'll check if this fixes it and keep you posted.

oscarvadillog:
Woa, amazing Lorenzo. Thanks very much!

lminiero:
Committed the fix here: https://github.com/meetecho/janus-gateway/commit/3f4960d3ed2ffa4b863ce5e0677ef7a827b46122
This seems to do the trick both in the VideoCall and the NoSIP demos (it happened there too for the same reason). Closing this PR as it's not needed anymore, but thanks for contributing it anyway!

PR #1508, opened 2019-02-04
Fixes for RTP issues discovered while fuzzing.

atoppi:
This PR adds many fixes to RTP parsing functions discovered while fuzz testing Janus.


lminiero:
Looks good to me! :+1:

PR #1511, opened 2019-02-13
janus_streaming: drop write-only audio_ssrc variable

tmatth:
This PR drops the `audio_ssrc` member of the `janus_streaming_rtp_source` struct as it never seems to be used. 

Otherwise it is intended for some future purpose it would be helpful to document that.


lminiero:
I think it was there in case in the future we add RTCP support for audio too: that's what we currently use `video_ssrc` for (PLIs towards the source).

lminiero:
Closing for the aforementioned reason, for now. We can revisit later. Thanks!

Issue #1517, opened 2019-02-19
Failure to load event handler plugin


Aidenhjj:
I've compiled Janus for Tx2 (aarch64), with [this](https://pastebin.com/QMFmGXSq) configuration.

When set:

`events: {
    broadcast = true
}`

In janus.jcfg, and:

`general: {
    enabled: true
}`

In janus.eventhandler.sampleevh.jcfg, I get:

```
[ERR] [janus.c:main:4038]   Couldn't load event handler plugin  \
    'libjanus_sampleevh.so': /opt/janus/lib/janus/events/libjanus_sampleevh.so: \
     undefined symbol: pow
```

lminiero:
Looks like a missing `-lm`. Try adding it [here](https://github.com/meetecho/janus-gateway/blob/master/Makefile.am#L307) and recompiling.

Aidenhjj:
Thanks. Great project by the way. Have added https://github.com/meetecho/janus-gateway/pull/1519 to resolve. If the pull request is badly formatted, please tell me, as I have 0 experience with github.

lminiero:
Thanks, I've added the fix myself then.

PR #1519, opened 2019-02-19
Fix missing libm linkage for events sample

Aidenhjj:
Resolution of https://github.com/meetecho/janus-gateway/issues/1517.

Following recommendation, added -lm to the right place in the Makefile.am.


lminiero:
I've already added the fix, thanks anyway!

PR #1520, opened 2019-02-19
rtp: use defines for audio and video timestamp increments

tmatth:


lminiero:
Actually the only place where they're used is in the last argument of `janus_rtp_header_update`, and that value is deprecated and never used. I'll update the signature of the method in the unified-plan branch, where there's enough changes that plugin implementors will have to update that too. It's a good idea to use a define in the AudioBridge, though, where it's used as the default number of samples: I'll update it shortly.

As such, I'm closing this PR, but thanks as usual for the contribution!

lminiero:
Correction: I already changed the signature of the method in unified-plan! That just left me with the define in the AudioBridge, that I just did.

PR #1524, opened 2019-02-25
Configurable TWCC feedback period

lminiero:
As the title says, this patch makes the Transport Wide Congestion Control RTCP feedback period configurable. By default, we send this feedback once per second, as we do for other RTCP packets like SR and RR. Since it might make sense for the TWCC feedback to be more frequent though (or less so, depending on the requirements), this adds an option both in `janus.jcfg` and the command line to do that.

Specifically, there's a new property called `twcc_period` in the `media` container of `janus.jcfg` to override it. For instance:

	media: {
		[..]
		twcc_period = 200
		[..]
	}

will set the period to 200ms (5 feedback messages per second). The same result can be achieved using the `-B` command line option, e.g.:

	/opt/janus/bin/janus -B 200

It is backwards compatible, meaning that if you don't set anything (the default) we'll keep on sending once per second exactly as before (it's pretty much the same code).

I'm planning to merge soon, so please let me know if you notice any regression.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,8 @@ or on the command line:
                                   notifies you about this
 	-r, --rtp-port-range=min-max  Port range to use for RTP/RTCP (only available
 								  if the installed libnice supports it)
+	-B, --twcc-period=number      How often (in ms) to send TWCC feedback back to
+                                  senders, if negotiated (default=1s)
 	-n, --server-name=name        Public name of this Janus instance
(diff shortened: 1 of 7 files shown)


lminiero:
Pinging @kangsik as I remember him telling they modified the code to send feedback every 200ms, and so this patch may be similar to what they did internally.

lminiero:
Merging.

PR #1525, opened 2019-02-25
Added option to negotiate inband FEC for Opus in VideoRoom and EchoTest

lminiero:
As the title suggests, very simple change to the SDP utils to allow providing custom fmtp attributes in negotiation. This is used in this patch to negotiate `useinbandfec=1` in both EchoTest and VideoRoom.

For the VideoRoom, this is tied to a new `opus_fec=true` property when creating rooms (default is still FALSE, for backwards compatibility). When set to TRUE, if publishers are negotiating inband FEC for Opus, we set it in the answer as well, and in all the offers we send to their subscribers. Anyway, notice that in principle the amount of redundancy to add to Opus packets is derived from the lost packets that are notified back: in the current implementation, this would be the lost packets on the publisher "leg", but in a one-to-many scenario it makes more sense to convey the lost packets on the subscribers' side instead. This is not envisaged in this pull request, and would need to be done in a later stage.

Planning to merge soon, so please stop me before that if you have questions or doubts.

### DIFF
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -958,7 +958,7 @@ janus_sdp *janus_sdp_generate_offer(const char *name, const char *address, ...) 
 	gboolean do_audio = TRUE, do_video = TRUE, do_data = TRUE,
 		audio_dtmf = FALSE, video_rtcpfb = TRUE, h264_fmtp = TRUE,
 		data_legacy = TRUE;
-	const char *audio_codec = NULL, *video_codec = NULL, *audio_fmtp;
+	const char *audio_codec = NULL, *video_codec = NULL, *audio_fmtp = NULL;
 	int audio_pt = 111, video_pt = 96;
 	janus_sdp_mdirection audio_dir = JANUS_SDP_SENDRECV, video_dir = JANUS_SDP_SENDRECV;
 	int property = va_arg(args, int);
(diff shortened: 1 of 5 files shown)


sdp-utils.c
```diff
@@ -1033,6 +1035,11 @@ janus_sdp *janus_sdp_generate_offer(const char *name, const char *address, ...)
 			janus_sdp_attribute *a = janus_sdp_attribute_create("rtpmap", "%d %s", dtmf_pt, janus_sdp_get_codec_rtpmap("dtmf"));
 			m->attributes = g_list_append(m->attributes, a);
 		}
+		/* Check if there's a custom fmtp line to add for audio */
+		if(audio_fmtp) {
```
tmatth: This read may be of an uninitialized `audio_fmtp`, it should set it to NULL as is done for `audio_codec` and `video_codec`.

sdp-utils.c
```diff
@@ -1095,7 +1102,7 @@ janus_sdp *janus_sdp_generate_answer(janus_sdp *offer, ...) {
 	/* Let's see what we should do with the media */
 	gboolean do_audio = TRUE, do_video = TRUE, do_data = TRUE,
 		audio_dtmf = FALSE, video_rtcpfb = TRUE, h264_fmtp = TRUE;
-	const char *audio_codec = NULL, *video_codec = NULL;
+	const char *audio_codec = NULL, *video_codec = NULL, *audio_fmtp;
```
tmatth: This `audio_fmtp` should also be initialized to NULL.

sdp-utils.c
lminiero: Good catch, thanks! Fixing now :+1:

PR #1542, opened 2019-03-11
fuzzers: make jobs and workers configurable via environment

tmatth:


tmatth:
@atoppi this can be specified with `WORKERS=M JOBS=n ./run.sh`

fuzzers/run.sh
```diff
@@ -3,6 +3,8 @@
 set -eu
 
 TARGET=${1-"rtcp_fuzzer"}
+JOBS=${JOBS:-4}
+WORKERS=${WORKERS:-4}
```
lminiero: I'll let @atoppi comment on the PR in general, but why is the value `-4` there? Shouldn't it be `4`?

tmatth: @lminiero in bash, that's the syntax to specify default values if missing, it's dash 4, not minus 4.

lminiero: Thanks for the clarification, I'm an idiot :smile:

tmatth: @lminiero not at all, I have to look this up every time ;)

atoppi:
@tmatth Thanks for the PR.
I think JOBS and WORKERS should default to a `nproc` related value.

tmatth:
@atoppi updated

fuzzers/run.sh
```diff
@@ -3,8 +3,9 @@
 set -eu
 
 TARGET=${1:-"rtcp_fuzzer"}
-JOBS=${JOBS:-4}
-WORKERS=${WORKERS:-4}
+HALF_NCORES=$(expr $(nproc) / 2)
```
atoppi: To avoid zero values for single core machines:
```
if [[ $HALF_NCORES -eq 0 ]]; then
 HALF_NCORES=1
fi
```

atoppi:
@tmatth looks good to me, merging.

PR #1543, opened 2019-03-12
Support for mid RTP extension, and better extmap negotiation in SDP utils

lminiero:
In Janus master, we currently reject the negotiation of the mid RTP extension, that is `urn:ietf:params:rtp-hdrext:sdes:mid`. The reason for that is that, when negotiated, browsers use that to verify which track the packet is for, rather than the SSRC. Normally, this is fine, but if the mid in the packet doesn't match any in the SDP, the packet is silently discarded: considering the plugin-based architecture we have, this can happen in a few scenarios, e.g., in the VideoCall where the caller may use `0` and `1` as mid (set by the browser) while the callee may use `audio` and `video` (set by Janus); even using numbers as browsers do, as we now do in the unified-plan branch, this can be an issue is we bridge streams in ways that result in a mid mismatch (e.g., a VideoRoom publisher packet with mid=0, may be sent on a VideoRoom subscriber track with mid=2). For this reason, until now we decided to simply drop that RTP extension, in order to avoid issues.

That said, recent developments in the spec made it clear that the mid RTP extension will need to be supported. In order to start patching this behaviour, we now do negotiate the mid RTP extension for incoming offers, and remove it for our own offers instead: this way, for PeerConnections where the user is the initiator (e.g., VideoRoom publishers), the extension will be negotiated and usable, while for PeerConnections where Janus offers instead (e.g., VideoRoom subscribers) we make sure the extension is not negotiated instead. This will allow us to make use of that info in some cases, and avoid the issues I presented in the previous paragraph.

As a side note, I took advantage of this changes to also streamline the way we negotiate RTP extensions via SDP utils as well. Some plugins (EchoTest, AudioBridge, VideoRoom), for instance did some awkward and manual manipulation of the SDP, that I now made much more easier. What I added was basically the possibility to specify which extensions should be accepted when generating an answer, and the SDP utils will do the rest: for offers, instead, it still is a manual process, since IDs would need to be specified as well. Here's an example of how this works in the EchoTest:

	janus_sdp *answer = janus_sdp_generate_answer(offer,
		JANUS_SDP_OA_AUDIO_CODEC, json_string_value(audiocodec),
		JANUS_SDP_OA_AUDIO_FMTP, opus_fec ? "useinbandfec=1" : NULL,
		JANUS_SDP_OA_VIDEO_CODEC, json_string_value(videocodec),
		JANUS_SDP_OA_ACCEPT_EXTMAP, JANUS_RTP_EXTMAP_MID,
		JANUS_SDP_OA_ACCEPT_EXTMAP, JANUS_RTP_EXTMAP_RTP_STREAM_ID,
		JANUS_SDP_OA_DONE);

As you can see, we're passing two `JANUS_SDP_OA_ACCEPT_EXTMAP` flags that tell the SDP utils to accept the provided RTP extensions, in this case `JANUS_RTP_EXTMAP_MID` and `JANUS_RTP_EXTMAP_RTP_STREAM_ID`: those two values are strings defined in `rtp.h`, but you can provide some custom ones as well.

I plan to merge soon, so feedback welcome!


lminiero:
Merging.

PR #1545, opened 2019-03-13
Fix and enhance RTCP stats and link quality metrics.

atoppi:
This PR heavily refactors the way Janus calculates some RTCP stats, estimate link metrics and generate receiver reports.

- Before this PR the lost count was basically incremented before sending the first NACK for a packet.
Now the calculation happens whenever a RTCP Receiver Report is being sent, that is once in a second.
The cumulative loss is now obtained as the accumulation of the _packets lost_ in every 1 second time window.
That is, for every RR being sent:
```
lost_interval = expected_interval - received_interval
cumulative_loss += lost_interval
```

- The `expected` packets semantics has not been changed and is calculated from the highest sequence number in the stream.

- The `received` packet counter is now incremented upon the reception of one of the followings:

1. in-order RTP media packet
2. out-of-order RTP media packet for RFC4588 enabled RTP streams
3. out-of-order, not too late (*), RTP media packet for non-RFC4588-enabled and NACKs-enabled RTP streams

- The `retransmitted` packet counter  is now incremented upon the reception of one of the followings:

1. RTP RFC4588 packet
2. out-of-order, too late (*), RTP media packet for non-RFC4588-enabled and NACKs-enabled RTP streams

(*) The late evaluation is done with a heuristic based on the difference between the last in-order RTP timestamp and the out-of-order packet RTP timestamp

- Jitter is now calculated only for in-order packet.

- In link quality estimation has been fixed in order to consider only the difference between `expected` and `received`

- In media link quality has been fixed in order to consider also the `retransmitted` packet.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -2453,12 +2453,16 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 
 				/* Update the RTCP context as well */
 				rtcp_context *rtcp_ctx = video ? stream->video_rtcp_ctx[vindex] : stream->audio_rtcp_ctx;
-				gboolean count_lost = ((!video && !component->do_audio_nacks) || (video && !component->do_video_nacks)) ? TRUE : FALSE;
-				janus_rtcp_process_incoming_rtp(rtcp_ctx, buf, buflen, count_lost);
+				gboolean retransmissions_disabled = (!video && !component->do_audio_nacks) || (video && !component->do_video_nacks);
+				janus_rtcp_process_incoming_rtp(rtcp_ctx, buf, buflen,
+						(video && rtx) ? TRUE : FALSE,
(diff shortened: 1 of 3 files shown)


lminiero:
Pinging @pitkonst as this probably has an impact on how he originally conceived the media/link quality contribution.

pitkonst:
Thank you for maintaining this feature. We have experienced problems with measurements in new versions of Janus. I'll check and test this changes ASAP.

pitkonst:
I've checked inbound link quality code and evereything is fine. Tests shows that mesurments are accurate. `received` and `retransmitted` semantics changed since my initial commit, and there were problems in calculations. And this PR fixes this. @atoppi, thank you for this one.

lminiero:
Looks good, then, merging :+1:

PR #1547, opened 2019-03-14
Better support for rid-based simulcasting

lminiero:
Considering that, in the transition to unified-plan, Chrome is slowly starting to support rid-based simulcasting support, this patch aims at improving the way we handle that. In fact, while Firefox already does rid-based simulcasting and we handle that already, Chrome advertises no SSRC when doing that, which broke our code. This effort is not aimed at fixing this issue altogether (rid-based simulcast is currently partly broken in Chrome as well, so it's hard to test), but to make that process much easier: now we don't rely on knowing SSRCs in advance when rids are used, and we find them out from RTP extensions containing those rid instead.

Planning to merge quite soon (I only need to update a couple more plugins, which I'll do tomorrow), so if you want to stop me because you found regression you better hurry :wink:


lminiero:
Delayed the merging process for a bit as I had to fix a couple more things. More specifically, I've also added support for the repaired RTP stream ID, in order to make sure we can recognize rtx packets properly as well when SSRCs are missing. In order to avoid issues in bidirectional PeerConnections, when a packet is recovered we don't only restore the original RTP header, but also replace the rid-related extension ID with the regular rid extension ID instead.

Before merging I plan to see if I can integrate this new approach in the SIP/SIPre/NoSIP plugins as well.

lminiero:
Merging.

PR #1549, opened 2019-03-16
1522 fix null session

naeem:
Fix for https://github.com/meetecho/janus-gateway/issues/1522


lminiero:
I'm not exactly sure how that fixes anything, actually. Checking if `session` is valid may fix that specific call, but we dereference the same pointer a lot in the following lines. Are you sure the issue you experienced wasn't related to something different?

naeem:
Unfortunately we can't re-produce the problem as it doesn't happen consistently (could be an issue in our app)
However, It definitely fixed our issue, and also looking at the ASAN [error log](https://gist.github.com/naeem/011ce9a306887d225c37dac45ef42a4b) verifies where it fails.
When it was crashing the same line number occurred in the logs each time.

lminiero:
But that line only tells you that `session->stack->s_nh_r` causes the crash, not that it's `session` or `session->stack`, and you only check the former. If it was `session->stack` it would still crash, and in both cases it will crash later anyway if `send_register` is TRUE, as both are dereferenced in that branch (`session` also in the `else` case). As such, I don't think this really fixes anything: it looks more like a way not to make it crash on that specific line, and my guess is that if you had an intermittent issue related to that, it would still crash for you, just later.

lminiero:
Closing for the reasons stated above. I'll think of a better fix in the next few days. Thanks for contributing anyway!

PR #1551, opened 2019-03-19
Support for multiple datachannel streams in the same PeerConnection

lminiero:
Historically, in Janus we've always supported a single stream in a PeerConnection datachannel: this stream was called "JanusDataChannel" and had to be created by the client, with Janus simply reacting to this; then, each message exchanged by Janus and client would be sent over this stream.

Of course, while simple and effective in most scenarios, this could be limiting in others, e.g., in custom plugins where a more sophisticated approach might be needed. This is definitely the case for the unified-plan version of the VideoRoom plugin, where a single PeerConnection can stream multiple publishers at the same time, and so being limited to a single stream would basically break datachannel broadcasting (when I receive a datachannel message, is it coming from Bob or Alice?).

This PR aims at fixing this, and adds multichannel support to our datachannel implementation. The feature is quite simple, and based on labels: in JavaScript we extended the existing API in a backwards compatible way (more on that below), while plugins are now notified from which label a new message originated in `incoming_data`, and at the same time plugins can specify a custom label that should be used for sending messages via `relay_data`. In case the label doesn't exist, a new stream is created automatically. Plugins are free to pass `NULL` as the label in `relay_data`: this will use the default stream that we've been used since the beginning, and so should ensure backwards compatibility.

As I was saying, we slightly changed the JavaScript API to support this new feature. Specifically, the two existing callbacks simply have a new `label` property in the signature:

    ondataopen() ---> ondataopen(label)
    ondata(text) ---> ondata(text, label)

This way, you know which label a new available stream is using, and you also know which label originated an incoming message. You're free to ignore those if you won't use the feature. To send a message using a custom label, instead, just add a `label` property to a `data` request, e.g.:

	echotest.data({
		text: "hello",
		label: "mylabel", <--- custom stream
		error: function(reason) { bootbox.alert(reason); },
		success: function() { $('#datasend').val(''); },
	});

In case a stream with that label doesn't exist yet, it will be created for you.

Tested briefly and it seems to be working as expected. Unless I find breaking bugs or relevant regressions, I plan to merge soon, so please let me know if think there's anything I should fix.

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -955,10 +955,10 @@ int janus_dtls_verify_callback(int preverify_ok, X509_STORE_CTX *ctx) {
 }
 
 #ifdef HAVE_SCTP
-void janus_dtls_wrap_sctp_data(janus_dtls_srtp *dtls, char *buf, int len) {
+void janus_dtls_wrap_sctp_data(janus_dtls_srtp *dtls, char *label, char *buf, int len) {
 	if(dtls == NULL || !dtls->ready || dtls->sctp == NULL || buf == NULL || len < 1)
 		return;
-	janus_sctp_send_data(dtls->sctp, buf, len);
+	janus_sctp_send_data(dtls->sctp, label, buf, len);
 }
 
 int janus_dtls_send_sctp_data(janus_dtls_srtp *dtls, char *buf, int len) {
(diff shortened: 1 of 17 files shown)


lminiero:
Merging.

PR #1567, opened 2019-04-01
audio/videoTransceiver direction property error in Safari

alienpavlov:
I've got an error `Unhandled Promise Rejection: TypeError: Attempted to assign to readonly property.
` when **.direction** property was set in Safari (both mac and iOS). But I found that Safari has `.setDirection()` setter method, so I used it to set a direction if available

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -2412,24 +2412,40 @@ function Janus(gatewayCallbacks) {
 			if(!audioSend && !audioRecv) {
 				// Audio disabled: have we removed it?
 				if(media.removeAudio && audioTransceiver) {
-					audioTransceiver.direction = "inactive";
+					if (audioTransceiver.setDirection) {
+						audioTransceiver.setDirection("inactive");
+					} else {
+						audioTransceiver.direction = "inactive";
+					}
 					Janus.log("Setting audio transceiver to inactive:", audioTransceiver);
 				}
 			} else {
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for this contribution! Have you checked if this has an impact on Chrome and/or Firefox?

alienpavlov:
> Thanks for this contribution! Have you checked if this has an impact on Chrome and/or Firefox?

yes, also checked in Microsoft Edge

lminiero:
Excellent, merging then :+1:

lminiero:
Mh this seems not to work with my Safari Technology Preview, which may be a bit older though. Which version did you test this against?

alienpavlov:
> Mh this seems not to work with my Safari Technology Preview, which may be a bit older though. Which version did you test this against?

I was able to reproduce it in Safari Version 12.1 (14607.1.40.1.4) and on iOS 12.1.4 (not sure what exact version of Safari)

Do you mean your Safari Technology Preview supports writing to RTCRtpTransceiver **.direction** property and you don't see **.setDirection()** method? or you got an error with the code I commited?

Here is some specification about it https://www.w3.org/TR/2016/WD-webrtc-20160531/#rtcrtptransceiver-interface but actually I found this method through javascript console in Safari - so this was a hint for me to use this setter method instead of assigning the .direction property

PR #1577, opened 2019-04-10
janus_streaming: guard against NULL rtpmaps

tmatth:
Reported by Ilmars Lacplesis


lminiero:
Looks good, thanks! I'll think if it might make sense to implement more generic fences and checks, as that indeed shouldn't happen. Merging!

PR #1579, opened 2019-04-11
Update README.md

najd-mrabet:


lminiero:
I don't think that's needed, as the libnice installation makes it clear when you configure there, and I don't want to list all the dependencies for all dependencies :slightly_smiling_face: 
As such, thanks for the contribution, but closing without merging.

PR #1581, opened 2019-04-12
Ignore encrypted extensions when processing SDP (fixes #1575).

### LINKED ISSUE #1575: Wrong extension id for encrypted headers
In some cases, Janus is sending an answer with a wrong extension id if encrypted extension headers are used. You can test this by enabling the flag "Negotiation with encrypted header extensions for SRTP in WebRTC" in `chrome://flags`.

Offer from Chrome:
```
...
m=video 9 UDP/TLS/RTP/SAVPF 96 97 98 99 100 101 102 122 127 121 125 107 108 109 124 120 123
...
a=extmap:2 urn:ietf:params:rtp-hdrext:toffset
a=extmap:3 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:4 urn:3gpp:video-orientation
a=extmap:5 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:6 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=extmap:7 http://www.webrtc.org/experiments/rtp-hdrext/video-content-type
a=extmap:8 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
a=extmap:10 http://tools.ietf.org/html/draft-ietf-avtext-framemarking-07
a=extmap:12 http://www.webrtc.org/experiments/rtp-hdrext/color-space
a=extmap:13 urn:ietf:params:rtp-hdrext:encrypt urn:ietf:params:rtp-hdrext:toffset
a=extmap:11 urn:ietf:params:rtp-hdrext:encrypt urn:3gpp:video-orientation
a=extmap:9 urn:ietf:params:rtp-hdrext:encrypt http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:0 urn:ietf:params:rtp-hdrext:encrypt http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=extmap:0 urn:ietf:params:rtp-hdrext:encrypt http://www.webrtc.org/experiments/rtp-hdrext/video-content-type
...
```

Answer from Janus:
```
...
m=video 9 UDP/TLS/RTP/SAVPF 96
...
a=extmap:11 urn:3gpp:video-orientation
a=extmap:0 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
...
```

Note that the extensions got mapped to the id of the encrypted version from the offer (that has the `urn:ietf:params:rtp-hdrext:encrypt` as prefix before the actual extension.

The problem is caused by the code that checks if the extension name is part of the m-line attribute in https://github.com/meetecho/janus-gateway/blob/ce8255e56b4a1cecc992cd55633bcdb1ef88cfe7/plugins/janus_videoroom.c#L5893

In this case this matches both the unencrypted and the encrypted extension and the last matching id (here the encrypted) is used.
(On a side note, the id `0` means that an unused number should be assigned in the answer, so replying with `extmap:0` is invalid, too).

When setting the answer, Chrome fails with "Failed to set remote answer sdp: Failed to set remote video description send parameters.".

fancycode:


lminiero:
Thanks! But I think this needs to be extended to a few more plugins than just the VideoRoom. The AudioBridge also accesses extensions to check the one used for audio levels, for instance.

That said, it might be easier to just drop the encrypted extensions attributes from the SDP when doing the anonymize process: this way, plugins never get to see them at all, and nothing breaks. I'll have to check if the same issue can happen in the core too, though, as we do traverse extensions looking for mid and rid for instance.

lminiero:
Implemented as I said, by basically removing the encrypted extensions before passing the SDP to plugins, and viceversa. Verified that it seems to be working when I tested. I borrowed your definition in `rtp.h` since it was indeed useful, and we'll need it if/when we'll add proper support for encrypted extensions. Closing your PR, but thanks again for your precious help!

Issue #1583, opened 2019-04-12
n.setDirection is not a function. (In 'n.setDirection(\"sendonly\")', 'n.setDirection' is undefined) on Safari 12.1


Tobiasz-Rumian:
When connecting to Janus from Safari 12.1 the bellow errors appears:

`[Error] WebRTC error: – {code: undefined, name: "TypeError", message: "n.setDirection is not a function. (In 'n.setDirection(\"sendonly\")', 'n.setDirection' is undefined)"}
	error
	error (videoroomtest.js:419)
	(anonymous function) (janus.js:2307)
	promiseReactionJob`

`[Error] WebRTC error: – "enumerateDevices error"
	error
	error (videoroomtest.js:419)
	(anonymous function) (janus.js:2313)
	promiseReactionJob`

After that Janus is not receiving any streams from Safari (other clients can't see Safari) but Safari is able to see all other clients.
This errors only appears on Safari (I didn't test it on older versions). I've tested it on Chrome 73 and Firefox 67 but the errors doesn't appear there.

lminiero:
Yeah, it looks like it's related to the fix @alienpavlov made in #1567. You don't see it on Chrome and Firefox because it's specific to Safari. Have you checked if reversing that PR fixes it for you?

Tobiasz-Rumian:
Unfortunately reversing that PR doesn't help. Now there is no `setDirection` in code but the error is still the same. It looks like some internal function is failing but it makes no sens to me.

lminiero:
Have you checked if it isn't a caching issue?

Tobiasz-Rumian:
Yep. Twice.

lminiero:
Then no idea, I'm not a Mac OS or Safari guy :slightly_smiling_face: 
I'll try to have a look later with the old machine we have to see if can spot the reason. If you manage to get it working in the meanwhile, pull requests are always welcome of course!

Tobiasz-Rumian:
I found that the error is in implementation of `createOffer` (line 2532 `config.pc.createOffer()`) on Safari. 
When you pass media constraints to it the implementation fails. With no constraints the connection is working fine but I think they were there for a reason so it's not a proper fix.

Tobiasz-Rumian:
#1584

alienpavlov:
[...]

PR #1584, opened 2019-04-12
Use createOffer with no constraints if browser is safari.

Tobiasz-Rumian:
Use createOffer with no constraints if browser is safari. This is due to Safari not being able to handle media constraints.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -2529,8 +2529,13 @@ function Janus(gatewayCallbacks) {
 				sender.setParameters(parameters);
 			}
 		}
-		config.pc.createOffer(mediaConstraints)
-			.then(function(offer) {
+		var offer;
+		if(Janus.webRTCAdapter.browserDetails.browser !== "safari"){
+			offer = config.pc.createOffer(mediaConstraints)
+		}else{
+			offer = config.pc.createOffer()
+		}
+		offer.then(function(offer) {
 				Janus.debug(offer);
 				Janus.log("Setting local description");
 				if(sendVideo && simulcast) {


lminiero:
Thanks! Can you fill in the form for our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

Tobiasz-Rumian:
Done

lminiero:
Good catch. Maybe something like this instead?

```
var constraints;
if(Janus.webRTCAdapter.browserDetails.browser !== "safari")
	constraints = mediaConstraints;
config.pc.createOffer(constraints)
			.then(function(offer) {
```

Anyway, I'm not convinced that Safari doesn't support constraints. Iñaki from mediasoup for instance confirmed that you do need them in some cases (e.g., ICE restarts), which means that just never passing constraints would simply break things. My guess is that we should just avoid setting any _direction_ constraint here, namely `offerToReceiveAudio` and `offerToReceiveVideo`.

lminiero:
Closing this PR then, I'll work on a different fix.

mtltechtemp:
> Good catch. Maybe something like this instead?
> 
> ```
> var constraints;
> if(Janus.webRTCAdapter.browserDetails.browser !== "safari")
> 	constraints = mediaConstraints;
> config.pc.createOffer(constraints)
> 			.then(function(offer) {
> ```
> Anyway, I'm not convinced that Safari doesn't support constraints. Iñaki from mediasoup for instance confirmed that you do need them in some cases (e.g., ICE restarts), which means that just never passing constraints would simply break things. My guess is that we should just avoid setting any _direction_ constraint here, namely `offerToReceiveAudio` and `offerToReceiveVideo`.

I believe we should use `getSupportedConstraints()` instead of just passing nothing...

lminiero:
See discussion in #1583, I'm already working on a fix.

PR #1587, opened 2019-04-13
allow selecting h264 pt variant with maximum profile level when answering

itaibenari:
This allows to automatically answer with the offered h264 pt with the highest profile-level specification rather than with the lowest.
It also allows for getting all pt's for a specific codec found in an sdp.

### DIFF
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -573,11 +573,28 @@ int janus_sdp_remove_payload_type(janus_sdp *sdp, int pt) {
 }
 
 int janus_sdp_get_codec_pt(janus_sdp *sdp, const char *codec) {
-	if(sdp == NULL || codec == NULL)
+	GList * pt_list = NULL;
+	int pt = 0;
+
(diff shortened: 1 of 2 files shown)


lminiero:
Sorry but I can't merge this. Browsers are only mandated to support the baseline profile, everything else is optional: your patch could result in completely broken sessions if what the offerer sends the recipient doesn't support. If you want to provide a different patch that allows to specify which profile should be accepted, instead of the behaviour we're following now, I'd be happy to review that.

lminiero:
Closing for the above mentioned reason. Thanks for the contribution anyway, and looking forward to a different take on that!

itaibenari:
@lminiero this is why this is behind a flag defaulting to FALSE. the scenario targeted is where you have a server side recipient  that can decode whatever is coming. the same argument may be given for the automatic selection of codecs (where the recipient may not support vp9 for example).
btw, i think there are other nice things in the PR that you should consider adding anyway.
thanks

Issue #1591, opened 2019-04-17
Sip plugin, Re-Invite from remote part and Chrome DTMF


StephaneG31:
Hi,
I will try to explain a problem in our use case.
We use Janus just as a SIP Gateway to join some conference call.
So we need to use DTMF.
With Firefox it works, DTMF also.

With Chrome, DTMF are ignored or not received by our audio conf call bridge.

UserAgent <=> Janus <=> SBC....SBC <=> AudioBridge Application Server

When negotation SDP, Chrome for DTMF send 
`a=rtpmap:126 telephone-event/8000`

Audio bridge accepts. But Audio Bridge workflow are configured for sending re-INVITE to "renegociate" some SDP, and audio bridges and send in the re-INVITE packet
`a=rtpmap:101 telephone-event/8000`

It's the same SDP type but payload type ID is different (we don't know why the audio bridge of our partner work like this and we can't change this).

The re-INVITE is received by Janus and Janus send juste un ACCEPT message without forwarding the re-INVITE to browser to "renegociate".

So after that Chrome keeps initial INVITE, so DTMF are sent with payload type 126, and we think (but we didn't make a capture on our mutulized SBC) that SBC will drop this kind of packet which are not ACCEPT in any INVITE or RE-INVITE.

I think in the code the interesting lines are from here :
https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_sip.c#L3498

Maybe if one of SDP in a re-Invite message is different from initial negociation, Janus must forward request to the browser ?

I think Janus detect SDP changed with the `changed `variable https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_sip.c#L3481, but this one seems to not be use after

lminiero:
I don't remember exactly how we integrated the renegotiation support in the SIP plugin, but we do forward the SDP if there are media changes IIRC (e.g., you add or remove audio/video), and don't otherwise (e.g., for onhold we don't). It's very likely that this change in payload type doesn't trigger that renegotiation towards the browser, and I honestly don't know if the browser would accept the renegotiation if we did.

I'll add to the TODO list the fact we have to check what's happening there. In the meanwhile, there are a couple of things you can do to mitigate the issue until this is solved:

1. with the SIP plugin you can send DTMF tones via SIP INFO as well; if your bridge supports them, you can use those as a fallback for the time being;
[...]

PR #1597, opened 2019-04-24
Added proper management of incoming re-INVITEs to SIP plugin (see #1591)

lminiero:
This patch attempts to fix the issue described in #1591, by basically introducing proper support of incoming re-INVITEs in the SIP plugin. In fact, when we added support for renegotiations in #1099 we modified the SIP plugin to be able to send re-INVITEs, while we still handled incoming re-INVITEs the same way we did before: sending a 200 OK without involving the client. Of course, while this _kinda_ worked for scenarios like onhold, it broke scenarios like the one described in #1591 (changing the payload type for DTMF tones), and probably more.

The patch fixes it by introducing a new property you can set when sending a `call` or `accept`, named `autoaccept_reinvites`. By default, if you don't pass it, it will be `TRUE`, meaning that the SIP plugin will keep on behaving as it always has: this is to ensure backwards compatibility with clients written before this PR. If set to `FALSE`, instead, the plugin will forward the new SDP to the client, and expect an answer to send back: this explains why we default to `TRUE` (not answering an incoming re-INVITE, which is what "older" clients would do, would break the session). The event to signal the re-INVITE is called `updatingcall`, and will include a JSEP with the new offer pretty much as `incomingcall` does; as a client, you're expected to send your answer back with an `update` request. In case you implemented active renegotiations before, you may have noticed that `update` is also the request you use to send a re-INVITE yourself: I basically just updated the method to do both, meaning that if you send an offer it will be an INVITE, while if you send an answer it will be a 200 OK instead.

I tested this with a simple SIPp-based UAS that triggers a renegotiation to change the payload type (as, again, #1591 was what brought me here), and it seems to be working as expected (Chrome does change the payload type for outgoing DTMF tones after the re-INVITE). Of course, I'd love some feedback to be sure this works as expected in other scenarios as well, and that most importantly it doesn't break anything that was working before. If I don't get any feedback, I'll just assume it works for you guys and merge.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1954,6 +1954,7 @@ function Janus(gatewayCallbacks) {
 			// If we're updating and keeping all tracks, let's skip the getUserMedia part
 			if((isAudioSendEnabled(media) && media.keepAudio) &&
 					(isVideoSendEnabled(media) && media.keepVideo)) {
+				pluginHandle.consentDialog(false);
 				streamsDone(handleId, jsep, media, callbacks, config.myStream);
 				return;
 			}
@@ -2041,6 +2042,7 @@ function Janus(gatewayCallbacks) {
 			}
 			// Skip the getUserMedia part
 			config.streamExternal = true;
+			pluginHandle.consentDialog(false);
 			streamsDone(handleId, jsep, media, callbacks, stream);
 			return;
 		}
@@ -2296,6 +2298,7 @@ function Janus(gatewayCallbacks) {
 					};
 					Janus.debug("getUserMedia constraints", gumConstraints);
 					if (!gumConstraints.audio && !gumConstraints.video) {
+						pluginHandle.consentDialog(false);
 						streamsDone(handleId, jsep, media, callbacks, stream);
 					} else {
 						navigator.mediaDevices.getUserMedia(gumConstraints)
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -127,7 +127,7 @@ $(document).ready(function() {
 									Janus.debug("Consent dialog should be " + (on ? "on" : "off") + " now");
 									if(on) {
 										// Darken screen and show hint
-										$.blockUI({ 
+										$.blockUI({
 											message: '<div><img src="up_arrow.png"/></div>',
 											css: {
 												border: 'none',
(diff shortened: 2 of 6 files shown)


GerardM22:
Hi Lorenzo,

I'm Gerard, the developer who is working with StephaneG31 on the implementation of a Janus SIP Gateway for conference calls.

I tried your patch for transmitting the reinvite to the browser but I encountered some problems when using the SIP Gateway demo.

When I click on the "Call" button, I receive an error popup "Missing SDP update" and the communication is closed. I found the same error in the Janus logs:
[Thu Apr 25 17:57:50 2019] [ERR] [plugins/janus_sip.c:janus_sip_handler:2791] Missing SDP update

The client receives the reinvite message but it seems that something is missing in the response.
It seems that Janus expected a parameter "update" : true in the jsep element of the update request. I tried to add it in siptest.js but I had the same error.

When removing this test by commenting lines 2790 to 2795 in janus_sip.c, it works fine: DTMF are OK with Chrome.
Can you explain what this test is for and why it doesn't work in my case ? Can I remove it safely ? Have I forgotten something on the client side ? 

Thank you
Gerard

lminiero:
The "update" property is set by the core when it detects a renegotiation. It's only set if the PeerConnection exists, though, which makes me think your SIP device is sending a re-INVITE too soon, before the WebRTC negotiation (ICE+DTLS) is completed. We may have to add a new flag in the core to know when we're renegotiating before the setup is complete, even though I suspect that will introduce other issues.

lminiero:
@GerardM22 can you check again with this commit?

GerardM22:
Sorry but it still doesn't work after your last commit.
I send the log file of my last attempt.

[janus_test_1597.log](https://github.com/meetecho/janus-gateway/files/3121549/janus_test_1597.log)
 commit

lminiero:
Yeah my bad, the new offer from the plugin (the re-INVITE) resets the `got-answer` flag, so my check was silly... I hoped I could avoid a new flag, but I guess we'll need it.

lminiero:
@GerardM22 hopefully this new commit does the trick.

GerardM22:
Yes ! it works !
Thank you very much for your reactivity !

lminiero:
Excellent, merging then!

PR #1599, opened 2019-04-26
Add an interception callback to js API

carlcc:
I want to add NACK support for audio publisher, so I added this callback to modify sdp before pass it to the webrtc engine. I think a number of users may need it!

The new version is compatible with the formal version, one can ignore this callback if he does not need it.


lminiero:
Useful feature for SDP munging, thanks! Although I must say I don't like the name you chose for the callback much... details, anyway, we can think of a better one before merging :wink: It should also be documented in `mainpage.dox`. Besides, please notice the indentation is broken in one of the `jsep` objects you create.

That said, please fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this when it's ready.

lminiero:
Thinking about it, why not make that method part of the createOffer/createAnswer properties, instead of being global? It makes more sense that way, I think, e.g.:

	echotest.createOffer(
		{
			media: { audio: true, video: true },
			mungeSdp: function(jsep) {
					//
			},
			success: function(jsep) {
					//
			}
		}

carlcc:
> Thinking about it, why not make that method part of the createOffer/createAnswer properties, instead of being global? It makes more sense that way, I think, e.g.:
> 
> ```
> echotest.createOffer(
> 	{
> 		media: { audio: true, video: true },
> 		mungeSdp: function(jsep) {
> 				//
> 		},
> 		success: function(jsep) {
> 				//
> 		}
> 	}
> ```

Yep, this is exactly what I mean, I think the only difference between us is callback name.

Honestly speeking, I don't like the word 'munge'... which reminds me my hair losing... and no munge.. ;-( Any way， Ithink 'munge' is kind of obscure. What about using a simple name 'customizeSdp'?

lminiero:
That's a better name indeed! Anyway, you currently set the callback globally, when creating the session, while it's only invoked by janus.js when doing a createOffer/answer. I think it should only be set when calling those methods instead, so that it's specific to that specific call, otherwise you lose the flexibility of customizing the SDP differently depending on the PeerConnection you're creating.

carlcc:
> That's a better name indeed! Anyway, you currently set the callback globally, when creating the session, while it's only invoked by janus.js when doing a createOffer/answer. I think it should only be set when calling those methods instead, so that it's specific to that specific call, otherwise you lose the flexibility of customizing the SDP differently depending on the PeerConnection you're creating.

It seems to be a problem with the github diff tool.

Actually, the callback is added under function `createOffer` and `createAnswer`, the diff tool shows that the callback is under function `Janus`. I guess it is because both `createOffer` and `createAnswer` function are embbeded under function `Janus`. You can expand the code to check that.

lminiero:
Ah thanks for the clarification. I'll try do so a better review later.

lminiero:
PS: we still need you to sign the CLA before this can be merged though.

lminiero:
Thanks, merging then!

Issue #1600, opened 2019-04-27
Crash in janus_ice_outgoing_traffic_handle


amnonbb:
#1459 
For me it's happened something like three times in last month ( libnice 0.1.14.1 )
https://pastebin.com/2xxAApCU

With last master it's occur more often.

lminiero:
Please provide a libasan dump.

lminiero:
Just noticed you mentioned libnice 0.1.14, have you tried master instead? Pretty sure libnice 0.1.14 precedes many fixes they made after that.

lminiero:
Closing as we merged #1601. Feel free to reopen if still an issue with that patch and libnice master.

PR #1601, opened 2019-04-27
ice: check if rtcp_ctx is NULL before use

tmatth:
Every other use of rtcp_ctx has an equivalent check.

Refs #1600.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3845,16 +3845,17 @@ static gboolean janus_ice_outgoing_traffic_handle(janus_handle *handle, janus_ic
 							medium->first_ntp_ts[0] = (gint64)tv.tv_sec*G_USEC_PER_SEC + tv.tv_usec;
 							medium->first_rtp_ts[0] = timestamp;
 						}
-						if(pkt->type == JANUS_ICE_PACKET_AUDIO) {
-							/* Let's check if this was G.711: in case we may need to change the timestamp base */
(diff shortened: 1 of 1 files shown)


tmatth:
Maybe wait until an adequate backtrace is provided to make sure that this is really a case of `rtcp_ctx` being NULL.

lminiero:
Thanks! We don't have that check in master either, though... anyway, in unified-plan it might indeed be happening because of how more often renegotiations can happen. Let's wait for feedback from @amnonbb  and maybe for him to check if this patch fixes it for him.

tmatth:
> Thanks! We don't have that check in master either, though... anyway, in unified-plan it might indeed be happening because of how more often renegotiations can happen. Let's wait for feedback from @amnonbb and maybe for him to check if this patch fixes it for him.

If @amnonbb confirms it, I could make the equivalent patch in master and you could merge it into this branch (if that makes more sense).

amnonbb:
I puted this fix on committed day. Until now issue did not back, but i would wait some weeks. I don't know how to reproduce it and when exactly it was happen.

lminiero:
I'll merge then. I'll apply a similar fix to master when I have some time. Kudos to @tmatth for promptly taking care of this and other issues lately!

lminiero:
Applied to master as well here: https://github.com/meetecho/janus-gateway/commit/b8d21988216d676d55a8bab70277332255fbcf1a

PR #1602, opened 2019-04-29
ice: fix inline warning

tmatth:
Fixes 3 instances of the following warning:
```
ice.c:446:20: warning: inlining failed in call to ‘janus_ice_free_queued_packet’: call is unlikely and code size would grow [-Winline]
```
with gcc (Ubuntu 8.2.0-7ubuntu1) 8.2.0

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -443,7 +443,7 @@ static inline void janus_ice_free_rtp_packet(janus_rtp_packet *pkt) {
 	g_free(pkt);
 }
 
-static inline void janus_ice_free_queued_packet(janus_ice_queued_packet *pkt) {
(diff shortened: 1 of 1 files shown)


lminiero:
But removing inline isn't fixing the warning :)
We had it there to possibly improve performance for that method, which is called often. Not sure if that's indeed the case, but AFAIK the warning is harmless.

tmatth:
@lminiero AFAICT it won't be inlined because it's using global static vars (this is why there's no such warning for  `janus_ice_free_rtp_packet`), so to me it doesn't make sense to add the inline specifier if it can't be used.

lminiero:
Thanks for the clarification! I'll have a look later, currently traveling.

lminiero:
Looks good, merging, thanks!

PR #1604, opened 2019-04-29
Alignment formatting

Joe-Palmer:


lminiero:
Mh, apart from the `JANUS_RABBITMQEVH_ERROR_UNKNOWN_ERROR` line, they don't seem misaligned in my editor though...

Joe-Palmer:
Fair enough. I was basing it on the GitHub editor but happy to leave it. Which editor do you use out of interest?

lminiero:
I'm using Geany, and I don't see the misalignment there. I do see it when using `less` on a terminal, though... Not sure what's going on where.

RSATom:
@lminiero, The reason is simple, you are using tabs equal to 4 chars in your editor, but by default terminal uses tabs equal to 8 chars... It's better use spaces for formatting instead of tabs to avoid such issues.

lminiero:
> It's better use spaces for formatting instead of tabs

Never going to happen! :smile: 
Thanks for the clarification, though, I wasn't aware of that!

RSATom:
It's very old  holy war - "tabs" vs "spaces" vs "tabs mixed with spaces"... Nobody win yet... so don't care :wink:

PR #1609, opened 2019-05-08
Allow payload type override when creating RTSP mountpoints

lminiero:
This is a simple fix, that addresses something I thought we had already and was actually missing, that is the ability to force a different payload type when creating an RTSP mountpoint. For some reason we only allowed the fmtp and rtmpap properties to be overridden, but not the payload type, which caused issues like the one described [here](https://groups.google.com/forum/#!topic/meetecho-janus/t25kHCx56Dc/discussion), where a camera was using a reserved payload type outside of the dynamic range for its video payload type, and the browser was obviously rejecting it.

I expect to merge soon, so please let me know if you notice any regression or typos.


lminiero:
Merging.

PR #1611, opened 2019-05-10
A couple of fixes on SIP race conditions after hangups

lminiero:
This patch includes a few fixes on some race conditions we experienced with the SIP plugin, especially in cases where new calls would be received immediately after hanging up a call. Due to the asynchronous nature of the messaging, and the management of the state, this could at times result in calls being hung up right after they were accepted (due to a previous cleanup still taking place), or broken audio.

Please let me know if you notice any regression with these fixes, as I plan to merge soon.


lminiero:
Merging.

PR #1615, opened 2019-05-13
H.264 temporal scalability support via frame-marking extension

lminiero:
While Chrome has supported H.264 simulcast for a while, temporal scalability support as in VP8 was missing, as there wasn't a way to convey that information to SFUs. In order to do that, [frame marking](https://bugs.chromium.org/p/webrtc/issues/detail_ezt?id=10579) support would be needed instead, that in the meanwhile has been [implemented](https://webrtc-review.googlesource.com/c/src/+/86200).

As such, to complement our existing simulcast support, this patch now uses the [frame marking RTP extension](https://tools.ietf.org/html/draft-ietf-avtext-framemarking-07) when H.264 simulcast is used to involve temporal scalability as well. The demos using simulcast have been updated to now show the temporal scalability buttons for H.264 as well: notice, though, that if you're using an older version of Chrome, those buttons will do nothing at all of course. Besides, you most likely need rid-based simulcasting for this to work (`?simulcast2=true` in the demos query string) for this to work.

Planning to merge soon, so as usual feedback is welcome.

### DIFF
--- a/html/echotest.js
+++ b/html/echotest.js
@@ -214,7 +214,7 @@ $(document).ready(function() {
 									if((substream !== null && substream !== undefined) || (temporal !== null && temporal !== undefined)) {
 										if(!simulcastStarted) {
 											simulcastStarted = true;
-											addSimulcastButtons(msg["videocodec"] === "vp8");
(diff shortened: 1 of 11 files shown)


lminiero:
Merging.

PR #1618, opened 2019-05-14
Add version of dependencies to server info (Janus and Admin API)

lminiero:
Since most of the times, when debugging, we ask people which libraries they're using, which version, and so on, we thought about making this process more straightforward. This patch adds the main dependencies and their versions in the information returned by a `"janus":"info"`, `/janus/info` or `/admin/info` request. In my setup, it looks something like this:

```
   [..]
   "dependencies": {
      "glib2": "2.56.4",
      "jansson": "2.11",
      "libnice": "0.1.14.1",
      "libsrtp": "libsrtp2 2.1.0",
      "libcurl": "7.59.0",
      "crypto": "OpenSSL 1.1.0 (compatible; BoringSSL)"
   },
   [..]
```

Very simply, a new JSON object before the list of modules, listing the version of the main libraries. Most of the libraries have a programmatic way of exposing this information: libnice doesn't though (at least not that I'm aware of) so for that I had to add an external `pkg-config` call in the Makefile to have it added to our `version.c` generated file.

For the moment this only lists the main libraries, but in the future it might be useful to have the same done in all plugins as well, e.g., via a new methods plugins might expose to export their internal dependencies, if any.

Unless you have reasons why this shouldn't be done (breaks something? info should not be exposed?) I'm planning to merge soon, so "be quick or be dead" :wink: .

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -195,6 +195,7 @@ version.c: FORCE | $(dir_target)
 	echo "$(build_date)" | awk 'BEGIN {} {print "const char *janus_build_git_time = \""$$0"\";"} END {} ' >> version.c
 	echo "$(JANUS_VERSION)" | awk 'BEGIN {} {print "int janus_version = "$$0";"} END {} ' >> version.c
 	echo "$(JANUS_VERSION_STRING)" | awk 'BEGIN {} {print "const char *janus_version_string = \""$$0"\";"} END {} ' >> version.c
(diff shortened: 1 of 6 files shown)


lminiero:
To those that REALLY don't want to disclose the dependencies, a new property called `hide_dependencies` can be set in `janus.jcfg` to prevent that.

lminiero:
Merging.

PR #1622, opened 2019-05-14
New experimental debug mode with disabled WebRTC encryption

lminiero:
While we have native `.pcap` support in Janus, as explained in this [blog post](https://www.meetecho.com/blog/capturing-webrtc-traffic-in-janus/) I wrote some time ago, to look at unencrypted traffic for debugging purposes, sometimes it may be useful to do the same thing in a live session. Chrome unstable/canary has a `--disable-webrtc-encryption` command line flag that allows it to disable the WebRTC encryption layers (DTLS and SRTP) but keep all the rest, in order to exchange RTP/RTCP packets in the clear. Anyway, so far this couldn't interoperate with Janus, as we require all the layers, including encryption, to be active.

This patch tries to fix that, adding an experimental debug mode to allow Janus to act the same way too. If you pass the `-w` or `--no-webrtc-encryption` parameter on the command line, or set `no_webrtc_encryption=true` in `janus.jcfg`, then Janus will disable DTLS and SRTP, and do plain RTP/RTCP exactly as Chrome with the property introduced before enabled. Of course, if you enable that mode, regular WebRTC sessions will _NOT_ work, as we'll skip the DTLS negotiation and handshake entirely. Besides, notice that datachannels won't be usable either, as they do require DTLS as an application level transport protocol.

If you don't pass anything (the default) Janus should work exactly as it has always worked so far.

I tested this briefly with EchoTest and VideoRoom and it seems to do its job. I didn't test much more and to be honest I wouldn't really care if this mode is unstable or not, as it's really meant just for debugging under particular circumstances. Anyway, please let me know if you think this is useful, and if you have any feedback.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -195,6 +195,7 @@ version.c: FORCE | $(dir_target)
 	echo "$(build_date)" | awk 'BEGIN {} {print "const char *janus_build_git_time = \""$$0"\";"} END {} ' >> version.c
 	echo "$(JANUS_VERSION)" | awk 'BEGIN {} {print "int janus_version = "$$0";"} END {} ' >> version.c
 	echo "$(JANUS_VERSION_STRING)" | awk 'BEGIN {} {print "const char *janus_version_string = \""$$0"\";"} END {} ' >> version.c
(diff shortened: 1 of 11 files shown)


lminiero:
Merging.

Issue #1616, opened 2019-05-13
Ability to configure a port range for incoming RTP connections when using RTSP


aluedeke:
When using RTSP as source in the streaming plugin the RTP and RTCP ports used to handle the media streams are chosen randomly. It would be desired to configure a range where Janus picks a pair of ports from, to make firewall configurations possible and more secure.

@lminiero is that you would accept a PR for?

lminiero:
@aluedeke sure, it sounds useful! If possible please use the same syntax the SIP/SIPre/NoSIP plugins use, just to make things simpler. You can probably reuse most of the code in there as well to speed up things, e.g., for the config processing part. Thanks!

PR #1623, opened 2019-05-15
Make it possible to set RTSP client port range in the streaming config

mykola-mokhnach:
In order to fine-tune a local firewall it is necessary to have a possibility to set port ranges for RTSP connections. Currently these ports are chosen randomly. We've added a new general setting into streaming config called `rtsp_client_port_range`. The setting accepts valid port ranges and if set then RTSP listening ports will be chosen from this range. In case the setting is not present in the streaming config then the server behaviour is preserved.

We can also observe that janus always opens one random UDP port on startup for both IPv4 and IPv6 (the latter is always +1 greater), although we didn't see any particular `bind` call in the code where this might be happening. All plugins except of streaming and echo are disabled. @lminiero Could you please give us a hint on where to look for?

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -1074,6 +1074,7 @@ typedef struct janus_streaming_mountpoint {
 GHashTable *mountpoints;
 janus_mutex mountpoints_mutex;
 static char *admin_key = NULL;
+static char *rtsp_client_port_range = NULL;
 
 typedef struct janus_streaming_helper {
 	janus_streaming_mountpoint *mp;
@@ -1363,6 +1364,9 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 		if(!notify_events && callback->events_is_enabled()) {
 			JANUS_LOG(LOG_WARN, "Notification of events to handlers disabled for %s\n", JANUS_STREAMING_NAME);
 		}
+		janus_config_item *rtsp_port_range_key = janus_config_get(config, config_general, janus_config_type_item, "rtsp_client_port_range");
+		if(rtsp_port_range_key != NULL && rtsp_port_range_key->value != NULL)
+			rtsp_client_port_range = g_strdup(rtsp_port_range_key->value);
 		/* Iterate on all mountpoints */
 		GList *clist = janus_config_get_categories(config, NULL), *cl = clist;
 		while(cl != NULL) {
@@ -1869,6 +1873,7 @@ void janus_streaming_destroy(void) {
 
 	janus_config_destroy(config);
 	g_free(admin_key);
+	g_free(rtsp_client_port_range);
 
 	g_atomic_int_set(&initialized, 0);
 	g_atomic_int_set(&stopping, 0);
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
After some time experimenting with configuration options we were able to figure out these random port numbers are opened by data channels functionality and they are not shown in `lsof` output if `--disable-data-channels` argument is present. Although, it is still unclear which part of the code is responsible for opening these ports.

lminiero:
Please let's not mix different things in the same ticket. This is for your PR (that I'll try to review when I have some time), not for ports that may be opened somewhere else.

plugins/janus_streaming.c
```diff
@@ -1363,6 +1364,9 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 		if(!notify_events && callback->events_is_enabled()) {
 			JANUS_LOG(LOG_WARN, "Notification of events to handlers disabled for %s\n", JANUS_STREAMING_NAME);
 		}
+		janus_config_item *rtsp_port_range_key = janus_config_get(config, config_general, janus_config_type_item, "rtsp_client_port_range");
+		if(rtsp_port_range_key != NULL && rtsp_port_range_key->value != NULL)
+			rtsp_client_port_range = g_strdup(rtsp_port_range_key->value);
```
lminiero: This looks very inefficient. You're basically parsing the same setting over and over again, any time an RTSP mountpoint is created. Please look at how the SIP/SIPre/NoSIP plugins do it instead.

plugins/janus_streaming.c
```diff
@@ -5052,41 +5057,98 @@ static int janus_streaming_rtsp_parse_sdp(const char *buffer, const char *name,
...
+		if(min_port > max_port) {
+			uint16_t temp_port = min_port;
+			min_port = max_port;
+			max_port = temp_port;
+		}
+		if(min_port == 0)
+			min_port = 0x400;
+		if(max_port == 0)
+			max_port = 0xFFFF;
+		JANUS_LOG(LOG_INFO, "RTSP client port range: %u -- %u\n", min_port, max_port);
```
lminiero: See above: very inefficient.

plugins/janus_streaming.c
```diff
@@ -5052,41 +5057,98 @@ static int janus_streaming_rtsp_parse_sdp(const char *buffer, const char *name,
...
+					if(fds->rtcp_fd < 0) {
+						close(fds->fd);
+						is_port_found = FALSE;
+					} else if(getsockname(fds->rtcp_fd, (struct sockaddr *)&address, &len) < 0) {
+						close(fds->fd);
+						close(fds->rtcp_fd);
+						is_port_found = FALSE;
+					}
+				}
+			}
```
lminiero: This looks like a lot of code that could be greatly simplified. Check the `janus_nosip_allocate_port_pair` and `janus_nosip_bind_socket` methods in NoSIP for instance. Doing something similar would make it quite easy to do the same here, and not only for RTSP, but for all random ports in RTP mountpoints too (that is, any time you pass 0 there).

plugins/janus_streaming.c
mykola-mokhnach: I was following the `admin_key` example, since it is also defined in `general` section of the module, but I agree that keeping a string is inefficient and it would be better to cache the actual min and max port numbers.

plugins/janus_streaming.c
mykola-mokhnach: thanks, will check

lminiero:
@mykola-mokhnach I just prepared a different patch that aims at doing the same thing, but in general: that is, not only RTSP, but any kind of RTP-based mountpoint where the port to choose is random. Please test that one as well and let me know if it does indeed cover the requirement you described in #1616. Thanks!

lminiero:
@mykola-mokhnach did you have a chance to test the patch I mentioned?

mykola-mokhnach:
not yet @lminiero

I think we can do it this week and I'll let you know then. Thanks for taking care of this

PR #1631, opened 2019-05-21
Added status messages to MQTT transport

feymartynov:
This PR adds optional sending of connect and disconnect messages to MQTT transport plugin.

The reason why the same functionality in MQTT EVH plugin is not enough is that event handler plugins are being loaded before other plugins. So it sends the connect status message before the transport is ready which creates a race condition: when another service that sees the status and immediately sends a message janus may skip it if the transport is not ready yet.

The configuration is slightly different from the EVH plugin's. Personally I think it's more clear.

LWT message is also being sent and there's no option to disable it because I can't imagine any use case when one needs to send a regular disconnect message but doesn't need LWT.

### DIFF
--- a/transports/janus_mqtt.c
+++ b/transports/janus_mqtt.c
@@ -60,6 +60,9 @@ void janus_mqtt_session_created(janus_transport_session *transport, guint64 sess
 void janus_mqtt_session_over(janus_transport_session *transport, guint64 session_id, gboolean timeout, gboolean claimed);
 void janus_mqtt_session_claimed(janus_transport_session *transport, guint64 session_id);
 
+#define JANUS_MQTT_DEFAULT_STATUS_TOPIC	"status"
+#define JANUS_MQTT_DEFAULT_STATUS_QOS   1
+
 /* Transport setup */
 static janus_transport janus_mqtt_transport_ =
 	JANUS_TRANSPORT_INIT (
@@ -112,6 +115,15 @@ typedef struct janus_mqtt_context {
 	struct {
 		int timeout;
 	} disconnect;
+	/* If we loose connection, the will is our last publish */
+	struct {
+		gboolean enabled;
+		char *connect_message;
+		char *disconnect_message;
+		char *topic;
+		int qos;
+		gboolean retain;
+	} status;
 	struct {
 		char *topic;
 		int qos;
@@ -119,6 +131,7 @@ typedef struct janus_mqtt_context {
 	struct {
 		char *topic;
 		int qos;
+		gboolean retain;
 	} publish;
 	struct {
 		struct {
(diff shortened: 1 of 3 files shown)


transports/janus_mqtt.c
```diff
@@ -60,6 +60,9 @@ void janus_mqtt_session_created(janus_transport_session *transport, guint64 sess
 void janus_mqtt_session_over(janus_transport_session *transport, guint64 session_id, gboolean timeout, gboolean claimed);
 void janus_mqtt_session_claimed(janus_transport_session *transport, guint64 session_id);
 
+#define JANUS_MQTT_DEFAULT_STATUS_TOPIC	"status"
+#define JANUS_MQTT_DEFAULT_STATUS_QOS						1
```
lminiero: Indentation looks weird here?

transports/janus_mqtt.c
```diff
@@ -377,6 +401,43 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
...
+	/* Status messages configuration */
+	janus_config_item *status_enabled_item = janus_config_get(config, config_status, janus_config_type_item, "enabled");
+	if(status_enabled_item && status_enabled_item->value && janus_is_true(status_enabled_item->value)) {
+		ctx->status.enabled = TRUE;
+
+		janus_config_item *status_connect_message_item = janus_config_get(config, config_status, janus_config_type_item, "connect_message");
+		if(status_connect_message_item && status_connect_message_item->value) {
+			ctx->status.connect_message = g_strdup(status_connect_message_item->value);
+		} else {
+			ctx->status.connect_message = NULL;
```
lminiero: I don't think you need the `else`s here, `ctx` is created via `g_malloc0` which also initializes everything to `0` and `NULL` by default.

transports/janus_mqtt.c
```diff
@@ -377,6 +401,43 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
...
+		}
+
+		janus_config_item *status_disconnect_message_item = janus_config_get(config, config_status, janus_config_type_item, "disconnect_message");
+		if(status_disconnect_message_item && status_disconnect_message_item->value) {
+			ctx->status.disconnect_message = g_strdup(status_disconnect_message_item->value);
+		} else {
+			ctx->status.disconnect_message = NULL;
```
lminiero: Same here.

transports/janus_mqtt.c
```diff
@@ -377,6 +401,43 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
...
+		janus_config_item *status_topic_item = janus_config_get(config, config_status, janus_config_type_item, "topic");
+		if(status_topic_item && status_topic_item->value) {
+			ctx->status.topic = g_strdup(status_topic_item->value);
+		} else {
+			ctx->status.topic = g_strdup(JANUS_MQTT_DEFAULT_STATUS_TOPIC);
+		}
+
+		janus_config_item *status_qos_item = janus_config_get(config, config_status, janus_config_type_item, "qos");
+		if(status_qos_item && status_qos_item->value) {
+			ctx->status.qos = atoi(status_qos_item->value);
```
lminiero: What happens if you put a negative number here?

lminiero:
As a side note, before we can merge this we'd also need you to fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/).

lminiero:
@feymartynov ping :slightly_smiling_face:

feymartynov:
@lminiero pong :)
Thank you for the review. I'm a bit busy at the moment. I'm going to go through your comments as soon as I have time.

lminiero:
Ack, thanks for the feedback!

transports/janus_mqtt.c
feymartynov:
```
[FATAL] [transports/janus_mqtt.c:janus_mqtt_init:465] Can't connect to MQTT broker, return code: -9
```

lminiero:
Sorry, missed your latest fixes, apparently I didn't get a notification... merging then, thanks again for your contribution!

PR #1635, opened 2019-05-23
Set ICE remote credentials when receiving remote SDP, instead of later

lminiero:
The subject basically says it all. So far we've been setting the remote ICE credentials, using `nice_agent_set_remote_credentials()`, only when calling `janus_ice_setup_remote_candidates()`, that is when we believe the ICE checks should start. Anyway, it makes more sense to set them as soon as we're aware of them, that is when we receive the first remote SDP offer or answer, which should also help in case of early prflx candidates.

The only reason this is a pull request and not a direct commit is that I want to make sure it doesn't break anything. There may have been a reason why we were delaying setting the remote credentials to a later stage, but I couldn't find anything in my notes: I've checked with a few of our demos and everything seems to work as expected, but if you can make a quick test as well and let me know that would help. Planning to merge soon if I get no objection.


lminiero:
Merging :+1:

Issue #1627, opened 2019-05-16
Janus crash libwebsockets


danflu:
Hi, Janus is crashing randomly when using websockets and videoroom plugin. 
The stacktrace is provided below.

When query git HEAD version by using command:
git rev-parse HEAD

It returns:
ddde5e28e93a2b42483963297f7ccfc94b0b0ce1

Any ideas ? Thanks a lot!

```
`[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by /opt/janus/bin/irm-janus.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f5595dcca6e in rops_callback_on_writable_ws () from /usr/lib/libwebsockets.so.14
[Current thread is 1 (Thread 0x7f55bd8a2700 (LWP 8129))]
(gdb) bt


#0  0x00007f5595dcca6e in rops_callback_on_writable_ws () from /usr/lib/libwebsockets.so.14
#1  0x00007f5595dc4cc6 in lws_callback_on_writable () from /usr/lib/libwebsockets.so.14
#2  0x00007f55ac0ed13b in janus_websockets_send_message (transport=0x7f5576bc3120, request_id=<optimized out>, 
    admin=<optimized out>, message=0x7f5578016940) at transports/janus_websockets.c:904
#3  0x0000556814679e28 in janus_process_success (request=0x7f5576bc6f60, payload=0x7f5578016940) at janus.c:2247
#4  0x000055681467c59d in janus_process_incoming_request (request=0x7f5576bc6f60) at janus.c:1005
#5  0x0000556814681db8 in janus_transport_requests (data=<optimized out>) at janus.c:2649
#6  0x00007f55c30de3d5 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#7  0x00007f55c1bbf494 in start_thread (arg=0x7f55bd8a2700) at pthread_create.c:333
#8  0x00007f55c1901acf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:97`
```

lminiero:
Try compiling with libasan, as that will give much more relevant info on the crash: https://janus.conf.meetecho.com/docs/debug

danflu:
Thanks. I will update Janus to the latest version and recompile with libasan.

atoppi:
Might be related, coming from #1588 
[...]

PR #1638, opened 2019-05-24
Improvements on writable notifications in WebSockets transport

lminiero:
Attempt to solve #1627, please try if you're encountering issues with websockets.


lminiero:
Merging, as this does seem to be fixing things (assuming libwebsockets >= 3.0 is used, otherwise the same thing we were doing before still applies).

Issue #1641, opened 2019-05-26
No media path established when SIP call traversed Janus 


oofp:
Hello, I have developed Haskell client library for Janus and successfully use for more then a year. Now I wanted to create demo for simple back to back SIP call trough Janus. I have attached the client twice to SIP plugin and then receive "incomingcall" on one plugin and created outbound call on another plugin with SDP of incoming call. Similarly I forwarded  SDP answer of accepted call from outbound leg to inbound. On signalling level everything worked fine,  but media path was not established. I am attaching log for the issue. Not sure, if it is expected to work, so please reject this ticket, if it does not work by design.
[janus.log](https://github.com/meetecho/janus-gateway/files/3219970/janus.log)

Thanks a lot.

lminiero:
This isn't the type of issue you'll solve by looking at logs. You'll have to capture traffic somehow (especially SIP signalling) to see why media streams can't be established. Pretty sure it's a setup issue and not a Janus one (which means this question probably belongs to the [group](https://groups.google.com/forum/#!forum/meetecho-janus) and not here) but I'll keep the issue open until you can confirm that's the case.

oofp:
Thanks Lorenzo. I captured tcpdump (attached along with new janus logs). SIP signalling shows successful setup of two SIP dialogs and SDP are properly exchanges. I also see two RTP stream incoming sent by SIP provider to Janus, but there is no RTP traffic from Janus to SIP provider. 


[janus2.zip](https://github.com/meetecho/janus-gateway/files/3223674/janus2.zip)
![RtpStreams](https://user-images.githubusercontent.com/25211514/58419356-61ee5f00-8058-11e9-9a16-cb9c303910d8.PNG)


Please let me know if I can provide any further information.
Thanks

oofp:
Just wanted to add that log shows that both relay threads started in incoming RTP streams were detected by SIP plugin instances

[Mon May 27 11:55:18 2019] Starting relay thread (112229_11 <--> sip:19058833463@toronto2.voip.ms)
[Mon May 27 11:55:18 2019] Got SIP peer audio SSRC: 82182365

[Mon May 27 11:55:18 2019] Starting relay thread (112229_10 <--> sip:112229_5@158.85.70.149)
[Mon May 27 11:55:29 2019] Got SIP peer audio SSRC: 1401247833

lminiero:
[...]

PR #1642, opened 2019-05-28
Fix trickle issue detected at issue #1641

oofp:
Don't set trickle at SDP if all candidates are provided; Don't assume trickle of not specified at SDP

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -1161,7 +1161,7 @@ int janus_process_incoming_request(janus_request *request) {
 			type = NULL;
 			gboolean do_trickle = TRUE;
 			json_t *jsep_trickle = json_object_get(jsep, "trickle");
-			do_trickle = jsep_trickle ? json_is_true(jsep_trickle) : TRUE;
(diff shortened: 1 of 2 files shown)


lminiero:
We actually set the default to TRUE because, especially at the time, not all browsers would advertise trickle support and then would trickle anyway. As a result, we'd give up sooner than we had to (not all candidates had arrived yet) and ICE would fail. As such, I'm not going to merge this as it is. No objection on only advertising trickle on our side if full trickle is enabled, instead.

As a side note, can you fill in the form for our [CLA](https://cla.conf.meetecho.com/janus/) and sign it? We can't merge otherwise.

lminiero:
Actually, looking at the [specification](https://tools.ietf.org/html/draft-ietf-ice-trickle-21), the `ice-options: trickle` attribute is needed to state you support trickle, which can mean either sending them or receiving them. Since it's true for Janus no matter if we full-trickle or not, we have to put the attribute, so we're doing the right thing.

PR #1643, opened 2019-05-28
Getting 'complteted:true' is received start ICE check if not started already 

oofp:
If trickle with 'complteted:true' is received  start ICE check  not started already


oofp:
Need a bit more tetsing

PR #1644, opened 2019-05-28
Ensure that ICE check is in progress when all trickled candidates are received

oofp:


oofp:
I deployed janus with this change on cloud server and tried several other plugins (AudioBridge, VideoRoom) and encountered no problem.

lminiero:
I understand this fixes the issue for you, but I have concerns about this PR, and I have to study carefully the patch and its implications: just doing `janus_ice_setup_remote_candidates` when it's not the right time may break things. I'm still not sure what your setup is, and who's talking to who WebRTC-wise. IIUC, you're not trickling from the client side: if that's the case, in the JSEP you send to Janus you have to manually add a `trickle:false` property and then we'll ignore the SDP value (which is what we do in [janus.js](https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L2834) as well for instance).

oofp:
Thanks Lorenzo. I will try describe my setup. Here is simplest scenario. My app is attaching to two plugins: SIP and echotest. I receive incoming call in SIP side and forward received JSEP offer to echotest as it is. Then I forward back JSEP Answer from echotest back to SIP side. Without this fix, the media path is not established and webrtcup event is not received. The fix solves the problem. I tried similar test for SIP-SIP, SIP-AudioRoom, SIP-VideoCall. All started working with fix applied. I tested in both my local dev environment and with Janus deployed at cloud with public IP. I perfectly understand your concern, please let me know if I can assist in any way.

lminiero:
Then this means you're trying to have two Janus instances talk to each other (or Janus talk to itself) to create a PeerConnection. This is not an usual scenario, and one that is very different from how pretty much everyone else uses Janus. Please follow my suggestion on the trickle property to see if that works without changes to the code, as that's the documented way of negotiating a PeerConnection where neither party trickles their candidates.

oofp:
Thanks Lorenzo. I tried you advise with adding trickle:false attributed to JSEP, It did not help . But, I think I found less intrusive solution. I added the following logic to parsing SDP:
```c
if (!strcasecmp(a->name, "end-of-candidates")){
    //* FIXME Should we do something with this?
    JANUS_LOG(LOG_VERB, "[%"SCNu64"] end-of-candidates received\n", handle->handle_id);
    janus_flags_set(&handle->webrtc_flags,JANUS_ICE_HANDLE_WEBRTC_ALL_TRICKLES);
```
My understanding is that  JANUS_ICE_HANDLE_WEBRTC_ALL_TRICKLES flag indicates that all candidates have been received and that end-of-candidates also indicates that all candidates have been supplied. 

I tried this changed worked, it perfectly with no other adjustments required. 

I hope this change makes more sense to you? 

Thanks a lot for your patience,
Boris

lminiero:
My feeling is that the change you proposed won't work either, at least not without some tweaks, because there might be more than one `end-of-candidates` in an SDP: specifically, there will be one per m-line, which means that for instance if you have both audio and video you'll have two.

Anyway, my colleagues managed to replicate the issue yesterday, and found a pattern that seems to be the cause, which is basically a race condition between some events that leads to `janus_ice_setup_remote_candidates` never being called. Your patch solves the issue in your specific case, but is not enough apparently for the generic issue they found. I'll let @atoppi chime in on this on the exact problem, as he's the one that identified the pattern. We'll try to come up with a patch soon, so that you can check if it solves your problem as well.

oofp:
Thanks Lorenzo and @atoppi . It is great news that problematic pattern was  identified and proper fix is coming. For time being I will stay with fix at my fork that set JANUS_ICE_HANDLE_WEBRTC_ALL_TRICKLES on "end-of-candidates". I think that this flag is applied only after sdp parsing is completed. 
Setting flag more then twice (when it appears for both audio and video) has no negative impact. The problem will be when end-of-candidates appears for one type media but not for another. But practically, I believe  this case is extremely rare.  
Once proper fix is available I get rid of my fork.

lminiero:
Makes sense, and thanks for the precious feedback so far!

atoppi:
I have replied in the original issue #1641

oofp:
Thanks I will try that tonight

oofp:
closing this PR as proper solution has been provided at https://github.com/meetecho/janus-gateway/pull/1652

PR #1647, opened 2019-05-30
New Admin API method to make synchronous requests to plugins

lminiero:
As the title explains, this introduces a new Admin API request, called `message_plugin`, that allows you to make synchronous requests to plugins using the Admin API instead of the Janus API.

The idea behind it is to allow plugins to expose a few selected requests via Admin API as well, so that they're easier to use: for instance, right now creating a room in the VideoRoom requires creating a session and a handle, which is not hard but is longer than it should (3 requests at least), and has the session-related overhead as part of that; having the ability to create a room via Admin API means that all you need is a single REST call and you're done, which means some management-related operations can be made much more straightforward to implement for developers.

Right now this only lays the foundation for this: I implemented the API mechanism itself and extended the plugin API with a new callback (`handle_admin_message`), but only implemented the callback in the EchoTest as a proof of concept. The EchoTest simply sends back whatever you put in the request, which made it easy for me to check everything works. Next step would be modifying the relevant plugins to support this, which will probably require a bit of a refactoring (e.g., moving the create/edit/destroy code to separate methods, since they might now be invoked in different places).

The Admin API method is very easy to use, as it looks like this:

```
{
        "janus": "message_plugin",
        "transaction": "<unique transaction ID>",
        "admin_secret": "<Admin API secret>",
        "plugin": "<plugin namespace, e.g., janus.plugin.videoroom>",
        "request": {
                // Content of the request
         }
}
```

which means a one-liner with curl is straightforward to implement, e.g.:

```
curl -X POST -H "Content-Type: application/json" -d '{"janus": "message_plugin", "transaction": "123", "admin_secret": "janusoverlord", "plugin": "janus.plugin.videoroom", "request": {"request": "list"}}' http://localhost:7088/admin/
```

To make things easy to test and use I also modified the `admin.html` page to include a `Plugins` tab, from where you can craft requests for a plugin in a visual way, e.g.:

admin-message-plugin

Feedback welcome!

### DIFF
--- a/html/admin.html
+++ b/html/admin.html
@@ -41,6 +41,7 @@
 					<li role="presentation" class="active"><a href="#home" aria-controls="home" role="tab" data-toggle="tab">Home</a></li>
 					<li role="presentation" class="disabled"><a href="#serverinfo" aria-controls="serverinfo" role="tab" data-toggle="tab">Server Info</a></li>
 					<li role="presentation" class="disabled"><a href="#settings" aria-controls="settings" role="tab" data-toggle="tab">Settings</a></li>
+					<li role="presentation" class="disabled"><a href="#plugins" aria-controls="plugins" role="tab" data-toggle="tab">Plugins</a></li>
 					<li role="presentation" class="disabled"><a href="#handlesinfo" aria-controls="handlesinfo" role="tab" data-toggle="tab">Handles</a></li>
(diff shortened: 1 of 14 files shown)


lminiero:
This is now available in all the plugins I thought it made sense to add this to. Looking forward to your feedback! As usual, lacking feedback I'll merge soon.

lminiero:
Merging.

PR #1654, opened 2019-06-05
Support plan-b and unified in Safari dependant on Safari Experimental Features Settings

aberrios85:
Safari users might not have unified plan experimental feature turned on. Check if its on and set SDP Semantics accordingly.

We noticed that janus.js worked fine if users had WebRTC Experimental Features Unified Plan turned on however required the use of plan-b is not.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1677,6 +1677,11 @@ function Janus(gatewayCallbacks) {
 				// For Chrome versions before 72, we force a plan-b semantic, and unified-plan otherwise
 				pc_config["sdpSemantics"] = (Janus.webRTCAdapter.browserDetails.version < 72) ? "plan-b" : "unified-plan";
 			}
+			if(Janus.webRTCAdapter.browserDetails.browser === "safari") {
+				//Safari might have experimental Unified Plan turned on, if its on set semantics to Unified plan if its off use plan-b:
+				pc_config["sdpSemantics"] = (RTCRtpTransceiver.prototype.hasOwnProperty("currentDirection")) ? "unified-plan" : "plan-b";
+				Janus.log("Safari unified plan setting is: " + RTCRtpTransceiver.prototype.hasOwnProperty("currentDirection") + " have set sdpSemantics to " + pc_config["sdpSemantics"]);
+			}
 			var pc_constraints = {
 				"optional": [{"DtlsSrtpKeyAgreement": true}]
 			};


lminiero:
Thanks for this contribution! Please fill our form and sign our [CLA](https://cla.conf.meetecho.com/janus/), though, as this is needed before we can merge. I'll have a look at your patch in the next few days, hopefully tomorrow, and provide feedback.

aberrios85:
Completed the CLA.

The check in the PR is mentioned here:

https://webkit.org/blog/8672/on-the-road-to-webrtc-1-0-including-vp8/

lminiero:
Thanks! Looking at your patch, it may not be enough though. This will indeed create the PeerConnection with the right semantics, but there are other places where we check if the browser supports unified-plan, e.g., when we add tracks and need to see if transceivers are supported or not. Currently, the check we have for Safari is:

```
Janus.webRTCAdapter.browserDetails.browser === "safari" && window.RTCRtpSender.prototype.replaceTrack
```

Not sure if those checks will need to be changed with your new check as well, or if some are better kept as they are.

lminiero:
@aberrios85 any feedback on my note above? Thanks!

aberrios85:
Hi Lorenzo,

I've just been looking over those other checks in the code. Forgive my ignorance, I'm fairly new on the WebRTC scene. As I understand it transceivers are only supported in Unified Plan? So you check for transceiver support in the browser and use them as necessary? But without changing sdp semantics.

The reason we came across this and added the check in my PR is when using VideoRoom on Safari, if the user did not enable Experimental Unified Plan support in Safari they did not receive any streams, only transmitted. Enabling Unified Plan support in Safari solved the issue, and changing SDP to Plan-B fixed the issue if they did not have the feature on, so we added that check.

lminiero:
My point is that, if we use the SDP semantics (which your patch does) but then don't use the unified-plan stuff itself (transceivers), then we're only doing half the job. So maybe the same kind of check you added should be part of those checks too, to make sure we use transceivers if we know that Safari supports unified-plan. Unfortunately I'm very much not a Mac OS guy, and so adding/testing this stuff myself is hard for me: we do have a Mac OS machine for testing, but it's old and we use it rarely.

aberrios85:
Aha, I understand now :bulb:.

It will take me some time to get around to checking those as I have other priorities at the moment :family_man_girl_boy: I will get to look at it one evening soon.

aberrios85:
I've just got to look at the code referencing Transceivers when replacing the Video Track, and have been having a play with Safari, I've added in the check to these parts however, Safari (and Edge for that matter) seem to have issues when replacing the video track without using transceivers. This is probably a seperate issue and may be to do with my implementation in the client. I'm still looking into it.

fippo:
never adapter versions have a helper to determine if unified is supported in safari:
`adapter.browserDetails.supportsUnifiedPlan
see https://github.com/webrtcHacks/adapter/pull/954

Safari does not (afaik; it would be better if they did) understand the non-spec sdpSemantics so setting it shouldn't result in anything.

lminiero:
@fippo but didn't newer versions (7.x?) also stop supporting some older browsers though? At least from what I remember reading some time ago.

lminiero:
Mh, and it looks like cdnjs (what we currently use to include the adapter) only has up to v6.4.0...

lminiero:
That said, we do have a similar check in the `unified-plan` version of `janus.js`:

https://github.com/meetecho/janus-gateway/blob/39a996070ce55a910997ed696f931a6b607b180f/html/janus.js#L378

I guess this could be backported to master as well.

lminiero:
I've ported the check I mentioned to `janus.js`, and started using it across the code: https://github.com/meetecho/janus-gateway/commit/fe8c88b24184a93f7c406ff82b372a082c2cfdff
@aberrios85 can you check if that works for you now? If so, I guess this replaces your patch as well.

lminiero:
Closing.

PR #1656, opened 2019-06-06
Refactored janus-pp-rec to support command line options

lminiero:
As the title says. It should be 100% compatible with the way before, with the difference that the values we previously only exposed via environment variables (which were undocumented, incidentally) are now exposed via command line as well (the env variables are still there, they're just ugly :slightly_smiling_face: ). Besides, the existing command line arguments are now more flexible. The new usage of the tool now prints:

```
Usage: janus-pp-rec [OPTIONS] source.mjr [destination.[opus|wav|webm|mp4|srt]]

  -h, --help                    Print help and exit
  -V, --version                 Print version and exit
  -j, --json                    Only print JSON header  (default=off)
  -H, --header                  Only parse .mjr header  (default=off)
  -p, --parse                   Only parse and re-order packets  (default=off)
  -m, --metadata=metadata       Save this metadata string in the target file
  -r, --postreset-trigger=count Number of packets needed to detect a timestamp
                                  reset (default=200)
  -i, --ignore-first=count      Number of first packets to ignore when
                                  processing, e.g., in case they're cause of
                                  issues (default=0)
  -a, --audiolevel-ext=id       ID of the audio-levels RTP extension
                                  (default=none)
  -v, --videoorient-ext=id      ID of the video-orientation RTP extension
                                  (default=none)
  -d, --debug-level=1-7         Debug/logging level (0=disable debugging,
                                  7=maximum debug level; default=4)
  -D, --debug-timestamps        Enable debug/logging timestamps  (default=off)
  -o, --disable-colors          Disable color in the logging  (default=off)
```

I'm planning to merge quite soon, so if you encounter any regression with this tool, please let me know. Pinging @tgabi333 in particular, as I know he uses the tool a lot.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,8 @@ Makefile.in
 /events/*.la
 /events/.libs
 /postprocessing/*.o
+/postprocessing/pp-cmdline.c
+/postprocessing/pp-cmdline.h
 /fuzzers/*.a
 /fuzzers/*.o
 /fuzzers/out
--- a/Makefile.am
+++ b/Makefile.am
@@ -515,6 +515,8 @@ if ENABLE_POST_PROCESSING
 bin_PROGRAMS += janus-pp-rec
 
 janus_pp_rec_SOURCES = \
+	postprocessing/pp-cmdline.c \
+	postprocessing/pp-cmdline.h \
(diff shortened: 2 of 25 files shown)


lminiero:
Merging.

PR #1657, opened 2019-06-07
Adding reference to Haskell binding of Janus client protocol 

oofp:
using WebSocket transport with examples


lminiero:
Looks good, thanks for working on that project and for sharing it!

lminiero:
Updated the online docs as well: https://janus.conf.meetecho.com/docs/resources

PR #1660, opened 2019-06-08
enable dtls window size on non-MacOS machines

garry81:
because HAS_DTLS_WINDOW_SIZE is *NOT* defined anywhere, we can't
change dtls window size. define HAS_DTLS_WINDOW_SIZE for non-MacOS
machines


sobo08:
BTW, is there any reason to disable HAS_DTLS_WINDOW_SIZE for MacOS? I had no issues on MacOS with it.

lminiero:
Apparently it's been there for a very long time, since #125 was merged. IIRC it was there because MacOS was the one system where you couldn't set the window size, not the other way around: probably older versions of the library there had this problem, but I have no idea since I'm not a Mac user. But I agree that the way it was used didn't really work, as it was defining a conditional that the configure would then not act upon. Let me check if this works in my setup and in case I'll merge :+1:

lminiero:
Thanks for the fix, merging!

PR #1661, opened 2019-06-09
[WIP] added ability act websocket transport as client

RSATom:
Hi, @lminiero. 
This work is not finished yet, but I just curious your opinion about add ability to websocket transport act as client. What do you think if it could be useful someone else except me?

### DIFF
--- a/html/streamingtest.js
+++ b/html/streamingtest.js
@@ -42,11 +42,20 @@
 // in the presented order. The first working server will be used for
 // the whole session.
 //
+const useWebSockets = true;
 var server = null;
-if(window.location.protocol === 'http:')
-	server = "http://" + window.location.hostname + ":8088/janus";
-else
-	server = "https://" + window.location.hostname + ":8089/janus";
+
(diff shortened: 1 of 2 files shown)


lminiero:
I think this came up once or twice in the past, e.g., for some customers that had IoT scenarios where Janus may be deployed as a box somewhere behind a NAT and needs to contact an external service itself.

Anyway, I don't feel this requires a change in Janus: this can be done quite simply with a small companion WS proxy application, that one one side connects to Janus itself, on the other side connects to the external service, and simply bridges the messages back and forth; disconnection on one side triggers a disconnection on the other, so no risk of getting orphaned sessions. Besides, I'm very hesitant in allowing any of the existing transports to change to accomodate this, as 1. we'd need to mantain two different modes ourselves here (which I'd really rather not do), and 2. people would likely start asking the same feature for other transports as well, and that's a can of worms.

RSATom:
Understand. Thank you for detailed answer!

PR #1663, opened 2019-06-11
fix SEDS length when adding to compound

glenn-hpcnt:
The compound rtcp packet is discarded in client side because the sdeslen value is invalid. (too long) I changed temporarily this value to matching the current cname length. In essence, sdeslen should be adjusted depending on cname dynamically.


lminiero:
Mh, this might be because we changed the CNAME from `janusaudio`/`janusvideo` to simple `janus` for both streams, and this means 5 less bytes, so two less words: so yes, your change from `24` to `16` seems indeed correct here. Thanks for the fix, merging!

PR #1664, opened 2019-06-13
Changed slowlink to use lost packets instead of NACKs, and made it configurable

lminiero:
Historically, the trigger for the Janus API `slowlink` events has always been the number of NACKs that were sent or received. Since NACKs are optional, though (and never negotiated by browsers for audio, incidentally), this made the feature a bit too frail.

This patch changes the behaviour, and now uses the count of lost packets to send the event instead. Specifically, every second Janus now checks how many packets Janus or the user lost since the last check, and if they exceed a specified threshold, the `slowlink` event is sent. The checks are different for audio and video, and for the direction of media: as such, if the threshold is 4 and then 2 outgoing packets and 2 incoming packets are lost, or 3 incoming audio packets are lost and 1 incoming video packet is lost, the threshold is _NOT_ reached. It _MUST_ be lost packets for the same media in the same direction.

The threshold is now also configurable, statically and dynamically. The static configuration can be done via command line (`-W, --slowlink-threshold`) and in `janus.jcfg` (`slowlink_threshold` property in `[media]`), the dynamic configuration via Admin API (`set_slowlink_threshold` request). This should allow you guys to tinker with the settings, to find the value you prefer to avoid getting spammed with notifications in cases where it might not be needed. The default value is `4`.

Notice that the format of the event changes a bit. It now includes a `media` property as well, to specify if it's related to audio or video, and instead of a `nacks` property (which is gone) we include a `lost` property instead to indicate the count of recently lost packets that led to the event.

Please let me know if you have feedback on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,9 @@ or on the command line:
 	-t, --no-media-timer=number   Time (in s) that should pass with no media
                                   (audio or video) being received before Janus
                                   notifies you about this
+	-W, --slowlink-threshold=number
+                                  Number of lost packets (per s) that should
+                                  trigger a 'slowlink' Janus API event to users
 	-r, --rtp-port-range=min-max  Port range to use for RTP/RTCP (only available
 								  if the installed libnice supports it)
 	-B, --twcc-period=number      How often (in ms) to send TWCC feedback back to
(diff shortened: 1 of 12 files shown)


trankkilou:
We will test it AZAP as we use it a lot

lminiero:
@trankkilou any feedback? I'd like to merge this soon. Thanks!

trankkilou:
Hi lorenzo,

No sorry We didn’t have the time to test. 
I think we won’t before a while so if you are hurry to merge your PR don’t wait for us. 
I will give a feedback anyway when we will have test it. 

Julien

lminiero:
@trankkilou ack, thanks for the feedback. I'll merge then :+1:

PR #1666, opened 2019-06-18
 Fix reference leak

uxmaster:
The leak was discovered while playing with the `voicemail` plugin with `#define REFCOUNT_DEBUG`.


lminiero:
Mh, have you tested this with other plugins? The VoiceMail plugin is very rarely used, and so the issue may actually lie there, e.g., by it not releasing the plugin session and so keeping the handle itself alive as well. I'm afraid this could actually result in crashes, because in other scenarios this might end up freeing the handle much sooner than it should. I often use the reference count debugging, and with the most commonly used plugins I don't have any leak at all right now.

uxmaster:
Actually, the VoiceMail plugin is the only one that uses this `end_session` function. That is why you don't have any leaks in other plugins. Just try your VoiceMail plugin demo.

BTW. I think there is a similar leak while handling `leave` request in the AudioBridge plugin. This request is also never used in the demos, so I didn't dig much into it.

lminiero:
Got it, thanks for the clarification! :+1: 
I'll have a look tomorrow in the morning.

lminiero:
I verified the issue, and actually fixed an issue in the process: for some unknown reason (copy/paste from `close_pc()` is my guess :grin:) we were not going on with `end_session()` if `JANUS_ICE_HANDLE_WEBRTC_ALERT` was set, which of course made no sense, so that's now fixed in https://github.com/meetecho/janus-gateway/commit/05ba4561dd84a05958beaf415bdc9a003ba6cb03.

Thanks for the fix, merging!

PR #1673, opened 2019-06-22
Fix release sctp resources if the creation of the association fails

ggarber:
When the creation of the sctp association fails (f.e. `usrsctp_connect` fails) the sctp object is released with `janus_refcount_decrease` but the socket is not closed (no call to `usrsctp_close`) and the address is not unregistered (no call to`usrsctp_deregister_address`).

These can leak some resources although those failure cases are probably very unlikely.

The current patch calls `janus_sctp_association_destroy` instead of `janus_refcount_decrease` to make sure the full cleanup is done and stores early the socket handle in `sctp->sock` so that it is always stored when calling `janus_sctp_association_destroy`.   In case the sctp socket creation fails (`usrsctp_socket`)  the sctp address is unregistered but no `usersctp_close` is done.

How to reproduce: Didn't try to reproduce it, only looked at the code.
How to test the patch: Tested the VideoRoomDemo and the TextRoomDemo still work but didn't explicitly test the error cases.

### DIFF
--- a/sctp.c
+++ b/sctp.c
@@ -172,20 +172,22 @@ janus_sctp_association *janus_sctp_association_create(janus_dtls_srtp *dtls, jan
 		sctp->stream_buffer[i] = 0;
 	}
 	sctp->stream_buffer_counter = 0;
-	sctp->sock = NULL;
 
 	usrsctp_register_address((void *)sctp);
 	usrsctp_sysctl_set_sctp_ecn_enable(0);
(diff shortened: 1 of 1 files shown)


ibc:
Hi @ggarber!

lminiero:
@ggarber thanks, this looks good! Can you fill in the form for our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

lminiero:
Thanks, merging! :+1:

PR #1677, opened 2019-06-27
Update bower.json and package.json

ancorgs:
The files `bower.json` and `package.json` were outdated, specially their dependencies. This fixes that. 

Those files allow to bundle `janus.js` in other JavaScript projects. The pull request also includes some documentation about how to do that without copying a modified version of `janus.js` into the project (which is possible and documented, but kind of ugly).

As a bonus track, this includes a fix for a minor bug found by ESLint.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,7 +33,6 @@
     "tests"
   ],
   "dependencies": {
-    "jquery": "*",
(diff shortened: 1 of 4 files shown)


lminiero:
Thanks for taking care of all this, merging!

PR #1678, opened 2019-06-27
ice: avoid dereferencing component if NULL

tmatth:
This fixes a crash observed repeatedly in stress testing Janus (under valgrind) with aiortc, where this dereference was segfaulting:
https://github.com/meetecho/janus-gateway/blob/4b07eadd6ed48db6f16e5c747bcd6fba09dd8fa3/ice.c#L3831

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3826,6 +3826,9 @@ static gboolean janus_ice_outgoing_traffic_handle(janus_ice_handle *handle, janu
 			JANUS_LOG(LOG_WARN, "[%"SCNu64"] WebRTC encryption disabled, skipping DTLS handshake\n", handle->handle_id);
 			janus_ice_dtls_handshake_done(handle, component);
 			return G_SOURCE_CONTINUE;
+		} else if(!component) {
+			JANUS_LOG(LOG_WARN, "[%"SCNu64"] ICE component not initialized, aborting DTLS handshake\n", handle->handle_id);
+			return G_SOURCE_CONTINUE;
 		}
 		/* Start the DTLS handshake */
 		janus_dtls_srtp_handshake(component->dtls);


lminiero:
How can that be NULL at that point? The `janus_ice_dtls_handshake` is queued by the selected-pair event, which is in the same event loop as the outgoing traffic handler: besides, the DTLS stack is actually initialized before ICE even starts. I only see two possible causes:

1. since we create the DTLS stack *after* `nice_agent_attach_recv` is called, we may be in the very unlikely scenario where starting the loop gets ICE to succeed before we get to that next line? (race condition) if that's the case, we can try moving that ahead;
2. the DTLS stack is destroyed between the selected-pair and the outgoing handler, which I think can't happen (DTLS is among the last things to be freed)

Anyway, thanks for identifying this issue, and for providing a fix: I'll merge, and think about the causes above later. Merging!

tmatth:
> How can that be NULL at that point? The `janus_ice_dtls_handshake` is queued by the selected-pair event, which is in the same event loop as the outgoing traffic handler: besides, the DTLS stack is actually initialized before ICE even starts. I only see two possible causes:
> 
>     1. since we create the DTLS stack _after_ `nice_agent_attach_recv` is called, we may be in the very unlikely scenario where starting the loop gets ICE to succeed before we get to that next line? (race condition) if that's the case, we can try moving that ahead;
> 
>     2. the DTLS stack is destroyed between the selected-pair and the outgoing handler, which I think can't happen (DTLS is among the last things to be freed)
>

Yeah, this wasn't clear to me either, now that I'm logging it instead of crashing it, I'm always seeing this pattern:
```
[Thu Jun 27 18:58:12 2019] [4224716446593592] WebRTC resources freed; 0x1d90aa50 0x1b6e0de0
[Thu Jun 27 18:58:12 2019] [WARN] [4224716446593592] ICE component not initialized, aborting DTLS handshake
```

lminiero:
So it is indeed case 2.

The good thing is that `janus_ice_webrtc_free` is always called by the libnice loop, so there's no risk of race conditions here: the check you added should be enough to keep things clean, and just as the message says simply abort a handshake that won't be needed anymore.

PR #1693, opened 2019-07-10
[Fix]: janus.js client bug, 'for in loop on array' is a risk since it…

itamayo:
'for in loop on array' is a risk since it's take in account any object defined on array as key, for example, if you has prototyped a array herite it, as 
```js
Array.prototype.mean = function (){... } 

for (var i in config.candidate){
 /* it will try with 'mean'  config.candidates['mean'] */
var candidate = config.candidates[i];
if(!candidate || candidate.completed === true) {
  // end-of-candidates
  config.pc.addIceCandidate({candidate:''});
  } else {
      // New candidate
      config.pc.addIceCandidate(candidate);
 }

}

```
it will fail, it happens to me =(

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1829,7 +1829,7 @@ function Janus(gatewayCallbacks) {
 					config.remoteSdp = jsep.sdp;
 					// Any trickle candidate we cached?
 					if(config.candidates && config.candidates.length > 0) {
-						for(var i in config.candidates) {
+						for(var i = 0; i< config.candidates.length; i++) {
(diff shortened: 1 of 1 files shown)


lminiero:
Sorry, apparently I missed this... thanks for the fix! Can you sign our [CLA](https://cla.conf.meetecho.com/janus/) so that we can merge this?

itamayo:
Done ! Congrats for the good job done !! Nice toolkit =)

lminiero:
Thanks, merging! :+1:

PR #1695, opened 2019-07-11
Fix Janus not sending DATA_CHANNEL_ACK when requested stream id = 0.

atoppi:
This PR fixes a bug that prevented Janus to send an ACK response to a datachannel open request with id=0.

According to the draft:

> After the DATA_CHANNEL_OPEN message has been sent, the sender of the
   DATA_CHANNEL_OPEN can start sending messages containing user data
   without waiting for the reception of the corresponding
   DATA_CHANNEL_ACK message.  However, before the DATA_CHANNEL_ACK
   message or any other message has been received on a Data Channel, all
   other messages containing user data and belonging to this Data
   Channel MUST be sent ordered, no matter whether the Data Channel is
   ordered or not.  After the DATA_CHANNEL_ACK or any other message has
   been received on the Data Channel, messages containing user data MUST
   be send ordered on ordered Data Channels and MUST be sent unordered
   on unordered Data Channels.

This means that the browsers opening the first datachannel with Janus, always considered the channel ordered, no matter the flag specified in the JS API, because of the absence of the ACK from the Server.

### DIFF
--- a/sctp.c
+++ b/sctp.c
@@ -819,26 +819,22 @@ void janus_sctp_handle_open_request_message(janus_sctp_association *sctp, janus_
 	channel->stream = stream;
 	channel->flags = 0;
 	sctp->stream_channel[stream] = channel;
-	if(stream == 0) {
-		janus_sctp_request_more_streams(sctp);
+	if(janus_sctp_send_open_ack_message(sctp->sock, stream)) {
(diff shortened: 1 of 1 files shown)


lminiero:
LGTM :+1:

PR #1706, opened 2019-07-19
add configurable maxBitrate values for simulcast encodings

vivaldi-va:
While trying to simulcast higher resolution streams, I've ran into an issue whereby the hard-coded values for maxBitrate were not sufficient. I've added a option to configure these values individually when initializing Janus. 

By default they will remain as they were, but are now set in a single variable so they also remain consistent should the default values need changing for any reason.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -512,6 +512,22 @@ function Janus(gatewayCallbacks) {
 	if(isNaN(longPollTimeout))
 		longPollTimeout = 60000;
 
+	// overrides for default maxBitrate values for simulcasting
+	var simulcastMaxBitrates = {
+		high: 900000,
+		medium: 300000,
+		low: 100000,
+	};
+
+	if (gatewayCallbacks.simulcastMaxBitrates !== undefined && gatewayCallbacks.simulcastMaxBitrates !== undefined ) {
+		if (gatewayCallbacks.simulcastMaxBitrates.high)
+			simulcastMaxBitrates.high = gatewayCallbacks.simulcastMaxBitrates.high;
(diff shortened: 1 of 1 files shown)


lminiero:
The idea is nice, but I'm not sure about the implementation. Overriding a property once will override it for all  PeerConnections interested in simulcast for the same session, which seems broken. The defaults should be immutable, and invoked when an override is not available.

vivaldi-va:
ah, you're right. Added a function to return the values instead, which will evaluate any set config values each time and return accordingly.

lminiero:
I hate to be a pain, but can `getMaxBitrates()` take an argument, instead of accessing the `gatewayCallbacks` object itself? Seems more solid that way, and would make it easier should we update janus.js in the future to be more "object-oriented".

vivaldi-va:
no worries, updated

lminiero:
Oh, just noticed you're using `gatewayCallbacks`: why not use `callbacks` instead? The way you're doing it requires the simulcast properties to be set when creating the session, rather when doing `createOffer`. Apologies for doing this after commit, but I'm in Montreal for the IETF meeting and I don't have much time to look at stuff in detail at the moment, and missed this before...

lminiero:
Thanks for the quick fixes! I'll make a quick test to see if it works for me, and merge in case. Before I can do that, can you sign our [CLA](https://cla.conf.meetecho.com/janus/) though?

vivaldi-va:
Done. Thanks for taking the time, enjoy the conf!

lminiero:
Thanks, merging! :v:

PR #1707, opened 2019-07-22
Fix typo in textroom_handle_admin_message to stop segfault

MvEerd:
This typo could cause a segfault when an admin message was sent with a ```request``` parameter but without ```textroom``` parameter because the validation in ```JANUS_VALIDATE_JSON_OBJECT``` is checking for ```request``` but the code was still attempting to access ```textroom```
```==16220==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f8fde859608 bp 0x7f8fcf2f43d0 sp 0x7f8fcf2f3b30 T9)
==16220==The signal is caused by a READ memory access.
==16220==Hint: address points to the zero page.
    #0 0x7f8fde859607  (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x5b607)
    #1 0x7f8fcd5e312c in janus_textroom_handle_admin_message plugins/janus_textroom.c:1173```


lminiero:
Good catch, thanks for fixing it! Merging :+1:

PR #1709, opened 2019-07-24
Add MQTT v5 support

feymartynov:


lminiero:
Thanks! I won't be able to review until I get back home (currently abroad for a conference for the whole week), but froom a quick glance you definitely broke the indentation of the configuration file. Please fix that.

lminiero:
@feymartynov ping on the config file fix :wink: 

On the rest of the patch, I don't think it's ok the way it is now. You updated the README to say that only v1.1 of the library is required, and that v1.3 is only needed if you want MQTTv5. Anyway, the code uses the `*5` version of methods and structures (e.g., `MQTTAsync_successData5`) no matter what, which means that if someone has the v1.1 installed, compilation will fail because of the unrecognized structures.

This means the patch should be updated to:

* either mandate support for v1.3 of the library (but I don't know how widespread that is, and it would break retrocompatibility to those with an older version of the library)
* or add a new condition that you can `#ifdef` on to see if you can use the new stuff (but that will make the code "ugly", since it would be in a lot of places).

I personally don't want to break retrocompatibility (we've tried to ensure that for other transports too across the years), so I'd go for the second option, unless there's a strong and compelling point on 1.3 being pretty much universally deployed and available. Pinging @oej as well since AFAIK he's still using the MQTT code.

feymartynov:
You're right. I would also like to save back compatibility so I'm going to add missing `#ifdef`s and try to build it with Paho 1.1.

feymartynov:
@lminiero updated the PR.
Tested it with both Paho 1.1.0 (MQTT 3.1.1) and Paho 1.3.0 (MQTT 3.1.1 & 5).

conf/janus.transport.mqtt.jcfg.sample
```diff
@@ -7,6 +7,7 @@ general: {
 	url = "tcp://localhost:1883"		# The connection URL of the MQTT broker: if you want
 										# to use SSL, make sure you type ssl:// instead of tcp://,
 										# and that you configure the SSL settings below
+  #mqtt_version = "3.1.1"     # Protocol version. Available values: 3.1, 3.1.1 (default), 5.
```
lminiero: Indentation still broken here... please use tabs as per the guidelines.

transports/janus_mqtt.c
```diff
@@ -157,27 +162,95 @@ int janus_mqtt_client_message_arrived(void *context, char *topicName, int topicL
 void janus_mqtt_client_delivery_complete(void *context, MQTTAsync_token token);
 int janus_mqtt_client_connect(janus_mqtt_context *ctx);
 void janus_mqtt_client_connect_success(void *context, MQTTAsync_successData *response);
+#ifdef MQTTVERSION_5
+	void janus_mqtt_client_connect_success5(void *context, MQTTAsync_successData5 *response);
```
lminiero: No need to indent here, it's a function declaration as the others; same for the others below.
Thinking about it, it might make sense to group all the v5 declarations together in a single `#ifdef`? Would make everything more readable IMO.

transports/janus_mqtt.c
```diff
@@ -306,6 +379,25 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
 	}
 
 	/* Connect configuration */
+	janus_config_item *mqtt_version = janus_config_get(config, config_general, janus_config_type_item, "mqtt_version");
+	const char *mqtt_version_str = (mqtt_version && mqtt_version->value) ? g_strdup(mqtt_version->value) : JANUS_MQTT_VERSION_DEFAULT;
```
lminiero: No need to `g_strdup` here, you're not keeping `mqtt_version_str`: it's only declared and used here. The way it is now you're introducing a memory leak.

transports/janus_mqtt.c
```diff
@@ -440,12 +532,21 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
 	}
 
 	/* Creating a client */
-	if(MQTTAsync_create(
+	MQTTAsync_createOptions create_options = MQTTAsync_createOptions_initializer;
+
+	#ifdef MQTTVERSION_5
+		if (ctx->connect.mqtt_version == MQTTVERSION_5) {
+			create_options.MQTTVersion = MQTTVERSION_5;
+		}
```
lminiero: Indentation is broken. The way the code is structured in the rest of the repo, the `#ifdef`, `#endif`, etc. are all shifted to the left (no indentation), while what they contain should have the same indentation as if they were not wrapped in there. I won't write it under all of the changes, so please update this wherever this occurs.

feymartynov:
Updated the PR.

lminiero:
Thanks a lot for the very quick fixes! I'll give it a test later to try and merge before I leave :+1:

conf/janus.transport.mqtt.jcfg.sample
```diff
@@ -7,6 +7,7 @@ general: {
 	url = "tcp://localhost:1883"		# The connection URL of the MQTT broker: if you want
 										# to use SSL, make sure you type ssl:// instead of tcp://,
 										# and that you configure the SSL settings below
+	#mqtt_version = "3.1.1"     			# Protocol version. Available values: 3.1, 3.1.1 (default), 5.
```
lminiero: It looks like it's still spaces between the value and the comment. Can you fix that too?

transports/janus_mqtt.c
```diff
@@ -306,6 +350,25 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
...
+
+	if(strcmp(mqtt_version_str, JANUS_MQTT_VERSION_3_1) == 0) {
+		ctx->connect.mqtt_version = MQTTVERSION_3_1;
+	} else if(strcmp(mqtt_version_str, JANUS_MQTT_VERSION_3_1_1) == 0) {
+		ctx->connect.mqtt_version = MQTTVERSION_3_1_1;
+	} else if(strcmp(mqtt_version_str, JANUS_MQTT_VERSION_5) == 0) {
+	#ifdef MQTTVERSION_5
+		ctx->connect.mqtt_version = MQTTVERSION_5;
+	#else
+		JANUS_LOG(LOG_FATAL, "Using MQTT v5 requires compilation with Paho >= 1.3.0");
```
lminiero: There's a `\n` missing here.

transports/janus_mqtt.c
```diff
@@ -306,6 +350,25 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
...
 	/* Connect configuration */
+	janus_config_item *mqtt_version = janus_config_get(config, config_general, janus_config_type_item, "mqtt_version");
+	const char *mqtt_version_str = (mqtt_version && mqtt_version->value) ? mqtt_version->value : JANUS_MQTT_VERSION_DEFAULT;
+
+	if(strcmp(mqtt_version_str, JANUS_MQTT_VERSION_3_1) == 0) {
+		ctx->connect.mqtt_version = MQTTVERSION_3_1;
+	} else if(strcmp(mqtt_version_str, JANUS_MQTT_VERSION_3_1_1) == 0) {
+		ctx->connect.mqtt_version = MQTTVERSION_3_1_1;
+	} else if(strcmp(mqtt_version_str, JANUS_MQTT_VERSION_5) == 0) {
+	#ifdef MQTTVERSION_5
```
lminiero: This block of `#ifdef #else #endif` still has a broken indentation.

feymartynov: Why? `#ifdef #else #endif` are aligned one tab left and the code itself is aligned like it would be without these directives.

feymartynov:
Updated the PR.

lminiero:
Thanks, merging :+1:

PR #1713, opened 2019-07-26
rtp: drop unused step parameter from plugins

tmatth:
@atoppi @lminiero I've left the paratmer in place, but set it to zero everywhere to be able to drop some dead code.


lminiero:
Actually this is already addressed in the `unified-plan` branch, so since it can be considered an API change, I'd rather do it as part of that larger effort than on master. Pretty sure that if I merge this, I'll have some headaches aligning that branch because of conflicts :smile:

tmatth:
@lminiero superseded by https://github.com/meetecho/janus-gateway/pull/1715

Issue #1716, opened 2019-07-27
janus-pp-rec broken g711 conversion


danielferrer:
Converting a .mjr g711 file to wav, result in a corrupted .wav file.

The problem is that the generated wav header is incorrect, check the following chart for wav header:
http://soundfile.sapp.org/doc/WaveFormat/

You're not setting "Chunksize" and "SubChunk2Size".

I've made a patch to fix this. The result is a wav file with same length but this can be played. The original file can't be played, for example Audacity doesn't open it.

PR #1717, opened 2019-07-27
Fix wav header on g711 conversion

danielferrer:
Fixes mjr to g711 conversion.
closes issue #1716

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -857,7 +857,7 @@ int main(int argc, char *argv[])
 				exit(1);
 			}
 		} else if(g711) {
-			if(janus_pp_g711_create(destination, metadata) < 0) {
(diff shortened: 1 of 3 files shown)


lminiero:
Thanks for spotting the problem and working on a fix! But I don't think this is a valid solution.

What you're doing is passing the size of the .mjr file to the .wav creator, which is wrong. The .mjr file has a lot of overhead (headers for each packet, RTP header, extensions, etc.), which means the size of the .wav file will be much smaller. What should happen is that, when we have to close the file, we should check how large it is and seek back to the beginning and update the header. I was pretty sure we were doing this already (I took some code I used in another project for saving to .wav, and I did do it there), but I may have missed that part. I'll take note that this needs fixing.

lminiero:
Fixed in the commit above. It seems to be working for me, please let me know if that's not the case.

PR #1718, opened 2019-07-29
Tool to convert .mjr files to .pcap

lminiero:
As the title says, this introduces a simple tool, part of the post-processing utilities, to convert an existing `.mjr` file to a `.pcap` readable, e.g., with Wireshark. Obviously network levels are simulated, and at the moment the timestamp for each packet is fake as well: this still makes the tool useful for inspecting RTP packets in a visual way, and so nice for debugging.

We're thinking of modifying the `.mjr` format slightly to include the time at which each specific packet was received. This will be useful for several reasons, one of which would be that such a generated `.pcap` would have "real" timing information as well. Not sure yet if I'll just modify this PR for that effort or if that will come later, though.

Planning to merge soon, so feedback welcome!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ docs/html/
 /janus
 /janus-cfgconv
 /janus-pp-rec
+/mjr2pcap
 /plugins/*.so
 /transports/*.so
 /events/*.so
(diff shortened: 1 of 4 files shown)


lminiero:
Merging.

PR #1719, opened 2019-07-30
Added arrival time of packets to .mjr files (backwards compatible)

lminiero:
This complements the feature added in #1718, and lays the foundation for more advanced processing in the future. As the title says, now for all packets we save to an `.mjr` file, we also store a timestamp of when the packet was received. The timestamp is relative to the `u` (updated) info already available in `.mjr` files, and is in milliseconds: this means the first packet will always have value `0`, and others will have a value that is relative to that. If you need an absolute value (which is what we do when using `mjr2pcap` for instance) then you can use the `u` value in the global header as a starting point.

This is all backwards compatible, meaning that recordings saved with this version of Janus will still be processable by older versions of `janus-pp-rec`. In fact, the way we did this was by "abusing" the way too large `MEETECHO` string we use to separate packets: we left 4 characters for the string (`MEET`), and used the other 4 for the timestamps, which is a 32-bit unsigned integer. Since older versions of the tool only check the first two characters to check if it's a global or packet header, this still works, as they'll simply ignore the whole 8 bytes no matter what they contain. The new `janus-pp-rec`, instead, is aware of the new `MJR00002` global header, which when present means timestamps will be included.

At the moment we only use this information to "enrich" `.pcap` files created with `mjr2pcap` with a more realistic time for packets (which is in ms instead of us, but that doesn't really matter). In the future we may use it for more advanced features, e.g., better skew compensation in some edge cases.

Planning to merge soon, so please test and give feedback!

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -285,6 +285,7 @@ int main(int argc, char *argv[])
 	/* Pre-parse */
 	if(!jsonheader_only)
 		JANUS_LOG(LOG_INFO, "Pre-parsing file to generate ordered index...\n");
+	gboolean has_timestamps = FALSE;
 	gboolean parsed_header = FALSE;
 	gboolean video = FALSE, data = FALSE;
 	gboolean opus = FALSE, g711 = FALSE, g722 = FALSE,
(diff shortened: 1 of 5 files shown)


lminiero:
Merging.

PR #1720, opened 2019-08-01
Handling of split audio/video connection addresses

pawnnail:
This request addresses a SIP plugin defect. When peer offers or answers with different connection media addresses for audio and media then either video or audio packet relay process is broken. This applies when setting up a new session or re-negotiating. It seems to stem from the fact there is a single session media field `media.remote_ip` set from the c= lines in the offer/answer. It is always set from last media description c= line so depending on the order of audio/video media descriptions either audio or video packet relay process is broken, i.e. audio or video stream is missing. The fundamental problem is that there is no provision for audio and video media addresses to be different. 
The fix is somewhat sizable, however, relatively straightforward otherwise. I preserve `media.remote_ip` field as default connection address that is used for both audio and video when they share the same address (or there is either audio or video media available but not both). I have added `media.remote_alt_video_ip`. It is set from media video description c= line only if different from the `media.remote_ip` and both audio and video media are available.  The fix also encompasses the change to the signature of `janus_sip_connect_sockets()` where it now requires distinct audio and video address, both can be optional.

I am not sure if the same fixes should be added to SIPre plugin. It seems to be a bit behind the SIP plugin feature wise? I am pretty happy to merge the fixes into SIPre if this request is approved.


lminiero:
Thanks for this effort! I'd ask you to please split the fixes in two separate pull requests: the re-INVITE fix in fact is much smaller, trivial, and could be merged right away; the other one I need more time to review, and test as well. I guess the easiest thing is to just create a new PR for the re-INVITE one, and leave this one as it is (so that it can be rebased to master when we merged the first).

From a quick glance, not sure why the properties are called `remote_ip` and `remote_alt_video_ip` now, instead of just `remote_audio_ip` and `remote_video_ip` that would be clearer and much less ambiguous. I'll try to do a deeper review of the rest as soon as I have some time :+1:

pawnnail:
Thanks for that, In hindsight, I should have kept one fix per PR. I have now removed the `autoaccept_reinvites` related commit from the PR and will open a separate one for the `autoaccept_reinvites` bug. I will also look into fixing the same bug in SIPre when I have a chance to test that plugin.

RE `remote_alt_video_ip` attribute vs `remote_video_ip` I don't have a strong argument for either. My motivation was to preserve the use semantics of the `remote_ip` attribute that applies when there is a single connection address for both video & audio, which is most of the time. It is only when video IP address differs I set the `remote_alt_video_ip` attribute. Now that you've said it's more intuitive to use pair `remote_audio_ip`/`remote_video_ip` I am pretty happy to change it and always use these attributes for audio/video event if they are same. It definitely looks cleaner. So please let me know and I will rework the fix.

lminiero:
I agree it looks cleaner! If you're going to update SIPre, could you update NoSIP as well? Media-wise, they basically all share the same code, since I based them both on the Sofa SIP plugin at the time. Thanks!

lminiero:
@pawnnail please let me know when you've made the change, and I'll look it at so that we can merge this soon, possibly before I go on vacation for the summer. I'm at a conference abroad this week, but I'll try to find time to review.

As to NoSIP and SIPre, feel free to ignore them for now. If this works for the SIP plugin, I can take care of updating those two myself.

pawnnail:
I will push the changes into a new branch in about 14 hours from now. I have  prepared the fixes for sipre and nosip as well. More fixes are coming but will submit a pull request for split audio/video first. Thanks!

lminiero:
Thanks! :+1:

pawnnail:
Have created new branch `split-audio-video-media-addresses` and new pull request #1727. I will close this pull request. Please take a look PR #1727. Thanks you.

PR #1721, opened 2019-08-02
Re-INVITE isn't offered to the called party

pawnnail:
When `autoaccept_reinvites=FALSE`  is added to accept request re-INVITE are not passed to the called party for handling. Added a fix so that the `autoaccept_reinvites=FALSE` is now respected and passed to the called party for handling.


lminiero:
Thanks, merging! :+1:

PR #1723, opened 2019-08-06
feat(textroom): listparticipants

mtltechtemp:
Everything's in the title.

CLA signed.

### DIFF
--- a/plugins/janus_textroom.c
+++ b/plugins/janus_textroom.c
@@ -21,7 +21,7 @@
  * a "setup" message, by which the user initializes the PeerConnection
  * itself. Apart from that, all other messages can be exchanged directly
  * via Data Channels. For room management purposes, though, requests like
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! I think `listparticipants` should ask for the `pin` if it's configured, though. It may be a password-protected room, and this way everyone can just see who's in even without it. I guess I'll have to do the same for other plugins, if it's missing there too.

mtltechtemp:
@lminiero Good catch!
I'm not comfortable enough with the codebase to do that, I'm afraid I might create security holes.
But, feel free to modify my PR.

lminiero:
> Edit: for AudioBridge, VideoRoom, no pin's required to list participants too.

Yep, that's why I said «I guess I'll have to do the same for other plugins, if it's missing there too» :smile: 

Anyway, that's fine. Since I can take care of that myself.

lminiero:
Merging! :+1:

PR #1727, opened 2019-08-06
Split audio video media addresses

pawnnail:
When SDP is processed single session attribute `remote_ip` is used for both audio and video connection addresses. It leads to a problem when audio and video addresses are different as specified by their respective media descriptions: either audio or video streams are missing.

The fix introduces dedicated connection addresses for audio and video, `remote_audio_ip `and `remote_video_ip` that are set from c= line of at the session level or media description level. The fix applied to SIP plugin as well as noSIP and SIPre plugins.


lminiero:
Thanks for this great effort! From a quick glance I can't spot anything wrong or out of place, I'll give it a test later, and merge if it looks fine :+1:

lminiero:
Looks good, thanks again for working on this, merging! :v:

PR #1729, opened 2019-08-08
Fix macro names to not use reserved identifiers (fixes #1725)

### LINKED ISSUE #1725: reserved identifier violation
I would like to point out that an identifier like “[`_JANUS_CORE_H`](https://github.com/meetecho/janus-gateway/blob/c6a1255ec5084afb8158cda972eeb09cec50372b/janus.h#L18 "Update candidate")” [does not fit](https://www.securecoding.cert.org/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier#DCL37-C.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28IncludeGuard%29 "Do not use identifiers which are reserved for the compiler implementation.") to the expected naming convention of the C language standard.
Would you like to adjust your selection for unique names?

lminiero:
As the title says. No feature change, just a fix on that. Please let me know if that works @elfring and if I missed anything.

### DIFF
--- a/apierror.h
+++ b/apierror.h
@@ -10,8 +10,8 @@
  * \ref core
  */
 
-#ifndef _JANUS_API_ERROR_H
-#define _JANUS_API_ERROR_H
+#ifndef JANUS_API_ERROR_H
+#define JANUS_API_ERROR_H
 
 /*! \brief Success (no error) */
 #define JANUS_OK								0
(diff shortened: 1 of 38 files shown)


elfring:
Thanks for your adjustment of affected include guards.

lminiero:
Merging then.

PR #1732, opened 2019-08-12
Better async managament of new mountpoints with temp map of IDs

lminiero:
As the title says. Not tested, so feedback welcome.


lminiero:
Merging.

PR #1734, opened 2019-08-12
Minor streaming fixes

ssalenik:
Just noticed a couple of possible small mistakes.


lminiero:
Thanks but I'm on vacation. I'll review when I'm back.

lminiero:
Thanks for catching those, merging! :+1:

PR #1737, opened 2019-08-14
Fix automatic reconnection in MQTT transport

feymartynov:
Paho calls `onSuccess` callback of `MQTTAsync_connect` call only on the first connection attempt.
In case of reconnection due to network problems or broker restart it wouldn't be called hence when back online the transport would not notify event handlers, send status message and resubscribe to the topic (which is necessary when `cleansession` flag is enabled in config).
Instead of `onSuccess` we [should](https://www.eclipse.org/lists/paho-dev/msg04046.html) set another callback with `MQTTAsync_setConnected` which is being called on reconnection and do all the above stuff there.

### DIFF
--- a/transports/janus_mqtt.c
+++ b/transports/janus_mqtt.c
@@ -157,9 +157,10 @@ typedef struct janus_mqtt_context {
 } janus_mqtt_context;
 
 /* Transport client methods */
+void janus_mqtt_client_connected(void *context, char *cause);
+void janus_mqtt_client_disconnected(void *context, MQTTProperties *properties, enum MQTTReasonCodes reasonCode);
 void janus_mqtt_client_connection_lost(void *context, char *cause);
 int janus_mqtt_client_message_arrived(void *context, char *topicName, int topicLen, MQTTAsync_message *message);
-void janus_mqtt_client_delivery_complete(void *context, MQTTAsync_token token);
 int janus_mqtt_client_connect(janus_mqtt_context *ctx);
 int janus_mqtt_client_reconnect(janus_mqtt_context *ctx);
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! I'm currently on vacation, I'll have a look when I'm back at the office.

lminiero:
Apologies, I've just returned from my vacation. I'll have a look either today or on Monday. Thanks!

transports/janus_mqtt.c
```diff
@@ -720,48 +780,13 @@ int janus_mqtt_client_connect(janus_mqtt_context *ctx) {
 }
 
 void janus_mqtt_client_connect_success(void *context, MQTTAsync_successData *response) {
-	janus_mqtt_client_connect_success_impl(context);
```
lminiero: Please don't keep methods empty. At the very least, there should be a comment explaining why we don't care about this callback.

transports/janus_mqtt.c
```diff
@@ -720,48 +780,13 @@ int janus_mqtt_client_connect(janus_mqtt_context *ctx) {
...
 }
 
 #ifdef MQTTVERSION_5
 void janus_mqtt_client_connect_success5(void *context, MQTTAsync_successData5 *response) {
-	janus_mqtt_client_connect_success_impl(context);
```
lminiero: See above.

lminiero:
@feymartynov please ping me as soon as you fix those couple of nits above, as I think that apart from that this is ready to be merged. Thanks!

feymartynov:
@lminiero updated the PR.

lminiero:
Thanks, merging! :+1:

PR #1740, opened 2019-08-15
Fixed handling of offerless reINVITE.

pawnnail:
The fix is straightforward and is around introducing two new session states marking re-INVITE transaction progress. If I've missed or got something wrong please let me know and I will make corrections as required.

"Update" request is generated by client in two cases:
1. when initiating a session update, i.e. sending a re-INVITE request
2. when handling a session update, i.e. sending a re-INVITE response
There is a problem with the second case. When a re-INVITE with SDP offer is received everything is fine. However, if an offerless re-INVITE is received it is responded to with a re-INVITE request, which is wrong. The problem seems to be that the "update" request with sdp of type "offer" is hard-wired with sending a re-INVITE request and sdp of "answer" is hard-wired with sending a re-INVITE response. That obviously doesn't work when handling offerless re-INVITEs.

To fix the problem the re-INVITE response is sent when there is a re-INVITE request is in progress, otherwise, a new re-INVITE request is sent. The decision is based on the state of the session that now
recognises client or server re-INVITE progress. Accordingly, two new states are introduced and maintained: `janus_sip_call_status_incall_reinviting` and `janus_sip_call_status_incall_reinvited` as extension to the existing state `janus_sip_call_status_incall`. Additionally, having the new states
is helpful in preventing "glare" type of scenarios, for example, where client "update" request collides with SIP re-INVITE request.

For reliability, the new states transitions are driven by nua state events "ready", "calling" or when receiving a re-INVITE request. Also, added method update_session_status() that logs session state changes to assist debugging.

Handling of some client request types such as "hangup", "update", "hold", "unhold", "info", "dtmf_info" are allowed in session state janus_sip_call_status_inviting that seems to be a copy-and-paste error. Have corrected this as well.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -758,10 +758,10 @@ static void update_session_status(janus_sip_session *session, janus_sip_call_sta
 	}
 }
 
-static int janus_sip_call_is_established(janus_sip_session *session) {
+static gboolean janus_sip_call_is_established(janus_sip_session *session) {
 	return (session->status == janus_sip_call_status_incall ||
 		session->status == janus_sip_call_status_incall_reinviting ||
-		session->status == janus_sip_call_status_incall_reinvited) ? 1 : 0;
+		session->status == janus_sip_call_status_incall_reinvited) ? TRUE : FALSE;
 }
 
 static void janus_sip_session_destroy(janus_sip_session *session) {
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -627,6 +627,8 @@ typedef enum {
 	janus_sip_call_status_inviting,
 	janus_sip_call_status_invited,
 	janus_sip_call_status_incall,
+	janus_sip_call_status_incall_reinviting,
+	janus_sip_call_status_incall_reinvited,
 	janus_sip_call_status_closing,
 } janus_sip_call_status;
 
@@ -640,6 +642,10 @@ static const char *janus_sip_call_status_string(janus_sip_call_status status) {
 			return "invited";
 		case janus_sip_call_status_incall:
 			return "incall";
+		case janus_sip_call_status_incall_reinviting:
+			return "incall_reinviting";
+		case janus_sip_call_status_incall_reinvited:
+			return "incall_reinvited";
 		case janus_sip_call_status_closing:
 			return "closing";
 		default:
@@ -743,6 +749,10 @@ static GHashTable *callids;
 static janus_mutex sessions_mutex = JANUS_MUTEX_INITIALIZER;
 
 static void janus_sip_srtp_cleanup(janus_sip_session *session);
+
+
+
+static void update_session_status(janus_sip_session *session, janus_sip_call_status new_status);
 
 static void janus_sip_session_destroy(janus_sip_session *session) {
 	if(session && g_atomic_int_compare_and_exchange(&session->destroyed, 0, 1))
@@ -1713,7 +1723,9 @@ void janus_sip_incoming_rtp(janus_plugin_session *handle, int video, char *buf, 
 			JANUS_LOG(LOG_ERR, "No session associated with this handle...\n");
 			return;
 		}
-		if(session->status != janus_sip_call_status_incall)
+		if(session->status != janus_sip_call_status_incall &&
+			session->status != janus_sip_call_status_incall_reinviting &&
+			session->status != janus_sip_call_status_incall_reinvited)
 			return;
 		/* Forward to our SIP peer */
 		if(video) {
@@ -1830,7 +1842,9 @@ void janus_sip_incoming_rtcp(janus_plugin_session *handle, int video, char *buf,
 			JANUS_LOG(LOG_ERR, "No session associated with this handle...\n");
 			return;
 		}
-		if(session->status != janus_sip_call_status_incall)
+		if(session->status != janus_sip_call_status_incall && 
+			session->status != janus_sip_call_status_incall_reinviting &&
+			session->status != janus_sip_call_status_incall_reinvited)
 			return;
 		/* Forward to our SIP peer */
 		if(video) {
@@ -1959,7 +1973,9 @@ static void janus_sip_hangup_media_internal(janus_plugin_session *handle) {
 	janus_mutex_unlock(&session->rec_mutex);
 	if(!(session->status == janus_sip_call_status_inviting ||
 		 session->status == janus_sip_call_status_invited ||
-		 session->status == janus_sip_call_status_incall)) {
+		 session->status == janus_sip_call_status_incall ||
+		 session->status == janus_sip_call_status_incall_reinviting ||
+		 session->status == janus_sip_call_status_incall_reinvited)) {
 		g_atomic_int_set(&session->establishing, 0);
 		g_atomic_int_set(&session->established, 0);
 		g_atomic_int_set(&session->hangingup, 0);
@@ -1973,7 +1989,7 @@ static void janus_sip_hangup_media_internal(janus_plugin_session *handle) {
 		session->media.autoaccept_reinvites = TRUE;
 		session->media.ready = FALSE;
 		session->media.on_hold = FALSE;
-		session->status = janus_sip_call_status_closing;
+		update_session_status(session, janus_sip_call_status_closing);
 		nua_bye(session->stack->s_nh_i, TAG_END());
 		g_free(session->callee);
 		session->callee = NULL;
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! I'm on vacation, though, so I'll only be able to review even I'm back.

lminiero:
In the meanwhile, can you fix the conflicts introduced by some recent merge?

plugins/janus_sip.c
```diff
@@ -744,6 +750,10 @@ static janus_mutex sessions_mutex = JANUS_MUTEX_INITIALIZER;
 
 static void janus_sip_srtp_cleanup(janus_sip_session *session);
 
+
+
+static void update_session_status(janus_sip_session *session, janus_sip_call_status new_status);
```
lminiero: Why the extra line feeds above?

plugins/janus_sip.c
```diff
@@ -1713,7 +1723,9 @@ void janus_sip_incoming_rtp(janus_plugin_session *handle, int video, char *buf,
 			JANUS_LOG(LOG_ERR, "No session associated with this handle...\n");
 			return;
 		}
-		if(session->status != janus_sip_call_status_incall)
+		if(session->status != janus_sip_call_status_incall &&
+			session->status != janus_sip_call_status_incall_reinviting &&
+			session->status != janus_sip_call_status_incall_reinvited)
```
lminiero: I see there are a lot of checks now that include all the three states, since `janus_sip_call_status_incall` is not enough anymore. Maybe it would be simpler and more readable to add a static helper method that does that? e.g., `janus_sip_is_incall(session)`?

plugins/janus_sip.c
```diff
@@ -4916,3 +4967,9 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 	return NULL;
 }
 
+static void update_session_status(janus_sip_session *session, janus_sip_call_status new_status) {
```
lminiero: Why not avoid the declaration above, and just define the method directly there right away? It's small enough, and the declaration isn't really needed.

As a side note, I guess you added this method to track the chanegs when debugging?

plugins/janus_sip.c
pawnnail: No reason, just a typo.

plugins/janus_sip.c
pawnnail: Agree. And, correct, to make it easier to follow the session state, which is useful beyond just debugging. I hope we can keep the output?

plugins/janus_sip.c
pawnnail: I did consider doing it and but thought that incorporating all three states into `janus_sip_is_incall` might sound misleading  or confusing because method `janus_sip_is_incall` would sound like a shortcut for a single state `janus_sip_call_status_incall`. So wasn't sure it would be a good idea. Maybe, we can name the new method `janus_sip_call_is_established`?

plugins/janus_sip.c
```diff
@@ -749,10 +749,20 @@ static GHashTable *callids;
...
+	if (session->status != new_status) {
+		JANUS_LOG(LOG_VERB, "[%s], session status change [%s]-->[%s]\n", session->account.username == NULL ? "null" : session->account.username, janus_sip_call_status_string(session->status), janus_sip_call_status_string(new_status));
+		session->status = new_status;
+	}
+}
 
-
-static void update_session_status(janus_sip_session *session, janus_sip_call_status new_status);
+static int janus_sip_call_is_established(janus_sip_session *session) {
```
lminiero: I think this should be boolean, not int.

lminiero:
I'll need some time to test this, as it's quite a lot of changes. Thanks for the quick fixes!

plugins/janus_sip.c
pawnnail: Have changed. Thanks.

plugins/janus_sip.c
```diff
@@ -1959,7 +2010,7 @@ static void janus_sip_hangup_media_internal(janus_plugin_session *handle) {
 	janus_mutex_unlock(&session->rec_mutex);
 	if(!(session->status == janus_sip_call_status_inviting ||
 		 session->status == janus_sip_call_status_invited ||
-		 session->status == janus_sip_call_status_incall)) {
+		 janus_sip_call_is_established(session))) {
```
lminiero: [nit] Looks like there's an extra space here?

plugins/janus_sip.c
```diff
@@ -1973,7 +2024,7 @@ static void janus_sip_hangup_media_internal(janus_plugin_session *handle) {
 		session->media.autoaccept_reinvites = TRUE;
 		session->media.ready = FALSE;
 		session->media.on_hold = FALSE;
-		session->status = janus_sip_call_status_closing;
+		update_session_status(session, janus_sip_call_status_closing);
```
lminiero: Can you rename this function to `janus_sip_call_update_status`? It makes it more consistent with the code style of other calls.

plugins/janus_sip.c
```diff
@@ -3040,7 +3103,8 @@ static void *janus_sip_handler(void *data) {
 			json_object_set_new(result, "event", json_string("hangingup"));
 		} else if(!strcasecmp(request_text, "recording")) {
 			/* Start or stop recording */
-			if(!(session->status == janus_sip_call_status_inviting || session->status == janus_sip_call_status_incall)) {
+			if(!(session->status == janus_sip_call_status_inviting || /* Presume it makes sense to start recording with early media? */
+				janus_sip_call_is_established(session))) {
```
lminiero: Since this part of the `if` is on a second line, it needs to be indented with one more tab per our code style.

plugins/janus_sip.c
```diff
@@ -3242,8 +3306,9 @@ static void *janus_sip_handler(void *data) {
 			json_object_set_new(result, "event", json_string("infosent"));
 		} else if(!strcasecmp(request_text, "message")) {
 			/* Send a SIP MESSAGE request: we'll only need the content */
-			if(!(session->status == janus_sip_call_status_inviting || session->status == janus_sip_call_status_incall)) {
-				JANUS_LOG(LOG_ERR, "Wrong state (not in a call? status=%s)\n", janus_sip_call_status_string(session->status));
+			if(!(session->status == janus_sip_call_status_inviting || 
+				janus_sip_call_is_established(session))) {
```
lminiero: See above on indentation guidelines.

plugins/janus_sip.c
```diff
@@ -3458,6 +3523,13 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				session->transaction = NULL;
 				if(g_atomic_int_get(&session->establishing) || g_atomic_int_get(&session->established))
 					gateway->close_pc(session->handle);
+			} else if (session->stack->s_nh_i == nh && callstate == nua_callstate_calling && session->status == janus_sip_call_status_incall) {
+				/* Have just sent re-INVITE */
+				update_session_status(session, janus_sip_call_status_incall_reinviting);
+			} else if (session->stack->s_nh_i == nh && callstate == nua_callstate_ready && 
+				(session->status == janus_sip_call_status_incall_reinviting || session->status == janus_sip_call_status_incall_reinvited)) {
```
lminiero: See above on indentation guidelines.

plugins/janus_sip.c
```diff
@@ -4513,6 +4595,9 @@ static void janus_sip_media_cleanup(janus_sip_session *session) {
 	}
 	/* Clean up SRTP stuff, if needed */
 	janus_sip_srtp_cleanup(session);
+
+	/* Media fields not cleaned up elsewhere */
+	janus_sip_media_reset(session);
```
lminiero: Any reason why this is a separate function, if it's only called here?

pawnnail: For all three plugins (writing it just here for the sake of brevity), it may be simpler to create a helper method that includes all that stuff, e.g., `janus_sipre_media_reset` or something like that. Doing this all inline looks ugly.

_Originally posted by @lminiero in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMTkyMDU0MDQ0OnYy/pull_request_review_threads/discussion_

plugins/janus_sip.c
pawnnail: Fixed.

plugins/janus_sip.c
pawnnail: Scanned and fixed all occurrences. BTW, there is seems to be inconsistency in spacing between if and () throughout the code. Anyway, changed all ifs so that it's not followed by a space.

plugins/janus_sip.c
pawnnail: Fixed as per prev comment.

pawnnail:
If there is anything else that looks amiss please let me know. Thanks for the review!

lminiero:
Looks good, now, thanks! I'll only tweak the text printed in `janus_sip_call_update_status` a bit, but something I can easily do myself. Merging, thanks again for this huge effort! :heart:

pawnnail:
That's great, thanks Lorenzo!

PR #1749, opened 2019-08-26
Let Janus keep recording a disabled mountpoint

atoppi:
This PR adds a new parameter to `disable` mountpoint API, named `stop_recording`.
By setting `stop_recording` to `false` while disabling a rtp streaming mountpoint, Janus will not close all the active recorders on that mountpoint, disabling only the packet relaying to the clients.
The `stop_recording` is `true` by default if omitted.

The scenario is a temporary "pause" for the live streaming while not losing the ability to record the incoming packets.

While adding this feature some other fixes/changes were done:
- close the recorders in case of a polling error from the mp sockets
- drop the `data` packets for a disabled mountpoint
- drop the `audio_rtcp` packets for a disabled mountpoint
- drop the `video_rtcp` packets for a disabled mountpoint

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -385,13 +385,29 @@ rtspiface = network interface IP address or device name to listen on when receiv
  *
 \verbatim
 {
-	"request" : "enable|disable",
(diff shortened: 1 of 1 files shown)


lminiero:
Looks good, merging :+1:

PR #1751, opened 2019-08-27
Add audio level dBov average to talk events in VideoRoom plugin

aconchillo:
This is has been handy to set the right audio level and active packets settings.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -634,6 +634,19 @@ room-<unique room ID>: {
 }
 \endverbatim
  *
+ * When configuring the room to request the ssrc-audio-level RTP extension,
+ * ad-hoc events might be sent to all publishers if \c audiolevel_event is
+ * set to true. These events will have the following format:
+ *
+\verbatim
+{
+	"videoroom" : <"talking"|"stopped-talking", whether the publisher started or stopped talking>,
+	"room" : <unique numeric ID of the room the publisher is in>,
+	"id" : <unique numeric ID of the publisher>,
+	"audio-level-dBov-avg" : <average value of audio level, 127=muted, 0='too loud'>
+}
+\endverbatim
+ *
  * An interesting feature VideoRoom publisher can take advantage of is
  * RTP forwarding. In fact, while the main purpose of this plugin is
  * getting media from WebRTC sources (publishers) and relaying it to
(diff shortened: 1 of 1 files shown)


lminiero:
Apologies, I've just returned from my vacation. I'll have a look either today or on Monday. In the meanwhile, can you fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so I can merge this if/when ready? Thanks!

plugins/janus_videoroom.c
```diff
@@ -632,6 +632,20 @@ room-<unique room ID>: {
 	"videoroom" : "event",
 	"configured" : "ok"
 }
+\endverbatim
+ *
+ * When configuring the room to request the ssrc-audio-level RTP
+ * extension, we saw that events could be sent to the publishers if
+ * \c audiolevel_event is set to true. These events will have the
+ * following format:
```
lminiero: This text ("we saw...") doesn't work for the documentation: here the feature must be explained, not "discovered". It should look something like this instead:

```
When configuring the room to request the ssrc-audio-level RTP
extension, ad-hoc events might be sent to all publishers if
\c audiolevel_event is set to true. These events will have the
following format:
```

plugins/janus_videoroom.c
```diff
@@ -4237,6 +4252,7 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
 					janus_mutex_lock(&videoroom->mutex);
 					json_t *event = json_object();
 					json_object_set_new(event, "videoroom", json_string(participant->talking ? "talking" : "stopped-talking"));
+					json_object_set_new(event, "audio-level-dBov-avg", json_real(audio_dBov_avg));
```
lminiero: Semantically, this should be added after "id": not really important, but since we encode properties in order and in other parts of the code we do the same, I'd prefer it to be that way, for style.

plugins/janus_videoroom.c
```diff
@@ -4246,6 +4262,7 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, int video, char
 					if(notify_events && gateway->events_is_enabled()) {
 						json_t *info = json_object();
 						json_object_set_new(info, "videoroom", json_string(participant->talking ? "talking" : "stopped-talking"));
+						json_object_set_new(event, "audio-level-dBov-avg", json_real(audio_dBov_avg));
```
lminiero: See above.

lminiero:
@aconchillo any news on review and CLA?

plugins/janus_videoroom.c
aconchillo: done.

plugins/janus_videoroom.c
aconchillo: done.

aconchillo:
CLA signed, rebased and changes done. Apologies for the delay.

plugins/janus_videoroom.c
aconchillo: done.

lminiero:
Thanks, merging then!

PR #1753, opened 2019-08-30
Adding properties to config must replace old ones (fixes #1738)

### LINKED ISSUE #1738: Janus takes the wrong "apisecret" when it's specified within config file and command line
Hello! 

When specifying an "apisecret" using command line (-a) and config file, Janus only accepts the one specified within the config file, although the docs say "Options passed through the command line have the precedence on those specified in the configuration file.". 

As soon as the secret is removed from the config file, the one from the command line starts being accepted.

Thanks!
Valentin

lminiero:
As the title says, this should fix a broken behaviour the config utils had. Adding properties would simply append them to a list, but would not remove the old property with the same name: while most of the times whis still worked, it could happen that the new value wouldn't be enforced. This fixes it by implicitly removing the property with the same name from the specified container, before adding it again.

### DIFF
--- a/config.c
+++ b/config.c
@@ -449,6 +449,10 @@ int janus_config_add(janus_config *config, janus_config_container *container, ja
 		return -1;
 	if(container != NULL && container->type != janus_config_type_category && container->type != janus_config_type_array)
 		return -2;
+	if(item->name) {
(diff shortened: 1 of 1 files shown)


lminiero:
Merging.

PR #1755, opened 2019-09-02
reads possible new nfo files

ebaioni:


ebaioni:
Sorry, PR opened against the wrong repo

PR #1756, opened 2019-09-02
SIP plugin: Hangup reason_text

ihusejnovic:
Added content of reason_text from SIP Reason header ([RFC3326](https://tools.ietf.org/html/rfc3326)) in SIP hangup event.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -76,7 +76,7 @@
 		"event" : "<name of the error event>",
 		"code" : <SIP error code>,
 		"reason" : "<SIP error reason>",
-		"reason_text" : "<SIP reason text; optional>"
+		"reason_header" : "<SIP reason header; optional>"
 	}
 }
 \endverbatim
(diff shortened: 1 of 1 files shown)


lminiero:
Isn't that what the reason string we already have is for?

ihusejnovic:
I didn’t want to change the existing behavior. Currently, the reason string is used for a SIP code phrase (text corresponding to status code).

Here is an example of the reason string and the reason_text (from header Reason) so you can see the difference:
`{"event":"hangup","code":487,"reason":"Request Terminated","reason_text":"Call completed elsewhere"},"call_id":"4515e0000ef5-5d6ccd2c-50b6236e-231838f0-e68abe2"}`

lminiero:
In that case, I think naming it `reason_text` is going to be ambiguous, due to the `reason` we already have there. Since you mentioned RFC3326, a better name might be `reason_header` or `sip_reason`?

ihusejnovic:
Yes, renamed to `reason_header`.

lminiero:
Thanks, merging then! :+1:

PR #1758, opened 2019-09-02
Added new announcement request to TextRoom

lminiero:
As the title says, this adds a new simple feature to the TextRoom plugin, to basically send room announcements. The syntax is quite basic, and very similar to `message`:

```
{
	"textroom" : "announcement",
	"room" : <unique numeric ID of the room this announcement will be sent to>,
	"secret" : "<room secret; mandatory if configured>",
	"text" : "<content of the announcement to send, as a string>"
}
```

The room secret is required since an announcement is seen as coming from the room, and not a specific user: as such, only those in control of the room should be able to send one. Users will see the announcement as an `announcement` event:

```
{
	"textroom" : "announcement",
	"room" : <room ID the announcement was sent to>,
	"date" : "<date/time of when the announcement was sent>",
	"text" : "<content of the announcement>"
}
```

Again, unlike `message` there's no `from` because the sender is the room itself. The patch also updates the demo code to show announcements properly in the chat box.

The feature is available through the Admin API as well, which should make it very easy to script announcements, e.g., using `curl`.

### DIFF
--- a/html/textroomtest.js
+++ b/html/textroomtest.js
@@ -163,6 +163,14 @@ $(document).ready(function() {
 											$('#chatroom').append('<p>[' + dateString + '] <b>' + participants[from] + ':</b> ' + msg);
 											$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;
 										}
+									} else if(what === "announcement") {
+										// Room announcement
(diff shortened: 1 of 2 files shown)


lminiero:
Merging.

PR #1763, opened 2019-09-09
Fix dynamic RTSP mountpoint creation hang when there are no free ports.

### LINKED ISSUE #1762: Configurable RTP port range may cause hanging when creating RTSP mountpoint via an API
Reproduction scenario:
1. Define rtp_port_range = "50000-50010" in janus.plugin.streaming.jcfg
2. Define 4 static rtsp streams in janus.plugin.streaming.jcfg
3. Run janus and create new rtsp mountpoint with an API, then destroy it with an API, **twice**
4. Try to create rtsp mountpoint with an API - call hangs

I was stepping through the code with the debugger and it seems that janus_streaming_create_fd() may loop infinitely because **if(!use_range && !quiet)** will never break.

remgodow:
Fixes #1762, also returns early when there is neither video nor audio sdp result.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -4619,8 +4619,10 @@ static int janus_streaming_create_fd(int port, in_addr_t mcast, const janus_netw
 		if(bind(fd, (struct sockaddr *)(&address), sizeof(struct sockaddr)) < 0) {
 			close(fd);
 			fd = -1;
-			if(!use_range && !quiet) {
+			if(!quiet){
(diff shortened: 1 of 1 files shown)


lminiero:
@remgodow thanks! Can you fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

lminiero:
@remgodow I'd really like to merge this, but until we get feedback from you and you sign the CLA, this won't be possible. I'll wait a few more days, and then I'll close the PR and address the issue on my own.

lminiero:
As anticipated, I ended up applying the fix myself. Thanks for bringing up the issue, closing.

PR #1765, opened 2019-09-10
Update README.md

indatawetrust:


lminiero:
Not sure why you added that, as it's not needed. Closing.

indatawetrust:
Cloning does not work without writing .git. If you don't care, don't add it.

lminiero:
I tried without the `.git` after seeing your pull request and it worked as expected for me. What did you test this on? If this is indeed not working in some environments, I'll add it :+1:

PR #1766, opened 2019-09-10
Minor edits of VideoRoom doc

giavac:


giavac:
> Thanks! But the "listener" parts should be removed, as that's a legacy syntax we're trying to get rid of, and we're only keeping in the code for backwards compatibility.

Ah, that explains a doubt I had. I'll create a new PR without that change.

PR #1771, opened 2019-09-13
Allow for capturing desktop audio when sharing screen

syntheticore:
Janus.js already has the option of adding a microphone audio track to screenshares, but capturing audio directly from the desktop wasn't possible. This feature only works in Chrome on Windows at the moment, but it's still nice being able to configure it.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -2207,7 +2207,7 @@ function Janus(gatewayCallbacks) {
 						// The new experimental getDisplayMedia API is available, let's use that
 						// https://groups.google.com/forum/#!topic/discuss-webrtc/Uf0SrR4uxzk
 						// https://webrtchacks.com/chrome-screensharing-getdisplaymedia/
-						navigator.mediaDevices.getDisplayMedia({ video: true })
+						navigator.mediaDevices.getDisplayMedia({ video: true, audio: media.captureDesktopAudio })
 							.then(function(stream) {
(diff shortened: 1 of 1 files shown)


lminiero:
What happens when the `audio` property is in the `getDisplayMedia` of other browsers? I want to make sure we won't cause exceptions or broken behaviours just to support a Chrome+Windows thing.

syntheticore:
I tested using FF and Chrome on osx and both ignored the audio property when it was set. In Chrome/Win it merely shows an additional checkbox to the user and still allows desktop audio to be deactivated. But you're right, I should do further testing on Windows and using Safari. I'll do that next week and make sure that it works everywhere and let you know. It shouldn't break existing code though, as it only makes a difference when somebody actually sets the new 'catureDesktopAudio' property deliberately.

lminiero:
@syntheticore thanks! I'll wait for that feedback and in case it looks fine I'll merge it :+1: 
I did a check with Chrome on Linux and it didn't complain (audio didn't work, of course, but that was to be expected, since it's a Windows only feature).

syntheticore:
I did some further testing now and could not find any problems in any of the major webRTC enabled browsers using any OS. I think this is safe to fly. We're actually using it in production already at my company.

lminiero:
Thanks for checking, merging then!

arpu:
is there any documentation of this chrome flag captureDesktopAudio ?

Issue #981, opened 2017-08-23
Simultaneous SIP calls


iclems:
I've been investigating managing simultaneous SIP calls. It seems reasonable to attach another SIP plugin as soon as you're on an active call to be able to receive incoming call requests separately during the current call. But it could actually be interesting to have an option for Janus SIP to not decline incoming calls if there's an active call but rather notify of "incoming call" to give us the opportunity to attach another plugin when this happens and not preventively in advance.

iclems:
Indeed that would need to be tested. I believed that's how CallKit works on iOS for instance: receive a notification, re-register and get a RE INVITE

andresico:
The way a solved the simultaneous SIP calls issue is by registering a new handler whenever I do or receive a new call. This way, you have always an available and registered handler for new incoming and outgoing calls. Busy handlers can be detached once the call is finished, but that forces you to create a new handler and register it, because detaching a handler unregisters the SIP peer. It would be great if we could detach a handler without unregistering the SIP peer.

shrhoads:
Right you just start a new instance of janus with a new server side session.

jmordica:
Is there a way to initiate the new instance without registering again?

Using this: https://github.com/meetecho/janus-gateway/issues/274

lminiero:
Hi all, it took a while but we finally started working on this. Please read the PR above to learn how to use it, test and provide feedback. The sooner I know it works, the sooner I can merge.

lminiero:
@iclems @andresico @shrhoads @jmordica have you all given up on Janus, or you just don't care about the feature anymore? :innocent:

jmordica:
Sorry for the late response here. I had to use frafos but now that this feature exists I’d like to try it and hopefully swap back over to Janus.

lminiero:
Cool! Looking forward to your feedback, if you'll start playing with this. There's some more updates coming to the SIP plugin as well, that I hope to be able to publish soon.

jmordica:
This may not be the best place to ask this but what other sip updates are you planning?

lminiero:
SIP transfer support, mostly.

lminiero:
Plus some other things I'll discuss at JanusCon next week.

jmordica:
Sweet! That will be awesome to have instead of having to use the PBX/Asterisk to handle it.

Thanks!

lminiero:
[...]

PR #1772, opened 2019-09-13
Support for simultaneous calls in SIP plugin

lminiero:
As the title says, this pull request is aimed at adding support for simultaneous calls to/from the same account in the SIP plugin. This was first asked for in #981, and has been a commonly asked feature since, e.g., to build more effective Janus-based softphones in the web, for contact/call centers, etc. This is part of a new series of developments, sponsored by a Colombian Company with 21 years of experience on developing Telecommunication Solutions.

Since this is part of a set of activities I've been working on related to the SIP plugin, this is actually based on another PR I published a couple of days ago (#1768), that is the one that implements SIP SUBSCRIBE. As such, testing this effort you can test that one as well (and I do need feedback on that too). The patch includes a documentation of the feature, in a dedicated section that can be generated when compiling the docs via doxygen and that I'll briefly describe here. 

Attaching to the SIP plugin with a Janus handle means creating a SIP stack on behalf of a user or application: this typically means registering an account, and being able to start or receive calls, handle subscriptions, and so on. This also means that, since in Janus each core handle can only be associated with a single PeerConnection, each SIP account is limited to a single call per time: if a user is in a SIP session already, and another call comes in, it's automatically rejected with a `486 Busy`.

While usually not a big deal, there are use cases where it might make sense to be able to support multiple concurrent calls, and maybe switch from one to the other seamlessly. This is possible in the SIP plugin using the so-called "helper sessions". Specifically, these helper sessions work under the assumption that there's a "master" session that is registered normally (the "regular" SIP plugin handle you've been using for years, that is), and that these helper sessions can simply be associated to that: any time another concurrent call is needed, if the "master" session is busy one of the "helpers" can be used; the more helper sessions are available, the more simultaneous calls can be established.

The way this works is simple:

1. you create a SIP session the usual way, and send a regular `register` there; this will be the master session, and will return a `master_id` property when successfully registered;
2. for each helper you want to add, you attach a new Janus handle to the SIP plugin, and send a `register` request with `type: "helper"` and providing the same `username` as the master, plus a `master_id` attribute referencing the main session;
3. at this point, the new helper is associated to the master, meaning it can be used to start new calls or receive calls exactly as the main session, and using the same account information, credentials, etc.

Notice that, as soon as the master unregisters, or the Janus handle it's on is detached, all the helper sessions associated to it are automatically torn down as well. Specifically, the plugin will forcibly detach the related handles. Should you need to register again, and want some helpers there too, you'll have to add them again. Notice that by default, if no helper is added, incoming calls when you're busy will still return a `486`: Janus **WILL NOT** automatically create these helpers for you; it's up **TO YOU** to create additional helpers **IN ADVANCE** if you anticipate you'll need more channels for your account. Putting this in bold because it's important, and I want to avoid repeating this over and over again at any "it doesn't work" report.

If you want to see this in practice, the SIP plugin demo has a "hidden" function you can invoke from the JavaScript console to play with helpers: calling the `addHelper()` function will add a new helper, and show additional controls. You can add as many helpers as you want.

Hopefully this should be enough to let you experiment with the feature: while I've been playing with this a bit and it seems to be working as expected, I didn't make extensive tests, and most importantly not in as many SIP scenarios as you guys may have put Janus into. If you're interested (and I know many of you are, it was a commonly asked feature), please give this a try and let me knows if it works as expected or if there's things we need to fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ docs/html/
 /janus
 /janus-cfgconv
 /janus-pp-rec
+/mjr2pcap
 /plugins/*.so
 /transports/*.so
 /events/*.so
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,6 @@ Makefile.in
 
 # OS X
 .DS_Store
+
+# CLion
+.idea
--- a/Makefile.am
+++ b/Makefile.am
@@ -513,6 +513,7 @@ endif
 
 if ENABLE_POST_PROCESSING
 bin_PROGRAMS += janus-pp-rec
+bin_PROGRAMS += mjr2pcap
 
 janus_pp_rec_SOURCES = \
 	postprocessing/pp-cmdline.c \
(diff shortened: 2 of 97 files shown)


lminiero:
Merging.

PR #1779, opened 2019-09-16
Typo fix in JANUSSDP.removePayloadType

l7s:


lminiero:
Good catch, thanks! I see you signed the CLA already, so merging :+1:

PR #1781, opened 2019-09-17
JavaScript logging improved

l7s:


lminiero:
Why not add something like an `int level` to the `duk__console_log_helper` definition? This would allow you to pass `level` instead of forcing `LOG_INFO`: this way, you can have `duk__console_log` pass `LOG_INFO`, `duk__console_warn` pass `LOG_WARN`, etc., something that would map our log errors more organically.

l7s:
Good point, here is the [commit](https://github.com/meetecho/janus-gateway/pull/1781/commits/ac4eaf17ad65da720498b6bf0a9693bd9ca2722d) maddping `console.info/warn/error` to `LOG_INFO/LOG_WARN/LOG_ERR` accordingly.

lminiero:
Thanks for the quick enhancement, merging then! :+1:

PR #1782, opened 2019-09-17
Hook Lua print function(s) to Janus logger

lminiero:
This is the Lua counterpart of what #1781 does for Duktape, and basically ensures that, if you use the integrated Lua logger, it ends up in the Janus logging mechanism, and not on regular stdout: this ensures, for instance, that if you're logging to file, the Lua prints end there too. Besides, the patch adds a `logger.warn` and a `logger.err` as well, that map to the warning/error flags in our logs (and so use the related prefix in the logs too); `logger.print` works exactly as before, but in now maps to `LOG_INFO`: so, should you want to reduce the verbosity to, let's say, `LOG_VERB`, you just have to change the `4` to `5` in `janus-logger.lua`.

Feedback welcome.

### DIFF
--- a/plugins/janus_lua.c
+++ b/plugins/janus_lua.c
@@ -383,6 +383,23 @@ static void *janus_lua_async_event_helper(void *data) {
 
 
 /* Methods that we expose to the Lua script */
+static int janus_lua_method_januslog(lua_State *s) {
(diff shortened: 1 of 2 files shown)


lminiero:
Merging.

PR #1783, opened 2019-09-17
moved destroySession connection condition

alexandreneves:
Moving the connection condition from destroySession allows handles to be cleaned locally even if there's no connection. When a connection is (temporarily) lost and destroySession is called (using the public method), it should clean the handles anyway, in case the connection gets restored. This avoids possible leaks with unexpected active PC.


lminiero:
Makes sense, thanks for the fix! Can you fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

alexandreneves:
> Makes sense, thanks for the fix! Can you fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

Done : ) thanks

lminiero:
Thanks, merging then! :+1:

PR #1784, opened 2019-09-19
SIP plugin: add local interface for SDP binds

razvancrainea:
Add the `local_sdp_ip` parameter for the SIP plugin that allows Janus
to bind the media stack on a different interface than `local_ip`.

### DIFF
--- a/conf/janus.plugin.sip.jcfg.sample
+++ b/conf/janus.plugin.sip.jcfg.sample
@@ -5,7 +5,7 @@ general: {
 
 	# Specify which local IP address to bind for the media stack.
 	# If not set it will be automatically set to the value of local_ip
-	#local_sdp_ip = "1.2.3.4"
+	#local_media_ip = "1.2.3.4"
 
 	# Specify which (public) IP address to advertise in the SDP.
 	# If not set, the value above or anything autodetected will be used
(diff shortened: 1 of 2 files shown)


razvancrainea:
I've noticed that the `local_ip` is a common parameter for all SIP plugins (noSIP, SIPre), but I've only added it for the SIP plugin. If necessary, I can also add it for the other plugins as well.

lminiero:
@razvancrainea thanks! No worries about the other plugins, I can take care of porting the same changes to those as well once we merge :+1: 
Can you fill in the form and sign our [CLA](https://cla.conf.meetecho.com/janus/)? I'm out of office today, but the patch looks fine, and so I can merge when that's done.

razvancrainea:
I've just signed it, thanks!

lminiero:
Thanks! Looking at the patch, the only comment I have is of a semantic nature: rather than `local_sdp_ip`, I'd call this something like `local_media_ip`, since what it really does is specifying which interface/IP the plugin should bind to for media. It *might* end up being used in the SDP too, but we have another property called `sdp_ip` that explicitly can override that, and so calling this `local_sdp_ip` might sound ambiguos, expecially if they're both set. What do you think?

razvancrainea:
I completely agree, it does indeed only specify the media bind, which is more generic than SDP. I will push a new commit for this soon, thanks for the review!

lminiero:
Thanks for the quick fix, merging then!

PR #1785, opened 2019-09-19
sdp-utils: check that janus_sdp_get_codec_rtpmap succeeded

tmatth:
We've seen the following in our logs in production:
```
[Wed Sep 18 21:33:23 2019] ^[[31m[ERR]^[[0m [sdp-utils.c:janus_sdp_get_codec_rtpmap:763] Unsupported codec 'none'
```
While we're not sure where that's coming from (or what the full consequences are), I audited the uses of `janus_sdp_get_codec_rtpmap` and noticed a few places where a NULL pointer could be added to the list of attributes.

### DIFF
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -1351,46 +1351,52 @@ janus_sdp *janus_sdp_generate_answer(janus_sdp *offer, ...) {
 			/* Add the related attributes */
 			if(m->type == JANUS_SDP_AUDIO) {
 				/* Add rtpmap attribute */
-				janus_sdp_attribute *a = janus_sdp_attribute_create("rtpmap", "%d %s", pt, janus_sdp_get_codec_rtpmap(codec));
-				am->attributes = g_list_append(am->attributes, a);
-				/* Check if we need to add a payload type for DTMF tones (telephone-event/8000) */
-				if(audio_dtmf) {
-					int dtmf_pt = janus_sdp_get_codec_pt(offer, "dtmf");
-					if(dtmf_pt >= 0) {
-						/* We do */
-						am->ptypes = g_list_append(am->ptypes, GINT_TO_POINTER(dtmf_pt));
-						janus_sdp_attribute *a = janus_sdp_attribute_create("rtpmap", "%d %s", dtmf_pt, janus_sdp_get_codec_rtpmap("dtmf"));
+				const char *codec_rtpmap = janus_sdp_get_codec_rtpmap(codec);
+				if(codec_rtpmap) {
+					janus_sdp_attribute *a = janus_sdp_attribute_create("rtpmap", "%d %s", pt, codec_rtpmap);
+					am->attributes = g_list_append(am->attributes, a);
+					/* Check if we need to add a payload type for DTMF tones (telephone-event/8000) */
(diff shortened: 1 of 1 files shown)


sdp-utils.c
```diff
@@ -1371,7 +1374,9 @@ janus_sdp *janus_sdp_generate_answer(janus_sdp *offer, ...) {
 				}
 			} else {
 				/* Add rtpmap attribute */
-				janus_sdp_attribute *a = janus_sdp_attribute_create("rtpmap", "%d %s", pt, janus_sdp_get_codec_rtpmap(codec));
+				const char *codec_rtpmap = janus_sdp_get_codec_rtpmap(codec);
+				if(!codec_rtpmap) continue;
```
lminiero: Can you split this on two lines, per our code style?

lminiero: As a side note, why is there a `continue` for video, while for audio just a conditional? The `continue` as it is looks dangerous too, since this is in a `while` and you're not updating the pointer.

tmatth: > As a side note, why is there a `continue` for video, while for audio just a conditional? The `continue` as it is looks dangerous too, since this is in a `while` and you're not updating the pointer.

Yeah the continue is bad, re-pushing.

lminiero: Thanks for the quick fix! I'll need to make a few tests before I can merge, specifically in cases where no common codec is found. We recently fixed some issues in that scenario, and I want to be sure there won't be a regression. I'll do that when I get back to the office tomorrow.

lminiero:
I still see the continue in the changes? Is it github not showing things properly?

tmatth:
> I still see the continue in the changes? Is it github not showing things properly?

Whoops, thought I'd pushed. Should be there now.

sdp-utils.c
```diff
@@ -1351,8 +1351,11 @@ janus_sdp *janus_sdp_generate_answer(janus_sdp *offer, ...) {
 			/* Add the related attributes */
 			if(m->type == JANUS_SDP_AUDIO) {
 				/* Add rtpmap attribute */
-				janus_sdp_attribute *a = janus_sdp_attribute_create("rtpmap", "%d %s", pt, janus_sdp_get_codec_rtpmap(codec));
-				am->attributes = g_list_append(am->attributes, a);
+				const char *codec_rtpmap = janus_sdp_get_codec_rtpmap(codec);
+				if(codec_rtpmap){
```
lminiero: Missing space (code style).

sdp-utils.c
```diff
@@ -1371,26 +1374,29 @@ janus_sdp *janus_sdp_generate_answer(janus_sdp *offer, ...) {
 				}
```
lminiero: Shouldn't the two above be part of the `if(codec_rtpmap)` too, especially `audio_fmtp` since it relies on a payload type that may not be associated to a codec at all? I see you did that for video.

tmatth: I wasn't sure for the DTMF (could you have an invalid/disabled audio codec but an independent  DTMF line)? You're definitely right about the fmtp though.

lminiero:
Will test later, thanks!

lminiero:

PR #1787, opened 2019-09-20
Mute participants

klanjabrik:


lminiero:
Nice! Definitely useful to have, especially considering it would be a synchronous request available via Admin API as well. I'll review and test later. In the meanwhile, make sure you compile and sign our [CLA](https://cla.conf.meetecho.com/janus/), which is a prerequisite to merging.

lminiero:
The feature seems to be working nicely, but I think there's potential ambiguity from a semantic perspective. For instance, I sent a `request:"mute"` request with no `muted` property, and it did nothing: that's because apparently the default value for `muted` is `false`, but it makes no sense considering the call is called `mute`.

I'd change the syntax to get rid of the `muted` property, and use two different values for `request` instead: if it's called `mute`, you mute the user, otherwise if it's `unmute` you unmute them. Much simpler and cleaner from an API perspective. You wouldn't need to change the code that much either, as it would simply be a matter of changing the main check to handle the requests:

    } else if(!strcasecmp(request_text, "mute") || !strcasecmp(request_text, "unmute"))) {

and then instead of `json_is_true(muted)` look at the value of `request_text` to see what to do.

klanjabrik:
Hi @lminiero . I just changed the file as you suggested. Please review dan test it. Thanks!

plugins/janus_audiobridge.c
```diff
@@ -2329,6 +2354,120 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+		if(participant == NULL) {
+			janus_mutex_unlock(&audiobridge->mutex);
+			janus_refcount_decrease(&audiobridge->ref);
+			JANUS_LOG(LOG_ERR, "No such user %"SCNu64" in room %"SCNu64"\n", user_id, room_id);
+			error_code = JANUS_AUDIOBRIDGE_ERROR_NO_SUCH_USER;
+			g_snprintf(error_cause, 512, "No such user %"SCNu64" in room %"SCNu64, user_id, room_id);
+			goto prepare_response;
+		}
+		
+		participant->muted = muted;
```
lminiero: I'd add a further check before doing this. If we're trying to perform an operation that has already been done (i.e., trying to mute an already muted user, or trying to unmute a user that is not mute), then we should do nothing, and definitely not send another event to other participants; the only thing we should so is send a success back.

As such, I'd add something like

```
if(participant->muted == muted) {
	/* Nothing to do, just prepare response */
	response = json_object();
	json_object_set_new(response, "audiobridge", json_string("success"));

	/* Done */
	janus_mutex_unlock(&audiobridge->mutex);
	janus_refcount_decrease(&audiobridge->ref);
	goto prepare_response;
}
```

This makes the requests idempotent.

klanjabrik: Ah yes, sorry for that. Should've check it also.

klanjabrik: Hi @lminiero . I just added the condition.

lminiero:
Thanks, merging then! :+1:

PR #1788, opened 2019-09-21
Gelf event handler

mirkobrankovic:
Simple TCP/UDP socket event handler to gelf interface.
Config file is simple, you need filter, host, port

events
will look like this in Graylog interface

Still need to think about message structure and custom fields.


lminiero:
Cool, thanks! You may want to rename this PR to event handler, rather than forwarder, though, as I thought this was media-related.

Being UDP based, how does this handle big events? SDP offer/answers can go way beyond the MTU, and it will be even worse once we merge the unified-plan branch. I'm not familiar with GELF, so I'll look into it and review the code.

mirkobrankovic:
Indeed, there is udp chunking with magic number and packet sequence:
https://docs.graylog.org/en/3.1/pages/gelf.html#gelf-via-udp
Need to look at that also

lminiero:
Yep, the chunking part looks useful, and so does compression (but that would be true for other plugins as well, I'll have to check that). Do you plan to add that to the PR?

mirkobrankovic:
I will give it a try :D

mirkobrankovic:
@lminiero can you help me a bit because I got stuck at this chunking and can't resolve it.
Here is the test I did:
I have a simple [server.c](https://gist.github.com/mirkobrankovic/d2392aea519a3f9e3d2a54cea7b87494) listener and [client.c](https://gist.github.com/mirkobrankovic/d18bc4af88cef1ecd0cc579e845b80b4) that is suppose to take 2404 length SDP and chunk it into 512 packets and send it to server.

You can see the result in [this result](https://gist.github.com/mirkobrankovic/9c7f9be4b24aeba75e66bbb4c889bd7d)
So I calculate that I need 4*512 and last 356 length.

1. Problem is that my "empty" chunk is not empty and reports 518 length at first.
2. Problem is that first iteration makes a good chunk, except it has additional 6 bytes (for example:L �FE	�), but then in second, buf pointer is moved correctly to position 512, but chunk stays the same after strncpy, only those 6 last bytes are removed. Third iteration for chunk var is actually what was suppose to happen in second one.

I'm doing something wrong with this pointer or this 6 bytes are messing everything up, but i'm not sure where they come from.

Hope you have time to help,
Thanks,
Mirko

lminiero:
Probably some strings manipulation error? Sorry, can't look into this right now.

atoppi:
Hi @mirkobrankovic,
your client loop should be like this
```
    //--------------NETWORK PART---------------------------

    int total = len / MAX_GELF_MSG_LEN + 1;
    fprintf(stderr, "Total packets: %d\n", total);
    fprintf(stderr, "----------------------------------------------------------------------------------------------------------\n");

    int offset = 0;
    int sent = 0;
    char *rnd = randstring(8);

    //fprintf(stderr, "Buffer: %s\n", buffer);

    for (int i = 0; i < total; i++) {
        fprintf(stderr, "Packet No: %d\n", i+1);
        fprintf(stderr, "offset: %d\n", offset);
        
        int bytesToSend = offset + MAX_GELF_MSG_LEN < len ? MAX_GELF_MSG_LEN : len - offset;
        fprintf(stderr, "Bytes to send: %d\n", bytesToSend);
        
        sent = write(sockfd, buf+offset, bytesToSend);
        if (sent < 0) {
            fprintf(stderr, "ERROR writing to socket\n");
            printf(" Value of errno: %s\n ", strerror(errno));
            return 1;
        }
        
        fprintf(stderr, "Bytes sent: %d\n", sent);
        offset += sent;
        fprintf(stderr, "----------------------------------------------------------------------------------------------------------\n");
    }
```

mirkobrankovic:
@atoppi thanks a lot for pointers :)

mirkobrankovic:
TCP is working, now to fix UDP.
I will keep it running on our servers and logging to Graylog to see if something bad will happen :D

lminiero:
If this Gelf thing supports gzip, you may want to have a look at the utils I added recently for the purpose, which should help reduce the size of data on UDP (unless Gelf has its own fragmentation mechanisms).

mirkobrankovic:
@lminiero thanks I saw it the other day and thought about it.
they do have custom [UDP chunking](https://docs.graylog.org/en/3.1/pages/gelf.html#chunking) imitating what TCP does, but for [compression](https://docs.graylog.org/en/3.1/pages/gelf.html#compression) they do recommend Gzip or Zlib

mirkobrankovic:
TCP compression is not supported cause of the '\0' but for UDP it is possible, just need to see if then I need to stop chunking or not ....

mirkobrankovic:
wow it is all connected :D

mirkobrankovic:
Tested all cases and there is the network sniff:
https://pastebin.com/47KqcBus
@lminiero I would like your code x-ray input, I know you have at least 10 improvements :D

lminiero:
@mirkobrankovic ack, I'll try to review when I have some time :+1: 
Probably won't happen before next week though.

mirkobrankovic:
I think i have a mem leak, says valgrind:
```
==142035== 13 bytes in 1 blocks are definitely lost in loss record 31 of 683
==142035==    at 0x40307F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==142035==    by 0x5124718: g_malloc (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==142035==    by 0x513D4EE: g_strdup (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==142035==    by 0x10F7788A: janus_gelfevh_init (in /opt/janus/lib/janus/events/libjanus_gelfevh.so.0.0.0)
==142035==    by 0x417880: main (in /opt/janus/bin/janus)
```

mirkobrankovic:
now only reported definite mem leak is in videoroom:
```
==151507== 72 (24 direct, 48 indirect) bytes in 1 blocks are definitely lost in loss record 179 of 307
==151507==    at 0x40307F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==151507==    by 0x5124718: g_malloc (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==151507==    by 0x513B922: g_slice_alloc (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==151507==    by 0x511B3C3: g_list_append (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==151507==    by 0x42002D: janus_config_get_categories (in /opt/janus/bin/janus)
==151507==    by 0x127AB72A: janus_videoroom_init (in /opt/janus/lib/janus/plugins/libjanus_videoroom.so.0.0.0)
==151507==    by 0x414D81: main (in /opt/janus/bin/janus)
```

mirkobrankovic:
Seems to be working for more then 20-30 days.
I'll try to push it to Production with higher load of messages

lminiero:
Is this Gelf server open source?

mirkobrankovic:
It is part of the Graylog, so this is the licence:
https://github.com/Graylog2/documentation/blob/3.1/LICENSE.md

mirkobrankovic:
Damn, I resetted my master in order to start new thing and this PR gets closed :d

lminiero:
> It is part of the Graylog, so this is the licence:
> https://github.com/Graylog2/documentation/blob/3.1/LICENSE.md

That looks like the license for the documentation, though, not the server.

mirkobrankovic:
sorry :)
https://github.com/Graylog2/graylog2-server/blob/master/COPYING
GNU GPL 3.0

lminiero:
Thanks the clarification! Will you let us know when you consider this done, so that we can start a review process?

mirkobrankovic:
I do consider it done now, since it supports tcp and udp (chunking and compression)

lminiero:
Thanks for the heads up, I'll try and review ASAP :+1:  (may be after FOSDEM though)

configure.ac
```diff
@@ -1028,6 +1038,9 @@ echo "Event handlers:"
 AM_COND_IF([ENABLE_SAMPLEEVH],
 	[echo "    Sample event handler:  yes"],
 	[echo "    Sample event handler:  no"])
+AM_COND_IF([ENABLE_GELFEVH],
+	[echo "    Gelf event handler:    yes"],
+	[echo "    Gelf event handler:    no"])
```
lminiero: Just a nit, but this should appear at the end of the list. The order is not alphabetic, but depending on when they were added.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
+/*! \file   janus_gelfevh.c
+ * \author Lorenzo Miniero <lorenzo@meetecho.com>
```
lminiero: This should be you, not me :wink:

configure.ac
lminiero: As a side note, isn't it called GELF (all-caps)?

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+ * \copyright GNU General Public License v3
+ * \brief  Janus GelfEventHandler plugin
+ * \details  This is a trivial event handler plugin for Janus, which is only
+ * there to showcase how you can handle an event coming from the Janus core
+ * or one of the plugins. This specific plugin forwards every event it receives
+ * to a web server via an UDP request, using libcurl.
```
lminiero: The description seems incorrect (possibly a copy of the sample event handler?). You should clarify exactly what this handler is supposed to talk to, possibly with links to Graylog, the GELF format, etc. If multiple protocols are supported (I think you mentioned UDP and TCP?) and there are custom features (compression?) they should be mentioned too. Whatever you write here is converted to a documentation web page with Doxygen, and eventually pushed to the docs online.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+ *
+ * \ingroup eventhandlers
+ * \ref eventhandlers
+ */
+
+#include "eventhandler.h"
+
+#include <math.h>
+// #include <curl/curl.h>
```
lminiero: Is libcurl used or not? It's linked in the Makefile, and checked in configure.ac, but you're not including it here. As a side note, please don't use `//` for comments, but use `/*  */` instead (code style).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+	if(!event || event == &exit_event)
+		return;
+	json_decref(event);
+}
+
+/* Gelf backend to send the events to */
+static char *backend = NULL;
+static char *port = NULL;
+static int sockfd;
+//static struct sockaddr_in servaddr;
```
lminiero: Same here on comments style (please fix wherever they appear like this, or remove the lines if they're not needed anymore).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	memset(&servaddr, 0, sizeof(servaddr));
+
+	servaddr.sin_family = AF_INET;
+	servaddr.sin_port = htons(atoi(port));
+	servaddr.sin_addr.s_addr = inet_addr(host);
+
+	if(connect(sockfd, (struct sockaddr *)&servaddr, sizeof(servaddr)) < 0) {
+		JANUS_LOG(LOG_WARN, "Connect to Gelf host failed\n");
```
lminiero: See previous note on `Gelf` vs. `GELF`, especially considering the output doesn't seem consistent throughout the rest of the code (it's all caps just a couple of lines below, for instance).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+		janus_network_address_to_string_buffer(&addr, &addr_buf) != 0) {
+		if(res)
+			freeaddrinfo(res);
+		JANUS_LOG(LOG_ERR, "Could not resolve address (%s)...\n", backend);
+		return;
+	}
+	const char *host = g_strdup(janus_network_address_string_from_buffer(&addr_buf));
+	freeaddrinfo(res);
+
+    if((sockfd = socket(AF_INET, SOCK_DGRAM, 0)) < 0 ) {
```
lminiero: Indentation seems broken here (spaces?)

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+/* Plugin implementation */
+static void janus_gelfevh_connect(void) {
+	struct addrinfo *res = NULL;
+	janus_network_address addr;
+	janus_network_address_string_buffer addr_buf;
+	struct sockaddr_in servaddr;
+
+	if(getaddrinfo(backend, NULL, NULL, &res) != 0 ||
+		janus_network_address_from_sockaddr(res->ai_addr, &addr) != 0 ||
+		janus_network_address_to_string_buffer(&addr, &addr_buf) != 0) {
```
lminiero: The two lines above should have another tab of indentation, since they belong to the `if` above.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+	JANUS_LOG(LOG_INFO, "Connected to GELF backend: [%s:%s]\n", host, port);
+}
+
+static int janus_gelfevh_send(char *message) {
+	//JANUS_LOG(LOG_WARN, "Sending event to GELF: %s\n", message);
+	if(!message) {
+		JANUS_LOG(LOG_WARN, "Message is NULL, not sending to Gelf!\n");
+		return -1;
+	}
+	if (write(sockfd, message, MAX_GELF_LOG_LEN) < 0) {
```
lminiero: Code style: please remove space between `if` and bracket.

lminiero: Why are you trying to send `MAX_GELF_LOG_LEN` bytes, when the message is very likely shorter?

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+		/* Events */
+		if(json_object_get(request, "events"))
+			req_events = json_string_value(json_object_get(request, "events"));
+		/* Backend stuff */
+		if(json_object_get(request, "backend"))
+			req_backend = json_string_value(json_object_get(request, "backend"));
+		if(json_object_get(request, "port"))
+			req_port = json_string_value(json_object_get(request, "port"));
+		if(req_backend && req_port) {
+			/* Invalid backend address, port */
```
lminiero: Not sure why this is a failed check..? Did you mean to place a `!` before one of the two?

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+			error_code = JANUS_GELFEVH_ERROR_INVALID_ELEMENT;
+			g_snprintf(error_cause, sizeof(error_cause), "Invalid backend URI '%s:%s'", req_backend, req_port);
+			goto plugin_response;
+		}
+		/* If we got here, we can enforce */
+		if(req_events)
+			janus_events_edit_events_mask(req_events, &janus_gelfevh.events_mask);
+		if(req_backend) {
+			janus_mutex_lock(&evh_mutex);
+			if(req_backend && req_port) {
```
lminiero: This will never be executed, because of the check above.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+				case JANUS_EVENT_TYPE_CORE:
+				case JANUS_EVENT_TYPE_EXTERNAL:
+					short_message = "JANUS_EVENT_TYPE_CORE";
+					break;
+				default:
+					JANUS_LOG(LOG_WARN, "Unknown type of event '%d'\n", type);
+					short_message = "UNKNOWN_JANUS_EVENT";
+					break;
+			}
+			output = event;
```
lminiero: This is going to cause issues, since you're modifying the event, and then extracting portions of the original event and setting them on itself. Just create a new object here (`output = json_object()`), and copy what you need from the original event if this needs to be in a format Graylog expects.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+		if(event == NULL)
+			continue;
+		if(event == &exit_event)
+			break;
+		output = NULL;
+
+		while(TRUE) {
+			/* Handle event */
+			int type = json_integer_value(json_object_get(event, "type"));
+			switch(type) {
```
lminiero: There's no need for all that: you can just use `janus_events_type_to_label` or `janus_events_type_to_name` from `events.h`(unless they really need to be in the format you're using here?)

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+				struct timeval t;
+				gettimeofday(&t, NULL);
+				double micro_timestamp = (double)(1000000 * t.tv_sec + t.tv_usec) / 1000000;
+				json_object_set(output, "timestamp", json_real(micro_timestamp));
+			}
+			json_object_set(output, "version", json_string("1.1"));
+			json_object_set(output, "host", json_string("janus"));
+			json_object_set(output, "level", json_integer(1));
+			json_object_set(output, "short_message", json_string(short_message));
+			json_object_set(output, "full_message", json_object_get(event, "event"));
```
lminiero: I believe that when you use `json_object_set`, a reference is already added to that, so you shouldn't have issues when switching to a new object for `output`. Anyway, you may want to make sure that `full_message` is indeed there when you make that change.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+		if(res)
+			freeaddrinfo(res);
+		JANUS_LOG(LOG_ERR, "Could not resolve address (%s)...\n", backend);
+		return;
+	}
+	const char *host = g_strdup(janus_network_address_string_from_buffer(&addr_buf));
+	freeaddrinfo(res);
+
+    if((sockfd = socket(AF_INET, SOCK_DGRAM, 0)) < 0 ) {
+		JANUS_LOG(LOG_ERR, "Socket creation failed");
```
lminiero: Please add info on why socket operations fail (see usage of `errno` and `strerror` from other plugins).

events/janus_gelfevh.c
lminiero: See above on socket errors.

events/janus_gelfevh.c
lminiero: Also, not sure why you're using `write` instead of `send`, which you can use now that the socket is connected. Reading around, it looks like `send` gives more info when it fails (which we do care about, see above on socket errors).

events/janus_gelfevh.c
lminiero: Side note, no support for TCP then? Or does Graylog only support UDP? What about compression via gzip, which we now provide as part of core utils?

events/janus_gelfevh.c
```diff
@@ -0,0 +1,476 @@
...
+			if(janus_gelfevh_send(event_text) < 0) {
+				JANUS_LOG(LOG_WARN, "Couldn't send event to GELF reconnect ... ?: %s\n", event_text);
+			}
+			break;
+		}
+		/* Done, let's unref the event */
+		json_decref(output);
+		output = NULL;
+	}
+	JANUS_LOG(LOG_VERB, "Leaving Gelf Event handler thread\n");
```
lminiero: See Gelf vs GELF.

events/janus_gelfevh.c
lminiero: Just noticed the other handlers don't have a good description either... I'll fix those to give a good example :smile:

events/janus_gelfevh.c
mirkobrankovic: no was not necessary since it would help only for tcp support, and it is an easy send srteam function

conf/janus.eventhandler.gelfevh.jcfg.sample
```diff
@@ -0,0 +1,21 @@
+# This configures the gelf event handler. Appending necessary headers 
+# and sending messages via TCP or UDP (maybe quic soon :D)
```
lminiero: I missed this part here: `gelf` should be `GELF`. I understand it's cute, but the QUIC reference and the smiley should be removed :wink:

configure.ac
```diff
@@ -617,10 +626,12 @@ AC_CHECK_LIB([nanomsg],
              ])
 AM_CONDITIONAL([ENABLE_RABBITMQ], [test "x$enable_rabbitmq" = "xyes"])
 AM_CONDITIONAL([ENABLE_RABBITMQEVH], [test "x$enable_rabbitmq_event_handler" = "xyes"])
+AM_CONDITIONAL([ENABLE_GELFEVH], [test "x$enable_gelf_event_handler" = "xyes"])
 AM_CONDITIONAL([ENABLE_MQTT], [test "x$enable_mqtt" = "xyes"])
 AM_CONDITIONAL([ENABLE_MQTTEVH], [test "x$enable_mqtt_event_handler" = "xyes"])
 AM_CONDITIONAL([ENABLE_NANOMSG], [test "x$enable_nanomsg" = "xyes"])
 AM_CONDITIONAL([ENABLE_NANOMSGEVH], [test "x$enable_nanomsg_event_handler" = "xyes"])
+AM_CONDITIONAL([ENABLE_GELFEVH], [test "x$enable_gelf_event_handler" = "xyes"])
```
lminiero: There are two of these (copy/paste instead of cut/paste?)

configure.ac
```diff
@@ -1028,6 +1039,9 @@ echo "Event handlers:"
 AM_COND_IF([ENABLE_SAMPLEEVH],
 	[echo "    Sample event handler:  yes"],
 	[echo "    Sample event handler:  no"])
+AM_COND_IF([ENABLE_GELFEVH],
+	[echo "    Gelf event handler:    yes"],
+	[echo "    Gelf event handler:    no"])
```
lminiero: Same here, it's there twice (only the last one should remain).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
+/*! \file   janus_gelfevh.c
+ * \author Mirko Brankovic <mirkobrankovic@gmail.com>
+ * \copyright GNU General Public License v3
+ * \brief  Janus GelfEventHandler plugin
+ * \details  This is a GELF event handler plugin for Janus, which is suppose
```
lminiero: s/is suppose/is supposed

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
+/*! \file   janus_gelfevh.c
+ * \author Mirko Brankovic <mirkobrankovic@gmail.com>
+ * \copyright GNU General Public License v3
+ * \brief  Janus GelfEventHandler plugin
+ * \details  This is a GELF event handler plugin for Janus, which is suppose 
+ * to send json events to GELF (Graylog logger), Necessary headers are prepended.
```
lminiero: Comma should be a full stop?

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+ * For sending, you can use TCP which is not recomanded in case there will be
+ * a lot of messages. There is also UDP support, but you need to limit the payload 
+ * size with max_message_len + remember to leave space for 12 bytes for special
```
lminiero: s/+ remember/and remember
s/leave space/leave room

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+ * \copyright GNU General Public License v3
+ * \brief  Janus GelfEventHandler plugin
+ * \details  This is a GELF event handler plugin for Janus, which is suppose 
+ * to send json events to GELF (Graylog logger), Necessary headers are prepended.
+ * For sending, you can use TCP which is not recomanded in case there will be
+ * a lot of messages. There is also UDP support, but you need to limit the payload 
+ * size with max_message_len + remember to leave space for 12 bytes for special 
+ * headers. UDP messages will be chunked automatically.
+ * There is also compression available for UDP protocol, to save network bandwith
+ * for cpu sacrifice. This is not available for TCP due to GELF limitations
```
lminiero: s/for cpu sacrifice/while using a bit more CPU
There should be a full stop at the end of the sentence.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+	if(!event || event == &exit_event)
+		return;
+	json_decref(event);
+}
+
+/* Gelf backend to send the events to */
+static char *backend = NULL;
+static char *port = NULL;
+static int max_gelf_msg_len = 500;
+static int janus_gelfevh_socket_type = 1;
```
lminiero: This should either be an `enum`, to explicitly say "this is UDP, this is TCP" (helpful in case this can be expanded later on with more protocols) or a boolean with the protocol name in it (e.g., `janus_gelfevh_use_tcp` that when `TRUE` means you'll use TCP). A generically named variable with a number value doesn't tell who reads the code what that variable does.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+static int sockfd;
+//static struct sockaddr_in servaddr;
```
lminiero: This has been marked as resolved, but the comment is still in there (and using a syntax that breaks our code style).

events/janus_gelfevh.c
lminiero: This hasn't been resolved yet either.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+	}
+	JANUS_LOG(LOG_INFO, "Connected to GELF backend: [%s:%s]\n", host, port);
+}
+
+static int janus_gelfevh_send(char *message) {
+	if(!message) {
+		JANUS_LOG(LOG_WARN, "Message is NULL, not sending to Gelf!\n");
+		return -1;
+	}
+	/* TCP */
```
lminiero: This should be inside the if, and indented like the related code.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+				return -1;
+			} else if (n == 0) {
+				JANUS_LOG(LOG_WARN, "Connection has been unexpectedly closed by remote side: \n", strerror(errno));
+				close(sockfd);
+				return -2;
+			} else {
+				out_bytes += n;
+			}
+		}
+	/* UDP chunking with headers */
```
lminiero: Same here, it should be inside the else, and indented like the code.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+		char *buf = message;
+		if(compress) {
+			char compressed_text[8192];
+			size_t compressed_len = 0;
+			compressed_len = janus_gzip_compress(compression,
+				message, strlen(message),
+				compressed_text, sizeof(compressed_text));
+			if(compressed_len == 0) {
+				JANUS_LOG(LOG_ERR, "Failed to compress event (%zu bytes)...\n", strlen(message));
+				/* Sending message uncompressed */
```
lminiero: Not sure we should send the message uncompressed in case compression failes: in that case, it's not an error, but a warning, and the log line should say it will fallback to uncompressed data.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+			} else {
+				len = compressed_len;
+				buf = compressed_text;
+			}
+		}
+
+		int total = len / max_gelf_msg_len + 1;
+		if (total > MAX_GELF_CHUNKS) {
+			JANUS_LOG(LOG_WARN, "Gelf allows %d number of chunks, try increasing max_gelf_msg_len\n", MAX_GELF_CHUNKS);
```
lminiero: Since you're returning with an error, here, a `LOG_ERR` is probably better suited. The log line should clarify the message has been dropped. Gelf should be GELF.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+			return -1;
+		}
+		/* do we need to chunk the message */
+		if(total == 1) {
+			int n = write(sockfd, buf, len);
+			if(n < 0) {
+				JANUS_LOG(LOG_WARN, "Sending UDP message failed: %s \n", strerror(errno));
```
lminiero: This should probably be a `LOG_ERR` too. For the output, we usually print `errno` itself too, besides the description, as it makes it easier to debug calls, e.g.:

    JANUS_LOG(LOG_WARN, "Sending UDP message failed: %d (%s)\n", errno, strerror(errno));

As a side note, it looks like you're still using `write` instead of `send`, which is preferred instead, especially when errors happen.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+				return -1;
+			}
+			return 1;
+		} else {
+			int offset = 0;
+			char *rnd = randstring(8);
+			for (int i = 0; i < total; i++) {
+				int bytesToSend = offset + max_gelf_msg_len < len ? max_gelf_msg_len : len - offset;
+				/* prepend the necessary headers (imitate TCP) */
```
lminiero: Just a nit: it should be Prepend with a capital P, since it's starting the sentence

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+				chunk[1] = 0x0f;
+				memcpy(chunk + 2, rnd, 8);
+				chunk[10] = (char)i;
+				chunk[11] = (char)total;
+				char *head = chunk;
+				memcpy(head+12, buf, bytesToSend);
+				buf += bytesToSend;
+				int n = write(sockfd, head, bytesToSend + 12);
+				if(n < 0) {
+					JANUS_LOG(LOG_WARN, "Sending UDP message failed: %s \n", strerror(errno));
```
lminiero: See above on write + errno.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+		if(!item || !item->value) {
+			JANUS_LOG(LOG_WARN, "Missing or invalid port\n");
+			goto done;
+		}
+		port = g_strdup(item->value);
+		item = janus_config_get(config, config_general, janus_config_type_item, "protocol");
+		if (item && item->value) {
+			if (strcasecmp(item->value, "udp") == 0){
+				janus_gelfevh_socket_type = 2;
+			}
```
lminiero: There should be validation here: it should make sure that it's either TCP or UDP, and other values should show a warning saying it's not an accepted value. Whether this should result in failure or fallback to a default value is up to you (I think in most plugins we show that as a warning and fall back).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+		}
+		item = janus_config_get(config, config_general, janus_config_type_item, "max_message_len");
+		if (item && item->value) {
+			int mml = atoi(item->value);
+			max_gelf_msg_len = mml;
```
lminiero: This should make sure the value is valid (e.g., a positive integer).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+			}
+		}
+		/* Done */
+		enabled = TRUE;
+	}
+done:
+	janus_config_destroy(config);
+	config = NULL;
+	if(!enabled) {
+		JANUS_LOG(LOG_FATAL, "Gelf event handler not enabled/needed, giving up...\n");
```
lminiero: s/Gelf/GELF

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+		return -1;	/* No point in keeping the plugin loaded */
+	}
+	JANUS_LOG(LOG_VERB, "Gelf event handler configured: %s:%s\n", backend, port);
```
lminiero: s/Gelf/GELF

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+	/* Launch the thread that will handle incoming events */
+	GError *error = NULL;
+	handler_thread = g_thread_try_new("janus gelfevh handler", janus_gelfevh_handler, NULL, &error);
+	if(error != NULL) {
+		g_atomic_int_set(&initialized, 0);
+		JANUS_LOG(LOG_ERR, "Got error %d (%s) trying to launch the GelfEventHandler handler thread...\n", error->code, error->message ? error->message : "??");
+		return -1;
+	}
+	JANUS_LOG(LOG_INFO, "%s initialized!\n", JANUS_GELFEVH_NAME);
+	janus_gelfevh_connect();
```
lminiero: I think the connect should happen before spawning the thread, and have a return value, so that if connecting fails you can return a failure from the init too. Unless you plan to reconnect on a regular basis when an initial connection fails (e.g., Janus started before Graylog): "regular basis" here may also be "if we're not connected when trying to send a message, try connecting first"

events/janus_gelfevh.c
```diff
@@ -0,0 +1,617 @@
...
+
+	/* Do NOT handle the event here in this callback! Since Janus notifies you right
+	 * away when something happens, these events are triggered from working threads and
+	 * not some sort of message bus. As such, performing I/O or network operations in
+	 * here could dangerously slow Janus down. Let's just reference and enqueue the event,
+	 * and handle it in our own thread: the event contains a monotonic time indicator of
+	 * when the event actually happened on this machine, so that, if relevant, we can compute
+	 * any delay in the actual event processing ourselves. */
+	json_incref(event);
+	//JANUS_LOG(LOG_WARN, "Got event: %s\n", json_dumps(event, json_format));
```
lminiero: See previous notes on comments syntax.

events/janus_gelfevh.c
lminiero: This hasn't been addressed yet: can you please clarify what this check is for?

events/janus_gelfevh.c
lminiero: This is quite important, so please make sure this is addressed in the next commits.

mirkobrankovic:
Hey Lorenzo, thanks for the detailed review (as always :)

I have made a mistake, first one my making a PR from my forked master, which I later on reset to start a new feature audioroom noise canceling ... this reset my PR completely   :/

I saw when I restored it that it was from older branch. I have fixed most of the stuff but I will like to try it out on my acc hosts to make sure all is working fine.

Give me some time to address this, thanks :)

events/janus_gelfevh.c
mirkobrankovic: Done, thanks for a great find

events/janus_gelfevh.c
mirkobrankovic: this saves me a lot of code :D

events/janus_gelfevh.c
mirkobrankovic: I did push this change, but I noticed that it will only work if tcp is chosen, UDP will be considered open :D

mirkobrankovic:
@lminiero can you please review once more.
I tested it and hopefully addressed all comments :)
Thanks

conf/janus.eventhandler.gelfevh.jcfg.sample
lminiero: This was marked as resolved, but it looks it wasn't?

events/janus_gelfevh.c
lminiero: s/recomanded/recommended

events/janus_gelfevh.c
lminiero: This note hasn't been addressed.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,583 @@
...
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include "../ip-utils.h"
+
+/* Plugin information */
+#define JANUS_GELFEVH_VERSION			1
+#define JANUS_GELFEVH_VERSION_STRING 	"0.0.1"
+#define JANUS_GELFEVH_DESCRIPTION 		"This is event handler plugin for Janus, which forwards events via TCP/UDP to GELF server."
+#define JANUS_GELFEVH_NAME 				"JANUS GelfEventHandler plugin"
+#define JANUS_GELFEVH_AUTHOR 			"Meetecho s.r.l."
```
lminiero: This should be you as well.

events/janus_gelfevh.c
lminiero: Please address this too.

events/janus_gelfevh.c
lminiero: Memo :wink:

events/janus_gelfevh.c
lminiero: Memo pt.2 :smile:

events/janus_gelfevh.c
lminiero: This hasn't been addressed, and I think it's an important point.

events/janus_gelfevh.c
lminiero: Please address this too, as it can be helpful when debugging issues.

events/janus_gelfevh.c
lminiero: Memo pt.X :smile:

events/janus_gelfevh.c
lminiero: :arrow_up:

events/janus_gelfevh.c
lminiero: Check still missing here.

events/janus_gelfevh.c
lminiero: Reminder about this.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,583 @@
...
+			json_t *output = json_object();
+
+			int type = json_integer_value(json_object_get(event, "type"));
+			const char *short_message = janus_events_type_to_name(type);
+			json_t *microtimestamp = json_object_get(event, "timestamp");
+			if(microtimestamp && json_is_integer(microtimestamp)) {
+				double created_timestamp = (double)json_integer_value(microtimestamp) / 1000000;
+				json_object_set(output, "timestamp", json_real(created_timestamp));
+			}
+			else {
```
lminiero: Please put the `else` on the same line as the closing bracket (code style).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,583 @@
...
+				struct timeval t;
+				gettimeofday(&t, NULL);
+				double micro_timestamp = (double)(1000000 * t.tv_sec + t.tv_usec) / 1000000;
+				json_object_set(output, "timestamp", json_real(micro_timestamp));
```
lminiero: Why not use `janus_get_real_time()` here?

events/janus_gelfevh.c
mirkobrankovic: I will change it, but it is mentioned that the only difference is in the flags:

> With a zero flags argument, send() is equivalent to write(2)

events/janus_gelfevh.c
mirkobrankovic: ok lets re-build this to enum

mirkobrankovic: I can't seem to set it as enum, since it is already set that way, so I get this error:
```
events/janus_gelfevh.c:111:2: error: redeclaration of enumerator ‘SOCK_STREAM’
  SOCK_STREAM,
  ^
/usr/include/x86_64-linux-gnu/bits/socket_type.h:26:3: note: previous definition of ‘SOCK_STREAM’ was here
   SOCK_STREAM = 1,  /* Sequenced, reliable, connection-based
   ^
events/janus_gelfevh.c:112:2: error: redeclaration of enumerator ‘SOCK_DGRAM’
  SOCK_DGRAM
  ^
/usr/include/x86_64-linux-gnu/bits/socket_type.h:29:3: note: previous definition of ‘SOCK_DGRAM’ was here
   SOCK_DGRAM = 2,  /* Connectionless, unreliable datagrams
   ^
```
I know currently is not so nicely done, but if I put an enum as TCP/UDP then I need to do transformation to SOCK_* or numbers 1/2 to pass to connect, so one more if/else.
Not sure what is smart here.

mirkobrankovic: ok solved

events/janus_gelfevh.c
mirkobrankovic: Is it ok if I check if atoi is > then 0?

mirkobrankovic:
@lminiero I think this should be "almost" ready :dagger: 
Thanks for all the reviews :)

lminiero:
> @lminiero I think this should be "almost" ready 
> Thanks for all the reviews :)

Ack, let me know when it's ready and you want me to do another review :+1:

mirkobrankovic:
Take a look at it now, I think I addressed all the topics mentioned, even the whole section i missed last time :)

lminiero:
Most of my comments have not been addressed, and yet they have been marked as resolved.

conf/janus.eventhandler.gelfevh.jcfg.sample
mirkobrankovic: done, i missed this one completely  and not sure how

events/janus_gelfevh.c
mirkobrankovic: missed most of these, should be ok now

events/janus_gelfevh.c
mirkobrankovic: done :)

events/janus_gelfevh.c
mirkobrankovic: removed

events/janus_gelfevh.c
mirkobrankovic: done

events/janus_gelfevh.c
mirkobrankovic: done

events/janus_gelfevh.c
mirkobrankovic: done :d

mirkobrankovic:
I reviewed all 3 segments of the comments and I think I addressed most of them.

This one if puzzling me, what did you mean by this, to decide if message should be sent uncompressed or not?:

```
				JANUS_LOG(LOG_ERR, "Failed to compress event (%zu bytes)...\n", strlen(message));
				/* Sending message uncompressed */

lminiero on Feb 24 Member
This hasn't been addressed, and I think it's an important point.
```

Thanks a lot for your time to do this, I really appreciate it

events/janus_gelfevh.c
```diff
@@ -0,0 +1,589 @@
...
+static int janus_gelfevh_send(char *message) {
+	if(!message) {
+		JANUS_LOG(LOG_WARN, "Message is NULL, not sending to GELF!\n");
+		return -1;
+	}
+	if(transport == TCP) {
+		/* TCP */
+		unsigned int out_bytes = 0;
+		while (out_bytes < strlen(message)+1) {
+			int n = send(sockfd, message, strlen(message) + 1, 0);
```
lminiero, line 200: I think this is broken. You're looping until a number of bytes equal to the size of the string is sent, but then in case you only sent part of them, you're sending the whole string again. This should try to send the part of the string that wasn't sent (using an offset, and the updated length). As a side note, no need for the +1, you typically don't need to send the terminator.

events/janus_gelfevh.c
lminiero: There should be a validation here, to make sure that only "udp" or "tcp" are provided. If a different value is provided, it should either fail or show a warning and fallback to the default (udp).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,589 @@
...
+
+	while(g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping)) {
+		event = g_async_queue_pop(events);
+		if(event == NULL)
+			continue;
+		if(event == &exit_event)
+			break;
+
+		/* Handle event */
+		while(TRUE) {
```
lminiero, line 559: `while(TRUE)` is a bad idea, there should be checks on whether things are shutting down. Check the other handlers to see how loops are handled there.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,589 @@
...
+				buf = compressed_text;
+			}
+		}
+
+		int total = len / max_gelf_msg_len + 1;
+		if (total > MAX_GELF_CHUNKS) {
+			JANUS_LOG(LOG_ERR, "Event not sent! GELF allows %d number of chunks, try increasing max_gelf_msg_len\n", MAX_GELF_CHUNKS);
+			return -1;
+		}
+		/* do we need to chunk the message */
```
lminiero, line 238: Just a nit: please capitalize the first letter of comments.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,589 @@
...
+}
+
+/* GELF backend to send the events to */
+static char *backend = NULL;
+static char *port = NULL;
+
+typedef enum janus_gelfevh_socket_type {
+	TCP = 1,
+	UDP = 2
+} janus_gelfevh_socket_type;
```
lminiero, line 115: Per code style, the enum values should be prefixed too, e.g., `JANUS_GELFEVH_SOCKET_TYPE_UDP`, just to ensure there are not conflicts with anything else.

events/janus_gelfevh.c
mirkobrankovic: this is the event handling thread that I took from simple:
https://github.com/meetecho/janus-gateway/blob/master/events/janus_sampleevh.c#L461
and i see it also in new ws event handler:
https://github.com/meetecho/janus-gateway/blob/master/events/janus_wsevh.c#L543

mirkobrankovic: there is a check before that, if it is not stopping and initialized

> while(g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping)) {

events/janus_gelfevh.c
mirkobrankovic: I think it should be ok like this

mirkobrankovic:
addressed invalid values:
```
[Fri Apr 24 12:40:31 2020] [janus.eventhandler.gelfevh.jcfg]
[Fri Apr 24 12:40:31 2020]     general: {
[Fri Apr 24 12:40:31 2020]         enabled: true
[Fri Apr 24 12:40:31 2020]         events: all
[Fri Apr 24 12:40:31 2020]         backend: graylog.coligo.com
[Fri Apr 24 12:40:31 2020]         port: 12201
[Fri Apr 24 12:40:31 2020]         protocol: udpp
[Fri Apr 24 12:40:31 2020]         max_message_len: safdsf512
[Fri Apr 24 12:40:31 2020]         compress: false
[Fri Apr 24 12:40:31 2020]         compression: 6
[Fri Apr 24 12:40:31 2020]     }
[Fri Apr 24 12:40:31 2020] [WARN] Missing or invalid transport, using default:UDP
[Fri Apr 24 12:40:31 2020] [WARN] Missing or invalid max_message_len, using default: 500
```
@lminiero when you get time, can you review new changes ? thanks

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+ * \brief  Janus GelfEventHandler plugin
+ * \details  This is a GELF event handler plugin for Janus, which is supposed 
+ * to send json events to GELF 
+ * (Graylog logger https://docs.graylog.org/en/3.2/pages/gelf.html). 
+ * Necessary headers are prepended.
+ * For sending, you can use TCP which is not recommended in case there will be
+ * a lot of messages. There is also UDP support, but you need to limit the payload 
+ * size with max_message_len and remember to leave room for 12 bytes for special 
+ * headers. UDP messages will be chunked automatically.
+ * There is also compression available for UDP protocol, to save network bandwith
```
lminiero, line 13: Small typo: bandwith --> bandwidth

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+#define JANUS_GELFEVH_ERROR_MISSING_ELEMENT		412
+#define JANUS_GELFEVH_ERROR_INVALID_ELEMENT		413
+#define JANUS_GELFEVH_ERROR_UNKNOWN_ERROR		499
+
+/* Plugin implementation */
+static char *randstring(size_t length) {
+	static char charset[] = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+	char *randomString = NULL;
+	if (length) {
+		randomString = malloc(sizeof(char) * (length + 1));
```
lminiero, line 143: This should be `g_malloc`, because you're using `g_free` to free it later on. Using `g_free` on something not allocated by glib will likely cause issues.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+				janus_network_address_to_string_buffer(&addr, &addr_buf) != 0) {
+		if(res)
+			freeaddrinfo(res);
+		JANUS_LOG(LOG_ERR, "Could not resolve address (%s)...\n", backend);
+		return -1;
+	}
+	const char *host = g_strdup(janus_network_address_string_from_buffer(&addr_buf));
+	freeaddrinfo(res);
+
+    if((sockfd = socket(AF_INET, transport, 0)) < 0 ) {
```
lminiero, line 172: Looks like it's indented with spaces and not tabs, here?

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+		/* TCP */
+		int out_bytes = send(sockfd, message, strlen(message), 0);
+		if (out_bytes < 0) {
+			JANUS_LOG(LOG_WARN, "Sending TCP message failed, dropping event: %s \n", strerror(errno));
+			return -1;
+		} else if (out_bytes == 0) {
+			JANUS_LOG(LOG_WARN, "Connection has been unexpectedly closed by remote side: %s\n", strerror(errno));
+			close(sockfd);
+			return -1;
+		} else {
```
lminiero, line 206: Shouldn't there be a check here to see if we sent less than we wanted to send, which can happen with TCP? This will sometimes result in only a portion of the data to be sent, or am I mistaken?

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+		}
+		port = g_strdup(item->value);
+		item = janus_config_get(config, config_general, janus_config_type_item, "protocol");
+		if (item && item->value) {
+			if (strcasecmp(item->value, "udp") == 0){
+				transport = JANUS_GELFEVH_SOCKET_TYPE_UDP;
+			} else if (strcasecmp(item->value, "tcp") == 0){
+				transport = JANUS_GELFEVH_SOCKET_TYPE_TCP;
+			} else {
+				JANUS_LOG(LOG_WARN, "Missing or invalid transport, using default:UDP\n");
```
lminiero, line 323: Missing space in `:UDP`.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+		backend = g_strdup(item->value);
+		item = janus_config_get(config, config_general, janus_config_type_item, "port");
+		if(!item || !item->value) {
+			JANUS_LOG(LOG_WARN, "Missing or invalid port\n");
+			goto done;
+		}
+		port = g_strdup(item->value);
+		item = janus_config_get(config, config_general, janus_config_type_item, "protocol");
+		if (item && item->value) {
+			if (strcasecmp(item->value, "udp") == 0){
```
lminiero, line 318: Missing space in `){` (code style).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+				transport = JANUS_GELFEVH_SOCKET_TYPE_UDP;
+			} else if (strcasecmp(item->value, "tcp") == 0){
```
lminiero, line 320: Missing space in `){` (code style).

events/janus_gelfevh.c
lminiero: Actually, for code style also the `if (` should not have a space instead (happens a few times in the code, just a nit though).

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+			compress = TRUE;
+			item = janus_config_get(config, config_general, janus_config_type_item, "compression");
+			if(item && item->value) {
+				int c = atoi(item->value);
+				if(c < 0 || c > 9) {
+					JANUS_LOG(LOG_WARN, "Invalid compression factor '%d', falling back to '%d'...\n", c, compression);
+				} else {
+					compression = c;
+				}
+			}
```
lminiero, line 353: This should have a check for TCP, and show a warning that, if TCP is used and compression enabled, data will actually not be compressed (as explained in the comment at the beginning of the code). People unaware of the limitation may expect the data to be compressed while it won't be, so a warning will clarify the actual behaviour.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,590 @@
...
+			error_code = JANUS_GELFEVH_ERROR_INVALID_ELEMENT;
+			g_snprintf(error_cause, sizeof(error_cause), "Invalid backend URI '%s:%s'", req_backend, req_port);
+			goto plugin_response;
+		}
+		/* If we got here, we can enforce */
+		janus_mutex_lock(&evh_mutex);
+		if(req_events)
+			janus_events_edit_events_mask(req_events, &janus_gelfevh.events_mask);
+		if (req_compress > -1)
+			compress = req_compress ? TRUE : FALSE;
```
lminiero, line 516: Same considerations about the warning here on enabling compression when TCP is used.

events/janus_gelfevh.c
lminiero: Fair point, I should actually check the code I wrote before complaining :slightly_smiling_face:

events/janus_gelfevh.c
mirkobrankovic: I hate VS code :D, thanks

events/janus_gelfevh.c
mirkobrankovic: good point:

> If no error occurs, send returns the total number of bytes sent, which can be less than the number requested to be sent in the len parameter. Otherwise, a value of SOCKET_ERROR is returned,

mirkobrankovic:
@lminiero I refactored tcp send to make sure all bytes are sent, but while debugging I haven't seen sent less then size (every time is sent in one go, even when messages are 2500 chars long), but better to be safe.
Also I needed to add NULL terminator, otherwise it was skipping every second message.

Can you do one more round?

events/janus_gelfevh.c
```diff
@@ -0,0 +1,606 @@
...
+	janus_network_address addr;
+	janus_network_address_string_buffer addr_buf;
+	struct sockaddr_in servaddr;
+
+	if(getaddrinfo(backend, NULL, NULL, &res) != 0 ||
+				janus_network_address_from_sockaddr(res->ai_addr, &addr) != 0 ||
+				janus_network_address_to_string_buffer(&addr, &addr_buf) != 0) {
+		if(res)
+			freeaddrinfo(res);
+		JANUS_LOG(LOG_ERR, "Could not resolve address (%s)...\n", backend);
```
lminiero, line 166: Can you add errno here too? This is just to align to a commit I made recently, since we were not logging why resolution may fail: https://github.com/meetecho/janus-gateway/commit/d3494b98355710d78073c2d89023e4fc41bd41e8

events/janus_gelfevh.c
```diff
@@ -0,0 +1,606 @@
...
+		return -1;
+	}
+	const char *host = g_strdup(janus_network_address_string_from_buffer(&addr_buf));
```
lminiero, line 169: It looks like you're not doing a `g_free` of `host` when an error occurs (the `return -1` branches in the following lines), or before returning the success. While it's probably a very small leak, it's still a leak, so I'd add that.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,606 @@
...
+	servaddr.sin_family = AF_INET;
+	servaddr.sin_port = htons(atoi(port));
+	servaddr.sin_addr.s_addr = inet_addr(host);
+
+	if(connect(sockfd, (struct sockaddr *)&servaddr, sizeof(servaddr)) < 0) {
+		JANUS_LOG(LOG_ERR, "Connect to GELF host failed\n");
+		return -1;
+	}
+	JANUS_LOG(LOG_INFO, "Connected to GELF backend: [%s:%s]\n", host, port);
+	return 1;
```
lminiero, line 188: Any reason this returns `1` instead of `0`, since you're already returning negative integers in case of errors? Not an issue, just curiosity, since this is different from how we do things in other parts of the code. It looks like the same is done in a few other methods as well.

events/janus_gelfevh.c
```diff
@@ -0,0 +1,606 @@
...
+			if(n < 0) {
+				JANUS_LOG(LOG_WARN, "Sending UDP message failed, dropping event: %s \n", strerror(errno));
+				return -1;
+			}
+			return 1;
+		} else {
+			int offset = 0;
+			char *rnd = randstring(8);
+			for (int i = 0; i < total; i++) {
+				int bytesToSend = offset + max_gelf_msg_len < len ? max_gelf_msg_len : len - offset;
```
lminiero, line 247: I think it's better to wrap the items with a math operation in brackets, here, or there may be ambiguities: if not for the compiler, for who reads the code for sure.

events/janus_gelfevh.c
mirkobrankovic: thanks, will add it

lminiero:
I think this has been around long enough (I see you opened the PR during JanusCon :scream: ), and I think it's fine as it is (I can fix small nits myself, if I find any). Thanks again, merging!

lminiero:
Made some minor changes (mostly code style), and changed the configure to by default compile the module: your patch had `maybe` in there, but that's actually only supposed to be used when there are dependencies to take into account, and a check later might turn the `maybe` into a `yes`. Since your module doesn't depend on any library, it can be a `yes` right away (unless something else sets it to `no`).

mirkobrankovic:
Thanks a lot, I have a lot of problems with this VS Code style and auto-correct on different languages :D

PR #1789, opened 2019-09-22
Add test for -Wunused-but-set-variable

ffontaine:
The -Wunused-but-set-variable option does not exist in some old gcc
versions (gcc 4.5.x), so using it unconditionally breaks the build
with such compilers.

This commit introduces the AX_CHECK_COMPILE_FLAG m4 macro taken from
the autoconf-archive
(http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_check_compile_flag.m4),
and uses it to detect if the -Wunused-but-set-variable option is
supported, and only uses it in this case.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>
(rebased against v0.1.0)
Signed-off-by: Bernd Kuhls <bernd.kuhls@t-online.de>
(rebased against v0.6.3)
Signed-off-by: Adam Duskett <Aduskett@gmail.com>
[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/janus-gateway/0002-Add-test-for-Wunused-but-set-variable.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -65,8 +65,10 @@ clang*)
 	CFLAGS="$CFLAGS \
 		-Wcast-align \
 		-Wno-override-init \
-		-Wunsafe-loop-optimizations \
-		-Wunused-but-set-variable"
+		-Wunsafe-loop-optimizations"
+
+	AX_CHECK_COMPILE_FLAG([-Wunused-but-set-variable],
+		[CFLAGS="$CFLAGS -Wunused-but-set-variable"])
 esac
 
 JANUS_VERSION=75


ffontaine:
I updated the PR as requested

ffontaine:
OK done

atoppi:
Thanks @ffontaine, I have just noticed that `autoconf-archive` is not a dependency in the `README` and we are not relying on it in our environments, so this may break the toolchains we use today.
Adding the macro to the repo is also something we want to avoid.

I'll wait for @lminiero to have the final word on the approval, and also @alexamirante to understand the effects on existing Janus installation routines.

lminiero:
> This definitely adds the new `autoconf-archive` dependency that should be properly advertised as it would break build scripts people have in place.

This seems overkill to me, especially to address something that can be considered legacy (I'm on gcc 9.x right now, not sure how old 4.x and 5.x are). Isn't there a simpler way to detect the gcc version at configure time, to then update the CFLAGS property programmatically? If not, I'd much rather remove the `-Wunused-but-set-variable` flag instead.

atoppi:
Hi @ffontaine,
I'm sorry but after an internal discussion we have decided to reject the PR.
GCC 4.5.x is from 2010-2012 and we can not add a dependency or remove a useful warning just for a legacy compiler.
You'll have to manually remove that flag when compiling with gcc versions older than 4.6.
Thanks the same for contributing!

PR #1777, opened 2019-09-16
Add command line option to janus-pp-rec to specificy the output format

rscreene:
I made a small change to janus-pp-rec to allow the user to specify the destination format as a separate option on the command line (rather than inferring the format from the destination filename).

This enables the user to pipe the output to stdout like:
```
janus-pp-rec -f opus -d 0 plugins/recordings/rec-sample-audio.mjr /dev/stdout > output.opus
```

PR #1791, opened 2019-09-26
Write moov atom at the head of the MP4 file

rscreene:
I modified janus-pp-rec so that it will add the MOOV atom to the header of the output file.

Note that this may be incompatible with my previous PR (https://github.com/meetecho/janus-gateway/pull/1777) if the output file "/dev/stdout" is used.  Let me know if you want this behind a command line flag.  (Or I could fopen the destination file and see if it is seekable?)

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -114,6 +114,7 @@ Usage: janus-pp-rec [OPTIONS] source.mjr [destination.[opus|wav|webm|mp4|srt]]
 int janus_log_level = 4;
 gboolean janus_log_timestamps = FALSE;
 gboolean janus_log_colors = TRUE;
+gboolean janus_faststart = FALSE;
 
 static janus_pp_frame_packet *list = NULL, *last = NULL;
 static char *metadata = NULL;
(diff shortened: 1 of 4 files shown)


lminiero:
What is the new option needed for?

rscreene:
My previous PR was to add a command line option to overwrite the output file format (and not derive it from the output filename).  The thought behind this was to allow us to write to /dev/stdout

The PR tells the avcodec to seek back to the header of the .MP4 file and write the MOOV atom at the start.

Obviously if the output is defined as stdout then the seek will not work.

lminiero:
> The PR tells the avcodec to seek back to the header of the .MP4 file and write the MOOV atom at the start.

Yeah, but I mean, what's the purpose of this change?

lminiero:
The reason I'm asking is that I'm unfamiliar with this MOOV thing. What is it for? Is this something the webm container can also use? At any rate, whatever is it I'd make it an optional setting.

rscreene:
Ah, sorry my mis-understanding.  The MOOV atom contains (amongst other things) the duration of the stream.  janus-pp-rec will write this atom at the end of file, my patch will move it to the front.  If it is not placed at the start of the file then some players (eg. Safari) have to spool through to the end of the file to get this infomation.  I believe that some players will even refuse to play the file.

I'm quite happy to make it optional.  But, I cannot see any existing method that I could use to pass the command line arguments to the protocol file (pp-h264.c).  Can you give me any pointers?

lminiero:
You can just edit the signature of the `janus_pp_h264_create()` method to add the new property, e.g., a boolean flag of some sorts. There's no real convention for the different codecs the tool supports, so you have more freedom there. But again, if this is something that can be used in webm files too, please consider adding it there too.

rscreene:
It doesn't look like it relevant for WebM
Hold off on this PR and I'll update it with a command line flag.

lminiero:

Issue #1685, opened 2019-07-05
Janus 0.7.2 crash when leaving videoroom


pgj007:
```
(gdb) bt
#0  0x00007f3c6f5b9a1f in janus_videoroom_leave_or_unpublish () at /opt/janus/lib/janus/plugins/libjanus_videoroom.so
#1  0x00007f3c6f5dfc54 in janus_videoroom_handler () at /opt/janus/lib/janus/plugins/libjanus_videoroom.so
#2  0x00007f3c9523d4f0 in g_thread_proxy (data=0x96a8a0) at gthread.c:784
#3  0x00007f3c93d50dd5 in start_thread (arg=0x7f3c6b6dd700) at pthread_create.c:307
#4  0x00007f3c93a79ead in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```

```
asm:
cmp    DWORD PTR [rbp-0x12c],0x0                                                                                                                                        
je     0x7f3c6f5b9a5d <janus_videoroom_leave_or_unpublish+1329>                                                                                                         
mov    rax,QWORD PTR [rbp-0x128]                                                                                                                                        
lea    rdx,[rax+0x18]                                                                                                                                                   
mov    rax,QWORD PTR [rbp-0x128]                                                                                                                                        
mov    rax,QWORD PTR [rax+0x8]                                                                                                                                          
mov    rax,QWORD PTR [rax+0xa8] //////////////////////////////////////////rax=0, so crash                                                                                                                                   
mov    rsi,rdx                                                                                                                                                          
mov    rdi,rax                                                                                                                                                          
call   0x7f3c6f5b25a0 <g_hash_table_remove@plt>                                                                                                                        
mov    rax,QWORD PTR [rbp-0x128]                                                                                                                                        
mov    eax,DWORD PTR [rax+0x20]                                                                                                                                         
mov    eax,eax                                                                                                                                                          
mov    rdx,rax                                                                                                                                                          
mov    rax,QWORD PTR [rbp-0x128]                                                                                                                                        
mov    rax,QWORD PTR [rax+0x8]                                                                                                                                          
mov    rax,QWORD PTR [rax+0xb0]                                                                                                                                         
mov    rsi,rdx                                                                                                                                                          
mov    rdi,rax                                                                                                                                                          
call   0x7f3c6f5b25a0 <g_hash_table_remove@plt>                                                                                                                         
[...]

PR #1795, opened 2019-09-30
Better handling of participant's room pointer

atoppi:
This PR tries to address issues like #1685 where concurrent calls to `janus_leave_or_unpublish` and/or room pointer setting to NULL might lead to a crash when destroying and leaving a session at the same time.

I've tested this in common scenarios and it seems to not break anything. Also the room gets correctly freed, so everything should just work.
Nonetheless I'm most interested in feedbacks from people hitting issues similar to #1685.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -2373,14 +2373,21 @@ static void janus_videoroom_leave_or_unpublish(janus_videoroom_publisher *partic
 		return;
 	}
 	janus_mutex_unlock(&rooms_mutex);
-	if(!participant->room || g_atomic_int_get(&participant->room->destroyed))
+	janus_videoroom *room = participant->room;
+	if(!room || g_atomic_int_get(&room->destroyed))
 		return;
+	janus_refcount_increase(&room->ref);
+	janus_mutex_lock(&room->mutex);
+	if (!participant->room) {
+		janus_mutex_unlock(&room->mutex);
+		janus_refcount_decrease(&room->ref);
+		return;
+	}
 	json_t *event = json_object();
 	json_object_set_new(event, "videoroom", json_string("event"));
 	json_object_set_new(event, "room", json_integer(participant->room_id));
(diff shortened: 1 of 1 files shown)


mayankag87:
Hi Alessandro,

I tested the fix but unfortunately it is not good. This is causing a server crash as soon as janus session is destroyed or the leave+detach is done from client side. Could you please look into the issue?

Thanks
Mayank

lminiero:
Please provide a libasan dump. Nothing we can look into, without that.

atoppi:
Hi @mayankag87,
thanks for testing, but as Lorenzo said we need at least a gdb backtrace.

mayankag87:
Sure. Let me collect debugging info and get back.

mayankag87:
Hi Lorenzo/Alessandro,

A strange thing I observe is that I never get a core dumped when my server crashes. When I check the janus process on the server, it is killed proving that the crash occurred but no core dump. I grepped the entire directory for "*core* and found nothing. Can you throw some light on this as to why I don't get a core dumped? 

While I am trying to install libasan on my centOS 7 box, I would like to inform that this issue should be easy to reproduce. I just patched the diff, started the videoroom demo, and then just pressed Stop. Pressing stop leads to server crash. Can you try reproducing the issue at your end if possible while I collect libasan dump?

Thanks

atoppi:
I tried but I am not able to reproduce any crash locally.

lminiero:
@mayankag87 why? This patch has nothing to do with unified-plan! Please test this pull request on master. If it's merged, I'll port it to unified-plan later on.

mayankag87:
I was able to locate the core file on my server. After analyzing the bt, I found out that the crash was caused due to my changes conflicting with the changes done to fix this issue. The crash is resolved now after modifying my changes in accordance with the changes for this fix. I will try to test this on master and share updates.

lminiero:
Merging :+1:

PR #1799, opened 2019-10-02
Added new WebSockets event handler

lminiero:
As the title says. Nothing fancy, you can configure the websockets backend to connect to and, optionally, the subprotocol to use. Reconnection if the connection is lost is done automatically, using an exponential back-off. I only checked plain websockets, so I don't know if secure websockets will work out of the box, or if additional work would be required.

Feedback welcome, I plan to merge soon.


lminiero:
Merging.

PR #1801, opened 2019-10-02
configure.ac: add --disable-stack-protector option

ffontaine:
This option will allow the user to disable -fstack-protector-all as some
toolchains don't support it.
Option is enabled by default to keep current behavior

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,7 +20,6 @@ LT_INIT
 # Common CFLAGS
 CFLAGS="$CFLAGS \
 	-fPIC \
-	-fstack-protector-all \
 	-fstrict-aliasing \
 	-ggdb3 \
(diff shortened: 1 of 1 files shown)


atoppi:
Hi @ffontaine,
sorry but disabling stack protection might lead to serious security risks and we can't permit this by configuring the Janus building script.
If you know what you are doing, changing that flag must be your responsibility.

ffontaine:
For the record, this patch was a tentative to push an upstreamable solution to remove the patch that we have in buildroot since 4 years: https://git.buildroot.net/buildroot/tree/package/janus-gateway/0001-disable-ssp.patch

PR #1804, opened 2019-10-04
--cwd-path (Current Working Directory) CLI option added

l7s:
In some cases (for example when running Janus as non-root user) it might be useful to change daemon process working directory to other than `/`. For example for `core` dumps to be saved to a filesystem, non-root process cannot have `/` as `CDW` but rather a directory writable for that user.

Now it is possible to start daemon as say `janus` process as below (assuming `/opt/janus/` directory is owned by `janus` user):

```
runas="janus"
cwd="/opt/janus/var"
pidfile="/opt/janus/var/run/janus.pid"
logfile="/opt/janus/var/log/janus.log"
daemon="/opt/janus/bin/janus"

su -c "$daemon -b -p $pidfile -N -L $logfile -D -o -d $loglevel -H $cwd" $runas
```

Regards,
Chris

### DIFF
--- a/janus.ggo
+++ b/janus.ggo
@@ -37,3 +37,4 @@ option "token-auth" A "Enable token-based authentication for all requests" flag 
 option "token-auth-secret" - "Secret to verify HMAC-signed tokens with, to be used with -A" string typestr="randomstring" optional
 option "event-handlers" e "Enable event handlers" flag off
 option "no-webrtc-encryption" w "Disable WebRTC encryption, so no DTLS or SRTP (only for debugging!)" flag off
+option "cwd-path" H "Working directory for Janus daemon process (default=/)" string typestr="path" optional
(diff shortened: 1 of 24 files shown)


janus.ggo
```diff
@@ -37,3 +37,4 @@ option "token-auth" A "Enable token-based authentication for all requests" flag
 option "token-auth-secret" - "Secret to verify HMAC-signed tokens with, to be used with -A" string typestr="randomstring" optional
 option "event-handlers" e "Enable event handlers" flag off
 option "no-webrtc-encryption" w "Disable WebRTC encryption, so no DTLS or SRTP (only for debugging!)" flag off
+option "cwd-path" H "Working directory for Janus daemon process (default=/)" string typestr="path" optional
```
lminiero: Please put this where the other daemon-related property is. The file is also missing an endline, apparently, whichshould be fixed. Besides, any new property added here must be added to the README (output of `janus -h`) and the `janus.1` file (man page) as well.

l7s: Missing does added here https://github.com/meetecho/janus-gateway/pull/1804/commits/14d0818ac47bdcf5bca7d866eeb5fd0fd5b361e2

lminiero:
Thanks, looks good! Merging :+1:

Issue #1703, opened 2019-07-18
Modular logging


jalaziz:
The current logging mechanisms in Janus doesn't leave much room for alternative backends.

While stdout and file logging covers most use cases, there are advanced use cases that would be possible with a plugin-based logging mechanism. For example, logging to systems like Kafka or logstash require scraping the logs from the host. Also, while still possible, handling multi-line logs or converting to structured logs is non-trivial.

It would be great if Janus could adopt a modular architecture for logging that allows custom logging backends, much like it currently does for event handlers.

lminiero:
That's an interesting idea, but I'm a bit concerned about performance. Right now, all we have to worry about is some local I/O. The moment plugins appear to do more complex things, we risk dragging the server down as it will have to wait for a print operation to get somewhere before it can move on. I guess this would be up to plugins to implement properly, but it still is a concern.

There's also a problem in configuration and command line settings. Right now we have command line properties to log to stdout, or to file, go in background, etc. The moment this becomes plugin-based, those options will disappear (since they're modules they can't have a static setting, they might not exist in a deployment), which I'll expect will be cumbersome in some deployments. It's even more true for debugging-related settings (e.g., log level, whether timestamps should be logged, enabling lock/refcount debugging, etc.), all of whom have dedicated controls, and even Admin API commands to tweak them; again, with a plugin based approach, this would mean different settings for each of them, and a more cumbersome (and definitely not backwards compatible) way of tweaking them via Admin API.

I'll add an `enhancement` label but to be honest, considering things "work" right now and this change feels pretty complex, it would be low priority for us at the moment, so not sure when we'd be able to tackle this.

jalaziz:
I totally expect it to be low priority for now.

However, I should add that performance is one of the many reasons why modular logging would be better. Logging to local I/O is full of potential performance bottlenecks. Verbose logging to a file (especially on the cloud) can easily kill IOPs capacity. Using a buffer/queue and logging asynchronously can remove the performance burden from Janus. As you said, though, it would be the up to the plugins to do the right thing. Still, there are plenty of great examples out there to model upon.

As for configuration, I understand. For normal deployments, stdout and file logging could remain as CLI options and built-in as default. I'm not suggesting those get removed. For more complex logging needs, I think it's entirely fair to expect users to provide a configuration file as is expected with transports, event handlers, etc.

[...]

PR #1814, opened 2019-10-09
New category of plugins for modular logging

lminiero:
This pull request addresses the enhancement proposed in #1703, namely by adding a new category of plugins that allow you to create custom logger modules. The existing stdout- and file-based logging we had in Janus are still there, and I left them integrated in the core itself. What you can do, now, is add new modules to do more advanced things.

If you have a look at the API for these new plugins, you'll find it's nothing new: it's almost identical to the one for event handlers, so if you worked on such a plugin you should be right at home working on a new logger. The only thing you really need to do is implement the `incoming_logline()` callback, which will provide you with a monotonic timestamp of when the log line was written, and a string containing the log line itself. Then just copy that string, and add it to an internal queue to then do whatever you want with it. Make sure you don't do any I/O or networking inside that callback, or you'll slow the whole Janus logging (the callback is invoked by the log thread in `log.c`); besides, don't mess with the log line string you get in the callback, make a copy and work on that.

As a proof of concept, I added a quite dumb module to save log lines as JSON objects in a local file. The result is something like this:

```
{
   "timestamp": 30006141564,
   "line": "JANUS JSON logger plugin initialized!\n"
}
{
   "timestamp": 30006141622,
   "line": "---------------------------------------------------\n"
}
{
   "timestamp": 30006141625,
   "line": "  Starting Meetecho Janus (WebRTC Server) v0.7.5\n"
}
{
   "timestamp": 30006141628,
   "line": "---------------------------------------------------\n\n"
}
[..]
{
   "timestamp": 30248093214,
   "line": "Bye!\n"
}
```

Nothing fancy, but I just wanted to confirm I could actually handle log lines in a custom way. You'll probably want to make more interesting things with them, e.g., sending them to popular logging frameworks. If you want to give this a try, compile the sample logger by passing the `--enable-json-logger` to the `configure` script, and enable it in its configuration file.

Feedback welcome!


lminiero:
Other important note. You can NOT have different log levels for different loggers: the log level you configure right now impacts all logging, whether it's the integrated one (stdout, file) or external loggers. This is because that filtering and prefixing is done way before getting to the logging functionality, in `debug.h`: when the log line to print gets to the logger, it's already been processed to take the log level into account. Since we just extended the log functionality in `log.c`, external loggers get exactly what integrated loggers get, which seems fair.

If you're interested in figuring out if a log line is, e.g., a warning or an error, you'll have to parse the content yourself: it should be easy enough to have a look at the the string to see how it starts.

lminiero:
Added a new sample logger plugin, one that can send logs to Homer using the HEP protocol. For now it only supports plain UDP and TCP, although I've only tested the UDP one (but not against a real Homer server, so it will likely need fixing). It can be improved along the way with more protocols.

jalaziz:
This looks great! A couple questions/comments though:
1. It _would_ be nice to have access to some core metadata. For example, the janus version and host information. Host information can be derived in the plugin, so I understand if you'd prefer to leave such things out of core.
2. It _would_ also be nice to have the log line deconstructed, but as you mention, it can be parsed out at the plugin level. If leaving it up to the plugin is preferred, would it make sense to ship some utility function to parse the log message?
3. A good example/built-in plugin would be a UDP/TCP syslog plugin.

Other than that, the overall requirements make sense.

lminiero:
> 1. It _would_ be nice to have access to some core metadata. For example, the janus version and host information. Host information can be derived in the plugin, so I understand if you'd prefer to leave such things out of core.

Some info would definitely be useful: e.g., I have an uncommitted change where we pass the server instance name via `init`. That said I'd rather not overdo it. Version can be accessed via `version.h` already (I'll have to check if it's installed in the headers folder, especially for third-party plugins). As to host info, we don't even collect it in the core, so I'd rather leave this up to plugins.

> 2\. It _would_ also be nice to have the log line deconstructed, but as you mention, it can be parsed out at the plugin level. If leaving it up to the plugin is preferred, would it make sense to ship some utility function to parse the log message?

No, I don't think that's needed. I think we can leave this up to plugins that need it.

> 3\. A good example/built-in plugin would be a UDP/TCP syslog plugin.

I'm not really familiar with that, so for now I'll stop at the samples I prepared, as they should be enough as a reference for other loggers. We can add others later on, either ourselves or hoping for contributions.

lminiero:
I chose to only keep the sample JSON logger, for now, and add the Homer logger in a later pull request: in fact, I haven't found the time to test it yet, and the HTTP integration would require some changes in Homer as well before it can be used. As such, it's keep to merge the foundation for modular logging compact, and expand on it later on (which is why everything is modular, after all).

Merging.

PR #1816, opened 2019-10-14
Extension for logging handle_id in janus_sip.c

akarukov:
Logging in janus_sip.c mainly cannot be correlated with specific handle (it easily could be extended to log session_id too).
This PR is to append handle_id (and/or session_id) for each log line of sip plugin where it possible.

•	save handle_id in session struct at session creation time and use this ID for logging.
•	define additional logging macro and replace in most cases
JANUS_LOG(...
for 
JANUS_LOG_SESSION(session,  .....
so logging will look like
2019-10-11 15:34:50 [INFO] [8632242147942951] SIP session created.

For brevity I added only handle_id to logging, since it most descriptive, however it is very easy to modify JANUS_LOG_SESSION macro to include session_id.

2019-10-11 09:32:38 [INFO]  Creating new session: 197745138456214; 0x7f8b80018d90
2019-10-11 09:32:38 [INFO]  Creating new handle in session 197745138456214: 7263169726096724; 0x7f8b80018d90 0x7f8b80006ad0

2019-10-11 09:32:40 [INFO] [7263169726096724] SIP session created.

This way we could correlate and trace logs for specific session/handles.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -491,6 +491,8 @@
 #define JANUS_SIP_AUTHOR			"Meetecho s.r.l."
 #define JANUS_SIP_PACKAGE			"janus.plugin.sip"
 
+#define JANUS_LOG_SESSION(session, level, format, ...) JANUS_LOG(level, "[%"SCNu64"] " format, (session ? session->log_handle_id : 0), ##__VA_ARGS__)
+
 /* Plugin methods */
 janus_plugin *create(void);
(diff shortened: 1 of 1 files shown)


lminiero:
Sorry, but I can't merge this. First of all, the separation of IDs space between core and plugins is intentional: one is not meant to know about the other. Besides, what you're casting to `guint64` is not the handle ID. Closing.

PR #1820, opened 2019-10-15
Logging videoroom session

akarukov:
(Second version. I've addressed your comments/suggestions from #1817 )

Logging in janus_videoroom.c mainly cannot be correlated with specific session.
This PR is to append log_session_id for each log line of videoroom plugin where it is possible.

• save log_session_id in session struct at session creation time and use this ID for logging.
• define additional logging macro and replace in most cases
JANUS_LOG(...
for
JANUS_LOG_SESSION(session, .....
so logging will look like

2019-10-11 09:32:40 [INFO] [7263169726096724] Videoroom session created.

This way we could correlate and trace logs for specific session.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1070,7 +1070,7 @@ room-<unique room ID>: {
 #define JANUS_VIDEOROOM_AUTHOR			"Meetecho s.r.l."
 #define JANUS_VIDEOROOM_PACKAGE			"janus.plugin.videoroom"
 
-#define JANUS_LOG_SESSION(session, level, format, ...) JANUS_LOG(level, "[%"SCNu64"] " format, (session ? session->log_handle_id : 0), ##__VA_ARGS__)
+#define JANUS_LOG_SESSION(session, level, format, ...) JANUS_LOG(level, "[%"SCNu64"] " format, (session ? session->log_session_id : 0), ##__VA_ARGS__)
 
 /* Plugin methods */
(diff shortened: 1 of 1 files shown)


akarukov:
@lminiero , we're open to your suggestions as to what exact ID to show in logs (current time, structure pointer, or whatever you decide more appropriate).

The goal of this commit is to have strict correlation of all log lines from this plugin for specific session (some log lines include various IDs, but they not standardized).
That would help to improve support and troubleshooting.

lminiero:
You should use event handlers for that, not logging.

akarukov:
@lminiero 
Events of course are very good mechanism.

I understand.
I'm not lowering the essence of events framework for troubleshooting solution.
But logging itself is also very critical part of janus.
And there are lot of cases when janus throws an error in log without firing any event.

The ability to group log lines for one user/session/connection is very important.
This PR just adds a session unique identifier for all log lines generated by sip/videoroom plugins.

lminiero:
Again, this can't be merged as it is currently envisaged. It definitely can't be a plugin-specific mechanism as you're doing right now (other plugins except VideoRoom and SIP being second class citizens), and it can't highjack the separation of contexts between core and plugins we're enforcing. If you *really* need this kind of correlation, then this probably needs to happen somewhere in the core, e.g., in `debug.h` with a new macro that involves a plugin session object to add custom prefixes based on its contents (like the handle ID and the plugin name). Closing.

PR #1821, opened 2019-10-15
Logging sip session

akarukov:
(Second version. I've addressed your comments/suggestions from #1816 )

Logging in janus_sip.c mainly cannot be correlated with specific session.
This PR is to append log_session_id for each log line of sip plugin where it is possible.

• save log_session_id in session struct at session creation time and use this ID for logging.
• define additional logging macro and replace in most cases
JANUS_LOG(...
for
JANUS_LOG_SESSION(session, .....
so logging will look like

2019-10-11 09:32:40 [INFO] [7263169726096724] SIP session created.

This way we could correlate and trace logs for specific session.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -491,7 +491,7 @@
 #define JANUS_SIP_AUTHOR			"Meetecho s.r.l."
 #define JANUS_SIP_PACKAGE			"janus.plugin.sip"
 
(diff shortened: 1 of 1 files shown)


akarukov:
@lminiero , we're open to your suggestions as to what exact ID to show in logs (current time, structure pointer, or whatever you decide more appropriate).

The goal of this commit is to have strict correlation of all log lines from this plugin for specific session (some log lines include various IDs, but they not standardized).
That would help to improve support and troubleshooting.

lminiero:
Closing per comment in #1820.

PR #1823, opened 2019-10-16
Update getStats() to use a promise instead of callback

oscarvadillog:
Small update in janus.js for getStats() method. Now, it returns a promise which resolves the stats instead of a callback function.

https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/getStats


lminiero:
Have you verified this works on all browsers? We can't merge something that won't leave some of them behind.

oscarvadillog:
It's the same as here https://github.com/meetecho/janus-gateway/pull/1828#issuecomment-544119476

lminiero:
Merging then!

Issue #1662, opened 2019-06-10
SIP plugin issue - SDP parsing fails with multiple b-lines


zayim:
SDP parsing fails for valid SDP with error:

`[Mon Jun 10 14:43:24 2019] [31m[ERR][0m [sdp-utils.c:janus_sdp_parse:583] Multiple m-line b= lines: b=RS:1000`

`[Mon Jun 10 14:43:24 2019] [31m[ERR][0m [plugins/janus_sip.c:janus_sip_sofia_callback:3813]     Error parsing SDP! Multiple m-line b= lines: b=RS:1000`

SDP content:

```
v=0
o=- 3790417309 42960157160000 IN IP4 X.Y.Z.W
s=-
c=IN IP4 X.Y.Z.W
t=0 0
m=audio 28594 RTP/AVP 8 126
b=RR:3000
b=RS:1000
a=rtpmap:8 PCMA/8000
a=rtpmap:126 telephone-event/8000
a=fmtp:126 0-15
a=maxptime:40
```

This is most probably introduced with this piece of code merged on April 19th, 2019:

```
if(mline->b_name) {
	if(error)
		g_snprintf(error, errlen, "Multiple m-line b= lines: %s", line);
	success = FALSE;
	break;
}
```

which does not allow SDP to have multiple b-lines, which is perfectly valid according to RFC https://tools.ietf.org/html/rfc4566 which says:

`b=* (zero or more bandwidth information lines)`

lminiero:
We can limit that check to AS, but those RS/RR values the SIP client is setting won't get to the browser, as we don't parse them and so don't regenerate them either. Pretty sure they're not supported by browsers anyway.

lminiero:
Fixed in the above commit.

Issue #1277, opened 2018-06-14
Bandwidth limit not working on firefox


groupboard:
The bitrate limit seems to be having no effect on firefox, either in the echotest or videoroom pages, both on the meetecho demo server and our server. It works fine in this page, so I think there is an error in the janus implementation:

https://webrtc.github.io/samples/src/content/peerconnection/bandwidth/

It looks like the REMB bitrate is having no effect for some reason, even though firefox does support REMB.

I managed to get it to limit the bitrate to the value specified in the room config by fixing a few bugs in plugins/janus_videoroom.c. This is good enough for us, although it still doesn't let the user change the bandwidth in the middle of the session:

In janus_videoroom.c, change:

`if(strstr(msg_sdp, "Mozilla")) {
`
to:

`if(strstr(msg_sdp, "mozilla")) {
`
(should be lower case M).

And replace:

```
m->b_name = g_strdup("AS");
m->b_value = (int)(videoroom->bitrate/1000);
```

with:

```
m->b_name = g_strdup("TIAS");
m->b_value = (int)(videoroom->bitrate);
```

(Firefox uses b=TIAS rather than b=AS, and the bitrate for TIAS is in bps rather than kbps).

lminiero:
Just tried the EchoTest with an updated version of both Firefox 60 (stable) and 62 (nightly), and the bandwidth controls seem to be working just fine here.

lminiero:
Notice that adapting to the bitrate might be slow, but it's definitely happening in my case.

groupboard:
I'm using firefox 60 on a macbook pro. Resolution is at 1280x720. If I go to https://janus.conf.meetecho.com/echotest.html and change to 128kbps, even after a minute it's still at about 600kbps.

groupboard:
Just tried Nightly (62) with the same result (capping to 128kbps had no effect...remote stream stayed at 600kbps).

groupboard:
I tested on Windows 10 and it works fine, so it looks like it's an issue specific to either macOS, or the camera on the macbook pro.

lminiero:
Closing as not a Janus issue, then.

groupboard:
The issue is that Firefox doesn't always adjust bandwidth dynamically when receiving REMB packets:

https://bugzilla.mozilla.org/show_bug.cgi?id=1362456

I suspect it is because the Macbook pro camera only has one hardware resolution, and perhaps Firefox doesn't adjust the bandwidth in the case.

[...]

PR #1824, opened 2019-10-17
Fix bandwidth_cap

MvEerd:
I noticed that b=TIAS was no longer in the answer sent to Firefox when bandwidth_cap was true in master
Upon using git bisect I traced this back to the following commit; aa676253bf74f20ee2ab641f49b9b6c2bbf87e4c

It seems that the strcmp() function can return a positive integer when partially matching (and 0 when exactly matching) and thus evaluates to true when TIAS is matched against AS.
The break statement then prevents the line from being set instead of bypassing the error conditional as seemed to be the intention in (#1662), any other b= lines would still trigger the error conditional after the first one is set

Removing this break fixes the bandwidth_cap setting but to further address #1662 I think multiple b-lines should be supported as stated in RFC 4566
I'd imagine a g_list similar to the sdp attributes would be best suited for this but will leave that decision to you as you're much more familiar with all your sdp-utils use cases

### DIFF
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -514,10 +514,6 @@ janus_sdp *janus_sdp_parse(const char *sdp, char *error, size_t errlen) {
 							break;
 						}
 						*semicolon = '\0';
-						if(strcmp(line, "AS")) {
-							/* We only support b=AS, skip */
-							break;
-						}
 						mline->b_name = g_strdup(line);
 						mline->b_value = atol(semicolon+1);
 						*semicolon = ':';


lminiero:
That commit fixed an issue (it mentions #1662), so simply reverting as you did is out of the question. You should make sure the check is still in place but doesn't affect TIAS.

MvEerd:
As mentioned in the PR I am aware of the intention in that commit to fix #1662 but from what I see and noticed it does not actually match as you intended.
It would not return true for ``AS`` since strcmp would return 0 if it exactly matches the given string so it does not currently affect the b=AS line at all

From what I can tell the current break just skips setting the b= line if ``AS`` is included in the line (but does not exactly match)
So it currently prevents ``TIAS`` from being set but does not skip the multiple b= line check for ``AS`` so I think this has been a regression. I'll try to munge some SDPs to further confirm.

Ideally multiple b= lines should be allowed so I think a g_list is the proper way to go but I just wanted to fix/mention the regression since the original issue of #1662 didn't seem addressed and it broke the fix for #1277

lminiero:
Mh, that's weird though, because `strcmp` is not `strstr`, so I don't get why it should "succeed" even if the string doesn't match completely. Maybe it should be a `strcasecmp` instead?

But again, I cannot accept the patch as it currently is. It may not be an issue for you, but it was in some SIP calls (which is where the problem was happening), as apparenty they had more than one b= attribute and it would break things. Limiting the check to b=AS was what fixed it here. I guess we can extend it to b=TIAS, which also means we'd have to extend the SDP structures to store its value (or it will disappear once we regenerate the SDP anyway).

lminiero:
> It seems that the strcmp() function can return a positive integer when partially matching (and 0 when exactly matching) and thus evaluates to true when TIAS is matched against AS.

Re-reading your point, I think the patch is working exactly as intended. `strcmp` is only supposed to return zero when the strings are exactly equal, which is what we want here: we're skipping anything that is not `AS`, so that includes `TIAS` (anything that is not exactly `AS` is skipped). This is because, again, in our SDP structures we only support `AS` anyway: it makes no sense parsing and accepting something if we cannot regenerate it again when we have to convert the SDP structure to a string again.

The fix here, if you need it, is extend the support to `TIAS` as well: not only in parsing the line, but also in the struct fields and converting the SDP to a string.

PR #1828, opened 2019-10-17
Improved attach/reattach MediaStream helpers avoiding browser versions

oscarvadillog:
As you know, the `createObjectURL()` is deprecated in some browsers. 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/srcObject

With this PR we remove all browser and version checks and we try to use `srcObject` whenever possible.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -297,31 +297,29 @@ Janus.init = function(options) {
 		}
 		// Helper methods to attach/reattach a stream to a video element (previously part of adapter.js)
 		Janus.attachMediaStream = function(element, stream) {
-			if(Janus.webRTCAdapter.browserDetails.browser === 'chrome') {
-				var chromever = Janus.webRTCAdapter.browserDetails.version;
-				if(chromever >= 52) {
-					element.srcObject = stream;
-				} else if(typeof element.src !== 'undefined') {
+			try {
+				element.srcObject = stream;
+			}
+			catch {
+				try {
 					element.src = URL.createObjectURL(stream);
-				} else {
+				}
+				catch {
 					Janus.error("Error attaching stream to element");
 				}
-			} else {
-				element.srcObject = stream;
 			}
 		};
 		Janus.reattachMediaStream = function(to, from) {
-			if(Janus.webRTCAdapter.browserDetails.browser === 'chrome') {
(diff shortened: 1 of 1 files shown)


lminiero:
Not sure what harm there is in keeping the check we had in place, if it allows to still support older browsers? Not sure how widespread those still might be, but older distros may be stuck with that. If this doesn't work with all browsers we support, we can't merge this.

oscarvadillog:
I have checked the latest versions of Chrome, Firefox, Opera, Safari and it works well. But I'm not sure about older versions of these browsers. I understand that using try catch will always go the right way, but maybe we should check older versions, just in case.

Do you have a list or table with the browsers and versions Janus supports?

lminiero:
No list, sorry, so that's actually a fair point. Having to chase fixes any time a new browser breaks something, there are good chances that very old browsers wouldn't even work on some other thing. I'll have a look on Monday to see if we still need to worry about those.

html/janus.js
```diff
@@ -297,31 +297,29 @@ Janus.init = function(options) {
...
 		Janus.attachMediaStream = function(element, stream) {
-			if(Janus.webRTCAdapter.browserDetails.browser === 'chrome') {
-				var chromever = Janus.webRTCAdapter.browserDetails.version;
-				if(chromever >= 52) {
-					element.srcObject = stream;
-				} else if(typeof element.src !== 'undefined') {
+			try {
+				element.srcObject = stream;
+			}
+			catch {
```
lminiero: Please put the `catch` on the same line as the closing bracket (code style).

html/janus.js
```diff
@@ -297,31 +297,29 @@ Janus.init = function(options) {
...
-			if(Janus.webRTCAdapter.browserDetails.browser === 'chrome') {
-				var chromever = Janus.webRTCAdapter.browserDetails.version;
-				if(chromever >= 52) {
-					to.srcObject = from.srcObject;
-				} else if(typeof to.src !== 'undefined') {
+			try {
+				to.srcObject = from.srcObject;
+			}
+			catch {
+				try {
```
lminiero: See above.

oscarvadillog:
Done

lminiero:
Sorry for the late response, but I've been away all week due to a conference. Merging, thanks! :+1:

PR #1830, opened 2019-10-18
[Suggestion] Started the refactoring of the janus.js

oscarvadillog:
I admire this project and all the work and effort behind it, in fact it's a key component of my current project.

I would like to participate in the refactoring of the janus.js library. In my opinion, it can be improved a lot in terms of good practices of JavaScript (ES5) without affecting the behavior with the Janus Server.

This is a first commit where I've shown my idea. If you find it interesting let me know and I'll continue refactoring and advancing in this pull request. 😆

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -110,7 +110,7 @@ Janus.useDefaultDependencies = function (deps) {
 			if(options.withCredentials !== undefined) {
 				fetchOptions.credentials = options.withCredentials === true ? 'include' : (options.withCredentials ? options.withCredentials : 'omit');
 			}
-			if(options.body !== undefined) {
+			if(options.body) {
 				fetchOptions.body = JSON.stringify(options.body);
 			}
 
@@ -280,8 +280,7 @@ Janus.init = function(options) {
 							var tracks = stream.getTracks();
 							for(var i in tracks) {
 								var mst = tracks[i];
-								if(mst !== null && mst !== undefined)
-									mst.stop();
+								if(mst) mst.stop();
 							}
 						} catch(e) {}
 					});
@@ -333,8 +332,7 @@ Janus.init = function(options) {
 		window.addEventListener(eventName, function(event) {
 			Janus.log("Closing window");
 			for(var s in Janus.sessions) {
-				if(Janus.sessions[s] !== null && Janus.sessions[s] !== undefined &&
-						Janus.sessions[s].destroyOnUnload) {
+				if(Janus.sessions[s] && Janus.sessions[s].destroyOnUnload) {
 					Janus.log("Destroying session " + s);
 					Janus.sessions[s].destroy({asyncRequest: false, notifyDestroyed: false});
 				}
@@ -410,12 +408,11 @@ Janus.init = function(options) {
 
 // Helper method to check whether WebRTC is supported by this browser
 Janus.isWebrtcSupported = function() {
-	return window.RTCPeerConnection !== undefined && window.RTCPeerConnection !== null;
+	return window.RTCPeerConnection ? true : false;
 };
 // Helper method to check whether devices can be accessed by this browser (e.g., not possible via plain HTTP)
 Janus.isGetUserMediaAvailable = function() {
-	return navigator.mediaDevices !== undefined && navigator.mediaDevices !== null &&
-		navigator.mediaDevices.getUserMedia !== undefined && navigator.mediaDevices.getUserMedia !== null;
+    return (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) ? true : false;
 };
 
 // Helper method to create random identifiers (e.g., transaction)
@@ -605,11 +602,9 @@ function Janus(gatewayCallbacks) {
 			Janus.debug("Got an ack on session " + sessionId);
 			Janus.debug(json);
 			var transaction = json["transaction"];
-			if(transaction !== null && transaction !== undefined) {
+			if(transaction) {
 				var reportSuccess = transactions[transaction];
-				if(reportSuccess !== null && reportSuccess !== undefined) {
-					reportSuccess(json);
-				}
+				if(reportSuccess) reportSuccess(json);
 				delete transactions[transaction];
 			}
 			return;
@@ -618,11 +613,9 @@ function Janus(gatewayCallbacks) {
 			Janus.debug("Got a success on session " + sessionId);
 			Janus.debug(json);
 			var transaction = json["transaction"];
-			if(transaction !== null && transaction !== undefined) {
+			if(transaction) {
 				var reportSuccess = transactions[transaction];
-				if(reportSuccess !== null && reportSuccess !== undefined) {
-					reportSuccess(json);
-				}
+				if(reportSuccess) reportSuccess(json);
 				delete transactions[transaction];
 			}
 			return;
@@ -645,7 +638,7 @@ function Janus(gatewayCallbacks) {
 			if(config.pc && config.remoteSdp) {
 				// Add candidate right now
 				Janus.debug("Adding remote candidate:", candidate);
-				if(!candidate || candidate.completed === true) {
+				if(!candidate || candidate.completed) {
 					// end-of-candidates
 					config.pc.addIceCandidate(Janus.endOfCandidates);
 				} else {
@@ -802,11 +795,11 @@ function Janus(gatewayCallbacks) {
 
 	// Private helper to send keep-alive messages on WebSockets
 	function keepAlive() {
-		if(server === null || !websockets || !connected)
+		if(!server || !websockets || !connected)
 			return;
 		wsKeepaliveTimeoutId = setTimeout(keepAlive, keepAlivePeriod);
 		var request = { "janus": "keepalive", "session_id": sessionId, "transaction": Janus.randomString(12) };
-		if(token !== null && token !== undefined)
+		if(token)
 			request["token"] = token;
 		if(apisecret)
 			request["apisecret"] = apisecret;
@@ -1289,7 +1282,7 @@ function Janus(gatewayCallbacks) {
 			request["token"] = pluginHandle.token;
 		if(apisecret)
 			request["apisecret"] = apisecret;
-		if(jsep !== null && jsep !== undefined)
+		if(jsep)
 			request.jsep = jsep;
 		Janus.debug("Sending message to plugin (handle=" + handleId + "):");
 		Janus.debug(request);
@@ -1314,7 +1307,7 @@ function Janus(gatewayCallbacks) {
 					return;
 				} else if(json["janus"] !== "ack") {
 					// Not a success and not an ack, must be an error
-					if(json["error"] !== undefined && json["error"] !== null) {
+					if(json["error"]) {
 						Janus.error("Ooops: " + json["error"].code + " " + json["error"].reason);	// FIXME
 						callbacks.error(json["error"].code + " " + json["error"].reason);
 					} else {
@@ -1351,7 +1344,7 @@ function Janus(gatewayCallbacks) {
 					return;
 				} else if(json["janus"] !== "ack") {
 					// Not a success and not an ack, must be an error
-					if(json["error"] !== undefined && json["error"] !== null) {
+					if(json["error"]) {
 						Janus.error("Ooops: " + json["error"].code + " " + json["error"].reason);	// FIXME
 						callbacks.error(json["error"].code + " " + json["error"].reason);
 					} else {
@@ -1843,7 +1836,7 @@ function Janus(gatewayCallbacks) {
 						for(var i = 0; i< config.candidates.length; i++) {
 							var candidate = config.candidates[i];
 							Janus.debug("Adding remote candidate:", candidate);
-							if(!candidate || candidate.completed === true) {
+							if(!candidate || candidate.completed) {
 								// end-of-candidates
 								config.pc.addIceCandidate(Janus.endOfCandidates);
 							} else {
@@ -2096,8 +2089,7 @@ function Janus(gatewayCallbacks) {
 						for(var i in tracks) {
 							var mst = tracks[i];
 							Janus.log(mst);
-							if(mst !== null && mst !== undefined)
-								mst.stop();
+							if(mst) mst.stop();
 						}
 					} catch(e) {
 						// Do nothing if this fails
@@ -2418,7 +2410,7 @@ function Janus(gatewayCallbacks) {
 						for(var i = 0; i< config.candidates.length; i++) {
 							var candidate = config.candidates[i];
 							Janus.debug("Adding remote candidate:", candidate);
-							if(!candidate || candidate.completed === true) {
+							if(!candidate || candidate.completed) {
 								// end-of-candidates
 								config.pc.addIceCandidate(Janus.endOfCandidates);
 							} else {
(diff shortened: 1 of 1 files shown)


lminiero:
Before even looking at this, I see a lot of changes to code style that are not needed at all, and make it pretty much impossible to just focus on the changes: e.g., a ton of `if(` turned to `if (`. Our code style is the former, so please don't change that and others.

oscarvadillog:
I only tried to align and use always the same code style. The current code has lines with `if(` and others with `if (`. [Example.](https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L272). You also use double and single quotes at the same time.

On the other hand, I think there are many conditions that could be simplified, for example:
`if(plugin === undefined || plugin === null) {` turn to `if (!plugin)`
or `if(apisecret !== null && apisecret !== undefined)` turn to `if (apisecret)`

Of course I respect your style guide, my intention was only to align the code and always use the same rules. I personally like [Airbnb guide style](https://github.com/airbnb/javascript), but I understand you have yours.

In any case, this PR was just a suggestion, supposed to be unnecessary. We can close it if you want 😄

lminiero:
I'm ok keeping it open, and I appreciate any attempt to improve the code, so thanks for that! But I won't look at it if more than half the changes have nothing to do with the patch, because I honestly don't have the time to discriminate actual improvements from visual things. If you update the patch to just keep the meaningful changes, happy to review.

oscarvadillog:
What about changes like this? https://github.com/meetecho/janus-gateway/pull/1830/commits/4ba444a02d2eda8b9dfe5da18d36606168b83634

lminiero:
Thanks, much more readable and easier to follow this way!

oscarvadillog:
WIP 👍

oscarvadillog:
It may be enough for now. There are many lines already modified.
All demos tested in Chrome and FF and I didn't detect anything broken.

lminiero:
@oscarvadillog thanks! Apologies for the late reply, we've been busy with a few conferences. @atoppi @GaijinKa can you have a look?

html/janus.js
```diff
@@ -108,9 +108,9 @@ Janus.useDefaultDependencies = function (deps) {
 				fetchOptions.headers['Content-Type'] = 'application/json';
 			}
 			if(options.withCredentials !== undefined) {
-				fetchOptions.credentials = options.withCredentials === true ? 'include' : (options.withCredentials ? options.withCredentials : 'omit');
```
lminiero: I'm not sure changes like this are ok. The API requires that value to be `true`, but that check will succeed for anything that in a check resolves to `true`, so basically anything that is not `0`, `null` or `undefined` I guess? I'm ok with simplifying the code, but not at the expense of a weaker API.

html/janus.js
```diff
@@ -469,15 +466,11 @@ function Janus(gatewayCallbacks) {
...
-	if(iceServers === undefined || iceServers === null)
-		iceServers = [{urls: "stun:stun.l.google.com:19302"}];
+	var iceServers = gatewayCallbacks.iceServers || [{urls: "stun:stun.l.google.com:19302"}];
 	var iceTransportPolicy = gatewayCallbacks.iceTransportPolicy;
 	var bundlePolicy = gatewayCallbacks.bundlePolicy;
 	// Whether IPv6 candidates should be gathered
-	var ipv6Support = gatewayCallbacks.ipv6;
-	if(ipv6Support === undefined || ipv6Support === null)
-		ipv6Support = false;
+	var ipv6Support = gatewayCallbacks.ipv6 || false;
```
lminiero: See above: IMO this should be an explicit check on `true`, and a `false` otherwise.

html/janus.js
```diff
@@ -649,7 +638,7 @@ function Janus(gatewayCallbacks) {
 			if(config.pc && config.remoteSdp) {
 				// Add candidate right now
 				Janus.debug("Adding remote candidate:", candidate);
-				if(!candidate || candidate.completed === true) {
+				if(!candidate || candidate.completed) {
```
lminiero: In the Janus API, `completed` in `trickle` is a boolean, so I think the `true` check we had was correct and strict as it was supposed to be.

html/janus.js
```diff
@@ -1539,30 +1523,26 @@ function Janus(gatewayCallbacks) {
...
 			return;
 		}
-		var duration = dtmf.duration;
-		if(duration === null || duration === undefined)
-			duration = 500;	// We choose 500ms as the default duration for a tone
-		var gap = dtmf.gap;
-		if(gap === null || gap === undefined)
-			gap = 50;	// We choose 50ms as the default gap between tones
+		var duration = dtmf.duration || 500; // We choose 500ms as the default duration for a tone
+		var gap = dtmf.gap || 50; // We choose 50ms as the default gap between tones
```
lminiero: `duration` and `gap` are allowed to be `0`: this change breaks that, and incorrectly sets them to the default value instead when `0` is provided.

html/janus.js
```diff
@@ -3081,17 +3038,17 @@ function Janus(gatewayCallbacks) {
 								if(inStats) {
 									config.bitrate.bsnow = res.bytesReceived;
 									config.bitrate.tsnow = res.timestamp;
-									if(config.bitrate.bsbefore === null || config.bitrate.tsbefore === null) {
+									if(!config.bitrate.bsbefore || !config.bitrate.tsbefore) {
```
lminiero: Again, this ignores the case of those being valid integers but `0`.

html/janus.js
```diff
@@ -3164,14 +3121,13 @@ function Janus(gatewayCallbacks) {
...
-				if(!config.streamExternal && config.myStream !== null && config.myStream !== undefined) {
+				if(!config.streamExternal && config.myStream) {
 					Janus.log("Stopping local stream tracks");
 					var tracks = config.myStream.getTracks();
 					for(var i in tracks) {
 						var mst = tracks[i];
 						Janus.log(mst);
-						if(mst !== null && mst !== undefined)
-							mst.stop();
+						if(mst) mst.stop();
```
lminiero: Please don't collapse condition and action. As per our code style, they're always on different lines.

html/janus.js
```diff
@@ -3482,19 +3438,17 @@ function Janus(gatewayCallbacks) {
...
-		if(Janus.webRTCAdapter.browserDetails.browser == "edge") {
+		if(Janus.webRTCAdapter.browserDetails.browser === "edge") {
 			Janus.warn("Edge doesn't support data channels yet");
 			return false;
 		}
-		if(media === undefined || media === null)
+		if(!media)
 			return false;	// Default
-		return (media.data === true);
+		return media.data;
```
lminiero: Users may have provided garbage in the `data` property: this should only return `true` if `data` was really `true`.

atoppi: Agree, and the check is the same for all of the `isEnabled` methods.

lminiero: Actually, not for *all* of them, since `audio` and `video` can be objects (e.g., to specify the deviceId) or strings (e.g., "screen").

lminiero:
I would personally object to anything that sensibly transforms the code, since again, **I'm** the one supposed to maintain it, and I have no intention of doing so with a much more complex JS code I'd likely not understand. Besides, I don't want any dependency in there (if that's what you meant by Babel).

mtltechtemp:
> I would personally object to anything that sensibly transforms the code, since again, **I'm** the one supposed to maintain it, and I have no intention of doing so with a much more complex JS code I'd likely not understand. Besides, I don't want any dependency in there (if that's what you meant by Babel).

I think you're wrong. ES6 looks like C, C#.
I believe it be easier for you since you're a C guy.

lminiero:
Be that as it may, I won't accept any major change to janus.js until the unified-plan branch is merged. It would generate a ton of conflicts with the changes we have there, and I have to fight with those enough as it is.

mtltechtemp:
> Be that as it may, I won't accept any major change to janus.js until the unified-plan branch is merged. It would generate a ton of conflicts with the changes we have there, and I have to fight with those enough as it is.

It's not about to be fussioned since it's been sleeping since December 8, 2018. (Except for bugfixes)
Or am I wrong?

lminiero:
I still want it to be merged, and it will be merged. I wanted to avoid doing that until enough people test it and tell me it's ok, as I wanted to be sure it wouldn't break too many things. Unfortunately that hasn't happened, so one day I'll just do it.

mtltechtemp:
> I still want it to be merged, and it will be merged. I wanted to avoid doing that until enough people test it and tell me it's ok, as I wanted to be sure it wouldn't break too many things. Unfortunately that hasn't happened, so one day I'll just do it.

Well, you did your job, you warned and asked for feedback.
It's been almost a year, I think it's time.

lminiero:
> Well, you did your job, you warned and asked for feedback.
> It's been almost a year, I think it's time.

That's what I wanted to hear! :smile: 
Jokes apart, I'll probably send a last reminder after the IETF (last week of November), to then merge as the first thing on 2020. The plan is to keep the latest 0.x version as a separate branch anyway, rather than a tag as we did so far with versions: while we won't develop new features or accept pull requests on that, it would still allow us to push fixes when needed while people transition to the new version.

oscarvadillog:
Thanks for all your comments @atoppi 

What do we do with the `for ... in` loops? Right now there are 11 loops in the janus.js file

atoppi:
Maybe @lminiero has some comments about this but if you ask me I'd switch the `for ... in` loops on the arrays with the appropriate `for ... of` constructs.

oscarvadillog:
Ok, we wait for @lminiero 
If you agree I can make that change in all the loops.

lminiero:
Sure: as far as I've understood it should be a simple change in the way the for loop is invoked, and so nothing too disruptive. Thanks again for taking care of all this, it's hugely appreciated!

html/janus.js
```diff
@@ -3164,14 +3121,13 @@ function Janus(gatewayCallbacks) {
 			config.bitrate.value = null;
 			try {
 				// Try a MediaStreamTrack.stop() for each track
-				if(!config.streamExternal && config.myStream !== null && config.myStream !== undefined) {
+				if(!config.streamExternal && config.myStream) {
 					Janus.log("Stopping local stream tracks");
 					var tracks = config.myStream.getTracks();
 					for(var i in tracks) {
```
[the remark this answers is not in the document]
oscarvadillog: Already done

oscarvadillog:
Already done

atoppi:
Hi @oscarvadillog,
there two issues:

1) When using the `for ... of` loop you could already get the element of the array.

This means that the old
```javascript
for(var index in config.pc.getSenders()) {
  // get "s"
  var s = config.pc.getSenders()[index];
```
become
```javascript
for(var s of config.pc.getSenders()) {
  // already got "s"
```

But you are still reading the element
```javascript
for(var index of config.pc.getSenders()) {
  var s = config.pc.getSenders()[index];
```
and this is breaking the code.

2) You have to change the construct only for the Arrays, not also for the Object.
For example `Janus.sessions` is an Object, so you need to keep the `for ... in` (unless you are not inspecting an iterable object, but this is not our case).

oscarvadillog:
That's right! I'll check back later.

oscarvadillog:
Already done, your review @atoppi please

lminiero:
He's currently abroad for a conference, so he'll probably only be able to review when he's back, towards the end of the week. From a quick glance, it looks fine to me, but I don't trust myself much when it comes to JavaScript :slightly_smiling_face:

html/janus.js
```diff
@@ -609,11 +600,9 @@ function Janus(gatewayCallbacks) {
 			Janus.debug("Got an ack on session " + sessionId);
 			Janus.debug(json);
 			var transaction = json["transaction"];
-			if(transaction !== null && transaction !== undefined) {
+			if(transaction) {
 				var reportSuccess = transactions[transaction];
-				if(reportSuccess !== null && reportSuccess !== undefined) {
-					reportSuccess(json);
-				}
+				if(reportSuccess) reportSuccess(json);
```
lminiero: Please put the action on a different line and indented (code style). The same seems to be happening in a couple of other instances, so please fix it there too.

html/janus.js
```diff
@@ -622,23 +611,21 @@ function Janus(gatewayCallbacks) {
 			Janus.debug("Got a success on session " + sessionId);
 			Janus.debug(json);
 			var transaction = json["transaction"];
-			if(transaction !== null && transaction !== undefined) {
+			if(transaction) {
 				var reportSuccess = transactions[transaction];
-				if(reportSuccess !== null && reportSuccess !== undefined) {
-					reportSuccess(json);
-				}
+				if(reportSuccess) reportSuccess(json);
```
lminiero: See above.

html/janus.js
```diff
@@ -1752,11 +1726,11 @@ function Janus(gatewayCallbacks) {
 					pluginHandle.iceState(config.pc.iceConnectionState);
 			};
 			config.pc.onicecandidate = function(event) {
-				if (event.candidate == null ||
+				if (!event.candidate ||
 						(Janus.webRTCAdapter.browserDetails.browser === 'edge' && event.candidate.candidate.indexOf('endOfCandidates') > 0)) {
 					Janus.log("End of candidates.");
 					config.iceDone = true;
-					if(config.trickle === true) {
+					if(config.trickle) {
```
lminiero: This should actually remain a `true` check, the `trickle` property must be a boolean.

html/janus.js
```diff
@@ -1771,7 +1745,7 @@ function Janus(gatewayCallbacks) {
 						"sdpMid": event.candidate.sdpMid,
 						"sdpMLineIndex": event.candidate.sdpMLineIndex
 					};
-					if(config.trickle === true) {
+					if(config.trickle) {
```
lminiero: See above.

html/janus.js
```diff
@@ -2134,15 +2101,15 @@ function Janus(gatewayCallbacks) {
 			var constraints = { mandatory: {}, optional: []};
 			pluginHandle.consentDialog(true);
 			var audioSupport = isAudioSendEnabled(media);
-			if(audioSupport === true && media != undefined && media != null) {
+			if(audioSupport && media) {
 				if(typeof media.audio === 'object') {
```
lminiero: Maybe this `if` can be collapsed with the parent one?

html/janus.js
```diff
@@ -2134,15 +2101,15 @@ function Janus(gatewayCallbacks) {
...
 				}
 			}
 			var videoSupport = isVideoSendEnabled(media);
-			if(videoSupport === true && media != undefined && media != null) {
-				var simulcast = callbacks.simulcast === true ? true : false;
-				var simulcast2 = callbacks.simulcast2 === true ? true : false;
+			if(videoSupport && media) {
+				var simulcast = (callbacks.simulcast === true);
+				var simulcast2 = (callbacks.simulcast2 === true);
 				if((simulcast || simulcast2) && !jsep && (media.video === undefined || media.video === false))
```
lminiero: Shouldn't the `(media.video === undefined || media.video === false)` become a `!media.video` too? Or did you leave it out on purpose?

oscarvadillog:
Resolved. Thx @lminiero 😄

oscarvadillog:
Waiting for the green light from @atoppi ❤️

atoppi:
Sorry guys for the late reply, LGTM :+1:

lminiero:
Merging then, thanks @oscarvadillog for the great work!

PR #1832, opened 2019-10-21
Fix bitrate_cap in Firefox

MvEerd:
This PR simply extends the conditional added in #1662 to include TIAS to fix the bandwidth_cap functionality in Firefox

This has been tested and confirmed working on Firefox and Chrome, both the b=TIAS and b=AS lines are set correctly in the answer depending on the browser

### DIFF
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -514,8 +514,8 @@ janus_sdp *janus_sdp_parse(const char *sdp, char *error, size_t errlen) {
 							break;
 						}
 						*semicolon = '\0';
-						if(strcmp(line, "AS")) {
-							/* We only support b=AS, skip */
+						if(strcmp(line, "AS") && strcmp(line, "TIAS")) {
+							/* We only support b=AS and b=TIAS, skip */
 							break;
 						}
 						mline->b_name = g_strdup(line);


lminiero:
As you pointed out in your note on the other PR, since you're reusing the same properties in the struct, this will cause issues when either they're both present, or we're re-generating the SDP. For instance, if the peer sent TIAS, if we parse and re-create the SDP it will be rendered as AS instead. I'd make them separate properties, so that there's no ambiguity.

lminiero:
Edit: regenerating shouldn't be an issue, now that I look at the code again, as we store the AS vs TIAS in the `name` property. There still remains the issue of what to do if they're both there. Not sure if that can happen, though, but the fact Firefox doesn't put them together is unfortunately not conclusive to declare it won't.

lminiero:
The RFC seems to indicate they might very well appear together: https://tools.ietf.org/html/rfc3890#section-6.7

MvEerd:
Indeed, #1662 also highlighted https://tools.ietf.org/html/rfc4566 which states ```b=* (zero or more bandwidth information lines)```
so the lack of multiple b= line support was not yet addressed

This was merely an attempt at making the bitrate cap in Firefox functional again as this is something I need in production. And since Janus sets the TIAS for Firefox, replacing the AS, I don't see this actually being an issue in any of the current use cases?

lminiero:
I'll be abroad for a conference for a few days, but as soon as I'm back (or if I have cycles when I'm there) I'll have a deeper look and maybe craft a fix for both things. Let's keep this open in case it turns out it can be merged safely. Thanks!

lminiero:
> Since Janus only sets b=TIAS for Firefox, and does not set b=AS alongside it, I don't see this actually being an issue in any of the current use cases for the b= lines?

I guess you're talking of the VideoRoom plugin here? Plugins simply relaying the SDP (e.g., EchoTest or VideoCall) would relay the property as it is: only the VideoRoom generates an ad-hoc property differently for Chrome and Firefox. Anyway, after having a look at the code I agree your patch is fine for the moment: we can always make it more advanced later on, if needed.

Thanks, merging!

PR #1835, opened 2019-10-23
Updated libnice recommended version

oscarvadillog:
Minor change to update the libnice recommended version >= 0.1.15 in the README and documentation files.


lminiero:
Good point, and makes sense since we already print a warning in the logs if the version of libnice is not at least that. Merging :+1:

PR #1840, opened 2019-10-27
ice: ignore/enforce only when IP starts with the partial-string from the list

uxmaster:
instead of ignoring/enforcing every IP containing partial-string at whatever position.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -238,7 +238,7 @@ gboolean janus_ice_is_enforced(const char *ip) {
 	GList *temp = janus_ice_enforce_list;
 	while(temp) {
 		const char *enforced = (const char *)temp->data;
-		if(enforced != NULL && strstr(ip, enforced)) {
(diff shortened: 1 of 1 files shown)


lminiero:
While this makes sense, this will fail if IPs/interfaces to split have any spaces in between, e.g.:

    192.168., 10.1.1.8, eth2

This is because we split on the comma, and so some of the strings would actually be something like `[space]10.1.1.8`, that would NOT match with your check. As such, I can't merge until this is taken into account.

uxmaster:
Actually, since neither `strstr()` nor `g_strsplit()` trim any spaces, `[space]10.1.1.8` does NOT match now, as well. So there will be no change in this regard.

lminiero:
Mh, good point... I guess we have been lucky so far then! I'll take note that this should be fixed then, and merge this in the meanwhile. Thanks again for the contribution!

PR #1846, opened 2019-10-30
Gzip compression utility in the core (and sample event handler)

lminiero:
Pretty basic PR that adds a new method to the core `utils.c` to compress a string to `gzip`. The idea is that should be of help to all the plugins that may have a need for that, e.g., event handlers or, in the future, loggers (see #1814).

In this patch, as a proof of concept I only integrated it in the sample event handler: you can configure it to compress the events to send out (and the compression factor to use), and in case, HTTP requests will use `application/gzip` instead of `application/json`. Tested briefly with a node.js backend and it seems to be working as expected. In the future, this may be extended to other plugins as well.

For the moment I only added a way to compress a string to gzip-ed data, and not the other way around, as I didn't see a real need for that. Again, in the future we may want to add that part too if it turns out there's use cases where it may be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ A couple of plugins depend on a few more libraries:
 Additionally, you'll need the following libraries and tools:
 
 * [GLib](http://library.gnome.org/devel/glib/)
+* [zlib](https://zlib.net/)
 * [pkg-config](http://www.freedesktop.org/wiki/Software/pkg-config/)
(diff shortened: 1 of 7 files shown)


mtltechtemp:
What about [Brotli](https://github.com/google/brotli) ?

lminiero:
Not familiar with it, sorry.

lminiero:
Merging.

PR #1849, opened 2019-11-04
VP9 SVC fixes

lminiero:
This PR tries to address some of the shortcomings we had in our existing VP9 SVC support.

Specifically, we only supported SVC when capturing webcams (regular video), but not for screensharing: this is because, in Chrome, different modes are used for the RTP VP9 packetization in the two cases (non-flexible for regular video, flexible for screensharing), and while we were parsing both, we were not using the parsed information properly for the packet pass/drop decisions. Besides, even for regular video the support was sometimes flaky, with video freezing at times when doing transitions from one layer to another.

This patch makes both modes work, now, and should improve the transition support as well. There are some things that can still be improved, like notifying the application when we drop to a lower layer if a higher layer is not available: in fact, unlike simulcast (where the transition must be done explicitly in that case), with SVC video will just work, since you always relay packets coming from lower layers anyway; what's missing is telling the application that you're now *only* sending those lower layers. We do have an "age" kind of property for the incoming layers (that we currently use to notify when we go up a layer after we receive a keyframe), so we may use that for the purpose in the future.

Since Chrome is working on making SVC usable via API as well and not only behind a flag as it is now, this patch should facilitate testing, hopefully at the upcoming IETF hackathon in Singapore as well.

### DIFF
--- a/html/vp9svctest.js
+++ b/html/vp9svctest.js
@@ -646,7 +646,7 @@ function addSvcButtons(feed) {
 				$('#sl' + index + '-1').removeClass('btn-primary btn-info').addClass('btn-primary');
 			if(!$('#sl' + index + '-0').hasClass('btn-success'))
 				$('#sl' + index + '-0').removeClass('btn-primary btn-info').addClass('btn-primary');
-			feeds[index].send({message: { request: "configure", spatial_layer: 1 }});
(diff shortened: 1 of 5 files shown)


lminiero:
Merging.

Issue #1536, opened 2019-03-07
Latching is not working for RTSP stream


ololobster:
See #1218 and 49416c2.

Janus uses wrong IP 0.0.0.0 to send UDP packages. Log:
```
RTSP video latching: 0.0.0.0:20008
  -- RTCP: 0.0.0.0:20009
```
Janus receives `c=IN IP4 0.0.0.0` from the camera, which is the cause of IP 0.0.0.0.
So there are no UDP packages from the camera (I am looking with wireshark). Log:
```
[WARN] [rtsp-test] 5s passed with no media, trying to reconnect the RTSP stream
```
I tried to hardcode real ip into `source->rtsp_vhost` [here](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_streaming.c#L5326) and it worked: UDP packages from the camera started flowing.

I am using branch master.

lminiero:
Looks like a problem in the RTSP server, not Janus. If `0.0.0.0` is what they put in the SDP, then clearly our latching packet will not reach the destination. They should insert a valid IP instead. Closing.

vashi:
Yes its maybe server bug. But its a plain web camera that can't fixed anyway. Moreover VLC plays this stream without any doubt. So, I think is better to create small workaround in Janus.

lminiero:
We can't "invent" the remote address, so not sure which workaround we should create here... if you share (in a gist/pastebin) the SDP coming from the camera we can check if the address is available in some other field.

ololobster:
Check out [RFC 4317](https://tools.ietf.org/html/rfc4317#section-5.3):

> but the answer from Bob contains the connection address 0.0.0.0 and a random port
> number, which means that Alice can not send media to Bob (the media
> stream is "black holed" or "bh")

So camera's behaviour is correct.

> We can't "invent" the remote address

We already have the remote address, we are sending DESCRIBE, SETUP and PLAY somehow.

lminiero:
> So camera's behaviour is correct.

I think you're confusing a bit of concepts and putting them together. The excerpt you mentioned described the hold functionality, which is typically used in SIP when you want to temporarily put a call on hold: it has nothing to do with RTSP. In this case, yes, the camera is saying that it doesn't want any media in, which is exactly why we can't do latching.

> We already have the remote address, we are sending DESCRIBE, SETUP and PLAY somehow.

[...]

PR #1851, opened 2019-11-04
Address issue #1536.

gcbartlett:
Fall back to using the RTSP SETUP response's 'source' field, if present, for the video latching host address, if the SDP field in invalid (0.0.0.0).

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -5590,6 +5590,20 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 			JANUS_LOG(LOG_VERB, "  -- SSRC (video): %"SCNu32"\n", ssrc);
 			source->video_ssrc = ssrc;
 		}
+		if(strcmp(vhost, "0.0.0.0") == 0) {
+			/* SDP does not contain a valid host address, so check for an RTSP 'source' address instead */
+			const char *source = strstr(curldata->buffer, ";source=");
+			if(source != NULL) {
+				const char *start = source+strlen(";source=");
+				const char *semicolon = strchr(start, ';');
+				const size_t source_len = semicolon ? (size_t)(semicolon-start) : strlen(start);
+				if (source_len < sizeof(vhost)) {
+					strncpy(vhost, start, source_len);
+					vhost[source_len] = '\0';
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! While I find the time to review, can you sign our [CLA](https://cla.conf.meetecho.com/janus/) to get this merged?

plugins/janus_streaming.c
```diff
@@ -5590,6 +5590,20 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 			JANUS_LOG(LOG_VERB, "  -- SSRC (video): %"SCNu32"\n", ssrc);
 			source->video_ssrc = ssrc;
 		}
+		if(strcmp(vhost, "0.0.0.0") == 0) {
```
lminiero: I see a couple of problems here:

1. You're only handling video for some reason: the same should be done for audio too.
2. Your trigger for checking `source` is the c-line being `0.0.0.0`, but it should actually also be when no c-line is there at all (maybe unlikely, but better handle that). To do that, both `ahost` and `vhost` should be initialized after they're declared, e.g.., `ahost[0] = '\0';`, as this way, if parsing the SDP we can't find any c-line, then we can check if the host string is empty as another trigger.

plugins/janus_streaming.c
```diff
@@ -5590,6 +5590,20 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
...
+			/* SDP does not contain a valid host address, so check for an RTSP 'source' address instead */
+			const char *source = strstr(curldata->buffer, ";source=");
+			if(source != NULL) {
+				const char *start = source+strlen(";source=");
+				const char *semicolon = strchr(start, ';');
+				const size_t source_len = semicolon ? (size_t)(semicolon-start) : strlen(start);
+				if (source_len < sizeof(vhost)) {
```
lminiero: If there's a check for that, there should be a warning if the source exceeds that value: right now it's just ignored.

lminiero:
I prepared a new patch that addresses the `source` parameter (for both audio and video), and extends the functionality to resolving the signalling server address if that's missing too. Closing this PR as it's obsoleted by that one.

gcbartlett:
Thanks Lorenzo.  Sorry I couldn't finish up my patch (am waiting for approval from our legal department on CLA).

PR #1852, opened 2019-11-04
Use strtol more, and add checks when atoi is used

lminiero:
As the title says, this is an attempt to rely less on `atoi`, and more on `strtol` instead: we didn't just replace it everywhere, but only where it made most sense, e.g., when parsing ports or when a type range is important. The patch adds a few helpers (e.g., `janus_string_to_uint16`) to make it all simpler to use in core and plugins.

Please test to make sure it doesn't break your applications (it shouldn't).


lminiero:
Merging then :+1:

Issue #987, opened 2017-09-01
Janus exits if STUN server can't be reached


mzanetti:
If a STUN server is configured, Janus tries to connect to it on startup. If that fails, for instance because of the device isn't online just yet, Janus exits with a FATAL error.

This is problematic because the service manager (in our case systemd) keeps on respawning Janus while Janus keeps on exiting. This causes a start/stop loop causing high CPU/Resource demand and slowing down other services.

If the network connection can't be established, Janus should instead wait for a bit and just try to reconnect instead of exiting.

Here's an excerpt of the log when this happens:

Aug 29 10:46:42 loop-box snap[3904]: STUN server to use: numb.viagenie.ca:3478
Aug 29 10:46:42 loop-box snap[3904]: [ERR] [ice.c:janus_ice_set_stun_server:759] Could not resolve numb.viagenie.ca...
Aug 29 10:46:42 loop-box snap[3904]: [FATAL] [janus.c:main:3725] Invalid STUN address numb.viagenie.ca:0

lminiero:
I'm personally not a big fan of waiting until things exist (which might never happen), this should be up to the service setup and can be done externally. It's a fatal error and not a warning because warnings may not be seen, and then things don't work and you don't know why: a fatal error allows you to check the configuration for mistakes and fix that.

mzanetti:
I see your point. But this isn't a configuration error. The configuration is valid in this case but for some reason internet connectivity is down atm. It doesn't seem sensible to me to just crash whenever the internet connection goes down for a moment. Obviously printing a warning to the log that the STUN connection failed makes sense so in case of a configuration error one could still find that in the logs which one has to read anyways, even when it exits with a fatal error.

uasan:
You just need correctly to set directive `After` in Systemd unit file. 
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Before=

mzanetti:
yeah, I had that thought too...  "After=network-online.target" would probably work around this for this particular case. Need to figure out how to do that as we're shipping janus in an Ubuntu Snap package which autogenerates the systemd service, but I'm sure there is a way somehow.

Still just exiting on temporary failures seems wrong to me...

lminiero:
Closing as there's a way to get around that via Systemd, and we have no immediate plan to change the current behaviour.

jeremyarr:
[...]

PR #1854, opened 2019-11-05
Feature/ignore unreachable ice server

sjkummer:
Adding an additional nat config item 'ignore_unreachable_ice_server' which allows to start Janus on edge devices even if WAN connectivity is temporarily unavailable. The default behaviour (without the new config item) remains the same as before.
https://github.com/meetecho/janus-gateway/issues/987

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1011,17 +1011,13 @@ int janus_ice_set_stun_server(gchar *stun_server, uint16_t stun_port) {
 	/* Test the STUN server */
 	janus_network_address public_addr = { 0 };
 	if(janus_ice_test_stun_server(&addr, janus_stun_port, 0, &public_addr, NULL) < 0) {
-		if(janus_stun_server) {
-        	g_free(janus_stun_server);
-        }
+		g_free(janus_stun_server);
 		janus_stun_server = NULL;
 		return -1;
 	}
 	if(janus_network_address_to_string_buffer(&public_addr, &addr_buf) != 0) {
 		JANUS_LOG(LOG_ERR, "Could not resolve public address...\n");
-		if(janus_stun_server) {
-			g_free(janus_stun_server);
-		}
+		g_free(janus_stun_server);
 		janus_stun_server = NULL;
 		return -1;
 	}
--- a/ice.c
+++ b/ice.c
@@ -991,10 +991,13 @@ int janus_ice_set_stun_server(gchar *stun_server, uint16_t stun_port) {
 
 	/* Test the STUN server */
 	janus_network_address public_addr;
-	if(janus_ice_test_stun_server(&addr, janus_stun_port, &public_addr) < 0)
+	if(janus_ice_test_stun_server(&addr, janus_stun_port, &public_addr) < 0) {
+		janus_stun_server = NULL;
 		return -1;
+	}
 	if(janus_network_address_to_string_buffer(&public_addr, &addr_buf) != 0) {
 		JANUS_LOG(LOG_ERR, "Could not resolve public address...\n");
+		janus_stun_server = NULL;
 		return -1;
 	}
 	const char *public_ip = janus_network_address_string_from_buffer(&addr_buf);
--- a/ice.c
+++ b/ice.c
@@ -1011,11 +1011,17 @@ int janus_ice_set_stun_server(gchar *stun_server, uint16_t stun_port) {
 	/* Test the STUN server */
 	janus_network_address public_addr = { 0 };
 	if(janus_ice_test_stun_server(&addr, janus_stun_port, 0, &public_addr, NULL) < 0) {
+		if(janus_stun_server) {
+        	g_free(janus_stun_server);
+        }
 		janus_stun_server = NULL;
 		return -1;
 	}
 	if(janus_network_address_to_string_buffer(&public_addr, &addr_buf) != 0) {
 		JANUS_LOG(LOG_ERR, "Could not resolve public address...\n");
+		if(janus_stun_server) {
+			g_free(janus_stun_server);
+		}
 		janus_stun_server = NULL;
 		return -1;
 	}
--- a/janus.c
+++ b/janus.c
@@ -4106,7 +4106,7 @@ gint main(int argc, char *argv[])
 #endif
 	const char *nat_1_1_mapping = NULL;
 	uint16_t rtp_min_port = 0, rtp_max_port = 0;
-	gboolean ice_lite = FALSE, ice_tcp = FALSE, full_trickle = FALSE, ipv6 = FALSE;
+	gboolean ice_lite = FALSE, ice_tcp = FALSE, full_trickle = FALSE, ipv6 = FALSE, ignore_unreachable_ice_server = FALSE;
 	item = janus_config_get(config, config_media, janus_config_type_item, "ipv6");
 	ipv6 = (item && item->value) ? janus_is_true(item->value) : FALSE;
 	item = janus_config_get(config, config_media, janus_config_type_item, "rtp_port_range");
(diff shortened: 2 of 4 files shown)


lminiero:
This has been discussed in the past, and I'm not sure it's wise. That startup test can be very helpful to figure out if the connection details will be usable, and allowing people to skip it will cause issues in some cases: what if it's not just "temporarily unreachable", but unreachable-period, and you're starting the server anyway? Besides, in case of STUN it means Janus won't have access to the public IP, which shouldn't be an issue in most cases (it's only used for nat-1-1, which should not be used in conjunction with STUN anyway), but still implies incomplete information when you ask Janus for info.

Anyway, if you guys really want to shoot yourself in the foot, I'll think about it. That said, please sign the [CLA](https://cla.conf.meetecho.com/janus/), as it's required for merging.

sjkummer:
Hello Lorenzo, thanks for your fast response. 

I fully agree that reachability problems should not be ignored if Janus runs as cloud server. I have added a (hopefully) very clear warning in the example-config and added a note about using the systemd 'After' directive as best-practice. Further, I have signed the CLA.

I really think this change should be merged to make Janus ready for Edge-Computing and IoT use cases. As an example: Janus can be used to add WebRTC-capabilities to legacy audio/video streams in order to make them interoperable with modern browsers. This streams can be consumed by on-site applications (which do not require TURN) and optionally remote clients (which require TURN). Why should we disallow the on-site usage of an application just because the internet connectivity (of the company / factory / organisation / public office / ... ) is unreliable?

This change should improve the reliability and the range of application for Janus and its community.

Best regards,
Sebastian

lminiero:
> Why should we disallow the on-site usage of an application just because the internet connectivity (of the company / factory / organisation / public office / ... ) is unreliable?

We wouldn't be preventing anything: if the application doesn't require TURN, they wouldn't put any TURN server in the config, and there would be no attempt to connect to anything, and so no risk of getting a failure doing that, whether they're connected to the internet or not. This can only happen if one actively sets a STUN or TURN server in the settings.

Notice that skipping the STUN/TURN check because it fails does NOT mean they won't be used: they'll still be involved any time a PeerConnection needs to be established, because they were in the configuration. So, just for the sake of clarity, if the purpose of this PR is disabling STUN if that check fails and just go on, that's not what would happen: the STUN/TURN server would still be in the configuration properties, and would be used accordingly (and possibly fail again). If that's not the purpose, then that's ok, and this can be merged.

lminiero:
PS: to clarify, the STUN server would still be attempted for a new PeerConnection if what failed was the STUN connectivity check; if the failure was in resolving the address, then Janus would not have reached the part of the code where it takes note of the STUN server address, and so it would NOT be used further. For TURN, we never to any connectivity check, and so the only failure can come from an error trying to resolve the address (which means that TURN would indeed be always be disabled, in case of a failure at startup).

lminiero:
To solve this ambiguity, I think the stun/turn init methods, before returning the error, should free the strings containing the resolved address. This way, no matter the cause, if the server is configured to go on the user knows STUN/TURN support are disabled because of that.

sjkummer:
> PS: to clarify, the STUN server would still be attempted for a new PeerConnection if what failed was the STUN connectivity check; if the failure was in resolving the address, then Janus would not have reached the part of the code where it takes note of the STUN server address, and so it would NOT be used further. For TURN, we never to any connectivity check, and so the only failure can come from an error trying to resolve the address (which means that TURN would indeed be always be disabled, in case of a failure at startup).

Thanks for your further clarifications! My original intention was to use the configured STUN and TURN servers for all new PeerConnections after the ice servers have become reachable. But I realise that this actually would mean to resolve the ice server addresses at a later point (again) which would increase the complexity of this change. 
I agree to free up the resolved STUN address after the connectivity check failed to achieve a consistent behaviour. I'll push a change.

lminiero:
Thanks! It shouldn't just be set to NULL, though: the string is strdup-ed and do should be freed. The same should be done in the TURN setup too, since the new ignore property is involved for that method was well.

As a side note, it looks like there are some conflicts now?

sjkummer:
Okay, I have resolved the conflicts and freed the stun server string.
For TURN, no string needs to be freed, because no further test is done after the TURN server address was resolved, right?

ice.c
```diff
@@ -1010,10 +1010,19 @@ int janus_ice_set_stun_server(gchar *stun_server, uint16_t stun_port) {
 
 	/* Test the STUN server */
 	janus_network_address public_addr = { 0 };
-	if(janus_ice_test_stun_server(&addr, janus_stun_port, 0, &public_addr, NULL) < 0)
+	if(janus_ice_test_stun_server(&addr, janus_stun_port, 0, &public_addr, NULL) < 0) {
+		if(janus_stun_server) {
```
lminiero: No need for the `if`, `g_free` does nothing if it's `NULL`.

ice.c
```diff
@@ -1010,10 +1010,19 @@ int janus_ice_set_stun_server(gchar *stun_server, uint16_t stun_port) {
...
+        	g_free(janus_stun_server);
+        }
+		janus_stun_server = NULL;
 		return -1;
+	}
 	if(janus_network_address_to_string_buffer(&public_addr, &addr_buf) != 0) {
 		JANUS_LOG(LOG_ERR, "Could not resolve public address...\n");
+		if(janus_stun_server) {
```
lminiero: Same as above.

janus.c
```diff
@@ -4235,12 +4238,20 @@ gint main(int argc, char *argv[])
 	/* Initialize the ICE stack now */
 	janus_ice_init(ice_lite, ice_tcp, full_trickle, ipv6, rtp_min_port, rtp_max_port);
 	if(janus_ice_set_stun_server(stun_server, stun_port) < 0) {
-		JANUS_LOG(LOG_FATAL, "Invalid STUN address %s:%u\n", stun_server, stun_port);
-		exit(1);
+		if(!ignore_unreachable_ice_server) {
+			JANUS_LOG(LOG_FATAL, "Invalid STUN address %s:%u\n", stun_server, stun_port);
+			exit(1);
+		} else {
+			JANUS_LOG(LOG_ERR, "Invalid STUN address %s:%u\n", stun_server, stun_port);
```
lminiero: I'd add a "STUN will be disabled" text to the error, so that it's more explicit.

janus.c
```diff
@@ -4235,12 +4238,20 @@ gint main(int argc, char *argv[])
...
+		}
 	}
 	if(janus_ice_set_turn_server(turn_server, turn_port, turn_type, turn_user, turn_pwd) < 0) {
-		JANUS_LOG(LOG_FATAL, "Invalid TURN address %s:%u\n", turn_server, turn_port);
-		exit(1);
+		if(!ignore_unreachable_ice_server) {
+			JANUS_LOG(LOG_FATAL, "Invalid TURN address %s:%u\n", turn_server, turn_port);
+			exit(1);
+		} else {
+			JANUS_LOG(LOG_ERR, "Invalid TURN address %s:%u\n", turn_server, turn_port);
```
lminiero: Same here.

lminiero:
> For TURN, no string needs to be freed, because no further test is done after the TURN server address was resolved, right?

You're right! I should probably add a STUN check for TURN servers too, to make sure they're reachable (as a TURN server is a STUN server too), but I can do that after this is merged.

ice.c
sjkummer: good to know :-)

sjkummer:
Should all be cleaned up now.

lminiero:
Thanks for your patience and quick fixes, looks good now! Merging then :+1:

sjkummer:
Great thanks 🙂

Issue #1853, opened 2019-11-05
Segmentation fault in Janus Streaming plugin 


digitalsamba:
We have been able to regularly reproduce (largely on demand) a segmentation fault inside the Janus Streaming plugin.

It happens most often on a rapid start/stop of a stream in a rtp forwarding scenario. Basically we are rapidly destroying a streaming mount point and creating a new one, meaning users in a room are trying to subscribe immediately to every new streaming mountpoint, which is destroyed soon after.

The culprit line is "if(mountpoint->streaming_source == janus_streaming_source_rtp) {" inside "janus_streaming_setup_media" method in janus_streaming.c.

It seems to us that the mountpoint variable is NULL and is being referenced causing a segmentation fault.
i.e. It sounds like Janus is trying to finish the setup of a streaming mountpoint, which is already destroyed and mountpoint variable was set to NULL by another concurrent thread.

gdb backtraces are attached for 0.7.4 and 0.7.5. 
This segfault is occurring in both versions. 

0.7.4 

gdb threads -> https://pastebin.com/iGzYCir6
gdb bt -> https://pastebin.com/hTEXE8bt

0.7.5 

gdb threads -> https://pastebin.com/BdDFUfjN
gdb bt -> https://pastebin.com/MfQd0fmt

lminiero:
Can you check if the PR we merged above fixes the issue for you?

digitalsamba:
@lminiero Thank you very much! This has worked well for us. 
We tested this heavily over the last several hours and have not seen a segfault since. 

Would it be possible to get this merged into 0.7.4 & 0.7.5 as using tagged stable versions is a better fit for our deployment model right now?

lminiero:
You have to thank @garry81, not me :wink: 

Tagged versions are tags, not branches, so they can't get patches. As a side note, tagged versions are **NOT** stable: they're just tags of a specific commit before merging something substantial. There is no such thing as a stable version, since with WebRTC tomorrow a browser update could break anything. As far as I'm concerned, the only stable version is master, whatever it is, as it's the only one receiving bug fixes and updates.

Closing.

PR #1855, opened 2019-11-06
fix nullptr dereference in streaming plugin

garry81:


lminiero:
I guess this is related to #1853? Not sure if that will be enough, as if the cause of the problem is a mountpoint being destroyed while people are connecting to it, a crash might still occur after the check. Anyway, it's definitely a good starting point, and the changes make sense, so I'll merge. Thanks! :+1:

PR #1865, opened 2019-11-13
libnice-devel required

wasphin:
Test on RHEL distributions.


lminiero:
Sorry, but can't accept this. In the README we clarify that you need a recent version of libnice, and most of the ones available in RH repos (CentOS in particular) are ancient, which means that it should be compiled manually. Closing.

PR #1866, opened 2019-11-13
Fixes to RTSP latching procedure (fixes #1536, replaces #1851)

### LINKED ISSUE #1536: Latching is not working for RTSP stream

lminiero:
As the title says, this PR improves the way we support RTSP latching in the Streaming plugin.

At the moment, to figure out the address the RTSP server will send media from, we only look at the c-line in the SDP returned in response to a DESCRIBE request: some servers, though (e.g., some cameras), don't put a valid value in there. Per the standard, the `source` param in the `Transport` header in the response to a SETUP request should be used instead, when available. That said, there are servers that don't advertise the address neither in a c-line, nor in a `source` parameter: the only way to figure out the address, in that case, is simply resolve the address of the server from the RTSP url itself.

Starting from these considerations, the patch changes the behaviour like this:

1. we look at the SDP, and if a valid address is in the c-line, we take note of that;
2. if we detect a `source` attribute, we use the address reported there, whether we found an address in the c-line or not;
3. in case we didn't get any address from the above, we resolve the host part of the RTSP url we used to contact the server (in case both audio and video are there, we only do that once).

As a side note, the patch also fixes another incorrect behaviour in the plugin. Namely, when configuring an RTSP mountpoint, you can specify whether audio or video should be ignored: the code in master doesn't do that, meaning that if the RTSP stream is audio/video, but you want video only (`audio=false`), audio would be negotiated with the RTSP server anyway. The patch fixes that.

Feedback welcome! I tested this with a few RTSP servers (a local one and a couple of public streams) and it seems to be working as expected, but of course I'm interested to figure out if this still results in latching not working as expected with some camera models. Please notice that I'm only interested to figure out if Janus is performing latching to the right address: if the camera/RTSP server ignores it and streaming doesn't work, nothing we can do there.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -5813,11 +5813,7 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 						pi++;
 					char name[50], value[50];
 					if(sscanf(pi, "%49[a-zA-Z_0-9]=%49s", name, value) == 2) {
-						if(!strcasecmp(name, "timeout")) {
-							/* Take note of the timeout, for keep-alives */
-							ka_timeout = atoi(value);
-							JANUS_LOG(LOG_VERB, "  -- RTSP session timeout (audio): %d\n", ka_timeout);
-						} else if(!strcasecmp(name, "ssrc")) {
+						if(!strcasecmp(name, "ssrc")) {
 							/* Take note of the audio SSRC */
 							uint32_t ssrc = strtol(value, NULL, 16);
 							JANUS_LOG(LOG_VERB, "  -- SSRC (audio): %"SCNu32"\n", ssrc);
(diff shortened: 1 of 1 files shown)


gcbartlett:
Looks like this now requires libcurl >= 7.62.0.  Before it would compile with earlier versions (and would point out when it was less than 7.45.0 - for RTSP authentication support).  Are you OK with the new build-time requirements this introduces, or is there another way to get the server address?

lminiero:
There are no new requirements, I didn't touch the libcurl stuff.

gcbartlett:
Your fix adds use of `curl_url`:
```CURLU *url = curl_url();```
which according to https://curl.haxx.se/libcurl/c/curl_url.html was added in curl 7.62.0.

Now, I'm only currently using the 0.7.1 release of janus, so maybe you upped the libcurl requirement since then?

lminiero:
You're right, I forgot about that part. I honestly have no will or intention to write a manual parser for RTSP urls: since I'm using the same libcurl parser for a modular logger in #1814 as well (which needs to support generic urls, making manual parsing much more cumbersome, I think it's easier to just raise the libcurl version requirement as you pointed out.

If someone can provide a zero effort and reliable mechanism to do the same thing, I'll gladly accept a pull request, otherwise I'll just modify the configure script to require 7.62.0.

Thanks for pointing this out!

lminiero:
Updated the patch so that we only go through the `CURLU` code if libcurl is recent enough. This means that, if libcurl is too old, we only try 1. and 2., and not 3.

lminiero:
These header parsing methods are getting awfully long, and repetitive... we should probably put those in helper functions to keep the code clean and compact. Will do that when I'm back.

lminiero:
Merging.

PR #1867, opened 2019-11-17
Dynamically update NACK queue size depending on RTT

lminiero:
You may be familiar with a property in Janus called `max_nack_queue`: that's a property you can configure at startup (and dynamically, via Admin API) to tell Janus how long sent RTP packets should be kept in memory, in case NACKs arrive. As such, it's not a count on the number of packets, but a time value instead, in milliseconds to be precise. This is fine, but the problem is that this value is global, and not per PeerConnection: in fact, the ability to react to NACKs is tied to the RTT (the browser needs time to detect the missing packet and send a NACK for it), but different users/PeerConnections may have wildly different RTT values. As such, it's hard to find a value that fits all: make it too low, and NACKs may be unusable in some cases; make it too large, and you'll be wasting memory, besides reacting to NACKs for packets that are way too old and so potentially cause delays.

This patch tries to address that, by introducing a dynamic aspect to it. Basically, `max_nack_queue` is now gone, and has been replaced by a new property called `min_nack_queue` instead, which defaults to 200ms. Then, for each PeerConnection we regularly check the RTT we get from RTCP: we use the RTT info to compute an updated NACK queue size value for that specific PeerConnection, and use that instead. If the computed value is lower than `min_nack_queue`, then we stick to `min_nack_queue`: this allows you to still be more conservative if you want, that is, even if we detect sun and rainbows, you can still keep a bit of queue to stay safe. Notice that we're already a bit conservative in calculating the queue size from the RTT, as the "math" we use is `RTT+100`ms, meaning that even in case of no RTT, you'll still see a 100ms queue. We can probably re-evaluate all this later on, but I didn't want to be too disruptive in this first change.

Please test and provide feedback.


lminiero:
As a side note, this patch means you cannot completely disable NACKs as you could before with a `max_nack_queue=0`. Since all you can do is set a minimum value, even setting that to 0 means that a queue size will still be set according to the RTT, and the math I provided above. I don't see any good reason to disable NACKs entirely, so I don't see that as an issue, but if you have opinions, please share them.

atoppi:
I have a couple of comments:
1) we should still have a MAX length on the queue to avoid an uncontrolled growth, because just keeping up with RTT RTCP RR statistic is risky (that might be an issue due to a buggy or a malicious client)
2) the queue length should not be setted to the exact value calculated by the RTT stat, but somehow smoothed with multiple samples (e.g. moving average) in order to avoid big jumps of the queue size, followed by a lot of work to cleanup and refill the buffer (think about jittery networks)

fippo:
told @lminiero via twitter but that might be get lost with jetlag+travel: note that SRTP has replay protection window which seems to be used when HAS_DTLS_WINDOW_SIZE is defined. But then, that only applies for plain resends, not rtx ones

lminiero:
@fippo thanks for the reminder! Shouldn't be an issue for us though: if rtx is not used, the packet we keep in the buffer is already encrypted, so we don't use libsrtp again; if it is, we encrypt on a different ssrc/seq/ts space, and so shouldn't be an issue either.

lminiero:
Changed the old `-q` to `-Q`, since it has now a different meaning. Keeping it the same would have meant ambiguities and unexpected behaviour for people using the old setting, e.g., in their scripts.

itamayo:
Do you plan to add this option to the master ? It could be interesting ...

lminiero:
@itamayo well I haven't received any feedback at all, so it's hard to tell. If you're willing to give this a try and let us know, that will help things go faster.

itamayo:
We've test using videoroom plugin, with 4 videos 720p  bitrate 320kb/s published with gstreamer/webrtc client. 

Using Master branch the experience is quite good, but it freeze each video at least once per minute. Increasing video numbers this freeze increase equally. 

Using dinamyc_nac_min branch freezes are reduced by half. It's freeze every 2 minutes, better is nothing, thanks.

**PD: Freeze -> freezeCount on webrtc-internals**

lminiero:
Merging.

matiaspl:
The dynamic NACK queue management doesn't work as expected. It penalizes packets from the second half of a GOP effectively rendering them more likely to be non-recoverable using the NACK queue. More details here: https://groups.google.com/g/meetecho-janus/c/8cg4QntHeIs/m/jKKoe9wDBQAJ

PR #1870, opened 2019-11-20
Add audio skew compensation to janus-pp-rec.

atoppi:
Thanks to the enhancements done in #1719 (timestamp markers in mjr packets) we can now import the skew compensation algorithm (for details see the original PR #1153) into the janus post processing tool.
This initial effort just covers the audio skew compensation, as we wish to collect some feedback before trying to implement the same on the video packets.

We have added a new `S` parameter to `janus-pp-rec` where basically the user needs to specify a threshold in milliseconds. This value represents the maximum drift between remote and local clock before the algorithm executes a compensation (that might be either a packet drop or a sequence number / RTP timestamp jump). 
The default value is `0` (skew compensation is disabled).

By choosing high values the rate of false positives (due to packet jitter) is reduced, while lower values let the algorithm compensate more often.
Our preliminary tests have been conducted with a `-S 120` value, compensating tracks drifted by several seconds and restoring the lip-sync to an acceptable range.


tgabi333:
:+1:

lminiero:
Merging :+1:

PR #1873, opened 2019-11-25
SIP plugin: custom (non-standard) headers on incoming events (requests)

ihusejnovic:
This is the implementation of adding the option to receive custom (non-standard) headers on incoming SIP events (requests).
You need to specify the array of header prefixes in register request as 'incoming_header_prefixes'. Then, during some incoming SIP event (request), the headers that starts with some prefix from mentioned array will be forwarded further.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2708,18 +2708,18 @@ static void *janus_sip_handler(void *data) {
 				janus_mutex_unlock(&sessions_mutex);
 			}
 
-                        json_t *header_prefixes_json = json_object_get(root, "incoming_header_prefixes");
-                        if(header_prefixes_json && json_is_array(header_prefixes_json)) {
-                                size_t index;
-                                json_t *value;
+			json_t *header_prefixes_json = json_object_get(root, "incoming_header_prefixes");
+			if(header_prefixes_json && json_is_array(header_prefixes_json)) {
+				size_t index;
+				json_t *value;
 
-                                json_array_foreach(header_prefixes_json, index, value) {
-                                        const char *header_prefix = g_strdup(json_string_value(value));
-                                        session->incoming_header_prefixes = g_list_append(session->incoming_header_prefixes, header_prefix);
-                                }
-                        }
+				json_array_foreach(header_prefixes_json, index, value) {
+					const char *header_prefix = g_strdup(json_string_value(value));
+					session->incoming_header_prefixes = g_list_append(session->incoming_header_prefixes, header_prefix);
+				}
+			}
 
-                        /* If this is a refresh, get rid of the old values */
+			/* If this is a refresh, get rid of the old values */
 			if(refresh) {
 				/* Cleanup old values */
 				if(session->account.identity != NULL) {
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -119,6 +119,7 @@
 	"proxy" : "<server to register at; optional, as won't be needed in case the REGISTER is not goint to be sent (e.g., guests)>",
 	"outbound_proxy" : "<outbound proxy to use, if any; optional>",
 	"headers" : "<array of key/value objects, to specify custom headers to add to the SIP REGISTER; optional>",
+	"incoming_header_prefixes" : "<array of strings, to specify custom headers to read on incoming SIP events; optional>",
 	"refresh" : <true|false; if true, only uses the SIP REGISTER as an update and not a new registration; optional>",
 	"master_id" : <ID of an already registered account, if this is an helper for multiple calls (more on that later); optional>
 }
@@ -205,6 +206,7 @@
 	"call_id" : "<value of SIP Call-ID header for related call>",
 	"result" : {
 		"event" : "ringing",
+		"headers" : "<array of key/value objects, optional>"
 	}
 }
 \endverbatim
@@ -220,7 +222,8 @@
 	"call_id" : "<value of SIP Call-ID header for related call>",
 	"result" : {
 		"event" : "accepted",
-		"username" : "<SIP URI of the callee>"
+		"username" : "<SIP URI of the callee>",
+		"headers" : "<array of key/value objects, optional>"
 	}
 }
 \endverbatim
@@ -241,7 +244,8 @@
 	"call_id" : "<value of SIP Call-ID header for related call>",
 	"result" : {
 		"event" : "progress",
-		"username" : "<SIP URI of the callee>"
+		"username" : "<SIP URI of the callee>",
+		"headers" : "<array of key/value objects, optional>"
 	}
 }
 \endverbatim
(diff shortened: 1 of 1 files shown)


plugins/janus_sip.c
```diff
@@ -1383,6 +1397,28 @@ static void janus_sip_remove_quotes(char *str) {
...
+                                const char *header_name = g_strdup(unknown_header->un_name);
+                                json_object_set(headers, header_name, json_string(unknown_header->un_value));
+                                break;
+                        }
+                        temp = temp->next;
+                }
+                unknown_header = unknown_header->un_next;
+        }
+        return headers;
+}
```
lminiero: Indentation seems broken in this function: I see spaces instead of tabs. Besides, just for code style, please remove the extra space between `if`/`while` and the curly bracket.

plugins/janus_sip.c
```diff
@@ -1383,6 +1397,28 @@ static void janus_sip_remove_quotes(char *str) {
...
+	json_t *headers = json_object();
+	if (!sip) {
+		return headers;
+	}
+        sip_unknown_t *unknown_header = sip->sip_unknown;
+        while (unknown_header != NULL) {
+                GList *temp = session->incoming_header_prefixes;
+                while (temp != NULL) {
+                        char *header_prefix = (char *) temp->data;
+                        if (strncmp(unknown_header->un_name, header_prefix, strlen(header_prefix)) == 0) {
```
lminiero: This should check if `header_prefix` and `unknown_header->un_name` are not NULL.

plugins/janus_sip.c
```diff
@@ -2671,6 +2707,18 @@ static void *janus_sip_handler(void *data) {
...
 				janus_mutex_unlock(&sessions_mutex);
 			}
+
+			json_t *header_prefixes_json = json_object_get(root, "incoming_header_prefixes");
+			if(header_prefixes_json && json_is_array(header_prefixes_json)) {
+				size_t index;
+				json_t *value;
+
+				json_array_foreach(header_prefixes_json, index, value) {
+					const char *header_prefix = g_strdup(json_string_value(value));
```
lminiero: No validation is done on the JSON. First of all, it looks like `incoming_header_prefixes` is not in `register_parameters` (see how we do for `headers`), which means the plugin will not automatically reject it if it's not an array. Besides, you're assuming the content of that array will be a string (`json_string_value`), which can't be given for granted.

ihusejnovic:
Thanks for comments. I fixed all issues. Let me know if you find anything else.

plugins/janus_sip.c
```diff
@@ -1399,24 +1400,24 @@ static void janus_sip_remove_quotes(char *str) {
...
 	json_t *headers = json_object();
-	if (!sip) {
+	if(!sip)
 		return headers;
+	sip_unknown_t *unknown_header = sip->sip_unknown;
+	while(unknown_header != NULL) {
+		GList *temp = session->incoming_header_prefixes;
+		while(temp != NULL) {
+			char *header_prefix = (char *) temp->data;
+			if(header_prefix != NULL && unknown_header->un_name != NULL)
```
lminiero: There's missing brackets for this if.

ihusejnovic: Fixed.

plugins/janus_sip.c
```diff
@@ -1385,6 +1400,29 @@ static void janus_sip_remove_quotes(char *str) {
...
+	json_t *headers = json_object();
+	if(!sip)
+		return headers;
+	sip_unknown_t *unknown_header = sip->sip_unknown;
+	while(unknown_header != NULL) {
+		GList *temp = session->incoming_header_prefixes;
+		while(temp != NULL) {
+			char *header_prefix = (char *) temp->data;
+			if(header_prefix != NULL && unknown_header->un_name != NULL) {
+				if(strncmp(unknown_header->un_name, header_prefix, strlen(header_prefix)) == 0) {
```
lminiero: This should be case insensitive (`strncasecmp`).

plugins/janus_sip.c
```diff
@@ -2706,6 +2744,19 @@ static void *janus_sip_handler(void *data) {
 				g_hash_table_insert(masters, GUINT_TO_POINTER(session->master_id), session);
 				janus_mutex_unlock(&sessions_mutex);
 			}
+
+			json_t *header_prefixes_json = json_object_get(root, "incoming_header_prefixes");
+			if(header_prefixes_json) {
+				size_t index;
+				json_t *value;
```
lminiero: Both the above should be initialized.

plugins/janus_sip.c
```diff
@@ -206,6 +207,7 @@
 	"call_id" : "<value of SIP Call-ID header for related call>",
 	"result" : {
 		"event" : "ringing",
+		"headers" : "<array of key/value objects, optional>"
```
lminiero: The description (here and in the other occurrences) should be a bit more elaborate: e.g., something like "array of custom headers as key/value objects". Besides, the "optional" should come after a semicolon, not a comma (just as a convention, since it's how it's done in the rest of the docs).

plugins/janus_sip.c
```diff
@@ -1385,6 +1400,29 @@ static void janus_sip_remove_quotes(char *str) {
 	}
 }
 
+static json_t *janus_sip_get_incoming_headers(const sip_t *sip, const janus_sip_session *session) {
+	json_t *headers = json_object();
+	if(!sip)
+		return headers;
+	sip_unknown_t *unknown_header = sip->sip_unknown;
```
lminiero: Any reason why you're only traversing the unknown headers? I was under the assumption that this could be used to intercept any header, but then it wasn't returning anything when I attempted to intercept some common headers (e.g., "User-Agent") and noticed this. I guess this is what you mean by "custom" headers in the documentation, but it's actually not clear at all. As such, it should either be made more explicit in the documentation, or the `janus_sip_get_incoming_headers` implementation should be extended to any header, whether it's standard or not (which would sound more natural to me).

lminiero: After looking at the Sofia SIP [structure for messages](http://sofia-sip.sourceforge.net/refdocs/sip/structsip__s.html) in the docs, though, extending this to all headers seems pretty much impossible... Known headers seem to be all parsed to their own structure, and not available as a key/value list anywhere else like unknown headers are in `sip_unknown`. If so, then I'd just recommend to make it more explicit in the docs (and PR title) that the new feature only impacts non-standard headers.

ihusejnovic: Yes, you are right. I will update documentation.

lminiero:
Sorry, one more thing I noticed just now... `headers` in events is described as an "array of custom headers", but it's a JSON object, not an array. Thanks for the quick fixes done so far!

ihusejnovic:
> Sorry, one more thing I noticed just now... `headers` in events is described as an "array of custom headers", but it's a JSON object, not an array. Thanks for the quick fixes done so far!

No problem, changed.

lminiero:
Thanks, merging then! :+1:

PR #1875, opened 2019-11-26
Fix RTSP SETUP when url includes query string parameters (fixes #1869)

### LINKED ISSUE #1869: RTSP SETUP for URL with parameters fails
I'm currently testing #1866, but I'm quite positive this bug is happening on master as well (I could not find changes to relevant code parts when looking for the problem); if there is any doubt, I'm happy to actually test it with the latest release as well.

When configuring a RTSP streaming mountpoint URL

`url = "rtsp://192.168.2.10/axis-media/media.amp?videopsenabled=1"`

, a RTSP `DESCRIBE` is responded with the correct (video) control URL `rtsp://192.168.2.10/axis-media/media.amp/stream=1?videopsenabled=1`.

Unfortunately, when using URL parameters, Janus fails to check that [The control attribute already contains the whole URL](https://github.com/meetecho/janus-gateway/blob/b72992335e7a87c2161c69ed9b5f80ad3543ea97/plugins/janus_streaming.c#L5584):

```
if(strstr(vcontrol, source->rtsp_url) == vcontrol) {
	/* The control attribute already contains the whole URL? */
	g_snprintf(uri, sizeof(uri), "%s", vcontrol);
} else {
	/* Append the control attribute to the URL */
	g_snprintf(uri, sizeof(uri), "%s/%s", source->rtsp_url, vcontrol);
}
```

Since Janus doesn't ignore URL parameters when looking for `rtsp_url` in `vcontrol`, the **wrong branch** will be **executed**, sending this request:

`SETUP rtsp://192.168.2.10/axis-media/media.amp?videopsenabled=1/rtsp://192.168.2.10/axis-media/media.amp/stream=1?videopsenabled=1 RTSP/1.0`

which is answered with a 404.

I (or the camera, I guess) would expect a request like

`SETUP rtsp://192.168.2.10/axis-media/media.amp/stream=1?videopsenabled=1 RTSP/1.0`

lminiero:
As the title says. This is supposed to fix #1869, but I don't have RTSP servers that accept query string parameters to test against, so looking forward to feedback.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -5589,14 +5589,17 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 		g_free(curldata->buffer);
 		curldata->buffer = g_malloc0(1);
 		curldata->size = 0;
+		gboolean add_qs = (rtsp_querystring != NULL);
+		if(add_qs && strstr(vcontrol, rtsp_querystring) != NULL)
+			add_qs = FALSE;
 		if(strstr(vcontrol, rtsp_url) == vcontrol) {
 			/* The control attribute already contains the whole URL? */
 			g_snprintf(uri, sizeof(uri), "%s%s%s", vcontrol,
-				rtsp_querystring ? "?" : "", rtsp_querystring ? rtsp_querystring : "");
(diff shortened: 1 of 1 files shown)


lminiero:
Merging.

PR #1877, opened 2019-11-27
Remove option to enable rtx (now always supported, when negotiated)

lminiero:
For a while, we've kept RFC4588 (rtx) support behind a property you had to enable yourself, either via config or command line: without it, retransmissions would be sent the "old" way. Enabling it didn't mean it would be used, of course, since this still had to be negotiated via SDP: as such, the "old" way would still be used for endpoints that don't support it.

Considering there are advantages to using it when it's available, it doesn't make much sense to keep it that way anymore, so this patch removes that property from the configuration settings, and enables it by default. Again, endpoints not supporting it won't use it, so no harm done.

Submitted as a pull request just to isolate the changes on that part: I'm planning to merge soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,8 +281,6 @@ or on the command line:
 	-T, --ice-tcp                 Whether to enable ICE-TCP or not (warning: only
                                   works with ICE Lite)
                                   (default=off)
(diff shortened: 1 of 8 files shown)


lminiero:
Merging.

PR #1878, opened 2019-11-27
Binary data support in data channels

lminiero:
As the title says, this patch tries to add support for binary data to datachannels in Janus. As you probably know if you've been using datachannels in the past, so far we only supported text, so this should complement that part. Notice that this support is not extended to recordings, where we currently still only support postprocessing to `.srt` files, and so will only work as expected for text.

If you have your own plugin, notice that this introduces a change in both `incoming_data` and `relay_data`, both of which now include a new `textdata` boolean property to specify whether the data is text or binary. Check the diff on "plugin.h" for details: all stock plugins have been updated to take this into account, so you can check their diffs too for more info.

If you're just talking to Janus, instead, it should be pretty much transparent, since both sending and receiving data will depend on the type of data. Nothing changes in janus.js, for instance, where:

    echotest.data({data: "ciao"});

will send text data since the data is a string, while:

    echotest.data({data: new Uint8Array(16)});

will send binary data instead. Same thing for when you receive data: just look at the type of data you got in the `ondata` callback to understand what you just received.

I tested this briefly and it seems to be working as expected, but I'm not a great expert on using data from JavaScript, so not sure if there are shortcomings and pitfalls in our current bnary support. Feedback more than welcome!

### DIFF
--- a/conf/janus.plugin.streaming.jcfg.sample.in
+++ b/conf/janus.plugin.streaming.jcfg.sample.in
@@ -47,6 +47,7 @@
 #		passed, which would then change the current SSRC (0=disabled)
 # dataport = local port for receiving data messages to relay
 # dataiface = network interface or IP address to bind to, if any (binds to all otherwise)
+# datatype = text|binary (type of data this mountpoint will relay, default=text)
 # databuffermsg = true|false (whether the plugin should store the latest
 #		message and send it immediately for new viewers)
 # threads = number of threads to assist with the relaying part, which can help
@@ -162,6 +163,7 @@ file-ondemand-sample: {
 	#video = false
 	#data = true
 	#dataport = 5008
+	#datatype = "text"
 #}
 
 #
--- a/dtls.c
+++ b/dtls.c
@@ -965,10 +965,10 @@ int janus_dtls_verify_callback(int preverify_ok, X509_STORE_CTX *ctx) {
 }
 
 #ifdef HAVE_SCTP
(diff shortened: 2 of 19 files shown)


lminiero:
I should also clarify that this does have an impact on the Lua and Duktape APIs too: in both cases, `incomingData` and `relayData` are gone, and are now replaced by `incomingTextData`, `incomingBinaryData`, `relayTextData` and `relayBinaryData`. As such, if you're writing a Lua or Duktape plugin and are using datachannels, this would slightly change the way you do it.

oscarvadillog:
@lminiero I use the datachannel to send files as base64. What are the advantages/diffs of doing so instead of base64? I'll test soon... Thanks!

lminiero:
@oscarvadillog Base64 encoding adds overhead of about 33%, which means you send a lot more data that way.

lminiero:
> I should also clarify that this does have an impact on the Lua and Duktape APIs too: in both cases, `incomingData` and `relayData` are gone, and are now replaced by `incomingTextData`, `incomingBinaryData`, `relayTextData` and `relayBinaryData`. As such, if you're writing a Lua or Duktape plugin and are using datachannels, this would slightly change the way you do it.

To partially rectify this, I've made sure that the old callbacks and functions are still there in Lua and Duktape: when you try to use them, a warning is displayed on the console, and text datachannels are used automatically. Not planning to keep those forever, but they should keep backwards compatibility for the time being.

lminiero:
Merging.

PR #1880, opened 2019-11-28
Support for strings as unique IDs in AudioBridge, VideoRoom, TextRoom

lminiero:
All plugins use numeric identifiers for their resources: this is especially true for plugins that support the concept of rooms and participants, so AudioBridge, VideoRoom and TextRoom. This dates back to the very beginnings of Janus, where using numeric identifiers seemed to make more sense from a management and performance perspective, especially considering they'd be used as keys in hashtables, and compared often. That said, I've often seen requests in the past to support strings too, as many people rely on other unique identifiers, e.g., UUIDs, for their resources: having a different space in plugin forces people to implement their own mappings and keep them aligned, which can be cumbersome.

This patch is a first attempt to try and solve that. Specifically, it aims at adding support for string-based identifiers to AudioBridge, VideoRoom and TextRoom: it will NOT change how the core works (sessions and handles will still be numeric), nor will it touch other plugins (don't be greedy :wink: ). It does that by introducing a new static configuration properties to these plugins, called `string_ids`: this means that, when you start Janus, it's up to you to tell plugins if you'll use numeric identifiers as before (`string_ids=false`, the default), or strings (`string_ids=true`). You cannot mix the two, e.g., use strings in one room and numbers in another: it's one or another, decided at startup.

For the time being, I've only updated the AudioBridge, in order to figure out if/how this works and if it's worth it. As you can see from the amount of changes, it was a lot of work, so I won't start with the other plugins until I know that:

1. this is something people would be interested in (the lack of feedback on other PRs has been annoying);
2. the existing patch doesn't introduce regressions for numeric IDs (the plugin must still work exactly as expected if you change nothing);
3. it works exactly the same way if you use strings in all cases (feature parity);
4. we have no leaks or crashes due to the usage of strings.

Thanks, feedback welcome.


uxmaster:
I'm not interested :wink: since (in most of the use cases) there should be a middleman between Janus and the outside world anyway to orchestrate everything, including IDs. I like simplicity. 

_Disclamer: I've presented my point of view just to avoid the annoying lack of feedback._ :wink:

lminiero:
@uxmaster thx! Yeah, that's how I feel (and what I do) as well. But since I've seen this asked more than once, I thought I'd give a stab at it anyway. Thanks for the feedback!

akilude:
It would be a nice to have feature for us, it will simplify our rooms logic as we won't have to map room name to numeric ids.

lminiero:
It should now be available for all plugins, so AudioBridge, VideoRoom and TextRoom. I tested this with a few requests and it seems to be working as expected, after I fixed some bugs, but obviously this needs better testing, and by someone different than me to make sure I didn't miss anything.

@akilude since you mentioned you could make use of it, any chance you can give this a test? Which of those plugins are you using, by the way?

lminiero:
We're not going to enforce uniqueness on display names, only IDs.

chpl:
Having string IDs in Streaming plugin too will help in our case - and it will justify changing our system to test it.

lminiero:
I'm not going to work on yet another plugin until these have been tested.

akilude:
@lminiero i will test it soon, we use the videoroom plugin, Sorry for the late reply.

akilude:
@lminiero i have tested the video room plugin with string ids, it works fine, is there a list of characters that are disallowed in the id, for example, i think space and the "-" character don't work.

lminiero:
@akilude apologies, I think I missed your last message... we don't have limitations, they're just strings for us; not sure if libconfig enforces limits, though (for rooms added to the configuration file).

lminiero:
Merging.

lminiero:
@chpl please see the PR above for the Streaming plugin.

akilude:
@lminiero I tested videorooms by creating dynamic rooms by messaging the plugin not via the plugin config file.

I see that's it pushed to master, will test again soon.

Thanks.

Jeevhi:
> I'm not interested  since (in most of the use cases) there should be a middleman between Janus and the outside world anyway to orchestrate everything, including IDs. I like simplicity.
> 
> _Disclamer: I've presented my point of view just to avoid the annoying lack of feedback._

I agree! string changes are unnecessary.

PR #1881, opened 2019-11-28
added gtk-doc-tools package for ubuntu/debian deps

euskadi31:
libnice required `gtk-doc-tools`

see: 

```
./autogen.sh: 26: ./autogen.sh: gtkdocize: not found
```


lminiero:
Since it's a libnice dependency (and libnice must be installed manually, with its own set of instructions), I don't think it needs to be listed in our README.

tmatth:
@euskadi31 if you grab a release tarball from here: 
https://gitlab.com/libnice/libnice/releases
it already contains a configure script (no need to call autogen.sh), so if you run it with `./configure --disable-gtk-doc` you can drop the gtk-doc-tools dependency.

lminiero:
Good point, closing then :+1:

itsaboutcode:
I don't see this change in master repo.

PR #1885, opened 2019-12-03
Add CHANGELOG.md file into the project

oscarvadillog:
The idea is to add a CHANGELOG file that contains all the Janus releases notes, as it is currently only in the mailing list group.

What do you think of the following structure? Do you think it's good to separete into categories: Added, Removed, Fixed, Updated...?


lminiero:
Thanks! But not sure why master is there: I won't update the Changelog any time I commit something. I'd rather also get rid of the added/fixed/removed/etc. thing at the end, as I'm not going to get into the details of that either. What I'm willing to do is update the Changelog any time we tag a new version, by adding the same info I post on the group in the related section.

oscarvadillog:
Great! Updated CHANGELOG file.
Releases earlier than v0.2.1 are missing. Couldn't find on the group.

lminiero:
Yeah, we didn't advertise the very first versions. Maybe replace the question mark with a "TBD." or "Missing info", or something like that. I can fill in the blanks when I have some time.

oscarvadillog:
Already done 👍

oscarvadillog:
Already done 👍

lminiero:
Awesome, thanks! I guess that "Changelog added" will be one of the lines in the next summary :smile: 
Merging :+1:

PR #1886, opened 2019-12-04
New external logger to send logs to HOMER

lminiero:
This is a new logger that extends the functionality we added in #1814. Specifically, it allows you to send the Janus logs in real-time to a remote [HOMER](https://github.com/sipcapture/) instance. It can be compiled with a `--enable-homer-logger` passed to the configure script, and supports two different ways to push logs to HOMER:

1. Using native HEP, over either UDP or TCP;
2. Using HTTP, which actually "piggybacks" the existing HTTP event handler syntax by leveraging a currently unused event type.

You choose the protocol and backend in the `janus.logger.homerlog.jcfg` configuration file. The HTTP support actually depends on whether libcurl is available or not: if it isn't, the logger will be built with only HEP support. I plan to add gzip support as well (since log lines can be large, and HOMER seems to support it, at least on UDP/TCP), but I haven't come to that yet.

This was originally part of #1814 itself, but eventually, to keep things simple, I only merged the reference JSON-based logger implementation. One more reason is that, to be honest I haven't properly tested the logger either: I did test HEP/UDP against a barebone nodejs application using the `hep-js` stack, and it seems to work, but I did not test TCP, and neither did I test if this works with an actual HOMER instance due to lack of time. Besides, HTTP support actually needs some change in HEpop, since it introduces a new event type they're not aware of, and that should now be mapped to logs somehow. Pinging the awesome @lmangani to bring him up to speed.

Looking forward to feedback, especially with respect to testing this with the real thing!


lminiero:
Closing as this has fallen a bit behind, and we still don't have a clear idea of what the best way to sent logs to Homer is (maybe something that extends the existing event handlers API instead?). We can get back to this later.

lmangani:
Why closing instead of looping us in? It would be quite simple to arrange a meeting with our team and discuss this topic.

lminiero:
Hey Lorenzo, apologies, you're right! Just cleaning some old issues, and this was almost a year old :hand_over_mouth: 
I was indeed going to write you privately later today, as I wanted to pitch the idea of a new event handler type sent from the logger, which should make it easy to handle logs along regular events, so with minimum changes in the way you handle events already.

lmangani:
It's been a little while but hey, we'd like to revive this since technologies have advanced in the meantime! Is this still a good PR template or shall we rather start over @lminiero <3

lminiero:
@lmangani as far as the logger plugin API is concerned, this should indeed be a good template to start, as nothing changed there! Pretty sure that whatever I wrote at the time for the Homer logging format will be obsolete, though, so that part may need to be changed. And not sure if you support something other than HTTP that may be more optimized? Even when aggregating multiple events together, this may still result in multiple HTTP requests to be sent, and so may be inefficient as it is written here.

lminiero:
> And not sure if you support something other than HTTP that may be more optimized?

I mean, I know you do, but I meant specifically for remote logging :smile:

Issue #1861, opened 2019-11-08
Videocall: Remote video with poor quality/bitrate


oscarvadillog:
I'm testing Janus demos and have detected an issue with the quality of the video on one side.

The user who initiates the call sees the remote video with a poor quality (low bitrate ~30kbps). However, it doesn't happen on the other side, the peer who accepts the call has no problem with the remote video. I've tried different computers and networks, it happens all the time.

What could have changed in Janus to cause this? Previous releases didn't have this problem.

Tested here: https://janus.conf.meetecho.com/videocalltest.html

![videocall-issue](https://user-images.githubusercontent.com/4169970/68473039-3506c080-0222-11ea-8326-5cd331027cd2.JPG)

lminiero:
We currently support sender-side bandwidth estimation on the PeerConnection that sends the offer (the caller), not on the other side. Good chances are that this is the cause. We already have an open issue to track this requirement, but we don't know yet if and when we'll add support for BWE in Janus, as it's a very complex task. Closing.

oscarvadillog:
Thanks @lminiero 

I've tried the [demos ](https://janus.conf.meetecho.com/videocalltest.html) again and now the bitrate is good on both sides. What do you think was the problem? This morning I tried with different computers, networks and Janus servers. 

Do you think it can be improved by using the "set" request to configure the BW limit value (e.g: 256000) in the onremotestream event?

oscarvadillog:
This afternoon it worked well for me, but now I have the same problem again.
Why does it sometimes work and sometimes it doesn't? Remote video at 30kbps is very poor and laggy. Don't you think it's something important and makes the Videocall plugin unusable? I've been using the latest version of Chrome...

oscarvadillog:
I've tested in several browsers and it only happens in Chrome 78 (the latest stable version).
FF has always worked well for me.
Any idea what's going on @lminiero? Thanks!

oscarvadillog:
> We currently support sender-side bandwidth estimation on the PeerConnection that sends the offer (the caller), not on the other side. Good chances are that this is the cause. We already have an open issue to track this requirement, but we don't know yet if and when we'll add support for BWE in Janus, as it's a very complex task. Closing.

@lminiero I can't find the open issue you mentioned. I think this problem affects all users who use Videocall with Chrome...

oscarvadillog:
Below some stat charts using chrome insternals:

[...]

Issue #1473, opened 2018-12-14
Check if transport wide CC is supported


IvanMorozoff:
I'm testing janus/videocalltest.html with my local build.

```
commit 175eea78b2592382261190569ca3ff4d9b1608c1
Date:   Tue Nov 27 11:26:41 2018 +0100
```

I've discovered that videocalltest.htm and videoroomtest.html behave different.
In case of videoroomtest.htm I get less number of NACKs in my test environment.

In case of videocalltest.html 
handle->stream->do_transport_wide_cc are different (0 and 1) for the connected peers,
but both local and remote SDP contain extmap:5 and "transport-cc" lines.

Remote SDP:
```
a=extmap:5 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=rtcp-fb:101 transport-cc
```

Local SDP:
```
a=extmap:5 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=rtcp-fb:107 transport-cc
```

I suspect do_transport_wide_cc is not set properly in case of answer. See janus.c:

```
				if(!offer) {
...
				} else {
					/* Check if transport wide CC is supported */
					int transport_wide_cc_ext_id = janus_rtp_header_extension_get_id(jsep_sdp, JANUS_RTP_EXTMAP_TRANSPORT_WIDE_CC);
					handle->stream->do_transport_wide_cc = transport_wide_cc_ext_id > 0 ? TRUE : FALSE;
					handle->stream->transport_wide_cc_ext_id = transport_wide_cc_ext_id;
				}

```

lminiero:
Yes, initially it was done on purpose, as the main target was the VideoRoom, so publishers only (who always offer). We loosened that constraint to allow other plugins to use it, but I never thought about scenarios with equal peers like the VideoCall.

Honestly I'll have to think about that change: we only support the receiving part of this mechanism (we receive the sequence numbers in the RTP extension, and send RTCP feedback), not the other way around, and I don't know if this will mean the browser will start assuming we do, and possible make performance worse. Have you tried forcing it for answers as well to see if it fixes your use case?

IvanMorozoff:
Please clarify the following question:
Does this mean that in the case of VideoCall and if both peers support transport-cc
we always have connections as below:
```
caller<->janus: "transport_wide_cc"
answerer<->janus: "remb" 
```
?

lminiero:
[...]

PR #1889, opened 2019-12-05
Support for transport-wide CC on outgoing streams

lminiero:
This is an extension to the work started in #1884, where we did a bit of refactoring to give us more control on RTP extensions. That effort payed off here, as it allowed us to inject our own transport-wide RTP extension, which is the first step to implementing some form of bandwidth estimation in Janus.

More specifically, this patch implements the following changes:

1. It allows plugins to negotiate transport-wide CC when offering themselves, which was prevented before.
2. When negotiated, and when video is used, we add a global sequence number to the transport-wide CC extension (`http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01`). Before we only parsed the extension set by clients, in order to craft our RTCP feedback to send back: this now allows clients (e.g., browsers) to send us feedback via RTCP instead.
3. There is code in place to parse the transport-cc RTCP feedback sent by clients: we parse the whole message and build a list, which is currently ignored.

This whole effort will be the starting point to implement some actual BWE, which will happen in a separate PR. As it is, even ignoring the feedback we get back as we do now, it still improves some use cases that were partially broken before, e.g., sender-side BWE from both caller and callee in the VideoCall plugin (see #1473 and #1861). In fact, ensuring we negotiate the extension also when we offer ourselves, means browsers will add the extension in those sessions as well, and use the feedback we send back to adapt their bitrate: before, this was only possible when the client was offering (EchoTest, VideoCall caller, VideoRoom publisher, etc.)

Please make sure you test this and provide feedback, as it's an important step forward to make considerable improvements to the bandwitdh management in Janus. Since it's based on #1884, testing this branch you actually test both, and I **do** need feedback on both. Once I have a clearer idea on whether it works as expected and causes no issues, I'll be able to focus on the more complex stuff, like integrating some algorithm to do basic bandwidth estimation, and provide plugins (and applications) with more precise information, and possibly even automated behaviours.


janus.c
```diff
@@ -3407,6 +3409,10 @@ json_t *janus_plugin_handle_sdp(janus_plugin_session *plugin_session, janus_plug
 		}
 		if(ice_handle->stream && ice_handle->stream->mid_ext_id != mid_ext_id)
 			ice_handle->stream->mid_ext_id = mid_ext_id;
+		if(ice_handle->stream && ice_handle->stream->transport_wide_cc_ext_id != transport_wide_cc_ext_id) {
+			ice_handle->stream->do_transport_wide_cc = transport_wide_cc_ext_id > 0 ? TRUE : FALSE;
```
tmatth: nit: this could be simplified to `transport_wide_cc_ext_id > 0`

oscarvadillog:
Well done @lminiero! The problem with the videocall was solved. 
Tested on Chrome 78 🎉 🚀

tmatth:
@lminiero this just got reported and may be related to the various issues ppl have opened in janus around this topic:
https://bugs.chromium.org/p/webrtc/issues/detail?id=11196

lminiero:
@tmatth the issue people reported was caused by the fact VideoCall callees can't use transport-cc in master, because there we only allow it when browsers are the offerers. Since callees are answerers, in master transport-cc is not used for them, and the browser can't do its BWE, probably resulting in those low bitrates. It was confirmed that with this PR the issue is fixed, and that's because this patch enables transport-cc no matter who offers.

Not sure if we're impacted by the issue you mentioned, as we ignore padding packets (but we take them into account for transport-cc), and we currently don't generate our own (we'll probably need to in the future, for probing).

zayim:
Hi, Lorenzo! We tested this branch and it mainly works great, video quality is much better!

We did found 2 issues, however:

- Video gets rotated. Happens in video calls involving mobile phones (tested on Android), video stream coming from front camera is rotated 90° CW (it happens right from the beginning, without rotating phone, etc.). Does not happen on master branch.

- Sometimes, when making video calls from Chromium on Ubuntu (not Chrome) we get lots of:

`[ice.c:janus_ice_outgoing_traffic_handle:4096] [6423777842486859] Error setting transport wide CC sequence number...`

error logs. This happens when using SIP plugin (and when terminating to something that is not WebRTC, i.e. FreeSwitch).

lminiero:
> Hi, Lorenzo! We tested this branch and it mainly works great, video quality is much better!

Excellent, thanks for testing!



> * Video gets rotated. Happens in video calls involving mobile phones (tested on Android), video stream coming from front camera is rotated 90° CW (it happens right from the beginning, without rotating phone, etc.). Does not happen on master branch.

That may be because of something we're doing wrong handling the video orientation extension: we parse it, set it in the structure, and are then supposed to re-add it again, but that may not be happening for some reason.

> * Sometimes, when making video calls from Chromium on Ubuntu (not Chrome) we get lots of:
> 
> `[ice.c:janus_ice_outgoing_traffic_handle:4096] [6423777842486859] Error setting transport wide CC sequence number...`
> 
> error logs. This happens when using SIP plugin (and when terminating to something that is not WebRTC, i.e. FreeSwitch).

Not sure what's wrong there, browser shouldn't matter. I'll have a look.

lminiero:
@zayim do both of the issues happen on the SIP plugin, or were you testing on another plugin? Is the video-orientation header negotiated?

lminiero:
@zayim thinking about it, if this is happening with the SIP plugin, then it's expected. As explained in #1884, we strip all extension when we call `relay_rtp()`, and we only recreate them if they're set in the new structure. So, assuming A is using Janus and B is a regular SIP endpoint, this is what's happening:

1. A and B negotiate the video-orientation extension in their SIP call;
2. A adds extension to RTP packet;
3. Janus core parses extension, sets values in the struct, passes packet+struct to plugin;
4. SIP plugin relays packet (that still has extension, we don't remove it on the way in) via plain RTP to the peer;
5. B receives the extension, so everything's fine there;
6. B adds extension to its own RTP packet;
7. SIP plugin receives the packet, and relays it to the Janus core (extension still there);
8. Janus core strips all extensions, and adds them it is aware of (e.g., `mid`); since the SIP plugin didn't use the new struct to pass info on the video-orientarion extension, it is not re-added to the packet;
9. A receives RTP packet *without* extension.

Issue happens to B as well if they're behind Janus too, for the same reason.

Again, this is expected due to the changes in #1884, and the rationale and motivations are explained in detail there. It's up to plugins to fill the struct with info on extensions they want WebRTC users to receive, which means in this case the SIP plugin would need to be aware of extension IDs, parse packets on the way in to set the struct accordingly (a bit like the core does before calling `incoming_rtp`) and pass that metadata with the packet.

zayim:
Hi, Lorenzo! Thanks for the reply. Yes, both of issues are on SIP plugin. I understand what is issue with this rotated video. Do you guys have any plan to implement parsing packets and adding extensions for SIP plugin some times soon?

lminiero:
> Do you guys have any plan to implement parsing packets and adding extensions for SIP plugin some times soon?

Not right away, no. We're going on our holiday break soon, so it's likely this will have to wait until we get back.

lejlasolak:
Hi, we implemented adding video-orientation extension for SIP plugin and it seems to work. There is pull request for this.

jbboisseau:
Hi @lminiero 
This branch works great on our side. Do you know when you can merge it with the master?

lminiero:
@jbboisseau thanks for the feedback! We'll merge soon but not sure when: I still need to merge a couple of things first, then we'll create a new tag and merge this and other more "breaking" PRs in a new version.

lminiero:
Merging as, while we currently don't do anything with the feedback, it's still useful as it is anyway.

tmatth:
@lminiero I don't understand why yet, but this change is greatly reducing our outgoing FPS (the previous commit b501e170c5826b5fdb04493be38188a14dfdce79 is fine though). I will dig further on my end.

lminiero:
@tmatth yeah someone mentioned very low bitrates on the demos too. My guess is that something causes the extension NOT to be negotiated, or it is but there's no feedback, and so Chrome falls back to the minimum bitrate. This is weird because I remember the VideoCall working fine, so it may be a regression in monodirectional streams (unless VideoCall is broken too now). I'll have a look as soon as I am at the office.

lminiero:
@tmatth fixed in https://github.com/meetecho/janus-gateway/commit/58a8042c792dd4dbe06ee9a5d05a7bc4e1ce7bfa
It was a stupid typo, where I was checking whether `video_send` was TRUE, and I should have checked if `video_recv` was TRUE instead, since we only send the feedback when we _receive_ media from the peer. It was working as expected for bidirectional PeerConnections (e.g., EchoTest and VideoCall) because in that case `video_send` was indeed TRUE.

tmatth:
@lminiero thanks for the quick fix, confirmed that it's working as expected now.

PR #1898, opened 2019-12-18
Added support for real-time text (T.140 with or without red) to the SIP plugin

lminiero:
Real-time text is specified in [RFC4103](https://tools.ietf.org/html/rfc4103), and is basically a way to have live text conversations on RTP: the core difference with regular instant messaging is that you see the text as it's being typed, rather than when the other person sends the completed message. While it's been around for a while, it will be even more important in the future since it will be part of the so called "Next-Generation Emergency Services".

I'm writing a blog post on the whole effort, so I won't spend too many words on it here (you can refer to that once I share the link). The main thing you need to be aware of, from a Janus perspective, is that if you negotiate data channels in the SIP plugin, now, you'll attempt to establish a real-time text session (the data channel part of the SDP will be transformed to an `m=text` portion instead). Data will then be exchanged as an ArrayBuffer containing the T.140 blocks with the real-time text: I made a quick (and very ugly) update to the SIP demo to show this in action.

I only tested this for incoming calls from an old client called TIPcon1, and negotiating text only (no audio/video along text): the plan is to test this with other media and for outgoing calls as well (which should simply require us to offer data channels from the web application, nothing more), and possibly with other endpoints as well. Since I don't have access to other endpoints, though, I'm looking forward to feedback from people who do instead.


lminiero:
The blog post has been published, so you can find more info here:
https://www.meetecho.com/blog/realtime-text-sip-and-webrtc/

lminiero:
Extracted the changes to support data channel subprotocols in a separate PR (now merged), so this PR should now be much more focused on the SIP plugin changes to get T.140 working instead.

lminiero:
I'm closing this PR, as I don't see it being merged anytime soon. Should people be interested in reviving this, please let me know.

PR #1901, opened 2019-12-19
fix janus crash when pidfile is locked by another process

aisatis:


lminiero:
Not sure what fix you wanted to contribute, but this pull request is broken. It tries to modify a ton of files, and looks like it's based on either #1884 or #1889 rather than master. Please provide a PR based on master instead. Closing.

PR #1903, opened 2019-12-20
startup: only close the logger directory if it was opened

kazzmir:
Janus opens the `loggers` directory on startup to scan for logger plugins, but in the case that the `loggers` directory does not exist janus will still try to close the directory, leading to a segfault.
```
DIR * dir = opendir();
if (!dir){
   ...
} else {
   ...
}
closedir(dir);
```
This simply moves the closedir() call to the else branch where the dir is known to be open.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -3965,8 +3965,8 @@ gint main(int argc, char *argv[])
 				g_hash_table_insert(loggers_so, (gpointer)janus_logger->get_package(), event);
 			}
 		}
+		closedir(dir);
 	}
-	closedir(dir);
 	if(disabled_loggers != NULL)
 		g_strfreev(disabled_loggers);
 	disabled_loggers = NULL;


lminiero:
Sorry for the late response, just saw this as I was on vacation for the holidays. I'll have a look tomorrow and review. That said, please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus/) as that's needed for me to be able to merge.

lminiero:
Looks like the same fix is needed for event handlers as well: media and transport plugins, instead, have an `exit` when the folder is missing or invalid, so the `closedir` in that case always happens it has indeed been opened.

lminiero:
I can see the CLA has been signed, so I'll merge right away, thanks spotting the issue and providing a fix! I'll apply the fix on event handlers myself.

PR #1906, opened 2019-12-27
Fix out of bounds array access for last_spatial_layer

tmatth:
Caught by clang:
```
plugins/janus_streaming.c:3981:2: warning: array index 2 is past the end of the array (which contains 2 elements) [-Warray-bounds]
        session->last_spatial_layer[2] = 0;
```


lminiero:
Sorry for the late response, going back to the office tomorrow and I'll review :+1:

lminiero:
You're right, some time ago we added support for 3 SVC layers, but didn't update that array. Thanks for spotting this, merging!

PR #1911, opened 2020-01-03
[videoroom] Fix race condition when joining and destroying the session

atoppi:
Fix a possible race condition when joining as a subscriber and destroying the peer session.


lminiero:
Merging :+1:

PR #1918, opened 2020-01-08
Use sendBeacon instead of sync XHR in onbeforeunload (fixes #1902)

### LINKED ISSUE #1902: Update janus.js for session destroy function onbeforeunload not working in browser's newest versions
Chrome has "disabled" sending Sync XHR requests to prevent end user staying too long in a page where has a sync request when leaving the page, which can impact in end user experience (in their opinion of course).

To prevent errors when calling the session destroy function on page unload, the function must be updated to use `sendBeacon()`, which is made for this kind of tasks. The function does an async call, but can be used on before unload event in almost any browser (except IE, as always...)

Here are the links related to this problem:

- Chrome feature removal: https://www.chromestatus.com/feature/4664843055398912
- MSDN: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon
- W3C: https://w3c.github.io/beacon/
- CanIUse: https://caniuse.com/#feat=beacon

lminiero:
See #1902 for details. Seems to be working fine in my tests, and apparently `sendBeacon` is pretty much [universally available](https://caniuse.com/#feat=beacon) now, but feedback is still welcome in case this doesn't work on some browsers.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -968,7 +968,26 @@ function Janus(gatewayCallbacks) {
 		}
 		if(!connected) {
 			Janus.warn("Is the server down? (connected=false)");
+			sessionId = null;
 			callbacks.success();
+			return;
+		}
+		if(unload) {
+			// We're unloading the page: use sendBeacon for HTTP instead,
+			// or just close the WebSocket connection if we're using that
+			if(websockets) {
+				ws.onclose = null;
+				ws.close();
+				ws = null;
+			} else {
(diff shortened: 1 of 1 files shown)


oscarvadillog:
Great @lminiero - it works properly in my tests. ❤️ **sendBeacon** ❤️

lminiero:
Thanks @oscarvadillog, merging then :+1:

PR #1919, opened 2020-01-09
Add new configuration property to add protected folders not to save to

lminiero:
This PR tries to address a feature we've been missing so far, that is the configuration of a list of folders we shouldn't write to, at least when it comes to files that may be written as a consequence of an API call. To make a simple example, you might ask for a .mjr recording or a .pcap capture, and saved to, e.g., `/usr/bin/`, at the risk of causing problems to the instance. This patch adds a new property you can add to `janus.jcfg`, an array called `protected_folders`, which contains the base folders you don't want Janus to write to, e.g.:

	protected_folders = [
		"/bin",
		"/boot",
		"/dev",
		"/etc",
		"/initrd",
		"/lib",
		"/lib32",
		"/lib64",
		"/proc",
		"/root",
		"/sbin",
		"/sys",
		"/usr",
		"/var",
		"/opt/janus/bin",
		"/opt/janus/etc",
		"/opt/janus/include",
		"/opt/janus/lib",
		"/opt/janus/lib32",
		"/opt/janus/lib64",
		"/opt/janus/sbin",
	]

With such a list in effect, trying to save an .mjr file in `/usr/bin` would fail since the `/usr` exception would be triggered. At the moment, this is only enforced on attempts to start a new recorder and new pcap/text2pcap captures, that is what you can usually ask for via API: as such, it does NOT include, for instance, configuration files, or files that plugins open themselves using folders specified in their configuration files (although plugins may of course be modified to use that check too). This is because configuration files are supposed to be only accessible to who manages the Janus instance, and so if you write a wrong folder there, you're basically shooting your own foot, while API calls are outside of your control.

By default, no folder is protected, Janus only relies on what is written in `janus.jcfg`: since by default, even after updates, the existing `janus.jcfg` is not touched, existing deployments would need an update to their configuration file to start taking advantage of this. New instances instead should be able to use it right away, as the value of `protected_folders` is uncommented in the sample configuration file. Not sure whether we should have a hardcoded list to fallback to when such a value is missing in the code itself: on one side it would start protecting deployments, but on the other end it might break existing applications that for one reason or another are using those folders.

The underlying implementation uses [realpath](http://man7.org/linux/man-pages/man3/realpath.3.html) under the curtains, meaning it's supposed to be smart enough to take into account relative paths and symbolic links. That said, it might not be perfect, so any improvement here would be welcome, as well as generic feedback on whether you think this is useful/effective or not as it is.

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -138,7 +138,6 @@ general: {
 		"/sbin",
 		"/sys",
 		"/usr",
-		"/usr/sbin",
 		"/var",
 			# We add what is usually the folders Janus is installed too
 			# as well: we don't just put "/opt/janus" because that would
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -124,6 +124,8 @@ general: {
 		# you can configure an array of folders that Janus should prevent
 		# creating files in. By default no folder is protected, unless the
 		# 'protected_folder' property below is uncommented and configured.
+		# Notice that at the moment this only covers attempts to start
+		# an .mjr recording and pcap/text2pcap packet captures.
 	protected_folders = [
 		"/bin",
 		"/boot",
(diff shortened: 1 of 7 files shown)


atoppi:
lgtm :+1:

lminiero:
Merging.

PR #1922, opened 2020-01-16
Fixed bug about parsing RTSP response

kefir266:
'Session:' point was newer reached after 'Transport:' processing because of changed \n to 0 character in 'Transport:' line

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -5713,11 +5713,6 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 			header = strstr(curldata->buffer, "transport:");
 		if(header != NULL) {
 			char *end = strchr(header, '\r');
-			if(end != NULL)
-				*end = '\0';
-			end = strchr(header, '\n');
-			if(end != NULL)
-				*end = '\0';
 			/* Iterate on all params */
 			char *p = header, param[100], *pi = NULL;
 			int read = 0;
@@ -5753,7 +5748,7 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 				}
 				/* Move to the next param */
 				p += read;
-				if(*p != ';')
+				if(p >= end)
 					break;
 				while(*p == ';')
 					p++;
@@ -5765,11 +5760,6 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 			header = strstr(curldata->buffer, "session:");
 		if(header != NULL) {
 			char *end = strchr(header, '\r');
-			if(end != NULL)
-				*end = '\0';
-			end = strchr(header, '\n');
-			if(end != NULL)
-				*end = '\0';
 			/* Iterate on all params */
 			char *p = header, param[100], *pi = NULL;
 			int read = 0;
@@ -5793,7 +5783,7 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 				}
 				/* Move to the next param */
 				p += read;
-				if(*p != ';')
+				if(p >= end)
 					break;
 				while(*p == ';')
 					p++;


lminiero:
I'm not sure this is a valid fix. It may solve your specific issues (camera possibly not using \r\n as it should per the standard?), but I suspect it will cause problems with all other ones. Please share an example of parsed RTSP message that caused problems to you (possibly a pcap capture as otherwise termination characters will be lost).

kefir266:
I don't see any reason of `\r` . I rewrite this code only for not to loop when \r encountered. It's not my case I don't have this one in my response. Maybe it should be removed.

My case:
`00000000│ 52 54 53 50 2f 31 2e 30 ┊ 20 32 30 30 20 4f 4b 0a │RTSP/1.0┊ 200 OK_│
│00000010│ 43 53 65 71 3a 20 32 0a ┊ 54 72 61 6e 73 70 6f 72 │CSeq: 2_┊Transpor│
│00000020│ 74 3a 20 52 54 50 2f 41 ┊ 56 50 3b 75 6e 69 63 61 │t: RTP/A┊VP;unica│
│00000030│ 73 74 3b 63 6c 69 65 6e ┊ 74 5f 70 6f 72 74 3d 31 │st;clien┊t_port=1│
│00000040│ 30 30 30 30 2d 31 30 30 ┊ 30 31 3b 73 65 72 76 65 │0000-100┊01;serve│
│00000050│ 72 5f 70 6f 72 74 3d 35 ┊ 30 30 33 32 2d 35 30 30 │r_port=5┊0032-500│
│00000060│ 33 33 3b 73 73 72 63 3d ┊ 32 31 43 33 37 36 46 45 │33;ssrc=┊21C376FE│
│00000070│ 3b 6d 6f 64 65 3d 22 50 ┊ 4c 41 59 22 0a 53 65 72 │;mode="P┊LAY"Ser│
│00000080│ 76 65 72 3a 20 47 53 74 ┊ 72 65 61 6d 65 72 20 52 │ver: GSt┊reamer R│
│00000090│ 54 53 50 20 73 65 72 76 ┊ 65 72 0a 53 65 73 73 69 │TSP serv┊er_Sessi│
│000000a0│ 6f 6e 3a 20 33 4b 6f 72 ┊ 71 70 77 50 5a 43 68 55 │on: 3Kor┊qpwPZChU│
│000000b0│ 70 31 4e 39 3b 74 69 6d ┊ 65 6f 75 74 3d 36 30 0a │p1N9;tim┊eout=60_│
│000000c0│ 44 61 74 65 3a 20 57 65 ┊ 64 2c 20 31 35 20 4a 61 │Date: We┊d, 15 Ja│
│000000d0│ 6e 20 32 30 32 30 20 31 ┊ 31 3a 35 34 3a 34 30 20 │n 2020 1┊1:54:40 │
│000000e0│ 47 4d 54 0a ┊ │GMT ┊ │`

before when it encountered \n it changed to \0 in header **Transport:**
`		/* Find the Transport header and parse it */
		char *header = strstr(curldata->buffer, "Transport:");
		if(header == NULL)
			header = strstr(curldata->buffer, "transport:");
		if(header != NULL) {
			char *end = strchr(header, '\r');
			if(end != NULL)
				*end = '\0';
			end = strchr(header, '\n');
			if(end != NULL)
				*end = '\0';`
then in header **Session**: `header = strstr(curldata->buffer, "Session:");` will not work because of \0 after **Transport:** line

If you give me example with \r\n a would take decision what to do with \r

lminiero:
> I don't see any reason of `\r` .

You may not see a reason for it, but per the standard, lines are terminated by a CRLF (Carriage Return, Line Feed), which is a `\r\n`. We replace `\r` with `\0` because we want to parse a single line, and so terminating it there that's where our string "ends". From your snippet, the RTSP message only includes a `\n` (`0a`), which is a broken, non-standard message: https://tools.ietf.org/id/draft-ietf-mmusic-rfc2326bis-33.html#rfc.section.5

That's why the code is not working as expected in your case. While we can try to add an exception for broken clients that don't follow the standard (we have a similar thing in the SDP utils in the core too), I want to make sure that exception won't break RTSP parsing of _legitimate_ messages instead, which I'm not sure your patch ensures. I'll take note of this, and see if I can come up with a simple fix: again, though, if you can share a .pcap dump of the broken RTSP message (instead of the text copy you pasted here), that would help.

kefir266:
You are right about \r. It's there, I changed limiter on the \r, but I try to say we mustn't to change this characters to \0 because of *p points to the same string curldata->buffer and when we look up strstr(curldata->buffer, "Session:") It can't reach it because we put \0 after **Transport:** string

lminiero:
@kefir266 please see #1925, as I used a similar approach to the one we have in the SDP utils to be transparent to `/n` vs `/r/n`: the code is also more "compact", now, which should help too.

kefir266:
It works for, but I don't have audio. Most of all it would  work for a stream with audio as ka_timeout was set in video section. There is the same block for audio since 5838 line

lminiero:
> There is the same block for audio

Whoops, you're right, I only did it for one of the two... I'll fix it shortly.

lminiero:
@kefir266 it should be fixed now, thanks for the quick feedback :+1:

lminiero:
Closing as we have the other PR. Thanks for bringing this to our attention and proposing a fix!

PR #1925, opened 2020-01-17
Better parsing of RTSP messages (see #1922)

lminiero:
Attempt to improve RTSP parsing, especially when dealing with non-standard-compliant RTSP endpoints that only use `/n` instead of `/r/n`. Replaces #1922.

I don't have a way to test it with such a broken RTSP server, so please let me know if this does indeed work as expected. I tested this with a couple of public RTSP servers successfully, so there should be no regression, but again, please make sure you don't see any issue with your RTSP usage.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -5707,97 +5707,87 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 			return -5;
 		}
 		JANUS_LOG(LOG_VERB, "SETUP answer:%s\n", curldata->buffer);
-		/* Find the Transport header and parse it */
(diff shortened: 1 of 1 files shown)


lminiero:
@kefir266 did you have a chance to test this after I fixed the missing audio part? I'd like to merge this soon.

kefir266:
> @kefir266 did you have a chance to test this after I fixed the missing audio part? I'd like to merge this soon.

I have done the second commit too but only with video stream. I don't have appropriate source. If you give me some link with audio I could check it quickly.

lminiero:
I don't have RTSP streams to test against. I'll assume this fixes it, then.

lminiero:
Merging.

PR #1928, opened 2020-01-20
added video orientation and audio level extensions

lejlasolak:


lminiero:
Hoped the new title would update the commit message, but nope... nevermind then, thanks again for the contribution! I'll update the `transport-cc` branch shortly.

PR #1929, opened 2020-01-22
Fix g_async_queue usage

tmatth:
This PR is a continuation of https://github.com/meetecho/janus-gateway/pull/1121 with the addition that it checks that every queue was allocated beforehand. The deleted `if (popped) continue` are therefore guaranteed to be dead code.

The main rationale for deleting the `if (popped) continue` blocks is that these obscure the fact that calls to `g_async_queue_pop` are in fact blocking.

### DIFF
--- a/events.c
+++ b/events.c
@@ -269,8 +269,6 @@ void *janus_events_thread(void *data) {
 	while(eventsenabled) {
 		/* Any event in queue? */
 		event = g_async_queue_pop(events);
-		if(event == NULL)
-			continue;
 		if(event == &exit_event)
 			break;
 
--- a/events/janus_mqttevh.c
+++ b/events/janus_mqttevh.c
@@ -964,10 +964,6 @@ static void *janus_mqttevh_handler(void *data) {
 	while(g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping)) {
 		/* Get event from queue */
 		event = g_async_queue_pop(events);
-		if(event == NULL) {
-			/* There was nothing in the queue */
-			continue;
-		}
 		if(event == &exit_event) {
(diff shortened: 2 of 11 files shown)


lminiero:
But won't Glib just abort when it can't allocate something? If so, we can simply get rid of the checks where they don't matter (in some cases they were needed because queues might be deallocated in the meanwhile). I don't think the "memory errors" you added in some places are needed.

tmatth:
> But won't Glib just abort when it can't allocate something? 

Ack, you're totally right. Well, at least I'll be able to shorten this PR quite a bit ;)

lminiero:
Let me know when the PR has been shortened and I'll review again! :+1:

tmatth:
@lminiero updated

lminiero:
I thought you'd remove some of the checks we have on whether the queue exists, before reading it, as it's supposed to always exist: not sure about removing the NULL check after a `g_async_queue_pop` though. That is indeed blocking and supposed to return "something", i.e., whatever is in the queue, but what if we actually enqueued NULL? Does `g_async_queue_push` do nothing, when you queue NULL as a pointer, or is it a valid value?

tmatth:
> That is indeed blocking and supposed to return "something", i.e., whatever is in the queue, but what if we actually enqueued NULL? Does `g_async_queue_push` do nothing, when you queue NULL as a pointer, or is it a valid value?

lminiero:
> From the doc (https://developer.gnome.org/glib/stable/glib-Asynchronous-Queues.html#g-async-queue-push):

I've read all those doc pages so many times, that I should really be ashamed for missing that... thanks for the clarification! Merging then :+1:

PR #1932, opened 2020-01-29
Janus Travis CI integration

atoppi:
This PR integrates Travis CI in Janus and basically resumes and extends the #955 by @fippo.

We're checking three different Janus builds:
- complete setup (w/ data channels, all plugins, all transports, all event handlers, all loggers etc.)
- complete setup w/o data channels
- complete setup w/o libcurl

All of these are compiled with Linux amd64 on both gcc and clang.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,13 @@ env:
   global:
     - JANUS_COMMON_OPTS="--prefix=/opt/janus --disable-docs --enable-post-processing --enable-plugin-lua --enable-plugin-duktape --enable-json-logger"
   jobs:
-    - JANUS_CONFIG_OPTS="" LIBCURL=YES
-    - JANUS_CONFIG_OPTS="--disable-data-channels" LIBCURL=YES
-    - JANUS_CONFIG_OPTS="--disable-turn-rest-api --disable-sample-event-handler" LIBCURL=NO
+    - JANUS_CONFIG_OPTS="" LIBCURL=YES DATACHANNELS=YES
+    - JANUS_CONFIG_OPTS="--disable-data-channels" LIBCURL=YES DATACHANNELS=NO
+    - JANUS_CONFIG_OPTS="--disable-turn-rest-api --disable-sample-event-handler" LIBCURL=NO DATACHANNELS=YES
 
 git:
   quiet: true
+  depth: 1
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,67 @@
...
+  - git clone -b master https://github.com/sctplab/usrsctp usrsctp
+  - pushd usrsctp && ./bootstrap && ./configure --prefix=/usr --disable-static --disable-debug --disable-programs && make -j$(nproc) && sudo make install && popd
+  - git clone -b v3.2-stable https://github.com/warmcat/libwebsockets.git libwebsockets
+  - pushd libwebsockets && mkdir -p build && pushd build && cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr -DLWS_WITH_STATIC=OFF -DLWS_WITHOUT_CLIENT=ON -DLWS_WITHOUT_TESTAPPS=ON -DLWS_WITHOUT_TEST_SERVER=ON -DLWS_WITH_HTTP2=OFF .. && make -j$(nproc) && sudo make install && popd && popd
+  - git clone -b master https://github.com/alanxz/rabbitmq-c rabbitmq-c
+  - pushd rabbitmq-c && git submodule update --init && mkdir build && pushd build && cmake -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_EXAMPLES=OFF -DBUILD_TESTS=OFF -DBUILD_TOOLS=OFF .. && sudo cmake --build . --target install -- -j$(nproc) && popd && popd
+  - git clone -b master https://github.com/eclipse/paho.mqtt.c.git paho.mqtt.c
+  - pushd paho.mqtt.c && cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr -DPAHO_WITH_SSL=TRUE -DPAHO_BUILD_SAMPLES=FALSE -DPAHO_BUILD_DOCUMENTATION=FALSE && make -j$(nproc) && sudo make install && popd
```
tmatth: You may want to replace some of these clone commands with e.g., `curl -fSL https://github.com/warmcat/libwebsockets/archive/v${LIBWEBSOCKETS_VERSION}.tar.gz`.

 Also sticking to specific versions may be easier to debug (but github does let you grab tarballs of `master` or arbitrary  .commits too).

Otherwise, if you want to keep doing `git clone` I'd suggest adding `--depth 1`.

alexamirante: +1 for `--depth 1`. I don't see any advantage in getting rid of `git clone` in favor of getting a tarball via http.

atoppi:
Thanks for the heads up @tmatth.
I have added depth 1 to git configuration.

.travis.yml
tmatth: Yeah it'd only be in a situation where you didn't want the extra dependency of git or were doing http caching (via a proxy) and so needed to download over plain http.

README.md
```diff
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/meetecho/janus-gateway.svg?branch=master)](https://travis-ci.org/meetecho/janus-gateway)
+
```
lminiero: This should be below the name of the project, not above.

lminiero: You can add it before the new badge I just added for Coverity.

lminiero:
Looks good to me, merging :+1:

PR #1938, opened 2020-02-06
AudioBridge: The target should also recevied latest status

klanjabrik:
The target of participant should also acknowledge the latest mute/unmute status which has been made by administrator.


lminiero:
Thanks for the contribution! The title of the pull request should clarify this is for the AudioBridge, though.
As a side note, it looks like your patch is failing to build on Travis, so you may want to fix that.

lminiero:
> As a side note, it looks like your patch is failing to build on Travis, so you may want to fix that.

Scratch that, looks like it's a git clone issue on one of the dependencies. We'll try to have the test repeated.

lminiero:
Thanks, merging! :+1:

PR #1940, opened 2020-02-06
Configurable global prefix for log lines

lminiero:
A simple patch that adds a new configuration property, to specify a prefix that should appear in all log lines. For instance, setting

    log_prefix = "[janus] "

in `janus.jcfg` will result in something like this:

```
Janus commit: 77b710c8efd1a93ecff82ba3c444ffc4d39622ec
Compiled on:  Thu  6 Feb 18:05:19 CET 2020

[janus] Logger plugins folder: /home/lminiero/test/logs/lib/janus/loggers
[janus] Loading logger plugin 'libjanus_jsonlog.so'...
[janus] [WARN] JSON logger disabled
```

Terminal colors are supported, so something like this:

    log_prefix = "[\x1b[32mjanus\x1b[0m] "

will do the same thing, but with "janus" in green.

This is probably not very useful, and maybe redundant considering that this kind of flexibility can be better achieved with a custom logger, but it might be helpful in some cases anyway. Will merge soon, unless I'm told something's very wrong.

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -18,6 +18,11 @@ general: {
 	#debug_timestamps = true				# Whether to show a timestamp for each log line
 	#debug_colors = false					# Whether colors should be disabled in the log
 	#debug_locks = true						# Whether to enable debugging of locks (very verbose!)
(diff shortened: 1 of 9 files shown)


lminiero:
Merging.

PR #1941, opened 2020-02-07
Travis libnice clang flags

atoppi:
Add `-Wno-cast-align` when building `libnice` with clang on Travis.


lminiero:
Looks like it fixes the Travis failure, so merging :+1:

PR #1946, opened 2020-02-15
Feature/allow ice tcp without ice lite

sjkummer:
We should allow to have ICE-TCP enabled without ICE Lite. Recent versions of libnice allow this combination and gather tcp passive candidates etc. in this setup. Let's keep a warning there if there are known problems in specific version. However, both settings (ICE-TCP and ICE Lite) are disabled by default - and it is still up to the configurator which combination should be used.


lminiero:
Unfortunately IIRC there's no version information available in `nice.h` that we can use to selectively disable the feature if we know it will not work. That said, your change to the config should hopefully make it clear that it can cause issues (to those who do read the documentation, that is, but that's a rant for another time :smile: ), so thanks for the suggestion!

Merging :+1:

PR #1952, opened 2020-02-18
conf: document events option

tmatth:


lminiero:
Re-reading the text, I think there may be some ambiguity, as by

    Whether to send event notifications

people may think they refer to asynchronous events through this transport. Could you change the text in the config comments to something like this instead?

    Whether to notify event handlers about transport-related events

Not sure if that text is clear enough: suggestions welcome, if not! Apart from this, this looks good so I'll merge as soon as this is updated :+1:

lminiero:
Thanks for the quick fix, merging! :+1:

PR #1956, opened 2020-02-20
Fix deadlock during dynamic mountpoint creation

### LINKED ISSUE #1955: Unified-plan: Deadlock during dynamic rtp mountpoint creation
Sending "Create" request to the streaming plugin causes a deadlock (there is no response for this an future requests).

I have found this caused by trying to get locked mutex (_mountpoints_mutex_ in _janus_streaming_create_rtp_source_ function).

k0ntre201:
Fixes #1955.
Before calling function _janus_streaming_create_rtp_source_ the _mountpoints_mutex_ has to been unlocked.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -2994,6 +2994,8 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 					goto prepare_response;
 				}
 			}
+			janus_mutex_unlock(&mountpoints_mutex);
+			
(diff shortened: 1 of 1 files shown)


lminiero:
As I was saying in response to your other issue, it makes little sense to apply fixes on #1459 as it is. It's not going to be merged the way it is now: I'll have to work on a new version. As soon as that's done, if this is still an issue we can apply it there.

lminiero:
Closing for now: we can reference it again when a new pull request for multistream is available.

PR #1715, opened 2019-07-26
plugins: drop dead code related to rtp_header_update

tmatth:

PR #1964, opened 2020-02-26
rtp: drop dead code in rtp_header_update callers

tmatth:
This is in the same vein as #1715


lminiero:
Makes sense, thanks! Merging :+1:

PR #1965, opened 2020-02-27
Detect H264 key frames with smaller SPS units

cameronlucas3:
Reduces the H264 keyframe length check from 16 to 6 bytes.
6 bytes seems to be the lower bound of any possibly valid SPS NAL unit,
based on Section 7.3 of the H264 specification.

For reference, we have been observing Chrome 80 producing SPS units
of 12 bytes or less.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -742,7 +742,7 @@ gboolean janus_vp9_is_keyframe(const char *buffer, int len) {
 }
 
 gboolean janus_h264_is_keyframe(const char *buffer, int len) {
-	if(!buffer || len < 16)
+	if(!buffer || len < 6)
 		return FALSE;
 	/* Parse H264 header now */
 	uint8_t fragment = *buffer & 0x1F;


lminiero:
Thanks for the clarification and the fix! Have you verified that this does indeed result in higher chances of detecting a keyframe? Wondering in particular if this was affecting H.274 simulcast support.

Can you sign our [CLA](https://cla.conf.meetecho.com/janus/) before we can merge this?

cameronlucas3:
CLA is signed. 

In regards of detecting a keyframe we have measurements of it taking over 6 seconds from the start of a stream for Chrome 80 to send a large enough SPS for a keyframe be detected. This causes it to miss the very first keyframe and sometimes multiple keyframes after. With this change it can reliably detect the first keyframe.

lminiero:
Thanks, merging then! :+1:

PR #1969, opened 2020-02-28
Support for strings as unique mountpoint IDs in Streaming plugin

lminiero:
Pretty much the same thing as #1880, but for the Streaming plugin. In particular, allows the usage of strings as mountpoint IDs if `string_ids` is set to `TRUE` in the configuration file. Just as in the other patch, numbers and strings are mutually exclusive in the running instance: you can choose to either use one, or the other, not both.

Not really tested, so feedback welcome.


lminiero:
Merging.

PR #1972, opened 2020-03-02
Remove onmute onunmute

groupboard:
The onmute/onunmute handlers were added on Jun 25, 2019. However, this causes unwanted behaviour on chrome. The behaviour of chrome seems to be that it mutes a stream when no data has been received for roughly 2 seconds. So, if there is any temporary loss of incoming video for a few seconds, this will result in "No remote video available" appearing in the video window, and then disappearing, which confuses the user.

It can be demonstrated in a videoroom page by opening a cpu intensive website in another tab of chrome (e.g. the weather network), then dragging that tab out into a separate browser window, and moving it around the screen. Try it a few times, and you'll sometimes see "no remote video available" flash up temporarily.

A better option is to just remove the onmute/onunmute methods, unless there is some reason they need to be handled. That way the video will just pause momentarily. If the connection is completely lost for whatever reason, the onended event handler will be triggered.

If we want to notify the user that the connection has paused, it might be better to just add some kind of icon onto the paused video feed.


groupboard:
Will the "ended" event not be called in those cases? The mute event is "sent to a MediaStreamTrack when the track's source is temporarily unable to provide media data. When the track is once again able to produce media output, an unmute event is sent." So it's always a temporary thing...if the track is permanently removed/ended, then you should always be getting an "ended" event fired.

lminiero:
ended is only valid when the port for the m-line becomes zero, which means it cannot be reused for renegotiations anymore.

DaGLiMiOuX:
I would like to point to my tests into this issue, as I've been working on it for a few days after updating to 0.9.2 from 0.8.X (I cannot exactly remember the version right now).

After a couple of days working with track events in videoroom, handled by janus.js itself in `onmute` and `onunmute`, I've found that is quite not intuitive for the end-user to understand what's happening with the streaming when black screens and a "No video" tag appears in the screen and a second later starts playing the video again.

In terms of logic/programming, sounds nice for me as we can know when a user is having a problem with the connection (losing frames for bad network), but end-users don't understand it very well at all... So we were studying how to properly make it work as it was in previous versions of janus.js. That said, we changed `onmute` and `onunmute` functions properties in the remote streams as `null` when we are receiving them.

Also, I would like to point our tests with Safari. You pointed in other issues that you do not do tests (or atleast not many, correct me if I'm wrong, please!) in Safari or iOS. I work for making our ecosystem work either in Chrome or Safari, no matter the OS (which is kinda hard), but I would like to report that when an iOS device with Safari connects to a room and calls for unpublish local video, in the other side (the subscriber) we are receiving a lot of logs of `muting` and `unmuting` events constantly like the worst of spams you ever received... That made the video tag where the stream is constantly spams from a "No video" into a black screen and viceversa every second.

I think that this is more from how browsers work than how janus works. I'll try to find more information and keep you noticed about this in Safari devices or if it is something related of how unpublishing works through janus.js, as it sets the stream track enabled as `false` and couldn't test enough if this is breaks Safari.

For now, we disabled those events in our code and I would like to suggest a change into those event handlers or taking Safari in count for it, as it works really bad for unpublished videos.

I will try to share some pastebins and debugging when I get to work tomorrow, so you can better follow up the problem.

If you need any screenshot or additional information, let me know and I'll be glad to share my debug tests.

Br,
D.

PR #1975, opened 2020-03-03
Make prebuffering in AudioBridge configurable

lminiero:
By default, the AudioBridge buffers 6 (value of the `DEFAULT_PREBUFFERING` define) packets before decoding a participant, in order to smooth things out when there's a bit of jitter. Some people have expressed an interest to make this configurable, rather than edit the code, so this patch does that.

Now you can set a `default_prebuffering` property when creating a room (both statically and dynamically), and that will act as the new default for that specific room; omitting it will leave `DEFAULT_PREBUFFERING` as the default. You can force a different value for individual participants by passing a `prebuffer` property in a `join` request.

### DIFF
--- a/conf/janus.plugin.audiobridge.jcfg.sample
+++ b/conf/janus.plugin.audiobridge.jcfg.sample
@@ -9,6 +9,7 @@
 # audiolevel_event = true|false (whether to emit event to other users or not, default=false)
 # audio_active_packets = 100 (number of packets with audio level, default=100, 2 seconds)
 # audio_level_average = 25 (average value of audio level, 127=muted, 0='too loud', default=25)
+# default_prebuffering = number of packets to buffer before decoding each particiant (default=6)
 # record = true|false (whether this room should be recorded, default=false)
 # record_file = "/path/to/recording.wav" (where to save the recording)
 #
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -37,6 +37,7 @@ room-<unique room ID>: {
 	audiolevel_event = true|false (whether to emit event to other users or not, default=false)
 	audio_active_packets = 100 (number of packets with audio level, default=100, 2 seconds)
 	audio_level_average = 25 (average value of audio level, 127=muted, 0='too loud', default=25)
+	default_prebuffering = number of packets to buffer before decoding each particiant (default=DEFAULT_PREBUFFERING)
 	record = true|false (whether this room should be recorded, default=false)
 	record_file =	/path/to/recording.wav (where to save the recording)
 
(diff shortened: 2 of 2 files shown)


chpl:
I see that this value is used both for prebuffering and as max queue size. Maybe separate values should be used so it's start to transmit earlier (less latency) but still be able to accumulate more packets.

lminiero:
This value has nothing to do with max queue size, that's a Janus core thing (that is now not even there anymore, we have a `min_size_queue` property that has a different meaning).

plugins/janus_audiobridge.c
```diff
@@ -3720,22 +3742,22 @@ void janus_audiobridge_incoming_rtp(janus_plugin_session *handle, janus_plugin_r
...
+			if(g_list_length(participant->inbuf) > participant->prebuffer_count) {
 				participant->prebuffering = FALSE;
 				JANUS_LOG(LOG_VERB, "Prebuffering done! Finally adding the user to the mix\n");
 			} else {
 				JANUS_LOG(LOG_VERB, "Still prebuffering (got %d packets), not adding the user to the mix yet\n", g_list_length(participant->inbuf));
 			}
 		} else {
 			/* Make sure we're not queueing too many packets: if so, get rid of the older ones */
-			if(g_list_length(participant->inbuf) >= DEFAULT_PREBUFFERING*2) {
+			if(g_list_length(participant->inbuf) >= participant->prebuffer_count*2) {
```
chpl: Here prebuffer count limits the inbuf capacit

atoppi: You're right, that value will also limit the buffer capacity (as double the value).

It could be better to configure a total buffer length and then compute a prebuffering threshold based on the total buffer length. I find the current configuration method a bit misleading (configure the pre-buffer and compute the length as 2 * pre-buffer).

Also having a buffer length upper bound (e.g. 32, equivalent to ~600 ms of audio delay) might be a good idea (this to avoid `g_list_length` and `g_list_insert_sorted` calls on longer lists, potentially hitting the performance).

lminiero: @chpl got what you meant now :+1: 
I thought you were referring to the size of the packets we buffer in the core (which we call NACK queue), which is unrelated to this value here.

conf/janus.plugin.audiobridge.jcfg.sample
```diff
@@ -9,6 +9,7 @@
 # audiolevel_event = true|false (whether to emit event to other users or not, default=false)
 # audio_active_packets = 100 (number of packets with audio level, default=100, 2 seconds)
 # audio_level_average = 25 (average value of audio level, 127=muted, 0='too loud', default=25)
+# default_prebuffering = number of packets to buffer before decoding each particiant (default=6)
```
chpl: typo in word "partici**p**ant" here and in plugins/janus_audiobridge.c

lminiero: Thanks, I'll fix this :+1:

lminiero:
Merging.

PR #1979, opened 2020-03-05
Add G.711 support to the AudioBridge plugin

lminiero:
This was asked a few times in the past, and I had some time to work on it, so here it is. The AudioBridge now supports G.711 as well as a codec, not just Opus: this means that, whenever the sampling rate is not 8000 (te only rate G.711 supports), resampling will take place for G.711 participants to make sure their contribution can be mixed with the others', and they can receive a mix made on a higher sampling rate instead.

You can force G.711 for a participant by adding a `codec` property to the `join` request, and set it to `pcma` (a-Law) or `pcmu` (mu-Law): the default, if the property is missing, is still Opus. As a result, the SDP answer will match the provided codec instead. It's important to set the `codec` property, because if the SDP offer from a new participant _only_ contains G.711, but the `codec` property is not `pcmu` or `pcma`, the negotiation will _fail_, because the plugin will try to negotiate Opus instead.

The new G.711 support is available in the AudioBridge RTP forwarding functionality as well, and the way to configure a G.711 forwarder is via a `codec` property (same values as before) in the `rtp_forward` request. If you're creating the forwarder statically, the property to set in the configuration file is `rtp_forwarder_codec`.

Feedback welcome!

### DIFF
--- a/conf/janus.plugin.audiobridge.jcfg.sample
+++ b/conf/janus.plugin.audiobridge.jcfg.sample
@@ -21,7 +21,8 @@
 # rtp_forward_host_family = "<ipv4|ipv6; by default, first family returned by DNS request>"
 # rtp_forward_port = port to forward RTP packets of mixed audio to
 # rtp_forward_ssrc = SSRC to use to use when streaming (optional: stream_id used if missing)
-# rtp_forward_ptype = payload type to use when streaming (optional: 100 used if missing)
+# rtp_forward_codec = opus (default), pcma (A-Law) or pcmu (mu-Law)
+# rtp_forward_ptype = payload type to use when streaming (optional: only read for Opus, 100 used if missing)
 # rtp_forward_srtp_suite = length of authentication tag (32 or 80)
 # rtp_forward_srtp_crypto = "<key to use as crypto (base64 encoded key as in SDES)>"
 # rtp_forward_always_on = true|false, whether silence should be forwarded when the room is empty (optional: false used if missing)
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -49,7 +49,8 @@ room-<unique room ID>: {
 	rtp_forward_host_family = ipv4|ipv6; by default, first family returned by DNS request
 	rtp_forward_port = port to forward RTP packets of mixed audio to
 	rtp_forward_ssrc = SSRC to use to use when streaming (optional: stream_id used if missing)
-	rtp_forward_ptype = payload type to use when streaming (optional: 100 used if missing)
+	rtp_forward_codec = opus (default), pcma (A-Law) or pcmu (mu-Law)
+	rtp_forward_ptype = payload type to use when streaming (optional: only read for Opus, 100 used if missing)
 	rtp_forward_srtp_suite = length of authentication tag, if SRTP is needed (32 or 80)
 	rtp_forward_srtp_crypto = key to use as crypto, if SRTP is needed (base64 encoded key as in SDES)
 	rtp_forward_always_on = true|false, whether silence should be forwarded when the room is empty (optional: false used if missing)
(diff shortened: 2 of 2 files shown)


lminiero:
I think that one of the people asking for this a while ago was @mirkobrankovic, so all I can say is: better late than never! :smile:

mirkobrankovic:
Hey @lminiero ,
What should I set in "request":"create" audiocodec property?

lminiero:
> What should I set in "request":"create" audiocodec property?

Not "create", it's a property you set in "join", because each participant can use a different codec. And the property is called "codec", not "audiocodec".

```
{
    "request": "join",
    "room": 1234,
    [..]
    "codec": "pcmu",
    [..]
}
```

Valid values for `codec` are `opus` (default), `pcma` (G.711 a-Law) and `pcmu` (G.711 mu-Law).

mirkobrankovic:
but isn't `request: create` limiting on which codec should be used/accepted by the room sdp answers?

lminiero:
No, G.711 is always allowed at the moment. Do you see any reason why it shouldn't?

mirkobrankovic:
> No, G.711 is always allowed at the moment. Do you see any reason why it shouldn't?

no, just wondering if room create can overwrite codecs that are allowed/set on join :)

mirkobrankovic:
I have tried placing PCMA in first place but Janus audio bridge decides to use Opus:
https://pastebin.com/V638Nd6d

Do I need to leave only pcam?

lminiero:
You're not sending the right request. You're sending a `configure` with the offer, and you have to send a `join`, and the request MUST contain a `codec: "pcma"`. The order in the SDP is irrelevant, the plugin won't look at the SDP: you HAVE to say which codec you want by specifying it in the API request.

lminiero:
See this snippet from the PR description:

> You can force G.711 for a participant by adding a `codec` property to the `join` request, and set it to `pcma` (a-Law) or `pcmu` (mu-Law): the default, if the property is missing, is still Opus. As a result, the SDP answer will match the provided codec instead. It's important to set the `codec` property, because if the SDP offer from a new participant _only_ contains G.711, but the `codec` property is not `pcmu` or `pcma`, the negotiation will _fail_, because the plugin will try to negotiate Opus instead.

mirkobrankovic:
Indeed, just i forgot to add full log.
Here it is from join request (https://pastebin.com/fqKLDZK6)

I hope this one is ok

lminiero:
Ah ok, so you're first using "join" to join, but without sending the offer, and you only send an offer in a "configure" later on. That's an unusual approach, as the demo (and the proper way to do it) is to send the offer in the "join" itself: the way you do it feels more like what you'd do in the VideoRoom.

Anyway, it should work anyway, so I'll check if there's a bug somewhere.

lminiero:
Edit: nevermind, I'm an idiot :laughing: 
What you do is exactly what the demo does too, so nevermind that...

lminiero:
Are you sure you're talking to a Janus compiled on this branch? What's the commit that's reported when you send an "info" request via Janus API?

mirkobrankovic:
yeah sorry I wrote that "short" code to test video room, and then just adopted to audiobridge, didn't look how to lower amount of messages/steps

lminiero:
I just double checked modifying `audiobridgetest.js`, specifically by adding the codec when preparing the join request in the in the `registerUsername` function, like this:

```
		var register = { "request": "join", "room": myroom, "display": username };
		register["codec"] = "pcma";
```

and it works as expected for me. This makes me think your Janus instance is not actually the one build against this code.

mirkobrankovic:
img
I build .deb with checkinstall but I do see my repo is back to master, so let me compile it again and try

lminiero:
I'll bump the version of the AudioBridge plugin (from 10 to 11), so you'll be able to double check via Admin API if it is the one you need.

mirkobrankovic:
I was on master :/ sorry for that.
https://pastebin.com/54ww8Jcz
I got a a proper response from Janus now on audio bridge v11

Thanks for patience :D

lminiero:
Please let me know if it works as expected, and I'll merge (maybe adding a new tag before that).

mirkobrankovic:
> Please let me know if it works as expected, and I'll merge (maybe adding a new tag before that).

let me test transcoding, since that is most important thing i guess :D

mirkobrankovic:
I got 3 participants all with different codecs, and audio was good, so first test went fine.
good
I will try also from FreeSwitch perspective :d

lminiero:
@mirkobrankovic that sounds good enough to me, for now, thanks! I'll merge, but please do let us know if interoperability tests fail for any reason.

mirkobrankovic:
Great work by the way :D

PR #1980, opened 2020-03-06
Reduced log level to info when logger and event handlers are not found

oscarvadillog:
Minor change, just reduce the log level in these cases due to they are not fatal.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -3964,7 +3964,7 @@ gint main(int argc, char *argv[])
 	dir = opendir(path);
 	if(!dir) {
 		/* Not really fatal, we don't care and go on anyway: loggers are not fundamental */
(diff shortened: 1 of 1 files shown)


lminiero:
I'm not sure about this: we have the `LOG_FATAL` for other module folders as well. Anyway, if you don't like that, it should at the very least be `LOG_WARN`, because it's something that a person installing Janus may not expect and if it's `LOG_INFO` it will not be visible enough.

oscarvadillog:
In my opinion it's not a fatal error because janus can still working without this folder, but it's in your hands :stuck_out_tongue_winking_eye: - I've modified it to `LOG_INFO`

lminiero:
Thanks for the quick fix, merging then! :+1:

PR #1986, opened 2020-03-10
Small fixes for TypeScript declaration file

pszhang:
Fixed a typo and an issue with the typing of iceServers.

Based on the usage in the js library and the RTCPeerConnection's constructor spec (https://w3c.github.io/webrtc-pc/#dom-rtcconfiguration), it seems like it should be an array of RTCIceServer instead of string. TypeScript provides this definition (https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts)

### DIFF
--- a/npm/janus.d.ts
+++ b/npm/janus.d.ts
@@ -23,7 +23,7 @@ declare namespace JanusJS {
 		dependencies?: Dependencies;
 	}
 
-	interface ConstuctorOptions {
+	interface ConstructorOptions {
 		server: string | string[];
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! Can you sign our [CLA](https://cla.conf.meetecho.com/janus/) so that I can merge this?

pszhang:
Signed, thanks!

lminiero:
Thanks, merging then :+1:

PR #1987, opened 2020-03-10
ice: ensure that stream is non-NULL

tmatth:
This fixes a crash on later stream checks (e.g., transport_wide_cc et al).

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -4379,7 +4379,7 @@ static void janus_ice_queue_packet(janus_ice_handle *handle, janus_ice_queued_pa
 }
 
 void janus_ice_relay_rtp(janus_ice_handle *handle, janus_plugin_rtp *packet) {
-	if(!handle || handle->queued_packets == NULL || packet == NULL || packet->buffer == NULL ||
+	if(!handle || !handle->stream || handle->queued_packets == NULL || packet == NULL || packet->buffer == NULL ||
 			!janus_is_rtp(packet->buffer, packet->length))
 		return;
 	if((!packet->video && !janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_HAS_AUDIO))


tmatth:
Observed a SIGSEGV under load testing:
```
#0  0x000000000044b110 in janus_ice_relay_rtp (handle=0x7f88e9807e20, packet=0x7f89086b1680) at ice.c:4401
...
$8 = (janus_ice_handle *) 0x7f88e9807e20
(gdb) print handle->stream
$9 = (janus_ice_stream *) 0x0
```

lminiero:
Thanks for spotting (and fixing) this! I guess we'll have to do the same for `relay_rtcp` and `relay_data`. I'll take care of this after I merge :+1:

lminiero:
Looks like we already have the check in place for RTCP, while it's not used in data, so no need to do anything there.

tmatth:
@lminiero just a heads up, I think I may still be seeing the same crash so it's possible that this is being modified by another thread in between the check and the read.

lminiero:
@tmatth you may want to move the check closer to where it's used:

```
	if((packet->video && handle->stream->transport_wide_cc_ext_id > 0) || handle->stream->mid_ext_id > 0 ||
```

which is what we do for RTCP as well. Won't be perfect, but will still be better than checking (only) when we enter the function. Does this make sense?

tmatth:
> @tmatth you may want to move the check closer to where it's used:
> 
> ```
> 	if((packet->video && handle->stream->transport_wide_cc_ext_id > 0) || handle->stream->mid_ext_id > 0 ||
> ```
> 
> which is what we do for RTCP as well. Won't be perfect, but will still be better than checking (only) when we enter the function. Does this make sense?

That had occurred to me as it may reduce the likelihood of the crash, but it'd still not be threadsafe. Maybe the handle mutex should be held here?

lminiero:
I'm **very** hesitant to put locks in relay_rtp. I'll have a look at where stream is set to NULL tomorrow, to see if I can spot something.

tmatth:
> I'm **very** hesitant to put locks in relay_rtp. I'll have a look at where stream is set to NULL tomorrow, to see if I can spot something.

Yeah same...hopefully there's something easy to fix (like refcounting) to avoid this.

PR #1990, opened 2020-03-11
Fix rare race condition when claiming sessions

lminiero:
We've been notified about a rare race condition that can occur when claiming sessions, specifically when a "claim" request for a session happens at the same time as an event pushed by a plugin to a handle belonging to the session itself. The root cause is that the `source` property of `session` (which contains the info to relate a session to a transport instance) is used by different threads concurrently, and only some parts of the code are locking it properly: specifically, the code for "claim" does, while the code that notifies the event doesn't.

Looking at the code, I found other instances where `source` was accessed without a lock, so this patch makes sure they're all fixed. The only place where this doesn't happen is `janus_check_sessions`, as when `source` is accessed there a session has already timed out, meaning no conflicts are likely to take place. The patch also makes "claim" do nothing if you're trying to claim a session you already own (meaning "claim" comes from the same transport instance that is managing the session already): before we were replacing it with itself, which made little sense and made the issue easier to replicate.

I don't expect this to cause issues, as I've followed the core/transports interaction for the calls that are invoked from within the lock, and I didn't find any way transports can go back to the core and try to lock the same mutex. As such, I'm planning to merge soon, but feedback that might prove me wrong is more than welcome.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -1169,6 +1169,18 @@ int janus_process_incoming_request(janus_request *request) {
 	} else if(!strcasecmp(message_text, "claim")) {
 		janus_mutex_lock(&session->mutex);
 		if(session->source != NULL) {
+			/* If we're claiming from the same transport, ignore */
+			if(session->source->instance == request->instance) {
+				janus_mutex_unlock(&session->mutex);
+				/* Prepare JSON reply */
+				json_t *reply = json_object();
+				json_object_set_new(reply, "janus", json_string("success"));
+				json_object_set_new(reply, "session_id", json_integer(session_id));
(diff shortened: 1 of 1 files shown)


lminiero:
> The implementation of `send_message` is up to the transport, so it might be blocking for a long time, this is not known ahead of time.

Fair point: we should be safe with the existing transports, though. All transports just add to a queue, right now, with the exception of MQTT which uses MQTTAsync_sendMessage that is asynchronous anyway. We'll probably want to add something to the documentation that says `send_message` should not be blocking, for people interested in contributing a new transport.

lminiero:
Merging.

PR #1992, opened 2020-03-12
Added option to enforce validation on DTLS certificates

lminiero:
As you probably know already, all (or almost all) of the certificates exchanged via DTLS to establish PeerConnections are self-signed, meaning there's no need for validating them (which would actually failed if you tried). For those that have complete control over their client side, including on the certificates they generate and use, it might nevertheless be helpful to have a way to enforce Janus to enforce such a validation instead, in order to ensure the certificate being presented is valid.

This is exactly what this patch does. By default, we still accept self-signed certificates, so nothing changes. If you set

    dtls_accept_selfsigned = false

in the `[certs]` section of `janus.jcfg`, though, Janus will be instructed to only accept valid certificates, and you should see this line at startup:

    [WARN] WebRTC PeerConnections with self-signed certificates will NOT be accepted

Of course, needless to say **ONLY USE THIS FEATURE IF YOU KNOW WHAT YOU'RE DOING**.

As a side note, when I say that nothing changes for existing PeerConnections, that's only partly true: the DTLS stack in Janus will now also reject certificates that have expired, even when self-signed. This shouldn't be an issue when talking to browsers, but you may want to double check what your endpoints are sending, to make sure you're not using expired certificates somewhere.

Feedback welcome!

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -167,11 +167,14 @@ general: {
 # want to enforce the DTLS stack in Janus to enforce valid certificates
 # from peers, though, you can do that setting 'dtls_accept_selfsigned' to
 # 'false' below: DO NOT TOUCH THAT IF YOU DO NOT KNOW WHAT YOU'RE DOING!
+# Finally, you can configure the DTLS ciphers to offer: the default if
+# not specified is "HIGH:!aNULL:!MD5:!RC4" (what Janus used so far).
 certificates: {
 	#cert_pem = "/path/to/certificate.pem"
 	#cert_key = "/path/to/key.pem"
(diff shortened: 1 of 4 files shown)


lminiero:
Note: we may want to still enforce all checks on the validity of the certificate even for self-signed one, with the exception of `X509_V_ERR_DEPTH_ZERO_SELF_SIGNED_CERT` and `X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN`. As Tim Panton pointed out, an invalid certificate is still an invalid certificate after all...

lminiero:
Took advantage of these changes to also add a property to set the DTLS ciphers to use, and override the default ones we've been using for a while.

lminiero:
Merging.

PR #1993, opened 2020-03-12
Fixes to leaks and race conditions in VoiceMail plugin

lminiero:
We've been notified about some issues with the VoiceMail plugin, which in part didn't surprise me as it hasn't been touched much since it was first created, and we very rarely use it ourselves. Specifically, there was a race condition that could cause a crash in the server. This patch addresses that issue, and also takes care of leaks that the VoiceMail plugin was apparently causing consistently, whether the issue showed up or not.

I can't replicate the issue anymore, so I'm planning to merge soon. Feedback welcome!


lminiero:
Merging.

PR #1994, opened 2020-03-12
Several fixes to session management in VideoCall plugin

lminiero:
We've been notified about some issues in the VideoCall plugin, specifically some race conditions that could lead to crashes in Janus. After some investigations, this seems to be caused by a broken management of sessions in the plugin, that are tied to the usernames that are registered: this can lead to weird states, where some references may be freed too early or too many times, causing trouble. Checking the code, this is also because the VideoCall was possibly the only plugin that didn't benefit from some fixes we did a long time ago, due exactly to the different session management: this patch addresses that, and should make the plugin much more stable.

This seems to be working as expected for me, and we're doing some more tests in the meanwhile. Should these tests confirm the issue is fixed, I'll merge soon. Feedback welcome!

### DIFF
--- a/plugins/janus_videocall.c
+++ b/plugins/janus_videocall.c
@@ -350,6 +350,7 @@ static gboolean notify_events = TRUE;
 static janus_callbacks *gateway = NULL;
 static GThread *handler_thread;
 static void *janus_videocall_handler(void *data);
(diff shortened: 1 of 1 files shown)


atoppi:
I have been stress-testing this in the last hour and still no crashes
definitely lgtm :+1:

lminiero:
Merging then :+1:

PR #1995, opened 2020-03-12
update dtls ciphers

fippo:
following https://github.com/versatica/mediasoup/issues/367
Been using the same set for years without interop problems

Thanks @ibc for the poke

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -105,7 +105,7 @@ gboolean janus_is_dtls(char *buf) {
 }
 
 /* DTLS stuff */
-#define DTLS_CIPHERS	"HIGH:!aNULL:!MD5:!RC4"
+#define DTLS_CIPHERS	"DEFAULT:!NULL:!aNULL:!SHA256:!SHA384:!aECDH:!AESGCM+AES256:!aPSK"
 /* Duration for the self-generated certs: 1 year */
 #define DTLS_AUTOCERT_DURATION	60*60*24*365
 


lminiero:
Thanks @fippo! I've actually made ciphers configurable in #1992, though, and I think the (commented) value in the config is that one indeed (the one Chrome uses I think?). If you say that works fine, I'll make it the default there then :+1:

ibc:
"I don't know what the f*** those ciphers mean so I let the users choose it"

fippo:
@ibc will be happy to hear there are more people using those cool ciphers.
Lets burn some trash :-)

lminiero:
Merging, I'll align the other PR accordingly :+1:

ibc:
AMAZING!

lminiero:
> AMAZING!

:stuck_out_tongue_closed_eyes:

ibc:
This PR is amazing. Now I have a second problem. Can you create simple code snippets? You must do it because I don't want to learn Janus in depth but just use it and make money with it.

lminiero:
Of cour :heavy_dollar_sign:  e!

ibc:
BUT this would be good for you because the open source.

lminiero:
Sorry, I'm busy!

ibc:
WHY? YOU ARE AT HOME YOU CANNOT LEAVE

fippo:
lets call the :policeman: -- guardia di finanza would be appropriate since this is about :heavy_dollar_sign: :heavy_dollar_sign:

PR #1997, opened 2020-03-13
Implement ECDSA Certificate generation

### LINKED ISSUE #1978: Generate ECDSA Private Key (instead of RSA) in janus_dtls_generate_keys
I was testing some stuff against Janus with Pion and noticed the cert size differences when looking at DTLS handshakes (RSA being larger). Would you be interested in a PR that adds ECDSA generation? I put the pseudo code below, but will be a very tiny diff.  Would mean smaller DTLS handshakes and faster cert generation!

I am happy to put it behind a configuration option etc... however you prefer the change.

```
        if ((eccGroup = OBJ_txt2nid("prime256v1") == NID_undef) {
          // error
        }

        if ((eccKey = EC_KEY_new_by_curve_name(eccGroup)) == NULL) {
          // error
        }

        EC_KEY_set_asn1_flag(eccKey, OPENSSL_EC_NAMED_CURVE);

        if (EC_KEY_generate_key(eccKey) == 0) {
          // error
        }
        if (EVP_PKEY_assign_EC_KEY(*ppPkey, eccKey) == 0) {
          // error
        }
```

Sean-Der:
By default generate certificates with NIST P-256. Allow
RSA to still be used by with dtls_generate_rsa_private_key
config option.

Resolves #1978

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -195,6 +195,7 @@ media: {
 	#slowlink_threshold = 4
 	#twcc_period = 100
 	#dtls_timeout = 500
+	#dtls_generate_rsa_private_key = false
 }
 
 # NAT-related stuff: specifically, you can configure the STUN/TURN
--- a/dtls.c
+++ b/dtls.c
@@ -108,6 +108,8 @@ gboolean janus_is_dtls(char *buf) {
 #define DTLS_CIPHERS	"DEFAULT:!NULL:!aNULL:!SHA256:!SHA384:!aECDH:!AESGCM+AES256:!aPSK"
 /* Duration for the self-generated certs: 1 year */
 #define DTLS_AUTOCERT_DURATION	60*60*24*365
+/* NIST P-256 elliptic curve used for private key generation */
+#define DTLS_ELLIPTIC_CURVE NID_X9_62_prime256v1
 
 /* DTLS timeout base to enforce: notice that this can currently only be
  * modified if you're using BoringSSL, as OpenSSL uses 1s (1000ms) and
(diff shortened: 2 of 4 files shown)


lminiero:
Thanks @Sean-Der, you rock! I'll make a couple of tests, and merge in case I don't see issues :+1:

lminiero:
Looks good! I noticed a couple of things I'd like tweaks on (e.g., name of the property and the section it is in), but nothing that needs to wait, so I'll take care of those after I merge.

Thanks again!

fippo:
this shouldn't take more than `SSL_CTX_set_ecdh_auto(somectx, 1);`
(if SSL_CTX_set_ecdh_auto is supported. other wise you need a bit work)

paging @ibc for cross-pollination

lminiero:
@fippo this patch is for _generating_ ECDSA certificates, though. Is that what `SSL_CTX_set_ecdh_auto` is for?

ibc:
We already do [SSL_CTX_set_ecdh_auto()](https://github.com/versatica/mediasoup/blob/v3/worker/src/RTC/DtlsTransport.cpp?ts=2#L466) in mediasoup (whatever it means).

ibc:
PLEASE STAY AT HOME

fippo:
ah... it does take more  than SSL_CTX_set_ecdh_auto (which is a nice things to do so i'll provide a patch shortly) so what this does looks good.
Lucky me generating the certs with a single line of JS before passing them on :-)

ibc:
Mmm, I don't remember the rationale about my code, but it does this:

```c++
// Enable ECDH ciphers.
// DOC: http://en.wikibooks.org/wiki/OpenSSL/Diffie-Hellman_parameters
// NOTE: https://code.google.com/p/chromium/issues/detail?id=406458
// NOTE: https://bugs.ruby-lang.org/issues/12324
//
// Nothing to be done in OpenSSL >= 1.1.0.
#if (OPENSSL_VERSION_NUMBER >= 0x10100000L)
// For OpenSSL >= 1.0.2.
#elif (OPENSSL_VERSION_NUMBER >= 0x10002000L)
		SSL_CTX_set_ecdh_auto(DtlsTransport::sslCtx, 1);
// Older versions.
#else
		ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1);

		if (!ecdh)
		{
			LOG_OPENSSL_ERROR("EC_KEY_new_by_curve_name() failed");

			goto error;
		}

		if (SSL_CTX_set_tmp_ecdh(DtlsTransport::sslCtx, ecdh) != 1)
		{
			LOG_OPENSSL_ERROR("SSL_CTX_set_tmp_ecdh() failed");

			goto error;
		}

		EC_KEY_free(ecdh);
		ecdh = nullptr;
#endif
```

so, for OpenSSL >= 1.0.2 the only that must be done is:

```c++
SSL_CTX_set_ecdh_auto(sslCtx, 1);
```

WHAT'S WRONG WITH THIS?

lminiero:
We're talking about different things: we've had the same code to enable ECDHE ciphers as well for quite some time already. This patch here is used to generate the certificate Janus itself will use in DTLS: you either specify an existing certificate in the configuration file, or ask Janus to generate one at startup, and this is what this patch affects. Before our certificates were generated using RSA, thanks to a contribution by @saghul, while @Sean-Der now extended it to support generating ECDSA certificates too.

ibc:
Ah, I see, thanks. Updating my code too.

ibc:
So, any reason to keep the RSA code other than allowing older OpenSSL versions?

lminiero:
It's there mostly as a fallback, because it's what we've used for a long time. Since this is a new feature and I'm not 100% sure it works with all OpenSSL versions, I thought it might be safer to keep it there.

ibc:
Careful [here](https://github.com/meetecho/janus-gateway/pull/1997/files#diff-087c8af1d9f50a7df2ea8b32698689a3R298). `BN_free(bne);` is always called, even if `bne = BN_new();` was not called above.

fippo:
i've had left a papertrail in my own changes. I'm a bureaucrat.
This was added to webrtc in https://bugs.chromium.org/p/webrtc/issues/detail?id=5795 in 2016 and i've been running in production with just this for months without (reported) issues.

ibc:
So I've made the same changes to mediasoup by making it always generate certs with EC keys instead of RSA:

https://github.com/versatica/mediasoup/commit/cc0dccc93ed45b257472c4873e19b5c0d1ab99ac?ts=2

It seems to work and blablabla. Thanks, author!

PR #1998, opened 2020-03-16
Added option to ignore mDNS candidates

lminiero:
You may be familiar with [mDNS candidates](https://tools.ietf.org/html/draft-ietf-rtcweb-mdns-ice-candidates-04), which we've supported in Janus for a while. We need to resolve them ourselves before passing them to libnice, though, and we currently do that using `getaddrinfo`, which is a blocking method that only returns when the resolution has been either a success or a failure. This usually happens very quickly, but we've seen instances where it can take many seconds for a request to timeout, e.g., because services like Avahi are turned off on the machine: since this happens while the handle mutex is locked, and on the API messages thread, this can cause Janus being unresponsive in the meanwhile.

The first step to address this is this patch, that adds a new property in `janus.jcfg` by which you can tell Janus to ignore mDNS candidates. Of course this should only be used when you know for a fact you won't have any use for mDNS candidates (e.g., all users will always be remote, and not in the Janus network), but that can still help in case this has been an issue for you.

In the next few weeks we'll also work on making the resolution more asynchronous, but in the meanwhile, I'm planning to merge quite soon so feedback welcome!

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -231,6 +231,14 @@ nat: {
 	#ice_lite = true
 	#ice_tcp = true
 
+	# By default Janus tries to resolve mDNS (.local) candidates: since
+	# this is currently done synchronously and might keep the API busy,
+	# especially in case mDNS resolution takes a long time to timeout,
+	# you can choose to drop all .local candidates instead, which is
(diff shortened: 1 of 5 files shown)


lminiero:
Merging.

PR #1999, opened 2020-03-16
Support for RTSP 'Content-Base' header in Streaming plugin

lminiero:
As the title says. When the response to an RTSP DESCRIBE request contains a `Content-Base` header, the related value is what should be used in the following SETUP: since we didn't support it previously, our SETUP would be addressed to the wrong address, leading to an error. This patch tries to address that by parsing (and using) the value accordingly.

I don't have a server using that property to test against, so please let me know if it works as expected, and that this doesn't introduce regressions for sessions that did work before.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -5654,7 +5654,7 @@ static int janus_streaming_rtsp_parse_sdp(const char *buffer, const char *name, 
 		cb = strstr(buffer, "content-base:");
 	if(cb != NULL) {
 		cb = strstr(cb, "rtsp://");
-		const char *crlf = cb ? strstr(cb, "\r\n") : NULL;
+		const char *crlf = (cb ? strstr(cb, "\r\n") : NULL);
 		if(crlf != NULL && base != NULL) {
 			gulong size = (crlf-cb)+1;
 			if(size > 256)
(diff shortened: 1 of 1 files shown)


plugins/janus_streaming.c
```diff
@@ -5646,8 +5646,25 @@ static size_t janus_streaming_rtsp_curl_callback(void *payload, size_t size, siz
...
-static int janus_streaming_rtsp_parse_sdp(const char *buffer, const char *name, const char *media, int *pt,
+static int janus_streaming_rtsp_parse_sdp(const char *buffer, const char *name, const char *media, char *base, int *pt,
 		char *transport, char *host, char *rtpmap, char *fmtp, char *control, const janus_network_address *iface, multiple_fds *fds) {
+	/* Start by checking if there's any Content-Base header we should be aware of */
+	const char *cb = strstr(buffer, "Content-Base:");
+	if(cb == NULL)
+		cb = strstr(buffer, "content-base:");
+	if(cb != NULL) {
+		cb = strstr(cb, "rtsp://");
+		const char *crlf = strstr(cb, "\r\n");
```
tmatth: Is cb guaranteed to not be NULL here (i.e. are you sure that cb contains `rtsp://` ?)?

lminiero: Good catch, I'll fix it :+1:

lminiero:
Merging.

PR #2004, opened 2020-03-18
Resolve mDNS candidates asynchronously with GResolver (see #1998)

lminiero:
I've explained in #1998 how resolving mDNS candidates is done synchronously at the moment (using `getaddrinfo` that is a blocking request), and how this can cause the API to be unresponsive when that resolution takes a long time to timeout. That PR added a way to ignore mDNS candidates entirely, but was not an actual solution (especially in cases where you DO need to be able to resolve them).

This patch tries to address that, using [GResolver](https://developer.gnome.org/gio/stable/GResolver.html) to resolve mDNS candidates asynchronously: when a `.local` address is detected, and mDNS candidates are not ignored (default), we ask for an asynchronous resolution of the address and then return a success right away, thus keeping the API lean. When the address is resolved, we update the candidate we received originally and parse it as usual; if the handle went away in the meanwhile, we just drop it.

This adds a new dependency to `gio`, but to my knowledge it is always automatically installed when you install `glib`, so you should actually not need to do anything at all to use it (the installations instructions haven't been updated). I did make a small change in the core, though: since we needed a main loop to get `GResolver` to work, I got rid of the ugly `usleep(250000)` we had before, and turned that into a `GMainLoop` we run indefinitely there instead. This still works, because when we hit CTRL+C we just call `g_main_loop_quit()` and leave as expected.

I tested this briefly and it seems to be working as expected, but due to the asynchronous nature of the functionality, and the fact it may try to use stuff that has gone (e.g., answer to resolution arrived after the related handle was detached), I'm interested to figure out if it can cause issue, so feedback (and tests) welcome!

### DIFF
--- a/sdp.c
+++ b/sdp.c
@@ -18,6 +18,8 @@
 
 #include <netdb.h>
 
+#include <gio/gio.h>
+
 #include "janus.h"
 #include "ice.h"
 #include "sdp.h"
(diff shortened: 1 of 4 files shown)


sdp.c
```diff
@@ -672,22 +719,19 @@ int janus_sdp_parse_candidate(void *ice_stream, const char *candidate, int trick
...
-			g_strlcpy(rip, janus_network_address_string_from_buffer(&addr_buf), sizeof(rip));
+			/* We'll resolve this address asynchronously, in order not to keep this thread busy */
+			JANUS_LOG(LOG_VERB, "[%"SCNu64"] Resolving mDNS address (%s) asynchronously\n",
+				handle->handle_id, rip);
+			janus_sdp_mdns_candidate *mc = g_malloc(sizeof(janus_sdp_mdns_candidate));
+			janus_refcount_increase(&handle->ref);
+			mc->handle = handle;
+			mc->candidate = g_strdup(candidate);
+			mc->local = g_strdup(rip);
+			mc->cancellable = g_cancellable_new();
```
atoppi: 1) No need to handle references to `cancellable` when the task has ended?
2) Are we leaving to future enhancements the capability to cancel an async task instead of e.g. waiting for it to complete before refcounting the handle to 0 ? In this case we could set the `mc` field to `NULL` for now.

lminiero:
Merging.

PR #2009, opened 2020-03-20
Add repos for openSUSE and SUSE

ancorgs:
We have maintained repositories with janus-gateway packages for quite some time (likely, the first Janus packages been available for any distro ever), but I have just realized we never submitted them to the list in the Janus webpage.

Let's fix that. :wink:


lminiero:
@ancorgs thanks! I see some sort of unicode error in your name, though? (Gonz�lez)

ancorgs:
In fact, Saúl name has a similar problem and it looks ok in the final page. Anyways, since I'm not sure about what encoding is used I just removed the tilde (my name will be equally understandable).

lminiero:
Actually Saul's name looked fine in my github preview... thanks for fixing it though, merging!

lminiero:
Updated docs online too :+1: https://janus.conf.meetecho.com/docs/resources

PR #2013, opened 2020-03-23
Update media direction in SIP plugin if remote address is 0.0.0.0

lminiero:
Pretty short patch, that is aimed at fixing an issue we sometimes apparently encounter when putting calls on hold. When we put a call on hold, we set the media direction to `sendonly`, as we don't want to receive anything until the call is resumed: apparently, though, this can sometimes cause issues when the remote party answers with a `0.0.0.0` c-address, that might result in the inability to send packets, ICMP errors, and as a result the call being torn down after a couple of seconds.

This patch adds a further check on the media address when (re)connecting sockets, updating the media direction to false for audio and/or video if their remote address is, indeed, `0.0.0.0`.

Feedback welcome!


lminiero:
Looks like this fixes the issue, so I'll merge.

PR #2014, opened 2020-03-23
Track pending nack cleanup tasks and cancel them when freeing a stream.

atoppi:
This PR aims at fixing (again) #1453 .

This time we are tracking the pending cleanup tasks in a map. When destroying a stream the map is traversed and all pending tasks get cancelled.
I've made a couple of tests and everything seems to work (e.g. no leaks, no crashes) however we'll definitely need more feedback before merging and marking this as solved.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -324,6 +324,7 @@ typedef struct janus_ice_nacked_packet {
 	janus_ice_handle *handle;
 	int vindex;
 	guint16 seq_number;
+	guint source_id;
 } janus_ice_nacked_packet;
 static gboolean janus_ice_nacked_packet_cleanup(gpointer user_data) {
 	janus_ice_nacked_packet *pkt = (janus_ice_nacked_packet *)user_data;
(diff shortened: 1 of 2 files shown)


ice.c
```diff
@@ -2642,9 +2656,12 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 								np->handle = handle;
 								np->seq_number = cur_seq->seq;
 								np->vindex = vindex;
+								if(stream->pending_nacked_cleanup == NULL)
+									stream->pending_nacked_cleanup = g_hash_table_new(NULL, NULL);
 								GSource *timeout_source = g_timeout_source_new_seconds(5);
 								g_source_set_callback(timeout_source, janus_ice_nacked_packet_cleanup, np, (GDestroyNotify)g_free);
-								g_source_attach(timeout_source, handle->mainctx);
+								np->source_id = g_source_attach(timeout_source, handle->mainctx);
```
tmatth: The one caveat here is that IIUC `source_id` is not guaranteed to be unique here:
```
More specifically: source IDs can be reissued after a source has been destroyed and therefore it is never valid to use this function with a source ID which may have already been removed. An example is when scheduling an idle to run in another thread with g_idle_add(): the idle may already have run and been removed by the time this function is called on its (now invalid) source ID. This source ID may have been reissued, leading to the operation being performed against the wrong source.
``` 
(see https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#g-main-context-find-source-by-id)

Although the odds of that happening are probably fairly low.

atoppi: I can't imagine any execution that would lead to this.
Source ids are guaranteed unique among the sources of the same main context (and they can be reused like the docs says, but only after a destroy).
Anyway in our implementation a source could be either triggered (and so be gracefully removed from the map) after 5 seconds or cancelled by a stream destroy (never having a chance to take any action).
What kind of critical scenario are you thinking about ?

tmatth: > Source ids are guaranteed unique among the sources of the same main context 
Oh I missed that guarantee, in that case this is fine.

atoppi:
```c
  /* The counter may have wrapped, so we must ensure that we do not
   * reuse the source id of an existing source.
   */
  do
    id = context->next_id++;
  while (id == 0 || g_hash_table_contains (context->sources, GUINT_TO_POINTER (id)));
```

tmatth:
Can no longer reproduce the crasher with the fix, so LGTM!

lminiero:
Merging then :+1:

PR #1996, opened 2020-03-13
Fix deadlock when feed parameter is missing

apietrzak:
When feed parameter is not set videoroom plugin hangs up.

PR #2026, opened 2020-03-26
Updates to mutex unlocking in textroom and videoroom plugins

agclark81:
We've found a few locations in the textroom and videoroom plugins that benefit from ensuring the rooms mutex is unlocked before locking a specific room mutex, as well as a couple of places in videoroom where an unlock was necessary and one place where a refcount_decrease was needed. It appears that one of the changes we made is also in PR #1996.

### DIFF
--- a/plugins/janus_textroom.c
+++ b/plugins/janus_textroom.c
@@ -1453,8 +1453,8 @@ janus_plugin_result *janus_textroom_handle_incoming_request(janus_plugin_session
 			g_snprintf(error_cause, 512, "No such room (%s)", room_id_str);
 			goto msg_response;
 		}
-		janus_mutex_unlock(&rooms_mutex);
 		janus_refcount_increase(&textroom->ref);
+		janus_mutex_unlock(&rooms_mutex);
 		janus_mutex_lock(&textroom->mutex);
 		janus_textroom_participant *participant = g_hash_table_lookup(session->rooms,
 			string_ids ? (gpointer)room_id_str : (gpointer)&room_id);
@@ -1634,8 +1634,8 @@ janus_plugin_result *janus_textroom_handle_incoming_request(janus_plugin_session
 			g_snprintf(error_cause, 512, "No such room (%s)", room_id_str);
 			goto msg_response;
 		}
-		janus_mutex_unlock(&rooms_mutex);
 		janus_refcount_increase(&textroom->ref);
+		janus_mutex_unlock(&rooms_mutex);
 		janus_mutex_lock(&textroom->mutex);
 		/* A PIN may be required for this action */
 		JANUS_CHECK_SECRET(textroom->room_pin, root, "pin", error_code, error_cause,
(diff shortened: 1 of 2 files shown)


plugins/janus_textroom.c
```diff
@@ -1453,9 +1453,9 @@ janus_plugin_result *janus_textroom_handle_incoming_request(janus_plugin_session
 			g_snprintf(error_cause, 512, "No such room (%s)", room_id_str);
 			goto msg_response;
 		}
+		janus_mutex_unlock(&rooms_mutex);
```
lminiero: Better keep the `janus_refcount_increase` call within the lock.

plugins/janus_textroom.c
```diff
@@ -1634,9 +1634,9 @@ janus_plugin_result *janus_textroom_handle_incoming_request(janus_plugin_session
...
```
lminiero: See above.

plugins/janus_textroom.c
```diff
@@ -1771,9 +1771,9 @@ janus_plugin_result *janus_textroom_handle_incoming_request(janus_plugin_session
...
```
lminiero: See above.

plugins/janus_textroom.c
```diff
@@ -2190,9 +2190,9 @@ janus_plugin_result *janus_textroom_handle_incoming_request(janus_plugin_session
...
```
lminiero: See above.

plugins/janus_videoroom.c
```diff
@@ -4127,9 +4127,9 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 			janus_mutex_unlock(&rooms_mutex);
 			goto prepare_response;
 		}
+		janus_mutex_unlock(&rooms_mutex);
```
lminiero: See above (I should just have said "do it everywhere" :smile: )

plugins/janus_videoroom.c
```diff
@@ -5354,9 +5354,9 @@ static void *janus_videoroom_handler(void *data) {
 				janus_mutex_unlock(&rooms_mutex);
 				goto error;
 			}
+			janus_mutex_unlock(&rooms_mutex);
```
lminiero: As above.

plugins/janus_videoroom.c
```diff
@@ -5379,7 +5379,11 @@ static void *janus_videoroom_handler(void *data) {
 						JANUS_VIDEOROOM_ERROR_MISSING_ELEMENT, JANUS_VIDEOROOM_ERROR_INVALID_ELEMENT);
 				}
 				if(error_code != 0)
+				{
```
lminiero: Code style: please put the open bracket on the same line as the `if`.

plugins/janus_videoroom.c
```diff
@@ -5640,7 +5644,10 @@ static void *janus_videoroom_handler(void *data) {
...
```
lminiero: Same code style nit.

plugins/janus_textroom.c
agclark81: I've updated the PR to take into consideration all of your comments.

lminiero:
Thanks for the quick fixes! Looks good to me, merging then :+1:

Issue #2038, opened 2020-04-01
Crash in Streaming Plugin when using threads > 0


BellesoftConsulting:
Hi there, I am reporting a crash in my Beloved Janus server. the crash only happens for me when using threads > 0 on a mount-point on the streaming plugin.
I am using libnice master, and the following:

```
Debian Buster. kernel: 4.19.0-6-amd64 #1 SMP Debian 4.19.67-2+deb10u2 (2019-11-11) x86_64 GNU/Linux
```

This server is running master janus , at commit: f3e1c1e4555c8b963b697d82c46f9e959317963d
Thanks!

Backtrace:

```
`[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/opt/janus/bin/janus -o'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fe73e917e65 in g_mutex_lock () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
[Current thread is 1 (Thread 0x7fe7377fe700 (LWP 22651))]
(gdb) bt
#0  0x00007fe73e917e65 in g_mutex_lock () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007fe73c48a9b7 in janus_streaming_handler (data=<optimized out>) at plugins/janus_streaming.c:4759
#2  0x00007fe73e8f6415 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007fe73e19efa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#4  0x00007fe73e0cf4cf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) q`
```

lminiero:
Edited your post as the hashtag+number were being turned into links to other issues. Looks like it's dying on

```
janus_mutex_lock(&helper->mutex);
```

which may mean the helper was destroyed, which should only happen on a `janus_streaming_mountpoint_destroy` though, so it may be a race condition with viewers trying to subscribe to a mountpoint while you're destroying it (does this ring a bell?). If so, adding reference counters to helpers might help.

If you can replicate the issue, please [compile Janus with libasan support](https://janus.conf.meetecho.com/docs/debug) and provide the output it prints after the crash.

BellesoftConsulting:
Hi Lorenzo,
[...]

PR #2039, opened 2020-04-01
Use refcount for Streaming plugin helper threads

lminiero:
This is supposed to help with issue #2038, which looks like a crash trying to access a helper that has been destroyed already. Since we were not using reference counters for `janus_streaming_helper`, we now added it, which should help ensure the issue doesn't happen again.

Not tested extensively, so feedback welcome.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -1108,7 +1108,20 @@ typedef struct janus_streaming_helper {
 	GAsyncQueue *queued_packets;
 	volatile gint destroyed;
 	janus_mutex mutex;
+	janus_refcount ref;
 } janus_streaming_helper;
+static void janus_streaming_helper_destroy(janus_streaming_helper *helper) {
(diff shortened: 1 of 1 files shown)


BellesoftConsulting:
testing this now. thanks!

lminiero:
@BellesoftConsulting any update?

BellesoftConsulting:
> @BellesoftConsulting any update?

I put this server back in production, but it has been picked up, I'll force traffic to it to day and report back! sorry for the delay

lminiero:
Ping :slightly_smiling_face:

BellesoftConsulting:
Hi Lorenzo. So far so good. But could not get nearly as many folks into the server as I would have hoped.
:). so up to you.   We can close this, and if there is a new crash I'll follow your instructions above to get you more data.

lminiero:
It's still an improvement to the code, so I'll merge. In case you get crashes again we can take it from there.

PR #2040, opened 2020-04-01
Added support for static Opus files to Streaming plugin

lminiero:
This patch adds support for static Opus files to the Streaming plugin, meaning the plugin can now read Opus files and stream them via RTP, pretty much as we've only supported so far for raw a-Law and mu-Law recordings. This support is available both in pseudo-live mountpoints (when you create a mountpoint, the plugin loops on the specificed file, and when people attach they start listening from where the playout thread is), and in on-demand mountpoints (each person attaching to the mountpoint starts listening to the Opus stream from the beginning, and that loops until the PeerConnection is closed). This feature is only available if `libogg` and its headers are installed (which is already a requirement for the VoiceMail plugin and the post-processor, so you probably already have it).

In the future, we may want to extend this functionality to the AudioBridge as well, e.g., to play pre-recorded announcements, but for the moment in this PR I'd rather keep this to the Streaming and ensure it works as expected. As part of the process, to test stereo Opus files I also modified the Streaming plugin demo so that it negotiates stereo support if stereo is offered (the `stereo=1` property must be both in the SDP offer and answer audio `fmtp` attribute, otherwise the stream falls back to mono).

Before merging, I'll probably also add a couple of Opus streams that can be used with some sample mountpoints, but in the meanwhile it should be easy enough for you to find existing Opus files and test how they work. Here's an example of an on-demand (each user gets their own playout) Opus mountpoint:

```
opus-ondemand-sample: {
	type = "ondemand"
	id = 4
	description = "Opus file source (on-demand)"
	filename = "/path/to/file.opus"
	audio = true
	audiopt = 100
	audiortpmap = "opus/48000/2"
	audiofmtp = "stereo=1"
	video = false
	secret = "adminpwd"
}
```

This is what a pseudolive stream looks like instead:

```
opus-live-sample: {
        type = "live"
        id = 5
        description = "Opus file source (live)"
	filename = "/path/to/file.opus"
        audio = true
        audiopt = 100
        audiortpmap = "opus/48000/2"
	audiofmtp = "stereo=1"
        video = false
        secret = "adminpwd"
}
```

Of course, you won't need the `audiofmtp` property if the file is not stereo.

Feedback welcome.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -460,9 +460,9 @@ endif
 if ENABLE_PLUGIN_STREAMING
 plugin_LTLIBRARIES += plugins/libjanus_streaming.la
 plugins_libjanus_streaming_la_SOURCES = plugins/janus_streaming.c
-plugins_libjanus_streaming_la_CFLAGS = $(plugins_cflags) $(LIBCURL_CFLAGS)
-plugins_libjanus_streaming_la_LDFLAGS = $(plugins_ldflags) $(LIBCURL_LDFLAGS) $(LIBCURL_LIBS)
-plugins_libjanus_streaming_la_LIBADD = $(plugins_libadd) $(LIBCURL_LIBADD)
+plugins_libjanus_streaming_la_CFLAGS = $(plugins_cflags) $(LIBCURL_CFLAGS) $(OGG_CFLAGS)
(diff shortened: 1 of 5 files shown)


lminiero:
I won't add Opus samples to the repo, as Opus support is optional (and only enabled if libogg is available). I'll just modify the demos online instead. Merging.

PR #2043, opened 2020-04-02
Support for additional constraints on screenshare media

agclark81:
Calls to createOffer for screensharing media ({ media: { video: 'screen', screenshareFrameRate: 15 } }) does not pass in the frame rate constraint to getDisplayMedia that's now supported across most modern browsers. This PR adds support for this existing constraint, as well as introducing new screenshareHeight and screenshareWidth constraints to set a maximum height/width, which is particularly helpful to limit high-resolution displays from sending 4K+ images.

At some point, it might be worth considering removing some of the fallback options for screensharing, as those are not likely to be used now that getDisplayMedia is quite prevalent.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -2182,7 +2182,11 @@ function Janus(gatewayCallbacks) {
 						// The new experimental getDisplayMedia API is available, let's use that
 						// https://groups.google.com/forum/#!topic/discuss-webrtc/Uf0SrR4uxzk
 						// https://webrtchacks.com/chrome-screensharing-getdisplaymedia/
-						navigator.mediaDevices.getDisplayMedia({ video: true, audio: media.captureDesktopAudio })
+						constraints.video = { frameRate: media.screenshareFrameRate };
+						if (media.screenshareHeight) constraints.video.height = media.screenshareHeight;
+						if (media.screenshareWidth) constraints.video.width = media.screenshareWidth;
+						constraints.audio = media.captureDesktopAudio;
+						navigator.mediaDevices.getDisplayMedia(constraints)
 							.then(function(stream) {
 								pluginHandle.consentDialog(false);
 								if(isAudioSendEnabled(media) && !media.keepAudio) {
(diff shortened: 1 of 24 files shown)


lminiero:
Thanks, I'll review later. On a related note, are you aware if the system audio can indeed be captured using `getDisplayMedia`? In my case it always gets the microphone anyway, but I'm on Linux; not sure if it's (as usual) a Windows only thing.

arpu:
@lminiero  on linux you need to enable pipewire in the chrome://flags to get this working

lminiero:
@arpu ah I didn't know, I'll give it a try thanks!

lminiero:
> ( i use it with janus on https://vrland.io/dome screensharing and it works fantastic)

I saw your tweet, I'll have to learn what you guys are doing there sooner or later :slightly_smiling_face:

html/janus.js
```diff
@@ -2182,7 +2182,11 @@ function Janus(gatewayCallbacks) {
 						// The new experimental getDisplayMedia API is available, let's use that
 						// https://groups.google.com/forum/#!topic/discuss-webrtc/Uf0SrR4uxzk
 						// https://webrtchacks.com/chrome-screensharing-getdisplaymedia/
-						navigator.mediaDevices.getDisplayMedia({ video: true, audio: media.captureDesktopAudio })
+						constraints.video = { frameRate: media.screenshareFrameRate };
```
lminiero, line 2185: We should change the default `media.screenshareFrameRate = 3;` we have a few lines above, or maybe get rid of it entirely, before this can be merged: at the moment, since we ignore that property the video is reasonably fluid, but enforcing that default 3fps will make the experience quite crappy.

html/janus.js
```diff
@@ -2182,7 +2182,11 @@ function Janus(gatewayCallbacks) {
...
+						if (media.screenshareHeight) constraints.video.height = media.screenshareHeight;
+						if (media.screenshareWidth) constraints.video.width = media.screenshareWidth;
```
lminiero, line 2187: Code style: please put condition and action on different lines (and without any space between `if` and bracket).

html/janus.js
lminiero: I'd say let's just get rid of it, and let the browser do whatever it wants by default.

lminiero:
@agclark81 just a ping in case you missed my notes. Thanks!

agclark81:
I'll be committing the updated code momentarily.

lminiero:
Thanks, merging :+1:

PR #2044, opened 2020-04-02
refactoring - clean up

alienpavlov:
- replaced some **const** declarations with **var** to keep it ES5 style
- removed unnecessary semicolons and added recommended ones
- replaced some == with === to keep it more strict
- removed second arg from **new RTCPeerConnection()** documentation says it expect only one optional configuration argument
- added {} for some single lined if-statements for easier read
- replaced several "return true/false" lines with simple version (on my opinion)
- renamed few variables because they had duplicate declaration in the same namespace

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -130,7 +130,7 @@ Janus.useDefaultDependencies = function (deps) {
 						return reject({message: 'Request timed out', timeout: options.timeout});
 					}, options.timeout);
 				});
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks, I'll review later.

lminiero:
Looks fine to me, but I'd like to wait for #2043 before this is merged.

lminiero:
Merging, thanks :+1:

Issue #2045, opened 2020-04-03
Crash on ICE renegotiation in race condition


agclark81:
We've observed occasional crashes that occur on ICE renegotiation in the current master when, possibly, the connection and resources are simultaneously being cleaned up. The crash is occurring at 844/845 in sdp.c:

https://github.com/meetecho/janus-gateway/blob/d324ac421373d6e9d12f7428ef241e7954774786/sdp.c#L844

AddressSanitizer output is here: https://pastebin.com/NjebVMsH

Your insights would be much appreciated.

agclark81:
Just for clarification, we're seeing this occurring specifically with ICE restarts.

agclark81:
One more bit of context, in case it's helpful.

We're initiating an ICE restart for the publisher when we see an ICE state of "failed" or "disconnected", calling the following:

```
    room.createOffer({
      iceRestart: true,
      media: { audioRecv: false, videoRecv: false, audioSend: false, videoSend: false },
      success: function(jsep)
      {
        room.send({
          'message': { 'request': 'configure' },
          'jsep': jsep,
```

For subscribers, if we get a disconnected/failed state on their ICE, we call:

```
              remoteFeed.send({
                message: { request: 'configure', restart: true },
```

Do you see any issues with this approach that might be related?

tmatth:
> 
> Do you see any issues with this approach that might be related?

Even if there are, we should be avoiding this use-after-free crasher here.

lminiero:
The asan dump doesn't say where the free happened (no line number), you may want to launch Janus with the `ASAN_OPTIONS=fast_unwind_on_malloc=0` env variable set to have that info added.

lminiero:
Now that I think about it, not sure why we're setting the credentials on the `NiceCandidate` when we add it in `janus_sdp_parse_candidate`: libnice should already have the credentials from `nice_agent_set_remote_credentials`, so it shouldn't be needed. I'll check if we can just get rid of that part.

lminiero:
I got rid of those in the PR above. Since they were the cause of the crash, it may fix the race condition, but please check and let me know (especially that it doesn't introduce regressions).

agclark81:
We just picked up an attempted double-free on master with the PR #2046 applied. It doesn't seem to be related to that change, unless we're just seeing this now because the earlier crash is being avoided. https://pastebin.com/SyV51nFd

lminiero:
[...]

PR #2046, opened 2020-04-03
Don't set ICE credentials when parsing remote credentials

lminiero:
Came up reviewing #2045: apparently, we're setting the credentials on the `NiceCandidate` instance we create when we call `janus_sdp_parse_candidate`, which shouldn't be needed since libnice already has access to them from a previous call to `nice_agent_set_remote_credentials`.

It seems to still work as expected for me, but even if it's a tiny change I prefer keeping it a PR for now, so that you can make sure everything works for you too.


agclark81:
Super, we're running in production now and will report back any observations.

lminiero:
Merging.

PR #2048, opened 2020-04-03
Always add remote candidates from the libnice loop (see #2045)

lminiero:
See https://github.com/meetecho/janus-gateway/issues/2045#issuecomment-608544762 for details. Since this changes the way we process remote candidates we receive (whether in the SDP or via trickle), please make sure this is tested properly: I did a few brief checks and it seems to work, but better safe than sorry (and you'll be sorry if I merge and then it breaks because you didn't test it).

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -316,8 +316,11 @@ typedef struct janus_ice_queued_packet {
 } janus_ice_queued_packet;
 /* A few static, fake, messages we use as a trigger: e.g., to start a
  * new DTLS handshake, hangup a PeerConnection or close a handle */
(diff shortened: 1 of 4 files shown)


lminiero:
Moved `nice_agent_gather_candidates()` to the loop as well when doing full-trickle. The only disadvantage here is that we don't know right away if gathering failed (which should only happen when the RTP port range was exhausted, anyway), so negotiation goes on and we have to tear it down after that.

We still do that inline when doing half-trickle instead, as we need candidates to be available before we can ship the SDP in that case.

PR #2053, opened 2020-04-06
rtp forward: create both a ipv4 and ipv6 udp socket. fixes #2051

### LINKED ISSUE #2051: rtp forward with ipv6 udp socket does not work on macos
When setting up an RTP forward in the videoroom plugin, janus will create an ipv6 udp socket but then select an ipv4 or ipv6 sockaddr address depending on how the rtp forward was set up. Unfortunately calling sendto() on an ipv6 socket to an ipv4 address does not work on macos. I realize macos is not a first tier citizen in terms of janus support, but it would be a fairly minor change to support macos here.

The setup is that I have a gst-launch-1.0 process listening with a `udpsrc` element on 0.0.0.0:4005, and I create an RTP forward in janus with `host="127.0.0.1"` and `port=4005`.

The error shows up in the janus log
```
Error forwarding RTP audio packet for me... Invalid argument (len=98)...
```

The following C program also demonstrates the issue
```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/socket.h>
#include <sys/types.h>
#include <netdb.h>
#include <errno.h>
#include <string.h>
#include <arpa/inet.h>

int main(){
    int ok = 0;
    int port = 5454;
    int udp = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
    // int udp = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
    if (udp <= 0){
        printf("Unable to create socket: %s\n", strerror(errno));
        return 1;
    }

    printf("Created udp socket %d\n", udp);

    /*
    int v6only = 0;
    int ok = setsockopt(udp, IPPROTO_IPV6, IPV6_V6ONLY, &v6only, sizeof(v6only));
    if (ok < 0){
        printf("Could not set IPV6_V6ONLY=%d: %s\n", v6only, strerror(errno));
        return 1;
    }
    */

    /*
    struct sockaddr_in address;
    address.sin_family = AF_INET;
    inet_pton(AF_INET, "127.0.0.1", &address.sin_addr);
    address.sin_port = htons(port);
    */
    struct sockaddr_in6 address;
    address.sin6_family = AF_INET6;
    inet_pton(AF_INET6, "::1", &address.sin6_addr);
    address.sin6_port = htons(port);

    char buffer[10];
    sprintf(buffer, "test");
    ok = sendto(udp, buffer, strlen(buffer), 0, (struct sockaddr *) &address, sizeof(address));
    if (ok < 0){
        printf("Unable to send data: %s\n", strerror(errno));
        return 1;
    } else {
        printf("Sent %d bytes\n", ok);
    }

    close(udp);
}
```
Test this with
```
$ nc -u -l 5454
# In another terminal
$ clang x.c -o x
$ ./x
```
The word "test" should show up in the nc output.

On linux the socket can be ipv6 and the address can be ipv4, but on macos the socket and address must both be ipv6 or both be ipv4, otherwise sendto() will return -EINVAL.

The behavior in janus was changed in https://github.com/meetecho/janus-gateway/pull/1778

I was able to resolve this issue by making the following small change to plugins/janus_videoplugin.c
```
                if(publisher->udp_sock <= 0) {
                        // publisher->udp_sock = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
                        publisher->udp_sock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
                        int v6only = 0;
                        if(publisher->udp_sock <= 0
                                        // || setsockopt(publisher->udp_sock, IPPROTO_IPV6, IPV6_V6ONLY, &v6only, sizeof(v6only)) != 0
                             ) {

```
A more general simple fix is to use the check used elsewhere, which is `if serv_addr.sin_family == AF_INET{ .. do ipv4 stuff .. } else { .. do ipv6 suff .. }`. Would that be a reasonable compromise rather than having some `#ifdef MACOS` ? I can make a PR for this if so.

BTW I also tried having gst-launch-1.0 listen on an ipv6 address rather than ipv4, but gst just dies as soon as it receives the first udp packet from janus.
```
$ gst-launch-1.0 udpsrc uri="udp://[::0]:4000"  ...
Setting pipeline to PAUSED ...
Pipeline is live and does not need PREROLL ...
Setting pipeline to PLAYING ...
New clock: GstSystemClock
ERROR: from element /GstPipeline:pipeline0/GstUDPSrc:udpsrc0: Internal data stream error.
Additional debug info:
gstbasesrc.c(3072): void gst_base_src_loop(GstPad *) (): /GstPipeline:pipeline0/GstUDPSrc:udpsrc0:
streaming stopped, reason not-linked (-1)
Execution ended after 0:00:04.675785000
Setting pipeline to PAUSED ...
Setting pipeline to READY ...
Setting pipeline to NULL ...
Freeing pipeline ...
```

kazzmir:
For issue, https://github.com/meetecho/janus-gateway/issues/2051, macos (and probably other non-linux systems) cannot send data with a udp AF_INET6 packet to an ipv4 AF_INET address.
```
int udp = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
struct sockaddr_in address;
address.sin_family = AF_INET;
sendto(udp, ..., &address);
```

Currently in plugins/janus_videoroom.c a single `udp_sock` socket is created with AF_INET6 and use to send to both ipv4 and ipv6 addresses. The solution this PR has is to create two udp sockets, udp4_socket and udp6_socket, where if the address being sent to is ipv4 then use the udp4 socket, and if the address is ipv6 then use udp6 socket.

The implementation is straight forward, although it could be cleaned up due to having two sockets. In particular, the lines of code that set up the udp socket and address to invoke sendto() on could be abstracted out into its own function.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1520,7 +1520,8 @@ typedef struct janus_videoroom_publisher {
 	GHashTable *rtp_forwarders;
 	GHashTable *srtp_contexts;
 	janus_mutex rtp_forwarders_mutex;
-	int udp_sock; /* The udp socket on which to forward rtp packets */
+	int udp4_sock; /* The udp socket on which to forward rtp packets */
+	int udp6_sock; /* Similar to udp4_sock but for ipv6 */
 	gboolean kicked;	/* Whether this participant has been kicked */
 	volatile gint destroyed;
 	janus_refcount ref;
@@ -1622,8 +1623,10 @@ static void janus_videoroom_publisher_free(const janus_refcount *p_ref) {
 	janus_recorder_destroy(p->vrc);
 	janus_recorder_destroy(p->drc);
 
-	if(p->udp_sock > 0)
-		close(p->udp_sock);
+	if(p->udp4_sock > 0)
+		close(p->udp4_sock);
+	if(p->udp6_sock > 0)
+		close(p->udp6_sock);
 	g_hash_table_destroy(p->rtp_forwarders);
 	p->rtp_forwarders = NULL;
 	g_hash_table_destroy(p->srtp_contexts);
@@ -3684,11 +3687,23 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 			goto prepare_response;
 		}
 		janus_refcount_increase(&publisher->ref);	/* This is just to handle the request for now */
-		if(publisher->udp_sock <= 0) {
-			publisher->udp_sock = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
+		if(publisher->udp4_sock <= 0) {
+			publisher->udp4_sock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+			if(publisher->udp4_sock <= 0) {
+				janus_refcount_decrease(&publisher->ref);
+				janus_mutex_unlock(&videoroom->mutex);
+				janus_refcount_decrease(&videoroom->ref);
+				JANUS_LOG(LOG_ERR, "Could not open UDP IPV4 socket for RTP stream for publisher (%s)\n", publisher_id_str);
+				error_code = JANUS_VIDEOROOM_ERROR_UNKNOWN_ERROR;
+				g_snprintf(error_cause, 512, "Could not open UDP IPV4 socket for RTP stream");
+				goto prepare_response;
+			}
+		}
+		if(publisher->udp6_sock <= 0) {
+			publisher->udp6_sock = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
 			int v6only = 0;
-			if(publisher->udp_sock <= 0 ||
-					setsockopt(publisher->udp_sock, IPPROTO_IPV6, IPV6_V6ONLY, &v6only, sizeof(v6only)) != 0) {
+			if(publisher->udp6_sock <= 0 ||
+					setsockopt(publisher->udp6_sock, IPPROTO_IPV6, IPV6_V6ONLY, &v6only, sizeof(v6only)) != 0) {
 				janus_refcount_decrease(&publisher->ref);
 				janus_mutex_unlock(&videoroom->mutex);
 				janus_refcount_decrease(&videoroom->ref);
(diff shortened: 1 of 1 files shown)


lminiero:
I don't think that's the way to go. One socket should be enough, we just need to know in advance which one we'll need. That's why I suggested moving the [host resolution](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_videoroom.c#L1877) sooner than where it currently is.

lminiero:
PS: Sorry, forgot to say thanks for this! :blush:

lminiero:
(isolation is killing my manners)

kazzmir:
Ok I see what you mean. I will do the host resolution part earlier and then create the udp sock based on that. I am thinking the simplest way is just to call all the janus_videoroom_rtp_forwarder_add_helper() functions first, then set an `is_ipv4` flag in the publisher, and then use the `is_ipv4` flag to create the udp_sock, and also to help select the right parameters when calling sendto().

kazzmir:
is_ipv4 is set outside of janus_videoroom_rtp_forwarder_add_helper, and passed in as a flag now.

lminiero:
Mh that still feels a bit too convoluted and verbose to me, and adds a property to the signature that is unneeded IMO. I'll prepare a parallel PR that anticipates the host resolution.

lminiero:
As a side note, I think we have the same exact code in the AudioBridge plugin too (which supports RTP forwarders as well), so I'll make sure it's also applied there.

lminiero:
Ah, but another thing dawned on me, now... we're actually using a shared socket for publishers, that we use for ALL forwarders belonging to that publisher. If we force it to be just IPv4 or IPv6, it won't work if we want the same publisher to be sent to multiple addresses of different families. The socket that is created when setting up the forwarder itself is only the RTCP one, and that can actually still be IPv6 to receive either (it's how the Streaming plugin works too), so what needs fixing is just the RTP one.

My guess is that we should get rid of the global `udp_sock` in the publisher structure, and move that one to the forwarder struct as well. It will mean more sockets if we have multiple forwarders for the same publisher (so two rather than one if we're sending audio and video), but that's probably not a big deal. As an alternative, your original proposal of having two sockets (one v4 and one v6) in the publisher now makes more sense, even though each should then only be created when really needed (and so it should still be aware of what we're going to send to).

I'll think about this a bit more and then share something in a new PR.

lminiero:
Apparently it should be possible to detect this at runtime by checking if `IPV6_V6ONLY` is defined. This way I can add a different behaviour for systems that don't support it, and keep what we have now for Linux (which I'd rather not touch). I'll try to do that, so that you can check if it works for you (I don't have access to non-Linux machines so I wouldn't be able to test it).

lminiero:
FYI, apologies for this very late comment: I started working on that at the time, and then got swamped in a ton of other things. At the current stage, it's unlikely I'll get back to it any time soon, especially considering we're going to work on the new, revamped, multistream branch: since that will be a considerable refactoring of its own, I'd reject any bug change to the code base until that's merged. That said, I don't think the PR as it is works as explained in the comments above, so it unfortunately cannot be merged: as such, I think we'll close this for now, and we can come back to this once the multistream work has been done. Thanks anyway for the contribution!

lminiero:
I'll keep the original issue open and add the `enhancement` label just as a reminder.

PR #2055, opened 2020-04-07
Add configurable DSCP ToS for PeerConnections

lminiero:
See [this thread](https://groups.google.com/forum/#!topic/meetecho-janus/B10hy9hsJ6E) for context. This PR adds a new property called `dscp_tos` to the `media` section of `janus.jcfg`: if you put something there, then it's set as the DSCP Type of Service (ToS) in all media packets we send, using the `nice_agent_set_stream_tos` method in libnice.

Apparently the [specification](https://tools.ietf.org/html/draft-ietf-tsvwg-rtcweb-qos-18#page-6) suggests using different values for different media, but in Janus we bundle everything on the same ICE stream, and libnice only allows configuring it per stream (which makes sense), so you only get a single value you can set. By default it's 0 meaning that we don't set anything: choosing the right value is up to you.

Right now this is a global property, meaning it's set on all WebRTC PeerConnections we create. In the future we might want to make this configurable (e.g., per-session or even per-handle), so that people have more control over the priority they want to give to different connections. That said, nothing I plan to do in the near future: test this first, let's see if this works, and then we can talk about improving it.

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -205,6 +205,16 @@ media: {
 	#slowlink_threshold = 4
 	#twcc_period = 100
 	#dtls_timeout = 500
+
+	# If you need DSCP packet marking and prioritization, you can configure
+	# the 'dscp_tos' property to a specific values, and Janus will try to
+	# set it on all outgoing packets using libnice. Normally, the specs
+	# suggest to use different values depending on whether audio, video
+	# or data are used, but since all PeerConnections in Janus are bundled,
+	# we can only use one. You can refer to this document for more info:
+	# https://tools.ietf.org/html/draft-ietf-tsvwg-rtcweb-qos-18#page-6
+	# That said, DON'T TOUCH THIS IF YOU DON'T KNOW WHAT IT MEANS!
+	#dscp_tos = 46
 }
 
 # NAT-related stuff: specifically, you can configure the STUN/TURN
@@ -311,7 +321,7 @@ nat: {
 	# your machine is ready. Note that Linux distributions offer such directives.
 	# You could use the following directive in systemd: 'After=network-online.target'
 	# https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Before=
-	# ignore_unreachable_ice_server = true
+	#ignore_unreachable_ice_server = true
 }
 
 # You can choose which of the available plugins should be
--- a/ice.c
+++ b/ice.c
@@ -447,6 +447,18 @@ uint janus_get_twcc_period(void) {
 	return twcc_period;
 }
 
+/* DSCP Type of Service, which we can set via libnice: it's disabled by default */
+static int dscp_tos = 0;
+void janus_set_dscp_tos(int tos) {
+	dscp_tos = tos;
+	if(dscp_tos > 0) {
+		JANUS_LOG(LOG_VERB, "Setting DSCP Type of Service to %ds\n", dscp_tos);
+	}
+}
+int janus_get_dscp_tos(void) {
+	return dscp_tos;
+}
+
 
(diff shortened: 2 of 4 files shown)


lminiero:
As a side note, we're currently setting this when we create the agent, which means the same DSCP TOS is set for connectivity checks and DTLS handshake too. Not sure if we should only set it once the handshake is complete and the PeerConnection is up and running.

StephaneG31:
@lminiero we try to mark also packet for the client side (sender).
It seems to user [googDscp = true](http://www.rtcbits.com/2017/01/using-dscp-for-webrtc-packet-marking.html) doesn't work for Windows on we have to user priority https://bugs.chromium.org/p/webrtc/issues/detail?id=5836#c29
Do you think is the job a Janus.js to expose that priority or the role to the app JS code to do that ?

lminiero:
Not sure, honestly, as I was only interested in addressing the Janus part here. I'm not against having something in janus.js to tweak the client side, especially if it means passing for instance constraints on the PeerConnection (which janus.js tends to hide from the application anyway). If you can find ways to make this configurable in an easy way in janus.js, I'd be glad to evaluate a pull request :smile:

StephaneG31:
> Not sure, honestly, as I was only interested in addressing the Janus part here. I'm not against having something in janus.js to tweak the client side, especially if it means passing for instance constraints on the PeerConnection (which janus.js tends to hide from the application anyway). If you can find ways to make this configurable in an easy way in janus.js, I'd be glad to evaluate a pull request 😄

So we have to look where do that, it must be done just after the addTrack call : https://github.com/w3c/webrtc-priority/blob/master/explainer.md

StephaneG31:
@lminiero in our use case we use Janus as a SIP gateway for audio only relay to an SBC (connected to PSTN and phone network)
I try to understand how we can user DSCP on all links (AF46)
We do a traffic capture on Janus Server.

- CLIENT => JANUS  : we use googDscp for Linux/MacOS (on UDP paquet are marked with AF31) and for Windows only a GPO or registry key on HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\Windows\QoS\ can do the job. (UDP packet are marked with AF46)
- JANUS => SBC : from PCAP capture it seems we have DSCP 0, **do you think there's any option to set the DSCP ?**
- SBC => JANUS : we already receive AF46 from SBC
- JANUS => CLIENT : it's the purpose of this pullrequest, so you try to use it but it's not concluant (dscp is still 0) so we have to investigate.

lminiero:
> * JANUS => SBC : from PCAP capture it seems we have DSCP 0
>   Maybe just after create the socket we have to do someting like that (in janus_sip.c) :

My PR only impacts the WebRTC side, and I verified it worked in my case. Plugins generating traffic on their own have to set that on their own. No idea how that can be done with plain sockets, though, since I used a libnice method for the Janus core.

lminiero:
Wireshark showed the right value when I checked.

lminiero:
Ah I think I got your point now. Anyway, rather than changing the property in janus.jcfg, it's probably easier to let, e.g., 46 there, and then shift two bits on the left to set the actual value (`46 << 2` = `184`) in the call. Does this make sense? In case, we can rename the property to something more suited (maybe just `dscp`), and accept the old one for backwards compatibility.

StephaneG31:
Yes I think DSCP is more valuable than TOS.
But in fact libNice use IP_TOS (like Socket in low level).
https://libnice.freedesktop.org/libnice/NiceAgent.html#nice-agent-set-stream-tos

So I think it's better to use DSCP value than TOS value.

lminiero:
Yes, I'd set the shifted value in libnice. Okay, I'll make the change shortly.

lminiero:
This should fix it now. The property is named just `dscp` in `janus.jcfg`, and if you set `46` there we correctly set `184` in the TOS.

PR #2059, opened 2020-04-09
Added support for multichannel Opus audio (surround)

lminiero:
This PR adds preliminary support for multichannel Opus, which means you can now get surround audio via WebRTC as well when using Janus. You can learn all about how this actually works under the hood in [this excellent blog post](https://webrtcbydralex.com/index.php/2020/04/08/surround-sound-5-1-and-7-1-in-libwebrtc-and-chrome/) by @agouaillard, who found out about this possibility, and we actually used some demos written by @murillo128 as part of that effort as a basis for our work, so kudos to them for that!

In a nutshell, for what concerns Janus this now adds support for `multiopus` RTP maps when negotiating PeerConnections, not only to the core SDP stack but also to the Lua and Duktape ones. This was actually a bit more problematic than negotiating Opus, since there you always use `opus/48000/2` no matter whether the audio is actually stereo or not, while with Multiopus you're supposed to negotiate the actual number of channels. Since Chrome currently supports both 5.1 and 7.1, for WebRTC Opus, this means you can get either `multiopus/48000/6` and `multiopus/48000/8`: since we don't really support channel negotiation in the SDP, right now (which would require a substantial refactoring), this first integration is hardcoded to 5.1 instead. Maybe in the future we'll make this more flexible and dynamic, but it's a start.

The PR comes with a new demo as well, that uses one of the [Fraunhofer multichannel AAC samples](https://www2.iis.fraunhofer.de/AAC/multichannel.html) as the source of media we use in a PeerConnection: specifically, we open the file in a modified EchoTest page, use `captureStream` to get a MediaStream out of the video element, and send that instead of mic/webcam: if you get surround audio back in the demo, it means it's working.

Feedback welcome!


lminiero:
Almost forgot: while we do support recording `multiopus` streams to .mjr (which means that if you modify the Record&Play plugin, you'll be able to replay those), we currently do NOT support it in the `janus-pp-rec` post processing tool. In fact, as you'll be able to read in Alex's blog post, the format for these new audio frames is quite peculiar (which partly involves the ogg format as well), and so coming up with ways to get those files playable will be a challenge.

sdp-utils.c
```diff
@@ -1462,6 +1476,24 @@ janus_sdp *janus_sdp_generate_answer(janus_sdp *offer, ...) {
...
+					strstr(audio_fmtp, "channel_mapping") == NULL)) {
+				/* Missing channel mapping for the multiopus m-line, check the offer */
+				GList *mo = m->attributes;
+				while(mo) {
+					janus_sdp_attribute *a = (janus_sdp_attribute *)mo->data;
+					if(a->name && strstr(a->name, "fmtp") && a->value) {
+						char *tmp = strchr(a->value, ' ');
+						if(tmp && strlen(tmp) > 1) {
+							tmp++;
+							custom_audio_fmtp = g_strdup(tmp);
```
tmatth, line 1489: What about (probably invalid) SDPs with multiple `fmtp` lines? In this case, I think `custom_audio_fmtp` will leak.

lminiero: Good point, thanks, I'll think whether adding a `g_free` before the `g_strdup` to get rid of the previous value, or to show a warning that we're ignoring following values instead.

lminiero: Fixed by just entering the `if` when `custom_audio_fmtp` is NULL.

lminiero:
Not planning to touch this for a while, so I'll merge, since it's already in a usable state.

pgasiorowski-es:
Is it still supposed to be working in Chrome 90?
I also have problem playing back multiopus from this demo page:
https://sgmedooze.cosmosoftware.io/multiopus/

lminiero:
It's all explained in the box I added here: https://janus.conf.meetecho.com/multiopus.html

pgasiorowski-es:
Ok, thank you for pointing me at the explanation. It seems to be working as expected (?) with Chrome 91 (from Beta channel)

lminiero:
I have 92 installed and I see the video now, but there's no audio, so there must still be some problem.

pgasiorowski-es:
Ah. Now that is very wired as I'm getting an impression I'm seeing something opposite eg.:
* The video of the remote preview is frozen at one frame (screenshot attached)
* The audio of the remote preview is there (with 100-250ms delay) and can be muted, turned down/up separately from AAC source (as long as the aac source's is not muted).

frame-q1ds

PR #2063, opened 2020-04-10
Added option to configure time needed to detect a missing simulcast substream

lminiero:
See [this](https://groups.google.com/forum/#!topic/meetecho-janus/wGRCC8qGYZI) for context.

So far we've had a 250ms hardcoded value for detecting a missing substream: if we don't receive packets for a substream we're subscribed to for more than 250ms, then we consider the substream currently unavailable and drop to a lower substream instead. This works fine in most cases, except apparently when screensharing, as in some cases the framerate there can drop considerably: since 250ms correspond to about 4fps, anything below that would mean packets would be sent with larger intervals, thus triggering false positives.

This patch makes this property configurable in most plugins. For Streaming and VideoRoom plugins, these are per-subscriber configurations, _NOT_ per-publisher ones: each subscriber has their own simulcast context, which means they have their own tweak. I made the syntax pretty much the same in all plugins, so all you need to do is pass a

    fallback: <timer in microseconds>

in whatever request supports it to enforce it. The default if you don't specify it (or if you pass `0`) is `250000` as before.

Tested briefly with large values to see if this was behaving as expected and it looks like it is. I'll let you guys play with it more and shower me with tons of feedback.

### DIFF
--- a/plugins/janus_echotest.c
+++ b/plugins/janus_echotest.c
@@ -464,6 +464,8 @@ json_t *janus_echotest_query_session(janus_plugin_session *handle) {
 		json_object_set_new(info, "substream-target", json_integer(session->sim_context.substream_target));
 		json_object_set_new(info, "temporal-layer", json_integer(session->sim_context.templayer));
 		json_object_set_new(info, "temporal-layer-target", json_integer(session->sim_context.templayer_target));
+		if(session->sim_context.drop_trigger > 0)
+			json_object_set_new(info, "fallback", json_integer(session->sim_context.drop_trigger));
 	}
 	if(session->arc || session->vrc || session->drc) {
 		json_t *recording = json_object();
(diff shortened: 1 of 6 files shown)


akilude:
Thank you very much, will deploy it on a test server and give you some feedback.

akilude:
Hi @lminiero ,

Firstly, thanks again for the feature, i tested it with the echo test plugin and the videoroom plugins, both worked well, i updated the fallback value during the calls and could see much less substream thrashing when increasing the fallback value and also when decreasing the fallback value during the call lead to more substream thrashing so it definitely works. 


One thing that i've notice that might be a bug is sometimes the substreams does not fallback, i don't think this is due to this PR as its reproducible in master as well. Here is the scenario

1. Open https://janus.conf.meetecho.com/echotest.html?simulcast=true in chrome and start the demo
2. if you are on good internet, the selected substream should be S2 
3. Use the bandwidth dropdown to cap the bitrate to something lower like 256 or 128 kbps
4. Instead of the substream dropping down to 1 or 0, the video gets muted and never recovers, the selected substream is still S2. (This happens 3/5 times, the other 2 times the stream drops down to S0 and continues)
 

 Thanks,
Akil

lminiero:
Different issue, and unrelated to this enhancement. Merging.

PR #2065, opened 2020-04-11
configure: test earlier for AES-GCM

tmatth:
On systems with properly built libsrtp2:
```
checking for srtp_crypto_policy_set_aes_gcm_256_16_auth in -lsrtp2... yes
```
otherwise
```
checking for srtp_crypto_policy_set_aes_gcm_256_16_auth in -lsrtp2... no
checking for srtp_crypto_policy_set_aes_gcm_256_16_auth in -lsrtp... no
configure: error: libsrtp will fail to link due to missing AES-GCM support
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,7 +130,9 @@ AC_ARG_ENABLE([aes-gcm],
               [AS_HELP_STRING([--disable-aes-gcm],
                               [Disable AES-GCM support in libsrtp(2)])],
               [],
-              [AC_DEFINE(HAVE_SRTP_AESGCM)])
+              [AC_CHECK_LIB(srtp2,srtp_crypto_policy_set_aes_gcm_256_16_auth,,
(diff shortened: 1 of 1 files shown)


lminiero:
We had that already, it won't work if the libsrtp library pkg-config gets is somewhere else. I added a --disable-aes-gcm flag to the configure to disable that, it's in the FAQ too.

tmatth:
> We had that already, it won't work if the libsrtp library pkg-config gets is somewhere else. I added a --disable-aes-gcm flag to the configure to disable that, it's in the FAQ too.

Yeah I saw that in the FAQ, let me see if I can find a clean solution (otherwise I'll close).

lminiero:
@tmatth any update on this, or can it be closed?

tmatth:
> @tmatth any update on this, or can it be closed?

Hey not going to have time to further tweak this for a while, so closing for now.

PR #2068, opened 2020-04-14
Add missing info to videoroom "list" response

alexamirante:
The response to the `list` request addressed to the VideoRoom plugin was not returning all the information about rooms, mainly about extensions that will be negotiated.


lminiero:
Looks good, merging :+1:

PR #2074, opened 2020-04-16
Initial support for end-to-end encryption via Insertable Streams

lminiero:
Chrome has recently added experimental support for a new feature called [Insertable Streams](https://www.chromestatus.com/feature/6321945865879552): quoting from that link, it's an "API [that] enables the insertion of user-defined processing steps in the encoding and decoding of a WebRTC MediaStreamTrack". You can learn more in [this blog post](https://webrtcbydralex.com/index.php/2020/03/30/secure-frames-sframes-end-to-end-media-encryption-with-webrtc-now-in-chrome/) by @agouaillard, and in [this other blog post](https://webrtchacks.com/true-end-to-end-encryption-with-webrtc-insertable-streams/) by @fippo.

In a nutshell, when you use Insertable Streams the web application is notified about any frame you're about to send or have just received, allowing you to manipulate it before it is actually sent, or before it is passed to the decoder. This means that one of it's key use cases is obviously end-to-end encryption, as it allows you to, e.g., "encrypt" any video frame you're sending: only participants with the right key to "decrypt" the media content would be able to consume it. This way, no intermediary (servers like Janus included) has access to the media, since they wouldn't have access to the decrypting information.

This is obviously a very high level (and possibly partially incorrect) description, but it gives you an idea of what Insertable Streams allow you to do. It's worth pointing out, though, that this effort is very much a BEGINNING, not THE REAL DEAL. If you're expecting to get true end-to-end encryption right away, think again: at the moment, only video can be processed through Insertable Streams, not audio; besides, it will take time to work on the right functions to do the job, especially considering key management (which is a HUGE part of true end-to-end encryption) is still an open discussion. Again, a very good overview is provided in the above mentioned blog post by @agouaillard, who introduces a possible authentication scheme called SFrames (Secure Frames), which has been [used by Google itself](https://mailarchive.ietf.org/arch/msg/perc/s9pkcAbzLj5-LgE-W6l7xAgHQhc/) for quite some time already. This PR doesn't use SFrames yet (more on that later), but the plan is to work on that in the future too.

That said, using Insertable Streams in JavaScript means creating two transform functions:

1. a sender transform, to manipulate media on the way out (e.g., encrypt);
2. a receiver transform, to manipulate media on the way in (e.g., decrypt).

and then configure the WebRTC PeerConnection to use them, by basically applying these transforms to outgoing and incoming tracks respectively. Technically speaking, this is something that could be done completely on the client side, without involving Janus at all. My first experiments using EchoTest and VideoRoom, in fact, did exactly that, modifying `janus.js` for the purpose, and kinda worked fine already: even though media is encrypted, it's still RTP, and most plugins don't even look at the payload, or only use it partially. That said, it still makes sense to make Janus _aware_ that media will be encrypted: in fact, recording may be involved, which means the `.mjr` files should have a flag to clarify the content is encrypted and cannot be processed (but _can_ be replayed, e.g., via Record&Play, if the viewer has the right decrypt function); besides, in the future a new RTP extension will be added (as part of the AV1 specification, which brings us back to SFrames).

As such, this PR is an attempt to lay the ground for all that: it can be basically seen as a spiritual successor to #933, which we worked on with @agouaillard and @murillo128 at the time and had pretty much the same end-to-end encryption target in mind. As anticipated, it updates `janus.js` to support Insertable Streams, even though it expects the web application to provide the transform functions: the PR includes a modified EchoTest demo to do just that, using some transforms from [one of the official demos](https://webrtc.github.io/samples/src/content/peerconnection/endtoend-encryption/). When transforms are provided, `janus.js` also adds an `e2ee: true` property to the JSEP object sent to Janus, so that Janus is made aware that media will be encrypted: Janus will notify plugins as well. Plugins can in turn set the same property in the JSEP objects passed to the core in offers and answers, which means web applications can detect whether, e.g., a Streaming mountpoint or VideoRoom subscriber will be encrypted just by looking at whether the `e2ee` property in the JSEP they receive is `true`. When a PeerConnection is marked as end-to-end encrypted in Janus, any attempt to record it will result in the `e: true` property being added to the `.mjr` header: as such, you won't be able to post-process those recordings via `janus-pp-rec` (it would make no sense, since a random transform function may have been applied and media will be unaccessible), but it still makes for a good way of recording encrypted content (e.g., for replay via WebRTC later on). It's worth noting that, since only video is supported by Insertable Streams right now, we do _not_ mark audio recordings for `e2ee` PCs as encrypted at the moment, simply because they aren't; data channel recordings are obviously not encrypted either.

This is how all plugins are affected:

* **EchoTest**: it just takes note the PC may be encrypted, to mark it in the recordings.
* **VideoCall**: it takes note the PC may be encrypted, to mark it in the recordings, and sets `e2ee:true` in the offer sent to the callee if the caller is encrypted. The session will _not_ fail if the other side doesn't support it, because the `e2ee` property is currently a notification, not a negotiation: a session will still be established, media will simply be garbled becaused not decodable by either.
* **Streaming**: you can add an `e2ee:true` property to RTP mountpoints to clarify the media will be encrypted (e.g., if you're broadcasting media coming from an `e2ee` VideoRoom RTP forwarder).
* **SIP** and **NoSIP**: attempts to setup an `e2ee` PC are rejected (simply because we assume legacy endpoints would have no idea what this all is about).
* **AudioBridge**: attempts to setup an `e2ee` PC are rejected (the plugin needs to decode, mix, and encode content itself).
* **VideoRoom**: if a publisher uses `e2ee`, then the property is set in all offers we send to subscribers; you can configure a room with `require_e2ee:true`, which means attempts to publish _without_ `e2ee:true` will be rejected (useful in the future when you'll only want truly end-to-end encrypted rooms).
* **VideoCall**: attempts to setup an `e2ee` PC are rejected (the plugin needs to decode, mix, and encode content itself).
* **TextRoom**: not affected (only does data channels).
* **Lua** and **Duktape**: they take note the PC may be encrypted, to mark it in the recordings, and then leave the signalling and logic details up to the respective scripts.

This should be all. I didn't test extensively: I mostly tested with the EchoTest demo you can find here, and a bit with a VideoRoom demo that is not part of this PR. I encourage you to play with the demos to add support for this new feature yourself. Notice that this will **ONLY** work with Chrome canary or very recet versions (due to bugs in the original Chrome implementation, it only started working as expected in the EchoTest for me in `83.0.4103.14`), and will **NOT** work with older versions of the browser, or other browsers. Besides, you'll need to start Chrome with a custom flag at command line:

```
--enable-experimental-web-platform-features --force-fieldtrials=WebRTC-GenericDescriptorAdvertised/Enabled/
```

or Insertable Streams won't be available in recent Chrome versions either. A simple check to verify whether this is available for you or not, after using that flag, is checking whether `RTCRtpSender.prototype.createEncodedVideoStreams` is defined.

Feedback welcome!


html/janus.js
```diff
@@ -1772,6 +1787,20 @@ function Janus(gatewayCallbacks) {
 				pluginHandle.onremotestream(config.remoteStream);
 				if(event.track.onended)
 					return;
+				if(config.receiverTransform) {
+					var receiverStreams = null;
+					if(event.track.kind === "audio") {
+						// Audio isn't supported by Insertable Streams yet
```
fippo, line 1793: it is but you're stuck on an old version :-)
You could do feature detection but... it should be there really soon

html/e2etest.js
```diff
@@ -0,0 +1,654 @@
...
+// the page, meaning that, ideally, we'll be the only one able to decrypt
+// it: Janus will not have access to the media. In a scenario involving
+// multiple participants (e.g., a video call or conference), to use these
+// same functions you'll want to distribute the secret among all participants
+// using an out of band mechanism, so that they can encrypt and decrypt
+// the media and keep Janus out of the loop. Notice that these functions
+// are just an example, and are basically exact copies from this demo:
+// 		https://github.com/webrtc/samples/blob/gh-pages/src/content/peerconnection/endtoend-encryption/js/main.js
+// That said, you're free (or actually, encouraged) to try and use other
+// custom, and more advanced, trasform functions as well (e.g., SFrames).
```
fippo, line 545: typo: trasform

html/e2etest.js
```diff
@@ -0,0 +1,654 @@
...
+		// Called on startup.
+		console.log("[Sender transform] Startup");
+	},
+	transform(chunk, controller) {
+		//~ console.log("[Sender transform] Transforming");
+		// Copy of the above mentioned demo's encodeFunction()
+		const view = new DataView(chunk.data);
+		const newData = new ArrayBuffer(chunk.data.byteLength + 5);
+		const newView = new DataView(newData);
+		for(let i=0; i<10; ++i) {
```
fippo, line 601: you might want to update that one, in particular for audio there are three-byte (and even 0-byte) frames.

html/e2etest.js
```diff
@@ -0,0 +1,654 @@
...
+			newView.setInt8(i, view.getInt8(i));
+		}
+		for(let i=10; i<chunk.data.byteLength; ++i) {
+			const keyByte = currentCryptoKey.charCodeAt(i % currentCryptoKey.length);
+			newView.setInt8(i, view.getInt8(i) ^ keyByte);
+		}
+		newView.setUint8(chunk.data.byteLength, currentKeyIdentifier % 0xff);
```
fippo, line 608: can you :cherries:-pick the note from https://github.com/webrtc/samples/pull/1286?
Nobody wants the xor in production :-)

html/janus.js
lminiero: Thanks for the clarification! Knowing how Chrome packagers hate Linux, I'll probably have to wait a couple of weeks at the very least for that to land in my `google-chrome-unstable`... Does this mean I can check if `RTCRtpSender.prototype.createEncodedAudioStreams` exists too, or is there a different check? By the time this is merged, I'll make sure Janus addresses encrypted audio streams too.

html/e2etest.js
lminiero: No `n` in Italian (trasformare), so I'll play the "not my first language" card :laughing:

html/e2etest.js
lminiero: Thanks, I think I copied the functions a couple of weeks ago so it's very likely I'm using outdated examples. As soon as audio is supported in my browser I'll update and test that too.

html/e2etest.js
lminiero: Will do!

html/janus.js
fippo: I hope that next weeks unstable goes to M84. 84.0.4108.0 is pretty good. And yes @ createEncodedAudioStreams

lminiero:
Weirdly enough, I got a Chrome update sooner than expected (`84.0.4115.5`), so I changed the patch to support audio encryption (and flag in the recording) too. Transform functions haven't been updated yet.

lminiero:
Since I'm about to start working on a considerable code refactoring, I'll merge this as it is: it already works as expected in my simple tests, and as explained in my [blog post](https://www.meetecho.com/blog/janus-e2ee-sframe/), modifying `janus.js` to support SFrame is quite easy too. I can take care of adding the SFrame bits in a new PR later on, and we can address any other bug the moment people start playing with insertable streams, so no need to hold this.

alexamirante:
>Since I'm about to start working on a considerable code refactoring

I smell `v1.0.0` :yum: :yum:

Issue #2071, opened 2020-04-15
Heap Use After Free in HTTP Transport


bitfirenetworks:
Janus crashes intermittently with ~25 feeds in the VideoRoom with a use after free. Took a while to get this to happen intentionally, but I finally got an asan output - https://pastebin.com/KhhqmnnB
Running compiled from git at tag 0.9.2. Doesn't appear to be fixed in master.

lminiero:
We already have two issues open on HTTP, of which this is very likely a duplicate: #2062 #2066. Please don't open issues that are already open next time.

Issue #2066, opened 2020-04-11
Janus crash: transports/janus_http.c:1824


taaltaal:
We had some crashes before, this time we compiled with libasan as suggested in the docs.

The output is here:
https://pastebin.com/j7ucdmyE

P.S. Running from master. I hope this information helps!

lminiero:
Sounds like a similar issue as #2062. As I said there, enable refcount debugging by uncommenting the `#define REFCOUNT_DEBUG` line in refcount.h and recompile: if it crashes again, we'll need both the libasan info and the last part of the log to track what happened to the affected transport (long enough to track the reference that went to 0 from when it was started). Notice the log will be VERY verbose with refcount debugging enabled, so you may want to make sure you have enough room for that.

bitfirenetworks:
@lminiero, if this is related to #2071 as you suspect, this may be of use in debugging. I captured it with the REFCOUNT_DEBUG enabled: https://pastebin.com/DvBknrpB

lminiero:
@bitfirenetworks thanks, I'll have a look later today :+1:

bitfirenetworks:
Here's the same log and asan output, but all the way back to the creation of the object, with all the noise removed: https://pastebin.com/YgL9nDDU

lminiero:
Please don't trim the file. There's no guarantee that the lines you kept are related to the pointer that was affected: in your case, it dies on a `janus_transport_session` instance (pointer `0x60400087cd70`), but 1. there's no init anywhere (so no refcount initialization), and 2. none of the refcount references are anywhere near that pointer.

bitfirenetworks:
Here is the un-pruned log as far back as I have it for that crash: https://pastebin.com/qX0ex9mL

lminiero:
Thanks :+1:
This isn't master, though: please test with that, as for sure you're missing [one of the fixes](https://github.com/meetecho/janus-gateway/commit/1915f0fc07988fa70624d0a63b09ae6d7bc7510d) we had for requests being unreffed too soon (which could lead to transports being unreffed too soon as well). If it happens on master too, I'll have a look to see what else needs fixing.

bitfirenetworks:
Crashes in this area were more frequent in 0.9.2, but it happens pretty reliably in master as well. I'm early in my own hunt still, but I'm looking into the janus_http_timeout function as a potential suspect.
Crash in master: https://pastebin.com/fKnRF098

lminiero:
[...]

Issue #2062, opened 2020-04-10
Crash in HTTP transport


NikSmith:
Hello.

I use Janus to broadcast video to other servers using rtp_forwarding.
I do not have accurate statistics of the crashes, but this happens after a few hours of work with a small load (about 5-7 videorooms, 1 participiant = 1 videoroom).

libmicrohttpd12: 0.9.59-1

logs: https://pastebin.com/raw/U53v9cT0

lminiero:
Why don't you guys READ THE GUIDELINES! NO HUGE INLINE TEXT!

NikSmith:
Sorry. I updated issue.

tmatth:
@NikSmith are you using Janus v0.9.2? There was a fix to add reference counting to requests that might avoid this issue.

NikSmith:
> @NikSmith are you using Janus v0.9.2? There was a fix to add reference counting to requests that might avoid this issue.

I use Janus from master. I'm not sure, but I don't get this problem in v0.9.2

lminiero:
That's not a crash in the VideoRoom, it's a crash in the HTTP transport. Enable refcount debugging by uncommenting the `#define REFCOUNT_DEBUG` line in `refcount.h` and recompile: if it crashes again, we'll need the last part of the log to track what happened to the affected transport.

lminiero:
PS: notice the log will be VERY verbose with refcount debugging enabled, so you may want to make sure you have enough room for that.

NikSmith:
> That's not a crash in the VideoRoom, it's a crash in the HTTP transport. Enable refcount debugging by uncommenting the `#define REFCOUNT_DEBUG` line in `refcount.h` and recompile: if it crashes again, we'll need the last part of the log to track what happened to the affected transport.

I got new logs: https://pastebin.com/raw/qDvN2Eat

NikSmith:
The chart shows the frequency of crashes:

![image](https://user-images.githubusercontent.com/6698935/79101755-e8e01f00-7d71-11ea-9a29-266ac18712b0.png)

alokazay:
![image](https://user-images.githubusercontent.com/2505029/79148000-80269000-7dcd-11ea-9803-adb796d57554.png)  
 
Why is this happening? Screenshot from pages https://janus.conf.meetecho.com/videoroomtest.html

NikSmith:
@lminiero New logs https://pastebin.com/RVWuLxaa
I hope this can help solve the problem.

[...]

PR #2075, opened 2020-04-16
Use custom GSource to handle HTTP request timeouts (see #2062 and #2066)

lminiero:
We've seen a few issues related to HTTP, lately (#2062 and #2066, plus the #2071 duplicate). They mostly seem to be caused by race conditions, and one in particular (which may be the root cause of the other issues too) is apparently caused by an attempt to send an event on a long poll while at the same time the long poll expired and we're sending an empty response back. This race condition can lead to a double unref of the `janus_transport_session` instance that acts as a glue between core and transport, thus causing havoc.

The race condition can happen when:

1. we send an event, notice a timeout timer is still active, and so we destroy it and clean the resources (unref the transport and session related to it)
2. the timeout fires at the same time, which uses/unrefs the same data.

This patch tries to solve it by changing the stock glib timeout source we were using with a custom one: this custom source behaves pretty much as the existing timeout source (we use the same way to trigger it), but keeps track of the transport and session itself: both are only unreffed when the source is finalized, however that happens (whether it's because 1. happened, or because the timeout fired because indeed there was no event to send in 30s).

I tested this briefly but I can't check if this really solves the issue myself, since I never managed to replicate the issue that have been reported. If this patch doesn't solve the problem and you have simple ways to consistently replicate, please let me know.

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -127,6 +127,7 @@ typedef struct janus_http_msg {
 	char *response;						/* The response from the core as a string */
 	size_t resplen;						/* Length of the response in octets */
 	GSource *timeout;					/* Timeout monitor, if any */
+	volatile gint timeout_flag;			/* Whether a timeout hasn't fired yet */
 	volatile gint destroyed;			/* Whether this session has been destroyed */
 	janus_refcount ref;					/* Reference counter for this message */
 } janus_http_msg;
@@ -996,10 +997,9 @@ int janus_http_send_message(janus_transport_session *transport, void *request_id
 			if(msg && g_atomic_pointer_compare_and_exchange(&msg->longpoll, session, NULL)) {
 				janus_refcount_increase(&msg->ref);
 				/* Send the events back */
-				if(msg->timeout != NULL) {
+				if(g_atomic_int_compare_and_exchange(&msg->timeout_flag, 1, 0)) {
 					g_source_destroy(msg->timeout);
 					g_source_unref(msg->timeout);
-					msg->timeout = NULL;
 				}
 				msg->timeout = NULL;
 				janus_http_notifier(msg);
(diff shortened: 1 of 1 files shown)


lminiero:
Notice that the patch currently has some verbose debugging ([WARN]) to track the state of a timeout source: no need to worry about that, it's not really a warning, I just need it to track what's happening to the source when I look at logs.

atoppi:
lgtm

bitfirenetworks:
This looks great, and addresses something very close to what I was suspecting as well. Thanks for your responsiveness and for putting this PR together so quickly! We'll test it out and put it into use, and I'll let you know how it goes.

NikSmith:
It seems to work. We started the server yesterday, there are no crashes. Little time has passed, we continue to test. In a couple of days I will write a feedback. Thanks for the help.

TomFFF:
We also noticed the same problem and are using this version on one of our servers since yesterday (with libasan). Looking good so far, we will give feedback on Monday.  Thanks for the help

lminiero:
Just as a heads up, I most likely won't wait Monday before merging, considering it's a critical fix. If nobody reports issues by the end of the day, I'll just merge. Other issues appearing later on, in case, can be addressed after that.

antoniotuzzi:
I've recompiled and using it since yesterday, I'll keep you informed if any problem
(today I recompiled and restarted again with the atomic flag too)

MvEerd:
Holding up here as well so far :+1:  will report back if we run into anything before the end of the day
Thanks for your continued efforts

lminiero:
Merging :+1:

NikSmith:
Unfortunately, this remains unstable. But the frequency of crashes has decreased significantly.

image

I got another error, maybe this is not related to the current problem.

> janus[31012]: (process:31012): GLib-ERROR **: 18:15:36.552: Creating pipes for GWakeup: Too many open files

NikSmith:
I will increase the number of open files and start testing again.

Issue #2070, opened 2020-04-15
"Allow" adding params to sendBeacon in janus.js for session destroy


DaGLiMiOuX:
After Chrome 80, a `sendBeacon` for session destroy on unload was the replacement for the XHR sync function #1902 .

I've found a problem with this function, because we require to add params in URL, like a token or whatever information we need. We've found that this is an issue, because we are not allowing accesing directly to janus server from client browser, instead requests are managed by our own app, then redirected to janus, acting as a proxy for janus requests.

Would be helpful to allow adding params for `sendBeacon` request. We have no problem with any other request, because we are using jQuery ajax to perform requests and we modify its default settings to add those params when janus.js methods are called to perform requests to the server.

I would like to hear from you about this or if you counted on it, so we can try to perform a PR into `janus.js`.

Br,
D.

lminiero:
Why not set your own page-closed detector, and do the sendBeacon there? If you'd rather provide a PR, I'll be glad to review it (assuming it doesn't break backwards compatibility).

DaGLiMiOuX:
> Why not set your own page-closed detector, and do the sendBeacon there? If you'd rather provide a PR, I'll be glad to review it (assuming it doesn't break backwards compatibility).

I have my own sendBeacon detector, but in case someone has a similar situation where a param must be set to handle requests, even if they have their own sendBeacon or not, I think that allowing some extra params in URL could be helpful.

Btw, I'm using the built-in closing session using sendBeacon to prevent modifying janus.js nor programming what is already done.

I'll try to post a simple but working PR this afternoon.

Br,
D.

DaGLiMiOuX:
Refering again to my comment into PR: Any modification into this function, even just the URL, would be enough to create your own sendBeacon.

PR #2086, opened 2020-04-19
Include params in uri for sendBeacon in janus.js

DaGLiMiOuX:
Now we can use a `key: value` object to add parameters into the uri for sendBeacon. Useful if you use JWT tokens, for example or need to send extra information when unload event is reached and you don't want/need to build your own sendBeacon, refered in #2070

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -497,8 +497,8 @@ function Janus(gatewayCallbacks) {
 		longPollTimeout = gatewayCallbacks.longPollTimeout;
 	if(isNaN(longPollTimeout))
 		longPollTimeout = 60000;
-  // Parameters for sendBeacon URI when unload event is reached
-  var unloadParams = gatewayCallbacks.unloadParams || {};
+	// Parameters for sendBeacon URI when unload event is reached
+	var unloadParams = gatewayCallbacks.unloadParams || {};
 
 	// overrides for default maxBitrate values for simulcasting
 	function getMaxBitrates(simulcastMaxBitrates) {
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @DaGLiMiOuX! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
@DaGLiMiOuX thanks! Please fix your patch so that is uses tabs and not spaces, though.

html/janus.js
```diff
@@ -988,7 +990,20 @@ function Janus(gatewayCallbacks) {
...
+        var uri = server + "/" + sessionId;
+        if (unloadParams) {
+          Object.keys(unloadParams)
+            .filter(function(key) { return unloadParams[key] != null; }) // Ignore null or undefined params
+            .forEach(function(key) {
+              // If there are more than 1 param, we need to use '&' instead of '?' after first param
+              if (uri.search(/\?.*=.*/) >= 0) {
+                uri += '&' + key + '=' + unloadParams[key];
+              } else {
+                uri += '?' + key + '=' + unloadParams[key];
```
lminiero, line 1002: Shouldn't there be an escaping of those values? Or will `sendBeacon` do that itself?

html/janus.js
```diff
@@ -988,7 +990,20 @@ function Janus(gatewayCallbacks) {
...
+              }
+            });
+        }
+				navigator.sendBeacon(uri, JSON.stringify(request));
```
lminiero, line 1006: Not sure what's the point of adding those query string arguments, if we're still sending the request to Janus. The Janus integrated web server will ignore them all.

html/janus.js
```diff
@@ -988,7 +990,20 @@ function Janus(gatewayCallbacks) {
...
 				ws = null;
 			} else {
-				navigator.sendBeacon(server + "/" + sessionId, JSON.stringify(request));
+        var uri = server + "/" + sessionId;
+        if (unloadParams) {
+          Object.keys(unloadParams)
+            .filter(function(key) { return unloadParams[key] != null; }) // Ignore null or undefined params
+            .forEach(function(key) {
+              // If there are more than 1 param, we need to use '&' instead of '?' after first param
+              if (uri.search(/\?.*=.*/) >= 0) {
```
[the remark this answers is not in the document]
DaGLiMiOuX: I will try to fix that to search everything after server + session id and search for "?<key>=<value>" to know if there is already a param with a value as first parameter. Thanks for pointing out that it could search for a question mark in any part of the uri.

I'll try to fix it ASAP.

html/janus.js
DaGLiMiOuX: I will add an escaping for those values using `encodeURIComponent` in case someone uses `&` in a param.

html/janus.js
DaGLiMiOuX: This PR was made to use params in URI for `sendBeacon` if the Janus server is not achieved directly, for example using something like a proxy. In my own case, we have developed an admin panel to handle all web requests, so janus is never reached directly (to prevent unwanted access/attacks to the public server and only access through our authentication system). That made us to improve how calls are made in JS client to reach janus and we found that the `sendBeacon` URI was not "customizable".

Indeed, as discussed in the related issue, we could create our own `sendBeacon`, but we thought that someone who wanted to use janus behind any kind of proxy or redirecting requests could find an easy way to implement the current janus.js unload function in order to prevent doing themselves (e.g. they don't know how to do it or w/e).

As a side note, I'll like to mention that this was just to know if it could be useful to be implemented. Just let me know if you think this couldn't be useful at all in a day-by-day scenario.

lminiero: I've given this some thought and I don't think this is generic enough to be part of janus.js, sorry. This seems more specific to your use case, which is fine, but doesn't mean it should be part of the repo. Closing.

DaGLiMiOuX:
I'm completely agree. I was working on this for some weeks, but even if it was working for our use case, we found that any modification into sendBeacon function call would be enough to create your own sendBeacon when closing the room.

I was about to close it yesterday, but I was a bit busy, sorry.

Thanks for your time again.
Br,
D.

PR #2088, opened 2020-04-20
Add support for playback of audio files in AudioBridge

lminiero:
This PR continues the work started in #2040: where that patch added support for Opus files to the Streaming plugin, this does the same for the AudioBridge. This can be usesul for several reasons, like playing short announcements, or some music in background in a loop while you wait for the session to start.

> **NOTE WELL:** Notice that I'm just using these are _examples_: they're NOT features of the AudioBridge as you'd expect them of a common bridge. If you want to have announcements, it's up to YOU to record those to some Opus file somewhere: the AudioBridge will just play whatever file you're asking to play. Same thing for music on hold: an announcement will NOT mute other users, it will just be ANOTHER contribution in the mix, and so will play along other participants. If you want something like true music-on-hold, it's again up to YOU to mute everyone and just let the file play. This feature is meant to be a low-level feature you can use for different purposes: how to integrate this in your application flow is out of scope here.

That said, the request to play files is synchronous, meaning it can be done via Admin API as well, and is protected by the room password, which means only the "admin" of the room is supposed to be able to do that. Optionally this can be protected via the global `admin_key` as well. The syntax is quite straightforward:

```
{
	"request": "play_file",
	"room": <room ID>,
	"secret": "<room secret, if required>",
	"file_id": "<unique string ID of the announcement; random if not provided>",
	"filename": "<path to the Opus file to play>",
	"loop": <true|false, depending on whether or not the file should be played in a loop forever>
}
```

Once you know the `file_id`, stopping a playback before it's over is quite easy too:

```
{
	"request": "stop_file",
	"room": <room ID>,
	"secret": "<room secret, if required>",
	"file_id": "<unique string ID of the announcement>"
}
```

The playback will stop on its own (and the announcement removed) when the file reaches its end and loop was set to false. An event is sent to all participants about announcements starting and stopping, which will contain the `file_id`.

Tested briefly and it seems to be working as expected for me, but feedback welcome.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -407,9 +407,9 @@ plugins_libadd = \
 if ENABLE_PLUGIN_AUDIOBRIDGE
 plugin_LTLIBRARIES += plugins/libjanus_audiobridge.la
 plugins_libjanus_audiobridge_la_SOURCES = plugins/janus_audiobridge.c
-plugins_libjanus_audiobridge_la_CFLAGS = $(plugins_cflags) $(OPUS_CFLAGS)
-plugins_libjanus_audiobridge_la_LDFLAGS = $(plugins_ldflags) $(OPUS_LDFLAGS) $(OPUS_LIBS)
-plugins_libjanus_audiobridge_la_LIBADD = $(plugins_libadd) $(OPUS_LIBADD)
+plugins_libjanus_audiobridge_la_CFLAGS = $(plugins_cflags) $(OPUS_CFLAGS) $(OGG_CFLAGS)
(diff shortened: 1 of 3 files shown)


lminiero:
I should clarify that the path `play_file` expects is a local path: HTTP urls or stuff like that won't work. If the file is remote, download it yourself somehow and then pass the file system path to the call.

RSATom:
@lminiero do you plan merge it to master in the near future?

lminiero:
The sooner I get feedback, the sooner this is merged.

RSATom:
Ok, then I'll do snap build for that. Thanks!

RSATom:
Added branch for snap.
It can be installed by 
`sudo snap install janus-gateway --channel=edge/audiobridge-audiofile`
or updated by
`sudo snap refresh janus-gateway --channel=edge/audiobridge-audiofile`

lminiero:
Merging.

Issue #2087, opened 2020-04-19
janus_videoroom_hangup_subscriber segfaults with lots of publishers / subscribers leaving at the same time


cb22:
I'm not sure if this is related to #2034, but it looks like there's quite a specific setup required to trigger this bug and I can't see any information there linking the two besides both involving the videoroom plugin.

The setup:
* Latest janus master
* 8 event loop threads
* private_ids on (mentioned due to the stacktrace)
* Lots of active rooms, with publishers / subscribers leaving due to being killed (not leaving gracefully)

Janus segfaults when a large amount of participants leave the videoroom [(stacktrace)](https://pastebin.com/AJkA9GDY). My testing setup here consists of ~100x Google Cloud VM instances, each running 3x Chrome instances per videoroom, to two janus servers. 

I can seemingly replicate this crash consistently. Interesting to note, is that the crash doesn't seem to happen if I turn off a fixed number of event loop threads, nor if the publishers / subscribers send leave requests. I'm not sure if those two points might just be a red herring due to an underlying race condition though.

The specific line mentioned in the stacktrace is [here](https://github.com/meetecho/janus-gateway/blob/02dd7da5aa251b4a93a0f1ba0a4c67921b6fb825/plugins/janus_videoroom.c#L5167)

Some speculation (I'm not familiar with GLib!) is perhaps there's a TOUTOC between the check for `s->room != NULL` and the usage of `g_hash_table_lookup(s->room->private_ids`. Anyhow, I'm going to do some more invetigation, but I'm happy to provide any more info or run tests as required!

lminiero:
I had to google `TOUTOC` to see what it meant... :hand_over_mouth: 
A libasan dump might help, in this case, as that would tell if that pointer was freed somewhere or just set to NULL.

cb22:
Managed to recreate it with a simpler setup than before - 1 publisher and 8 subscribers in the same room will do it fairly repeatedly, as long as the subscribers all leave at the same time (`killall -9 chromium` does the job quite well if they're all on the same machine)

I've got a libasan dump ([here](https://pastebin.com/MXJNYexA)) but it's not going to be too useful for others since I've added a few debugging statements and extra code to narrow down the issue (such as an extra check on `s->room != NULL`).

[...]

PR #2091, opened 2020-04-20
Fix use after free in janus_videoroom_handler with events

cb22:
When trying to get janus up and running with asan for #2087, it would report a heap-user-after-free error ([cleaned up output](https://pastebin.com/2ZBkztXN), [full output](https://pastebin.com/SJ3q6VRp))

It looks like `feed_id_str` was being destroyed by a `json_decref` / `janus_videoroom_message_free` before being used to send the event off. One solution which fixed the problem for me was to simply decref and free after processing the event.


lminiero:
Looks reasonable, merging :+1:

PR #2093, opened 2020-04-22
More checks when hanging up VideoRoom subscriber (see #2087)

### LINKED ISSUE #2087: janus_videoroom_hangup_subscriber segfaults with lots of publishers / subscribers leaving at the same time

lminiero:
As the title says, attempt to fix #2087. WIP.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -5156,22 +5156,25 @@ void janus_videoroom_hangup_media(janus_plugin_session *handle) {
 	janus_videoroom_hangup_media_internal(handle);
 }
 
-static void janus_videoroom_hangup_subscriber(janus_videoroom_subscriber * s) {
+static void janus_videoroom_hangup_subscriber(janus_videoroom_subscriber *s) {
 	/* Already hung up */
 	if (!s->feed) {
 		return;
 	}
 	/* Check if the owner needs to be cleaned up */
-	if(s->pvt_id > 0 && s->room != NULL) {
-		janus_mutex_lock(&s->room->mutex);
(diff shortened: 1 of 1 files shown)


lminiero:
Looks like it works, so I'll merge.

Issue #2015, opened 2020-03-23
ICE thread sometimes peg CPU


gfodor:
Hi all, I am an engineer working on the same project as @mqp was when this issue was opened: https://github.com/meetecho/janus-gateway/issues/1260

At the time, that issue was closed due to the fact the CPU pegging behavior we had seen seemed to discontinue, but in fact it hadn't, and we never followed up. In the interim, we decided to wait and see how things would fare when we upgraded janus and its various dependencies.

Unfortunately, after the upgrade, we still see the behavior of pegged threads. Some relevant info:

- We have upgraded to janus 0.7.6 - we cannot fully upgrade to the latest version due to plugin incompatibilities that need to be addressed. I realize it's frowned upon to open bugs for older versions, but at least would like to know if there is reason to believe this problem is fixed in a newer version so we can prioritize the plugin migration process.

- We have upgraded the various dependencies, notably, we have upgraded libnice to master HEAD as of a few days ago.

The behavior we see is occasionally an `hloop` thread gets pegged, reporting ~50% CPU utilization in `ps` and one of the CPUs becomes pegged at 100%. Attached is a perf trace ([perf.zip](https://github.com/meetecho/janus-gateway/files/4370876/perf.zip), showing that the time is spent polling in g_poll. These events seem relatively rare, and are hard to reproduce. We managed to reproduce it once under load testing but in general there's no systematic way to do it - it seems to be a very low probability event, though it does seem to happen more frequently against production traffic vs artificial load testing (our load test uses headless Chrome browsers.)

On one of our production nodes, we managed to grab all the relevant metadata from the admin API re: handles. The only fishy thing was one session had a relatively large number of handles (25 or so) compared to others and all of them but one had false flags around offer negotiation. Other sessions had similar non-negotiated handles but only this one 'degenerate' session had so many handles, and only a single successfully negotiated handle. Hard to say if this is useful info or not. Here is a dump of the handles: [all_handles.zip](https://github.com/meetecho/janus-gateway/files/4370883/all_handles.zip) I did a bunch of ad hoc analysis to see if there were any telltale signs that a session was 'corrupted' based upon its metadata - the best I could come up with is that there is a session with only a single handle successfully negotiated with a lot of other non-negotiated ones - the other sessions all had at least two successfully negotiated handles.

Some things we tried/discovered:

- Forcibly detaching the handles and/or destroying the sessions from the admin API failed to resolve the CPU being pegged.

[...]

PR #2101, opened 2020-04-24
Bugfix/cpu usage based on v0.8.2

sjkummer:
avoid calling nice_agent_attach_recv() with a NULL callback, see https://gitlab.freedesktop.org/libnice/libnice/issues/14

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1257,7 +1257,7 @@ gint janus_ice_handle_destroy(void *core_session, janus_ice_handle *handle) {
 		janus_flags_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_STOP);
 		if(handle->mainloop != NULL) {
 			if(handle->stream_id > 0) {
-				nice_agent_attach_recv(handle->agent, handle->stream_id, 1, g_main_loop_get_context(handle->mainloop), NULL, NULL);
+				// nice_agent_attach_recv(handle->agent, handle->stream_id, 1, g_main_loop_get_context(handle->mainloop), NULL, NULL);
 			}
 			if(static_event_loops == 0 && handle->mainloop != NULL && g_main_loop_is_running(handle->mainloop)) {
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! I'm wondering if this warrants some new check in `janus_ice_cb_nice_recv`, though, as the main reason we set that callback to `NULL` is make sure we don't pass any media to plugins when we want to hang up the PeerConnection. Keeping the callback as it is will let libnice keep on telling about incoming packets, which means `incoming_rtp` may still occur after `hangup_media`, unless we have other checks to block media even when we're told about it.

sjkummer:
Hm, we could use `nice_agent_remove_stream()` as Olivier Crête suggested (but I didn't test that yet) or implement an additional check in the callback function. What would you prefer?

lminiero:
I think the additional check is the safest approach, especially since you tested this approach already. One more thing I'm wondering, though, is related to a comment I made on the original issue: https://github.com/meetecho/janus-gateway/issues/2015#issuecomment-605130186
If this only happens when calls fail, then it may be enough to only set the callback if the gathering works, while at the moment we always set it and then remove it when hanging up: unless the failures you were experiencing were not in `nice_agent_gather_candidates` but somewhere else?

sjkummer:
Okay, I've updated the PR with an additional check.

Regarding gathering: In my logs I don't see `Error gathering candidates` ...

lminiero:
Thanks! I'll make a couple of quick tests myself later and then merge :+1:

lminiero:
Looks good, merging :+1:

sjkummer:
Great, thanks 🙂

lminiero:
@sjkummer did you manage to replicate the issue fairly easily, before this fix? If so, any chance you can capture a full libnice log (it should be enough to enable it in `janus.jcfg`) of when the problem happens, and share it with the libnice developers here? https://gitlab.freedesktop.org/libnice/libnice/-/issues/107
Thanks!

Issue #2099, opened 2020-04-24
Massive disconnect of publishing stream causes heap-use-after-free crash


agclark81:
We've seen at least one heap-use-after-free crash occur in janus_videoroom_hangup_subscriber, seemingly at the very beginning on the existence check of "if (!s->feed)". We're running the latest master including the changes from PR #2093.

https://github.com/meetecho/janus-gateway/blob/4fb459e5906177c4c7452ceb7ffd761286f54dba/plugins/janus_videoroom.c#L5161

Logs with asan output is here: https://pastebin.com/HHp8Qv5Z

lminiero:
See PR above, which we were working on for a customer, and may be related to your issue.

agclark81:
Great; we'll check this out and see how it fares. Thanks!

lminiero:
Any updates on this? I'd like to merge soon.

agclark81:
We’ve been running it in prod for several days now, and we’ve observed no new crashes, so I think it’s safe to say there’s no major regression or issue with the PR.

lminiero:
Thanks, merged :+1:

PR #2102, opened 2020-04-24
Protect with a mutex janus_hangup_subscriber and subscribers list

atoppi:
This PR tries to solve some crashes happening in the videoroom plugin when subscribers leave the room  while a publisher is being disconnected (like #2099).
We basically protected the `janus_hangup_subscriber` and the publisher `subscribers` list with the already existent `subscribers_mutex`.
We also protected the publisher own subscriptions (used for `private_ids`) with a new dedicated mutex and not reusing the same `subscribers_mutex` (that would lead to a double locking).


nysthi:
I have this PR up and running in the last hour
I'll report back for any news

lminiero:
Merging.

Issue #2023, opened 2020-03-26
RabbitMQ exchange type should be configurable


mspanc:
In 0.9.1 the RabbitMQ exchange type is not configurable and is set to `fanout`.

The initialization procedure seems to be the following:

1. Connect to the RabbitMQ
2. Establish the channel
3. Declare an exchange as `fanout`
4. Declare an incoming queue
5. Declare an outgoing queue

It raises the following problems:

1. Unless the channel is bound to the exchange, sending a message into it will not cause any routing - and here binding does not happen
2. If you bind these two channels to the same exchange but the exchange is set to fanout, a message sent to the exchange will be delivered to both queues. Effectively Janus will read its own messages. Moreover, that leads to an endless loop of errors, as if you create a session, and Janus will reply with `success` message it will be routed back to Janus and it will reply with `error` message (as `success` is not valid call),  which will be again routed back to Janus etc.
 
The proper default procedure should be the following:

1. Connect to the RabbitMQ
2. Establish the channel
3. Declare an exchange as `direct`
4. Declare an incoming queue
5. Declare an outgoing queue
6. Bind the incoming queue to the exchange and set routing_key to the incoming queue name
7. Bind the outgoing queue to the exchange and set routing_key to the outgoing queue name

But for more advanced setups it should be configurable (at least the exchange type).

lminiero:
We're extremely busy and won't certainly have time to fix all this, as this is quite low priority for us. Should you be willing to provide a patch, we'll be glad to review a pull request.

pitkonst:
We are using RabbitMQ transport as it is implemented, and do not encount porblems you described. Fanout exchange is used to deliver same result message to multiple servers.

lminiero:
@mspanc @pitkonst there's a new PR implemented by @voicenter in #2104 that should provide support for that. It looks fine to me, so please test and in case it works for you too I'll merge.

lminiero:
Closing as the PR has been merged.

PR #2104, opened 2020-04-24
rabbitmq exchange type as config value

voicenter:
By the feature request mention in #2023

### DIFF
--- a/conf/janus.eventhandler.rabbitmqevh.jcfg.sample
+++ b/conf/janus.eventhandler.rabbitmqevh.jcfg.sample
@@ -19,6 +19,7 @@ general: {
 	#vhost = "/"					# Virtual host to specify when logging in, if needed
 	#exchange = "janus-exchange"
 	route_key = "janus-events"		# Name of the queue for event messages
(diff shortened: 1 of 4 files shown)


lminiero:
Thanks! I'll notify the people in that issue about this effort so that they can test it.

pitkonst:
Looks fine for me. There is FIXME comment in both files (line 95 and 112), witch could be removed.

voicenter:
@pitkonst , let me know if any more change needed,
I didn't add the parameter to the config file as it is optional, 
let me know what is the project  policy about it,
should I have it in the doc and sample config files...
thanks again ...
Shlomi

pitkonst:
About project policies we better ask @lminiero

lminiero:
A commented value in both config files (just to let people know it exists) would definitely help. You can use the existing lines as a template for the new property. I also noticed your patch has a lot of empty lines in between code snippets, so please just keep one for those (even though it's an easy nit I can fix myself after merging).

lminiero:
Thanks for the quick fixes! This seems good to merge for me: unless I see any objection on the original issue, I'll merge tomorrow morning :+1:

lminiero:
I don't see any objection, so I'll merge, thanks again! :+1:

PR #2114, opened 2020-04-28
Audiobridge and Streaming plugins don't play last chunk of .opus file

RSATom:


lminiero:
Thanks for the quick fix, merging! :+1:

Issue #2106, opened 2020-04-25
Janus crash in streaming plugin : Destroy during switching mountpoints


TomFFF:
We experience a crash in janus "heap-use-after-free"

Version used: janus master branch (latest commit 4fb459e5906177c4c7452ceb7ffd761286f54dba)

Raw log from the start of the switch of mountpoints until the crash : https://pastebin.com/jZft3v52
Libasan output : https://pastebin.com/Q8dJGCWA
Log with some comments (including creation mountpoints/ref debug affected session) : https://pastebin.com/m1yBk8x4

This is what we think happens:
When a destroy is processed during a switch of mountpoints, janus crashes (heap-use-after-free) in janus_streaming.c on line 4418
>	session->mountpoint = NULL;

This is caused because `janus_streaming_hangup_media_internal` is executed twice ( in `janus_streaming_hangup_media` and  `janus_streaming_destroy_session`). The first time it's executed with the mountpoint before the switch, the second time with the new mountpoint.

This has a result that line 4394 of janus_streaming.c is also excuted twice.
>janus_refcount_decrease(&session->ref);

The second time the refcount will decrease to 0, the free function is called, and it will give a heap-use-after-free on line 4418.

lminiero:
> This is caused because `janus_streaming_hangup_media_internal` is executed twice

We have a fence in place:

	if(!g_atomic_int_compare_and_exchange(&session->hangingup, 0, 1))
		return;

that will prevent `janus_streaming_hangup_media_internal` to reach the point you mention twice in a row. As such, while the race condition you suggest may happen, I'm pretty sure it's not because of the callback being called twice.

TomFFF:
I'm able to reproduce the issue with this script (after creating 2 rooms in the streaming plugin config (10/11)) 
```
#!/bin/bash
SESSION_ID=`curl --data '{ "janus": "create", "transaction": "tester.py"}' http://127.0.0.1:8088/janus 2>&1 | grep 'id' | sed 's/.*: //'`;
HANDLE_ID=`curl --data '{ "janus": "attach","plugin": "janus.plugin.streaming","transaction": "test-attach" }'  http://127.0.0.1:8088/janus/$SESSION_ID  2>&1 | grep '"id"' | sed 's/.*: //'`
[...]

PR #2115, opened 2020-04-28
New session mutex in Streaming plugin (see #2106)

lminiero:
First attempt at addressing the race condition described in #2106. Apparently, the cause is always the `session->mountpoint` changing from when we check it to when we use it, which is problematic when we use that info to decrease references. This patch adds a new mutex to the `janus_streaming_session` structure, that we can use when updating or checking the `mountpoint` property in critical sections.

I didn't test this extensively: I'll let @atoppi do his magic with the script he had to replicate the issue. Hopefully that will solve the problem AND avoid regressions (e.g., deadlocks that may occur due to the way the new mutex works). Until then, consider this WIP.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -1198,6 +1198,7 @@ typedef struct janus_streaming_session {
 	gint64 last_spatial_layer[3];
 	int temporal_layer, target_temporal_layer;
 	gboolean stopping;
+	janus_mutex mutex;
 	volatile gint renegotiating;
 	volatile gint hangingup;
 	volatile gint destroyed;
@@ -2253,6 +2254,7 @@ void janus_streaming_create_session(janus_plugin_session *handle, int *error) {
 	session->mountpoint = NULL;	/* This will happen later */
 	session->started = FALSE;	/* This will happen later */
 	session->paused = FALSE;
+	janus_mutex_init(&session->mutex);
 	g_atomic_int_set(&session->destroyed, 0);
 	g_atomic_int_set(&session->hangingup, 0);
 	handle->plugin_handle = session;
@@ -2299,6 +2301,7 @@ json_t *janus_streaming_query_session(janus_plugin_session *handle) {
 	janus_mutex_unlock(&sessions_mutex);
 	/* What is this user watching, if anything? */
 	json_t *info = json_object();
+	janus_mutex_lock(&session->mutex);
 	janus_streaming_mountpoint *mp = session->mountpoint;
 	json_object_set_new(info, "state", json_string(mp ? "watching" : "idle"));
 	if(mp) {
@@ -2336,6 +2339,7 @@ json_t *janus_streaming_query_session(janus_plugin_session *handle) {
 		}
 		janus_refcount_decrease(&mp->ref);
 	}
+	janus_mutex_unlock(&session->mutex);
 	json_object_set_new(info, "hangingup", json_integer(g_atomic_int_get(&session->hangingup)));
 	json_object_set_new(info, "destroyed", json_integer(g_atomic_int_get(&session->destroyed)));
 	janus_refcount_decrease(&session->ref);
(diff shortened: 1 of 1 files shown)


atoppi:
Kudos to @TomFFF for the script.
I've been running 4 of them in the last 30 minutes, no locks or crashes detected so far.
So lgtm :+1:

TomFFF:
Thanks @atoppi  and @lminiero  for the quick fix !

I'm also testing it and I don't see any problems so far

lminiero:
Good to know, thanks for the feedback! I'll let you test it for a bit longer, then, and if everything is still ok by tomorrow morning I'll merge :+1:

lminiero:
@TomFFF any update with your tests in the past few hours? Can we merge this?

TomFFF:
For me it looks good ! Couldn't reproduce it anymore. So I think it can be
 merged and the issue closed.

lminiero:
Ack, merging then :+1:

TomFFF:
@lminiero  , not sure if you should continue here or open a new issue ?

But we found a deadlock in the streaming plugin caused by this change.

When there is no more data in the `janus_streaming_relay_thread`  there is some `Notify users this mountpoint is done ` code

This takes a lock on  the mountpoint mutex first, and then a lock on each session of the viewer (L7893)

https://github.com/meetecho/janus-gateway/blob/0f6849e8cb58194275cf3f21f3161b01f0432ed9/plugins/janus_streaming.c#L7878-L7893

At the same time` janus_streaming_hangup_media` is executed :

First a lock on sessions->mutex (so new requests that need it will wait)
https://github.com/meetecho/janus-gateway/blob/0f6849e8cb58194275cf3f21f3161b01f0432ed9/plugins/janus_streaming.c#L4375-L4376

And in` janus_streaming_hangup_media_internal `first a lock on the `session->mutex` and after that on` mp->mutex`

https://github.com/meetecho/janus-gateway/blob/0f6849e8cb58194275cf3f21f3161b01f0432ed9/plugins/janus_streaming.c#L4405-L4408

So if you are very unlucky  both threads will wait at each other, and because there is a lock on `&sessions->mutex` the entire server is more or less 'broken'. 

Stacktrace : https://pastebin.com/i7JCr5EU

I don't have a case with lock debug on, but if you prefer it I can try to reproduce it (But I guess that I will need to add a sleep between the locks to trigger the deadlock, otherwise it looks impossible to hit it)

lminiero:
Ok, I'll look into it.

lminiero:
Unfortunately hangup_media is not the only place where we lock the session first and the mountpoint after that: it also happens in the management of many requests (watch, switch, etc.). As such, the right fix is probably to change the only where we lock the mountpoint first, that seems to be the notification you mentioned.

That said, it might not be that simple, considering we work on a mountpoint list and that gives us the sessions we need. Locking the mountpoint just to clone the list of viewers, and then unlock to iterate on the copy might work, since that copy wouldn't need to be protected by the mountpoint lock; at the same time, it might cause reference issues for sessions unless we make sure the copy doesn't address its own references too. I'll prepare a PR for you to test: not sure it will be ready today, more likely tomorrow.

lminiero:
I was wrong: the same swapped logic happens in "destroy" too. I'll have to think about how to fix things there too.

lminiero:
@TomFFF please test the PR above. I'm of course interested in fixing the new deadlock, but even more in ensuring we're not causing the original issue you opened again. The patch should ensure we always lock the mountpoint first, and then the session, but since I didn't have time to test there may still be issues. Looking forward to your feedback.

TomFFF:
Thx again for the fast reply !
I will start testing it now and give some feedback tomorrow

TomFFF:
Did some 'stress' tests during the night on a test server and couldn't reproduce any issue.

I will start using this version on our real servers.

lminiero:
@TomFFF any updates on the production tests? Thanks!

TomFFF:
Still running without any problems so far. So it looks ok for me

lminiero:
Ack, I'll merge then :+1: Thanks for spotting the new issue and for the help testing the fix!

PR #2124, opened 2020-04-30
Enable / disable recording while conference is in progress

wheresjames:
This commit adds a command to the videoroom plugin called 'enable_recording' which can be used to enable or disable recording while the conference is in progress.  Use is thus...

>    {
>        "request" : "enable_recording",
>        "room" : <unique numeric ID of the room>,
>        "secret" : "<room secret; mandatory if configured>"
>        "record" : <true|false, whether this publisher should be recorded or not; optional>,
>        "filename" : "<if recording, the base path/file to use for the recording files; optional>",
>    }

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3562,6 +3562,7 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 #else
 		g_async_queue_push(handle->queued_packets, &janus_ice_start_gathering);
 #endif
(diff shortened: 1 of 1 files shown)


lminiero:
@wheresjames I'm not going to look at something with no description. Please take the time to document what you're doing, why, the changes to the API, etc.

wheresjames:
Will do, sorry about that.

PR #2125, opened 2020-04-30
Off by one error causing missed rtp extensions

wheresjames:
This fixes the last extension header being missed.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -1070,7 +1070,7 @@ static int janus_pp_rtp_header_extension_find(char *buf, int len, int id,
 		janus_pp_rtp_header_extension *ext = (janus_pp_rtp_header_extension *)(buf+hlen);
 		int extlen = ntohs(ext->length)*4;
 		hlen += 4;
-		if(len > (hlen + extlen)) {
+		if(len >= (hlen + extlen)) {
 			/* 1-Byte extension */
 			if(ntohs(ext->type) == 0xBEDE) {
 				const uint8_t padding = 0x00, reserved = 0xF;


lminiero:
Thanks, I'll check later.

lminiero:
Not sure this can be accepted. `len` (packet len) is only equal to `hlen` (header length) plus `extlen` (length of extensions block) if the packet has no payload. We dump all packets with no payload anyway.

atoppi:
Hi @wheresjames ,
could you provide a binary sample of a RTP packet that is not being correctly parsed?
You can use the Janus pcap feature to get it.

wheresjames:
@atoppi , Absolutely, and thanks for your time guys.  I'm attaching a video that demonstrates the issue.  Run the following and you should see it doesn't pick up the rotation header without the patch.

> ./janus-pp-rec rottest3.mjr rottest3.mp4 -v 4

[rottest3.zip](https://github.com/meetecho/janus-gateway/files/4574251/rottest3.zip)

atoppi:
We need the mjr file, that zip only contains the mp4.

wheresjames:
@atoppi , Uh, zipped the wrong file

[rottest4.zip](https://github.com/meetecho/janus-gateway/files/4574574/rottest4.zip)

atoppi:
Hi @wheresjames ,
we have just fixed the root issue in [this commit](https://github.com/meetecho/janus-gateway/commit/cf636b43e3c07c0afced6023c57ebe6a11c7d746).
I'm closing the PR, thanks a lot for reporting and helping.

wheresjames:
Works perfectly, thanks!

Issue #2126, opened 2020-04-30
An invalid plugin handle can result in a dangling stream


menelike:
### Problem

I have seen that in conjunction with a very flaky internet connection, streams did not become cleaned up, with the result, that the camera was still running e.g. the lights were on.

### Analysis

After debugging the code here are my findings:

1. While calling `getUserMedia` like in [here](https://github.com/meetecho/janus-gateway/blob/d3ba04960d82403045bab2887b938f3f8a6965c1/html/janus.js#L2373-L2380)

```JS
navigator.mediaDevices.getUserMedia(gumConstraints)
  .then(function(stream) {
    pluginHandle.consentDialog(false);
    streamsDone(handleId, jsep, media, callbacks, stream);
  }).catch(function(error) {
  pluginHandle.consentDialog(false);
  callbacks.error({code: error.code, name: error.name, message: error.message});
});
```
the plugin handle might have become invalid in the meantime.

2. When then `streamsDone(handleId, jsep, media, callbacks, stream);` is called (https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L1630-L1634)
```JS
// WebRTC stuff
function streamsDone(handleId, jsep, media, callbacks, stream) {
  var pluginHandle = pluginHandles[handleId];
  if(!pluginHandle || !pluginHandle.webrtcStuff) {
    Janus.warn("Invalid handle");
    callbacks.error("Invalid handle");
    return;
  }
```

it will call the error callback `callbacks.error("Invalid handle");`

But at this point, there is no way to stop the tracks of that stream, even not with an appropriate error callback, since the stream is not included in the args. This leaves the stream in a dangling state.

### Possible solution

I have added a cleanup routine before calling `callbacks.error("Invalid handle");`:

```JS
// WebRTC stuff
function streamsDone(handleId, jsep, media, callbacks, stream) {
  var pluginHandle = pluginHandles[handleId];
  if(pluginHandle === null || pluginHandle === undefined ||
    pluginHandle.webrtcStuff === null || pluginHandle.webrtcStuff === undefined) {
    Janus.warn("Invalid handle");
    if (stream) {
      try {
        // Try a MediaStreamTrack.stop() for each track
        var tracks = stream.getTracks();
        for(var mst of tracks) {
          Janus.log(mst);
          if(mst)
            mst.stop();
        }
      } catch(e) {
        // Do nothing if this fails
      }
    }
    callbacks.error("Invalid handle");
    return;
  }
```
[...]

PR #2134, opened 2020-05-04
stop all tracks when streamsDone fails

menelike:
This fixes https://github.com/meetecho/janus-gateway/issues/2126

The goal of this PR is to stop tracks only in cases where Janus initially created the stream (e.g. `getUserMedia()` or `getDisplayMedia()`). Another option would be to stop all tracks within `streamsDone()` itself, but this might stop tracks in cases where it is not desired (e.g. externally passed streams).

This PR needs to be properly reviewed as I don't exactly know in which cases tracks should or should not be stopped. 

I am not happy the way I solved it (overriding `callbacks.error()`), maybe there is a cleaner way, feedback is highly appreciated.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1629,6 +1629,8 @@ function Janus(gatewayCallbacks) {
 		var pluginHandle = pluginHandles[handleId];
 		if(!pluginHandle || !pluginHandle.webrtcStuff) {
 			Janus.warn("Invalid handle");
+			// close all tracks if the given stream has been created internally
+			if (!callbacks.stream) stopAllTracks(stream);
 			callbacks.error("Invalid handle");
 			return;
 		}
@@ -3083,6 +3085,20 @@ function Janus(gatewayCallbacks) {
 		Janus.error("WebRTC error:", error);
 	}
 
+	function stopAllTracks(stream) {
+		try {
+			// Try a MediaStreamTrack.stop() for each track
+			var tracks = stream.getTracks();
+			for(var mst of tracks) {
+				Janus.log(mst);
+				if(mst)
+					mst.stop();
+			}
+		} catch(e) {
+			// Do nothing if this fails
+		}
+	}
+
 	function cleanupWebrtc(handleId, hangupRequest) {
 		Janus.log("Cleaning WebRTC stuff");
 		var pluginHandle = pluginHandles[handleId];
@@ -3130,19 +3146,9 @@ function Janus(gatewayCallbacks) {
 			config.bitrate.tsnow = null;
 			config.bitrate.tsbefore = null;
 			config.bitrate.value = null;
-			try {
-				// Try a MediaStreamTrack.stop() for each track
-				if(!config.streamExternal && config.myStream) {
-					Janus.log("Stopping local stream tracks");
-					var tracks = config.myStream.getTracks();
-					for(var mst of tracks) {
-						Janus.log(mst);
-						if(mst)
-							mst.stop();
-					}
-				}
-			} catch(e) {
-				// Do nothing if this fails
+			if(!config.streamExternal && config.myStream) {
+				Janus.log("Stopping local stream tracks");
+				stopAllTracks(config.myStream);
 			}
 			config.streamExternal = false;
 			config.myStream = null;
--- a/html/janus.js
+++ b/html/janus.js
@@ -86,6 +86,20 @@ var defaultExtension = {
 	}
 };
 
+function stopAllTracks (stream) {
+	try {
+		// Try a MediaStreamTrack.stop() for each track
+		var tracks = stream.getTracks();
+		for(var mst of tracks) {
+			Janus.log(mst);
+			if(mst)
+				mst.stop();
+		}
+	} catch(e) {
+		// Do nothing if this fails
+	}
+}
+
 Janus.useDefaultDependencies = function (deps) {
 	var f = (deps && deps.fetch) || fetch;
 	var p = (deps && deps.Promise) || Promise;
@@ -276,13 +290,7 @@ Janus.init = function(options) {
 						Janus.debug(devices);
 						callback(devices);
 						// Get rid of the now useless stream
-						try {
-							var tracks = stream.getTracks();
-							for(var mst of tracks) {
-								if(mst)
-									mst.stop();
-							}
-						} catch(e) {}
+						stopAllTracks(stream);
 					});
 				})
 				.catch(function(err) {
@@ -2088,17 +2096,7 @@ function Janus(gatewayCallbacks) {
 			if(media.update) {
 				if(config.myStream && config.myStream !== callbacks.stream && !config.streamExternal) {
 					// We're replacing a stream we captured ourselves with an external one
-					try {
-						// Try a MediaStreamTrack.stop() for each track
-						var tracks = config.myStream.getTracks();
-						for(var mst of tracks) {
-							Janus.log(mst);
-							if(mst)
-								mst.stop();
-						}
-					} catch(e) {
-						// Do nothing if this fails
-					}
+					stopAllTracks(config.myStream);
 					config.myStream = null;
 				}
 			}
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @menelike! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

html/janus.js
```diff
@@ -1864,6 +1864,29 @@ function Janus(gatewayCallbacks) {
 		callbacks = callbacks || {};
 		callbacks.success = (typeof callbacks.success == "function") ? callbacks.success : Janus.noop;
 		callbacks.error = (typeof callbacks.error == "function") ? callbacks.error : webrtcError;
+
+		function stopTracksOnError(stream, callbacks) {
```
menelike, line 1868: do we need to pass callbacks? Shall we just take `callbacks.error` from above directly?

html/janus.js
```diff
@@ -1864,6 +1864,29 @@ function Janus(gatewayCallbacks) {
...
+					var tracks = stream.getTracks();
+					for(var mst of tracks) {
+						Janus.log(mst);
+						if(mst)
+							mst.stop();
+					}
+				} catch(e) {
+					// Do nothing if this fails
+				}
+				if (typeof callbacks.error == "function") {
```
menelike, line 1882: at this point `callbacks.error` is always a function because of `callbacks.error = (typeof callbacks.error == "function") ? callbacks.error : webrtcError;` above. This is just a additional safety check.

lminiero:
Thanks, I missed the original issue as we had a national holiday here in Italy right before the weekend so was off the grid for a while. I'll review the issue and your PR (including your questions) later.

menelike:
No worries, this is not a critical bug, take your time. 
Thanks a lot for your work in general, I know that your time is precious.

atoppi:
I'd rather handle the error in `streamsDone`.
IIUC in case of external streams `webrtcStuff` is valid and  `webrtcStuff.streamExternal` is set to true.

atoppi:
@menelike have you considered my suggestion? Am I missing something?

menelike:
@atoppi Thanks a lot for your review and sorry for my late response, I was neck-deep in work.

Using `pluginHandle.webrtcStuff` or `streamExternal` is a solid solution, I was not aware of those flags. I'll update this PR over the weekend.

menelike:
One question though, does it make sense to wrap `callbacks.error` since it is being used multiple times within `streamsDone()` or shall we just consider stopping all tracks only in the case of 

```JS
		if(!pluginHandle || !pluginHandle.webrtcStuff) {
			Janus.warn("Invalid handle");
			callbacks.error("Invalid handle");
			return;
		}
```

atoppi:
Wrapping `callbacks.error` sounds more like a hack to me. We should try to intercept any scenario that lead to the issue.
We could start by handling the tracks stopping in that branch and then test if that is enough.

menelike:
Sorry for the long delay. I just tried to implement your suggestions @atoppi 

After further inspection I am not sure if it will work only in the scope of an external stream.
To my understanding the only way to determine if a given stream is internally created we must look for `!pluginHandle.webrtcStuff.streamExternal`.

The problem is that in 

```JS
		if(!pluginHandle || !pluginHandle.webrtcStuff) {
			Janus.warn("Invalid handle");
			callbacks.error("Invalid handle");
			return;
		}
```

we never have a `pluginHandle`, so stopping tracks in that if clause does not allow you to determine the state of `streamExternal`, hence tracks are closed for internal and external streams. I am not sure if that is desired.

atoppi:
`callbacks` should contain the property `stream` in case of external streams provided:

```javascript
		if(!pluginHandle || !pluginHandle.webrtcStuff) {
			Janus.warn("Invalid handle");
                        if (!callbacks.stream)  {
                           /* This does not involve external streams */
                           /* CLEANUP media tracks */
                        }
			callbacks.error("Invalid handle");
			return;
		}

**Comment by menelike** [comment]
@atoppi hope it fits now. I've made a manual test, the fix works on my side.

**Comment by atoppi** [comment]
@menelike since you are wrapping track closing in a dedicated function, would you mind replacing  replicated code like [here](https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L279) and [here](https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L2112) with a call to `stopAllTracks` ?

**Comment by menelike** [comment]
I was thinking about that as well, but I was afraid this would have exceeded the goal of this PR. I am happy you came to the same conclusion, will do that!

File: html/janus.js [hunk 3]
```diff
@@ -86,6 +86,20 @@ var defaultExtension = {
 	}
 };
 
+function stopAllTracks (stream) {
```
menelike, line 89: Had to move this up, otherwise, it would not have been able to be called from `Janus.init()`.
Not sure where to put this, if something else is desired please let me know.

menelike:
@atoppi Hope this is fine, let me know if you want any changes.
Thanks a lot for your feedback, highly appreciated!

html/janus.js
atoppi: You're right, I didn't think about that.
Moving the function declaration at the top sounds not good to me.

IMHO it's better to revert the for loop stopping the tracks in the `Janus.init()` and declare `stopAllTracks` in the scope of the `Janus` function (like it was before your last edit). Then re-use `stopAllTracks` whenever its declaration is in scope.

atoppi:
@menelike reviewed the changes.
Thanks for the effort.

atoppi:
@menelike did you miss my review comment by any chance?
Once resolved the last point, we'll be able to merge.

lminiero:
@menelike ping :slightly_smiling_face:

menelike:
@atoppi @lminiero 

Sorry guys for the late reply, I had a hard time managing the `work`-`life`-`family`-`opensource projects` balance. Hopefully, the changes are sufficient, if not, changes will be pushed immediately next time.

atoppi:
Thanks a lot @menelike, it LGTM now.

@lminiero don't forget to merge by squashing

html/janus.js
```diff
@@ -1624,11 +1624,27 @@ function Janus(gatewayCallbacks) {
 		});
 	}
 
+	function stopAllTracks (stream) {
```
lminiero, line 1627: Not sure why this has to be a private function. Can't this be sort of an object function that can be used in other contexts? Something like:

    Janus.stopAllTracks = function(stream) {

This way it can be used easily wherever it's needed, whether it's `janus.js` or the application itself.

html/janus.js
```diff
@@ -1624,11 +1624,27 @@ function Janus(gatewayCallbacks) {
...
+		}
+	}
+
 	// WebRTC stuff
 	function streamsDone(handleId, jsep, media, callbacks, stream) {
 		var pluginHandle = pluginHandles[handleId];
 		if(!pluginHandle || !pluginHandle.webrtcStuff) {
 			Janus.warn("Invalid handle");
+			// close all tracks if the given stream has been created internally
+			if (!callbacks.stream) stopAllTracks(stream);
```
lminiero, line 1647: Code style nits:

1. Comments should start with a capital letter;
2. Please no spaces between `if` and bracket;
3. Please put condition and action on different lines.

lminiero:
Thanks for the quick fixes! This looks good to me too, so I'l merge :v:

menelike:
@lminiero Thank you for your review. I've aligned the code as suggested and removed the last missing code from `Janus.init()` as well.

Regarding the linting rules, I looked through the code, did I miss the linting configuration? I'd like to avoid unnecessary work on your side the next time.

menelike:
I just saw the double space in `Janus.stopAllTracks  = function(stream) {`.

Damn. Shall I create a second PR for this?

lminiero:
We don't really have one, I should probably look into ways to make it automated, or at least provide guidance somewhere... I feel silly any time I mention those anyway :smile:

lminiero:
> Damn. Shall I create a second PR for this?

No need! We can take care of that ourselves :+1:

menelike:
Thanks a lot for the explanation! It has been a pleasure guys!

PR #2135, opened 2020-05-04
Adding support for cipher suite selection in websockets transport

agclark81:
This PR is modeled after the commits from #1219 and adds cipher suite selection for the websockets transport by setting 'ssl_cipher_list' on the 'lws_context_creation_info' struct to a 'ciphers' value in the websocket transport config. If this value isn't configured, a null is passed in and there is no behavior change. We've tested it with the cipher suites recommended by OWASP for secure protocols and everything is working as expected.

Example config:

ciphers = "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256";

### DIFF
--- a/conf/janus.transport.websockets.jcfg.sample
+++ b/conf/janus.transport.websockets.jcfg.sample
@@ -40,9 +40,10 @@ admin: {
 # Certificate and key to use for any secure WebSocket server, if enabled (and passphrase if needed).
 # You can also disable insecure protocols and ciphers by configuring the
 # 'ciphers' property accordingly (no limitation by default).
+# Examples of recommended cipher strings at https://cheatsheetseries.owasp.org/cheatsheets/TLS_Cipher_String_Cheat_Sheet.html
 certificates: {
 	#cert_pem = "/path/to/cert.pem"
 	#cert_key = "/path/to/key.pem"
 	#cert_pwd = "secretpassphrase"
-	#ciphers = "DEFAULT"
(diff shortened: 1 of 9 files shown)


conf/janus.transport.websockets.jcfg.sample
```diff
@@ -38,8 +38,11 @@ admin: {
 }
 
 # Certificate and key to use for any secure WebSocket server, if enabled (and passphrase if needed).
+# You can also disable insecure protocols and ciphers by configuring the
+# 'ciphers' property accordingly (no limitation by default).
 certificates: {
 	#cert_pem = "/path/to/cert.pem"
 	#cert_key = "/path/to/key.pem"
 	#cert_pwd = "secretpassphrase"
+	#ciphers = "DEFAULT"
```
lminiero, line 47: I'd probably set a reasonable default here, like the one we have in the HTTP config file; something that gives people an idea of what the syntax looks like. The example you made in the description may be too long for that, but something along those lines maybe.

agclark81: I've updated the example cipher string to the two ECDHE ciphers that are supported universally across all modern browsers, and I've linked out to the OWASP page with additional details on cipher strings and browser compatibility.

lminiero:
Thanks, merging! :+1:

PR #2136, opened 2020-05-05
small HTML fixes

willamowius:
I fixed a few small errors in the HTML structure, eg. double attributes or closing tags that need no closing.


lminiero:
Thanks! I'll have a look later.

lminiero:
Looks good, merging :+1:

PR #2137, opened 2020-05-05
Enable / disable recording while conference is in progress

wheresjames:
This PR adds a command to the video room plugin to enable / disable recording while a conference is in progress.  The command is formatted as follows.

> {
>     "request" : "enable_recording",
>     "room" : < unique numeric ID of the room >,
>     "secret" : "< room secret; mandatory if configured >"
>     "record" : < true|false, whether this publisher should be recorded or not; optional >,
>     "filename" : "< if recording, the base path/file to use for the recording files; optional >",
> }

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -4427,39 +4427,43 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 		janus_mutex_lock(&videoroom->mutex);
 		janus_mutex_unlock(&rooms_mutex);
 		/* Set recording status */
-		videoroom->record = recording_active ? TRUE : FALSE;
-		/* Iterate over all participants */
-		gpointer value;
-		GHashTableIter iter;
-		g_hash_table_iter_init(&iter, videoroom->participants);
-		while (g_hash_table_iter_next(&iter, NULL, &value)) {
-			janus_videoroom_publisher *participant = value;
-			if(participant && participant->session) {
-				janus_mutex_lock(&participant->rec_mutex);
-				gboolean prev_recording_active = participant->recording_active;
-				participant->recording_active = recording_active;
-				JANUS_LOG(LOG_VERB, "Setting record property: %s (room %"SCNu64", user %"SCNu64")\n", participant->recording_active ? "true" : "false", participant->room_id, participant->user_id);
-				/* Do we need to do something with the recordings right now? */
-				if(participant->recording_active != prev_recording_active) {
-					/* Something changed */
-					if(!participant->recording_active) {
-						/* Not recording (anymore?) */
-						janus_videoroom_recorder_close(participant);
-					} else if(participant->recording_active && participant->sdp) {
-						/* We've started recording, send a PLI/FIR and go on */
-						janus_videoroom_recorder_create(
-							participant, strstr(participant->sdp, "m=audio") != NULL,
-							strstr(participant->sdp, "m=video") != NULL,
-							strstr(participant->sdp, "m=application") != NULL);
-						if(strstr(participant->sdp, "m=video")) {
-							/* Send a FIR */
-							janus_videoroom_reqpli(participant, "Recording video");
+		gboolean room_prev_recording_active = recording_active ? TRUE : FALSE;
+		if (room_prev_recording_active != videoroom->record) {
+			/* Room recording state has changed */
+			videoroom->record = room_prev_recording_active;
+			/* Iterate over all participants */
+			gpointer value;
+			GHashTableIter iter;
+			g_hash_table_iter_init(&iter, videoroom->participants);
+			while (g_hash_table_iter_next(&iter, NULL, &value)) {
+				janus_videoroom_publisher *participant = value;
+				if(participant && participant->session) {
+					janus_mutex_lock(&participant->rec_mutex);
+					gboolean prev_recording_active = participant->recording_active;
+					participant->recording_active = recording_active;
+					JANUS_LOG(LOG_VERB, "Setting record property: %s (room %"SCNu64", user %"SCNu64")\n", participant->recording_active ? "true" : "false", participant->room_id, participant->user_id);
+					/* Do we need to do something with the recordings right now? */
+					if(participant->recording_active != prev_recording_active) {
+						/* Something changed */
+						if(!participant->recording_active) {
+							/* Not recording (anymore?) */
+							janus_videoroom_recorder_close(participant);
+						} else if(participant->recording_active && participant->sdp) {
+							/* We've started recording, send a PLI/FIR and go on */
+							janus_videoroom_recorder_create(
+								participant, strstr(participant->sdp, "m=audio") != NULL,
+								strstr(participant->sdp, "m=video") != NULL,
+								strstr(participant->sdp, "m=application") != NULL);
+							if(strstr(participant->sdp, "m=video")) {
+								/* Send a FIR */
+								janus_videoroom_reqpli(participant, "Recording video");
+							}
 						}
 					}
+					janus_mutex_unlock(&participant->rec_mutex);
 				}
-				janus_mutex_unlock(&participant->rec_mutex);
-			}
 		}
+        }
 		janus_mutex_unlock(&videoroom->mutex);
 		response = json_object();
 		json_object_set_new(response, "videoroom", json_string("success"));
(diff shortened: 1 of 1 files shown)


plugins/janus_videoroom.c
```diff
@@ -785,6 +785,20 @@ room-<unique room ID>: {
...
+ * To enable or disable recording while the conference is in progress,
+ * you can make use of the \c enable_recording request, which has to be
+ * formatted as follows:
+ *
+\verbatim
+{
+	"request" : "enable_recording",
+	"room" : <unique numeric ID of the room>,
+	"secret" : "<room secret; mandatory if configured>"
+	"record" : <true|false, whether this publisher should be recorded or not; optional>,
```
tgabi333, line 799: as i see the documentation is wrong, its not optional.

wheresjames:
@tgabi333 , Thanks, fixed.

antoniotuzzi:
great extension!
May I suggest a new one related ? (and you are very strong currently in the field!)
"rec_dir"
should be a PATH that if used is prepended to the "filename" (with correct final slash checks)

using "rec_dir" we can leave empty, for example,  "filename" in the configure, and leave to JANUS the naming tasks, but we avoid a common DIR for all recordings (to filter by, day, conversation ID etc etc)

wheresjames:
That's a great idea, I'll find time to add this.

lminiero:
> "record" : < true|false, whether this publisher should be recorded or not; optional >,

Not sure why a specific publisher is addressed here, if this is supposed to be a global recording thing.

wheresjames:
@lminiero , I changed that in that last patch, you are correct.

lminiero:
Ah sorry for missing that, I was just looking at the description, didn't have time to review the code yet :+1:

wheresjames:
@ antoniotuzzi , Sorry, I think I missed part of what you were saying, I think I get it now.  It doesn't make sense here to accept the filename parameter since it would have the effect of pointing all participants to the same file.  So that should be removed.

antoniotuzzi:
@wheresjames  you are right, correct

I confused with a single endpoint configure... :(

plugins/janus_videoroom.c
```diff
@@ -785,6 +785,19 @@ room-<unique room ID>: {
 		// Other publishers
 	]
 }
+\endverbatim *
+ *
+ * To enable or disable recording while the conference is in progress,
```
lminiero, line 790: This should clarify that it's a way to start/stop recording all participants in a bulk. We have ways to enable/disable recordings individually for different participants (which is still the preferred way for us since it allows us to use different paths, and what we use in production), and the sentence above creates ambiguity.

plugins/janus_videoroom.c
```diff
@@ -785,6 +785,19 @@ room-<unique room ID>: {
...
+ * you can make use of the \c enable_recording request, which has to be
+ * formatted as follows:
+ *
+\verbatim
+{
+	"request" : "enable_recording",
+	"room" : <unique numeric ID of the room>,
+	"secret" : "<room secret; mandatory if configured>"
+	"record" : <true|false, whether this room should be recorded or not>,
```
lminiero, line 799: I'd say "whether participants in this room should be automatically recorded or not".

plugins/janus_videoroom.c
```diff
@@ -4387,6 +4406,65 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+		error_code = janus_videoroom_access_room(root, FALSE, TRUE, &videoroom, error_cause, sizeof(error_cause));
+		if(error_code != 0) {
+			JANUS_LOG(LOG_ERR, "Failed to access videoroom\n");
+			janus_mutex_unlock(&rooms_mutex);
+			goto prepare_response;
+		}
+		janus_mutex_lock(&videoroom->mutex);
+		janus_mutex_unlock(&rooms_mutex);
+		/* Set recording status */
+		videoroom->record = recording_active ? TRUE : FALSE;
```
lminiero, line 4430: I feel this shouldn't update the value right away, but instead check what the value was before and do nothing if the new one is the same. Unless if, e.g., `videoroom->record` was `FALSE`, and there were individual recordings taking place, you want to still have them stopped anyway?

plugins/janus_videoroom.c
```diff
@@ -4387,6 +4406,65 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+					/* Something changed */
+					if(!participant->recording_active) {
+						/* Not recording (anymore?) */
+						janus_videoroom_recorder_close(participant);
+					} else if(participant->recording_active && participant->sdp) {
+						/* We've started recording, send a PLI/FIR and go on */
+						janus_videoroom_recorder_create(
+							participant, strstr(participant->sdp, "m=audio") != NULL,
+							strstr(participant->sdp, "m=video") != NULL,
+							strstr(participant->sdp, "m=application") != NULL);
```
lminiero, line 4453: I know this is in the original VideoRoom code, but looking at it now I think that's a broken way of detecting whether the publisher is sending that media. The stream may have been rejected, or be inactive. Anyway, just saying it out loud as a reminder to myself, nothing that should really concern you here.

lminiero:
@wheresjames any update on my notes? If you don't address them I can't merge this...

wheresjames:
@lminiero , Sorry for the delay I've been swamped.  Getting on this now.

plugins/janus_videoroom.c
wheresjames: Agree, updating.

plugins/janus_videoroom.c
wheresjames: I'll make this change.  Currently, it ensures the users recording state, which is interesting.  But I suppose an out-of-sync user is a sign of another issue.

lminiero: I wouldn't call it an issue: currently participants are able to individually change the state of their recording via `configure` messages, so a state where everyone except 1-2 people are being recorded can happen. In the future we may want to add an optional room property to prevent users from changing their recording state, and only allow it through the global feature you're adding now: this would ensure this desync cannot happen.

wheresjames: Ah, true, great point.

plugins/janus_videoroom.c
wheresjames: I will leave this then.

plugins/janus_videoroom.c
```diff
@@ -4387,6 +4406,69 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
 		json_object_set_new(response, "rtp_forwarders", list);
 		goto prepare_response;
+	} else if(!strcasecmp(request_text, "enable_recording")) {
+		JANUS_VALIDATE_JSON_OBJECT(root, record_parameters,
+			error_code, error_cause, TRUE,
+			JANUS_VIDEOROOM_ERROR_MISSING_ELEMENT, JANUS_VIDEOROOM_ERROR_INVALID_ELEMENT);
+		if(error_code != 0)
+			goto prepare_response;
+		json_t *record = json_object_get(root, "record");
+		int recording_active = json_is_true(record);
```
lminiero, line 4416: Why `int`? You can use `gboolean` right away.

plugins/janus_videoroom.c
```diff
@@ -4387,6 +4406,69 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+		error_code = janus_videoroom_access_room(root, FALSE, TRUE, &videoroom, error_cause, sizeof(error_cause));
+		if(error_code != 0) {
+			JANUS_LOG(LOG_ERR, "Failed to access videoroom\n");
+			janus_mutex_unlock(&rooms_mutex);
+			goto prepare_response;
+		}
+		janus_mutex_lock(&videoroom->mutex);
+		janus_mutex_unlock(&rooms_mutex);
+		/* Set recording status */
+		gboolean room_prev_recording_active = recording_active ? TRUE : FALSE;
```
lminiero, line 4430: No need for that, `recording_active` is a boolean.

wheresjames: You are correct, I think I copied it that way.  Pushed a fix.

lminiero:
This looks good to me: I'll add the new property to property to prevent self-recording myself, after I merge :+1: 
Thanks again for your contribution and the quick fixes!

wheresjames:
Thank you sir!

lminiero:
See the `lock_record` property introduced in the commit above for a way to ensure `enable_recording` is the only way to mess with recordings.

PR #2141, opened 2020-05-05
Fix to rare deadlock in Streaming plugin (see #2115)

lminiero:
See https://github.com/meetecho/janus-gateway/pull/2115#issuecomment-623874864 for context. Only partly tested, so WIP until we can confirm it fixes the issue with no regression to #2115 itself.


lminiero:
Merging.

PR #2144, opened 2020-05-06
Added new tool to convert .pcap captures to .mjr recording

lminiero:
What the title says. It's basically the reverse of the `mjr2pcap` we added in #1718, that I implemented since I needed it for some internal use cases, but can actually be useful in other scenarios as well. It depends on `libpcap`, so if the library is unavailable the tool will not be built. It's currently hardcoded to IPv4/UDP parsing, so if the pcap capture contains something else (e.g., RTP packet on IPv6) at the moment those packets will simply be ignored. In the future we can make it more flexible, but for the time being this should cover most of the typical captures.

Unlike `mjr2pcap`, `pcap2mjr` expects two additional command line arguments:

1. the codec to write in the .mjr header (e.g., `opus`)
2. the SSRC to filter (since a .pcap file may contain several RTP sessions and we only need one)

Example:

    pcap2mjr -c opus -s 12345678 file.pcap file.mjr

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ docs/html/
 /janus-cfgconv
 /janus-pp-rec
 /mjr2pcap
+/pcap2mjr
 /plugins/*.so
 /transports/*.so
 /events/*.so
@@ -53,6 +54,8 @@ Makefile.in
 /postprocessing/*.o
 /postprocessing/pp-cmdline.c
 /postprocessing/pp-cmdline.h
+/postprocessing/p2m-cmdline.c
+/postprocessing/p2m-cmdline.h
 /fuzzers/*.a
 /fuzzers/*.o
 /fuzzers/out
--- a/Makefile.am
+++ b/Makefile.am
@@ -560,6 +560,9 @@ endif
 if ENABLE_POST_PROCESSING
 bin_PROGRAMS += janus-pp-rec
 bin_PROGRAMS += mjr2pcap
+if ENABLE_PCAP2MJR
+bin_PROGRAMS += pcap2mjr
+endif
 
 janus_pp_rec_SOURCES = \
 	postprocessing/pp-cmdline.c \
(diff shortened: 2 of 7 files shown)


lminiero:
Added basic support to IPv6 and Linux Cooked Captures. This feels complete to me, so I'll merge.

zyliu1985:
How to build the pcap2mjr tool? I have installed the libpcap and enabled post-processing, but there is no pcap2mjr geneated. There are only janus-pp-rec and mjr2pcap. Thank you.

`./configure --prefix=/opt/janus --disable-rabbitmq --disable-mqtt --enable-post-processing --disable-websockets --enable-pcap2mjr

checking for EVENTS... yes
checking for LOGGERS... yes
checking for npm... no
checking for POST_PROCESSING... yes
checking for PCAP... no
`

lminiero:
You need libpcap-dev(el). The log clearly says it can't find PCAP, so that's the cause.

zyliu1985:
Thanks for your answer. But I'm sure I have installed libpcap-dev by apt. 

> apt install libpcap-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
libpcap-dev is already the newest version (1.8.1-6ubuntu1.18.04.1).
0 upgraded, 0 newly installed, 0 to remove and 2 not upgraded.
> 

How can I check the libpcap is working correctly?
I wrote a lillte test code, it seems that libpcap is working.

`
#include <pcap.h>
#include <stdio.h>
int main()
{
    char errBuf[PCAP_ERRBUF_SIZE], * device;
    device = pcap_lookupdev(errBuf);
    if(device) {
        printf("success: device: %s\n", device);
    } else {
        printf("error: %s\n", errBuf);
    }
    return 0;
}
`

output:
# ./test 
success: device: docker0

Thank you.

zyliu1985:
I compiled and installed the libpcap by code. And the pkg checked is passed. And pcap2mjr is ok.

Thank you.

PR #2147, opened 2020-05-09
Only process mute events if a timer fired to avoid video flashing.

### LINKED ISSUE #2145: Janus.js handling of track muting/unmuting causes video flashing in Chrome
In janus.js, a media track's mute/unmute behavior is handled as follows:

```
event.track.onended = function(ev) {
	Janus.log("Remote track muted/removed:", ev);
	if(config.remoteStream) {
		config.remoteStream.removeTrack(ev.target);
		pluginHandle.onremotestream(config.remoteStream);
	}
};
event.track.onmute = event.track.onended;
event.track.onunmute = function(ev) {
	Janus.log("Remote track flowing again:", ev);
	try {
		config.remoteStream.addTrack(ev.target);
		pluginHandle.onremotestream(config.remoteStream);
	} catch(e) {
		Janus.error(e);
	};
};
```

Chrome signals `mute` and `unmute` events whenever video is temporarily suspended, such as when packet loss is occurring. The code above causes the video at the receiver to flash--disappear when the track is removed by `onmute`, and then re-appear when added back by `onunmute`. When high packet loss occurs, flashing happens every few seconds.

Firefox does not exhibit this behavior, because it does not raise the mute and unmute events. (Instead, the video just freezes.)

Also, the problem occurs in Chrome with any videoroom settings--it doesn't matter whether `transport_wide_cc_ext` is enabled or disabled, whether simulcast is turned on or off, etc.

This is straightforward to duplicate in the janus demo room. (We're using clumsy to inject packet loss.)

I changed `onmute` to use a 5-sec timeout before removing the track (and made corresponding changes in `onunmute`), and the problem went away.

alexamirante:
In the last couple of days I happened to have a crappy network as my ISP is having troubles, so I tried to better study the video flashing issue described in #2145. 

By printing the time passed between `mute` and `unmute` events raised by Chrome and plotting them into a graph, I found out that this is always a multiple of 834 ms:
timing
I have no idea where this value comes from, but it's important to take it into account when setting the timeout @schoolcoder suggested. I rounded that value to 840 ms and set the timeout to three times that value (~2.5 seconds) which seems acceptable to me. This made most video flashing disappear in my crappy network.

I suck pretty bad at JavaScript, but it seemed easy enough to arrange a Pull Request.

Fixes #2145.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1784,7 +1784,7 @@ function Janus(gatewayCallbacks) {
 				};
 				event.track.onmute = function(ev) {
 					Janus.log("Remote track muted:", ev);
-					if(config.remoteStream) {
+					if(config.remoteStream && trackMutedTimeoutId == null) {
 						trackMutedTimeoutId = setTimeout(function() {
 							Janus.log("Removing remote track");
 							config.remoteStream.removeTrack(ev.target);
@@ -1799,6 +1799,7 @@ function Janus(gatewayCallbacks) {
 					Janus.log("Remote track flowing again:", ev);
 					if(trackMutedTimeoutId != null) {
 						clearTimeout(trackMutedTimeoutId);
+						trackMutedTimeoutId = null;
 					} else {
 						try {
 							config.remoteStream.addTrack(ev.target);
--- a/html/janus.js
+++ b/html/janus.js
@@ -1775,21 +1775,26 @@ function Janus(gatewayCallbacks) {
 				var trackMutedTimeoutId = null;
 				Janus.log("Adding onended callback to track:", event.track);
 				event.track.onended = function(ev) {
-					Janus.log("Remote track muted/removed:", ev);
+					Janus.log("Remote track removed:", ev);
 					if(config.remoteStream) {
 						clearTimeout(trackMutedTimeoutId);
+						config.remoteStream.removeTrack(ev.target);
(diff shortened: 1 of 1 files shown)


html/janus.js
```diff
@@ -1772,23 +1772,36 @@ function Janus(gatewayCallbacks) {
...
 				if(event.track.onended)
 					return;
+				var trackMutedTimeoutId = null;
 				Janus.log("Adding onended callback to track:", event.track);
 				event.track.onended = function(ev) {
 					Janus.log("Remote track muted/removed:", ev);
 					if(config.remoteStream) {
-						config.remoteStream.removeTrack(ev.target);
-						pluginHandle.onremotestream(config.remoteStream);
+						clearTimeout(trackMutedTimeoutId);
```
lminiero, line 1780: Shouldn't there be a `if(trackMutedTimeoutId != null) {` here as well?

alexamirante: calling `clearTimeout(null)` just does nothing:
> Passing an invalid ID to clearTimeout() silently does nothing; no exception is thrown.

so I don't think is actually needed.

alexamirante:
@atoppi double-check please

atoppi:
lgtm :+1:

lminiero:
Merging.

PR #2150, opened 2020-05-11
Add DSCP on RTP audio packets in SIP plugin

GerardM22:
To complete #2055, I propose to do the same thing in the RTP audio packets sent by the SIP plugin.
Tested OK on my implementation of Janus server.

### DIFF
--- a/conf/janus.plugin.sip.jcfg.sample
+++ b/conf/janus.plugin.sip.jcfg.sample
@@ -32,4 +32,11 @@ general: {
 
 	# Whether events should be sent to event handlers (default=true)
 	#events = false
+   
+	# If you need DSCP packet marking and prioritization, you can configure
+	# the 'dscp_audio_rtp' property to specific values, and Janus SIP plugin
+	# will set it on all outgoing audio RTP packets.
+	# No packet marking is done if this parameter is undefined or equal to 0
+	#dscp_audio_rtp = 46
+	
 }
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @GerardM22! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Why only audio?

GerardM22:
We only use audio in our Janus server but I can do the same thing for video.
I see that there are 4 types of stream in the SIP plugin:
- RTP audio
- RTP video
- RTCP audio
- RTCP video
Do you think that it can be useful to do it for RTCP too or only for RTP ?

lminiero:
No idea if RTCP should have the same value: if you know what's typical in the industry you can try doing the same. But I'd definitely add an option for RTP video too. And there should probably a check on the `setsockopt` call: if that fails there should be at least a warning in the logs.

lminiero:
Any chance you can do the same for the NoSIP plugin too, now that you're at it since they're "related" as plugins and media-wise do many things the same way? Thanks!

PS: please don't forget about signing the CLA, as we can't merge otherwise.

GerardM22:
I added DSCP setting on video RTP socket in the the SIP plugin and the check of setsockopt as you suggested but I tested only the audio.
I think its not necessary to do it for RTCP sockets.
I never looked at the NoSIP plugin, so I prefer not to touch it for the moment.

lminiero:
> I added DSCP setting on video RTP socket in the the SIP plugin and the check of setsockopt as you suggested but I tested only the audio.
> I think its not necessary to do it for RTCP sockets.

Thanks!

> I never looked at the NoSIP plugin, so I prefer not to touch it for the moment.

Fair enough, I'll add it myself then. This looks fine to me, so I'll merge, thanks again :+1:

PR #2155, opened 2020-05-16
Fix menus in html documentation when using Doxygen > 1.8.14

i8-pi:
A flag was added in Doxygen 1.8.14 to restore the old style menus. This change sets that flag

The doxygen config file was updated to the latest version (1.8.18) using `doxygen -u`. Older doxygen versions that don't recognize the new options will ignore them. I've tested 1.8.11 and 1.8.18

I've also removed the external jQuery reference and the included dynsections.js. We should be using the versions generated with the docs as different version have different bundled dependencies


januscla:
Thanks for your contribution, @i8-pi! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks, this Doxygen thing had been a thorn in our side for quite some time. I'll test and in case it works merge :+1:

configure.ac
```diff
@@ -470,10 +470,16 @@ AM_CONDITIONAL([ENABLE_DOCS], [test "x$enable_docs" = "xyes"])
...
 	AS_VERSION_COMPARE([$doxygen_version], [1.8.11],
-					   [],
-					   [],
-					   [AC_MSG_ERROR([Doxygen $doxygen_version too new: versions higher than 1.8.11 are known to render poorly.])]
-					  )
+                           [],
+                           [],
+                           [
+                            AS_VERSION_COMPARE([$doxygen_version], [1.8.14],
+                                               [AC_MSG_ERROR([Doxygen $doxygen_version not usable: versions between 1.8.12 and 1.8.14 are known to render poorly.])],
```
lminiero, line 477: Won't this only appear for version < 1.8.14, and not when it's exactly 1.8.14? If I got the check right, this means that for 1.8.14 this will be considered a success, go on, and since 1.8.14 is affected too, cause broken docs. If so, the check should probably be on 1.8.15? (just as we have 11 instead of 12 in the previous one).

i8-pi: Sorry, I did not explain the situation well. The flag to opt out of the new style menu was released in 1.8.14, so 1.8.14 works.

lminiero: Oh, perfect then, thanks! I was just testing this with the 1.8.15 on Fedora and it works perfectly, so I'll merge :+1:

PR #2157, opened 2020-05-18
Added support for data channel subprotocol

lminiero:
This is actually a subset of #1898, which is where we implemented support for real-time text to the SIP plugin using data channels. Per the standard, RTT on data channels requires a dedicated subprotocol, and since we didn't support it in Janus, that PR also implemented that part. That said, this feature doesn't really need to be tied to that effort, so I decided to make a separate PR instead, that we could merge sooner.

It's important to point out that the subprotocol is used when a new channel is created, and is not something you can change dynamically after a specific channel has been created. As such, you should always use it in conjunction with a label: this way, if the label doesn't exist, a new channel for it will be created, and the subprotocol you specified be assigned to it and used for all messages exchanged on it.

At the moment this only implements support for the feature, but no plugin really makes use of it. Anyway, it's easy to test with the EchoTest, since if you send data channel messages to that plugin with a specific label and opening a new channel with a specific subprotocol, it will use it for the response as well. As such, you can modify the echotest.js code like this:

```
function sendData() {
	var data = $('#datasend').val();
	if(data === "") {
		bootbox.alert('Insert a message to send on the DataChannel');
		return;
	}
	echotest.data({
		text: data,
		label: "new-channel",		// <--- let's use a custom channel, as protocol is
		protocol: "my-subprotocol",	// <--- only used when a new channel is opened
		error: function(reason) { bootbox.alert(reason); },
		success: function() { $('#datasend').val(''); },
	});
}
```

and you should see the response coming from the label you specified, and on the subprotocol used for the new channel. The same can be done from a plugin code as well, by properly setting the `protocol` in the `janus_plugin_data` structure:

```
janus_plugin_data data = {
	.label = "new-channel",
	.protocol = "my-subprotocol",
	.binary = FALSE,
	.buffer = text,
	.length = strlen(text)
};
gateway->relay_data(handle, &data);
```

### DIFF
--- a/dtls.c
+++ b/dtls.c
@@ -1052,10 +1052,10 @@ void janus_dtls_sctp_data_ready(janus_dtls_srtp *dtls) {
 	janus_ice_notify_data_ready(handle);
 }
 
-void janus_dtls_wrap_sctp_data(janus_dtls_srtp *dtls, char *label, gboolean textdata, char *buf, int len) {
+void janus_dtls_wrap_sctp_data(janus_dtls_srtp *dtls, char *label, char *protocol, gboolean textdata, char *buf, int len) {
 	if(dtls == NULL || !dtls->ready || dtls->sctp == NULL || buf == NULL || len < 1)
 		return;
-	janus_sctp_send_data(dtls->sctp, label, textdata, buf, len);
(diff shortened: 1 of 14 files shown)


lminiero:
Merging.

PR #2158, opened 2020-05-18
User talking

mirkobrankovic:
This PR is adding a per-user override of global room values of audio_active_packets and audio_level_average, since different users will have different microphone input levels (volume and gain).
It is less likely in practice to find settings of one user to fit another one.

@lminiero , if this makes sense and is implemented "just right" , let me know to add it to videoroom also.

Thanks for the feedback,
Mirko


plugins/janus_audiobridge.c
```diff
@@ -960,7 +962,9 @@ static struct janus_json_parameter join_parameters[] = {
 	{"prebuffer", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"quality", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"volume", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
-	{"secret", JSON_STRING, 0}
+	{"secret", JSON_STRING, 0},
+	{"audio_level_averge", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
```
lminiero, line 966: Typo, should be `audio_level_averge`.

lminiero:
It does make sense, thanks! I think adding it to the VideoRoom too would indeed help.

plugins/janus_audiobridge.c
mirkobrankovic: great thx :)

mirkobrankovic:
added to videoroom also, I haven't tested Configure request with this values, but I hope it will work if someone wants to change tem later after joining room

lminiero:
Thanks! I noticed a couple of small things that need to be fixed on code style, but I can take care of those myself. Merging :+1:

mirkobrankovic:
oh thanks :D

PR #2161, opened 2020-05-20
List private rooms if valid admin_key was provided.

robby2016:
This simple enhancement allows admins to list private rooms as well using the existing "list" request.
To make this fully backwards compatible, if no "admin_key" is provided the list gets returned like before (without the private rooms).

The same solution could be applied to other plugins like the audiobrige too.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -3551,12 +3551,33 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 			if(!room)
 				continue;
 			janus_refcount_increase(&room->ref);
-			if(room->is_private) {
-				/* Skip private room */
-				JANUS_LOG(LOG_VERB, "Skipping private room '%s'\n", room->room_name);
-				janus_refcount_decrease(&room->ref);
-				continue;
-			}
+      if(room->is_private) {
+        /* only if admin_key isset */
+        if(admin_key != NULL) {
+          json_t *admin_key_json = json_object_get(root, "admin_key");
+          /* verify admin_key was provided */
+          if(admin_key_json != NULL && strlen(json_string_value(admin_key_json)) > 0) {
+            JANUS_CHECK_SECRET(admin_key, root, "admin_key", error_code, error_cause,
+            JANUS_VIDEOROOM_ERROR_MISSING_ELEMENT, JANUS_VIDEOROOM_ERROR_INVALID_ELEMENT, JANUS_VIDEOROOM_ERROR_UNAUTHORIZED);
+            if(error_code != 0) {
+              JANUS_LOG(LOG_VERB, "No room list, wrong admin_key provided\n");
+              goto end_loop;
+            }
+          }
+          else {
+            /* Skip private room */
+            JANUS_LOG(LOG_VERB, "Skipping private room '%s'\n", room->room_name);
+            janus_refcount_decrease(&room->ref);
+            continue;
+          }
+        }
+        else {
+          /* Skip private room */
+          JANUS_LOG(LOG_VERB, "Skipping private room '%s'\n", room->room_name);
+          janus_refcount_decrease(&room->ref);
+          continue;
+        }
+      }
 			if(!g_atomic_int_get(&room->destroyed)) {
 				json_t *rl = json_object();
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @robby2016! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks for the contribution! Makes sense, and to apply it to other plugins as well. That said, please use tabs, not spaces: if your editor supports it, the repo has an `.editorconfig` file with a few rules.

plugins/janus_videoroom.c
```diff
@@ -3551,12 +3551,33 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+            janus_refcount_decrease(&room->ref);
+            continue;
+          }
+        }
+        else {
+          /* Skip private room */
+          JANUS_LOG(LOG_VERB, "Skipping private room '%s'\n", room->room_name);
+          janus_refcount_decrease(&room->ref);
+          continue;
+        }
```
lminiero, line 3579: Pre-checking the admin_key you also simplify this duplicate error management.

plugins/janus_videoroom.c
```diff
@@ -3546,16 +3546,32 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
-			if(room->is_private) {
-				/* Skip private room */
-				JANUS_LOG(LOG_VERB, "Skipping private room '%s'\n", room->room_name);
-				janus_refcount_decrease(&room->ref);
-				continue;
+			if(room->is_private && lock_room_list) {
+					/* Skip private room if no valid admin_key was provided */
+					JANUS_LOG(LOG_VERB, "Skipping private room '%s'\n", room->room_name);
+					janus_refcount_decrease(&room->ref);
+					continue;
```
lminiero, line 3574: Looks like you added an extra tab for the lines above?

plugins/janus_videoroom.c
```diff
@@ -3546,16 +3546,32 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+			json_t *admin_key_json = json_object_get(root, "admin_key");
+			/* Verify admin_key if it was provided */
+			if(admin_key_json != NULL && strlen(json_string_value(admin_key_json)) > 0) {
+				JANUS_CHECK_SECRET(admin_key, root, "admin_key", error_code, error_cause,
+					JANUS_VIDEOROOM_ERROR_MISSING_ELEMENT, JANUS_VIDEOROOM_ERROR_INVALID_ELEMENT, JANUS_VIDEOROOM_ERROR_UNAUTHORIZED);
+				if(error_code != 0) {
+					janus_mutex_unlock(&rooms_mutex);
+					goto prepare_response;
+				}
+				else {
```
lminiero, line 3560: Please put the`else` on the same line as the closing bracket (only code style, sorry for the bother).

robby2016:
Thank you very much for the great feedback. Of course! You're absolutely right, the check should happen outside of the loop. Hope this second version is better.

plugins/janus_videoroom.c
```diff
@@ -3546,16 +3546,32 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 		GHashTableIter iter;
 		gpointer value;
 		g_hash_table_iter_init(&iter, rooms);
+		gboolean lock_room_list = TRUE;
+		if(admin_key != NULL) {
+			json_t *admin_key_json = json_object_get(root, "admin_key");
+			/* Verify admin_key if it was provided */
+			if(admin_key_json != NULL && strlen(json_string_value(admin_key_json)) > 0) {
```
lminiero, line 3553: This should also have a check on `json_is_string(admin_key_json)`, as there's no guarantee `admin_key` is a string in the JSON (it's not part of the objects we use for validation in "list", since list requires no arguments).

robby2016:
Not a problem at all and sorry for the styles issues.

lminiero:
No problem, we don't really have these rules written anywhere... I probably should do that!
Thanks again, merging :+1:

lminiero:
FYI, I just added the same changes to other plugins with private resources as well.

robby2016:
Awesome! Thank you :)

PR #2164, opened 2020-05-22
SIP plugin: add audio/video stream with an update request(or reinvite)

ihusejnovic:
Hi,
this pull request is about adding option in SIP plugin to add audio/video stream with update request (or SIP reinvite) if it wasn’t initially negotiated.
Of course, for SIP reinvite it will work only if you don’t use autoaccept_reinvites feature.
I have tested it with Kamailio and it works fine. We are using it in production for last 5 days and we didn’t notice any issue.

The main change is allocating local media ports on update request if audio or video stream is added.
Also, if update request is answer(not offer) session->media.updated is set to TRUE because we need to connect new media ports.

Let me know if you have any question, suggestion or anything else. Thanks.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -5545,7 +5545,7 @@ void janus_sip_sdp_process(janus_sip_session *session, janus_sdp *sdp, gboolean 
 		temp = temp->next;
 	}
 
-	if(update && changed && *changed && (answer || session->media.autoaccept_reinvites)) {
+	if(update && changed && *changed) {
 		/* Something changed: mark this on the session, so that the thread can update the sockets */
 		session->media.updated = TRUE;
 		if(session->media.pipefd[1] > 0) {
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -1368,7 +1368,7 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 void janus_sip_sdp_process(janus_sip_session *session, janus_sdp *sdp, gboolean answer, gboolean update, gboolean *changed);
 char *janus_sip_sdp_manipulate(janus_sip_session *session, janus_sdp *sdp, gboolean answer);
 /* Media */
-static int janus_sip_allocate_local_ports(janus_sip_session *session);
+static int janus_sip_allocate_local_ports(janus_sip_session *session, gboolean update);
 static void *janus_sip_relay_thread(void *data);
 static void janus_sip_media_cleanup(janus_sip_session *session);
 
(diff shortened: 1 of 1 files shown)


plugins/janus_sip.c
```diff
@@ -5785,10 +5804,6 @@ static void janus_sip_connect_sockets(janus_sip_session *session, struct sockadd
 	if(!session || (!audio_server_addr && !video_server_addr))
 		return;
 
-	if(session->media.updated) {
```
ihusejnovic, line 5788: This is removed because session->media.updated is always false on this place.

lminiero:
Thanks! I won't have time to look into this today, I'll try to do that next week :+1:

plugins/janus_sip.c
```diff
@@ -3612,6 +3612,27 @@ static void *janus_sip_handler(void *data) {
...
+			gboolean audio_added = strstr(msg_sdp, "m=audio") && !strstr(msg_sdp, "m=audio 0") && session->media.local_audio_rtp_port == 0;
+			gboolean video_added = strstr(msg_sdp, "m=video") && !strstr(msg_sdp, "m=video 0") && session->media.local_video_rtp_port == 0;
+
+			if(audio_added)
+				session->media.has_audio = TRUE;	/* FIXME Maybe we need a better way to signal this */
+
+			if(video_added)
+				session->media.has_video = TRUE;	/* FIXME Maybe we need a better way to signal this */
+
+			if(audio_added || video_added) {
```
lminiero, line 3625: I think you can remove the three empty lines above, since they belong to the same group of code.

plugins/janus_sip.c
lminiero: Not sure why you think it will always be false there? Anyway, since it only showed an additional log line, no harm in getting rid of it, I guess.

lminiero:
One more note: I guess this won't work if the SIP session is using SDES-SRTP, since I think we only set that up at the beginning of the call, and I don't see crypto-related changes here.

lminiero:
Merging.

lminiero:
PS: looks like the NoSIP plugin already supports all that, and using the same additional `update` property in `janus_nosip_allocate_local_ports`: possibly what you took as inspiration for this patch? Anyway, less work for me :smile:

ihusejnovic:
Hello Lorenzo,
Sorry for the late reply, it's a holiday here. Thank you for review and merge.

Yes, I have implemented this based on the NoSIP plugin implementation.

When you asked me about SDES-SRTP I realized that I didn't test it because we are not using it.
I tried it now and it doesn't work when a new stream is added. I will create a new pull request to fix this.

PR #2167, opened 2020-05-22
Assign an address family when no interface is specified.

PauKerr:
There is a code path in the janus_streaming plugin/janus_streaming_create_fd() where a multicast mount point (mcast and port) is specified but no interface (iface) is specified that leaves the family set to zero. This causes the bind() call to fail because the incomplete address6 structure is passed when an IPv4 multicast address is the target.


januscla:
Thanks for your contribution, @PauKerr! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

alexamirante:
LGTM.

lminiero:
Thanks, merging then :+1:

PR #2171, opened 2020-05-25
Small improvements in the documentation of videoroom

ancorgs:
The first commit fixes what, in my opinion, is an error in the documentation. With the current implementation (which is more convenient than the documented one) you receive two events when a publisher leaves and only one when a lurker leaves.

The second one adds no new information (the same is implicitly said in other parts of the document), but I think it would be a useful note/reminder for those looking for a way to secure their connections.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -842,17 +842,21 @@ room-<unique room ID>: {
 }
 \endverbatim
  *
- * Other participants will receive a different event depending on whether
- * you were currently an active publisher ("unpublished") or simply
- * lurking ("leaving"):
+ * Other participants will receive a "leaving" event to notify them the
+ * circumstance:
  *
 \verbatim
 {
 	"videoroom" : "event",
 	"room" : <room ID>,
-	"leaving|unpublished" : <unique ID of the publisher who left>
+	"leaving : <unique ID of the participant who left>
 }
 \endverbatim
+ *
+ * If you were an active publisher, other users will also receive the
+ * corresponding "unpublished" event to notify them the stream is not longer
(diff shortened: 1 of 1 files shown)


lminiero:
It's only an error if you see that as a response to a `leave`, but I think the commit makes the `unpublished` event more confusing now. If as a participant you just unpublish (e.g., hangup the PeerConnection or send "unpublish", but not leave the room), everyone will get an `unpublished` error, NOT a `leaving` one, and they should NOT assume the user left: they're still very much there, just not sending anything. If you check the code, "unpublished" is sent when `is_leaving` is `FALSE` in `janus_videoroom_leave_or_unpublish`, which is what happens when `hangup_media` is called on the publisher.

ancorgs:
Well, as I understand it, the documentation currently says that:

1) if you are a lurker and you leave, the others only get a "leaving" event. 
2) if you are a publisher and you leave, the others only get an "unpublished" event

Which I find confusing since it seems to imply that in the second case the corresponding "leaving" event will never be received, which is not the case.

ancorgs:
I think I understand how it works (I checked the code before writing the pull request), but maybe I'm wrong. So let's check if I got it right:

If an active publisher sends a `leave` request, two things happen:
  - the media is hung up and, as a consequence, other participants receive an `unpublished` event
  - the session is destroyed and, as a consequence, other participants receive a `leaving` event

On the other hand, if a lurker sends a `leave` request, only the second part happens:
  - the session is destroyed and, as a consequence, other participants receive a `leaving` event

Isn't it?

lminiero:
Yep, that's correct :+1: 
Anyway, your changes do make sense after reading them again, especially considering that's in the "leave" section, so, merging!

PR #2174, opened 2020-05-26
Update janus.js

izemize:
Little fix. for send back to error callback detailed info (reason, +code). Without this modification, users only listen to string for example, Unauthorized request (wrong or missing secret/token)

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -909,7 +909,7 @@ function Janus(gatewayCallbacks) {
 				Janus.debug(json);
 				if(json["janus"] !== "success") {
 					Janus.error("Ooops: " + json["error"].code + " " + json["error"].reason);	// FIXME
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @izemize! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks, but this feels a bit like a breaking change. Some of our demos do intercept errors when trying to send synchronous requests, and the way they handle/display it right now will be broken by your fix.

lminiero:
@izemize any update on this? Have you checked if this can break the demos?

lminiero:
Closing since you disappeared. Feel free to reopen once you have time to address our feedback.

PR #2175, opened 2020-05-27
Fix streaming plugin demo page when string_ids is true

lionelnicolas:
When `string_ids` is set to `true` in streaming plugin configuration, watching a stream from the janus demos webpage fails with:

```
Invalid element type (id should be a string)
```

This commit makes the browser fallback to string instead of sending `null`.


lminiero:
Thanks, merging! :+1:

PR #2176, opened 2020-05-27
janus_videoroom: fix bad copy paste of codec name

tmatth:


lminiero:
Thanks for catching that, merging :+1:

PR #2184, opened 2020-05-28
Protect session callee accesses through session mutex in SIP plugin

atoppi:
This is to fix some race conditions discovered when freeing session callee.


lminiero:
Looks good to me :+1:

PR #2187, opened 2020-05-30
Fix if not building from the top directory (example : yocto)

tezcatli:
The file "janus.ggo" is presumed to be in the same directory than the build directory.   It's not true when you don't build janus in the top directory.    For example, the yocto build system builds packages in separate directories from the source.

This fix just adds $top_srcdir to reference janus.ggo.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -191,7 +191,7 @@ BUILT_SOURCES = cmdline.c cmdline.h version.c
 cmdline.h: cmdline.c
 
 cmdline.c: janus.c
-	gengetopt --set-package="janus" --set-version="$(VERSION)" < janus.ggo
+	gengetopt --set-package="janus" --set-version="$(VERSION)" < $(top_srcdir)/janus.ggo
 
 directory = .git
 dir_target = $(directory)-$(wildcard $(directory))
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @tezcatli! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
We have two other .ggo files generated that way, should they be updated too? Have you checked where the sources are generated after that? Sounds like if you specify a different source folder for the .ggo file, the sources would be put somewhere else, but I may be wrong.

tezcatli:
I haven't noticed the two other files since they are related to post processing tools and i don't use is this feature.  I've updated my pull request to fix also Makefile.am for those two file.  I've tried to build those tools and it works.

The generated source files are located in the build directory in respect of the tree hierarchy.  So generated files for janus.ggo are in <buil_dir> and  generated files for postprocessing/janus-pp-rec.ggo and  postprocessing/pcap2mjr.ggo are in <build_dir>/posprocessing

lminiero:
@tezcatli ack, I'll give it a go later. In the meanwhile, please sign the CLA as asked by the bot, as we can't merge this otherwise.

lminiero:
@tezcatli still waiting on the CLA. If you'd rather not sign, please let us know so that we don't keep this open needlessly. Thanks.

lminiero:
Thanks, merging :+1:

PR #2189, opened 2020-06-01
Don't put session to stopping in watch

TomFFF:
Caused by typo in this commit in 2a98ec28f5f1caad2ce163f71c0123ec42e39d53

```
			- session->stopping = FALSE;
			+ g_atomic_int_set(&session->stopping, 1);
```


lminiero:
Good catch, thanks, merging :+1:

PR #2190, opened 2020-06-01
Fix RTSP parsing

lionelnicolas:
This fixes two issues in streaming plugin when pulling an RTSP stream.

When the RTSP server is not sending `fmtp` in its SDP, values returned to browsers are wrong because of uninitialized variables :

```
a=fmtp:111 ��\�
```

When the RTSP server is sending a incomplete `fmtp` line, fmtp parameters where taken from the following line.

For example, if RTSP server sends :

```
(...)
a=fmtp:98
a=control:streamid=0
(...)
```

The streaming plugin was sending to the browser :

```
a=fmtp:98 a=control:streamid=0
```

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -6308,7 +6308,9 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 	int ka_timeout = 0;
 	int vpt = -1;
 	char vrtpmap[2048];
+	vrtpmap[0] = '\0';
 	char vfmtp[2048];
+	vfmtp[0] = '\0';
 	char vcontrol[2048];
 	char uri[1024];
 	char vtransport[1024];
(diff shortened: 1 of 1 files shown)


plugins/janus_streaming.c
```diff
@@ -6717,15 +6727,15 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
...
-		mp->codecs.audio_fmtp = doaudio ? g_strdup(afmtp) : NULL;
+		mp->codecs.audio_fmtp = doaudio && strlen(afmtp) ? g_strdup(afmtp) : NULL;
 	if(mp->codecs.video_pt == -1)
 		mp->codecs.video_pt = dovideo ? vpt : -1;
 	if(mp->codecs.video_rtpmap == NULL)
-		mp->codecs.video_rtpmap = dovideo ? g_strdup(vrtpmap) : NULL;
+		mp->codecs.video_rtpmap = dovideo && strlen(vrtpmap) ? g_strdup(vrtpmap) : NULL;
 	if(mp->codecs.video_fmtp == NULL)
-		mp->codecs.video_fmtp = dovideo ? g_strdup(vfmtp) : NULL;
+		mp->codecs.video_fmtp = dovideo && strlen(vfmtp) ? g_strdup(vfmtp) : NULL;
```
lminiero, line 6738: Just as a matter of readability, please group the condition in brackets in the 4 lines above.

lminiero:
Thanks! I just added a small note, and then it's good to merge for me.

lminiero:
Thanks, merging :+1:

PR #2195, opened 2020-06-02
make libcurl follow RTSP 302 redirections

lionelnicolas:
Follow redirects in case of RTSP server replies something like :

```
RTSP/1.0 302 Try Other Server
CSeq: 123
Location: rtsp://other-rtsp-server/stream-key
```


lminiero:
Thanks, merging :+1:

PR #2196, opened 2020-06-03
NoSIP plugin: Fixed SRTP-SDES for "process" request and session update

ihusejnovic:
Hi,
As I promised in my previous pull request here is the fix for SRTP-SDES in the NoSIP plugin.

Two issues are fixed:
- SRTP-SDES for session update. It's the same fix applied in the SIP plugin.
- "srtp" property was ignored for "process" requests. For example, if you send a "process" request with "srtp" set to “sdes_mandatory” and SDP that doesn’t contain any crypto attributes, although it should fail, it would be processed.

Also, I updated the demo page for NoSIP plugin for easier testing session update and SRTP-SDES.

p.s. I hope I will not promise a new fix in this pull request :)

### DIFF
--- a/html/nosiptest.html
+++ b/html/nosiptest.html
@@ -73,7 +73,11 @@
 					<div class="col-md-6">
 						<div class="panel panel-default">
 							<div class="panel-heading">
-								<h3 class="panel-title">Caller</h3>
+								<h3 class="panel-title">Caller
+									<div class="btn-group btn-group-xs pull-right">
+										<button class="btn btn-danger" autocomplete="off" id="togglevideo">Disable video</button>
+									</div>
+								</h3>
 							</div>
 							<div class="panel-body" id="videoleft"></div>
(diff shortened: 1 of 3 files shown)


lminiero:
Thanks a lot, this is really appreciated! I'll try to test this later today but from a quick glance this looks great, and much more work than I expected it to be.

> p.s. I hope I will not promise a new fix in this pull request :)

Hahaha, hope you didn't really feel compelled to do it! You're doing a lot for the project, and your contributions are always appreciated :slightly_smiling_face:

lminiero:
Looks good! I was a bit confused during the demo at first, because disabling video only hid the local one but not the remote one: then I remembered how I wrote my own demo, and that the remove video is actually coming from the callee, that is indeed still sending video... I verified through webrtc-internals that renegotiation seems to be happening correctly :v:

Merging :+1:

PR #2197, opened 2020-06-03
Don't keep session in paused when switching mountpoints

TomFFF:
Don't keep session in paused when switching mountpoints

Caused by typo in this commit in 2a98ec2
```
			- session->paused = FALSE;
			+ g_atomic_int_set(&session->paused, 1);
```


lminiero:
Good catch, thanks, merging :+1:

PR #2198, opened 2020-06-03
Update libnice link

ssalenik:
Current link 404s


tmatth:
This redirects to https://libnice.freedesktop.org/ for me, maybe that'd be preferable?

ssalenik:
Oups, didn't notice the redirect. Yes, I'll do that.

lminiero:
Thanks! I noticed we have the same (broken) link in `mainpage.dox`, so I'll update it there too.

PR #2201, opened 2020-06-08
Send username when using TURN REST API (fixes #2199)

### LINKED ISSUE #2199: TURN REST API request does not supply username
The janus_turnrest_request function in turnrest.c does not send a username. Which means it does not work with the classical turnhttp-server. Is there a chance to supply it?

lminiero:
See #2199 for context. Untested, so feedback welcome.

### DIFF
--- a/turnrest.h
+++ b/turnrest.h
@@ -68,7 +68,7 @@ void janus_turnrest_response_destroy(janus_turnrest_response *response);
  * @note Use janus_turnrest_response_destroy to get rid of the response, once done
  * @param[in] user Username to provide in the TURN REST API request
  * @returns A valid janus_turnrest_response instance, if successful, NULL otherwise */
-janus_turnrest_response *janus_turnrest_request(char *user);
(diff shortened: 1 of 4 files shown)


turnrest.h
```diff
@@ -66,8 +66,9 @@ void janus_turnrest_response_destroy(janus_turnrest_response *response);
 
 /*! \brief Retrieve address and credentials for one or more TURN servers
  * @note Use janus_turnrest_response_destroy to get rid of the response, once done
+ * @param[in] user Username to provide in the TURN REST API request
  * @returns A valid janus_turnrest_response instance, if successful, NULL otherwise */
-janus_turnrest_response *janus_turnrest_request(void);
+janus_turnrest_response *janus_turnrest_request(char *user);
```
tmatth, line 71: nit:  As `char *user` is read-only it might as well be `const char *user`

lminiero: Makes sense, I'll fix that, thanks :+1:

lminiero:
Merging.

PR #2203, opened 2020-06-08
fix muted timeout race condition

vahidov:
In our production we have a lot of errors like `Cannot read property 'removeTrack' of null`
Try to fix this


lminiero:
Thanks, merging :+1:

Issue #2204, opened 2020-06-08
Response to streaming plugin "list" command doesn't contain metadata 


willamowius:
The documentation says that the channel list of the streaming plugin returned by "list" contains the metadata.

Just tested against master, that it is not included.

Either the code or the documentation should be changed. I think changing the code would be better, because its handy to have the metadata right away.

lminiero:
> I think changing the code would be better, because its handy to have the metadata right away.

I agree. Don't have time to do that now though, so if you can share a pull request I'll gladly review.

willamowius:
I can do that.

I would propose to show the "enabled" field in the list, too.

lminiero:
Closing as both matters were addressed.

PR #2205, opened 2020-06-08
add metadata field to list reposnse, as documented

willamowius:
solves issue #2204


lminiero:
There should be an if, though, because the metadata string is not always set.

lminiero:
Thanks, merging :+1:

PR #2206, opened 2020-06-08
RTCRtpSender unavailable on old browsers

vahidov:


lminiero:
How old are we talking about? :wink:  This should be available pretty much everywhere, now.

vahidov:
https://caniuse.com/#search=RTCRtpSender - official stats
http://joxi.ru/GrqndjVckZRPam - own stat for 20 minutes

lminiero:
I only see Firefox <=33 not supporting that: Firefox stable is 77 now. Not sure how to parse the second link you provided instead.

vahidov:
@lminiero 
Chrome 49 doesn't support this. But it is last update version for windows xp. Yep, I understand, that is old system, but we have clients on it and this chrome is working fine for rest janus.js. 
And I don't see problem in one additional if. On line 343 you have same control for safari :-)

lminiero:
The problem is we use senders and receivers a lot. The revived effort on multistream even more, almost exclusively so I believe. While a single conditional as provided here is fine, I honestly have no intention of adding tons of those to get a jurassic browser to work in the future.

lminiero:
Merging this as it's a minor change. Please notice I won't refactor janus.js in case major changes are required though. Thanks!

PR #2210, opened 2020-06-08
add enabled field to stream list

willamowius:
The "enabled" field should also be part of the stream list, see #2204


lminiero:
Ack, merging :+1:

PR #2211, opened 2020-06-08
Support for multistream PeerConnections (replaces #1459)

lminiero:
Thanks to our friends at [Zextras](https://www.zextras.com/) (thanks guys!), we finally found the time and resources to revive the old multistream effort we started in #1459!

If you've followed its evolution, you'll remember that we actually started working on it a couple of years ago: unfortunately, due to lack of feedback, testing, and more in general us being way too busy with other things, that branch eventually became severely out of sync, and ended up being full of conflicts that were basically unresolvable. For that reason, we tagged that branch as #deprecated#, and stated we'd get back to it sooner or later, to re-apply the same changes starting from the latest Janus instead.

This is exactly what happened here. We went through all the changes we did at the time, rolled up our sleeves, and slowly figured out how they could be re-integrated in the new code base. Actually, this new effort is actually a bit leaner: in fact, the original patch contained a lot of changes that were not really relevant (renaming structures, defines, code style fixes, etc.), that could actually be considered mostly noise in the grand schema of things. As such, I decided to ignore all those changes instead, and just focus on what really needed changing to get all this working, which should make looking at the changes in the code a bit easier (not easy, but definitely easier).

For what concerns plugin APIs and, most importantly, `janus.js` APIs, the changes are exactly the same as the ones done in old #1459 (apart of course from new things that were added in the meanwhile, and were updated to be used here). While I plan to write a new summary soon, for the time being you can refer to the long description I made on the original PR, as most, if not all, of the concepts explained there apply here too.

Please notice that I won't let this effort die like #1459 did. I encourage you to test this and provide feedback, because this WILL be merged soon, and will be our `1.0.0` version. Considering the disruptive nature of this branch (which is only partly backwards compatible, especially in a few plugins), we plan to create a separate branch for the existing `0.10.x` Janus, rather than simply create a tag: in fact, a tag cannot be updated (it's a snapshot), while a separate branch can still receive fixes and optionally enhancements. Please beware, though, that once multistream is merged we'll focus our development efforts on that: we'll still push and accept fixes for bugs on whatever branch `0.10.x` will be on, but we won't work on new features there. This should give you one additional motivation in exploring this branch as soon as possible, and ensure that it will work as expected for you.

I made a few tests and all plugins seem to be working as expected. I did notice a few leaks here and there, possibly caused by missing reference counters somewhere, that I plan to address in the next few days. That said, please test and provide feedback!


lminiero:
For some more context, you can refer to this talk I made last year at CommCon, where I described the relevant changes:
https://www.youtube.com/watch?v=xbsHYvBjsdY
https://www.slideshare.net/LorenzoMiniero/multistream-in-janus-commcon-2019

jamken:
hi，@lminiero 
The feature 'multistream in one PC' sounds like a good idea, but it harms the scalability of janus videoroom. In our janus-cloud (https://github.com/OpenSight/janus-cloud) solution, assume that Alice push his stream to JanusA, Bob push hist stream to Janus B, Alice need to pull Bob's stream from JanusB, Bob need to pull Alice's stream from Janus A. So Alice need two PeerConnection, one for JanusA, the other for JanusB, So does Bob. In current implementation of Janus, it's good to support this case, but after 'multistream in one PC',  how can Alice get the Bob's stream? 
Perhaps this feature 'multistream in one PC' can be a option feature, user can disable/enable it for his situation ( if he want to get the scalability, he can disable this feature and make janus works like before, one stream one pc).

lminiero:
@jamken it _IS_ optional. Read the first changes in janus_videoroom.c which is where the updated documentation is.

lnogueir:
When will this be merged and docs get updated? How will this affect videoroom plugin?

lminiero:
The sooner you provide feedback, the sooner I'll merge. I can't magically know it works for you if you don't try it and tell me. Check #1459 for docs and changes.

bwqr:
Hi there, I am kind of beginner about webrtc stuffs. I was trying to send both screen and camera on same peer connection and came across with this pr. I want to ask a question about mid and its description.

In order to differentiate the camera and screen, i want to pass descriptions. However I do not know anything about mid which belongs to my track. For example, user firstly publish its camera which is mostly assigned to mid "0". Then, when user also publishes its screen, via configure request, there are two streams listed with mids "1" and "2".  May be I need to add the new track into peer connection via addTrack function. However, again, there is no information about what does this track represents, screen or camera? And other users are not get notified about this track with event which has publishers array, when I have added track with configure request.

Do you have any suggestion about this or have any idea to solve problem? I think I am missing some points about publishing and adding tracks. I don't know if I should open an issue or comment under this pr.

And thanks for the efforts for the whole project and this pr.

lminiero:
> Do you have any suggestion about this or have any idea to solve problem? I think I am missing some points about publishing and adding tracks. I don't know if I should open an issue or comment under this pr.

You're right, this bit is still partly missing, mostly because the way you add streams has not been properly updated yet. I want to update `janus.js` so that when you add a new stream (e.g., add webcam video, and audio) you can specify the `mid` you'd like each track to have: if the `mid` exists you're replacing a track, otherwise you're adding a new one. This way, you'd  have control over that aspect.

Until that happens (I'm busy with other things for the moment) you'll have to take care of the mapping yourself: e.g., negotiate an audio/video publisher as usual (to send audio/video), and then manually add a new track yourself and send a new offer; if you inspect the PeerConnection object (`[handle].webrtcStuff.pc`) you can find out which `mid` each outgoing track has.

bwqr:
Thanks for the answer. Specifying the mid foreach track can be really good idea, and if mid already exists, overriding the mid will help the application, because we will not need description anymore. For now, I will try to implement your temporary suggestion.

roisasson:
Great initiative! I am trying to understand how the "msid" attribute will be handled in multistream. It appears that the SDP offer still assigns one hardcoded "msid" to every media stream. Won't that result in all the media streams being synchronized to each other which might unnecessarily delay some of the streams? Thank you.

kinserlin:
hi.I have deploy the multistream branch to server but serveral problems 

1. problems when starting：1）can not load plugin libjanus_echotest,libjanus_videocall,libjanus_recordplay 。2）when load plugin janus_streaming it throw error : Can't add 'rtp' mountpoint 'multistream-test', no audio, video or data have to be streamed . here is the start log below:
`Initializing ICE stuff (Full mode, ICE-TCP candidates disabled, half-trickle, IPv6 support disabled)
STUN server to use: 192.168.1.69:3478
  >> 192.168.1.69:3478 (IPv4)
  >> Our public address is 192.168.1.177
TURN server to use: 192.168.1.69:3478 (udp)
TURN REST API backend: (disabled)
Crypto: OpenSSL >= 1.1.0
[WARN] No cert/key specified, autogenerating some...
Fingerprint of our certificate: 9C:25:57:71:F1:D2:36:60:DC:29:B5:55:57:BD:0D:88:03:E8:4C:57:C6:A4:66:BE:E9:EC:FE:B0:23:74:67:35
Sessions watchdog started
[WARN] Event handlers support disabled
Plugins folder: /usr/local/janus/lib/janus/plugins
Joining Janus requests handler thread
Loading plugin 'libjanus_audiobridge.so'...
JANUS AudioBridge plugin initialized!
Loading plugin 'libjanus_sip.so'...
JANUS SIP plugin initialized!
Loading plugin 'libjanus_echotest.so'...
[ERR] [janus.c:main:5031]       Couldn't load plugin 'libjanus_echotest.so': /usr/local/janus/lib/janus/plugins/libjanus_echotest.so: undefined symbol: janus_sdp_find_first_codecs
Loading plugin 'libjanus_nosip.so'...
JANUS NoSIP plugin initialized!
Loading plugin 'libjanus_videocall.so'...
[ERR] [janus.c:main:5031]       Couldn't load plugin 'libjanus_videocall.so': /usr/local/janus/lib/janus/plugins/libjanus_videocall.so: undefined symbol: janus_sdp_find_first_codecs
Loading plugin 'libjanus_recordplay.so'...
[ERR] [janus.c:main:5031]       Couldn't load plugin 'libjanus_recordplay.so': /usr/local/janus/lib/janus/plugins/libjanus_recordplay.so: undefined symbol: janus_sdp_find_preferred_codecs
Loading plugin 'libjanus_voicemail.so'...
JANUS VoiceMail plugin initialized!
Loading plugin 'libjanus_textroom.so'...
[WARN] Data channels support not compiled, disabling TextRoom plugin
[WARN] The 'janus.plugin.textroom' plugin could not be initialized
Loading plugin 'libjanus_videoroom.so'...
JANUS VideoRoom plugin initialized!
Loading plugin 'libjanus_streaming.so'...
[ERR] [plugins/janus_streaming.c:janus_streaming_init:1776] Can't add 'rtp' mountpoint 'multistream-test', no audio, video or data have to be streamed...
JANUS Streaming plugin initialized!
Transport plugins folder: /usr/local/janus/lib/janus/transports
Loading transport plugin 'libjanus_pfunix.so'...
[WARN] Unix Sockets server disabled (Janus API)
[WARN] Unix Sockets server disabled (Admin API)
[WARN] No Unix Sockets server started, giving up...
[WARN] The 'janus.transport.pfunix' plugin could not be initialized
Loading transport plugin 'libjanus_websockets.so'...
[WARN] libwebsockets has been built without IPv6 support, will bind to IPv4 only
libwebsockets logging: 0
WebSockets server started (port 9190)...
Secure WebSockets server started (port 9191)...
[WARN] Admin WebSockets server disabled
[WARN] Secure Admin WebSockets server disabled
JANUS WebSockets transport plugin initialized!
Loading transport plugin 'libjanus_http.so'...
WebSockets thread started
HTTP transport timer started
HTTP webserver started (port 9090, /janus path listener)...
HTTPS webserver started (port 9091, /janus path listener)...
Admin/monitor HTTP webserver started (port 7088, /admin path listener)...
Admin/monitor HTTPS webserver started (port 7889, /admin path listener)...
JANUS REST (HTTP/HTTPS) transport plugin initialized!`

2.problems when run the mvideoroomtest demo . when I click start , the janus server down. Here is the log before server down:

Creating new session: 1410407300713655; 0x7f3c600029c0
Creating new handle in session 1410407300713655: 5161599129367172; 0x7f3c600029c0 0x7f3c94005f10
[ERR] [ice.c:janus_plugin_session_is_alive:614] Invalid plugin session (0x7f3c7c00346b)
[5161599129367172] Creating ICE agent (ICE Full mode, controlled)
[ERR] [ice.c:janus_plugin_session_is_alive:614] Invalid plugin session (0x7f3c7c00292b)
[ERR] [sdp-utils.c:janus_sdp_get_codec_pt_full:755] **Unsupported codec '`**
[ERR] [sdp-utils.c:janus_sdp_get_codec_pt_full:755] **Unsupported codec 'p**
 [WARN] Unknown property -1 for preparing SDP offer, ignoring...
[WARN] Unknown property 17 for preparing SDP offer, ignoring...
[WARN] Unknown property -1726152288 for preparing SDP offer, ignoring...

lminiero:
@kinserlin please check the guidelines again, before dropping a megaton of text in here.

Bug-Fairy:
Firefox seems to be having issues with simulcast on this branch. I'm seeing constant `SRTP protect error... srtp_err_status_replay_fail` errors for Firefox clients on echotest (and custom plugin). Happens with VP8 when substream 0 or 1 is selected. Localhost server.

Bug-Fairy:
Is there any specific reason sdp-utils.c has JANUS_BUFSIZE defined as 16384? It's causing SDP parsing to be cutoff in janus core when trying to negotiate multiple streams (5 audio and 5 video in my test case) in one SDP.

roisasson:
This is a great addition to Janus. Thank you for taking it on.
 
I am wondering how multistream will affect synchronization. My understanding is that all MediaStreamTracks in the same MediaStream are synchronized when rendered. The binding between MediaStreams, MediaStreamTracks and the RTP Stream (SSRC) is signaled in SDP using the msid attribute. Janus generates SDP offers that assign all MediaStreamTracks to the same MediaStream with the msid identifier “janus”. This works fine with the VideoRoom plugin today because audio and video Tracks that share the same PeerConnection to a subscriber always originate from the same publisher and typically require synchronization.
 
When using the VideoRoom plugin with multistream, one PeerConnection to a subscriber could carry MediaStreamTracks from different publishers. If they are all assigned the same msid identifier they will be treated as part of the same MediaStream and synchronized when rendered. That might cause media from one publisher to unnecessarily delay the rendering of media from all other publishers. Is my understanding correct? Sorry if I am missing something obvious.

amnonbb:
> Is there any specific reason sdp-utils.c has JANUS_BUFSIZE defined as 16384? It's causing SDP parsing to be cutoff in janus core when trying to negotiate multiple streams (5 audio and 5 video in my test case) in one SDP.

I puted 32768 there. Tested with 25 feeds in room

lminiero:
> Is there any specific reason sdp-utils.c has JANUS_BUFSIZE defined as 16384? It's causing SDP parsing to be cutoff in janus core when trying to negotiate multiple streams (5 audio and 5 video in my test case) in one SDP

Simply a value that we thought was large enough. Can't remember if we had a higher value in #1459. Anyway, I have tried a VideoRoom subscription with ~20 videos in and it seemed to work. If you experiment with different values and have proposals, happy to hear them.

lminiero:
> I am wondering how multistream will affect synchronization. [..] It seems like the msid identifiers have to somehow be configurable

Not sure. Making them configurable would mean putting this burden on plugins, making them more complicated, as the core obviously doesn't know the relations between tracks: pretty sure it would have an impact on other parts of the core too (possibly RTCP SDES?). I'd rather not worry about that now, and get everything else working before that and merged, and in case it's an issue fix it later on.

lminiero:
> Firefox seems to be having issues with simulcast on this branch

From what I've read in another issue, Firefox has other issues too. I'll have a look probably next week.

roisasson:
> > I am wondering how multistream will affect synchronization. [..] It seems like the msid identifiers have to somehow be configurable
> 
> Not sure. Making them configurable would mean putting this burden on plugins, making them more complicated, as the core obviously doesn't know the relations between tracks: pretty sure it would have an impact on other parts of the core too (possibly RTCP SDES?).

Hi Lorenzo,
 
Complexity is certainly a concern. The scenario I am worried about is, for example, a wireless endpoint with very high jitter publishing audio into a room. That audio stream might delay the playback and rendering of all audio and video streams for all of its subscribers.
 
Just a thought – instead of making the msid identifiers configurable, would it be easier to just copy them over from the incoming SDP offer to the outgoing SDP offer? I believe that would avoid the problem since browsers generate SDP offers with random msid identifiers. Thank you.

Bug-Fairy:
> > Is there any specific reason sdp-utils.c has JANUS_BUFSIZE defined as 16384? It's causing SDP parsing to be cutoff in janus core when trying to negotiate multiple streams (5 audio and 5 video in my test case) in one SDP
> 
> Simply a value that we thought was large enough. Can't remember if we had a higher value in #1459. Anyway, I have tried a VideoRoom subscription with ~20 videos in and it seemed to work. If you experiment with different values and have proposals, happy to hear them.

I've hit the limit by letting the browser generate the offer. Most of the SDP is taken up by multiple h264 profiles for each video mline. Note that I am not using janus.js but using my own front end.
Did some performance tests and profiling with JANUS_BUFSIZE 65536 and encountered no issues.

lminiero:
@Bug-Fairy JANUS_BUFSIZE is used for building SDPs, though, not for parsing. So increasing it for SDPs that come from browsers is useless: there's no limit to how large SDPs coming from there can be.

thevaliio:
If I create a multistrem with more than 19 videos I get the error:
[ERR] [sdp-utils.c:janus_sdp_parse:656] Invalid line (2 bytes): a=
[ERR] [sdp.c:janus_sdp_preparse:42]   Error parsing SDP? Invalid line (2 bytes): a=
[ERR] [janus.c:janus_plugin_handle_sdp:3377] [6617817997017288] Couldn't parse SDP... Invalid line (2 bytes): a=
[ERR] [janus.c:janus_plugin_push_event:3295] [6617817997017288] Cannot push event (JSON error: problem with the SDP)
 >> Pushing event: 465 (Invalid SDP)

lminiero:
That's because the generated (written) SDP is incomplete (cut to ~16k) and so parsing fails because the SDP is invalid: but again, that property impacts generating SDPs, *not* parsing

thevaliio:
> That's because the generated (written) SDP is incomplete (cut to ~16k) and so parsing fails because the SDP is invalid: but again, that property impacts generating SDPs, _not_ parsing

I'm not a developer so if that question shouldn't be here please delete it. So we have JANUS_BUFSIZE to 64k but you sayin g this is for generating only.  So who is cutting the SDP to 16k? What is the fix for this?

lminiero:
Small summary (mostly a reminder to myself) of things that apparently need to be fixed:

1. Some inconsistencies with Firefox, that someone is looking into in #2324 already. Honestly, a low priority for me at the moment, as I want the whole infrastructure to work fine initially: any tweak to make specific browsers happy can wait. Of course, if you guys look into it in parallel (as @Bug-Fairy is doing in that PR), it's more than welcome, as we can converge at the end on activities that are actually independent of each other.
2. JANUS_BUFSIZE not being large enough when we have to generate SDPs with many m-lines. This is quite low priority for me as well, especially considering it's a low hanging fruit that you guys are already experimenting with, and so you'll probably find a solution sooner than I will (I have no time to make tests with many streams). I'm not sure just increasing the buffer size will really fix the issue, as it would just "move" the problem: we may have to look into using something like [g_alloca](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-alloca) to make that part dynamic instead (e.g., if we're generating a small SDP, use a smaller buffer), but I don't have a clear idea yet on how to do that properly. Maybe one of you would like to look into it in the meanwhile?
3. Renegotiations in a publisher (e.g., an already active publisher that at a certain point decides to start sending video when he wasn't before, or add a new video streams, or stop sending audio) are currently not notified to subscribers: this is indeed a limitation that I didn't take into account even in the original PR, and something that will need to be fixed. It's what I'm currently looking into, with a few doubts in place (autodetect changes vs. forcing changes to be notified via API) on how to take care of that. I hope to have a commit for that soon.
4. We still don't have an easy way to add/remove/replace local tracks in janus.js, which indeed makes it harder to experiment with the multistream features (to test things in the EchoTest I always have to do manual changes on the PeerConnection). No sense on fixing this until I've fixed the VideoRoom Publishers renegotiation part, though, otherwise you'll start tinkering with that, it won't work as you think it does, and it will just generate noise. As such, I'll improve janus.js only after point 3. is taken care of.

Not sure if I'm missing something, but if so, my priorities are laid down for the moment, so whatever I may have forgotten will have to wait (unless someone wants to help address it).

Bug-Fairy:
> That's because the generated (written) SDP is incomplete (cut to ~16k) and so parsing fails because the SDP is invalid: but again, that property impacts generating SDPs, _not_ parsing

It's truncating the SDP before it's passed to the plugin as there's a call to janus_sdp_write as part of the SDP anonymizing. So while the core might be seeing the whole SDP, plugins get passed a truncated version.

lminiero:
FYI, the commit adds a new notification about changes in a publisher, that is, in case a publisher renegotiation takes place. The notification basically looks exactly like the event you already receive when the publisher first creates the PeerConnection, meaning you'll get a `publishers` array with just the publisher info: the new object will contain the updated information on what the publisher is sending.

A couple of important points related to this:

1. First of all, you'll of course need to make sure you don't subscribe again to media that you're subscribed to already: in fact, the event you receive may be a duplicate of what you know already (e.g., publisher just renegotiated without adding or removing anything, or they removed video but kept audio); this is something you have to take care of at the application level.
2. Streams can now have a new property, called `disabled`: when `disabled: true`, it means the related m-line identified by that `mindex/mid` is now gone (e.g., the publisher removed the video). I decided to keep the stream there and mark it as disabled, rather than removing it from the list, since it's more correct (the m-line is indeed there, it's inactive). Again, it's up to you, in the web application, to use that information accordingly, e.g., sending an `unsubscribe` for the streams that are now disabled (by default you'll remain subscribed to an inactive stream, so you won't receive media from that stream).

I started modifying the `mvideoroomtest` demo, but work on that still isn't complete: I'll probably work on improving that in the next few days. Please let me know though if the changes make sense to you, and if you can see problems with the way they have been implemented.

bwqr:
Hi there, @lminiero, I was testing this new commit. I found an issue that disabled property for streams does not exist for joined events. That means, if a user attaches a new plugin with ptype publisher, joined event for this plugin contains all the streams without disabled property. If another publisher updates its configuration, new event will be containing the disabled property for streams. Is it intentional or a missed case?

lminiero:
`disabled` is only set if the m-line is inactive.

tomleavy:
@lminiero I am testing this branch, and one thing I noticed is that the subscribe request to add additional streams will return an ack that is tagged with the transaction id, but the subsequent "updated" event will not have the transaction id included in its properties. Is that by design? JoinAndConfigure works similarly, and in that case both the ack and the "joined" event have the transaction id tagged

lminiero:
@tomleavy events are not required to include a transaction ID, and you shouldn't expect one. A `subscribe` may not trigger an update right away, but schedule one instead (e.g., we've sent an update we haven't received an answer yet, so wait until that happens), or may never send an update at all. When scheduled, the transaction ID would most definitely not be there, as it would be lost already. As such, for consistency we never set it in "updated" events.

tomleavy:
Thanks for the explanation, makes sense @lminiero 👍 . Really appreciate all the hard work that is going into this

tomleavy:
When joining a video room as a publisher on this branch, the `joined` event has conflicting information in terms of the `talking` property for audio. In this case the user in question is actually talking, but the new property under `streams` has `false`, while the legacy property has `true`

```
"publishers": [
            {
               "id": 8371564275892102,
               "display": "iPhone SE (2nd generation)",
               "audio_codec": "opus",
               "video_codec": "vp8",
               "streams": [
                  {
                     "type": "audio",
                     "mindex": 0,
                     "mid": "0",
                     "codec": "opus",
                     "talking": false
                  },
                  {
                     "type": "video",
                     "mindex": 1,
                     "mid": "1",
                     "codec": "vp8"
                  }
               ],
               "talking": true
            }
         ]
```

lminiero:
I just rebased to the latest master and force-pushed the changed: this should include some simulcast-related fixes and enhancements, besides a fix on bundled streams that was affecting multistream. Please let me know if with this new update you notice any regression or bug.

ZeXtrasJay:
We found two regressions, the first is on audiobride plugin, a nullpointer in https://github.com/meetecho/janus-gateway/pull/2211/files#diff-489855b0c310258c85ffcd789ce59a22bd40ba5286f52255e185b506079abaadR6307

And sometimes video plugin sends truncated json, we are able to reproduce it consistently

lminiero:
@ZeXtrasJay the link seems to point to a much larger diff, so we can't really check what's wrong that way. Please share traces or libasan dumps for the AudioBridge NULL pointer you mention.

On the truncated JSON, seems much more likely this is an issue in the transport plugins, as those are the ones serializing the jansson objects. Possibly something we fixed recently in master (especially if you're using websockets) that hasn't landed in this branch yet. You can verify that by copying the `janus_websockets.c` file from master to your multistream code and recompile, and see if it fixes it. I plan to realign the code soon, as soon as a couple of PRs are merged first though.

lminiero:
Hi all,

I've been made aware that there seems to be an issue when unsubscribing from an audio stream, and then subscribing to a new audio stream that replaces the previous m-line: apparently, the fact that the SSRC changes when the m-line is updated to host a different stream confuses the browser, and causes playout or processing issues. The issue seems to go away by preventing the SSRC to change when the m-line changes "owner", which I'm not confident about (I found old notes where I actually had to add the "reset the SSRC" behaviour in that case exactly to prevent playout issues).

I don't remember this issue happening when I tested myself, but it may be dependant on the browser and/or the browser version. Has anyone noticed something like this happening in their tests? Pinging @ZeXtrasJay @amnonbb @Bug-Fairy in particular as I remember them being more actively involved in testing.

Bug-Fairy:
@lminiero There definitely exist some issues with multistream renegotiations. Something to do with browsers relying primarily on mid while Janus relies on mline order I believe.

amnonbb:
@lminiero I tried to test the same with video stream and got some strange result.
I found problem with subscription process, that not simple to reproduce. I'll try and report the result.

M0Rf30:
> @ZeXtrasJay the link seems to point to a much larger diff, so we can't really check what's wrong that way. Please share traces or libasan dumps for the AudioBridge NULL pointer you mention.
> 
> On the truncated JSON, seems much more likely this is an issue in the transport plugins, as those are the ones serializing the jansson objects. Possibly something we fixed recently in master (especially if you're using websockets) that hasn't landed in this branch yet. You can verify that by copying the `janus_websockets.c` file from master to your multistream code and recompile, and see if it fixes it. I plan to realign the code soon, as soon as a couple of PRs are merged first though.

Actually the link refers to the single line 6307 of janus audiobridge. Anyway I report here:
```
janus_sdp *offer = NULL, *answer = NULL;
if(got_offer) {
janus_sdp *answer = janus_sdp_generate_answer(offer);
/* Only accept the first audio line, and reject everything else if offered */
GList *temp = offer->m_lines;
```
With versions upper to 0.10.5 (at least until 0.10.8) this code brings issues

lminiero:
@M0Rf30 @ZeXtrasJay apologies for the late feedback. Your snippet did indeed help me find out where the issue was, which I should have now fixed.

M0Rf30:
@lminiero thanks for your efforts. Will give you a feedback ASAP.

wdv4758h:
Hi, I'm testing this `multistream` branch. The multistream videoroom works on my end, and I can dynamically create new multistream streaming mountpoint.

Is there a way to change the multistream mountpoint after creation? like adding some more media in on the fly.

lminiero:
> Is there a way to change the multistream mountpoint after creation? like adding some more media in on the fly.

No, Streaing mountpoints are static in this version. We don't have plans to change that, definitely not before the PR is merged.

wdv4758h:
> No, Streaing mountpoints are static in this version. We don't have plans to change that, definitely not before the PR is merged.

Thanks for clarification. It sounds reasonable.
And I notice that there are some similar discussion in the threads before, sorry for the duplication 😅

> I don't plan to change the way the new Streaming plugin works right now. One thing we might do in the future is make the VideoRoom itself support remote publishers, but that won't happen here. It's something we'll only look into once this effort has stabilized and it has been merged.

This VideoRoom remote publisher is exactly the thing I'm looking for.

sanjeev0291:
@wdv4758h This looks like a great feature. When do you see this getting merged this PR? We are eagerly waiting for this PR.

lminiero:
As I said other times, I need more feedback. Start using it and testing it. Feature-wise, I still need to add a couple of things in `janus.js` (like the ability to add/replace multiple streams), but server-side it's complete.

wdv4758h:
Yeah, I got multistream version in-house deployment. It works great with demo site, the connections are good. We are integrating with our own frontend to test more.

lminiero:
@wdv4758h thanks, looking forward to your feedback!

Someone on the group already pointed out our 16k buffer when crafting SDPs may be too little (it wasn't enough for the 25 streams PC he created), something we were already aware of since https://github.com/meetecho/janus-gateway/pull/2211#issuecomment-671353903. I'm hesitant to just put a 2x or 4x value in there, because it means allocating a lot of static memory any time we craft an SDP (even little ones), but you may want to experiment with that too. I'll have to think of a way to do this better, maybe with smaller buffers for each m-line, and a `g_realloc` for the SDP itself.

cameronlucas3:
@lminiero regarding SDP buffer size, we have found with our own frontend deployment that not even 64k is always quite enough. We ended up just increasing the buffer size to 128k for the time being to error on the side of caution. Supporting a more dynamic memory allocation, as you proposed would be ideal. 

Overall this branch so far has been working great for us, with our initial internal testing.

lminiero:
@cameronlucas3 I don't like the idea of big static arrays, though, and I definitely don't trust them to always be reliable, e.g., when using different compilers. @atoppi for instance mentioned how musl has much more narrow limits when it comes to how big they can be, and so would definitely be an issue if it were used here (it was in libnice for some users for a similar reason). I'll probably come up with a dynamic solution to this instead, possibly with an allocated and growing buffer.

wdv4758h:
> @wdv4758h thanks, looking forward to your feedback!
> 
> Someone on the group already pointed out our 16k buffer when crafting SDPs may be too little (it wasn't enough for the 25 streams PC he created), something we were already aware of since [#2211 (comment)](https://github.com/meetecho/janus-gateway/pull/2211#issuecomment-671353903). I'm hesitant to just put a 2x or 4x value in there, because it means allocating a lot of static memory any time we craft an SDP (even little ones), but you may want to experiment with that too. I'll have to think of a way to do this better, maybe with smaller buffers for each m-line, and a `g_realloc` for the SDP itself.

We have ingerated multistream without our own frontend. I do meet the problem about SDP buffer size with 20+ users, it got cut off with 16k. Thanks for mentioning that, life saver. I'm going to simply increase the number for now, the memory usage is not our concern at this point. We will have tests with more users in a room after that.

So far it's working great, much better than before!

lminiero:
I just pushed a commit that changes the buffer for the generated SDP dynamically: as anticipated, I went for a static buffer for the current mline, which I then append to the SDP; in case the SDP buffer is to small to accomodate the current m-line, I realloc the buffer. This seems to be working as expected in the few tests I made, but of course feedback would be welcome, especially from those of you who've been experimenting with many tracks in the same PeerConnection and had encountered the bufsize issue before.

plugins/janus_videoroom.c
```diff
@@ -5095,61 +6239,29 @@ void janus_videoroom_setup_media(janus_plugin_session *handle) {
...
+					if(ps && ps->type == JANUS_VIDEOROOM_MEDIA_VIDEO && ps->publisher && ps->publisher->session) {
+						janus_videoroom_reqpli(ps, "New subscriber available");
 					}
+					temp = temp->next;
+				}
+				/* Also notify event handlers */
+				if(notify_events && gateway->events_is_enabled()) {
+					json_t *info = json_object();
+					json_object_set_new(info, "event", json_string("subscribed"));
+					json_object_set_new(info, "room", json_integer(s->room_id));
```
wdv4758h, line 6261: @lminiero one tiny bug in this branch, the event from videoroom for `subscribed` is still using integer room id even though config use string id.

With string id config, I get something this:
```
{"emitter":"...","type":64,"timestamp":1618694821191801,"session_id":5564909303519302,"handle_id":337004046279573,"opaque_id":"videoroomtest-Byx47uLLjCCO","event":{"plugin":"janus.plugin.videoroom","data":{"event":"subscribed","room":0}}}
```

The room should be string, but I get `"room":0`.

wdv4758h:
The update of dynamic SDP buffer is working great for me so far.

plugins/janus_videoroom.c
lminiero: Ah good catch, thanks! I'll fix this later.

tezet:
Hi Lorenzo,

I think there might be a bug in the videoroom's data channel handling on the multistream branch. 

Once we switched from master to multistream, we noticed that data channel was no longer relaying data.

On the publisher side, the data channel is negotiated, but the data stream for some reason is always disabled:

`{"videoroom":"event","room":6546377825620122,"configured":"ok","audio_codec":"none","video_codec":"h264","streams":[{"type":"audio","mindex":0,"mid":"0","disabled":true},{"type":"video","mindex":1,"mid":"1","codec":"h264"},{"type":"data","mindex":2,"mid":"2",**"disabled":true**}]}}`

I tried enabling it afterwards by sending configure message with data stream mid, but this didn't enabled the stream either.

I can see that the data channel is actually opened, because the messages are received on Janus side:
`SCTP data received of length 6 on channel with id 0.`

Unfortunately when I subscribe to the feed ("the old way" or specifying all streams), subscribers do not receive relayed data.

Is there anything I am missing in understanding the API, or could it be an issue?

Please let me know if you would like some more logs to look into.

We are not using janus.js.

Thanks,
Tomek

thatsmydoing:
We've also been trying out this branch and can confirm that subscribing to data channels doesn't work. The data channel never opens.

We tried fixing the `disabled` issue with this patch
```
diff --git a/plugins/janus_videoroom.c b/plugins/janus_videoroom.c
index ff6ff550..6980829a 100644
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -9739,6 +9739,8 @@ static void *janus_videoroom_handler(void *data) {
 								}
 							}
 							mdir = (ps->vcodec != JANUS_VIDEOCODEC_NONE ? JANUS_SDP_RECVONLY : JANUS_SDP_INACTIVE);
+						} else if(m->type == JANUS_SDP_APPLICATION) {
+							mdir = JANUS_SDP_RECVONLY;
 						}
 					}
 					ps->disabled = (mdir == JANUS_SDP_INACTIVE);
```

and data channels sort of work but not really. I'm not sure how it worked prior to multistream but it seems that both publishers and subscribers have to `createDataChannel`. When a publisher sends to its own channel, it triggers a `datachannel` event on all subscribers. The channel created by the subscriber does not receive these messages only the one from the `datachannel` event.

This almost works but we also found that we can only subscribe to one data channel stream. If we subscribe to more than one, it just excludes the new data stream based on the subscription response. If we subscribe to only a new data stream, the subscribe message only acks and never sends a success/error message.

lminiero:
> If we subscribe to more than one, it just excludes the new data stream based on the subscription response

That's normal and expected. You can't create more than one data channel m-line in WebRTC. This is how it's done in the multistream VideoRoom, as explained in #1459 (which as I wrote in the description was and remains the main reference for changes, please do read that long intro)

> In the VideoRoom with a multistream setup this might be seen as a problem, if we want to use datachannels, because it means we're getting data channel messages from multiple publishers, and we might not be able to say which belongs to whom. Anyway, now that Janus supports multiple streams in the same datachannel, this is not an issue anymore: now the VideoRoom plugin always uses the publisher's ID as the "label" for outgoing data channel messages, which means that you just need to look at the label of incoming messages to know where it's from.

I haven't tested datachannels in the VideoRoom since the new refactoring. I'll try to do that when I have some time.

lminiero:
Datachannel support should be fixed in the VideoRoom. It was actually working as expected when using the old/legacy demo, but not in the new one (and so the new API), indeed due to the stream resulting incorrectly `disabled`. There was a missing direction management for application m-lines, which the above commit fixes.

thatsmydoing:
Yeah, I only ran into problems when trying to connect a 3rd user to the data channel.

> As I've explained in a previous comment, though, it's normal for a single m-line to be negotiated

Hm, I hadn't looked at the SDP but the videoroom API response only contained one stream of type data even if I specify multiple data streams to subscribe to. I at least expect videoroom to allow multiple data streams even if the underlying webrtc has the limitation.

lminiero:
> the videoroom API response only contained one stream of type data even if I specify multiple data streams to subscribe to

Yes, and that's normal. You only need one m-line for data channels. A single datachannel can contain multiple sessions, which we identify via labels. On the subscriber side, you're notified about additional streams being opened and data on them, and the label tells you who it's from (in the VideoRoom, we use the publisher ID as the label there).

lminiero:
As a side note, I tried with 3 users and it only half worked: the third user could receive messages but only from the second user, while messages from the third user ended up nowhere. First and second user talked fine. It's probably an internal subscription/routing logic typo somewhere, because the datachannel is correctly initialized for them all and in both directions.

thatsmydoing:
I think I might be misunderstanding something with regards to streams.

Is it that the `streams` array in the `subscribe` response corresponds 1-to-1 with the SDP / webrtc tracks layout and not actually what you've logically subscribed to in terms of `publisher_id, mid` pairs?

But it is possible to subscribe/unsubscribe to particular publishers independently? It's just that that's all internal state not revealed in the API?

lminiero:
As I suspected it was an internal routing issue: to check if a source was actually in the subscriber list, we were only matching the very first data subscription in the list, and so subscribers would only receive messages from the users they subscriber to first. I fixed that and now it works as expected with three users for me.

thatsmydoing:
> I fixed that and now it works as expected with three users for me.

Thanks! I can confirm on my side as well.

We've been testing this using our own websocket based client (not `janus.js`) and for the most part it works well after the datachannel issue was fixed. There were a couple of things we thought could be improved though.

 * it's only possible to get the list of publisher streams by joining as a `publisher`, that means even if we only want to subscribe, we have to join as a publisher first to get the streams, and then we can subscribe
 * it's not possible to get the list of currently published / subscribed streams outside of events, our use case here is to confirm our state after reclaiming a disconnected session
 * we only receive a "confirmation" event (like `updated` from `subscribe`) only if the SDP changes / negotiation is needed. if the `subscribe` action results in no SDP changes, like subscribing to a data stream when we already have a data channel, we receive no confirmation from janus. the use case here is we're a server proxying requests for a client, so don't see anything that happens in the browser
 * not answering an SDP causes all further SDP related requests to block. there doesn't seem to be a way of resetting this aside from re-creating the handle. The use case here is if we're reclaiming a disconnected session, we might have missed events and will not have received the SDP to answer.

plugins/janus_videoroom.c
```diff
@@ -941,38 +1010,48 @@ room-<unique room ID>: {
...
-	"offer_video" : <true|false; whether or not video should be negotiated; true by default if the publisher has video>,
-	"offer_data" : <true|false; whether or not datachannels should be negotiated; true by default if the publisher has datachannels>,
-	"substream" : <substream to receive (0-2), in case simulcasting is enabled; optional>,
-	"temporal" : <temporal layers to receive (0-2), in case simulcasting is enabled; optional>,
-	"fallback" : <How much time (in us, default 250000) without receiving packets will make us drop to the substream below>,
-	"spatial_layer" : <spatial layer to receive (0-2), in case VP9-SVC is enabled; optional>,
-	"temporal_layer" : <temporal layers to receive (0-2), in case VP9-SVC is enabled; optional>
+	"streams" : [
+		{
+			"feed_id" : <unique ID of publisher owning the stream to subscribe to>,
```
thatsmydoing, line 1015: This is actually just `feed`

plugins/janus_videoroom.c
```diff
@@ -1007,15 +1100,116 @@ room-<unique room ID>: {
...
+ * The syntax of the \c subscribe mirrors the one for new subscriptions,
+ * meaning you use the same \c streams array to address the new streams
+ * you want to receive, and formatted the same way:
+ *
+\verbatim
+{
+	"request" : "subscribe",
+	"streams" : [
+		{
+			"feed_id" : <unique ID of publisher owning the new stream to subscribe to>,
```
thatsmydoing, line 1129: Same as above, this is also just `feed`.

thatsmydoing (commented):
We also noticed some mistakes in the documentation while we were implementing our own client.

lminiero:
>     * it's only possible to get the list of publisher streams by joining as a `publisher`, that means even if we only want to subscribe, we have to join as a publisher first to get the streams, and then we can subscribe

That's always been like this and it's not going to change. That said, you don't really _need_ Janus to notify you about streams: if you're in control of the signalling, then you know who's publishing what, meaning you can do presence information on your own and in your own way with your own protocol, just creating subscriber handles if you're not going to publish anything.

>     * it's not possible to get the list of currently published / subscribed streams outside of events, our use case here is to confirm our state after reclaiming a disconnected session

Not sure what this means? You mean a dedicated request that can be made to see what you're publishing/subscribing?

>     * we only receive a "confirmation" event (like `updated` from `subscribe`) only if the SDP changes / negotiation is needed. if the `subscribe` action results in no SDP changes, like subscribing to a data stream when we already have a data channel, we receive no confirmation from janus. the use case here is we're a server proxying requests for a client, so don't see anything that happens in the browser

As in simply some sort of an "ack" from the plugin, maybe with the transaction ID of the request? It can be added, but it still will not include the SDP.

>     * not answering an SDP causes all further SDP related requests to block. there doesn't seem to be a way of resetting this aside from re-creating the handle. The use case here is if we're reclaiming a disconnected session, we might have missed events and will not have received the SDP to answer.

That's expected. We have to enforce a rigid offer/answer cycle, or things will break in the browser too. There's no way plugins can reset the situation anyway, as he offer/answer state is in the core too, and they have no control over that.

plugins/janus_videoroom.c
lminiero: Thanks, fixed both (actually three, "unsubscribe" had the same typo).

thatsmydoing:
> Not sure what this means? You mean a dedicated request that can be made to see what you're publishing/subscribing?

Yes. Although upon further thought, it doesn't mean much for recovering from a disconnect since if we send a publish/subscribe request and disconnect then reconnect. We would have missed the SDP response which means the peer connection will be in limbo anyway and have to be recreated. It might make sense for data channels though which don't always have SDPs.

> As in simply some sort of an "ack" from the plugin, maybe with the transaction ID of the request? It can be added, but it still will not include the SDP.

Yes. This would at least allow us to know it wasn't an error or have to add a timeout (which might mean something else).

> That's expected. We have to enforce a rigid offer/answer cycle, or things will break in the browser too. There's no way plugins can reset the situation anyway, as he offer/answer state is in the core too, and they have no control over that.

I see. Would it at least be possible to have the currently pending offer be resent?

Thinking about all this a bit more, it might be better to just use a different transport if unexpected disconnection is a concern.

tezet:
Thank you, Lorenzo! The data channel is working now as expected.

mirkobrankovic:
Hey, one question regarding subscribe and unsubscribe:
UserA publishes 3 streams (1,2,3 these are objects for ease of writing) and userB subscribes to all 3. 
Later on, userB decides he wants only streams 1 and 2, is it the same if userB sends `subscribe(streams[1,2])` or `unsubscribe(streams: 3)`? Will this result in the same SDP in response?

lminiero:
No, it will not result in the same SDP response. If B is already subscribed to 1,2,3 and now only wants 1,2, B needs to unsubscribe from 3; sending subscribe(1,2) will subscribe to 1,2 again, meaning B will now receive 1,2,3,1,2 (duplicates).

mirkobrankovic:
thanks :)

mirkobrankovic:
so from browser perspective if userB wants to subscribe to streams 1,3,4 (assuming thar userA started the stream4) so removing stream2 and adding stream4, browser needs to do `unsubscribe(stream2)` and then `subscribe(stream4)` and do the SDP negotiation 2 times? Or can do first subscribe and then unsubscribe but the later sdp will be the final one browser wants/needs.
Seems like there can be api that can combine the 2 steps ? or maybe I don't see the full picture :)

lminiero:
Doing unsubscribe first and then subscribe has the advantage that we may be able to re-use m-lines (the ones freed by the previous unsubscribe), if it's the same medium and same codec (e.g., unsubscribing from VP8 video to then subscribe to other VP8 video). If you subscribe to the new one first and then unsubscribe from the other you'll have an empty unused m-line (until it's possibly reused later).

The API doesn't allow you to combine subscribe and unsubscribe, and I don't plan to do anything in that direction until this is merged.

mirkobrankovic:
fair points, thanks for clarifying :)

dewebdes:
> hi，@lminiero
> The feature 'multistream in one PC' sounds like a good idea, but it harms the scalability of janus videoroom. In our janus-cloud (https://github.com/OpenSight/janus-cloud) solution, the participants get each stream of the other publishers on an indivudual PC from the different Janus-Servers. Assume that Alice push his stream to JanusA, Bob push hist stream to Janus B, Alice need to pull Bob's stream from JanusB, Bob need to pull Alice's stream from Janus A. So Alice need two PeerConnection, one for JanusA, the other for JanusB, So does Bob. In current implementation of Janus, it's good to support this case, but after 'multistream in one PC', how can Alice get the Bob's stream?
> Perhaps this feature 'multistream in one PC' can be a option feature, user can disable/enable it for his situation ( if he want to get the scalability, he can disable this feature and make janus works like before, one stream one pc).

The last version of janus gateway has still bug for videoroom on pc, I will try ur repo and i wish it can fix this problem.
All thisng are greate but this disability goes on to stop me...

dewebdes:
> hi，@lminiero
> The feature 'multistream in one PC' sounds like a good idea, but it harms the scalability of janus videoroom. In our janus-cloud (https://github.com/OpenSight/janus-cloud) solution, the participants get each stream of the other publishers on an indivudual PC from the different Janus-Servers. Assume that Alice push his stream to JanusA, Bob push hist stream to Janus B, Alice need to pull Bob's stream from JanusB, Bob need to pull Alice's stream from Janus A. So Alice need two PeerConnection, one for JanusA, the other for JanusB, So does Bob. In current implementation of Janus, it's good to support this case, but after 'multistream in one PC', how can Alice get the Bob's stream?
> Perhaps this feature 'multistream in one PC' can be a option feature, user can disable/enable it for his situation ( if he want to get the scalability, he can disable this feature and make janus works like before, one stream one pc).

I was tested it.
Dont understand.
in ur demo u use: 8088, 8089
and in ur config u use: 8188
and nothing work, and no doc is there
bad job :(

xyndata:
Dear Lorenzo and team. Is there any rough estimation of when the multistream would be merged into the master? Any open tasks? Could I help with development? Thank you.

amnonbb:
I have to say that last and previous commits work stable enough. I did not meet any crash during last 2 month.
We use 12 servers with video room plugin. 9 servers with stream plugin. 2 servers with audiobridge.

lminiero:
The last two commits fix an issue with VideoRoom stream descriptions. Apparently, we were overwriting the descriptions for other publishers when providing our own, in a few cases, and a descriptions update without renegotiating was being ignored as well: both issues should now be fixed, even though I didn't test extensively.

stefanzvkvc:
Hi all :)

I am currently facing a problem related to the resubscribe process.
Flow is:
Alice publishes camera, Bob subscribes on Alice's camera stream and starts receiving media.
After that Bob unsubscribes and subscribes again to Alice's camera.
Result of re-subscribing is a black screen for Bob and track has readyState "ended". 

Has anyone had a similar experience during developing of resubscribe process?

mirkobrankovic:
IS it by design that all Subscriptions are combined tracks into one stream? In VideoRoom but also Echotest, if you add 2 tracks with 2 streams to PeerConnectin locally you will receive one bundled stream with 2 tracks:
`a=msid:janus janus0
`and
`a=msid:janus janus1
`
```
pc1.getLocalStreams()
(2) [MediaStream, MediaStream]
pc1.getRemoteStreams()
[MediaStream]
```
So `getRemoteStreams()[0].getTracks()` will show 2 tracks in one stream.

Is this on purpose? since it is confusing out client "a bit" :)

This makes even bigger problem when we start using Unsubscribe API and then MIDs are re-used.
When a transceiver is set to "inactive" stream.onremovetrack is fired with the track that is on the inactive transceiver
When you reuse track from  mid=1 to mid=0 after previous track on mid=0 was removed it will fire onremovetrack for both tracks

lminiero:
> S it by design that all Subscriptions are combined tracks into one stream? In VideoRoom but also Echotest, if you add 2 tracks with 2 streams to PeerConnectin locally you will receive one bundled stream with 2 tracks

I wouldn't say it's by design, but in practice we're not taking the msid into account, so the end result is what you see there. I haven't given much thought on how this should be addressed yet, to be honest, as I believe plugins may need to be involved (as they're the ones who know which tracks belong together from a plugin logic perspective), which might make negotiation more complex.

mirkobrankovic:
Seems to me that it would be useful to have msid as publisher_feed_id or sender_id and then track_id to be real track_id gotten from publisher. THen receiving PeerConnection on Subscribers would be split in streams per publisher and holding all track_ids.
Problems would be that SDP would be even larger and benefits of BUNDLE can't be used, but also in case publisher decides to use replaceTract() would lead to inconsistent state (since remote peer doesn't know track was replaced).

mirkobrankovic:
> In WebRTC all tracks with the same msid are synchronized when rendered. Assigning the same msid to tracks from different publishers will also create lip sync issues. Retaining the published msid and track identifier should solve the sync problem.

Ok so exactly grouping tracks of same publishers in one msid/stream would seem like a good practice (i'm also not very familiar with this multistream approach) :)

lminiero:
This is much easier said than done, unfortunately. There are different layers we need to worry about here. We terminate msid and msid-semantic, and leaving this up to plugins will require a lot of work. Besides, not sure if this would have an impact on cname, which would in turn impact RTCP. I won't have time to look into all this, this week, so in case you have further feedback (and maybe SDP examples of how you think it should work) feel free to add it for when I'll have time to address this.

mirkobrankovic:
yeah I figured it is not simple at all. Ok will provide a O/A once I figure out how it can work.

mirkobrankovic:
Just roughly:
https://pastebin.com/UrVf05Tp

The point is that lipsync groups are in msid-semantics as piblisher ids (feed ids are unique i guess for room for each publisher so to avoid coalision):
`a=msid-semantic: WMS PUBLISHER1_FEED_ID PUBLISHER2_FEED_ID
`
I did copy paste of mid 0,1 to 2,3 so many things are same like ports and ssrcs, but point it to have msids grouped by publihers:
`a=msid:PUBLISHER1_FEED_ID TRACK_0_ID
`
THe only thing I'm not sure about is:
`a=group:BUNDLE 0 1 2 3`
Will this be possible.

I think this should create, on client side, when PeerConnection set this as remote description 2 streams with 2 tracks in each of them.But again I'm not sure about RTCP and all the correction protocols that follow that.

lminiero:
Just noticed I didn't write it here explicitly as well: there is a date for when multistream will be merged, and a strategy for how to deal with those that prefer to keep on using the current 0.x branch. Please see PR #2855 for more information.

lminiero:
For those that are interested, I wrote a blug post to summarize the changes:
https://www.meetecho.com/blog/multistream/

M0Rf30:
> The moment has finally come: merging the multistream branch on the day of Janus' 8th birthday! partying_face cake

So great. I also saw your intervention to FOSDEM 2022. Congrats paisà!

tbence94:
This PR broke the previous `.mjr` file naming conventions by adding another number to the end of the filename.

Previous filenames ended like this:
- `...-audio.mjr`
- `...-video.mjr`

After the PR they end like this:
- `...-audio-0.mjr`
- `...-video-1.mjr`

This can break applications if they expect that the filenames will end in a certain way.
I think this sould be noted in the changelogs somewhere.

Related changes in the code: https://github.com/meetecho/janus-gateway/pull/2211/files#diff-d96a677f8b8c31f15d44bf3a1d83934e5a6b6bdfab05bd70fd7be6119fb26740R7010-R7012

PR #2202, opened 2020-06-08
Update janus.d.ts

Glandos:
add types for `attachMediaStream` and `reattachMediaStream` helpers.

PR #2215, opened 2020-06-09
Update janus.d.ts

Glandos:
Missing return types for `attachMediaStream` and `reattachMediaStream`

Sorry, I missed it in #2202 :/

### DIFF
--- a/npm/janus.d.ts
+++ b/npm/janus.d.ts
@@ -137,8 +137,8 @@ declare namespace JanusJS {
 		static warn(...args: any[]): void;
 		static error(...args: any[]): void;
 		static randomString(length: number): string;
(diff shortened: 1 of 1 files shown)


lminiero:
Sounds good, I never use those things anyway :hand_over_mouth: 
As a side note, do you think those will need to be updated too, before we the multistream branch, considering there are changes to `janus.js`? (PR 2211)

Glandos:
Good question :)

However, given the fact that these types are hint for Typescript compilation, but not a blocker, you can safely merge #2211 first. And if someone like me need these types to be updated, well, it doesn't seem too much to do afterwards.

Moreover, it's good to wait for the API to stabilize. But if @lminiero has time to do it, it could be a good addition.

Issue #2188, opened 2020-05-31
Segfault in SIP plugin


zaltar:
We're running into a use after free bug in the SIP plugin that seems to randomly happen when a user has multiple incoming calls and abruptly closes their browser window destroying the session.

GDB backtrace: https://pastebin.com/UTG2FbJe
AddressSanitizer output: https://pastebin.com/4D4ADzmT

These were produced on the latest master, commit 7514b78e36a8bfb97d9474d4d0182cd307c906d2. Let me know if there's any more info I can provide.

Thanks!

lminiero:
Sounds like a refcount error. The session has been already freed (refcount=0) even though the Sofia SIP loop thread is still alive, and that has a reference of its own. Try uncommenting the `#define REFCOUNT_DEBUG` in `refcount.h` and recompile: it will add A LOT of additional log lines (make sure you have room for those), but if it crashes again it will give us more info to track the life cycle of the session that went away.

zaltar:
From latest master commit 0bb11f33d6c143d05ffae79796e46962019f5621
Here is the log from the crash with `RECOUNT_DEBUG` defined: https://pastebin.com/nTryNTUX
GDB backtrace: https://pastebin.com/w4Uf6AjM

From what I can tell, the helper `janus_sip_session` is being freed when it's removed from the master, before the Sofia SIP loop ends.  If an event happens to come in on the `nua_helper_t` that the helper `janus_sip_session` was bound to it crashes since it was already freed.

lminiero:
Thanks, I'll look into this later to see if I can spot anything.

On a general note, it's normal for helper sessions to go away before the loop ends: the loop belongs to the master session, and stays up for as long as it needs, while helpers can come and go. As such, it can indeed happen that a helper session is still used by the loop somehow and cause that issue. I'll have to check if there's an event in the Sofia SIP stack we can use as an indication that the session is "free" to go: even in that case, though, there may be the risk of leaks, because if we interrupt the master goes away and the loop is closed, the event to get rid of a helper may never be fired/handled. I'll have to think about this.

lminiero:
See the PR above: as I wrote there, I don't have time to test this, so please make sure that 1. it doesn't crash anymore, 2. there's no new memory leaks.

lminiero:
Merged the fix, so I'll close.

PR #2216, opened 2020-06-09
Increase reference to session when handling SIP calls (see #2188)

lminiero:
This is supposed to help with the occasional crashes happening in the SIP plugin when using helpers, described in #2188.

The root cause there is that the `janus_sip_session` associated to a helper may be freed (helper closed), but it still "lives" in the Sofia SIP loop thread, e.g., because there are still events associated to it. This happens when we a call is originated via `nua_invite` on a helper, or when we pass an incoming call to a helper and so bind it to the helper session using `nua_bind_handle`: this causes all events associated to that call to include a pointer to that session, as the `hmagic`. What seems to be happening is that the call still has some state that can trigger events: since the Sofia loop belongs to the master session, it still runs, and so can still ship those events; trying to dereference the now freed helper session causes the crash.

This patch tries to address that by using reference counters to ensure the session is not destroyed until the calls it handles are gone. Specifically, we add a new reference when starting a new call, and when getting a new one: when helpers are involved, we make sure it's their reference that is updated. The reference is then only decreased when we get the `nua_i_terminated` callback, which means the call is definitely over as far as the Sofia SIP stack is concerned.

I tested briefly and it seems to be working. That said, I didn't test intensively: no time for that. As such, if you use the SIP plugin, make sure to test this properly. Notice that you should not only make sure it doesn't crash where it did before, but ideally also check whether this introduces leaks: to do that, please build with libasan support, which will print a summary of the leaked memory when you shut Janus down cleanly. You can also uncomment `REFCOUNT_DEBUG` in `refcount.h` to more specifically track if there is any reference still dangling when Janus is shut down cleanly.

The sooner you can provide me feedback on that (and why not, fixes if you find anything broken), the sooner we can fix this.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -4603,9 +4603,8 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 					/* Bind the call to the helper and handle it there */
 					JANUS_LOG(LOG_VERB, "Passing INVITE to helper %p\n", helper);
 					nua_handle_bind(nh, helper);
-					/* Pass the reference for this call to this helper */
+					/* This session won't need the reference anymore, the helper will */
 					janus_refcount_decrease(&session->ref);
-					janus_refcount_increase(&helper->ref);
 					janus_sip_sofia_callback(event, status, phrase, nua, magic, nh, helper, sip, tags);
 					break;
 				}
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -3416,6 +3416,7 @@ static void *janus_sip_handler(void *data) {
 			g_hash_table_insert(callids, session->callid, session);
 			janus_mutex_unlock(&sessions_mutex);
 			g_atomic_int_set(&session->establishing, 1);
+			janus_refcount_increase(&session->ref);	/* Add a reference for this call */
 			nua_invite(session->stack->s_nh_i,
 				SIPTAG_FROM_STR(from_hdr),
 				SIPTAG_TO_STR(uri_text),
@@ -4514,9 +4515,12 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				janus_sip_call_update_status(session, janus_sip_call_status_incall);
 			}
 			break;
-		case nua_i_terminated:
+		case nua_i_terminated: {
 			JANUS_LOG(LOG_VERB, "[%s][%s]: %d %s\n", session->account.username, nua_event_name(event), status, phrase ? phrase : "??");
+			/* We had a reference to this session for this call, get rid of it */
+			janus_refcount_decrease(&session->ref);
 			break;
+		}
 	/* SIP requests */
 		case nua_i_ack: {
 			JANUS_LOG(LOG_VERB, "[%s][%s]: %d %s\n", session->account.username, nua_event_name(event), status, phrase ? phrase : "??");
@@ -4548,6 +4552,7 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 		}
 		case nua_i_invite: {
 			JANUS_LOG(LOG_VERB, "[%s][%s]: %d %s\n", session->account.username, nua_event_name(event), status, phrase ? phrase : "??");
+			janus_refcount_increase(&session->ref);	/* Add a reference for this call */
 			if(ssip == NULL) {
 				JANUS_LOG(LOG_ERR, "\tInvalid SIP stack\n");
 				nua_respond(nh, 500, sip_status_phrase(500), TAG_END());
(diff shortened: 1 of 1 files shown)


zaltar:
I was able to test out this patch and it did prevent the crashing, however I found two memory leaks.

The first is line 4608 in `janus_sip.c`.  I don't believe this refcount increase is needed on the helper since the call to janus_sip_sofia_callback will also increase it.

The second issue I found is when the browser disconnects from Janus while a call is in progress.  A `nua_i_terminated` event for the call never occurs so the refcount on the session never decreases. 

Let me know if you need additional logs for these leaks or any other info. Thanks for the help with this issue.

lminiero:
> The first is line 4608 in `janus_sip.c`. I don't believe this refcount increase is needed on the helper since the call to janus_sip_sofia_callback will also increase it.

Good catch! I'll fix that.

> The second issue I found is when the browser disconnects from Janus while a call is in progress. A `nua_i_terminated` event for the call never occurs so the refcount on the session never decreases.

That's what I feared would happen, in my comment on the issue page: the loop being stopped before all events are stopped. Not sure what the proper way to handle this will be, since it will probably mean adding some complex state management. I'll have to think about it.

lminiero:
@zaltar can you check this update?

I tried to use the `nua_r_shutdown` event for the purpose, which is notified by the loop when a NUA shutdown we requested has been processed. The documentation says this gets rid of all calls, dialogs, etc. associated with that stack, but unfortunately that doesn't result in the related events to be sent before `nua_r_shutdown`, or at all actually. As such, I had to add a "manual" tracking of call-related references in the master session: any time we increase a reference for a session due to a call, we keep track of it (if it's a helper, all tracking is done in the master), and similarly when we decrease the reference we remove it from our list as well. When we get the `nua_r_shutdown`, if we still have call-related references pending (which would never be cleaned up by a `nua_i_terminated`) we get rid of them ourselves.

I tested this briefly and it seems to be working as expected, but you may want to stress it a bit more, especially with helpers involved. On helpers, I wonder if the same issue that is happening with calls can happen with subscriptions as well (e.g., if the helper sent a subscribe or originated a transfer), but we can worry about that another day.

zaltar:
Unfortunately the crashes are now back.  
AddressSanitizer output: https://pastebin.com/mpZ8V8Ei
Tail of the Janus log right before a crash: https://pastebin.com/bkRLseS2

From what I can tell, `janus_sip_destroy_session` is setting `session->master` to `NULL` which then causes the wrong condition to be executed in the `nua_i_terminated` case (`janus_sip.c:4534`). This causes the helper session to remain in master's `active_calls` list but the helper session's refcount is decremented. Later on when sessions remaining in `active_calls` are released in the `nua_r_shutdown` case (`janus_sip.c:5064`) the helper session has already been freed.

lminiero:
I guess one way to address that is by only unreffing if the session was in the list we're removing it from. That should solve the crash. Not sure if it will introduce a leak, though: maybe we shouldn't set `session->master` to `NULL` for helpers, so that the info can still be used until the session is really freed. I'll try and come up with a patch later.

lminiero:
@zaltar pushed a new tentative fix.

zaltar:
I think we're very close to a fix.  Locally I just had to modify one thing that AddressSanitizer caught.  In `janus_sip_unref_active_call` the `janus_refcount_decrease` call at (`janus_sip.c:1680`) could cause `session` itself to be freed and then the access to `session->master` causes a use after free bug.  I was able to fix this by storing `session->master` in a local variable.  With that one change I did not see any more crashes or leaks/bugs in AddressSanitizer in testing.

Thanks for helping with this!

ayty-adrianomartins:
> I think we're very close to a fix. Locally I just had to modify one thing that AddressSanitizer caught. In `janus_sip_unref_active_call` the `janus_refcount_decrease` call at (`janus_sip.c:1680`) could cause `session` itself to be freed and then the access to `session->master` causes a use after free bug. I was able to fix this by storing `session->master` in a local variable. With that one change I did not see any more crashes or leaks/bugs in AddressSanitizer in testing.
> 
> Thanks for helping with this!

Something like this?

```
static void janus_sip_unref_active_call(janus_sip_session *session) {
	if(session == NULL)
		return;
	janus_sip_session *master = session->master;
	if(master) {
		janus_mutex_lock(&master->mutex);
		if(g_list_find(master->active_calls, session) != NULL) {
			master->active_calls = g_list_remove(master->active_calls, session);
		}
		janus_mutex_unlock(&master->mutex);
	} else {
		if(g_list_find(session->active_calls, session) != NULL) {
			session->active_calls = g_list_remove(session->active_calls, session);
		}
		janus_mutex_unlock(&session->mutex);
	}
}
```

lminiero:
Ok, done, thanks for the feedback.

lminiero:
Since @zaltar mentioned that with that fix, he couldn't replicate crashes or leaks anymore, I'll merge and close the original issue. Thanks for the help!

ayty-adrianomartins:
It crashed for me again, unfortunately for me it only occurs in production, and it's more difficult to debug, the Janus itself only crash and not generate any log, the only thing I can get for now is the kernel log:

This was one crash:
`Jun 16 08:30:28 tup02aty011 kernel: [2713460.950748] sip handler[24203]: segfault at 120 ip 00007f5e93dcdea0 sp 00007f5e910fdfa8 error 4 in libpthread-2.28.so[7f5e93dc8000+f000]`

This was another:
`Jun 16 08:50:02 tup02aty011 kernel: [2714635.141565] traps: sip handler[24178] general protection ip:7f74ba1ee7f0 sp:7f74ba0c4f20 error:0 in libsofia-sip-ua.so.0.6.0[7f74ba129000+117000]`

I'll see if I can generate a proper debug log.

Issue #2192, opened 2020-06-01
janus_rtp_header_extension_parse_audio_level state problem


maxboehm:
As discussed in Pull Request #2172, we are currently trying to detect the audio level via janus.

After that pull request, we stumbled upon the next interesting behaviour. I invested quite some time, but have no final result.

So far:
**Case a)**
1. join (audio=false,video=false)
2. configure(audio=true)
-> desired behaviour

**Case b)**
1. join (audio=false,video=false)
2. configure(audio=false,video=true)
3. configure(audio=true,video=true)
-> It does not work any more, no audio level is found. This is happening in Chrome as well as in Firefox. Somehow, depending which of both I share first, the audio level recognition does not work anymore.

I digged into the function _janus_rtp_header_extension_parse_audio_level_.
- Case a): "id"=1
- Case b): "id" = -2, after changing that manually to 1, I get the desired results.

ID is determined by "stream->audiolevel_ext_id" in ice.c:2551. 

At this point, I dropped out. Any ideas?

Thank you!

lminiero:
`stream->audiolevel_ext_id` is set in `janus.c`:
https://github.com/meetecho/janus-gateway/blob/master/janus.c#L1412
using `janus_rtp_header_extension_get_id` to find the extension ID in the new SDP. If you see `-2` it means the extension was not found in the SDP:
https://github.com/meetecho/janus-gateway/blob/master/rtp.c#L77
As such, you should check if the SDP still has the related extmap.

maxboehm:
Thank you for the hints, will do that next week as I am on the road this week.

maxboehm:
Got it.

If I enable the mic after the webcam, the method `janus_rtp_header_extension_get_id` **will not** be called. It is rather a conditional problem: I get the log "Negotiation update, checking what changed" after enabling the mic and found, that in this part of the [condition (janus.c:1426)]( https://github.com/meetecho/janus-gateway/blob/master/janus.c#L1426) the extension is not handled. 

I've modified [janus.c in my fork](https://github.com/maxboehm/janus-gateway/blob/master/janus.c#L1466) and the audio level recognition is working now. Just not sure about the consequences. What do you think?

lminiero:
It depends on the changes. Have you checked if it works as expected in the new multistream branch instead? Renegotiations should work "better" there.

maxboehm:
To be honest, I don't understand why it should behave differently in the multistream branch.

[...]

PR #2223, opened 2020-06-12
Check extensions after renegotiation

maxboehm:
Extensions were not checked after a renegotation did happen. This is a problem in the case, that sb first enables his/her webcam and afterwards enables his mic as the audio level extension is not recognised.

Also see issue #2192

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -1463,6 +1463,23 @@ int janus_process_incoming_request(janus_request *request) {
 							janus_dtls_srtp_create_sctp(stream->component->dtls);
 						}
 					}
+				// We also have to check the extensions, if we do a renegotiation
+				} else {
+					/* Check if the mid RTP extension is being negotiated */
+					handle->stream->mid_ext_id = janus_rtp_header_extension_get_id(jsep_sdp, JANUS_RTP_EXTMAP_MID);
+					/* Check if the RTP Stream ID extension is being negotiated */
+					handle->stream->rid_ext_id = janus_rtp_header_extension_get_id(jsep_sdp, JANUS_RTP_EXTMAP_RID);
+					handle->stream->ridrtx_ext_id = janus_rtp_header_extension_get_id(jsep_sdp, JANUS_RTP_EXTMAP_REPAIRED_RID);
+					/* Check if the audio level ID extension is being negotiated */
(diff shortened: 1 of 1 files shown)


janus.c
```diff
@@ -1463,6 +1463,23 @@ int janus_process_incoming_request(janus_request *request) {
 							janus_dtls_srtp_create_sctp(stream->component->dtls);
 						}
 					}
+				// We also have to check the extensions, if we do a renegotiation
```
lminiero, line 1466: Comments should be in the `/* */` form.

janus.c
```diff
@@ -1463,6 +1463,23 @@ int janus_process_incoming_request(janus_request *request) {
...
+					/* Check if the audio level ID extension is being negotiated */
+					handle->stream->audiolevel_ext_id = janus_rtp_header_extension_get_id(jsep_sdp, JANUS_RTP_EXTMAP_AUDIO_LEVEL);
+					/* Check if the video orientation ID extension is being negotiated */
+					handle->stream->videoorientation_ext_id = janus_rtp_header_extension_get_id(jsep_sdp, JANUS_RTP_EXTMAP_VIDEO_ORIENTATION);
+					/* Check if the frame marking ID extension is being negotiated */
+					handle->stream->framemarking_ext_id = janus_rtp_header_extension_get_id(jsep_sdp, JANUS_RTP_EXTMAP_FRAME_MARKING);
+					/* Check if transport wide CC is supported */
+					int transport_wide_cc_ext_id = janus_rtp_header_extension_get_id(jsep_sdp, JANUS_RTP_EXTMAP_TRANSPORT_WIDE_CC);
+					handle->stream->do_transport_wide_cc = transport_wide_cc_ext_id > 0 ? TRUE : FALSE;
+					handle->stream->transport_wide_cc_ext_id = transport_wide_cc_ext_id;
```
lminiero, line 1482: This whole check is in the wrong place, since it's in an `#ifdef HAVE_SCTP`. This means it will never be used if data channels support via usrsctplib is not compiled. I guess it should just become a new `if(offer)` after the `#endif`, rather than an `else`.

lminiero:
Moved the check myself, so we can close this. Thanks for the heads up on the issue!

maxboehm:
Sorry, had not found time yet to resolve it...

Thank you, Lorenzo!

PR #2225, opened 2020-06-12
Allow simulcast ports to be picked randomly in Streaming mountpoints

lminiero:
Normally, when you pass `0` as a port to create a Streaming plugin mountpoint, this means a random port will be picked in the configured RTP range (the plugin's rane, not the core's). This is currently not true for the simulcast ports, though, that is `videoport2` and `videoport3`, which are considered to be disabled instead when you pass `0`. This patch changes that, and uses random ports for them when `videosimulcast=true` and they're not provided or set to 0.

This means you can't create a simulcast mountpoints with just two substreams anymore, but I don't think it's that big of a deal (not sure anyone was doing that anyway). I don't think this is controversial, so I plan to merge soon.


lminiero:
Merging.

PR #2230, opened 2020-06-15
Edit callback.error returning

TheSidSpears:
There are many places where callbacks.error return `string` and many where return `Error`. And  the only place, where callbacks.error return `string, Error`. That's why it needed to write redundancy conditions for right handling errors from createOffer(). But we can return only `Error` object like for `getUserMedia()` to avoid this

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -2454,7 +2454,7 @@ function Janus(gatewayCallbacks) {
 				})
 				.catch(function(error) {
 					pluginHandle.consentDialog(false);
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! Have you checked if this impacts some of our demos? Besides, there may be people using this out in the wild, which means this might be considered a breaking change.

lminiero:
Merging since, as you said, this provides more consistency. If this breaks something, people will let us know.

PR #2231, opened 2020-06-15
Fixed broken simulcast behaviour

lminiero:
While simulcast works mostly fine in plugins that support it, it was broken in some particular cases, in particular:

1. if only substreams 2 and 0 are sent, and 2 suddently goes away, it should fallback to 0 since 1 is missing; instead, it stays in the 1 limbo forever;
2. if simulcast drops to a lower layer (e.g., 1) because the target (e.g., 2) is unavailable, it doesn't go back to the target layer when it becomes available again (drop actually becomes the new target).

This PR tries to fix both those issues, making the code smarter when it comes to transitions. All the changes are in the core, so they should be used automatically by all plugins. There's also a fix on keyframe detection for VP8, which was broken in some cases (something we fixed in the postprocessor already, but not here). I could replicate both issues easily with the Streaming plugin, but I can't now, so I consider this fixed.

Planning to merge soon, so please let me know if you have feedback before that.

### DIFF
--- a/rtp.c
+++ b/rtp.c
@@ -921,6 +921,7 @@ void janus_rtp_simulcasting_context_reset(janus_rtp_simulcasting_context *contex
 	memset(context, 0, sizeof(*context));
 	context->rid_ext_id = -1;
 	context->substream = -1;
+	context->substream_target_temp = -1;
 	context->templayer = -1;
 }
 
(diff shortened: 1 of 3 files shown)


lminiero:
Merging.

PR #2236, opened 2020-06-18
Refactor videoroom hangup media internal.

atoppi:
This PR addresses a crash that occurred when a videoroom subscriber session gets quickly destroyed while being established.

When this race condition happens an invalid (freed) subscriber session gets appended to the subscribers list tied to a specific publisher. Once that publisher leaves, Janus traverses the list and references again that freed session, making the process halt.

This patch basically introduces a refactoring of the hangup media methods that will avoid appending an invalid session to the publisher list.

Notice that the signature of `janus_videoroom_hangup_media_internal` has been changed and the method now accepts a `gpointer` (aka `void*`) in place of a `janus_plugin_session*`.


lminiero:
Merged, as this was tested extensively already.

PR #2240, opened 2020-06-21
fix redundant condition

willamowius:
fixing some redundant condition checks in js


lminiero:
Actually those redundant checks were typos, as they were all supposed to be something like `(if X !== undefined && X !== null)`. I recently cleaned up all those checks to just leave `if(X)`, so can you do the same for the ones you fixed here? Thanks!

willamowius:
Updated

lminiero:
Thanks for the quick fixes, merging!

Issue #2220, opened 2020-06-11
Watch button in streaming demo needs to be clicked twice on mobile


willamowius:
The "Watch or Listen" button in the streaming demo needs to be clicked twice to work after you change the stream when you use it on mobile (tested Chrome on Android and Safari on iOS). On Desktop browsers it works the first time.

On mobile if you pick a  channel, then click (tap) "Watch or Listen" nothing happens. If you change the channel again, the next click won't work either, so its probably the updateStreamsList() function in streamingtest.js, but even after experimenting with it quite a while I have no clue, why.

lminiero:
To be perfectly honest, I have zero time (and interest) to debug this :slightly_smiling_face: 
If you can find out the cause, happy to hear what it is.

vincentfretin:
Generally speaking, on touch devices, if you need two touches to click on an element and on desktop only one click, it usually means the first touch is for the hover behavior, the second if for the real click. You may look at if there is a hover behavior with css or js on this button.

lminiero:
I know it's silly to come back to an issue two years after it was opened, but I finally found out what the problem was :hand_over_mouth: 
Apparently, the cause was an item with class `dropdown-backdrop` not being removed after we selected a stream from the list: this item was intercepting the click, and that's why you needed to click twice. As a test while debugging Firefox mobile, I've added code to remove that element when we select a stream, and the problem was fixed, so I'll push it shortly on both `master` and `0.x`.

PR #2241, opened 2020-06-21
fix Bootstrap issue on mobile that requires a double tap on Watch button

willamowius:
fixing issue #2220

### DIFF
--- a/html/streamingtest.html
+++ b/html/streamingtest.html
@@ -76,6 +76,7 @@
 								<div class="btn-group btn-group-sm">
 									<button class="btn btn-primary" id="watch">Watch or Listen</button>
 									<select class="btn btn-default btn-space" id="streamset">
+									<option value='invalid'>Streams list</option>
 									</select>
 								</div>
 							</div>
--- a/html/streamingtest.html
+++ b/html/streamingtest.html
@@ -74,14 +74,9 @@
 							</div>
 							<div class="panel-body" id="list">
 								<div class="btn-group btn-group-sm">
-									<button class="btn btn-primary" autocomplete="off" id="watch">Watch or Listen</button>
-									<div class="btn-group btn-group-sm">
-										<button autocomplete="off" id="streamset" class="btn btn-default dropdown-toggle" data-toggle="dropdown">
-											Streams list<span class="caret"></span>
-										</button>
-										<ul id="streamslist" class="dropdown-menu" role="menu">
-										</ul>
-									</div>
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks! I'll have a look at the changes later and review :+1:

lminiero:
I'm afraid this change breaks the UI. First of all, it starts with one of the streams already selected, rather than showing the placeholder/guidance we have now. Besides, the look&feel is broken too, because the form is not using the Bootstrap CSS. As such, I don't think this can be merged as it is.

willamowius:
I'd consider not having an invalid choice that always produces an error in the list as a feature, not a bug, but if you prefer it that way, I have added one.

I don't see how the new version is not using Bootstrap CSS. Bootstrap is about styling HTML which this version does. That the old version was mimicking an HTML Select with with other elements is not a feature you should want to keep, especially since it didn't work on any mobile browser (neither Android nor iOS).

lminiero:
> I'd consider not having an invalid choice that always produces an error in the list as a feature, not a bug, but if you prefer it that way, I have added one.

It is a feature to me. If people see a title right away, they might thing that's all there is. An "invalid choice", as you call it (I call it a guide) tells them to check whats available. We have the same in the SIP plugin, where we DO want people to see the different ways they can register. Our demos may be ugly as hell, but we did give some thought to them, so please don't change existing behaviours.

> I don't see how the new version is not using Bootstrap CSS. Bootstrap is about styling HTML which this version does. That the old version was mimicking an HTML Select with with other elements is not a feature you should want to keep, especially since it didn't work on any mobile browser (neither Android nor iOS).

Well, it isn't. This is how it looks now:

select1

This is how it looks after your fix:

select2

Again, I'd never call our demos sleek or pretty, not at all, but I definitely don't want them to look even uglier than they already are. Bootstrap styling is definitely not happening for the new `select` you added, so that should be fixed before I can merge this.

lminiero:
Notice how the "invalid choice" doesn't appear in the select, because it's really not supposed to be part of the choices. This is how Bootstrap styles drop-down menus, and why we used that approach for the selector in here as well.

willamowius:
Then I'm afraid you need to find somebody else to fix your style of select on mobile. I didn't see a way to do it and not having mobile working isn't an option for me.

lminiero:
No problem, we have the issue open, so either us or someone else will hopefully find time to look into it sooner or later. Thanks anyway!

Issue #2117, opened 2020-04-29
maybe suport nng as replacement for nanomsg


jonassmedegaard:
nanomsg has had no new release since October 2018, and no code changes at Github since July 2019. Their homepage prominently states this:
> This project has largely been superceded by the nng project.

[nng](https://github.com/nanomsg/nng) states this:
> This project offers both wire compatibility and API compatibility, so most nanomsg users can begin using NNG right away.
>
> Existing nanomsg and mangos applications can inter-operate with NNG applications automatically.

Seems relevant to explore support for nng in addition to (or replacement for) nanomsg.

lminiero:
Not planning to do that, sorry. If anyone wants to take the challenge, we welcome pull requests.

lminiero:
@jonassmedegaard FYI, I played a bit with this in the weekend. Unlike their claims, they changed quite a lot in NNG: the compatibility layer itself has many constraint, so isn't worth the trouble. As such, I decided to try updating the code to just use NNG itself, instead: it was not an easy task, and compiles correctly now, but still doesn't seem to work.

The main change is in their asynchronous management: they got rid of POSIX APIs, for a much more convoluted and obscure (to me) AIO mechanism. While they provide you with a way to get a "fake" descriptor associated to their sockets that you can use in a poll yourself (unlike the `nn_poll` they had previously), it doesn't seem to work, or at least not consistently. The "writable" part in particular seems broken. Refactoring the whole plugin to use their new AIO stuff seems way too much work for the trouble, especially considering I didn't find good documentation on how it's supposed to work.

More importantly, it doesn't look like there are many NNG bindings out there. I wanted to use a simple nodejs application for testing, and the only NNG binding available is an unofficial one that won't even compile on newer (>= 12) nodejs deployments. Not sure about bindings to other languages, but the lack of reliable nodejs bindings is kind of a big deal to me. I tried to write a barebone NNG application in C to talk to, but of course if I'm doing something wrong in the plugin code, I'm making the same mistakes there, so there's no way for me to know what I should fix or how.

As such, while I can share the effort I worked on so far as a (very) WIP, I don't think it would be merged anytime soon anyway.

jonassmedegaard:
Oh my.  Sorry I lured you down that rabbit hole :-(

Thanks for sharing the efforts and criticism.

lminiero:
[...]

PR #2242, opened 2020-06-22
Move Nanomsg to NNG (see #2117)

lminiero:
This is an attempt to address #2117. As discussed there, the Nanomsg developers will not keep on updating/improving Nanomsg anymore, and have moved all their efforts to NNG instead (Nanomsg Next Generation). While still incomplete and currently broken, this patch is an attempt to convert both the Nanomsg trasport plugin and the Nanomsg event handler to use NNG instead.

As I explained in a comment on that issue, Nanomsg and NNG are actually quite different, so a 1-1 translation wasn't really possible. A key change was made on the way asynchronous operations can be performed: where before they provided an `nn_poll` API that could be used pretty much as `poll` for knowing when you could read/write from their sockets, that has now gone. They're using a completely different pattern, now, called [nng_aio](https://nng.nanomsg.org/man/tip/nng_aio.5.html) to perform asynchronous operations instead. The developers insist that this new approach is much more efficient, and I'm sure it is: unfortunately, it's not very documented apparently, or at least I couldn't find any guidance on how to use it effectively. Any discussion on how to move from polling to that points to the link I shared above, where since I'm dumb I couldn't find any actually helpful information.

NNG still partly supports `poll`, though, via some kind of a hack. Using a method called `nng_getopt` you can get a "fake" file descriptor associated to their sockets (which are now not numeric anymore), and that you can in theory use in a regular `poll` to know when their sockets are readabe/writable. Unfortunately, it doesn't seem to be working as expected for me, and I don't know why: I wanted to test it against something more high level (a nodejs application), but the only NNG binding available there won't compile on node>=12, so I can't use it. Good chances are I'm doing something very wrong (very likely since I tried to reuse the communication patterns from the original plugins), but since I don't have any way to test this, I have no way to know.

As such, I'll stop here, and I'm sharing this just in case someone more familiar with NNG can have a look and possibly help improve/fix it (which is why I'm publishing as a draft, rather than a regular PR). I also welcome feedback, especially if you're using Nanomsg in Janus today, on whether you think this effort is needed/helpful in the first place.


conf/janus.eventhandler.nngevh.jcfg.sample
```diff
@@ -1,9 +1,9 @@
-# This configures the Nanomsg event handler. Since this plugin only
-# forwards each event it receives via Nanomsg, you simply need to
+# This configures the NNG event handler. Since this plugin only
+# forwards each event it receives via NNG, you simply need to
 # configure (i) which events to subscribe to, (ii) the address to use for
 # the communication, and (iii) whether the address should be used to bind
 # locally or to connect to a remote endpoint. Notice that the only supported
-# pattern is NN_PUBSUB, where the Nanomsg event handler is the publisher.
+# pattern is NN_PUBSUB, where the NNG event handler is the publisher.
```
hugolm84, line 6: This should probably say NNG_PAIR, https://github.com/meetecho/janus-gateway/pull/2242/files#diff-5b84e446e3e9cf404a19e5947bef4edfR205

lminiero:
There doesn't seem to be much interest in this, so no point in keeping this effort going and I'll close it.

PR #2247, opened 2020-06-24
Add more checks on validity of NUA before using it in SIP plugin

lminiero:
Apparently, there are some race conditions that can happen if the `s_nua` NUA object is destroyed after the stack loop ends, and requests to use it arrive after that. This is a first attempt to fix that, by adding more checks on the validity of the object before using it, taking advantage of the stack mutex for the purpose. I'm not able to replicate the issue myself (master works fine for me even when stress testing) so I can't check if this works: I'll wait for feedback before deciding whether this fix is enough, or if more needs to be done.


lminiero:
Looks like this patch fixed the issue, so I'll merge.

PR #2250, opened 2020-06-28
Fix opus silence potential to generate huge files

neilkinnish:
This may not be the most optimal way to address the issue and I may be missing something. But this seems to work and resolve all the issues I have encountered where huge (example: 28gb) opus files can be generated.

### DIFF
--- a/postprocessing/pp-opus.c
+++ b/postprocessing/pp-opus.c
@@ -64,13 +64,13 @@ int janus_pp_opus_create(char *destination, char *metadata) {
 	return 0;
 }
 
-int janus_pp_opus_process(FILE *file, janus_pp_frame_packet *list, int *working, uint32_t count) {
+int janus_pp_opus_process(FILE *file, janus_pp_frame_packet *list, int *working) {
 	if(!file || !list || !working)
 		return -1;
 	janus_pp_frame_packet *tmp = list;
 	long int offset = 0;
 	int bytes = 0, len = 0, steps = 0, last_seq = 0;
-	uint64_t pos = 0;
+	uint64_t pos = 0, nextPos = 0;
 	uint8_t *buffer = g_malloc0(1500);
 	while(*working && tmp != NULL) {
 		if(tmp->prev != NULL && ((tmp->ts - tmp->prev->ts)/48/20 > 1)) {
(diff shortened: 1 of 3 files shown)


neilkinnish:
Looking deeper at this I think I can improve, but I'd still be interested in opinion / downsides of the general approach.

lminiero:
I think it's already helpful to avoid those issues. Not sure about the impact on recordings, as sometimes they're caused by timestamp resets at unfortunate places and may mean some missing audio packets when the additional command line flags are not used, but it should be very few of them anyway. This looks fine to me already, but if you think this can be improved before we merge please do let me know.

neilkinnish:
Updated to use `tmp->next` rather than pass the count.

neilkinnish:
> Not sure about the impact on recordings, as sometimes they're caused by timestamp resets at unfortunate places and may mean some missing audio packets when the additional command line flags are not used

I have tested across audio recordings that are fine and also 3 with issues.

It definitely resolves the issue of massive files which can impact the server / service quite badly and that I think is key (at least for me).

In all the results the audio was fine and playable. In one case of the 3 with issues it reported an issue with `granulepos` but played back okay and could be trasncoded to another format fine - I'm assuming this is because it didn't resolve some data issues with PTS at the beginning. Using -i and setting an arbitrary `n` also works and resolved the issue as it drops the frames with the issue I suppose, but you have to kind of guess the `n` and I would prefer not to.

postprocessing/pp-opus.c
```diff
@@ -85,6 +85,15 @@ int janus_pp_opus_process(FILE *file, janus_pp_frame_packet *list, int *working)
 			int i=0;
 			for(i=0; i<silence_count; i++) {
 				pos = (tmp->prev->ts - list->ts) / 48 / 20 + i + 1;
+				if(tmp->next == NULL) {
+					JANUS_LOG(LOG_WARN, "[SKIP] pos: %06" SCNu64 ", skipping remaining silence, next is NULL\n", pos);
```
lminiero, line 89: I'm not sure this is ok. You're dropping the last audio packet, this way.

neilkinnish: Adjusted in next commit

lminiero:
Thanks! I'll have to double check how this behaves with "normal" losses to compensate with silence, since they'd be much more frequent than GB files and we don't want to break those.

neilkinnish:
@lminiero 👍

lminiero:
Seems to be working as the current `janus-pp-rec` on an Opus recording with some occasional losses, so this looks good to me. Thanks, merging! :+1:

PR #2252, opened 2020-06-29
Post processing ratio/dimension handling

neilkinnish:
When post processing a WEBM video recording with multi-resolutions the post processor can end up setting the dimensions to equal sizes largely because it currently takes the largest of each as it encounters.

This can affect mobile recordings more obviously where it registers rotation that didn't happen but I have seen it happen on desktop frequently where the final dimensions are sent through reversed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -593,6 +593,8 @@ janus_pp_rec_SOURCES = \
 	postprocessing/pp-rtp.h \
 	postprocessing/pp-srt.c \
 	postprocessing/pp-srt.h \
+	postprocessing/pp-utils.c \
+	postprocessing/pp-utils.h \
 	postprocessing/pp-webm.c \
 	postprocessing/pp-webm.h \
 	postprocessing/janus-pp-rec.c \
--- a/postprocessing/pp-h264.c
+++ b/postprocessing/pp-h264.c
@@ -24,6 +24,7 @@
 
 #include "pp-h264.h"
 #include "../debug.h"
+#include "pp-utils.h"
 
 
 #define LIBAVCODEC_VER_AT_LEAST(major, minor) \
@@ -251,6 +252,8 @@ static void janus_pp_h264_parse_sps(char *buffer, int *width, int *height) {
 int janus_pp_h264_preprocess(FILE *file, janus_pp_frame_packet *list) {
 	if(!file || !list)
 		return -1;
+	dimension_array dimension_arr;
+	init_dimension_array(&dimension_arr, 1);
 	janus_pp_frame_packet *tmp = list;
 	int bytes = 0, min_ts_diff = 0, max_ts_diff = 0;
 	int rotation = -1;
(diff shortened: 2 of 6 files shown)


lminiero:
I'm not sure that's the way to fix it: seems way too convoluted. As a side note, it's not a webm only thing: whatever fix will be applied, needs to be added to the H.264/mp4 code too.

neilkinnish:
@lminiero thinking further on this... another approach is to add the width and height to arrays as we process and then use the most frequent of both. I am working on an implementation of this, but let me know if you foresee any issues with this approach.

lminiero:
@neilkinnish do you have a recording you can share with these dynamically changing resolutions?

neilkinnish:
I'll generate some.

lminiero:
@neilkinnish any update on those samples?

neilkinnish:
@lminiero I've had trouble getting permission, but...

1. I have a simplified fix almost done.
2. I can supply example frame details for ref.

lminiero:
This seems extremely convoluted... way too more than needed, IMHO.

neilkinnish:
@lminiero I've adjusted the handling, I'm not sure if you will approve of my implementation, but hopefully it demonstrates the issue.

It essentially just checks the dimensions going through and has a sequence/frame threshold. For the dimension to be applied it has to be above the threshold (seq/frame). This stops random or incorrect dimensions which are reported from being applied.

I will try and get example source for you to try, I just need permission for the ones I am using.

neilkinnish:
Hah just saw your reply after I posted

> This seems extremely convoluted... way too more than needed, IMHO.

Okay fair enough. I may be over complicating due to my experience of c (and lack of time). I'll try and get back to this when I have time. Feel free to close.

lminiero:
Didn't mean to downplay the effort and contribution, which is really appreciated! It just looks much more code than I think would be needed, and honestly looks more than I'm willing to mantain. If you can share some mjr files that show the issue and that your code fixes, I'll try to come up with a simpler (and probably more naive) implementation. If the main problem is rotation, my feeling is that a couple of additional checks would be all that's needed, but I may be underestimating the issue (which is why the mjr files may help). In case it turns out it's indeed more complex than that, we can come back to your PR, so no need to close it for now.

neilkinnish:
Don't worry honestly, I know its not perfect. I've been trying to keep it simple and honestly I don't have masses of c/c++ experience so I know I may be over engineering due to lack of knowledge of more simple options. But actually it helps to go through this process.

I will try and get some files for you... do you have an option to share privately?

lminiero:
My gmail address is the same as my github username, so you can send it there. Thx!

tgabi333:
Any update on this?

lminiero:
I started looking at this and then got sidetracked. Now we have the IETF meeting which will keep us busy for the whole week, so I for sure won't be able to do anything until it ends, sorry.

neilkinnish:
@lminiero I'm just taking another look at this.

Regarding...

```
vEncoder = avcodec_alloc_context3(codec);
vEncoder->width = max_width;
vEncoder->height = max_height;
```

When processing VP9 it seems to ignore the resolution set (width and height) – but it works correctly when VP8. 

For example I have a file which reports multiple resolutions of 960x540 and 1920x1080. If 1920x1080 is set it seems to ignore and use 960x540. I'm trying to find more information on the reason behind this but was thinking you may know the answer and be able to point me in the right direction.

neilkinnish:
Looking a little deeper and the VP9 not using the supplied max width & height, I'm assuming it's because of the frame changes and would need to re-initiate the AVCodecContext each time or add some kind of filter graph, I could be way off.

lminiero:
Almost forgot about this... anyway, the resolution you set there doesn't matter much, AFAIK: it's simply what ends up being written in the container header, and players are already aware of resolution changes as they're part of the information provided with each keyframe. It's only an issue when players rely on that to allocate the buffers they need for handling frames, and so may crash when a larger resolution than the one they were expecting arrives.

As such, rather than tracking all resolutions as your patch does now, it may make sense to only update the "max" info when `width*height` of the current frame is higher than the last `width*height` we took note of. This should simplify the process greatly. If that makes sense for you, I can prepare a quick PR that does that, so that you can check if it fixes your issues.

neilkinnish:
@lminiero sounds good 👍

lminiero:
@neilkinnish please give the PR above a go, to see if it fixes your original issue.

neilkinnish:
@lminiero will test over the weekend/monday and comment in the PR, thanks!

lminiero:
Closing as addressed in a different PR.

PR #2253, opened 2020-06-29
Early add a reference to a subscriber in videoroom handler.

atoppi:
This PR tries to fix a race condition that might occur in the videoroom plugin when a subscriber gets destroyed while the plugin thread is handling a subscriber message.

The proposed fix is to early add a reference to the subscriber in  the message handler while the `sessions_mutex` is stil taken by the thread. This should avoid `heap-use-after-free` errors when trying to access a freed subscriber later in the handler code.


lminiero:
Merging.

PR #2257, opened 2020-06-30
Fix documentation for mute_room/unmute_room

willamowius:
The documentation has a copy/paste bug


lminiero:

PR #2260, opened 2020-07-01
Rnnoise, neural network noise canceling - AudioBridge

mirkobrankovic:
Based upon [rnnoise](https://github.com/xiph/rnnoise) library, and only as a poc, to see if it makes sense to use it or not.
Instalation instructions are updated in README.md and also audiobridge usage.

`rnnoise_complexity` room parameter is actually how often frames should go trough rnnoise.
0 - disabled
1 - every second frame will be denoised
2 - 2 frames will be denoised, third skipped
3 - every frame will be denoised
This is an option since denoising will decrees audio quality (add a bit robotic flavor or even sometimes cut end of words, recognizing them as noise)

The rnnoise readme states `It operates on RAW 16-bit (machine endian) mono PCM files sampled at 48 kHz` and from code example I see it does read from a file into frames and process them one by one. I only managed to get clear audio on 16000 sampling rate room.

It would be nice to get some feedback on how good it is, or even better if it can be made better in any way, since my understanding of sampling is very limited :+1: )

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -152,6 +152,7 @@ janus_CFLAGS = \
 	-DLOGGERDIR=\"$(loggerdir)\" \
 	-DCONFDIR=\"$(confdir)\" \
 	$(BORINGSSL_CFLAGS) \
+	$(RNNOISE_CFLAGS) \
 	$(NULL)
 
 janus_LDADD = \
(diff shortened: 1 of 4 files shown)


lminiero:
Sorry, but I'm not going to consider major changes like this, especially when they add new dependencies, until the multistream branch has been merged.

mirkobrankovic:
aah make sense, thanks :)

PR #2261, opened 2020-07-02
New demo to use canvas element with EchoTest plugin

lminiero:
Nothing fancy: just a new demo, again based on the EchoTest, to show how you can use a canvas element to manipulate the webcam video before you send it to Janus via WebRTC. Very basic, and probably not very efficient, but it should be enough for demo purposes.

This is actually the same code I used in an old [Kamailio World Dangerous Demo](https://youtu.be/zwYUojfm0hY?t=2140), that I documented more throroughly in a [dedicated blog post](https://www.meetecho.com/blog/firefox-webrtc-youtube-kinda/); that old demo used the canvas element to change the video we injected in a VideoRoom, and from there an RTP forwarder would feed an FFmpeg script to transcode it to RTMP and send it to YouTube. To keep it simple, I chose to just send it to the EchoTest and have it be bounced back instead, so that people could see the canvas editing in action in real-time.

Planning to merge soon as there's nothing new or groundbreaking here, hope you'll have fin with it!


lminiero:
Merging.

PR #2275, opened 2020-07-10
merging from upstream

groupboard:


lminiero:
Pretty sure that's not what you meant to do.

groupboard:
Yes, indeed. Just meant to merge the latest upstream changes with my own fork. Didn't think creating a pull request was the correct way, but that seemed to be the only option. Annoying github confusing UX :)

PR #2276, opened 2020-07-10
Check websocket readystate on destroy

groupboard:
Check websocket readyState before sending destroy message, otherwise it will generate a "InvalidStateError: The object is in an invalid state." exception if the websocket isn't in the open state.


html/janus.js
```diff
@@ -1108,7 +1108,13 @@ function Janus(gatewayCallbacks) {
...
 			ws.addEventListener('error', onUnbindError);
 
-			ws.send(JSON.stringify(request));
+                        if (ws.readyState === 1) {
+                            ws.send(JSON.stringify(request));
+                        }
+                        else {
+                            onUnbindError();
+                        }
+
```
lminiero, line 1117: Please use tabs, not spaces, for indentation. The `else` should be on the same line as the closing bracket too.

groupboard: Both fixed now.

lminiero: They don't seem fixed to me... I still see spaces used for indentation.

groupboard: Sorry, it was using 4-char indentation and I've changed it to 8 (so all tabs).

lminiero: It must not be char-indentation, only tab-indentation. I still see whitespaces.

lminiero: Edit: apologies, github was showing me an old thing... seems ok now! :+1:

lminiero:
Thanks, merging!

PR #2278, opened 2020-07-13
Improve building with BoringSSL

fancycode:
Fixes:
- DTLSv1.2 was not negotiated with BoringSSL
- Thread-locks are not necessary with BoringSSL


lminiero:
Thanks! This looks fine to me, merging then :+1:

PR #2281, opened 2020-07-13
Fix broken EchoTest demo for Firefox if datachannels are not supported

lminiero:
I was recently made aware that if you try to open the EchoTest demo with Firefox on a Janus that has been build without support for datachannels, the PeerConnection setup fails. I'm just mentioning the EchoTest here and in the commit/title for the sake of simplicity, but this actually is true for all PeerConnections where Firefox offers data channels and Janus doesn't support them (e.g., VideoRoom publishers sending data too). The cause is how we create the m-line to reject the media:

    m=application 0 UDP/DTLS/SCTP 0

which causes this error to appear on the Firefox JS console:

    WebRTC error: DOMException: SIPCC Failed to parse SDP: SDP Parse Error on line 53:  No webrtc-datachannel token in m= media line, parse failed.

It all works as expected in Chrome instead (haven't tried Safari, Edge, or mobile browsers).

This patch tries to address that, by ensuring that we always add the `webrtc-datachannel` format to the m-line even when we're rejecting the SDP, i.e.:

    m=application 0 UDP/DTLS/SCTP webrtc-datachannel

This seems to fix the broken PeerConnection in Firefox, and still works in Chrome, so I consider the issue fixed. I decided to push this as a pull request just in case you can find cases where this actually breaks something it shouldn't (as I mentioned, I didn't test extensively, also because I don't have access to some browsers at all). I plan to merge soon, though, so please notice I won't wait forever.

### DIFF
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -1504,7 +1504,6 @@ janus_sdp *janus_sdp_generate_answer(janus_sdp *offer, ...) {
 					char *fmt_str = (char *)fmt->data;
 					if(fmt_str)
 						am->fmts = g_list_append(am->fmts, g_strdup(fmt_str));
-					JANUS_LOG(LOG_FATAL, "[answer] %p, %p\n", am, am->fmts);
 				}
 				temp = temp->next;
 				continue;
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -430,7 +430,7 @@ janus_sdp *janus_sdp_parse(const char *sdp, char *error, size_t errlen) {
 						m->proto = g_strdup(proto);
 						m->direction = JANUS_SDP_SENDRECV;
 						m->c_ipv4 = TRUE;
-						if(m->port > 0) {
+						if(m->port > 0 || m->type == JANUS_SDP_APPLICATION) {
 							/* Now let's check the payload types/formats */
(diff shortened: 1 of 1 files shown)


sdp-utils.c
```diff
@@ -1498,6 +1498,14 @@ janus_sdp *janus_sdp_generate_answer(janus_sdp *offer, ...) {
 			if(!do_data || data > 1) {
 				/* Reject */
 				am->port = 0;
+				/* Add the format anyway, to keep Firefox happy */
+				GList *fmt = m->fmts;
+				if(fmt) {
+					char *fmt_str = (char *)fmt->data;
+					if(fmt_str)
+						am->fmts = g_list_append(am->fmts, g_strdup(fmt_str));
+					JANUS_LOG(LOG_FATAL, "[answer] %p, %p\n", am, am->fmts);
```
fancycode, line 1507: Is the log statement a debugging leftover? I'm not sure the pointers need to be logged as fatal message...

lminiero: Argh, good catch, I thought I had gotten rid of them all... It was just lazy debugging on my end :slightly_smiling_face:

lminiero: Fixed :+1:

lminiero:
Merging.

PR #2285, opened 2020-07-15
Pass along data channel label when sending packets.

fancycode:
Otherwise all packets will be received on the default "JanusDataChannel" channel.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1661,6 +1661,7 @@ typedef struct janus_videoroom_rtp_relay_packet {
 	gboolean svc;
 	janus_vp9_svc_info svc_info;
 	/* The following is only relevant for datachannels */
+	char *label;
 	gboolean textdata;
 } janus_videoroom_rtp_relay_packet;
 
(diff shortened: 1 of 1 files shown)


lminiero:
That's actually on purpose. The VideoRoom in master only supports receiving a single publisher via datachannel through a specific subscription, and so using labels is useless. This is already fixed, and actually used a lot, in the `multistream` branch instead, where the same datachannel relays multiple publishers at the same time: in that case, the unique publisher ID is used as a label. As such, I don't think I can merge this, sorry.

fancycode:
Too bad, because this is a regression from older versions where the behaviour was like this (publishers could send multiple data channels and they were received individually by the subscribers).

Would you accept a PR that adds an option to re-enable this behaviour?

lminiero:
I don't think it ever worked like this, because we never included the label in the VideoRoom relay struct (which is what your patch indeed does to fix it) and so we were most definitely not specifying a label on the way out. Which older version are you referring to?

fancycode:
I have a Janus 0.4.3 running here :see_no_evil: where the subscribers receive the data channel messages on a channel with the same label as the publisher sends them on.

lminiero:
We didn't even support labels back then :rofl:  Support was only added in #1551 which happened much later than 0.4.3, so whatever was happening back then sounds like a bug acting as a feature.

fancycode:
I see :laughing:, thanks for the information. So I'll have to figure out how to upgrade to the latest version with the (for me) changed behaviour.

PR #2273, opened 2020-07-09
Add MQTT v5 properties support

feymartynov:
This is related to #1872.
Sorry for taking a long time to send the PR. We had to postpone testing of this feature but now it's good to go.
Also this PR includes MQTT v5 support for the MQTT event handler in the similar way as for the MQTT transport.

PR #2286, opened 2020-07-16
Pass MQTT buffer settings to Paho

feymartynov:
This is out of scope of the [previous](https://github.com/meetecho/janus-gateway/pull/2273) PR so here is a separate one.

This feature allows passing a couple of options to Paho MQTT client through transport & evh configs: max inflight messages and max buffered messages. By default Paho sets the first one to 10 and the second one to 100 which may be insufficient for high load. So here're options to configure them.


lminiero:
Looks good to me, thanks for this new contribution! Merging :+1:

PR #2288, opened 2020-07-17
Fix candidates memory leaks

fancycode:
- Ignored peer-reflexive candidates were not released.
- All found local candidates after the one that was used were leaking.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3643,8 +3643,10 @@ void janus_ice_resend_trickles(janus_ice_handle *handle) {
 		handle->handle_id, g_slist_length(candidates), stream->stream_id, component->component_id);
 	for(i = candidates; i; i = i->next) {
 		NiceCandidate *c = (NiceCandidate *) i->data;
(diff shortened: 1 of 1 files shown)


lminiero:
Good catch, I thought `nice_agent_get_local_candidates` only allocated the list, but looking at the documentation it looks like the caller owns the candidate objects too. I wonder if this changed recently, though? I wouldn't want this to crash instances with older versions of libnice.

fancycode:
I don't think this is something new in libnice, most of the candidates returned by `nice_agent_get_local_candidates` were already freed in the existing code, e.g. here:
https://github.com/meetecho/janus-gateway/blob/3498214b85d4bc352aa9d516ef5b7b7d8dfa1c8a/ice.c#L2047
https://github.com/meetecho/janus-gateway/blob/3498214b85d4bc352aa9d516ef5b7b7d8dfa1c8a/ice.c#L3658

lminiero:
Ack, makes sense, merging then! Thanks :+1:

PR #2294, opened 2020-07-22
Fix memory leak in sdp-utils.c

Bug-Fairy:
GList returned from g_hash_table_get_keys() was never freed.


tmatth:
LGTM

lminiero:
Good catch, thanks! I wonder why libasan isn't catching these for me, though... I always build with libasan support exactly to try and find memory leaks when they appear. Anyway, merging :+1:

PR #2297, opened 2020-07-24
Fixed validation of subscriber_parameters with string_ids true in vid…

ZeXtrasJay:
It seems the field "feed" inside stream is always validated as a integer, while it could also be a string.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1670,7 +1670,6 @@ static struct janus_json_parameter subscriber_parameters[] = {
 	{"temporal_layer", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 };
 static struct janus_json_parameter subscriber_stream_parameters[] = {
-	{"feed", JANUS_JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE},
 	{"mid", JANUS_JSON_STRING, 0},
 	/* For VP8 (or H.264) simulcast */
 	{"substream", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -7233,6 +7233,13 @@ static void *janus_videoroom_handler(void *data) {
 					}
 					if(error_code != 0) {
 						janus_mutex_unlock(&videoroom->mutex);
+						/* Unref publishers we may have taken note of so far */
+						while(publishers) {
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @ZeXtrasJay! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Mh I don't think that's the way to go. We already have definitions of feed as integer or string:

```
static struct janus_json_parameter feedopt_parameters[] = {
	{"feed", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE}
};
static struct janus_json_parameter feedstropt_parameters[] = {
	{"feed", JSON_STRING, 0}
};
```

For all requests that involve a `feed` or `id`, we actually decouple those from all the other properties, in order to validate the feed/id first depending on what they are, and only then validate the rest. The same should be done for `subscriber_stream_parameters`, or you end up with huge duplicates like the one you created here.

ZeXtrasJay:
@lminiero I've updated the PR, but i'm not 100% sure it's what you meant

lminiero:
Actually I think the only thing you need to do in this patch is removing `feed` from the definition of the `subscriber_stream_parameters[]` properties: in fact, we already have the right checks in both the for loops (we check `subscriber_stream_parameters` first, and then `feed` separately to see if it's an integer or string).

ZeXtrasJay:
@lminiero
 I've simplified the patch, but i also found a memory leak (two commits)

plugins/janus_videoroom.c
```diff
@@ -1670,7 +1670,7 @@ static struct janus_json_parameter subscriber_parameters[] = {
 	{"temporal_layer", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 };
 static struct janus_json_parameter subscriber_stream_parameters[] = {
-	{"feed", JANUS_JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE},
+	//~ {"feed", JANUS_JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE},
```
lminiero, line 1673: Please remove the line entirely, rather than just commenting it out.

lminiero:
> I've simplified the patch, but i also found a memory leak (two commits)

Good catch, it was in the previous error management code but not here, thanks :+1:

ZeXtrasJay:
Removed the line :+1:

lminiero:
Thanks, merging then! :v:

PR #2298, opened 2020-07-25
Fix minor memory leak in participant inbuf of audiobridge plugin

JeckLabs:
According to glib documentation [g_list_remove_link](https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-remove-link) will not free element itself, which cause a minor memory leak in the audio bridge plugin. Probably it would be better to replace it with [g_list_delete_link](https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-delete-link).

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -1320,7 +1320,7 @@ static void janus_audiobridge_participant_free(const janus_refcount *participant
 	while(participant->inbuf) {
 		GList *first = g_list_first(participant->inbuf);
 		janus_audiobridge_rtp_relay_packet *pkt = (janus_audiobridge_rtp_relay_packet *)first->data;
-		participant->inbuf = g_list_remove_link(participant->inbuf, first);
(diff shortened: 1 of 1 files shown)


lminiero:
Have you verified if this is indeed leaking memory? It does make sense looking at the docs, but my libasan setup doesn't seem to be catching those when I close Janus and it cleans up.

atoppi:
glist links seems to be handled with memory slices, so maybe that could explain why libasan is not catching those leaks

lminiero:
Ah that would make sense. I'll make a couple of quick tests, then, and merge if I don't spot anything weird :+1:

JeckLabs:
@lminiero  yes, this is how I found it in the first place. I'm using Valgrind and it shows a leak in glib slice allocator. I may try to reproduce if it helps.

lminiero:
No need! Sounds good to me, thanks for spotting this: merging :+1:

PR #2302, opened 2020-07-29
Use unique IDs and internal hashtable to map SCTP associations with usrsctp

lminiero:
This patch aims at addressing a vulnerability that apparently still affects usrsctp, and so you should test this if you use datachannels in your application. You can learn more about the vulnerability (and the workaround we implemented) by reading [this thread on Twitter](https://twitter.com/juberti/status/1288218310941773824).

Basically, with usrsctp almost everyone (including us, libwebrtc developers, and others) passed our custom structures/classes as opaque pointers when creating associations, as this way usrsctp would provide the same pointer back in callbacks, making it easier to address, e.g., incoming messages and events and relate them to the right instance. Apparently, though, usrsctp is actually putting the address of that pointer in SCTP messages (as part of the cookie), which creates an obvious vulnerability. I was made aware of this issue some time ago, but at the same time the notice said this had been fixed in usrsctp itself about a year ago: since we always recommend installing the latest version, I assumed we were fine, but apparently that's not the case. As such, I'm implementing the same workaround the libwebrtc developers implemented, that is an internal map with unique IDs: we pass these harmless IDs to usrsctp, and then look for the right SCTP instance when we get an ID in one of the callbacks.

While it seems to be working as expected in the few tests I made, you're encouraged to test this more thoroughly yourself, especially if you rely on datachannels in your applications. This new map I added also adds a new mutex, and as such there's always the risk of deadlocks in unexpected circumstances (which should happen, but you never know).

### DIFF
--- a/sctp.c
+++ b/sctp.c
@@ -119,6 +119,13 @@ void janus_sctp_handle_remote_error_event(struct sctp_remote_error *sre);
 void janus_sctp_handle_send_failed_event(struct sctp_send_failed_event *ssfe);
 void janus_sctp_handle_notification(janus_sctp_association *sctp, union sctp_notification *notif, size_t n);
 
+/* We need to keep a map of associations with random IDs, as usrsctp will
+ * use the pointer to our structures in the actual messages instead */
+static janus_mutex sctp_mutex;
+static GHashTable *sctp_ids = NULL;
+static void janus_sctp_association_unref(janus_sctp_association *sctp);
+
+/* SCTP management code */
 static gboolean sctp_running;
 int janus_sctp_init(void) {
 	/* Initialize the SCTP stack */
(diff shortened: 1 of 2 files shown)


ibc:
blablabla but looks ok

lminiero:
Well, look at it this way: you can just open a PR with no text and say "read here"... I did you a favor!

ibc:
Thanks I'll do

lminiero:
Just warn them that the description might be a bit impolite...

lminiero:
I don't see any objection, so I'll merge.

ibc:
> I don't see any objection, so I'll merge.

Actually you don't see anything.

lminiero:
Yeah, that was my point. But I can't wait for feedback forever, so merging they HAVE to test.

chadfurman:
> Yeah, that was my point. But I can't wait for feedback forever, so merging they HAVE to test.

Yeah, once we update!  :)

PR #2313, opened 2020-08-07
Various minor typo fixes

biodranik:


plugins/janus_videoroom.c
```diff
@@ -87,7 +87,7 @@ room-<unique room ID>: {
 	notify_joining = true|false (optional, whether to notify all participants when a new
 				participant joins the room. The Videoroom plugin by design only notifies
 				new feeds (publishers), and enabling this may result extra notification
-				traffic. This flag is particularly useful when enabled with \c require_pvtid
+				traffic. This flag is particularly useful when enabled with require_pvtid
```
lminiero, line 90: Why did you remove the `\c` here?

plugins/janus_videoroom.c
```diff
@@ -554,7 +554,7 @@ room-<unique room ID>: {
  * the response to a \c joinandconfigure will be a \c joined event, and
  * will again be accompanied by a JSEP SDP answer as usual.
  *
- * However you decided to publish something, as soon as the PeerConnection
+ * However if you decided to publish something, as soon as the PeerConnection
```
lminiero, line 557: The "if" is wrong here: what I meant was "no matter how you decided to publish something" (since we just talked of different ways of doing that).

plugins/janus_videoroom.c
```diff
@@ -408,7 +408,7 @@ room-<unique room ID>: {
  * (although may choose not to, more on this later) publish media in the
  * room, and as such become feeds that you can subscribe to.
  *
- * To specify a handle will be associated with a publisher, you must use
+ * To specify a handle which will be associated with a publisher, you must use
```
lminiero, line 411: The "which" is wrong here: the sentence means to say that "to specify THAT a (core) handle will be associated (is going to be associated) with a publisher, you mus use...".

plugins/janus_videoroom.c
```diff
@@ -895,7 +895,7 @@ room-<unique room ID>: {
  * right info out of band (which is impossible in rooms configured with
  * \c require_pvtid).
  *
- * To specify a handle will be associated with a subscriber, you must use
+ * To specify a handle which will be associated with a subscriber, you must use
```
lminiero, line 898: See above.

plugins/janus_videoroom.c
```diff
@@ -933,7 +933,7 @@ room-<unique room ID>: {
  * or are not interested in getting all the media (e.g., they're ok with
  * just audio and not video, or don't have enough bandwidth for both),
  * they can use those properties to shape the SDP offer to their needs.
- * In case the publisher to subscribe to is simulcasting or doing VP9 SVC,
+ * In case the publisher is simulcasting or doing VP9 SVC,
```
lminiero, line 936: Why did you remove that?

lminiero (commented):
Apologies for the late response, we were on vacation. While there are some nice fixes on typos I wasn't aware of, there are some changes that are broken. I added them as notes below.

plugins/janus_videoroom.c
biodranik: Because this text is verbatim already, and \c [prints like "\c require_pvtid"](https://janus.conf.meetecho.com/docs/videoroom.html)

biodranik:
@lminiero Thanks for the review, I have reverted back changes you've mentioned.

plugins/janus_videoroom.c
lminiero: Oh right, good catch!

PR #2314, opened 2020-08-08
Protect from an error of serializing object who contains cyclic object value

Mehdi-Verfaillie:


atoppi:
I'm sorry, I don't think this is necessary at all.
Cyclic references are not an issue here, being `message` built from data read from a input box.

Mehdi-Verfaillie:
Oh ok. Well, I got an cyclic object error on iOS on the default code. That was a solution who works for me.

PR #2324, opened 2020-08-17
[multistream] Fix renegotiation with Firefox for inactive mlines

Bug-Fairy:
Firefox does not like having mids from disabled mlines in BUNDLE group.
Chrome does not care.


amnonbb:
With this patch i got in janus log:
[ERR] [dtls.c:janus_dtls_srtp_incoming_msg:701] [1158695869776823] Handshake error: error:1000041b:SSL routines:OPENSSL_internal:TLSV1_ALERT_DECRYPT_ERROR

Bug-Fairy:
Didn't encounter a single issue from multiple clients on multiple systems. Are you sure it doesn't happen on the multistream branch without the patch?

amnonbb:
I tried it without patch and with patch on CentOS 8.2.
But i got other issues with Firefox with multistream demo.
Does multistream demo work for you in Firefox?

Bug-Fairy:
Indeed this is causing odd behavior when Janus is sending the offer. Looking into it.

Multistream works more or less for all browsers, but there are oddities here and there. I'm using my own front end instead of janus.js though.

lminiero:
Thanks for this. I didn't test extensively with Firefox, before going on vacation, so I expected some issues. Despite the problems you guys already discussed for this patch, what happens if an m-line was inactive and then becomes active again? We always keep the same mid, so I don't know if that can cause issues.

lminiero:
> [ERR] [dtls.c:janus_dtls_srtp_incoming_msg:701] [1158695869776823] Handshake error: error:1000041b:SSL routines:OPENSSL_internal:TLSV1_ALERT_DECRYPT_ERROR

I guess this can happen if the very first m-line is inactive, since it's the one all bundling is done upon.

Bug-Fairy:
> > [ERR] [dtls.c:janus_dtls_srtp_incoming_msg:701] [1158695869776823] Handshake error: error:1000041b:SSL routines:OPENSSL_internal:TLSV1_ALERT_DECRYPT_ERROR
> 
> I guess this can happen if the very first m-line is inactive, since it's the one all bundling is done upon.

There definitely was an issue with my patch when Janus was generating the offer as medium direction was not settled yet, but fixed that with 144d380. 
Not sure how to handle the first or all lines becoming inactive though.

Bug-Fairy:
Seems I was premature with this PR. It's causing errors about removing from an "already-established BUNDLE group" for another webrtc client. 
Feels like every webrtc implementation has its own idea on how inactive mlines, bundling and renegotiation should be treated.

lminiero:
Mentioning #2211 explicitly, otherwise we don't have an explicit relation between the two PRs.

lminiero:
What's the status of this patch? There seems to be something wrong with it, since I see it includes basically all the multistream commits, rather than just your changes. I guess this might be because of some rebases I performed on the multistream branch? Anyway, I'll close this since it's not in-line with the branch anymore: besides, I recently made a fix to master in https://github.com/meetecho/janus-gateway/commit/04ffda9064a486f3ad064edd4ea58ba307a118bd to address bundle issues, so hopefully they'll be fixed in multistream too when I'll rebase.

PR #2328, opened 2020-08-20
Prevent generating huge opus files when received timestamp is 0

ihusejnovic:
Hi Meetecho,

Recently we noticed that janus-pp-rec tool generates huge opus files(20GB, 30GB...) for some SIP calls. 
We use the latest Janus version, so bugfix introduced in PR https://github.com/meetecho/janus-gateway/pull/2250 didn't resolve our issue.

We converted the problematic mjr files to pcap and found some packets with very large timestamps(almost max uint32). 
But packets before and after the problematic packet were fine. Here is an example: 

image

Then we captured RTP packets on the network and found out that those problematic packets are received with a timestamp set to 0 and Janus converts them to large integer due to negative subtraction of timestamp and base_ts values.

Here is an example from the network trace:

image

I couldn't find any RFC how this packet with timestamp=0 and marker=1 should be handled so I decided to calculate timestamp based on the time when the previous packet is received (the same solution was used for an SSRC change). If you have another idea or any suggestion let me know.

We have tested this fix last few days and it's working fine for us. 

@neilkinnish can you also please take a look at this? Maybe this might resolve your issue, too.

### DIFF
--- a/rtp.c
+++ b/rtp.c
@@ -605,6 +605,17 @@ int janus_rtp_skew_compensate_video(janus_rtp_header *header, janus_rtp_switchin
 	return exit_status;
 }
 
+gint64 janus_get_audio_time_diff(uint16_t codec_type, gint64 a_last_time) {
+	gint64 time_diff = janus_get_monotonic_time() - a_last_time;
+	int akhz = 48;
+	if(codec_type == 0 || codec_type == 8 || codec_type == 9)
+		akhz = 8;	/* We're assuming 48khz here (Opus), unless it's G.711/G.722 (8khz) */
+	time_diff = (time_diff*akhz)/1000;
+	if(time_diff == 0)
+		time_diff = 1;
+	return time_diff;
+}
+
 void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_context *context, gboolean video, int step) {
 	if(header == NULL || context == NULL)
 		return;
@@ -666,13 +677,7 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
 			context->a_base_seq = seq;
 			/* How much time since the last audio RTP packet? We compute an offset accordingly */
 			if(context->a_last_time > 0) {
-				gint64 time_diff = janus_get_monotonic_time() - context->a_last_time;
-				int akhz = 48;
-				if(header->type == 0 || header->type == 8 || header->type == 9)
-					akhz = 8;	/* We're assuming 48khz here (Opus), unless it's G.711/G.722 (8khz) */
-				time_diff = (time_diff*akhz)/1000;
-				if(time_diff == 0)
-					time_diff = 1;
+				gint64 time_diff = janus_get_audio_time_diff(header->type, context->a_last_ts);
 				context->a_base_ts_prev += (guint32)time_diff;
 				context->a_prev_ts += (guint32)time_diff;
 				context->a_last_ts += (guint32)time_diff;
@@ -689,7 +694,12 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
 		}
 		/* Compute a coherent timestamp and sequence number */
 		context->a_prev_ts = context->a_last_ts;
-		context->a_last_ts = (timestamp-context->a_base_ts) + context->a_base_ts_prev;
+		if(timestamp >= context->a_base_ts) {
+			context->a_last_ts = (timestamp - context->a_base_ts) + context->a_base_ts_prev;
+		} else {
+			gint64 time_diff = janus_get_audio_time_diff(header->type, context->a_last_ts);
+			context->a_last_ts += time_diff;
+		}
 		context->a_prev_seq = context->a_last_seq;
 		context->a_last_seq = (seq-context->a_base_seq)+context->a_base_seq_prev+1;
 		/* Update the timestamp and sequence number in the RTP packet */
--- a/rtp.c
+++ b/rtp.c
@@ -605,7 +605,7 @@ int janus_rtp_skew_compensate_video(janus_rtp_header *header, janus_rtp_switchin
 	return exit_status;
 }
 
-gint64 janus_get_audio_time_diff(uint16_t codec_type, gint64 a_last_time) {
+static gint64 janus_get_audio_time_diff(uint16_t codec_type, gint64 a_last_time) {
 	gint64 time_diff = janus_get_monotonic_time() - a_last_time;
 	int akhz = 48;
 	if(codec_type == 0 || codec_type == 8 || codec_type == 9)
--- a/rtp.c
+++ b/rtp.c
@@ -677,7 +677,7 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
 			context->a_base_seq = seq;
 			/* How much time since the last audio RTP packet? We compute an offset accordingly */
 			if(context->a_last_time > 0) {
-				gint64 time_diff = janus_get_audio_time_diff(header->type, context->a_last_ts);
+				gint64 time_diff = janus_get_audio_time_diff(header->type, context->a_last_time);
 				context->a_base_ts_prev += (guint32)time_diff;
 				context->a_prev_ts += (guint32)time_diff;
 				context->a_last_ts += (guint32)time_diff;
@@ -697,7 +697,7 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
 		if(timestamp >= context->a_base_ts) {
 			context->a_last_ts = (timestamp - context->a_base_ts) + context->a_base_ts_prev;
 		} else {
-			gint64 time_diff = janus_get_audio_time_diff(header->type, context->a_last_ts);
+			gint64 time_diff = janus_get_audio_time_diff(header->type, context->a_last_time);
 			context->a_last_ts += time_diff;
 		}
 		context->a_prev_seq = context->a_last_seq;
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -6378,8 +6378,16 @@ static void *janus_sip_relay_thread(void *data) {
 						}
 						bytes = buflen;
 					}
+					if(header->markerbit) {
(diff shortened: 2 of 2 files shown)


atoppi:
@alexamirante found out a rationale behind that packet (cfr. [https://tools.ietf.org/html/rfc3389#section-5.1]())

**Real-time Transport Protocol (RTP) Payload for Comfort Noise (CN)**

> The receiver can detect silence suppression on the first
   packet received after the silence by observing that the RTP timestamp
   is not contiguous with the end of the interval covered by the
   previous packet even though the RTP sequence number has incremented
   only by one.  The RTP marker bit is also normally set on such a
   packet.

AFAIK we never encountered this before, but given its existence and the possibility of SIP phones to implement the RFC, we should take it into account maybe by filtering/adjusting those packets in the sip plugin and not in the core methods (like this PR is proposing).

rtp.c
```diff
@@ -689,7 +694,12 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
 		}
 		/* Compute a coherent timestamp and sequence number */
 		context->a_prev_ts = context->a_last_ts;
-		context->a_last_ts = (timestamp-context->a_base_ts) + context->a_base_ts_prev;
+		if(timestamp >= context->a_base_ts) {
```
atoppi, line 697: This is not correct.
Any packet received with a TS lesser than the base ts will have its TS replaced with a value computed on the real time (`janus_get_audio_time_diff`).
E.g. a SSRC change imposes a new base ts with a high value (say 4000000000). Once the ts warps around, any new packet until 4000000000 will have its TS replaced with a new computed one.

ihusejnovic:
Thank you for your comments @atoppi. 
I agree, let's move this fix to SIP plugin. I will filter these packets based on the provided RFC, but do you have any idea how should we adjust it? Should we replace timestamp like I already did in this PR or something else?

rtp.c
ihusejnovic: When SSRC is changed, context->a_base_ts is set to the current timestamp. So, it will not be replaced with a new estimated timestamp. But, this is irrelevant now because we will move this code to the SIP plugin. :)

lminiero:
Not sure what the proper fix is here, but it looks like a severely broken SIP client, since it doesn't even change payload type or SSRC: it just inserts a completely broken timestamp in there. Not saving in the mjr packets that have timestamp 0 would definitely not be a proper fix, since there are legitimate packets starting like that. My guess is that you'll have to add a check even before the `janus_rtp_header_update` call, to check timestamp `0` packets on the same SSRC as before and check if the previous timestamp does not justify a reset.

lminiero:
> @alexamirante found out a rationale behind that packet (cfr. https://tools.ietf.org/html/rfc3389#section-5.1)
> 
> **Real-time Transport Protocol (RTP) Payload for Comfort Noise (CN)**
> 
> >

Apologies, missed this post, I didn't know about this either.

rtp.c
atoppi: The RTP TS will eventually wrap around (keeping the same SSRC). So after a certain amount of time you'll receive a TS lesser than the last known base TS.

lminiero:
> we should take it into account maybe by filtering/adjusting those packets in the sip plugin and not in the core methods

Giving it some thought, I actually think it might make sense to fix those packets in `janus_rtp_header_update` instead, so in the RTP core management. We would have to do the same fix in NoSIP anyway (so replicated code), and we don't know if browsers will decide to start supporting this ugly thing in the future. It's just some additional logic that needs to be enforced when the SSRC doesn't change.

What I don't get is how to handle LEGITIMATE cases where the timestamp is very different and the sequence number increases by one: examples are media not being sent for a long time and then starting again, where the sequence number would increase by one (so that the peer knows they didn't lose any packet), but the timestamp reflects the current time (and so would not be the expected increase). This would probably force the code to be aware of the interarrival of packets somehow, which means that the timestamp fix for CN should only be applied if the value changed a lot but the last packet arrived a little time ago.

ihusejnovic:
> What I don't get is how to handle LEGITIMATE cases where the timestamp is very different and the sequence number increases by one: examples are media not being sent for a long time and then starting again, where the sequence number would increase by one (so that the peer knows they didn't lose any packet), but the timestamp reflects the current time (and so would not be the expected increase). This would probably force the code to be aware of the interarrival of packets somehow, which means that the timestamp fix for CN should only be applied if the value changed a lot but the last packet arrived a little time ago.

This is stated in RFC:

> The RTP marker bit is also normally set on such a packet. 

Maybe we can use a marker bit to differentiate this case from legitimate one?

rtp.c
ihusejnovic: Yes, you are right, thanks.

lminiero:
> Maybe we can use a marker bit to differentiate this case from legitimate one?

I don't think so. The marker bit is also set when you start sending audio after a long time that you didn't, so exactly the legitimate use case we want to allow.

atoppi:
We might compare the received RTP TS with a computed TS based on the elapsed time, and if 

those values diverge too much
AND
seq num is greater than the previous one
AND
marker bit is set

we are pretty sure that this is a CN packet.

Anyway this will add a computational effort for any received packet, not sure if this is worth given the uncommon scenario.

lminiero:
@ihusejnovic can you send us a pcap of a call with this issue, a capture of the SIP-to-Janus side? This way we can feed it to a SIPp script and test whatever fix/solution we come up with.

ihusejnovic:
> @ihusejnovic can you send us a pcap of a call (RTP only) with this issue, a capture of the SIP-to-Janus side? This way we can feed it to a SIPp script and test whatever fix/solution we come up with.

Sure, shared on email.

atoppi:
@ihusejnovic we are considering to introduce a specific fix for this issue in `janus-pp-rec` once #2345 will be merged.

atoppi:
@ihusejnovic meanwhile, would you mind sharing the mjr capture that contained the very large timestamps(almost max uint32) ?

lminiero:
As a side note, the SDP offer sent by the SIP device (or was it an answer?) and how Janus answered would help too, to see if they were offering CN in the first place and how browser/Janus reacted. If this happens because the SIP endpoint offers CN and we reject it, apparently as @atoppi found out there's a `silenceSupp` property we can set to off in the SDP: knowing what to react to would allow us to change the code accordingly.

ihusejnovic:
> @ihusejnovic meanwhile, would you mind sharing the mjr capture that contained the very large timestamps(almost max uint32) ?

Sent on email.

> As a side note, the SDP offer sent by the SIP device (or was it an answer?) and how Janus answered would help too, to see if they were offering CN in the first place and how browser/Janus reacted. If this happens because the SIP endpoint offers CN and we reject it, apparently as @atoppi found out there's a `silenceSupp` property we can set to off in the SDP: knowing what to react to would allow us to change the code accordingly.

Here is SDP offer and answer from Janus: https://pastebin.com/4pyEU5f6

Thank you.

atoppi:
hi @ihusejnovic,
getting back to this, have you tried disabling silence suppression in freeswitch [through this setting](https://freeswitch.org/confluence/display/FREESWITCH/suppress_cng)?

ihusejnovic:
> hi @ihusejnovic,
> getting back to this, have you tried disabling silence suppression in freeswitch [through this setting](https://freeswitch.org/confluence/display/FREESWITCH/suppress_cng)?

Hi, 
I don't have access to this Freeswitch so I couldn't change any settings. I have tried to remove line
`a=rtpmap:13 CN/8000` from SDP answer on Janus but it didn't help.

atoppi:
> Hi,
I don't have access to this Freeswitch so I couldn't change any settings. I have tried to remove line
a=rtpmap:13 CN/8000 from SDP answer on Janus but it didn't help.

Removing that attribute will not help.
Your SIP endpoint is generating RTP comfort noise with no specific payload type. It is using the same PT of media, keeping seq numbers in order, but setting that 0 timestamp.

I am pasting here a untested quick hack that will add `silenceSupp off` to the SDP that Janus sends to the sip endpoint.
Please give it a try and let us know if it helps.

```diff
diff --git a/plugins/janus_sip.c b/plugins/janus_sip.c
index 1d363e9e..6a249a1d 100644
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -5896,6 +5896,11 @@ char *janus_sip_sdp_manipulate(janus_sip_session *session, janus_sdp *sdp, gbool
 				ma = ma->next;
 			}
 		}
+		/* Reject RTP silence suppression */
+		if(m->type == JANUS_SDP_AUDIO) {
+			janus_sdp_attribute *a = janus_sdp_attribute_create("silenceSupp","off","----");
+			janus_sdp_attribute_add_to_mline(m, a);
+		}
 		temp = temp->next;
 	}
 	/* If we need to remove some payload types from the SDP, do it now */
```
Save it to `patch.diff` and apply it with `git apply patch.diff` on top of current master.

ihusejnovic:
Hi @atoppi, 
I have tested it for a few days but it didn't resolve the issue. 
Here are SDPs from FS and Janus: https://pastebin.com/VAtJqiPA
I can share the MJR file by email but here is just a sample from PCAP(converted from MJR):

image

atoppi:
@ihusejnovic I guess FreeSwitch simply ignores that attribute then.
Well, it seems there is no explicit way of disabling this behavior in the SDP, so this leaves the reaction in either the core or pprec as the only chance.

I'd be more hesitant to do this fix in the core, as I think common rtp sources (e.g. browser) do not support this kind of CN. 
Maybe we could add a specific flag to pp-rec to enable CN detection, leaving to the user the responsibility to check if the final file is OK.

Anyway I'm putting on hold the decision for this fix until #2345 will be merged.

atoppi:
@ihusejnovic meanwhile would you mind sending another mjr recording of a call experiencing the issue?

ihusejnovic:
> @ihusejnovic meanwhile would you mind sending another mjr recording of a call experiencing the issue?
Sure, sent on email.

atoppi:
Closed the PR for now, let's continue the discussion on the issue #2414.
We'll come up with a proper solution in a new PR as soon as possible.

PR #2332, opened 2020-08-22
Kick

backface:
kicking a participant neither worked with setting "textroom":"kick" (missing request) nor "request":"kick" (unknown request "kick").
Did I get the docs wrong or is there something mixed up?
however this PR seems to work!


lminiero:
@backface good catch! We probably added the "kick" request code, but forgot to list it in the synchronous methods the plugin supports. Notice you should add the same fix to `janus_textroom_handle_admin_message` as well, or otherwise "kick" won't be available via Admin API. I see you signed the CLA already, so as soon as that's added I'll merge :+1:

lminiero:
@backface ping :slightly_smiling_face:

backface:
added - but not tested ;)

lminiero:
Thanks, merging then!

PR #2341, opened 2020-08-27
Easy support for one-to-many scenarios in videoroomtest

neilyoung:
A small patch to `videoroomtest` for quick and dirty test of one-to-many scenarios. If you specify the query parameter `subscriber-mode=yes` the local instance will neither touch nor send local media, but instead display the media other publishers provide. Doesn't harm and change anything for the default configuration.

### DIFF
--- a/html/videoroomtest.js
+++ b/html/videoroomtest.js
@@ -64,7 +64,7 @@ var bitrateTimer = [];
 
 var doSimulcast = (getQueryStringValue("simulcast") === "yes" || getQueryStringValue("simulcast") === "true");
 var doSimulcast2 = (getQueryStringValue("simulcast2") === "yes" || getQueryStringValue("simulcast2") === "true");
-var subscriber_mode = (getQueryStringValue("subscriber-mode") === "yes")
+var subscriber_mode = (getQueryStringValue("subscriber-mode") === "yes")  || getQueryStringValue("subscriber-mode") === "true");
 
 $(document).ready(function() {
 	// Initialize the library (all console debuggers enabled)
@@ -163,12 +163,13 @@ $(document).ready(function() {
 											myid = msg["id"];
 											mypvtid = msg["private_id"];
 											Janus.log("Successfully joined room " + msg["room"] + " with ID " + myid);
-											if (subscriber_mode) {
+											if(subscriber_mode) {
                                                 						$('#videojoin').hide();
                                                 						$('#videos').removeClass('hide').show();
                                             						}
-                                            						else
+                                            						else {
                                                 						publishOwnFeed(true);
+											}
 
 											// Any new feed to attach to?
 											if(msg["publishers"]) {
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @neilyoung! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

html/videoroomtest.js
```diff
@@ -64,6 +64,7 @@ var bitrateTimer = [];
 
 var doSimulcast = (getQueryStringValue("simulcast") === "yes" || getQueryStringValue("simulcast") === "true");
 var doSimulcast2 = (getQueryStringValue("simulcast2") === "yes" || getQueryStringValue("simulcast2") === "true");
+var subscriber_mode = (getQueryStringValue("subscriber-mode") === "yes")
```
lminiero, line 67: This is missing the `true` check, if you want to be consistent with the existing properties.

html/videoroomtest.js
```diff
@@ -162,7 +163,13 @@ $(document).ready(function() {
 											myid = msg["id"];
 											mypvtid = msg["private_id"];
 											Janus.log("Successfully joined room " + msg["room"] + " with ID " + myid);
-											publishOwnFeed(true);
+											if (subscriber_mode) {
```
lminiero, line 166: Coding style: no space between `if` and bracket please.

html/videoroomtest.js
```diff
@@ -162,7 +163,13 @@ $(document).ready(function() {
...
+                                                						$('#videojoin').hide();
+                                                						$('#videos').removeClass('hide').show();
+                                            						}
+                                            						else
```
lminiero, line 170: Coding style: `else` should be on the same line as the closing bracket, and since it's after a bracket the other branch should be in brackets too (even if it has a single line of "action").

lminiero (commented):
Thanks! Added some notes before this can be merged, though. Not sure I like the name of the query string parameter, though, which seems overly long, especially with the hyphen.

html/videoroomtest.js
```diff
@@ -162,7 +163,14 @@ $(document).ready(function() {
 											myid = msg["id"];
 											mypvtid = msg["private_id"];
 											Janus.log("Successfully joined room " + msg["room"] + " with ID " + myid);
-											publishOwnFeed(true);
+											if(subscriber_mode) {
+                                                						$('#videojoin').hide();
+                                                						$('#videos').removeClass('hide').show();
+                                            						}
+                                            						else {
```
lminiero, line 170: Sorry, still missing the `else` on the same line as the closing bracket :hand_over_mouth:

lminiero:
Thanks, merging :+1:

neilyoung:
Thanks for merging it
Regards

lminiero:
Just pushed a fix as the new check for both `yes` and `true` had a typo.

neilyoung:
What? Never ending story obviously :)

PR #2345, opened 2020-09-01
[janus-pp-rec] Enhance timestamp assignment and add payload-type option to CLI.

atoppi:
In `janus-pp-rec` the timestamp assigned to each packet is a `uint64` that is used to monotonically order the packets in a recording.
This PR introduces a new algorithm in `janus-pp-rec` to assign packet timestamps in order to avoid erratic re-orderings that might lead to wrongly generated media files.

Since today that part of `janus-pp-rec` has relied on the  _"post reset trigger"_ mechanism to avoid messing media files (e.g. huge files) when particular orderings of timestamps were present.
This approach worked quite well so far but still it does not solve all corner cases. Furthermore it is in my opinion difficult to read, confusing to use and over-engineered.

This patch makes that mechanism more straightforward:
- remove the parameter `postreset-trigger` (you can not use it anymore in the CLI)
- instantly detects a timestamp reset (aka wrap-around) without the need of a trigger threshold
- use wrap-around aware checks to understand when a new timestamp is before or after the highest one

It has been tested with some mjrs containing special cases (timestamp reset in the first packets, out of order packets near the wrap around etc.) and it seemed to work as expected.

**UPDATE**: this PR also includes the changes proposed in #2412.


lminiero:
Pinging @tgabi333 as I know he uses mjr files a lot.

tgabi333:
Seems ok for me, thanks for letting me know

tgabi333:
What about JANUS_PPREC_IGNOREFIRST?

atoppi:
That part is unchanged.
I suspect that some of the issues that needed that setting could be solved by the new approach without using that parameter.

lminiero:
I don't think the "ignore first" stuff is needed anymore with this fix.

tgabi333:
@atoppi do you think that "ignore first" could be dropped out as well with this pull request?

atoppi:
@tgabi333 not in this PR. 
Skipping the first packets of a mjr might still find a motivation in the future.

lminiero:
Merging.

PR #2354, opened 2020-09-07
New mechanism to tweak/query transport plugins via Admin API

lminiero:
This PR adds a new synchronous Admin API request, called `query_transport`, that allows you to communicate with a transport plugin, like for instance `query_eventhandler` and `query_logger` already did for event handlers and loggers. The idea was to allow both dynamic configurations of transports (e.g., change settings) and queries on their internal state, where applicable. As with the other plugin categories that already supported something like this, it's an optional callback, meaning that your custom transport doesn't need to implement it: you'll need to recompile it, though, since the transport plugin API version has been bumped. 

At the moment only very simple requests have been implemented: specifically, all transport plugins allow you to change the value of `events` and `json` you specified in the configuration files on the fly; the WebSocket transport also allows you to change the `libwebsockets` logging level; finally, some plugins also allow you to query the number of active connections (e.g., HTTP, WS, Unix Sockets). That's because this PR is, again, just opening the door to this new mechanism, and I didn't want to overdo it: I'm sure in the future we can add more meaningful tweaks, but for the time being everything else seemed to be more complex that this effort called for, so I went for the low hanging fruit instead.

For a couple of examples, this is the request you can send to the HTTP transport to get the number of active connections and the number of messages currently being served, and the response sent back by Janus:

```
curl -d '{"janus": "query_transport", "transaction": "123", "admin_secret": "janusoverlord", "transport": "janus.transport.http", "request": {"request": "connections"}}' http://localhost:7088/admin

{
   "janus": "success",
   "transaction": "123",
   "response": {
      "result": 200,
      "connections": {
         "http": 6,
         "admin_http": 1
      },
      "messages": 2
   }
}
```

while this is how you can change some of the settings (notice how the transport adds a note to say it handled the request, but won't be used since event handlers are globally disabled):

```
curl -d '{"janus": "query_transport", "transaction": "123", "admin_secret": "janusoverlord", "transport": "janus.transport.http", "request": {"request": "configure", "events": true}}' http://localhost:7088/admin

{
   "janus": "success",
   "transaction": "123",
   "response": {
      "result": 200,
      "notes": [
         "Event handlers disabled at the core level"
      ]
   }
}
```

### DIFF
--- a/html/admin.html
+++ b/html/admin.html
@@ -42,6 +42,7 @@
 					<li role="presentation" class="disabled"><a href="#serverinfo" aria-controls="serverinfo" role="tab" data-toggle="tab">Server Info</a></li>
 					<li role="presentation" class="disabled"><a href="#settings" aria-controls="settings" role="tab" data-toggle="tab">Settings</a></li>
 					<li role="presentation" class="disabled"><a href="#plugins" aria-controls="plugins" role="tab" data-toggle="tab">Plugins</a></li>
+					<li role="presentation" class="disabled"><a href="#transports" aria-controls="transports" role="tab" data-toggle="tab">Transports</a></li>
 					<li role="presentation" class="disabled"><a href="#handlesinfo" aria-controls="handlesinfo" role="tab" data-toggle="tab">Handles</a></li>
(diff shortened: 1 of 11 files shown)


lminiero:
Merging.

PR #2355, opened 2020-09-08
Send websocket message in multiple fragments when needed.

### LINKED ISSUE #2349: Crash when using Firefox with WebSockets
Hello Janus Team,

I'm getting random server crashes when using Firefox 80.0.1 (64-bit) + Secure WebSockets as transport.

Using janus v0.10.5 (b26cbb159a6afb59f1ce99f3352eb2ab98199361)

Pastebin link with janus log (level 7) + libasan output
https://pastebin.com/0NdfQ1J8

Pastebin link with core stack trace:
https://pastebin.com/FWw8gJLE

It appears to be something related with buffer overflow when realocating ws_client buffer.

The crash always happens consistently in the same place on file janus_websockets.c at line 904:

`g_free(ws_client->buffer);`

Please, let me know if there is anything else I can do to help debugging this issue.

Thanks a lot for your attention,
Daniel

atoppi:
This PR adds the support to WebSocket fragmentation ([RFC 6455](https://tools.ietf.org/html/rfc6455#section-5.4)) when sending large messages to Janus clients and fixes #2349.

The effort started after the issue #2349 where it has been proved that in particular scenarios that involved the encapsulation of websocket in h2 Janus would crash due to a buffer overflow in the underlying libraries.
We decided to follow @lws-team suggestion (thanks again!) to implement message fragmentation in order to avoid this crash.
The size of each fragment (aka message chunk ) has been set as suggested to about 2 MTU (so 2800 bytes) and it is defined in the `MESSAGE_CHUNK_SIZE` macro.

While working on this PR we have also refactored the event loop in order to avoid duplicate source and make the code cleaner and more readable.

Clients **should not** take any action as their websocket libraries are expected to take care of message fragmentation.

**Notable changes (for reviewers)**:
- we do not use anymore the `sent` integer returned by `lws_write` to move the buffer pointer `bufoffset`. We just check that `sent` bytes are not lesser than requested `amount` (that would mean an error occurred) and then increment the pointer by `amount`
- `bufoffset` is now initialized to `LWS_PRE` (IMHO it does more sense, given that the buffer has size equals to `LWS_PRE` + `message size`
- the `response` string (that is the message to be sent) is now being freed as soon as it has been copied into the buffer and not just before quitting a loop iteration as before
- we fixed some potential overflows by transforming _bufsize_-like variables from `int` type to `size_t`

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -165,9 +165,9 @@ typedef struct janus_websockets_client {
 	GAsyncQueue *messages;					/* Queue of outgoing messages to push */
 	char *incoming;							/* Buffer containing the incoming message to process (in case there are fragments) */
 	unsigned char *buffer;					/* Buffer containing the message to send */
-	int buflen;								/* Length of the buffer (may be resized after re-allocations) */
-	int bufpending;							/* Data an interrupted previous write couldn't send */
-	int bufoffset;							/* Offset from where the interrupted previous write should resume */
+	size_t buflen;								/* Length of the buffer (may be resized after re-allocations) */
+	size_t bufpending;							/* Data an interrupted previous write couldn't send */
+	size_t bufoffset;							/* Offset from where the interrupted previous write should resume */
 	volatile gint destroyed;				/* Whether this libwebsockets client instance has been closed */
 	janus_transport_session *ts;			/* Janus core-transport session */
 } janus_websockets_client;
@@ -1057,6 +1057,9 @@ static int janus_websockets_callback_https(
 	return janus_websockets_callback_http(wsi, reason, user, in, len);
 }
 
+/* Use ~ 2xMTU as chunk size */
+#define MESSAGE_CHUNK_SIZE 2800
+
 /* This callback handles Janus API requests */
 static int janus_websockets_common_callback(
 		struct lws *wsi,
@@ -1184,7 +1187,7 @@ static int janus_websockets_common_callback(
 				/* Check if Websockets send pipe is choked */
 				if(lws_send_pipe_choked(wsi)) {
 					if(ws_client->buffer && ws_client->bufpending > 0 && ws_client->bufoffset > 0) {
-						JANUS_LOG(LOG_WARN, "Websockets choked with buffer: %d, trying again\n", ws_client->bufpending);
+						JANUS_LOG(LOG_WARN, "Websockets choked with buffer: %zu, trying again\n", ws_client->bufpending);
 						lws_callback_on_writable(wsi);
 					} else {
 						gint qlen = g_async_queue_length(ws_client->messages);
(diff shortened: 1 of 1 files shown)


transports/janus_websockets.c
```diff
@@ -1198,56 +1201,63 @@ static int janus_websockets_common_callback(
 				}
 
 				/* Check if we have a pending/partial write to complete first */
-				if(ws_client->buffer && ws_client->bufpending > 0 && ws_client->bufoffset > 0
-						&& !g_atomic_int_get(&ws_client->destroyed) && !g_atomic_int_get(&stopping)) {
```
lminiero, line 1202: These atomic checks should not be removed.

transports/janus_websockets.c
```diff
@@ -1198,56 +1201,63 @@ static int janus_websockets_common_callback(
...
-					}
-					/* Done for this round, check the next response/notification later */
-					lws_callback_on_writable(wsi);
-					janus_mutex_unlock(&ws_client->ts->mutex);
-					return 0;
 				}
-				/* Shoot all the pending messages */
-				char *response = g_async_queue_try_pop(ws_client->messages);
-				if(response && !g_atomic_int_get(&ws_client->destroyed) && !g_atomic_int_get(&stopping)) {
+				else {
```
lminiero, line 1208: `else` should be on the same line as the closing bracket.

transports/janus_websockets.c
```diff
@@ -1198,56 +1201,63 @@ static int janus_websockets_common_callback(
...
+					/* Shoot all the pending messages */
+					char* response = g_async_queue_try_pop(ws_client->messages);
```
lminiero, line 1210: Code style: `*` should be next to `response`, not `char`.

transports/janus_websockets.c
```diff
@@ -1198,56 +1201,63 @@ static int janus_websockets_common_callback(
...
+					ws_client->bufpending = strlen(response);
+					ws_client->bufoffset = LWS_PRE;
+					/* We can get rid of the message */
+					free(response);
+				}
+
+				/* Evaluate amount of data to send according to MESSAGE_CHUNK_SIZE */
+				int amount = ws_client->bufpending <= MESSAGE_CHUNK_SIZE ? ws_client->bufpending : MESSAGE_CHUNK_SIZE;
+				/* Set fragment flags */
+				int flags = lws_write_ws_flags(LWS_WRITE_TEXT, ws_client->bufoffset <= LWS_PRE, ws_client->bufpending <= (size_t)amount);
```
lminiero, line 1236: I don't understand this:

    ws_client->bufoffset <= LWS_PRE

Since `ws_client->bufoffset` is initialized to `LWS_PRE` and can only increase, it can never be smaller.

transports/janus_websockets.c
```diff
@@ -1198,56 +1201,63 @@ static int janus_websockets_common_callback(
...
+				/* Send the fragment with proper flags */
+				int sent = lws_write(wsi, ws_client->buffer + ws_client->bufoffset, (size_t)amount, flags);
+				JANUS_LOG(LOG_HUGE, "[%s-%p]   -- First=%d, Last=%d, Requested=%d bytes, Sent=%d bytes, Missing=%zu bytes\n", log_prefix, wsi, ws_client->bufoffset <= LWS_PRE, ws_client->bufpending <= (size_t)amount, amount, sent, ws_client->bufpending - amount);
+				if(sent < amount) {
+					/* Error on sending, abort operation */
+					JANUS_LOG(LOG_ERR, "Websocket sent only %d bytes (expected %d)\n", sent, amount);
+					ws_client->bufpending = 0;
+					ws_client->bufoffset = 0;
+				}
+				else {
```
lminiero, line 1246: See above on `else`

transports/janus_websockets.c
```diff
@@ -1198,56 +1201,63 @@ static int janus_websockets_common_callback(
...
+					/* Fragment successfully sent, update status */
+					ws_client->bufpending -= amount;
+					ws_client->bufoffset += amount;
+					if(ws_client->bufpending > 0) {
+						/* We still couldn't send everything that was left, we'll try and complete this in the next round */
 						/* We couldn't send everything in a single write, we'll complete this in the next round */
```
lminiero, line 1252: Double comment above, one should be removed.

transports/janus_websockets.c
```diff
@@ -1198,56 +1201,63 @@ static int janus_websockets_common_callback(
...
 					}
-					/* We can get rid of the message */
-					free(response);
-					/* Done for this round, check the next response/notification later */
-					lws_callback_on_writable(wsi);
-					janus_mutex_unlock(&ws_client->ts->mutex);
-					return 0;
 				}
+				/* Done for this round, check the next response/notification later */
+				lws_callback_on_writable(wsi);
```
lminiero, line 1258: This seems to be calling `lws_callback_on_writable` even when we successfully sent the last fragment, which shouldn't happen (it will perform one more useless loop).

lminiero (commented):
Added some notes :+1:

transports/janus_websockets.c
atoppi: Added again the checks in the OR form to let the function early return.

transports/janus_websockets.c
atoppi: Done.

transports/janus_websockets.c
atoppi: I have preserved the previous behavior (https://github.com/meetecho/janus-gateway/blob/master/transports/janus_websockets.c#L1246).
A quick test showed some issues when sending asyncronous messages when the `writable` call gets moved from that position.

lminiero: Ack!

transports/janus_websockets.c
```diff
@@ -1198,56 +1201,70 @@ static int janus_websockets_common_callback(
...
-					lws_callback_on_writable(wsi);
+				}
+
+				if (g_atomic_int_get(&ws_client->destroyed) || g_atomic_int_get(&stopping)) {
 					janus_mutex_unlock(&ws_client->ts->mutex);
 					return 0;
 				}
+
+				/* Evaluate amount of data to send according to MESSAGE_CHUNK_SIZE */
+				int amount = ws_client->bufpending <= MESSAGE_CHUNK_SIZE ? ws_client->bufpending : MESSAGE_CHUNK_SIZE;
```
tmatth, line 1243: nit: why not use `size_t` for amount (to avoid casting later) as it seems it's never needed as an int.

tmatth: Also a more descriptive name like: `amount_to_send` or `bytes_to_send` might be preferable.

atoppi: Honestly I've used the same name (`amount`) and type (`int`) present in the [lws example](https://github.com/warmcat/libwebsockets/blob/ec9e8ab892f2dc569ae09336c80c2796fd14a6ff/minimal-examples/ws-server/minimal-ws-server-pmd-bulk/protocol_lws_minimal_pmd_bulk.c#L91).

tmatth: @atoppi sure but in their case they don't need to do the casts. It's a minor point but cleaner IMO.

atoppi: If `amount` type is set to `size_t` the following simple test will fail
```c
#include <stdio.h>
#include <inttypes.h>
#include <stdlib.h>
#include <limits.h>

int main(int argc, char *argv[]) {	
    size_t amount = 2400;
    /* Simulate error in lws_write */
    int sent = -1;
    if (sent < amount) {
        printf("sent < amount !\n");
        return 0;
    }
    return -1;
}
```

Any cast in the comparison ` if (sent < amount)` might have a problem:
- if we do `(size_t)sent`, we are casting a signed to unsigned
- if we do `(int)amount`, we are casting to a lower (at least on my machine) size type (8 vs 4)

Am I missing something ?

tmatth: So in that example, since you are comparing to signed you should be using `ssize_t` for amount, but AFAICT in your patch you were only ever comparing to unsigned.

atoppi: [here](https://github.com/meetecho/janus-gateway/blob/ws-send-fragments/transports/janus_websockets.c#L1249
) I am comparing `sent` (type `int`, return value of `lws_write`) with `amount`.
According to lws docs about `lws_write`:
```
Return may be -1 for a fatal error needing connection close
```

tmatth: Right so `ssize_t` would make sense then.

lminiero: I personally don't see the point: `lws_write` returns `int`, not `ssize_t`, and comparing `size_t` and `ssize_t` would result in a comparison between values with different sizes anyway. Using `int` for that part seems reasonable.

tmatth: Yeah didn't want to derail the discussion here as we know these values are capped pretty low anyway.

atoppi:
@lws-team while we wait for this to be merged I'd like to discuss another issue I've found out with Firefox in this scenario.

Since I was testing huge payloads, I tried even bigger sizes (e.g. ~1.2 MBytes messages).
What happens is the following:
- a new websocket gets opened on Janus
- the message from the client does not seem to reach Janus
- when closing the tab or refreshing the page the websocket still remains open in Janus and a TCP socket is still established on the server (confirmed with lsof)
- only when the whole browser gets closed all the open websockets (e.g. due to multiple tries) are closed on Janus

The lws log of this scenario can be found [here](https://pastebin.com/EYAVCZL5).

lws-team:
> [Wed Sep  9 12:21:47 2020] [libwebsockets][INFO] seen incoming LWS_H2_FRAME_TYPE_DATA start
> [Wed Sep  9 12:21:47 2020] [libwebsockets][INFO] Frame header DATA: sid 15, flags 0x0, len 14
> [Wed Sep  9 12:21:47 2020] [libwebsockets][INFO] DATA rx on state 3
> [Wed Sep  9 12:21:47 2020] [libwebsockets][INFO] lws_h2_parser: LWS_H2_FRAME_TYPE_DATA: fl 0x0
> [Wed Sep  9 12:21:47 2020] [libwebsockets][DEBUG] lws_read_h1: h1 path: wsi state 0x119
> [Wed Sep  9 12:21:47 2020] [libwebsockets][DEBUG] lws_parse_ws: received 14 byte packet
> [Wed Sep  9 12:21:47 2020] [libwebsockets][DEBUG] lws_parse_ws: exit with 0 unused
> [Wed Sep  9 12:21:47 2020] [libwebsockets][INFO] lws_h2_parse_end_of_frame: DATA flags 0x0

I would think this is a ffox rfc8441 thing... although the support in current ffox is good for normal use now, there were a couple of bugs that hung around for about a year (in their bugzilla)... eventually they figured out it was on their side and solved it.

He's just not sending us the bulk data, it looks like we just get the encapsulated ws frame header.  And after that the link is idle.

With rfc8441 the ws traffic is going via what is often an already-existing tcp + tls connection to the server, the h2 one that provided the html and js that told it to open the ws link back to the server.  This is highly desirable, since we can mux inside that h2 connection without any further transport setup or validation.  But it means that when streams close, that's just a logical action inside a socket / tls tunnel that stays up, it's no longer the case that we get unambiguous events at the transport level like read() blows and throws -1 or 0 length tls reads or the usual that means that socket has had it.  We do get an unambiguous event but it's happening at ws frame parser inside h2 parse layer... it's pretty different.

It is meaningful to also try chrome since their implementations are completely different... chrome converged on a stable rfc8441 solution quicker than firefox in the past.

atoppi:
hi @lws-team,
I don't know how to make a test for RFC8441 with Chrome because, as discussed in the linked issue, Chrome seems not to fall in the same scenario of Firefox (WSS + lws w/ h2 support).

Do you think that a fix/workaround is possibile to avoid those stalling websockets on Janus, or we just need to wait for Mozilla to tackle this?

lws-team:
If I understand it, ffox just isn't sending any bulk data on the ws link, nor closing it either.... if so that's all happening inside ffox and I don't see how we can do anything about it.

atoppi:
Ack, thanks again @lws-team !

lminiero:
Merging :+1:

transports/janus_websockets.c
chengmo03013106: @lminiero Hi，May I asy why these atomic checks should not be removed ? 
It had checked before of this code branch, isn't ? 
What if ws_client->destroyed is not volicate int rather than int ? 
if so, Shoud it checked here ?

PR #2356, opened 2020-09-09
Bugfix: prevent borked generated audio file if meetecho header is sent with no RTP data next

SamyCookie:
For some reason I ignore, Janus server created a MJR file, with `MEET..B?.T` (4D 45 45 54 00 05 42 D5 00 54) for final data, aka a MEETECHO header with an expected of 84 bytes RTP packet data next. But this RTP packet data has not been written.

Consequences: postprocessing fails to generate a correct audio file, and some random extra silence audio data is badly inserted.

The current patch prevent this behaviour to happen by stopping MJR parsing if no data is reachable next.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -632,6 +632,7 @@ int main(int argc, char *argv[])
 	times_resetted = 0;
 	post_reset_pkts = 0;
 	uint64_t max32 = UINT32_MAX;
+	uint16_t rtp_header_len;
 	/* Start loop */
 	while(working && offset < fsize) {
 		/* Read frame header */
@@ -679,6 +680,10 @@ int main(int argc, char *argv[])
 			/* Things are simpler for data, no reordering is needed: start by the data time */
 			gint64 when = 0;
 			bytes = fread(&when, sizeof(gint64), 1, file);
+			if(bytes < (int)sizeof(gint64)) {
+				JANUS_LOG(LOG_WARN, "Missing data timestamp header");
+				break;
+			}
 			when = ntohll(when);
 			offset += sizeof(gint64);
 			len -= sizeof(gint64);
@@ -709,7 +714,12 @@ int main(int argc, char *argv[])
 			continue;
 		}
 		/* Only read RTP header */
-		bytes = fread(prebuffer, sizeof(char), len > 24 ? 24: len, file);
+		rtp_header_len = len > 24 ? 24: len;
+		bytes = fread(prebuffer, sizeof(char), rtp_header_len, file);
+		if(bytes < rtp_header_len) {
+			JANUS_LOG(LOG_WARN, "Missing RTP packet header data (%d instead %"SCNu16")\n", bytes, rtp_header_len);
+			break;
+		}
 		janus_pp_rtp_header *rtp = (janus_pp_rtp_header *)prebuffer;
 		JANUS_LOG(LOG_VERB, "  -- RTP packet (ssrc=%"SCNu32", pt=%"SCNu16", ext=%"SCNu16", seq=%"SCNu16", ts=%"SCNu32")\n",
 				ntohl(rtp->ssrc), rtp->type, rtp->extension, ntohs(rtp->seq_number), ntohl(rtp->timestamp));


postprocessing/janus-pp-rec.c
```diff
@@ -709,7 +709,12 @@ int main(int argc, char *argv[])
 			continue;
 		}
 		/* Only read RTP header */
-		bytes = fread(prebuffer, sizeof(char), len > 24 ? 24: len, file);
+		uint16_t rtp_header_len = len > 24 ? 24: len;
+		bytes = fread(prebuffer, sizeof(char), rtp_header_len, file);
+		if (bytes < rtp_header_len) {
+			JANUS_LOG(LOG_WARN, "Missing RTP packet header data (%i instead %i)\n", bytes, rtp_header_len);
```
lminiero, line 715: Since `rtp_header_len` is defined as `uint16_t`, the printf format should be `%"SCNu16"`; for bytes you might want to use `%d` since we don't use `%i` anywhere else (so just a matter of consistency). As a side note, considering it's used in all iterations of the loop, it might make sense to define the `rtp_header_len` variable just once initially, as we do for `bytes`.

lminiero (commented):
Thanks! Added a couple of notes inline. I see you signed the CLA already, so for me this can be merged as soon as the notes are addressed.

atoppi:
Hi @SamyCookie ,
thanks for the contribution!
Can you add the same checks for [data frames](https://github.com/SamyCookie/janus-gateway/blob/rtp_packet_data_missing/postprocessing/janus-pp-rec.c#L681) too?

postprocessing/janus-pp-rec.c
SamyCookie: all done !

SamyCookie:
@atoppi done too !

postprocessing/janus-pp-rec.c
```diff
@@ -679,6 +680,10 @@ int main(int argc, char *argv[])
 			/* Things are simpler for data, no reordering is needed: start by the data time */
 			gint64 when = 0;
 			bytes = fread(&when, sizeof(gint64), 1, file);
+			if (bytes < 1) {
```
atoppi, line 683: 1) _code style_: do not insert the space between `if` and the parenthesis
2) please change to
```c
if(bytes < (int)sizeof(gint64))
```

postprocessing/janus-pp-rec.c
```diff
@@ -709,7 +714,12 @@ int main(int argc, char *argv[])
 			continue;
 		}
 		/* Only read RTP header */
-		bytes = fread(prebuffer, sizeof(char), len > 24 ? 24: len, file);
+		rtp_header_len = len > 24 ? 24: len;
+		bytes = fread(prebuffer, sizeof(char), rtp_header_len, file);
+		if (bytes < rtp_header_len) {
```
atoppi, line 719: _code style_: do not insert the space between if and the parenthesis

postprocessing/janus-pp-rec.c
SamyCookie: ouch ! done !

atoppi:
lgtm :+1:

lminiero:
Thanks for the quick fixes, merging then!

SamyCookie:
@lminiero Should I write a ticket to reference the original problem  ? (the fact that the server has written a MEETECHO header with no data next)

lminiero:
I don't think there's any issue to address: it might be that a recording was closed while writing the last packet. As long as the postprocessor deals with it properly I think we're fine. As a side node, Janus doesn't write `MEETECHO` as a per-packet header anymore, but only `MEET`, as the other 4 bytes are used to store a partial timestamp of when we received the packet.

SamyCookie:
> I don't think there's any issue to address: it might be that a recording was closed while writing the last packet. As long as the postprocessor deals with it properly I think we're fine.

OK, but IMHO, this behaviour is unconsistant and should be fixed is server side.

> As a side node, Janus doesn't write `MEETECHO` as a per-packet header anymore, but only `MEET`, as the other 4 bytes are used to store a partial timestamp of when we received the packet.

You are right and it is a typo from me cause I already saw this in `janus-pp-rec`. Also, should the last MEETECHO specification [here](https://janus.conf.meetecho.com/docs/recordings.html) be updated to last version (aka MJR00002) ?

SamyCookie:
And of course, thanks for your time and merging this !

lminiero:
Good catch, I reflected that change in some slides recently (e.g., see slides 49-50 [here](https://www.slideshare.net/LorenzoMiniero/janus-workshop-cluecon-2020), but forgot to update the docs. I'll do that shortly.

PR #2357, opened 2020-09-12
Prevent unnecessary "Unsupported codec 'none' error"

neilyoung:
In case a party does not send audio, `audio_codec` renders `none`, which causes an unnecessary error in `janus_sdp_get_codec_rtpmap`. Because `do_audio` is already `false` at this time, it is checked and the call to `janus_sdp_get_codec_rtpmap` is prevented. Similar for video


lminiero:
Thanks, merging :+1:

neilyoung:
Thanks, Lorenzo. You are welcome.

PR #2358, opened 2020-09-14
Fix transaction state vacuum in MQTT transport

feymartynov:
This fixes a bug of comparing microseconds returned by `janus_monotonic_time` with seconds from `vacuums_interval` config setting which leads to premature dropping of a transaction state which in part implies sending the response to the wrong topic: the default one instead of the response topic specified in the request.

### DIFF
--- a/transports/janus_mqtt.c
+++ b/transports/janus_mqtt.c
@@ -1612,7 +1612,8 @@ static gboolean janus_mqtt_vacuum(gpointer context) {
 
 	while (g_hash_table_iter_next(&iter, NULL, &value)) {
 		janus_mqtt_transaction_state* state = value;
-		if(janus_get_monotonic_time() - state->created_at > ctx->vacuum_interval) {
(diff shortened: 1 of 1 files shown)


transports/janus_mqtt.c
```diff
@@ -1612,7 +1612,9 @@ static gboolean janus_mqtt_vacuum(gpointer context) {
 
 	while (g_hash_table_iter_next(&iter, NULL, &value)) {
 		janus_mqtt_transaction_state* state = value;
-		if(janus_get_monotonic_time() - state->created_at > ctx->vacuum_interval) {
+		gint64 diff = (janus_get_monotonic_time() - state->created_at) / G_GINT64_CONSTANT(1000000);
+
```
lminiero, line 1616: It may be better to use `G_USEC_PER_SEC` instead of `1000000` here, just for readability. As a side note, please remove the extra space above, I don't think it's needed.

lminiero (commented):
Thanks! Added just a small note inline.

lminiero:
Thanks, merging!

PR #2360, opened 2020-09-14
Use uniform code construction to get callback function address

ycherniavskyi:
This inconsistent in getting the callback function address cause me to look for some hidden logic. But when I understand that there is no hidden logic, and it is just the utilization of two different code construction which provide the same action, I decided, at last, create this tiny PR to justify my investigation somehow 😊.


atoppi:
Thanks @ycherniavskyi !
Merging.

lminiero:
I personally think that if we wanted to be considtent, we should have removed the `&` from where it was, instead, since througout the rest of the Janus code we never use it when we involve function pointers. Anyway, no harm done I guess.

ycherniavskyi:
I see. Unfortunately, I made my conclusion about consistenсу of `&` usage by observing only the current file, and there `&` used everywhere instead of updated place.

PR #2361, opened 2020-09-15
Add support for helper threads to RTSP mountpoints (fixes #2359)

### LINKED ISSUE #2359: Streaming Plugin: RTSP does not support "threads" configuration like RTP
Since RTSP ultimately uses RTP for stream reception, the RTSP sources should support configuration for "threads" to add helper threads as is supported for RTP.  When I try this for RTP I can see the created helper threads using `ps -T`, but no such additional threads exist for the RTSP streams.

When I look at the code (streaming plugin v0.0.8, Janus v0.10.5) I can see that the thread creation occurs only in the janus_streaming_create_rtp_source() function, and not in the _create_rtsp_source() method.
https://github.com/meetecho/janus-gateway/blob/v0.10.5/plugins/janus_streaming.c#L6060
Additionally, the `threads` parameter is not even parsed for the RTSP type case.

To reproduce, add a streaming source with a configuration as below.  See that the sources is created and that when reviewing threads using `ps -T` that a main processor thread `mp 1` exists but no helper threads for that source.
```
thread_test: {
        type = "rtsp"
        id = 1
        description = "Thread Test RTSP"
        audio = false
        video = true
        url = "rtsp://192.168.1.123/0"
        rtsp_user = "username"
        rtsp_pwd = "password"
        threads = 2
}
```

lminiero:
As the title says, this PR is supposed to extend the existing helper threads in the Streaming plugin to RTSP mountpoints as well, rather than just RTP. Both mountpoint types actually work the same way behind the curtains, as far as RTP is concerned, so what was missing was only a way to configure helpers when creating RTSP mountpoints in the first place. Not tested yet, so feedback welcome.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -862,6 +862,7 @@ static struct janus_json_parameter rtsp_parameters[] = {
 	{"videofmtp", JSON_STRING, 0},
 	{"videopt", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"videobufferkf", JANUS_JSON_BOOL, 0},
+	{"threads", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"rtspiface", JSON_STRING, 0},
 	{"rtsp_failcheck", JANUS_JSON_BOOL, 0}
 };
@@ -1129,8 +1130,8 @@ typedef struct janus_streaming_mountpoint {
 	janus_streaming_codecs codecs;
 	gboolean audio, video, data;
 	GList *viewers;
-	int helper_threads;		/* Only relevant for RTP mountpoints */
-	GList *threads;			/* Only relevant for RTP mountpoints */
+	int helper_threads;		/* Only relevant for RTP/RTSP mountpoints */
+	GList *threads;			/* Only relevant for RTP/RTSP mountpoints */
 	volatile gint destroyed;
 	janus_mutex mutex;
 	janus_refcount ref;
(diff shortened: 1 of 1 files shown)


jreich888:
Hi Lorenzo, haven't tried the code yet, but I noticed three places where the helper threads are used but have a check for RTP type first.  I'm not sure if they are critical or not, or if they are unrelated to the RTSP case.  They seem to be too far from the diffs to make a comment in the github interface, but please take a look around lines 3794, 4517, and 4928.   Not sure if there are others.

lminiero:
RTSP mountpoints *are* RTP sources, so it will work.

lminiero:
Made a quick test with a gstreamer-based RTSP server and it seems to be working as expected. Looking forward to your feedback before we merge.

jreich888:
Hi Lorenzo, I merged it here and gave it a try.  Ran it for a few hours, was able to confirm with `ps` that the threads were created, and everything looked good.  However, I was also running with larger socket buffers, so not conclusive what helped more, but didn't see any problems.  Did lose a couple dozen packets over the few hours, but that is orders of magnitude less than what I was seeing before.  Overall, looks good, thanks for the quick response on this request!

lminiero:
@jreich888 any update on this?

lminiero:
Whoops, sorry, I must have missed your comment: didn't get a notification about that, so sorry for the ping. Merging then :+1:

jreich888:
> Whoops, sorry, I must have missed your comment: didn't get a notification about that, so sorry for the ping. Merging then 👍

No worries, and thanks again for making this change so quickly.  👍

PR #2362, opened 2020-09-15
Notify codecs on stream published/forwarded events

zevarito:
Notify VideoRoom publisher AV codecs for **published** and **rtp_forward** events.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -3923,6 +3923,7 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 				json_object_set_new(info, "room", string_ids ? json_string(room_id_str) : json_integer(room_id));
 				json_object_set_new(info, "publisher_id", string_ids ? json_string(publisher_id_str) : json_integer(publisher_id));
 				json_object_set_new(info, "media", json_string("audio"));
+				json_object_set_new(info, "codec", json_string(janus_audiocodec_name(publisher->acodec)));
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @zevarito! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

zevarito:
Please let me know if this change requires tests & documentation.

lminiero:
Looks good, thanks! We can't merge until you sign the CLA, though. Please let us know when that's done.

zevarito:
@lminiero just signed, thanks.

lminiero:
Ack, merging :+1:

lminiero:
Group please. But I don't think we'll enforce any templating. Each plugin must be free to implement whatever they want the way they want, as long as they use JSON and use the right internal callbacks. The fact most plugins seem to have a similar API is only a coincidence, and a consequence of the fact I happened to write them all, but the simple fact that each plugin could implement a completely different logic (and satisfy different requirements) from other plugins means any templating beyond the rules we already have now would be impossible.

zevarito:
Are you saying that add `show` request to VideoRoom plugin *only* is feasible?
Would be also accepted a function say `json_object_room` that wraps this code[0] for example?
Sure, I'll post next time into the group, I wasn't sure since most of time I've seen request for help in how to use janus instead of code questions/propositions.

[0] https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_videoroom.c#L3622-L3656

PR #2363, opened 2020-09-15
Fix js client Chrome unified plan check

zevarito:
Fix js client Chrome unified plan check, unified plan will be always false.

I think you guys maintain primary `janus.js`then build/wrap for several import methods, please let me know if isn't like this. Thanks.


januscla:
Thanks for your contribution, @zevarito! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Good catch, the default is `false` so that was indeed not kicking in. Waiting for the CLA to be signed.

zevarito:
@lminiero just signed, thanks.

lminiero:
Thanks, merging then! :+1:

PR #2369, opened 2020-09-22
Fix  #2368 null protect invalid PeerConnection

### LINKED ISSUE #2368: null protect janus.js createDataChannel
After cleanup (disconnection perhaps), peer connection is set to null. 
Further calls to `createDataChannel` might crash (!incoming). 
https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L1552

Maybe better warn than crash?

edoshor:


januscla:
Thanks for your contribution, @edoshor! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

html/janus.js
```diff
@@ -1545,6 +1545,10 @@ function Janus(gatewayCallbacks) {
 			// TODO
 		};
 		if(!incoming) {
+			if(!config.pc) {
+				Janus.warn("Invalid PeerConnection");
+				return true;
+			}
```
lminiero, line 1551: I think this whole check should be moved at the beginning of the method, where `config` is assigned. It the PeerConnection doesn't exist, it doesn't make sense to proceed with the method either way. As a side note, there's no return value for the method, so just `return;` is enough.

lminiero (commented):
Thanks! Added a comment.

html/janus.js
edoshor: Sure. Done

lminiero:
Thanks for the fix, merging :+1:

PR #2371, opened 2020-09-23
[multistream] Fix stats event firing for only one stream

Bug-Fairy:
With multistream stats event is being fired for only one stream (mline) per handle and at incorrect interval. This fixes it to fire for each stream at correct intervals.


lminiero:
Ah, good catch, thanks! Merging :+1:

PR #2374, opened 2020-09-24
Wait for graceful MQTT disconnect

feymartynov:
On destroy MQTT transport publishes a offline status message and then calls `MQTTAsync_disconnect` which gracefully disconnects from the broker by sending `disconnect` message. However this call is asynchronous and is not being awaited and the socket closes faster than the message is being sent. From the point of view of the broker it looks like ungraceful disconnection so it publishes another one offline status message as LWT. As a result subscribers get duplicated offline status messages.

This PR fixes this problem by synchronizing `MQTTAsync_disconnect` call using `janus_condition_wait_until`.

### DIFF
--- a/transports/janus_mqtt.c
+++ b/transports/janus_mqtt.c
@@ -121,6 +121,8 @@ typedef struct janus_mqtt_context {
 	} connect;
 	struct {
 		int timeout;
+		janus_mutex mutex;
+		janus_condition cond;
 	} disconnect;
 	/* If we loose connection, the will is our last publish */
 	struct {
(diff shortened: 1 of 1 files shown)


transports/janus_mqtt.c
```diff
@@ -1224,7 +1228,16 @@ int janus_mqtt_client_disconnect(janus_mqtt_context *ctx) {
 
 	options.context = ctx;
 	options.timeout = ctx->disconnect.timeout;
-	return MQTTAsync_disconnect(ctx->client, &options);
+	int rc = MQTTAsync_disconnect(ctx->client, &options);
+	if (rc == MQTTASYNC_SUCCESS) {
+		janus_mutex_lock(&ctx->disconnect.mutex);
+		// Block the thread awaiting asynchronous graceful disconnect to finish.
```
lminiero, line 1234: Code style: comments should be in `/* */` format.

lminiero (commented):
I assume you tested this, and that it now works as expected, so apart from the note I added this looks good to me. As soon as the note is addressed I'll merge, thanks :+1:

transports/janus_mqtt.c
feymartynov: Fixed

feymartynov:
Sure, I've tested it.

lminiero:
Thanks, merging then :+1:

PR #2376, opened 2020-09-26
Fixed mutex lock on subscribe event ( masterid session )

mgarbin:
- There is an incorrect use of mutex lock ( duplicate lock mutex in "if(!session->helper)" ) that block all the next plugin action as result you need to kill janus server

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -3101,6 +3101,7 @@ static void *janus_sip_handler(void *data) {
 			nua_handle_t *nh = NULL;
 			if(session->stack->subscriptions != NULL)
 				nh = g_hash_table_lookup(session->stack->subscriptions, (char *)event_type);
+			janus_mutex_unlock(&session->stack->smutex);
 			if(nh == NULL) {
 				/* We don't, create one now */
 				if(!session->helper) {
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @mgarbin! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

atoppi:
lgtm

lminiero:
Good catch, but your patch is not a proper fix. It leaves the subscriptions update unprotected by the mutex. The original unlock should remain where it is, and the other occurrences should be fixed instead.

lminiero:
Besides, I think we had to make this change because we can either work on `session` or `session->master`, and there could be deadlocks there.

mgarbin:
Sorry Lorenzo, that was a "on the fly" patch that i made for test on my server, i'm not a dev and i don't know the flux of lock/unlock mutex on janus ( i'm starting to study sip plugin stack in last week ) , i see only in this part there was a double lock logic that break all the plugin. If you can do a rework to fix this problem i will be happy :) p.s. great work!

atoppi:
@mgarbin would you please test the alternative patch proposed in #2387 ?

mgarbin:
@atoppi yep no problem, i push it on my janus server this morning, on monday i will give you a feedback

mgarbin:
@atoppi ok, all is working fine! thx

lminiero:
@mgarbin thanks for the update! I'll merge that one then.

PR #2385, opened 2020-10-01
Add JSEP flag to invert processing order of rid in SDP

lminiero:
I described in detail the way simulcast works in Janus in [this blog post](https://www.meetecho.com/blog/simulcast-janus-ssrc/), last year, so I won't spend too many words on that. One thing worth mentioning, though, is that when `rid` is used, Janus expects a specific order: high substream first, then medium, and finally low. This means that the encoding object should be initialized like this:

```
parameters.encodings = [
	{ rid: "h", active: true, maxBitrate: 900000 },
	{ rid: "m", active: true, maxBitrate: 300000, scaleResolutionDownBy: 2 },
	{ rid: "l", active: true, maxBitrate: 100000, scaleResolutionDownBy: 4 }
];
```

which would result in an SDP containing something like this:

```
a=rid:h send
a=rid:m send
a=rid:l send
a=simulcast: send rid=h;m;l
```

This assumption on the order is required, because otherwise Janus has no clue on what will constitute a high quality substream, and what would be something else: the name of the `rid` values are of course useless for that, as they're user-provided and so can be anything. As such, we simply assume that the order will always be high-medium-low, and expect that in the SDP we receive. That said, I've been made aware that there are some implementations that fail to initialize when that's the enforced order, and only work when doing it the other way around, that is low-medium-high. Of course, in that case Janus messes up the mappings, because it ends up mapping the SSRC of what actually is a low quality stream to what it thinks is the high quality one instead, and viceversa, which breaks the automated switches too.

As such, I've come up with this simple PR, which simply just adds a way for you to specify whether the order will be high-medium-low (the default) or low-medium-high, so that Janus can do the mapping properly. Specifically, this can be specified in a new property you can set in the `jsep` object when sending the offer, called `rid_order`: the _only_ allowed values are `"hml"` (which is the default if omitted) and `"lmh"`. Notice that these strings have nothing to do with how you actually called your `rid` identifiers: `"hml"` is simply a compact way of saying High-Medium-Low, and `"lmh"` does the other way around. Trying to pass strings that differ from those patterns will simply result in Janus ignoring them, and so assuming the default `"hml"` value. I'm not going to add any mapping more complex than that, so don't ask.

I tested this by messing with `janus.js` to invert the order of encodings, and passing the proper `rid_order` property, and it seems to be working as expected. I only tested with the EchoTest, though. Feedback welcome.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -104,6 +104,7 @@ static struct janus_json_parameter jsep_parameters[] = {
 	{"type", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"sdp", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"trickle", JANUS_JSON_BOOL, 0},
+	{"rid_order", JSON_STRING, 0},
 	{"e2ee", JANUS_JSON_BOOL, 0}
 };
 static struct janus_json_parameter add_token_parameters[] = {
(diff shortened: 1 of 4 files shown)


janus.c
```diff
@@ -1310,6 +1311,21 @@ int janus_process_incoming_request(janus_request *request) {
 			type = NULL;
 			json_t *jsep_trickle = json_object_get(jsep, "trickle");
 			gboolean do_trickle = jsep_trickle ? json_is_true(jsep_trickle) : TRUE;
+			json_t *jsep_rids = json_object_get(jsep, "rid_order");
+			gboolean rids_hml = TRUE;
+			if(jsep_rids != NULL) {
+				const char *jsep_rids_value = json_string_value(jsep_rids);
+				if(jsep_rids_value != NULL) {
+					if(!strcasecmp(jsep_rids_value, "hml")) {
```
tmatth, line 1319: nit: since you've already initialized it to TRUE, this branch has no effect and could be dropped.

lminiero:
Merging.

PR #2386, opened 2020-10-03
fix warning about being deployed on private IP

npikimasu:
When no STUN or TURN server is configured and janus is on a NAT, a warning about being deployed on a private IP appears, even when nat_1_1_mapping is set in janus.jcfg.

Not sure when the warning starting popping up--sometime after v0.9.5.

There is a variable "const char* nat_1_1_mapping". Before support for multiple IPs was added, this variable was set to the value of the nat_1_1_mapping config value. However, while the const char* variable still exists, it is not used and not set. That causes the code that checks for private IPs to ignore the nat_1_1_mapping config setting, and issue the warning incorrectly.

This fix removes the 'const char* nat_1_1_mapping' variable, and replaces the single private-IP test with a test of each public IP address.

The warning can be safely ignored, so not a big deal in any case.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -4553,7 +4553,6 @@ gint main(int argc, char *argv[])
 #ifdef HAVE_TURNRESTAPI
 	char *turn_rest_api_method = NULL;
 #endif
-	const char *nat_1_1_mapping = NULL;
 	uint16_t rtp_min_port = 0, rtp_max_port = 0;
 	gboolean ice_lite = FALSE, ice_tcp = FALSE, full_trickle = FALSE, ipv6 = FALSE,
 		ignore_mdns = FALSE, ignore_unreachable_ice_server = FALSE;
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @npikimasu! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

janus.c
```diff
@@ -4707,32 +4706,37 @@ gint main(int argc, char *argv[])
...
-					private_address = TRUE;
-				} else if(ip[0] == 172 && (ip[1] >= 16 && ip[1] <= 31)) {
-					/* Class B private address */
-					private_address = TRUE;
-				} else if(ip[0] == 192 && ip[1] == 168) {
-					/* Class C private address */
-					private_address = TRUE;
-				}
+		int num_ips = janus_get_public_ip_count();
+		if(num_ips == 0) num_ips++;	/* if nat_1_1_mapping is off, the first (and only) public IP is the local_ip */
```
lminiero, line 4710: Code style: `if` action must be on a separate line, and comments must start with a capitalized letter.

janus.c
```diff
@@ -4707,32 +4706,37 @@ gint main(int argc, char *argv[])
...
+		/* check each public IP */
+		for (int i = 0; i < num_ips; i++) {
```
lminiero, line 4712: Code style: no whitespace between `for` and the bracket.
Besides, the `i` variable must be initialized _before_ the `for` loop.

janus.c
```diff
@@ -4707,32 +4706,37 @@ gint main(int argc, char *argv[])
...
+			gboolean private_address = FALSE;
+			const gchar* test_ip = janus_get_public_ip(i);
```
lminiero, line 4714: Code style: the asterisk pointer must be next to the variable, not the type.

lminiero (commented):
Thanks, we must have missed the orphaned variable when accepting that multiple nat-1-1 contribution.
I've added some notes inline, basically all code style related. Once they're addressed I'll test it, and in case it works merge it :+1:

janus.c
lminiero: See above on comments style too (there are other comments that need to be addressed the same way).

janus.c
npikimasu: Roger! I've corrected the styling issues.

lminiero:
Apologies for this late reply, busy week... I've just tested quickly and it seems to work as expected, thanks for the quick fixes on code style too! Merging :+1:

PR #2387, opened 2020-10-07
Do not handle session stack mutex if helper has not been created.

atoppi:
This is an alternative fix to #2376.


lminiero:
Merging.

PR #2392, opened 2020-10-14
janus.js: allow configuring simulcast send encoding parameters

fippo:
by passing callbacks.sendEncodings to allow for custom send encodings
(e.g. a different number of layers)


fippo:
@lminiero i *think* the priorities in one of the firefox code paths were wrong so drive-by fix

lminiero:
Thanks, looks good, merging! :+1:

Issue #2381, opened 2020-09-29
janus crashed in janus_ice_nacked_packet_cleanup


zhiyong0804:
janus crashed since ice handle freed, but the nack cleanup timer try to clean nack packets after 5 sec with wild pointer of ice handle

ice.c line 2718 : create nack packets clean timer, malloc janus_ice_nacked_packet packet, and reference ice handle, but under weak network, after ice handle freed 5 sec, the janus_ice_nacked_packet_cleanup function still use ice handle to free nack packe with wild pointer.

zhiyong0804:
we setup janus server in chain, and customer connect the server in Europe for test.

atoppi:
IIRC we fixed a similar (maybe the same) months ago.
Are you using a recent version of Janus?
If that's the case please provide a call stack from libasan or gdb.

lminiero:
Any update?

zhiyong0804:
> IIRC we fixed a similar (maybe the same) issue months ago.
> Are you using a recent version of Janus?
> If that's the case please provide a call stack from libasan or gdb.

@atoppi, i used v0.9.3, but i checked the the code of recent version too,  it's easy to find the PR  if we review the code of ice.c line 2718 as i described above.
here is call stack:
`[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/usr/local/bin/janus -b --debug-level=7 --log-file=/var/log/janus.log --stun-se'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000561277a28f07 in janus_ice_nacked_packet_cleanup (user_data=0x7f59cc0e1400) at ice.c:327
327                     JANUS_LOG(LOG_HUGE, "[%"SCNu64"] Cleaning up NACKed packet %"SCNu16" (SSRC %"SCNu32", vindex %d)...\n",
[Current thread is 1 (Thread 0x7f59e4ff9700 (LWP 30399))]
(gdb) where
#0  0x0000561277a28f07 in janus_ice_nacked_packet_cleanup (user_data=0x7f59cc0e1400) at ice.c:327
#1  0x00007f59f5839d03 in ?? () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007f59f5839285 in g_main_context_dispatch () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
[...]

PR #2394, opened 2020-10-16
bugfix 2381: janus crashed since ice handle freed, but the nack clean…

zhiyong0804:
janus crashed since ice handle freed, but the nack cleanup timer try to clean nack packets after 5 sec with wild pointer of ice handle.
https://github.com/meetecho/janus-gateway/issues/2381


januscla:
Thanks for your contribution, @zhiyong0804! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

atoppi:
I am still trying to fully understand the scenario leading to the issue but there is a problem in this patch.
Every packet added to the nack structure increase the handle reference count but you are not decrementing the references when a handle gets destroyed with some packets still stored in the nacks map.
As a consequence Janus will leak handles in memory.

lminiero:
Closing as I don't think this will fix the issue: as Alessandro pointed out, this will actually cause memory leaks. Thanks for your contribution anyway! If you come up with a different patch that solves the problem without causing leaks, please do share it.

PR #2396, opened 2020-10-18
janus.d.ts: correct mediaState definition

markspolakovs:
The declaration for `PluginOptions.mediaState` in janus.d.ts is incorrect - it specifies that it takes an object, rather than three parameters. This PR corrects that.


lminiero:
Thanks, merging! :+1:

PR #2398, opened 2020-10-20
[multistream] Fix videoroom support for h264 / vp9

tomleavy:
Multistream support currently breaks h264 and vp9 usage due to not passing around the required profile information. Currently this fix only works if a preferred h264 / vp9 profile is set at the videoroom level. 

I tried to also make it work generically, but could not figure out the proper way to extract the negotiated publisher profile id to set if the room did not have a preferred profile set, which would allow for a mixed set of profiles. Any help there would be appreciated

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1950,6 +1950,8 @@ typedef struct janus_videoroom_publisher_stream {
 	janus_videocodec vcodec;				/* Video codec this publisher is using (if video) */
 	int pt;									/* Payload type of this stream (if audio or video) */
 	char *fmtp;								/* fmtp that ended up being negotiated, if any (for video profiles) */
+	char *h264_profile;						/* H264 profile used for this stream (if video and H264 codec) */
+	char *vp9_profile;						/* VP9 profile this publisher is using (if video and VP9 codec) */
 	guint32 ssrc;							/* Internal SSRC of this stream */
 	gint64 fir_latest;						/* Time of latest sent PLI (to avoid flooding) */
 	gint fir_seq;							/* FIR sequence number, if needed */
@@ -2020,6 +2022,8 @@ typedef struct janus_videoroom_subscriber_stream {
 	janus_videoroom_media type;			/* Type of this stream (audio, video or data) */
 	janus_audiocodec acodec;			/* Audio codec this publisher is using (if audio) */
 	janus_videocodec vcodec;			/* Video codec this publisher is using (if video) */
+	char *h264_profile;					/* H264 profile used for this stream (if video and H264 codec) */
+	char *vp9_profile;					/* VP9 profile this publisher is using (if video and VP9 codec) */
 	int pt;								/* Payload type of this stream (if audio or video) */
 	gboolean opusfec;					/* Whether this stream is using inband Opus FEC */
 	/* RTP and simulcasting contexts */
@@ -2075,6 +2079,8 @@ static void janus_videoroom_subscriber_stream_free(const janus_refcount *s_ref) 
 	/* This subscriber stream can be destroyed, free all the resources */
 		/* TODO Anything else we should free? */
 	g_free(s->mid);
+	g_free(s->h264_profile);
+	g_free(s->vp9_profile);
 	g_free(s);
 }
 
@@ -2119,6 +2125,8 @@ static void janus_videoroom_publisher_stream_free(const janus_refcount *ps_ref) 
 	g_free(ps->mid);
 	g_free(ps->description);
 	g_free(ps->fmtp);
+	g_free(ps->h264_profile);
+	g_free(ps->vp9_profile);
 	janus_recorder_destroy(ps->rc);
 	g_hash_table_destroy(ps->rtp_forwarders);
 	ps->rtp_forwarders = NULL;
@@ -2582,6 +2590,11 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(
 	stream->type = ps->type;
 	stream->acodec = ps->acodec;
 	stream->vcodec = ps->vcodec;
+	if(stream->vcodec == JANUS_VIDEOCODEC_H264 && ps->h264_profile) {
+		stream->h264_profile = g_strdup(ps->h264_profile);
+	} else if(stream->vcodec == JANUS_VIDEOCODEC_VP9 && ps->vp9_profile) {
+		stream->vp9_profile = g_strdup(ps->vp9_profile);
+	}
 	stream->pt = ps->pt;
 	stream->opusfec = ps->opusfec;
 	char mid[5];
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @tomleavy! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

plugins/janus_videoroom.c
```diff
@@ -2020,6 +2020,8 @@ typedef struct janus_videoroom_subscriber_stream {
 	janus_videoroom_media type;			/* Type of this stream (audio, video or data) */
 	janus_audiocodec acodec;			/* Audio codec this publisher is using (if audio) */
 	janus_videocodec vcodec;			/* Video codec this publisher is using (if video) */
+	char *h264_profile;                 /* H264 profile used for this stream (if video and H264 codec) */
+	char *vp9_profile;                  /* VP9 profile this publisher is using (if video and VP9 codec) */
```
lminiero, line 2024: Please fix indentation of the comments (you're using spaces, code style requires tabs).

plugins/janus_videoroom.c
```diff
@@ -2582,6 +2586,11 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(
 	stream->type = ps->type;
 	stream->acodec = ps->acodec;
 	stream->vcodec = ps->vcodec;
+	if (stream->vcodec == JANUS_VIDEOCODEC_H264) {
+        stream->h264_profile = g_strdup(ps->fmtp);
+	} else if (stream->vcodec == JANUS_VIDEOCODEC_VP9) {
+	    stream->vp9_profile = g_strdup(ps->fmtp);
+	}
```
lminiero, line 2593: There's no check to see if `ps->fmtp` exists: if it's `NULL`, I think `g_strdup` will segfault.

Please fix the indentation (you're using spaces, code style requires tabs). Besides, per code style there should be no space between `if` and the bracket (`if (` --> `if(`).

plugins/janus_videoroom.c
```diff
@@ -9231,6 +9243,7 @@ static void *janus_videoroom_handler(void *data) {
 										h264_profile = NULL;
 										ps->vcodec = videoroom->vcodec[i];
 										ps->pt = janus_videocodec_pt(ps->vcodec);
+										ps->fmtp = g_strdup(vp9_profile);
```
lminiero, line 9246: This is wrong. The profile we extracted is not the content of the fmtp line.

plugins/janus_videoroom.c
```diff
@@ -9242,6 +9255,7 @@ static void *janus_videoroom_handler(void *data) {
 										vp9_profile = NULL;
 										ps->vcodec = videoroom->vcodec[i];
 										ps->pt = janus_videocodec_pt(ps->vcodec);
+										ps->fmtp = g_strdup(h264_profile);
```
lminiero, line 9258: This is wrong. The profile we extracted is not the content of the fmtp line.

lminiero (commented):
Thanks for the patch, even though it's currently incomplete. It's lacking the negotiation of the profiles in the answer too, for instance (see what [master](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_videoroom.c#L7192) currently does for a reference).

plugins/janus_videoroom.c
tomleavy: Before I made these changes, the `fmtp` variable was not being used for anything in this branch. A question I have is how relevant it really is. Should it just be removed and replaced with a specific variable for `h264_profile` and `vp9_profile`?

tomleavy:
I have a question around the desired direction overall. Is the goal to force every participant to use the same profile every time? Or to allow for there to be multiple profiles used simultaneously, with the problem being that some parties may not be able to handle subscribing due to client side restrictions?

plugins/janus_videoroom.c
tomleavy: 👍

plugins/janus_videoroom.c
tomleavy: 👍

tomleavy:
@lminiero I think I have everything resolved now. The missing component you mentioned previously was actually implemented already by a previous author (I think) in the code starting on line 9325. Let me know if there is anything else you need me to fix. Thanks!

plugins/janus_videoroom.c
```diff
@@ -9173,6 +9188,7 @@ static void *janus_videoroom_handler(void *data) {
 						ps->acodec = JANUS_AUDIOCODEC_NONE;
 						ps->vcodec = JANUS_VIDEOCODEC_NONE;
 						ps->pt = -1;
+						ps->fmtp = NULL;
```
lminiero, line 9191: This is not needed, `g_malloc0` allocates memory and sets it all to `0` to that variable is `NULL` already.

plugins/janus_videoroom.c
```diff
@@ -9226,22 +9242,34 @@ static void *janus_videoroom_handler(void *data) {
...
+									int pt = janus_sdp_get_codec_pt_full(offer, -1, janus_videocodec_name(videoroom->vcodec[i]), vp9_profile);
+									if(pt != -1) {
 										/* It is */
 										h264_profile = NULL;
 										ps->vcodec = videoroom->vcodec[i];
 										ps->pt = janus_videocodec_pt(ps->vcodec);
+										const char *video_fmtp = janus_sdp_get_fmtp(offer, m->index, pt);
+										if(video_fmtp) {
+											ps->fmtp = g_strdup(video_fmtp);
+										}
```
lminiero, line 9254: I don't think this is needed anymore, since we're already tracking the profile. Besides, this will cause leaks if `ps->fmtp` exists already, since you're setting the value to a new allocated property and the old one will be lost.

plugins/janus_videoroom.c
```diff
@@ -9226,22 +9242,34 @@ static void *janus_videoroom_handler(void *data) {
...
+									int pt = janus_sdp_get_codec_pt_full(offer, -1, janus_videocodec_name(videoroom->vcodec[i]), h264_profile);
+									if(pt != -1) {
 										/* It is */
 										vp9_profile = NULL;
 										ps->vcodec = videoroom->vcodec[i];
 										ps->pt = janus_videocodec_pt(ps->vcodec);
+										const char *video_fmtp = janus_sdp_get_fmtp(offer, m->index, pt);
+										if(video_fmtp) {
+											ps->fmtp = g_strdup(video_fmtp);
+										}
```
lminiero, line 9271: Same here: no need for the property, and even if it is, leaks are likely.

lminiero (commented):
Thanks for the fixes, I added a couple of additional notes.

lminiero:
@tomleavy can you let me know when you'll have some time to address those notes? I'm planning some fixes on the multistream branch and a rebase to master, but I'd rather merge this first in order to avoid conflicts later on. Thanks!

tomleavy:
Hey @lminiero I can take care of this tomorrow

tomleavy:
@lminiero should be ready to go now, rebased into 1 commit to make it clean for you

lminiero:
Thanks, merging! :+1:

PR #2399, opened 2020-10-21
Add missing unref to mDNS resolver gobject.

atoppi:
This PR simply adds a missing unref after a call to `g_resolver_get_default`.

Quoting glib docs

>Gets the default #GResolver. You should unref it when you are done
with it. #GResolver may use its reference count as a hint about how
many threads it should allocate for concurrent DNS resolutions.


lminiero:
Merging.

Issue #2401, opened 2020-10-22
NACK queue mishandling


matiaspl:
Our use case - RTP source with H.264 video and Opus audio, bitrate not exceeding 3 Mbps for video and 100 kbps for audio. Streaming plugin, vanilla Janus from snapcraft.

Code included in https://github.com/meetecho/janus-gateway/pull/1867 (which introduced dynamic queue handling in February) prevents the NACK queue from being used when it's needed the most: on lossy links with high jitter. When Janus registers a keyframe in the incoming stream it flushes the queue, so it's often unable to serve packets he's asked for by the clients through RTCP. 

There's slightly more detail in my comment PR https://github.com/meetecho/janus-gateway/pull/1867#issuecomment-713747528. The change was merged into master at the time 0.8.2 came out.

Using the same RTP source with stream plugin on two different versions of Janus.
- 0.7.6 - visible peaks, but the counter of the lost packets decreases when it gets NACKed packets from the buffer:

<img width="209" alt="packets_lost_0 7 6" src="https://user-images.githubusercontent.com/1915126/96804562-9e6de800-140f-11eb-9033-0e72efa96a3b.png">

- 0.8.2 (tested against master with the same results) - lack of incoming NACKed packets, lost packets counter increases monotonically

<img width="212" alt="packets_lost_0 8 2" src="https://user-images.githubusercontent.com/1915126/96804590-b5acd580-140f-11eb-9fa8-f3449a967695.png">

lminiero:
> When Janus registers a keyframe in the incoming stream it flushes the queue, so it's often unable to serve packets he's asked for by the clients through RTCP.

As I explained on the group, that's not a bug: it's done on purpose, and many WebRTC implementations do the same. It's only an issue if you have keyframes that are too frequent, and so attempts to recover keyframe N-1 fail because in the meanwhile you already sent keyframe N. Any attempt to send keyframes more often than once every 5s or so in WebRTC is doomed to fail, or be incredibly frail at the very least.

lminiero:
[...]

PR #2402, opened 2020-10-22
Make NACK buffer cleanup on outgoing keyframe configurable (see #2401)

lminiero:
See [this post](https://groups.google.com/d/msg/meetecho-janus/8cg4QntHeIs/jKKoe9wDBQAJ) and #2401 for reference.

This PR adds a new `janus.jcfg` property called `nack_optimizations` to enable the behaviour master currently has, that is, the ability to clean the NACK buffer any time we're sending a keyframe to the peer. Quoting from the configuration file, which explains the property and the rationale:

	# Janus can do some optimizations on the NACK queue, specifically when
	# keyframes are involved. Namely, you can configure Janus so that any
	# time a keyframe is sent to a user, the NACK buffer for that connection
	# is emptied. This allows Janus to ignore NACK requests for packets
	# sent shortly before the keyframe was sent, since it can be assumed
	# that the keyframe will restore a complete working image for the user
	# anyway (which is the main reason why video retransmissions are typically
	# required). While this optimization is known to work fine in most cases,
	# it can backfire in some edge cases, and so is disabled by default.
	#nack_optimizations = true

The property can also be dynamically changed via Admin API, e.g.:

    curl -d '{"janus": "set_nack_optimizations", "admin_secret": "janusoverlord", "transaction": "123", "nack_optimizations": true}' http://localhost:7088/admin

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -213,7 +213,7 @@ media: {
 	# sent shortly before the keyframe was sent, since it can be assumed
 	# that the keyframe will restore a complete working image for the user
 	# anyway (which is the main reason why video retransmissions are typically
-	# required. While this optimization is known to work fine in most cases,
+	# required). While this optimization is known to work fine in most cases,
 	# it can backfire in some edge cases, and so is disabled by default.
 	#nack_optimizations = true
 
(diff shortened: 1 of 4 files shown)


atoppi:
lgtm :+1:

lminiero:
Merging.

Issue #2403, opened 2020-10-22
libnice build time version check doesn't account for different PKG_CONFIG_PATH


ssalenik:
This is a minor annoyance :)
This check here uses system's pkg-config:
https://github.com/meetecho/janus-gateway/blob/a64173f451140c5e2dd258595f935db07a263902/Makefile.am#L207-L211

If janus is configured with `PKG_CONFIG_PATH` which points to another pkgconfig location for libnice, then it could be compiled with a newer version, but this check will still result in a runtime warning about outdated libnice.

eg:
```
./configure PKG_CONFIG_PATH=/my/libnice/path/lib/pkgconfig
```

Not sure what the best approach here is.

ssalenik:
I think actually passing in the user configured PKG_CONFIG_PATH here would be enough, eg:
```
PKG_CONFIG_PATH=THE_ONE_FROM_CONFIGURE pkg-config --modversion nice ...
```

ssalenik:
OK, so I guess the simplest solution is just to run `make` and `make install` with the same `PKG_CONFIG_PATH` var and that fixes it... I dunno if its "kosher" in automake to have configure pass this as a var to Makefile.am. If not, then I guess just close this.

atoppi:
Hi @ssalenik thanks for looking into this.
If you already have a proper fix, would you mind submitting a PR ? Also don't forget to sign the CLA if you didn't already!

PR #2405, opened 2020-10-23
Use PKG_CONFIG_PATH as configured for nice version

ssalenik:
Its possible that libnice location was configured via PKG_CONFIG_PATH in the configure step.
Configure passes this var to the Makefile, we should use it so that we are checking the version of the correct lib (instead of possibly the system lib).
This will also fix a build error here if there is no system version installed.

https://github.com/meetecho/janus-gateway/issues/2403


atoppi:
lgtm :+1:

lminiero:
Thanks, looks good! Merging then :+1:

PR #2407, opened 2020-10-24
Insert disabled property for inactive m-line in streams for joined event

bwqr:
When an already streaming publisher removes one of its stream from handle, other participants will receive an update event with all publishers' active and inactive m-lines. Inactive m-lines will have a property called disabled with value true.

When a new publisher handle is attached to videoroom, this new handle will receive a joined event with all publishers' active and inactive m-lines. However for this event, inactive m-lines does not have a disabled property. This commit updates joined events to contain disabled property for inactive m-lines.


lminiero:
Good catch, thanks for the fix! Looks good to me, so I'll merge :+1:

PR #2409, opened 2020-10-27
Added support for simulcast and TWCC to Duktape and Lua plugins

lminiero:
This patch adds support for both simulcast and TWCC (transport-wide CC, i.e., sender side bandwidth estimation) negotiation to the Duktape and Lua plugins. This was made possible thanks to a generous sponsorship by our good friends at [Voicenter](https://www.voicenter.com/), so make sure to properly thank them if you needed these features in your scripts! :heart: 

In case you want to start using the simulcast features in a Duktape/Lua script, while negotiation and RTP management is automatic there's a couple of things you'll need to do:

1. You can use the `setSubstream(id, substream)` and `setTemporalLayer(id, temporal)` methods from your scripts to configure which layer you're interested in: by default both plugins will assume you'll want the highest quality (`2`, even though the maximum temporal layer is really `1` right now). Notice that it's up to you to ask for a keyframe after a layer change request, which you can do using the already existing `sendPli(id)` request.
2. You can be notified when a substream/layer actually changes via the `substreamChanged(id, substream)` and `temporalLayerChanged(id, temporal)` callbacks: this is especially useful when you want to be aware of when a layer change request was actually done (which may not be right away, e.g., if a keyframe was missing), and/or to notify users about when things change (which is what we do in our demos, for instance).

You can find an example of how those new methods and callbacks are used in Duktape's `echotest.js` and Lua's `echotest.lua` scripts, which will work nicely with the existing EchoTest demo if you have it connect to those plugins rather than the C EchoTest one.

For what concerns TWCC, instead, we now automatically negotiate a few RTP extensions in both plugins, namely `mid`, `rid` and `repaired-rid`: the TWCC RTP extension is also negotiated by default, but you can disable it by passing `disableTwcc: true` when calling `generateAnswer()`. I'd always suggest negotiating it, as it will make publisher streams more stable, but if for some reason you'd rather disable it, that's the way to do it.

I only tested this with the EchoTest demo, since that's the script available in both plugins, and it seems to be working as expected. Looking forward to your feedback to see if there's anything that needs to be fixed, tweaked or improved: the sooner that happens, the sooner we can merge!

### DIFF
--- a/plugins/duktape/echotest.js
+++ b/plugins/duktape/echotest.js
@@ -193,6 +193,26 @@ function dataReady(id) {
 	// to throttle outgoing data and not send too much at a time.
 }
 
+function substreamChanged(id, substream) {
+	// If simulcast is used, this callback is invoked when the substream
+	// we're sending to this session changes: 0=low, 1=medium, 2=high
+	console.log("Substream changed for session " + id + ": " + substream);
+	// Let's send an event so that the user is aware
+	var event = { echotest: "event", videocodec: "vp8", substream: substream };
+	var jsonevent = JSON.stringify(event);
+	pushEvent(id, null, jsonevent, null);
+}
+
+function temporalLayerChanged(id, temporal) {
+	// If simulcast is used, this callback is invoked when the temporal
+	// layer we're sending to this session changes: 0=lowfps, 1=maxfps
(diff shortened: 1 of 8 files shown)


lminiero:
Merging in master, thanks again to [Voicenter](https://www.voicenter.com/) for making this possible!

PR #2410, opened 2020-10-27
New option to enforce CORS in HTTP and WS transport plugins

lminiero:
As the title says, this PR adds a new option to both the HTTP and WebSocket transport plugins to actually enforce CORS. In fact, it's up to browsers to enforce CORS once the relevant headers have been exchanged, and in case of WebSockets that actually never happens anyway. This PR allows you to enforce it on the server side too, if you want: by returning a `403` in the HTTP plugin, and just closing the connection attempt in the WS plugin.

I haven't tested this much, so I'd appreciate some feedback on whether or not the way I'm dealing with this works. For instance, GET messages are known to rarely include an `Origin` header, which is why in the HTTP plugin I'm relying on the `Referer` header instead; besides, I only played a bit with Chrome, but not other browsers, so there may be other things that need fixing before this can be merged.


lminiero:
Merging.

PR #2411, opened 2020-10-28
Fix RTP header buffer read

isnumanagic:
This PR fixes short RTP header buffer in postprocessing module.

Currently, only first 24 bytes of RTP packet are read into buffer, which can omit some/all extension header data, or, in worse case, cause extension header corruption (which is how I found this bug). This fix reads in remaining header bytes into buffer after reading csrc count and extension header length.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -730,10 +730,26 @@ int main(int argc, char *argv[])
 		audiolevel = -1;
 		rotation = -1;
 		if(rtp->extension) {
+			if (16+skip > rtp_header_len) {
+				bytes = fread(prebuffer+rtp_header_len, sizeof(char), 16+skip-rtp_header_len, file);
+				if (rtp_header_len+bytes < 16+skip) {
+					JANUS_LOG(LOG_WARN, "Missing RTP packet header data (%d instead %"SCNu16")\n", rtp_header_len+bytes, 16+skip);
+					break;
+				}
+				rtp_header_len += bytes;
+			}
 			janus_pp_rtp_header_extension *ext = (janus_pp_rtp_header_extension *)(prebuffer+12+skip);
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @isnumanagic! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

postprocessing/janus-pp-rec.c
```diff
@@ -727,13 +727,31 @@ int main(int argc, char *argv[])
 			JANUS_LOG(LOG_VERB, "  -- -- Skipping CSRC list\n");
 			skip += rtp->csrccount*4;
 		}
+		if (rtp->csrccount || rtp->extension) {
```
lminiero, line 730: Code style: no space between `if` and `(` (there's a few occurrences of this).

postprocessing/janus-pp-rec.c
```diff
@@ -727,13 +727,31 @@ int main(int argc, char *argv[])
...
+			rtp_read_n = (rtp->csrccount + rtp->extension)*4;
+			bytes = fread(prebuffer+rtp_header_len, sizeof(char), rtp_read_n, file);
+			if (bytes < rtp_read_n) {
+				JANUS_LOG(LOG_WARN, "Missing RTP packet header data (%d instead %"SCNu16")\n",
+					rtp_header_len+bytes, rtp_header_len+rtp_read_n);
+				break;
+			} else
```
lminiero, line 737: Code style: the other condition of the `else` also should have a bracket, even of the action is simple (makes the code more readable and less ambiguous).

lminiero (commented):
Thanks! We did indeed not take too many extensions into account at the time, and I agree your approach handles this better. I just added a couple of notes on code style inline, to make it more consistent with the rest of the project.

lminiero:
Looks good, merging, thanks again :+1:

PR #2412, opened 2020-10-28
Added janus-pp-rec option to only parse packets with a specific payload type

lminiero:
As the title says, nothing too fancy: this can be used in case an .mjr file contains packets with different payload types (e.g., Opus and Comfort Noise) which would confuse the processor. The processor already skips packets when dealing with static payload types (G.711/G.722), so this is really only needed when the codec is negotiated using dynamic payload types instead.

Notice this is a patch on #2345, not master, so if you test this one you'll test that one too (and we'd appreciate feedback to expedite the merging process there).


lminiero:
Merging in the mentioned branch.

PR #2420, opened 2020-11-03
janus_sampleevh: add missing break

tmatth:
No functional change (the switch is a noop) but this was the only case that was falling through.


lminiero:

PR #2422, opened 2020-11-04
Save media extmaps to recordings

isnumanagic:
This PR allows recordplay plugin and recorder to save and read media extmaps in .mjr header as array with `x` key. This in turn allows recordings with video-orient extmap to replay with correct orientation (among others), and allows rtp header extensions to be correctly mapped in janus-pp-rec tool without needing to pass `--videoorient-ext` or `--audiolevel-ext` params.

### DIFF
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -400,12 +400,12 @@ typedef struct janus_recordplay_recording {
 	janus_audiocodec acodec;	/* Codec used for audio, if available */
 	char *afmtp;				/* Audio fmtp, if any */
 	int audio_pt;				/* Payload types to use for audio when playing recordings */
-	GHashTable *audio_extmaps;	/* Audio extmaps, saved as (char*)id:(char*)info table */
+	GHashTable *audio_extmaps;	/* Audio extmaps, saved as (int*)id:(char*)info table */
 	char *vrc_file;				/* Video file name */
 	janus_videocodec vcodec;	/* Codec used for video, if available */
 	char *vfmtp;				/* Video fmtp, if any */
 	int video_pt;				/* Payload types to use for audio when playing recordings */
-	GHashTable *video_extmaps;	/* Video extmaps, saved as (char*)id:(char*)info table */
+	GHashTable *video_extmaps;	/* Video extmaps, saved as (int*)id:(char*)info table */
 	char *offer;				/* The SDP offer that will be sent to watchers */
 	gboolean e2ee;				/* Whether media in the recording is encrypted, e.g., using Insertable Streams */
 	GList *viewers;				/* List of users watching this recording */
@@ -451,7 +451,7 @@ static GHashTable *sessions;
 static janus_mutex sessions_mutex = JANUS_MUTEX_INITIALIZER;
 
 /*! \brief Helper function to filter and map \ref janus_sdp_attribute list to
- * (char*)id:(char*)info extmap hash table
+ * (int*)id:(char*)info extmap hash table
  * \param extmaps Hash table to contain filtered and mapped extmaps
  * \param attributes List of sdp media attributes
  */
@@ -462,7 +462,7 @@ static void janus_recordplay_map_extmaps(GHashTable *extmaps, const GList *attri
 	for(; attributes != NULL; attributes = attributes->next) {
 		janus_sdp_attribute *attr = (janus_sdp_attribute *)attributes->data;
 		if(attr->name == NULL || attr->value == NULL) {
-			JANUS_LOG(LOG_WARN, "Malformed NULL attribute\n");
+			JANUS_LOG(LOG_WARN, "Extmap has null fields\n");
 			continue;
 		}
 		if(!strcasecmp(attr->name, "extmap")) {
@@ -471,7 +471,13 @@ static void janus_recordplay_map_extmaps(GHashTable *extmaps, const GList *attri
 				JANUS_LOG(LOG_WARN, "Extmap 'extmap:%s' is malformed\n", attr->value);
 				continue;
 			}
-			g_hash_table_insert(extmaps, g_strndup(attr->value, id_len), g_strdup(attr->value+id_len));
+			uint32_t *ext_id = g_new(uint32_t, 1);
+			if(!janus_string_to_uint32(attr->value, ext_id)) {
+				g_hash_table_insert(extmaps, ext_id, g_strdup(attr->value+id_len));
+			} else {
+				JANUS_LOG(LOG_WARN, "Extmap has invalid id\n");
+				g_free(ext_id);
+			}
 		}
 	}
 }
@@ -487,14 +493,15 @@ static void janus_recordplay_copy_extmaps(GHashTable *dst, GHashTable *src) {
 	if(src == NULL)
 		return;
 	GHashTableIter iter;
-	char *ext_id, *ext_info;
+	uint32_t *ext_id;
+	char *ext_info;
 	g_hash_table_iter_init(&iter, src);
 	while(g_hash_table_iter_next(&iter, (gpointer *)&ext_id, (gpointer *)&ext_info)) {
 		if(ext_id == NULL || ext_info == NULL) {
 			JANUS_LOG(LOG_WARN, "Extmap has null fields\n");
 			continue;
 		}
-		g_hash_table_insert(dst, g_strdup(ext_id), g_strdup(ext_info));
+		g_hash_table_insert(dst, janus_uint32_dup(*ext_id), g_strdup(ext_info));
 	}
 }
 
@@ -505,11 +512,13 @@ static void janus_recordplay_copy_extmaps(GHashTable *dst, GHashTable *src) {
 static void janus_recordplay_regenerate_extmaps(janus_sdp_mline *media, GHashTable *extmaps) {
 	if(media == NULL || extmaps == NULL)
 		return;
-	GHashTable *id2info = g_hash_table_new(g_str_hash, g_str_equal);
+	GHashTable *id2info = g_hash_table_new(g_int_hash, g_int_equal);
 	GHashTable *info2id = g_hash_table_new(g_str_hash, g_str_equal);
-	GHashTable *generated = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
+	GHashTable *generated = g_hash_table_new_full(g_int_hash, g_int_equal, g_free, g_free);
 	GHashTableIter iter;
-	char *ext_id, *ext_info;
+	uint32_t *ext_id, i;
+	char *ext_info;
+	GQueue *indices = g_queue_new();
 	/* Add saved extmaps */
 	g_hash_table_iter_init(&iter, extmaps);
 	while(g_hash_table_iter_next(&iter, (gpointer *)&ext_id, (gpointer *)&ext_info)) {
@@ -517,8 +526,13 @@ static void janus_recordplay_regenerate_extmaps(janus_sdp_mline *media, GHashTab
 			JANUS_LOG(LOG_WARN, "Extmap has null fields\n");
 			continue;
 		}
+		JANUS_LOG(LOG_INFO, " > %d:%s\n", *ext_id, ext_info);
 		g_hash_table_insert(id2info, ext_id, ext_info);
 		g_hash_table_insert(info2id, ext_info, ext_id);
+	}
+	for(i = 1; i < 15; i++) {
+		if(!g_hash_table_contains(id2info, &i))
+			g_queue_push_head(indices, janus_uint32_dup(i));
 	}
 	/* Add generated extmaps */
 	janus_recordplay_map_extmaps(generated, media->attributes);
@@ -528,17 +542,21 @@ static void janus_recordplay_regenerate_extmaps(janus_sdp_mline *media, GHashTab
 			JANUS_LOG(LOG_WARN, "Extmap has null fields\n");
 			continue;
 		}
-		int id = atoi(ext_id);
-		if(id < 1 || id > 14) {
+		if(*ext_id < 1 || *ext_id > 14) {
 			/* 0 is padding and 15 is reserved */
-			JANUS_LOG(LOG_WARN, "Extmap has invalid id %s\n", ext_id);
+			JANUS_LOG(LOG_WARN, "Extmap has invalid id %u\n", *ext_id);
 		} else if(g_hash_table_contains(info2id, ext_info)) {
 			/* Extmap info collision; use saved id as it's renegotiated */
 			JANUS_LOG(LOG_INFO, "Extmap with value '%s' already exists, skipping\n", ext_info);
 		} else if(g_hash_table_contains(id2info, ext_id)) {
-			/* Extmap id collision; use generated extamp as those have higher priority */
-			JANUS_LOG(LOG_INFO, "Extmap with id %s already exists, overwriting\n", ext_id);
-			g_hash_table_remove(info2id, g_hash_table_lookup(id2info, ext_id));
+			/* Extmap id collision; find new id for generated extmap */
+			JANUS_LOG(LOG_INFO, "Extmap with id %u already exists, trying to find new id\n", *ext_id);
+			if(g_queue_is_empty(indices)) {
+				JANUS_LOG(LOG_INFO, "Extmap ids populated, dropping\n");
+				break;
+			}
+			ext_id = g_queue_pop_tail(indices);
+			JANUS_LOG(LOG_INFO, "New extmap id %u\n", *ext_id);
 			g_hash_table_insert(info2id, ext_info, ext_id);
 			g_hash_table_insert(id2info, ext_id, ext_info);
 		} else {
@@ -558,13 +576,14 @@ static void janus_recordplay_regenerate_extmaps(janus_sdp_mline *media, GHashTab
 	}
 	g_hash_table_iter_init(&iter, id2info);
 	while(g_hash_table_iter_next(&iter, (gpointer *)&ext_id, (gpointer *)&ext_info)) {
-		janus_sdp_attribute *attr = janus_sdp_attribute_create("extmap", "%s%s", ext_id, ext_info);
+		janus_sdp_attribute *attr = janus_sdp_attribute_create("extmap", "%u%s", *ext_id, ext_info);
 		media->attributes = g_list_prepend(media->attributes, attr);
 	}
 	/* Cleanup */
 	g_hash_table_destroy(id2info);
 	g_hash_table_destroy(info2id);
 	g_hash_table_destroy(generated);
+	g_queue_free_full(indices, g_free);
 }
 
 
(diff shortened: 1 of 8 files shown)


plugins/janus_recordplay.c
```diff
@@ -448,6 +450,38 @@ typedef struct janus_recordplay_session {
 static GHashTable *sessions;
 static janus_mutex sessions_mutex = JANUS_MUTEX_INITIALIZER;
 
+static GList *janus_recordplay_filter_extmaps(const GList *attributes) {
+	GList *filtered = NULL;
+	while(attributes != NULL) {
+		janus_sdp_attribute *attr = (janus_sdp_attribute *)attributes->data;
+		if(!strcasecmp(attr->name, "extmap"))
```
lminiero, line 457: Action is multiline, per code style there should be brackets (less ambiguous, more readable).

plugins/janus_recordplay.c
```diff
@@ -562,16 +597,19 @@ static const char *janus_recordplay_parse_codec(const char *dir, const char *fil
 			offset += 2;
 			if(len > 0 && !parsed_header) {
 				/* This is the info header */
-				bytes = fread(prebuffer, sizeof(char), len, file);
+				char *json_header = malloc(sizeof(char)*(len+1));
```
lminiero, line 600: Please use `g_malloc`/`g_free`. That said, not sure why you're allocating? 1500 bytes should be more than enough even if extensions are in the json.

plugins/janus_recordplay.c
```diff
@@ -606,6 +644,21 @@ static const char *janus_recordplay_parse_codec(const char *dir, const char *fil
...
+				/* Any extmaps? */
+				json_t *ext_json = json_object_get(info, "x");
+				if(ext_json && json_is_array(ext_json) && json_array_size(ext_json) > 0) {
+					size_t ext_n = json_array_size(ext_json), i = 0;
+					for(i = 0; i < ext_n; i++) {
+						const char *ext_str = json_string_value(json_array_get(ext_json, i));
+						char delim = ':', **ext_split = g_strsplit(ext_str, &delim, 2);
+						janus_sdp_attribute *ext = janus_sdp_attribute_create(
+							g_strdup(ext_split[0]),
+							ext_split[1] ? g_strdup(ext_split[1]) : NULL);
```
lminiero, line 656: This is wrong. It's assuming `g_strsplit` will succeed, that there are at least two parts and that both are not `NULL`. This will segfault if any of those assumptions are incorrect. Besides, there's no need to `g_strdup` when doing `janus_sdp_attribute_create`, as that method already allocates new stuff for those strings: you're generating memory leaks, as those strings you're allocating are lost.

plugins/janus_recordplay.c
```diff
@@ -653,15 +706,24 @@ static int janus_recordplay_generate_offer(janus_recordplay_recording *rec) {
 		JANUS_SDP_OA_AUDIO_PT, rec->audio_pt,
 		JANUS_SDP_OA_AUDIO_FMTP, rec->afmtp,
 		JANUS_SDP_OA_AUDIO_DIRECTION, JANUS_SDP_SENDONLY,
-		JANUS_SDP_OA_AUDIO_EXTENSION, JANUS_RTP_EXTMAP_MID, 1,
```
lminiero, line 656: Do NOT remove this: `mid` is NOT one of the extensions you're notified about in the extmap. It's always added by the core, and so it must always be there.

plugins/janus_recordplay.c
```diff
@@ -653,15 +706,24 @@ static int janus_recordplay_generate_offer(janus_recordplay_recording *rec) {
...
 		JANUS_SDP_OA_VIDEO, offer_video,
 		JANUS_SDP_OA_VIDEO_CODEC, janus_videocodec_name(rec->vcodec),
 		JANUS_SDP_OA_VIDEO_FMTP, rec->vfmtp,
 		JANUS_SDP_OA_VIDEO_PT, rec->video_pt,
 		JANUS_SDP_OA_VIDEO_DIRECTION, JANUS_SDP_SENDONLY,
-		JANUS_SDP_OA_VIDEO_EXTENSION, JANUS_RTP_EXTMAP_MID, 1,
```
lminiero, line 662: See above.

plugins/janus_recordplay.c
```diff
@@ -653,15 +706,24 @@ static int janus_recordplay_generate_offer(janus_recordplay_recording *rec) {
...
+		JANUS_SDP_OA_VIDEO_RTCPFB_DEFAULTS,
```
lminiero, line 714: Not sure why you're adding this?

plugins/janus_recordplay.c
```diff
@@ -653,15 +706,24 @@ static int janus_recordplay_generate_offer(janus_recordplay_recording *rec) {
...
 		JANUS_SDP_OA_DATA, FALSE,
 		JANUS_SDP_OA_DONE);
+	if(offer_audio) {
+		janus_sdp_mline *audio = janus_sdp_mline_find(offer, JANUS_SDP_AUDIO);
+		audio->attributes = g_list_concat(audio->attributes,
```
lminiero, line 719: This is not checking if `audio` is not `NULL`: it will segfault if it is.

plugins/janus_recordplay.c
```diff
@@ -653,15 +706,24 @@ static int janus_recordplay_generate_offer(janus_recordplay_recording *rec) {
...
+			janus_recordplay_copy_extmaps(rec->audio_extmaps));
+	}
+	if(offer_video) {
+		janus_sdp_mline *video = janus_sdp_mline_find(offer, JANUS_SDP_VIDEO);
+		video->attributes = g_list_concat(video->attributes,
```
lminiero, line 724: See above.

plugins/janus_recordplay.c
```diff
@@ -1683,6 +1749,25 @@ static void *janus_recordplay_handler(void *data) {
...
 			answer->o_version = session->sdp_version;
+			/* Update media attributes */
+			if(audio) {
+				if(rec->audio_extmaps != NULL) {
+					g_list_free_full(rec->audio_extmaps, janus_sdp_attribute_destroy);
+					g_list_free_full(session->arc->extmaps, g_free);
+				}
+				GList *attr = janus_sdp_mline_find(answer, JANUS_SDP_AUDIO)->attributes;
+				rec->audio_extmaps = janus_recordplay_filter_extmaps(attr);
+				session->arc->extmaps = janus_recordplay_bake_extmaps(rec->audio_extmaps);
```
lminiero, line 1760: Not checking if `session->arc` exists.

plugins/janus_recordplay.c
```diff
@@ -1683,6 +1749,25 @@ static void *janus_recordplay_handler(void *data) {
...
+			}
+			if(video) {
+				if(rec->video_extmaps != NULL) {
+					g_list_free_full(rec->video_extmaps, janus_sdp_attribute_destroy);
+					g_list_free_full(session->vrc->extmaps, g_free);
+				}
+				GList *attr = janus_sdp_mline_find(answer, JANUS_SDP_VIDEO)->attributes;
+				rec->video_extmaps = janus_recordplay_filter_extmaps(attr);
+				session->vrc->extmaps = janus_recordplay_bake_extmaps(rec->video_extmaps);
```
lminiero, line 1769: See above.

record.h
```diff
@@ -48,6 +49,8 @@ typedef struct janus_recorder {
 	char *codec;
 	/*! \brief Codec-specific info (e.g., H.264 or VP9 profile) */
 	char *fmtp;
+	/*! \brief Media extmaps */
+	GList *extmaps;
```
lminiero, line 53: I think using a list here is wrong. It should be a table, if the purpose is mapping a numeric extension ID to the extension namespace. Not even sure what the value saved there is?

record.c
```diff
@@ -300,6 +303,16 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
...
 			json_object_set_new(info, "e", json_true());
+		/* Write extmaps */
+		if(recorder->extmaps) {
+			json_t *ext_json = json_array();
+			GList *ext = recorder->extmaps;
+			while(ext != NULL) {
+				json_array_append_new(ext_json, json_string(ext->data));
+				ext = ext->next;
+			}
+			json_object_set(info, "x", ext_json);
```
lminiero, line 314: If the purpose is having mappings, here, this should be a key/value object, not a serialized string.

lminiero (commented):
While I applaud the effort (thanks!), I think there are several things done wrong here.

* First of all, using a list to store these mappings is IMO incorrect: since you want to know which numeric ID corresponds to which extension, it should be a table instead (which would make it easier to save and process in mjr files too).
* I don't like how the Record&Play plugin is pregenerating `janus_sdp_attribute` elements and storing them in a list, to append them brutally to the SDP. This will most likely cause issues when renegotiations are done, since it may result in those attributes being deallocated and still referenced by the list in the recording object. If all we keep is a map, we can regenerate the attributes any time we need them.
* I added other notes inline on some of the things I noticed that are wrong or dangerous.

isnumanagic:
Thanks for the quick review, I'll try to update the PR to address the comments sometime today.

plugins/janus_recordplay.c
isnumanagic: In theory it's possible (even though it's highly unlikely) for JSON header to have more than 1500 chars if there are many extmaps with long URIs.

plugins/janus_recordplay.c
isnumanagic: I've slightly misread g_strsplit documentation, thanks for pointing this out.

plugins/janus_recordplay.c
isnumanagic: This is a remnant from earlier local iteration of this PR, it somehow got through. Thanks for spotting this.

isnumanagic:
Changed extmaps to use `GHashMap *` mappings (with key `char *id` and value `char *info` types, where info is extmap value after id). JSON .mjr header also updated to keeps object instead of array for extmap `x` key.

postprocessing/janus-pp-rec.c
```diff
@@ -439,20 +442,23 @@ int main(int argc, char *argv[])
 			offset += 2;
 			if(len > 0 && !parsed_header) {
 				/* This is the info header */
-				bytes = fread(prebuffer, sizeof(char), len, file);
+				char *json_header = g_new(char, len+1);
```
atoppi, line 445: I don't get the purpose of using `json_header` (heap allocation) in place of `prebuffer` (stack).
By using `json_header` freeing must be taken into account, thus potentially leading to future bugs.
Both `char *` will be NULL terminated, so I don't really get the point of this. Am I missing something?

postprocessing/mjr2pcap.c
```diff
@@ -233,10 +233,13 @@ int main(int argc, char *argv[])
 			offset += 2;
 			if(len > 0 && !parsed_header) {
 				/* This is the info header */
-				bytes = fread(prebuffer, sizeof(char), len, file);
-				prebuffer[len] = '\0';
+				char *json_header = g_new(char, len+1);
```
atoppi, line 236: same as above

postprocessing/janus-pp-rec.c
isnumanagic: See here: https://github.com/meetecho/janus-gateway/pull/2422#discussion_r517671882 (tl;dr: prebuffer's 1500 char size might not be enough)

atoppi: I see. Thanks for clarifying then!

isnumanagic:
Hi @lminiero @atoppi. Could I request review for updated PR?

lminiero:
Sorry, we're busy with the IETF meeting, so we won't be able to do much before next week.

record.c
```diff
@@ -300,6 +303,16 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
 		/* If media will be end-to-end encrypted, mark it in the recording header */
 		if(recorder->encrypted)
 			json_object_set_new(info, "e", json_true());
+		/* Write extmaps */
+		if(recorder->extmaps) {
+			json_t *ext_json = json_object();
+			GHashTableIter iter;
+			char *ext_id, *ext_info;
+			g_hash_table_iter_init(&iter, recorder->extmaps);
+			while(g_hash_table_iter_next(&iter, (gpointer *)&ext_id, (gpointer *)&ext_info))
```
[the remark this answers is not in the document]
lminiero: As a side note, this isn't checking whether either `ext_id` or `ext_info` are `NULL`.

postprocessing/janus-pp-rec.c
```diff
@@ -633,6 +664,13 @@ int main(int argc, char *argv[])
 	post_reset_pkts = 0;
 	uint64_t max32 = UINT32_MAX;
 	uint16_t rtp_header_len, rtp_read_n;
+	/* Set extmap ids from attr list if not set */
+	for(i = 0; i < ext_n; i++) {
+		if(audio_level_extmap_id < 1 && strstr(ext[i], JANUS_RTP_EXTMAP_AUDIO_LEVEL))
+			audio_level_extmap_id = atoi(ext[i]);
+		if(video_orient_extmap_id < 1 && strstr(ext[i], JANUS_RTP_EXTMAP_VIDEO_ORIENTATION))
+			video_orient_extmap_id = atoi(ext[i]);
```
lminiero, line 672: This isn't checking if `ext[i]` is `NULL`.

postprocessing/janus-pp-rec.c
```diff
@@ -576,6 +582,25 @@ int main(int argc, char *argv[])
...
+				json_t *ext_json = json_object_get(info, "x");
+				if(ext_json) {
+					if(!json_is_object(ext_json)) {
+						JANUS_LOG(LOG_WARN, "Malformed extmaps in info header...\n");
+						g_free(json_header);
+						cmdline_parser_free(&args_info);
+						exit(1);
+					}
+					ext_n = json_object_size(ext_json);
+					ext = g_new(char *, ext_n);
```
lminiero, line 595: What if this is an empty object? `ext_n` will be `0`.

plugins/janus_recordplay.c
```diff
@@ -606,6 +688,14 @@ static const char *janus_recordplay_parse_codec(const char *dir, const char *fil
 				json_t *f = json_object_get(info, "f");
 				if(f && json_is_string(f) && fmtp && fmtplen > 0)
 					g_snprintf(fmtp, fmtplen, "%s", json_string_value(f));
+				/* Any extmaps? */
+				json_t *ext_json = json_object_get(info, "x"), *ext_info;
+				const char *ext_id;
+				if(ext_json && json_is_object(ext_json)) {
+					g_hash_table_remove_all(extmaps);
+					json_object_foreach(ext_json, ext_id, ext_info)
+						g_hash_table_insert(extmaps, g_strdup(ext_id), (gpointer)json_string_value(ext_info));
```
lminiero, line 697: See above: not checking if those strings are `NULL`.

plugins/janus_recordplay.c
```diff
@@ -448,6 +450,80 @@ typedef struct janus_recordplay_session {
...
+		return;
+	/* Consolidate extmaps */
+	GHashTable *mapped = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
+	janus_recordplay_map_extmaps(mapped, media->attributes);
+	GHashTableIter iter;
+	char *ext_id, *ext_info;
+	g_hash_table_iter_init(&iter, extmaps);
+	while(g_hash_table_iter_next(&iter, (gpointer *)&ext_id, (gpointer *)&ext_info))
+		g_hash_table_insert(mapped, g_strdup(ext_id), g_strdup(ext_info));
+	/* Remove existing extmaps */
```
lminiero, line 507: I think this is quite wrong: generate offer adds a `mid` extmap, that you'd be removing here. The only thing that you should be doing here is possibly see if an attribute you're adding now already exists (maybe with a different ID), or if there are ID conflicts, so those are the only ones you should remove IMO.

lminiero (commented):
Added some more notes. Thanks!

record.c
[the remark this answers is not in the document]
isnumanagic: Because values saved within the table are `char *`, so i pass pointer to `char *` so it can be populated while iterating.

plugins/janus_recordplay.c
isnumanagic: Generated extmaps are merged with media attributes into temporary `merged` hashmap, with former having priority, so this would not remove `mid` extmap as it would be present in `merged` hashmap and re-added further down. However, upon inspecting this part of code, it is not very clear how it does what it does, and has some possible edge case errors, so i will rewrite it.

isnumanagic:
Thanks for the review @lminiero. Sorry for pinging you while you were attending IETF meeting, hope it went well though! I have updated the PR per your comments, and rebased it to master to check for possible conflicts.

sdp-utils.c
```diff
@@ -1282,7 +1282,7 @@ janus_sdp *janus_sdp_generate_offer(const char *name, const char *address, ...)
 				char *extmap = g_hash_table_lookup(audio_extids, iter->data);
 				if(extmap != NULL) {
 					janus_sdp_attribute *a = janus_sdp_attribute_create("extmap",
-						"%d %s\r\n", GPOINTER_TO_INT(iter->data), extmap);
```
isnumanagic, line 1285: Correct me if this was supposed to be like this, but it does not appear in any other `janus_sdp_attribute_create` calls. I removed it because it messes up extmap value matching (I could've just trimmed extmap values but this seemed to be more correct).

plugins/janus_recordplay.c
```diff
@@ -448,6 +450,143 @@ typedef struct janus_recordplay_session {
...
+	char *ext_info;
+	GQueue *indices = g_queue_new();
+	/* Add saved extmaps */
+	g_hash_table_iter_init(&iter, extmaps);
+	while(g_hash_table_iter_next(&iter, (gpointer *)&ext_id, (gpointer *)&ext_info)) {
+		if(ext_id == NULL || ext_info == NULL) {
+			JANUS_LOG(LOG_WARN, "Extmap has null fields\n");
+			continue;
+		}
+		JANUS_LOG(LOG_INFO, " > %d:%s\n", *ext_id, ext_info);
```
atoppi, line 529: I'd make this log more specific than just `id > info`.
Also consider lowering its level to verbose.

isnumanagic: This is one of debug logs I used for testing, it shouldn't be here. Thanks for spotting this.

lminiero:
To be honest after discussing with @atoppi this still feels very convoluted for us. It looks like everything is handled in a more complex way than it should. From what we gather, this basically needs these core functionality:

1. a hashmap in record.h to track extensions in recordings
2. a way to add/remove extensions to a new mjr (the same way we can mark a recording as `e2ee`, for instance)
3. a way for interested plugins (e.g., Record&Play) to use that info

Point 1. is indeed there, but 2. seems implicit (plugins work on the hashtable directly) which can be problematic. Point 3. is the part where I think it has been implemented in a way that is way more complex than necessary, with duplicated tables, stuff that is regenerated, etc., all things that will add a maintainance burden and will prevent an easy integration in other plugins in the future (just imagine doing all this in the VideoRoom!). It feels like all the plugin needs is a simple way to add extensions one by one when a recording is being created, and iterating on those from existing recordings to create the attributes the first time, without complex structures to back the functionality.

I understand we've asked for way too many changes already, so I don't expect you to completely refactor the effort now. I'll try to come up with a simpler way to address this requirement and publish a new PR (possibly by the end of the week), so that we can check if it solves your need in a cleaner way. We can keep this PR open for the time being until that happens.

isnumanagic:
Thanks for the update. I agree with what has been said, especially with the third point - the current way of adding extmaps to SDP offer is not very elegant. I look forward to your PR.

isnumanagic:
Hi @lminiero & @atoppi, just wanted to check on the status here. Is this in progress? If not, I can take over the refactor, but will need some pointers on how to do it best, i.e. what you guys had in mind.

lminiero:
@isnumanagic apologies for the lack of feedback. I've been swamped with work so I had no time to work on this. I'll try to come up with something in the next few days or next week.

lminiero:
@isnumanagic please see the PR mentioned above. I'll close this one in the meanwhile. Thanks again!

PR #2423, opened 2020-11-06
Differentiate between IPv4 and IPv6 NAT-1-1 addresses

lminiero:
As the title says, this new PR tries to add some awareness in Janus about the nature of the provided address(es) for NAT-1-1 mapping purposes. At the moment, when you specify one or more addresses, they're blindly set for all candidates, independently of the protocol family: this means that if you set, for instance, `1.2.3.4`, this would be set for IPv6 candidates as well, which is clearly wrong. This new patch corrects that, and ensures that a NAT-1-1 address is only enforced if it's the same family as the candidate it's being applied on.

There's no change in how you configure the feature, it works exactly the same way: it's Janus that will look at the address(es) you pass, and figures out if they're IPv4 or IPv6 to handle them accordingly.

I tested this in several scenarios and it seems to be working, but of course I encourage you to test this carefully, especially if you're using the NAT-1-1 feature a lot (e.g., deploying on AWS, Azure, or other cloud providers). If I don't get any feedback, I plan to merge this relatively soon.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -2063,25 +2063,32 @@ static void janus_ice_cb_new_local_candidate (NiceAgent *agent, NiceCandidate *c
 #endif
 	char buffer[200];
 	guint public_ip_index = 0;
+	gboolean ipv6 = (nice_address_ip_version(&candidate->addr) == 6);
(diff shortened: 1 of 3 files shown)


alexamirante:
I tested it and works as expected. :+1:

lminiero:
Merging.

PR #2427, opened 2020-11-08
fixing a couple of bugs in sctp.c in burst transfers

afshin2003:
In janus_sctp_pending_message_create the len attribute was not set properly.
In the burst transfers, packets got stuck in pending_messages queue!. To fix this the new function janus_sctp_data_ready is called in case of SCTP_SENDER_DRY_EVENT. I moved the first loop of janus_sctp_send_data function to the new created function to push pending messages to sctp stack. At the end of function I called the janus_dtls_sctp_data_ready function.

### DIFF
--- a/sctp.c
+++ b/sctp.c
@@ -75,6 +75,7 @@ static janus_sctp_pending_message *janus_sctp_pending_message_create(uint16_t id
 	if(buf != NULL && len > 0) {
 		m->buf = g_malloc(len);
 		memcpy(m->buf, buf, len);
+		m->len = len;
 	} else {
 		m->buf = NULL;
 		m->len = 0;
@@ -104,6 +105,7 @@ int janus_sctp_send_text_or_binary(janus_sctp_association *sctp, uint16_t id, gb
 void janus_sctp_reset_outgoing_stream(janus_sctp_association *sctp, uint16_t stream);
 void janus_sctp_send_outgoing_stream_reset(janus_sctp_association *sctp);
 int janus_sctp_close_channel(janus_sctp_association *sctp, uint16_t id);
+void janus_sctp_data_ready(janus_sctp_association *sctp);
 void janus_sctp_handle_open_request_message(janus_sctp_association *sctp, janus_datachannel_open_request *req, size_t length, uint16_t stream);
 void janus_sctp_handle_open_response_message(janus_sctp_association *sctp, janus_datachannel_open_response *rsp, size_t length, uint16_t stream);
 void janus_sctp_handle_open_ack_message(janus_sctp_association *sctp, janus_datachannel_ack *ack, size_t length, uint16_t stream);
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @afshin2003! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

sctp.c
```diff
@@ -422,21 +425,7 @@ static int janus_sctp_incoming_data(struct socket *sock, union sctp_sockstore ad
...
-				JANUS_LOG(LOG_WARN, "[%"SCNu64"] Got EAGAIN when trying to resend pending message on channel %"SCNu16"\n",
-					sctp->handle_id, m->id);
-				break;
-			}
-			(void)g_queue_pop_head(sctp->pending_messages);
-			janus_sctp_pending_message_free(m);
-			m = g_queue_peek_head(sctp->pending_messages);
-		}
-	}
+
```
lminiero, line 428: I'm not sure removing this is a good idea. The main reason why we tried to send all pending messages before sending the data we were just asked to send was to ensure the proper order of the messages. Since you removed this, and now pending messages are sent by the "dry" event, it means it's going to be very likely that a message sent now with `janus_sctp_send_data` will be sent _before_ a pending message waiting to be sent.

lminiero (commented):
Thanks! I added a comment inline.

sctp.c
atoppi: No, this can not happen.
`janus_sctp_send_data` will either enqueue the message (in case we are buffering in `pending_message`) or send it right away.
Once `DRY_EVENT` occurs, the `pending_message` will be traversed in FIFO order.

atoppi:
Thanks for the effort @afshin2003, I think this is a good enhancement over the current implementation.

What you are proposing is a way to avoid message dropping (aka flow control) in case of bursts and I see how you accomplish this by starting the en-queuing in `janus_sctp_send_data` in case of `EAGAIN` and continuing to en-queue if a queue is not empty.
Then the queue is emptied in FIFO order when `DRY_EVENT` is received from the stack.

However I have an observation.
Even being an improvement, there is still room for dropped messages. In case we buffer a huge amount of messages, once `DRY_EVENT` occurs, janus will try to send all the messages in the queue and any error like `EAGAIN` will lead to a drop. This is not a bug in the PR and this behavior comes from the existing implementation. Anyway I think it is worth noting since the PR claims to fix "dropped messages in case of bursts".

uxmaster:
It seems the messages are left in the queue in the case of `EAGAIN`, so they're not dropped, or am I missing something?

atoppi:
@uxmaster you're right, we are just peeking the head with `g_queue_peek_head`.

lminiero:
Apologies for the very late reply: we've been busy with the IETF meeting and I've fallen a bit behind with reviews. I think the doubts I had are cleared, after the discussion I read above, so this is good to go for me. Thanks again, @afshin2003, merging! :v:

afshin2003:
So sorry guys for very late response :(
I was so busy on my project and could not check here!
I'm really appreciated for the good job you have been doing on this project.

PR #2428, opened 2020-11-08
Add durable flag for rabbitmq event handler

sakshamgupta05:
Added configuration for RabbitMQ Event Handler (rabbitmqevh.jcfg) to enable or disable durable flag while declaring exchange & queue (default is false). 

RabbitMQ Durability:
https://www.rabbitmq.com/queues.html#durability
https://www.rabbitmq.com/reliability.html#broker-side

Configuration `rabbitmqevh.jcfg`:
```
durable = true  # Default is false
```

I have tested this with both values (durable true & false) and it is working as expected.

### DIFF
--- a/events/janus_rabbitmqevh.c
+++ b/events/janus_rabbitmqevh.c
@@ -101,6 +101,7 @@ static amqp_connection_state_t rmq_conn;
 static amqp_channel_t rmq_channel = 0;
 static amqp_bytes_t rmq_exchange;
 static amqp_bytes_t rmq_route_key;
+static gboolean durable = FALSE;
 
 static janus_mutex mutex;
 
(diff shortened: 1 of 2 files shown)


events/janus_rabbitmqevh.c
```diff
@@ -278,6 +279,14 @@ int janus_rabbitmqevh_init(const char *config_path) {
 		JANUS_LOG(LOG_INFO, "RabbitMQ event handler enabled, %s:%d (%s) exch: (%s) exchange_type:%s\n", rmqhost, rmqport, route_key, exchange,exchange_type);
 	}
 
+	item = janus_config_get(config, config_general, janus_config_type_item, "durable");
+	if(!item || !item->value || !janus_is_true(item->value)) {
+		JANUS_LOG(LOG_INFO, "RabbitMQEventHandler: Using non durable exchange & queue\n");
+	} else {
+		JANUS_LOG(LOG_INFO, "RabbitMQEventHandler: Using durable exchange & queue\n");
+		durable = TRUE;
+	}
```
lminiero, line 288: Can you move this block before the "exchange" property is read? The previous log line in fact is a summary of the whole configuration, and so assumes everything has been parsed already.

lminiero (commented):
Thanks for your contribution! I'm not familiar with the property, but I'm sure more people will benefit from this fix. Do you think it makes sense to add the same to the RabbitMQ transport plugin as well?

sakshamgupta05:
It makes sense to add this to transport as well. I will make the suggested changes.

chriswiggins:
Note that I've implemented this functionality (alongside autodelete and exclusive) in #2430 - we actually don't need a queue in the RabbitMQ event handler, because it never receives any events on a queue, just publishes them.

atoppi:
Is this still needed?

sakshamgupta05:
Closing this for now. Durable flag is applicable for exchanges too. I will raise another PR for that or re open this one once the @chriswiggins 's PR is merged.

PR #2434, opened 2020-11-12
Let SIP users cancel pending transactions without waiting for a 1xx

### LINKED ISSUE #2433: Hangup does not work between "calling" and "progress"
I was [asked on your mailing list](https://groups.google.com/g/meetecho-janus/c/XbZTqeO-lbk/m/H9JfoDioCgAJ) to open an issue here:

I am using the SIP plugin and it works mostly fine. However, if I start a call and hang up during the time it takes to go from the "calling" event to the "progress" event, then the hangup fails and the call is still getting connected.

I can reproduce this with your SIP demo:
- Go to https://janus.conf.meetecho.com/siptest.html
- Log in to our SIP provider
- Start a call to a phone number
- Click "Hangup" right away

What I expect to happen:
- The call terminates and does not even start ringing on the other end. I am able to do a new call.

What actually happens:
- The "Hangup" button becomes disabled, but the call continues. The phone starts ringing. I have no way to terminate the call. The JS console shows errors.

Here is a screenshot of this state:
Janus WebRTC Server: SIP Gateway Demo 2020-11-12 11-16-19

atoppi:
See #2433 for reference.

This PR lets SIP users register with a new option `rfc2543_cancel` in the `register` message (default is `false`).
The new option will make Janus abort pending transactions (like pending INVITEs) even when no provisional response has been received. This is useful e.g. to avoid stuck sessions/UI in case the SIP server is not answering to the INVITE.

Fixes #2433.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -110,6 +110,7 @@
 	"force_udp" : <true|false; if true, forces UDP for the SIP messaging; optional>,
 	"force_tcp" : <true|false; if true, forces TCP for the SIP messaging; optional>,
 	"sips" : <true|false; if true, configures a SIPS URI too when registering; optional>,
+	"rfc2543_cancel" : <true|false; if true, configures sip client to CANCEL pending INVITEs without having received a provisional response first; optional>,
 	"username" : "<SIP URI to register; mandatory>",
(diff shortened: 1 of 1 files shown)


lminiero:
Merging.

PR #2437, opened 2020-11-14
Fix incompatible-pointer-types compiler warnings

i8-pi:
When compiling with -Wincompatible-pointer-types, the compiler outputs a bunch of warnings like this one

    transports/janus_http.c: In function ‘janus_http_handler’:
    transports/janus_http.c:1337:58: warning: passing argument 3 of ‘MHD_get_connection_values’ from incompatible pointer type [-Wincompatible-pointer-types]
     1337 |   MHD_get_connection_values(connection, MHD_HEADER_KIND, &janus_http_headers, msg);
          |                                                          ^~~~~~~~~~~~~~~~~~~
          |                                                          |
          |                                                          int (*)(void *, enum MHD_ValueKind,  const char *, const char *)
    In file included from transports/janus_http.c:37:
    /usr/include/microhttpd.h:2699:49: note: expected ‘MHD_KeyValueIterator’ {aka ‘enum MHD_Result (*)(void *, enum MHD_ValueKind,  const char *, const char *)’} but argument is of type ‘int (*)(void *, enum MHD_ValueKind,  const char *, const char *)’
     2699 |                            MHD_KeyValueIterator iterator,
          |                            ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~

This change updates the declarations for MHD callback function to return MHD_Result

janus_http_notifier() should probably also be updated since it returns either an int or MHD_Result based on the code path, but the return value is not used anywhere

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -252,17 +252,17 @@ static GSource *janus_http_request_timeout_create(janus_transport_session *ts, j
 
 
 /* Callback (libmicrohttpd) invoked when a new connection is attempted on the REST API */
-static int janus_http_client_connect(void *cls, const struct sockaddr *addr, socklen_t addrlen);
+static enum MHD_Result janus_http_client_connect(void *cls, const struct sockaddr *addr, socklen_t addrlen);
 /* Callback (libmicrohttpd) invoked when a new connection is attempted on the admin/monitor webserver */
-static int janus_http_admin_client_connect(void *cls, const struct sockaddr *addr, socklen_t addrlen);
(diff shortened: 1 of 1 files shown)


lminiero:
I was aware of the warnings and their cause, but I explicitly didn't make this change, as it would break older versions of MHD (MHD_Result was only introduced very recently). This is confirmed by the Travis failure. As such, I'll have to close this. Thanks for the contribution anyway!

lminiero:
An alternative might be making your change, but also define MHD_Result to int if it doesn't exist. If you're willing to refactor the PR to do that, in order to remove the warning but keep previous versions working, happy to review!

PR #2440, opened 2020-11-18
duktape extensions

MaxKirillov:
Added JWT for duktape plugins and duktape version of videoroom with demo page.


lminiero:
Thanks for the PR! We're quite busy these days, so I won't be able to look at it right away. I see you added a new demo page, though, that I don't think should be there.

plugins/janus_duktape_extra.c
```diff
@@ -40,7 +198,15 @@ void janus_duktape_register_extra_functions(duk_context *ctx) {
...
-	duk_put_global_string(ctx, "testExtraFunction");
+
+	duk_push_c_function(ctx, http_get, 1);
+	duk_put_global_string(ctx, "http_get");
+
+	duk_push_c_function(ctx, http_post, 2);
+    duk_put_global_string(ctx, "http_post");
+
+    duk_push_c_function(ctx, jwt_validate, 2);
+    duk_put_global_string(ctx, "jwt_validate");
```
lminiero, line 211: All these changes don't belong to the `extra` file. They should be in a new *.c/*.h couple, like the existing duktape extensions we incorporated (see the console log stuff, for instance). Since you're adding two separate features (JWT and HTTP), I believe they should be two separate couples. Besides, as I pointed out, the whole jwt and http stuff shoulf be optional, depending on whether the libraries they depend on are available or not. We have plenty of examples in the code on how to check for libcurl presence and support, which you can use as a basis for both.

lminiero (commented):
Just checked the code, and I don't think it's in a state it can be merged, at the moment: there's several notes I added, that you can find inline.

I didn't comment yet on all the Duktape videoroom code, as there's more fundamental changes that need to be made first. I'm not sure they belong to the repo, though, as they seem to be more complex than the average demo (see videoroom.lua); besides, the code style would need to be changed to fit the one in our repo (indentation is different, for instance, and I see many references to a document generator that isn't our own). We can get back to this once the main points have been addressed, though.

plugins/janus_duktape_extra.c
voicenter: Many thanks for your comments,
We will do the necessary changes and do another push when the code will be ready for it , 
Many thanks...

lminiero:
I'm closing this as the codebase has changed a lot in the past few months. Should you decide to get back to this, please open a new PR.

PR #2444, opened 2020-11-22
Fix incompatible-pointer-types compiler warnings

i8-pi:
Continuing on from #2437 
> An alternative might be making your change, but also define MHD_Result to int if it doesn't exist.

Good idea. How about this new PR?

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -35,6 +35,13 @@
 #include <netdb.h>
 
 #include <microhttpd.h>
+#ifdef HAVE_ENUM_MHD_RESULT
+	/* enum MHD_Result introduced in libmicrohttpd v0.9.71 */
+	typedef enum MHD_Result MHD_Result;
+#else
+	typedef int MHD_Result;
+#endif
+
 
 #include "../debug.h"
 #include "../apierror.h"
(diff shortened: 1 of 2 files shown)


transports/janus_http.c
```diff
@@ -35,6 +35,13 @@
 #include <netdb.h>
 
 #include <microhttpd.h>
+#ifdef HAVE_ENUM_MHD_RESULT
+	/* enum MHD_Result introduced in libmicrohttpd v0.9.71 */
+	typedef enum MHD_Result MHD_Result;
+#else
+	typedef int MHD_Result;
+#endif
```
lminiero, line 43: Shouldn't this just be an `#ifndef HAVE_ENUM_MHD_RESULT` to define it to `int` for older MHD versions? Won't this redefine something that exists already, for newer versions, and possibly cause issues?

lminiero (commented):
Thanks! Added a single note inline.

transports/janus_http.c
lminiero: I guess MHD just uses `enum MHD_Result`, but doesn't typdef's it as `MHD_Result`, correct?

lminiero: Verified looking at `microhttpd.h` that they don't indeed typedef `MHD_Result`, so no issue in your code there.

lminiero:
Thanks, merging!

PR #2445, opened 2020-11-22
Cleanup calls to setsockopt et al

tmatth:
For constants, we can directly use anonymous arrays in C for calls to setsockopt and similar instead of declaring variables that aren't used otherwise..

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -1829,8 +1829,7 @@ static int janus_audiobridge_create_udp_socket_if_needed(janus_audiobridge_room 
 		JANUS_LOG(LOG_ERR, "Could not open UDP socket for RTP forwarder (room %s)\n", audiobridge->room_id_str);
 		return -1;
 	}
-	int v6only = 0;
-	if(setsockopt(audiobridge->rtp_udp_sock, IPPROTO_IPV6, IPV6_V6ONLY, &v6only, sizeof(v6only)) != 0) {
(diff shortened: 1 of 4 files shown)


lminiero:
Is this widespread and common enough? I wouldn't want to cause issues with some compilers that may not support it.

atoppi:
The code is perfectly fine but imho it is worse in terms of readability.

lminiero:
I tend to agree with Alessandro: I'd rather keep it as it is. Possibly a bit more verbose, but clearer. Closing. but thanks anyway!

tmatth:
> Is this widespread and common enough? I wouldn't want to cause issues with some compilers that may not support it.

Yes FWIW any C99 compiler must support anonymous arrays like this, and Janus implicitly requires C99 (i.e. it won't compile if you force strict C90). 

> The code is perfectly fine but imho it is worse in terms of readability.

Yeah I could go either way, I'd at least make the parameters const though, currently there's a mix of setsockopt where the parameters are really variables and others where they are read-only but not declared const.

atoppi:
@tmatth I agree on using `const` qualifier for parameters passed to `setsockopt`.
As of the C99 thing, I didn't know that. Do you think that making the code compatibile to C90 might be worth the effort ?

tmatth:
> As of the C99 thing, I didn't know that. Do you think that making the code compatibile to C90 might be worth the effort ?

No not really, I'd be more inclined to go the other route and explicitly require C99 support, as C99 and later have a lot of nice features (and 21 years old is pretty mature IMO).

PR #2450, opened 2020-11-25
Added configurable mechanism for manually setting static event loop to use for new handles

lminiero:
For some context, please refer to the exchange [this comment](https://github.com/meetecho/janus-gateway/pull/2295#issuecomment-709337448) belongs to, as it's where the idea came from.

To summarize, normally when you use the static `event_loops` configuration, new handles are automatically added to loops following a round robin distribution. This patch allows you to configure Janus so that you can also optionally use the Janus API itself to tell the core which loop a new handle you're attaching should be added to. The discussion above provides some information on why it may be useful in some cases.

To enable the functionality, `event_loops` must be configured, and `allow_loop_indication` set to `true` (it's `false` by default). In that case, adding a `loop_index` to an `attach` request will instruct the core to manually add the handle to the specified core, rather than doing it automatically. Providing an invalid index will fall back to round robin.

This PR is based on the `sendmmsg` branch, as #2295 is where we're trying to work on most optimizations. Please test and provide feedback, as that's the fastest way to get this merged soon.

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -105,6 +105,18 @@ general: {
 									# As such, if you want to use this you should
 									# provision the correct value according to the
 									# available resources (e.g., CPUs available).
+	#allow_loop_indication = true	# In case a static number of event loops is
+									# configured as explained above, by default
+									# new handles will be allocated on one loop or
+									# another by the Janus core itself. In some cases
+									# it may be helpful to manually tell the Janus
+									# core which loop a handle should be added to,
+									# e.g., to group viewers of the same stream on
+									# the same loop. This is possible via the Janus
+									# API when performing the 'attach' request, but
+									# only if allow_loop_indication is set to true;
+									# it's set to false by default to avoid abuses.
+									# Don't change if you don't know what you're doing!
 	#opaqueid_in_api = true			# Opaque IDs set by applications are typically
 									# only passed to event handlers for correlation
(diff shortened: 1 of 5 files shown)


lminiero:
Merging in the `sendmmsg` branch.

fancycode:
As #2295 is on hold, any chance this PR can be merged to master? It sounds like a good optimization on its own.

lminiero:
I actually forgot about this... I'll check how easy it is to backport to master, since this has already been merged in that other branch.

lminiero:
Done. I don't have time to test this, and to be honest it's very low priority for me. If you do care about this, please test and provide feedback. If I don't get any, I'll assume people don't care and just close it (no reason to merge and maintain it otherwise).

fancycode:
Thanks a lot! I'll test once I get to implement the API and will provide feedback in the other PR.

Out of curiosity: how do you configure servers to handle lots of streams? With dynamically created loops or a static `event_loops`  configuration? There seems to be no "best practices" document (or at least I could not find one).

cb22:
@fancycode FWIW, the way we run in production is with some tooling around routing different users to different janus instances, than running multiple instances each with `event_loops = 1`.

The per process overhead is quite low (a few MB RAM), it essentially does the same thing as mentioned in the comment linked to above (I think! I'll admit I don't have hard data here) and has the added benefit of meaning a crash / issue affects fewer users than if everyone were on a single instance.

PR #2451, opened 2020-11-25
[janus-pp-rec] Add in-band opus FEC support.

atoppi:
This is a highly experimental patch that lets janus-pp-rec use opus in-band FEC data to recover lost packets.
The implementation has been inspired by the audiobridge plugin, where the FEC  support has been already implemented.
The current janus-pp-rec detects RTP discontinuities in opus recordings and fills them with opus silence packets. 
With this patch janus-pp-rec accepts a new CLI parameter `--fec, -F` that will try to recover lost packets using opus FEC data.

In order to support FEC pp-rec **must decode and re-encode the packets**, thus leading to longer processing times. Conversely, by sticking to the traditional fill-with-silence approach, pp-rec will just output an ogg file with the payloads stored in the available RTP packets, filling the voids with blank opus packets.

Two important notes to consider about opus fec in WebRTC:
1) in-band opus FEC is only capable of recovering ONE single packet (the previous one in the stream)
2) in-band opus FEC seems to kick-in only when severe packet loss is being experienced

**What I expect from testers**
I'm having some difficulties trying to validate the effects of FEC recovery. Maybe someone (more experienced than me!) is able to verify that a packet is correctly restored by inspecting the bitstream.
What is weird is that in case of lost bursts, all the packets that might not be recovered (e.g. the N-8, N-7 ... N-2 ths) seems to contain some data. I'd expect silence in that spots, so this is making me question the validity of the implementation and approach.
Due to previous points 1) and 2) testing the patch is not very easy, so feedback is very very welcome.

### DIFF
--- a/postprocessing/pp-opus.c
+++ b/postprocessing/pp-opus.c
@@ -26,10 +26,13 @@
 #include "../debug.h"
 #include "../version.h"
 
+#define	OPUS_SAMPLES	960
+#define	BUFFER_SAMPLES	OPUS_SAMPLES*6
 
 /* OGG/Opus helpers */
 FILE *ogg_file = NULL;
 ogg_stream_state *stream = NULL;
+gboolean use_fec = FALSE;
 
 void le32(unsigned char *p, int v);
 void le16(unsigned char *p, int v);
@@ -41,7 +44,7 @@ int ogg_write(void);
 int ogg_flush(void);
 
 
-int janus_pp_opus_create(char *destination, char *metadata) {
+int janus_pp_opus_create(char *destination, gboolean enable_opus_fec, char *metadata) {
 	stream = g_malloc0(sizeof(ogg_stream_state));
 	if(ogg_stream_init(stream, rand()) < 0) {
 		JANUS_LOG(LOG_ERR, "Couldn't initialize Ogg stream state\n");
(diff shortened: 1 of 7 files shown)


lminiero:
My feeling is that this functionality should only optionally depend on libopus, since the library is not a hard requirement for Janus itself (the AudioBridge plugin is optional, and the Streaming plugin can be built without it). I'm not a fan of the idea of decoding/re-encoding, as besides the additional time/computing that would be needed we don't know what the original quality of the recording was (was the source 16kz? 48? stereo or mono? what complexity was used initially?) which will inevitably have an impact on the end result; that said, I guess there's not much else we can do about this, if we want to take advantage of the FEC data to fix broken recordings. In the future we may want to check if the redundant information can be extracted manually without transcoding, to see if we can restore some missing packets ourselves that way.

atoppi:
I've added the sampling rate and mono/stereo evaluation through bistream analysis.
I'm removing the draft label now as I think this should be complete now.

postprocessing/pp-opus.c
```diff
@@ -72,31 +78,147 @@ int janus_pp_opus_process(FILE *file, janus_pp_frame_packet *list, int *working)
...
+	int max_frame_size = 120;
+	int channels = 1;
+	int sampling_rate = 8000;
+	gsize buflen;
+	if(use_fec) {
+		/* Check content paramters (mono/stereo, bandwidth) */
+		while(*working && tmp != NULL) {
+			len = 0;
+			/* RTP payload */
+			offset = tmp->offset+12+tmp->skip;
```
fippo, line 96: 12 is just the default rtp header length, are extensions stripped elsewhere?

atoppi: `tmp->skip` should contain the additional offset

lminiero:
Closing for the time being, we may get back to it again in the future.

Issue #1360, opened 2018-09-11
RTP Input for Audiobridges (Enhancement)


austinmarkus:
It would be really useful for me to have the ability to input RTP  to an audiobridge in a similar manner to how there is RTP output via the  forwarder. The basic request is to have the ability to send an Opus encoded RTP stream from a non-browser such that it could be heard the members of the bridge. 

If there is already a way to accomplish this that I have overlooked I'm open to suggestion.

Thank you for your amazing project.

lminiero:
This is currently unsupported in the AudioBridge. I can add it to the TODO list, but I honestly don't know when I'd be able to work on that (busy times!). PRs always welcome, of course!

An easier way to accomplish the same thing with what's available today would be by taking advantage of the Streaming plugin: create an audio-only Streaming mountpoint, have your users create a PeerConnection with that (e.g., just for receiving announcements), and anytime you have external RTP to send you send it there. Suboptimal, of course (you need two PC instead of one) but solves your requirement without touching the code.

lminiero:
FYI, we recently added support for playing static Opus files in the AudioBridge in #2088, so that should already help with that you need. Actual support for incoming RTP streaming will take longer, as it's quite low in our priority list right now.

lminiero:
@austinmarkus please see #2464.

austinmarkus:
Thank you!

PR #2464, opened 2020-12-01
Add support for plain RTP participants in AudioBridge

lminiero:
This PR adds support for plain RTP participants to the AudioBridge. By "plain RTP" I mean that, up until now, to join an AudioBridge mix you _had_ to use WebRTC: this patch adds a new mode, where you can issue a "join" without actually establishing a PeerConnection, but exchange plain ip/ports instead to be able to send media via plain RTP directly to the mix, and receive it via plain RTP as well. Use cases include manual injection of external RTP input (which is what #1360 was about), but also facilitating the integration with legacy endpoints (e.g., to have SIP users join a Janus AudioBridge).

In order to re-use as much of the existing code as possible, this patch basically just extends the existing API. This means that, to create a plain RTP participant, you _still_ need to attach a Janus handle as you normally would, and _still_ use the `join` request to let the participant join the room: the difference is that you don't provide any JSEP SDP to negotiate a PeerConnection, but provide an `rtp` object instead with some info on how the manual RTP connection should be created; the plugin sends back a similar `rtp` object back in the `joined` event, and so the external RTP endpoint (which will be seen as a normal participant) and the AudioBridge plugin can send RTP packets to each other.

The syntax for the `rtp` object is quite simple. An example of a `join` request might be this:

	var join = {
		request: "join",
		room: 1234,
		display: "Plain RTP participant",
		rtp: {
			ip: "192.168.1.10",
			port: 2468,
			audiolevel_ext: 1,
			payload_type: 111,
			fec: true
		}
	}

This will tell the plugin that this participant will use plain RTP, and will be listening on the specified ip/port address; if ip/port are missing, it means this participant will _only_ send media, and will not receive any (e.g., for injecting audio), even though resources will still be created for that (encoder, etc.). There's a few additional properties you can provide, like the audio-level extension ID, the payload type to use, and whether FEC should be enabled (basically what you'd normally expect to find in the SDP).

The `joined` event from the plugin will look exactly as it normally does, but will include an `rtp` object of its own, e.g.:

```
{
        "audiobridge" : "joined",
        "room" : <numeric ID of the room>,
        "id" : <unique ID assigned to the participant>,
        "display" : "<display name of the new participant>",
        "participants" : [
                // Array of existing participants in the room
        ],
        "rtp": {
                "ip": "<ip of the plugin">,
                "port": <port the plugin will wait for RTP traffic on>
        }
}
```

As an application orchestrating an RTP participant, that's all you need to know to complete the setup (where to send RTP packets, and where to expect them from). Notice that the plugin doesn't do any ICE or complex stuff, here: just as the SIP and NoSIP plugins, it expects to be able to communicate on those addresses directly.

I've tested this by modifying the NoSIP demo to connect to the AudioBridge via plain RTP, and it seems to be working as expected. Of course, more testing will need to happen, as I expect issues/crashes/leaks to arise here and there: probably some fine tuning will be needed, which is why for the time being I'll open this as a draft PR. If this interests you, please start testing and provide feedback.


lminiero:
I've made some more tests for an internal scenario, and it seems to be working as expected for me. I'll wait for (more) feedback before merging, but I consider this effort complete. Unless I see objections, this will be merged soon.

lminiero:
I've changed the default for the new feature to `FALSE`, which means that out of the box the plugin will not accept plain RTP Participants: this is to ensure consistency with the current behaviour.

lminiero:
Merging.

PR #2465, opened 2020-12-02
SIP plugin: custom headers in "decline" request

ihusejnovic:
Hi,

This PR adds an option to send custom headers in the "decline" request.


lminiero:
Looks good, thanks! Sooner or later we'll end up with custom headers for every single request :hand_over_mouth: 
Merging :+1:

PR #2467, opened 2020-12-02
[janus-pp-rec] Drop audio RTP silence suppression packets.

### LINKED ISSUE #2414: [janus-pp-rec] Huge opus files when received timestamp is 0 (see #2328)
Since #2328 contained a patch with a not acceptable solution, we are creating this issue to keep track of the post-processor problem described in the PR.
Please refer to #2328 for more context.

atoppi:
This PR aims at detecting RTP silence suppression (aka Discontinuous Transmission DTX) packets in `janus-pp-rec` and dropping them in order to avoid huge files being generated like described in #2328.

Quoting RFC 3389
```
   RTP allows discontinuous transmission (silence suppression) on any audio payload
   format.  The receiver can detect silence suppression on the first
   packet received after the silence by observing that the RTP timestamp
   is not contiguous with the end of the interval covered by the
   previous packet even though the RTP sequence number has incremented
   only by one.  The RTP marker bit is also normally set on such a
   packet.
```

The added check basically looks for audio packets with marker bit=1, close enough (in range of 100) to the highest sequence number received. If that packet RTP timestamp is very different (> 10x) from the expected one, we simply drop that packet.

A new CLI option has been added (`--silence-distance`) to control the range of packets for which the silence suppression detection acts (default is 100, while 0 means disabled).

Tested against a recording with a couple of DTX packets and it seems to work.

Fixes #2414.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -635,6 +635,8 @@ int main(int argc, char *argv[])
 	int ignored = 0;
 	offset = 0;
 	gboolean started = FALSE;
+	/* DTX stuff */
+	gboolean dtx_on = FALSE;
 	/* Extensions, if any */
 	int audiolevel = 0, rotation = 0, last_rotation = -1, rotated = -1;
 	uint16_t rtp_header_len, rtp_read_n;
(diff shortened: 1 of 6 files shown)


postprocessing/janus-pp-rec.c
```diff
@@ -801,8 +802,29 @@ int main(int argc, char *argv[])
 			/* Simple enough... */
 			started = TRUE;
 			highest_rtp_ts = rtp_ts;
+			highest_seq = p->seq;
 			p->ts = (times_resetted*max32)+rtp_ts;
 		} else {
+			if(!video && !data && rtp->markerbit == 1) {
+				/* Try to detect RTP silence suppression */
+				int32_t seq_distance = abs((int16_t)(p->seq - highest_seq));
+				if(seq_distance < 100) {
```
lminiero, line 811: I guess the `100` here could be made configurable (default=100) via command line, in case it's too aggressive (or not aggressive enough) in some specific cases.

atoppi: done

lminiero:
Feedback is positive, so I'll merge :v:

PR #2468, opened 2020-12-02
Replay data channel recordings

ricardo-salgado-tekever:
Added capability to replay data recordings in recordplay plugin from streaming plugin.

### DIFF
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -316,7 +316,6 @@ $(document).ready(function() {
 								},
 								ondataopen: function(data) {
 									Janus.log("The DataChannel is available!");
-									$('#waitingvideo').remove();
 									$('#videobox').append(
 										'<input class="form-control" type="text" id="datarecv" disabled></input>'
 									);
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -147,7 +147,7 @@ $(document).ready(function() {
 												recordplay.createAnswer(
 													{
 														jsep: jsep,
-														media: { audioSend: false, videoSend: false },	// We want recvonly audio/video
+														media: { audioSend: false, videoSend: false, data: true },	// We want recvonly audio/video
 														success: function(jsep) {
 															Janus.debug("Got SDP!", jsep);
 															var body = { request: "start" };
@@ -313,6 +313,20 @@ $(document).ready(function() {
 										$('#videobox .no-video-container').remove();
 										$('#thevideo').removeClass('hide').show();
 									}
+								},
+								ondataopen: function(data) {
+									Janus.log("The DataChannel is available!");
+									$('#waitingvideo').remove();
+									$('#videobox').append(
+										'<input class="form-control" type="text" id="datarecv" disabled></input>'
+									);
+									if(spinner)
+										spinner.stop();
+									spinner = null;
+								},
+								ondata: function(data) {
+									Janus.debug("We got data from the DataChannel!", data);
+									$('#datarecv').val(data);
 								},
 								oncleanup: function() {
 									Janus.log(" ::: Got a cleanup notification :::");
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -319,9 +319,6 @@ $(document).ready(function() {
 									$('#videobox').append(
 										'<input class="form-control" type="text" id="datarecv" disabled></input>'
 									);
-									if(spinner)
-										spinner.stop();
-									spinner = null;
 								},
 								ondata: function(data) {
 									Janus.debug("We got data from the DataChannel!", data);
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -2687,7 +2687,7 @@ static void *janus_recordplay_playout_thread(void *sessiondata) {
 				int offset = data->offset;
 				fseek(dfile, data->offset, SEEK_SET);
 				bytes = fread(&when, sizeof(gint64), 1, dfile);
-				when = ntohll(when); // NOTE: not currently used - playback is interested in actual data packets.
+				when = ntohll(when); /* NOTE: not currently used - playback is interested in actual data packets. */
 				offset += sizeof(gint64);
 				len -= sizeof(gint64);
 
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -2711,8 +2711,33 @@ static void *janus_recordplay_playout_thread(void *sessiondata) {
 			}
 		}
 		if(data) {
-			if(data == session->dframes) {
-				/* First packet, send now */
+			u_int64_t prev_ts = 0; /* All timestamps for data are indexed to 0, since when parsing ts = when - c_time */
+			if(data->prev) 
+				prev_ts = data->prev->ts;
+			ts_diff = data->ts - prev_ts;
+
+			/* Check if it's time to send */
+			gettimeofday(&now, NULL);
+			d_s = now.tv_sec - dbefore.tv_sec;
+			d_us = now.tv_usec - dbefore.tv_usec;
+			if(d_us < 0) {
+				d_us += 1000000;
+				--d_s;
+			}
+			passed = d_s*1000000 + d_us;
+			if(passed < (ts_diff-5000)) {
+				dsent = FALSE;
+			} else {
+				/* Update the reference time */
+				dbefore.tv_usec += ts_diff%1000000;
+				if(dbefore.tv_usec > 1000000) {
+					dbefore.tv_sec++;
+					dbefore.tv_usec -= 1000000;
+				}
+				if(ts_diff/1000000 > 0) {
+					dbefore.tv_sec += ts_diff/1000000;
+					dbefore.tv_usec -= ts_diff/1000000;
+				}
 				/* Read data packet */
 				fseek(dfile, data->offset, SEEK_SET);
 				bytes = fread(buffer, sizeof(char), data->len, dfile);
@@ -2722,59 +2747,13 @@ static void *janus_recordplay_playout_thread(void *sessiondata) {
 				janus_plugin_data datapacket = {
 					.label = NULL,
 					.protocol = NULL,
-					.binary = rec->textdata? FALSE : TRUE, 
+					.binary = rec->textdata ? FALSE : TRUE, 
 					.buffer = (char *)buffer,
 					.length = bytes
 				};
 				gateway->relay_data(session->handle, &datapacket);
-				
-				gettimeofday(&now, NULL);
-				dbefore.tv_sec = now.tv_sec;
-				dbefore.tv_usec = now.tv_usec;
 				dsent = TRUE;
 				data = data->next;
-			} else {
-				/* What's the timestamp skip from the previous packet? */
-				ts_diff = data->ts - data->prev->ts;
-				/* Check if it's time to send */
-				gettimeofday(&now, NULL);
-				d_s = now.tv_sec - dbefore.tv_sec;
-				d_us = now.tv_usec - dbefore.tv_usec;
-				if(d_us < 0) {
-					d_us += 1000000;
-					--d_s;
-				}
-				passed = d_s*1000000 + d_us;
-				if(passed < (ts_diff-5000)) {
-					dsent = FALSE;
-				} else {
-					/* Update the reference time */
-					dbefore.tv_usec += ts_diff%1000000;
-					if(dbefore.tv_usec > 1000000) {
-						dbefore.tv_sec++;
-						dbefore.tv_usec -= 1000000;
-					}
-					if(ts_diff/1000000 > 0) {
-						dbefore.tv_sec += ts_diff/1000000;
-						dbefore.tv_usec -= ts_diff/1000000;
-					}
-					/* Read data packet */
-					fseek(dfile, data->offset, SEEK_SET);
-					bytes = fread(buffer, sizeof(char), data->len, dfile);
-					if(bytes != data->len)
-						JANUS_LOG(LOG_WARN, "Didn't manage to read all the bytes we needed (%d < %d)...\n", bytes, data->len);
-					/* Update payload type */
-					janus_plugin_data datapacket = {
-						.label = NULL,
-						.protocol = NULL,
-						.binary = rec->textdata ? FALSE : TRUE, 
-						.buffer = (char *)buffer,
-						.length = bytes
-					};
-					gateway->relay_data(session->handle, &datapacket);
-					dsent = TRUE;
-					data = data->next;
-				}
 			}
 		}
 	}
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -261,7 +261,6 @@
 
 #include <dirent.h>
 #include <arpa/inet.h>
-#include <netinet/udp.h>
 #include <sys/stat.h>
 #include <sys/time.h>
 #include <jansson.h>
@@ -406,7 +405,7 @@ typedef struct janus_recordplay_recording {
 	char *vfmtp;				/* Video fmtp, if any */
 	int video_pt;				/* Payload types to use for audio when playing recordings */
 	char *drc_file;				/* Data file name */
-	gboolean textdata;			/* Data format is text */
+	gboolean textdata;			/* Whether data format is text */
 	char *offer;				/* The SDP offer that will be sent to watchers */
 	gboolean e2ee;				/* Whether media in the recording is encrypted, e.g., using Insertable Streams */
 	GList *viewers;				/* List of users watching this recording */
@@ -498,7 +497,7 @@ void janus_recordplay_send_rtcp_feedback(janus_plugin_session *handle, int video
 #define AUDIO_PT		111
 #define VIDEO_PT		100
 
-#define UDPHDR_SIZE		8
+#define DATATS_SIZE	8
 
 /* Helper method to check which codec was used in a specific recording (and if it's end-to-end encrypted) */
 static const char *janus_recordplay_parse_codec(const char *dir, const char *filename, char *fmtp, size_t fmtplen, gboolean *e2ee) {
@@ -626,7 +625,7 @@ static const char *janus_recordplay_parse_codec(const char *dir, const char *fil
 				const char *c = json_string_value(codec);
 				if (data) {
 					/* Found! */
-					c = !strcasecmp(c, "text")? "text" : "bin";
+					c = !strcasecmp(c, "text") ? "text" : "binary";
 					json_decref(info);
 					fclose(file);
 					return c;					
@@ -2072,8 +2071,10 @@ void janus_recordplay_update_recordings_list(void) {
 			char *ext = strstr(rec->drc_file, ".mjr");
 			if(ext != NULL)
 				*ext = '\0';
-			rec->textdata = !strcasecmp("text", janus_recordplay_parse_codec(recordings_path,
-				rec->drc_file, NULL, sizeof(NULL), NULL));
+			const char *textcodec = janus_recordplay_parse_codec(recordings_path,
+				rec->drc_file, NULL, sizeof(NULL), NULL);
+			if (textcodec)
+				rec->textdata = !strcasecmp("text", textcodec);
 		}
 		rec->audio_pt = AUDIO_PT;
 		if(rec->acodec != JANUS_AUDIOCODEC_NONE) {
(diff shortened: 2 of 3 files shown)


januscla:
Thanks for your contribution, @ricardo-salgado-tekever! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

plugins/janus_recordplay.c
```diff
@@ -2618,6 +2677,74 @@ static void *janus_recordplay_playout_thread(void *data) {
...
+					/* Send now */				
+					fseek(dfile, data->offset+UDPHDR_SIZE, SEEK_SET);
+					bytes = fread(buffer, sizeof(char), data->len-UDPHDR_SIZE, dfile);
+					if(bytes != data->len-UDPHDR_SIZE)
+						JANUS_LOG(LOG_WARN, "Didn't manage to read all the bytes we needed (%d < %d)...\n", bytes, data->len);
+					/* Update payload type */
+					janus_plugin_data datapacket = {
+						.label = NULL,
+						.protocol = NULL,
+						.binary = rec->textdata? FALSE : TRUE,
```
lminiero, line 2738: Code style: missing space before the question mark.

plugins/janus_streaming.c
```diff
@@ -4000,7 +4000,7 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 			}
 			if(data) {
 				const char *datafile = json_string_value(data);
-				drc = janus_recorder_create(NULL, "text", (char *)datafile);
+				drc = janus_recorder_create(NULL, source->textdata ? "text" : "bin", (char *)datafile);
```
lminiero, line 4003: This is wrong. The recorder currently only supports "text", and will fail if you pass "bin" (or "binary"). Since we'll have to fix this in the core, please modify the PR so that only "text" is supported for the moment.

lminiero (commented):
Thanks! But I noticed several issues. You can find my comments inline.

html/recordplaytest.js
```diff
@@ -147,7 +147,7 @@ $(document).ready(function() {
 												recordplay.createAnswer(
 													{
 														jsep: jsep,
-														media: { audioSend: false, videoSend: false },	// We want recvonly audio/video
+														media: { audioSend: false, videoSend: false, data: true },	// We want recvonly audio/video
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: This has no effect on the demo - it will still work if there are no data channels on the recoring. On the other hand, if not present and someone wants to test using recordings from the streaming plugin, it would need a change to work - datachannel will not open from server side otherwise.

plugins/janus_recordplay.c
```diff
@@ -261,6 +261,7 @@
 
 #include <dirent.h>
 #include <arpa/inet.h>
+#include <netinet/udp.h>
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Removed it.

plugins/janus_recordplay.c
```diff
@@ -404,6 +405,8 @@ typedef struct janus_recordplay_recording {
 	janus_videocodec vcodec;	/* Codec used for video, if available */
 	char *vfmtp;				/* Video fmtp, if any */
 	int video_pt;				/* Payload types to use for audio when playing recordings */
+	char *drc_file;				/* Data file name */
+	gboolean textdata;			/* Data format is text */
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Corrected.

plugins/janus_recordplay.c
```diff
@@ -493,6 +498,8 @@ void janus_recordplay_send_rtcp_feedback(janus_plugin_session *handle, int video
 #define AUDIO_PT		111
 #define VIDEO_PT		100
 
+#define UDPHDR_SIZE		8
+
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Removed.

plugins/janus_recordplay.c
```diff
@@ -2618,6 +2677,74 @@ static void *janus_recordplay_playout_thread(void *data) {
...
+				
+				gettimeofday(&now, NULL);
+				dbefore.tv_sec = now.tv_sec;
+				dbefore.tv_usec = now.tv_usec;
+				dsent = TRUE;
+				data = data->next;
+			} else {
+				/* What's the timestamp skip from the previous packet? */
+				ts_diff = data->ts - data->prev->ts;
+				ts_diff = (ts_diff)/dhz;
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Removed.

plugins/janus_recordplay.c
```diff
@@ -2618,6 +2677,74 @@ static void *janus_recordplay_playout_thread(void *data) {
 				}
 			}
 		}
+		if(data) {
+			if(data == session->dframes) {
+				/* First packet, send now */
+				/* Data recording stores raw UDP packets */
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Corrected to extract recording timestamp. For the timebeing, the recording timestamp is not used, since the frame timestamp (data->ts) can be used directly, as it follows the recording timestamp.
It is always: recording_timestamp(current_frame) = timestamp(current_frame) + recording_timestamp(initial_frame).

plugins/janus_recordplay.c
ricardo-salgado-tekever: Corrected.

plugins/janus_streaming.c
ricardo-salgado-tekever: Corrected.

plugins/janus_recordplay.c
```diff
@@ -2482,15 +2539,17 @@ static void *janus_recordplay_playout_thread(void *data) {
 	if(audio_pt == 0 || audio_pt == 8 || audio_pt == 9)
 		akhz = 8;
 	int vkhz = 90;
+	int dhz = 1;
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Removed.

plugins/janus_recordplay.c
```diff
@@ -615,6 +624,13 @@ static const char *janus_recordplay_parse_codec(const char *dir, const char *fil
 					return NULL;
 				}
 				const char *c = json_string_value(codec);
+				if (data) {
+					/* Found! */
+					c = !strcasecmp(c, "text")? "text" : "bin";
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Corrected.

plugins/janus_recordplay.c
```diff
@@ -2041,6 +2067,14 @@ void janus_recordplay_update_recordings_list(void) {
 			if(e2ee)
 				rec->e2ee = TRUE;
 		}
+		if(data && data->value) {
+			rec->drc_file = g_strdup(data->value);
+			char *ext = strstr(rec->drc_file, ".mjr");
+			if(ext != NULL)
+				*ext = '\0';
+			rec->textdata = !strcasecmp("text", janus_recordplay_parse_codec(recordings_path,
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Corrected.
Since there is no other data codec supported besides "text" anyway, will just treat anything put here as binary.

plugins/janus_recordplay.c
```diff
@@ -2226,7 +2261,7 @@ janus_recordplay_frame_packet *janus_recordplay_get_frames(const char *dir, cons
 				}
 				w_time = json_integer_value(created);
 				/* Summary */
-				JANUS_LOG(LOG_VERB, "This is %s recording:\n", video ? "a video" : "an audio");
+				JANUS_LOG(LOG_VERB, "This is %s recording:\n", video ? "a video" : audio ? "an audio" : "a data");
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Corrected.

html/recordplaytest.js
```diff
@@ -314,6 +314,20 @@ $(document).ready(function() {
...
 								},
+								ondataopen: function(data) {
+									Janus.log("The DataChannel is available!");
+									$('#waitingvideo').remove();
+									$('#videobox').append(
+										'<input class="form-control" type="text" id="datarecv" disabled></input>'
+									);
+									if(spinner)
+										spinner.stop();
+									spinner = null;
```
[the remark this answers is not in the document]
ricardo-salgado-tekever: Removed.

html/recordplaytest.js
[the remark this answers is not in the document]
lminiero: You're right, sorry! I thought this was the recording part, not the playout one :see_no_evil:

plugins/janus_recordplay.c
```diff
@@ -615,6 +621,13 @@ static const char *janus_recordplay_parse_codec(const char *dir, const char *fil
 					return NULL;
 				}
 				const char *c = json_string_value(codec);
+				if (data) {
```
lminiero, line 624: Code style: no space between `if` and bracket.

plugins/janus_recordplay.c
```diff
@@ -615,6 +621,13 @@ static const char *janus_recordplay_parse_codec(const char *dir, const char *fil
...
+					/* Found! */
+					c = !strcasecmp(c, "text") ? "text" : "binary";
```
lminiero, line 626: Just put "text" here: as I said in a previous comment, we currently don't support saving binary data in mjr yet anyway. That will be done in another effort later on, so I'll take care of updating the code here when that happens.

plugins/janus_recordplay.c
```diff
@@ -2041,6 +2064,16 @@ void janus_recordplay_update_recordings_list(void) {
...
 				rec->e2ee = TRUE;
 		}
+		if(data && data->value) {
+			rec->drc_file = g_strdup(data->value);
+			char *ext = strstr(rec->drc_file, ".mjr");
+			if(ext != NULL)
+				*ext = '\0';
+			const char *textcodec = janus_recordplay_parse_codec(recordings_path,
+				rec->drc_file, NULL, sizeof(NULL), NULL);
+			if (textcodec)
```
lminiero, line 2074: See above for code style note on `if (`.

plugins/janus_recordplay.c
```diff
@@ -2618,6 +2677,98 @@ static void *janus_recordplay_playout_thread(void *data) {
...
+			if(data == session->dframes) {
+				/* First packet, send now */
+				/* Data recording stores recording timestamp in first 8 bytes - it follows the frame ts monotonically.
+					invariant: when = data->ts + when(initial packet) */
+				gint64 when = 0;
+				int len = data->len;
+				int offset = data->offset;
+				fseek(dfile, data->offset, SEEK_SET);
+				bytes = fread(&when, sizeof(gint64), 1, dfile);
+				when = ntohll(when); // NOTE: not currently used - playback is interested in actual data packets.
```
lminiero, line 2690: Code style: comments should always be in `/*  */` format.

plugins/janus_recordplay.c
```diff
@@ -2618,6 +2677,98 @@ static void *janus_recordplay_playout_thread(void *data) {
...
+				offset += sizeof(gint64);
+				len -= sizeof(gint64);
+
+				/* Read data packet */
+				fseek(dfile, offset, SEEK_SET);
+				bytes = fread(buffer, sizeof(char), len, dfile);
+				JANUS_LOG(LOG_INFO, "Sending data packet at rtp_timestamp = %lu, timestamp = %lu, delta = %lu\n", (data->ts), when, when - data->ts);
```
lminiero, line 2697: This shouldn't be `LOG_INFO`, but `LOG_HUGE` (unless it's just there for your own debugging?).

lminiero (commented):
Thanks for the quick fixes! Added a few other notes.

That said, I think this still needs work: I don't see any change to how `janus_recordplay_get_frames` processes frames, and I see the same method is used as it is for data channels. This means that the code will try to reorder frames working with a non-existing RTP header, which will not work, and will not time the messages correctly. Please see how `janus-pp-rec.c` traverses data recordings in a different way (it should be easy enough to port as it's a separate check), by skipping to the right offset and most importantly using the time information in the mjr to time the messages properly.

plugins/janus_recordplay.c
ricardo-salgado-tekever: Corrected.

plugins/janus_recordplay.c
ricardo-salgado-tekever: Corrected.

plugins/janus_recordplay.c
ricardo-salgado-tekever: Agreed!

plugins/janus_recordplay.c
ricardo-salgado-tekever: Yup, just for my debugging. Corrected.

plugins/janus_recordplay.c
ricardo-salgado-tekever: Corrected.

ricardo-salgado-tekever:
Interestingly, even without changes to `janus_recordplay_get_frames`, this does actually work. And the timing comes out OK, as odd as it may sound.
What is happening is that the non-existent RTP header fields end up aligning the timestamp field with the least significant word of the timestamp :-D - and it works!
Nonetheless, this is not a proper solution, so I will port from `janus-pp-rec.c`.

One more point, it looks like there is a bug there reading the timestamp: this `bytes = fread(&when, sizeof(gint64), 1, file);` should return 1  instead of bytes. 
Per documentation:

> **Return Value**
> The total number of elements successfully read are returned as a size_t object, which is an integral data type. If this number differs from the nmemb parameter, then either an error had occurred or the End Of File was reached.
>

ricardo-salgado-tekever:
Now with a proper implementation on ´ janus_recordplay_get_frames ´ for data recordings, based no ´janus-pp-rec.c´.
Note that ´janus-pp-rec.c´ has a bug in reading data packets header, I will submit a correction in another pull request.

ricardo-salgado-tekever:
Once this one is merged - but not before we dot all the i's and cross all the t's - I have seek capability already working on the playout.

plugins/janus_recordplay.c
```diff
@@ -2280,6 +2318,37 @@ janus_recordplay_frame_packet *janus_recordplay_get_frames(const char *dir, cons
 			offset += len;
 			continue;
 		}
+
+		if(data) {
+			/* Things are simpler for data, no reordering is needed: start by the data time */
+			gint64 when = 0;
+			bytes = fread(&when, sizeof(gint64), 1, file);
+			if(bytes < 1) {
```
lminiero, line 2326: See note on your other PR. I'd rather invert the arguments there, so that we still check the bytes.

plugins/janus_recordplay.c
```diff
@@ -2618,6 +2710,73 @@ static void *janus_recordplay_playout_thread(void *data) {
 				}
 			}
 		}
+		if(data) {
+			if(data == session->dframes) {
+				/* First packet, send now */
```
lminiero, line 2715: I don't think this is correct. The first data frame should not be sent until the timestamp says it should. You can have a recording created at time 0, and the first data message saved 15s later. The mjr parsing for data does indeed take that into account, by using `c_time` as a reference to calculate the relative time of each packet:

    p->ts = when-c_time;

where the first `p->ts` may very well not be `0`.

lminiero: I guess this can be solved by getting rid of the `first` branch of the if, and just stick to the second, where the `ts_diff` calculation uses something else (a monotonic timer variable created as soon as the streaming start, perhaps) instead of `data->prev->ts` for the first frame (where `data->prev` doesn't exist), e.g.

   ts_diff = data->ts - data->prev->ts;   --->   ts_diff = data->ts - start_time;

ricardo-salgado-tekever: Actually, since `data-ts` is calculated as `when-c_time`, it means that the timestamp are all indexed to 0.
So all we need is to use 0 the first time.

lminiero: No, because the timestamp in the packet (those 8 bytes) will be a monotonic time of when that packet was written to the file; `c_time` is the monotonic time of then the mjr file was first created. As such, `when-c_time` is the time that passed between since the recording file was created and when the first packet was written (which will often be > 0).

lminiero: Most of the times, you create a recording and no data is written to it for a while (much more likely for data, audio/video are there almost right away often).

ricardo-salgado-tekever: But in that case, we would need to correlate the `c_time` of the different files - audio, video - and assume that the one that starts first represents 0 time for the data... Is that a good assumption?
From what I can see in the streaming plugin (which is the only one generating data recordings), all files are created and header files written almost simultaneously, and `c_time` will be very close indeed.

plugins/janus_recordplay.c
```diff
@@ -2041,6 +2065,16 @@ void janus_recordplay_update_recordings_list(void) {
...
 		}
+		if(data && data->value) {
+			rec->drc_file = g_strdup(data->value);
+			char *ext = strstr(rec->drc_file, ".mjr");
+			if(ext != NULL)
+				*ext = '\0';
+			const char *textcodec = janus_recordplay_parse_codec(recordings_path,
+				rec->drc_file, NULL, sizeof(NULL), NULL);
+			if(textcodec)
+				rec->textdata = !strcasecmp("text", textcodec);
```
lminiero, line 2076: `rec->textdata` should be statically set to `TRUE` at the moment, since we don't support binary recordings at the moment.

lminiero (commented):
I've added a small note inline: including the missing swap of `sizeof(gint64), 1` addressed in the previous note, I think those are the last parts that need to be fixed before this is ready.

I'll make a couple of tests myself in the meanwhile to see if I spot anything else. Thanks!

plugins/janus_recordplay.c
lminiero: For audio/video it's rarely needed, as they flow continuously, so it can be pretty much safely ignored. Data channel messages are occasional, instead: just think of how they're used in the EchoTest demo, for instance, where audio and video flow all the time, and data only when you type something in.

ricardo-salgado-tekever: The recordings right now are started at the same time, even if data is only written much later.
However, in terms of scalability, it makes sense that all the `c_time` fields are kept at playout time, to make sure they align, for future use of recordings that are not started simultaneously.

So what I propose is for the `c_time` to be extracted out to the playout thread (from `get_frames` ), so that it can be used.

lminiero: Actually I just checked with a recording of the echotest, and timing of the data was 100% correct! I only sent data later, and it matched exactly the moment in audio I said I had just sent the message. There is another problem, though: Janus crashed for me when I tried playing a non-data recording after I had just closed the data one. I'll have a look at the asan traces and will provide info in a comment.

lminiero:
If you play a recording with data, and then play a recording without data, Janus crashes here:

```
==24334==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f5e6d8d1631 bp 0x000000000000 sp 0x7f5e63707428 T14)
==24334==The signal is caused by a READ memory access.
==24334==Hint: address points to the zero page.
    #0 0x7f5e6d8d1631 in __strstr_sse2_unaligned (/lib64/libc.so.6+0xa9631)
    #1 0x7f5e6e6fca30 in strstr (/lib64/libasan.so.6+0x8ba30)
    #2 0x7f5e6969271e in janus_recordplay_playout_thread plugins/janus_recordplay.c:2536
```

which is

```
	if(session->dframes) {
		char source[1024];
		if(strstr(rec->drc_file, ".mjr"))            <----
```

This is very likely because `session->dframes` is not cleared up when the playout ends, and so it's still there for the following playout that doesn't have any data recording. Since `rec->drc_file` is `NULL`, the method crashes.

The cleanup is missing at the end of `janus_recordplay_playout_thread`:

```
	audio = session->aframes;
	while(audio) {
		tmp = audio->next;
		g_free(audio);
		audio = tmp;
	}
	session->aframes = NULL;
	video = session->vframes;
	while(video) {
		tmp = video->next;
		g_free(video);
		video = tmp;
	}
	session->vframes = NULL;
```

This should now include the same for `dframes`.

lminiero:
One more note: you're never freeing `recording->drc_file`, so you're also introducing a memory leak. It should be done in `janus_recordplay_recording_free`, which is where `arc_file` and `vrc_file` are freed too.

ricardo-salgado-tekever:
Yeah, too much time spent on garbage-collected environments. Need to clean-up after myself.

plugins/janus_recordplay.c
ricardo-salgado-tekever: As I've seen on my side. 
I'd much rather leave is as it is - it assumes the recordings are done simultaneously. It's simpler and easy to understand the behaviour if you feed in recordings done at different time.

lminiero:
> Yeah, too much time spent on garbage-collected environments. Need to clean-up after myself.

C is ruthless!! :hand_over_mouth:

plugins/janus_recordplay.c
lminiero: Yep, since it behaves as it should, I agree there's no need to change it :+1: 
It's a first integration anyway: in case issues appear later on, they can be fixed in time.

plugins/janus_recordplay.c
ricardo-salgado-tekever: Done.

plugins/janus_recordplay.c
ricardo-salgado-tekever: Done.

plugins/janus_recordplay.c
ricardo-salgado-tekever: Done.

lminiero:
Thanks for the quick fixes! I don't see anything else that needs to be addressed, so this can be merged :+1:

PR #2471, opened 2020-12-03
Add an option to enable libmicrohttpd error logs

evorw:


conf/janus.transport.http.jcfg.sample
```diff
@@ -19,6 +20,7 @@ general: {
 	#secure_interface = "eth0"		# Whether we should bind this server to a specific interface only
 	#secure_ip = "192.168.0.1"		# Whether we should bind this server to a specific IP address (v4 or v6) only
 	#acl = "127.,192.168.0."		# Only allow requests coming from this comma separated list of addresses
+	#http_lib_debug = false			# Ask libmicrohttpd to write warning and error messages to stderr (default=false)
```
lminiero, line 23: I think the property should be called `mhd_debug`, since it's the library we're debugging.

transports/janus_http.c
```diff
@@ -117,6 +117,7 @@ static janus_transport_callbacks *gateway = NULL;
 static gboolean http_janus_api_enabled = FALSE;
 static gboolean http_admin_api_enabled = FALSE;
 static gboolean notify_events = TRUE;
+static enum MHD_FLAG mhd_debug_flag = MHD_NO_FLAG;
```
lminiero, line 120: Hve you checked if this breaks older versions of libmicrohttpd? We had to do something like this for `enum MHD_result` recently.

conf/janus.transport.http.jcfg.sample
evorw: Done!

transports/janus_http.c
evorw: It was `enum MHD_FLAG` in MHD version 0.9.59, so unless they changed it in an intervening version and then changed it back we should be ok

conf/janus.transport.http.jcfg.sample
```diff
@@ -5,6 +5,7 @@
 # all the web servers will try and bind on both IPv4 and IPv6: if you
 # want to only bind to IPv4 addresses (e.g., because your system does not
 # support IPv6), you should set the web server 'ip' property to '0.0.0.0'.
+# To see debug logs from the HTTP server library, set 'http_lib_debug'.
```
lminiero, line 8: This comment still references the old name.

lminiero:
Thanks for the fixes, merging!

PR #2472, opened 2020-12-03
Corrected janus pp rec

ricardo-salgado-tekever:
Using fread() with a `size` != 1 byte does not return the number of bytes, but the number of elements of `size` that where read.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -671,7 +671,7 @@ int main(int argc, char *argv[])
 		if(data) {
 			/* Things are simpler for data, no reordering is needed: start by the data time */
 			gint64 when = 0;
-			int bytes = fread(&when, 1, sizeof(gint64), file);
+			bytes = fread(&when, 1, sizeof(gint64), file);
 			if(bytes < sizeof(gint64)) {
 				JANUS_LOG(LOG_WARN, "Missing data timestamp header");
 				break;
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -671,8 +671,8 @@ int main(int argc, char *argv[])
 		if(data) {
 			/* Things are simpler for data, no reordering is needed: start by the data time */
 			gint64 when = 0;
-			bytes = fread(&when, sizeof(gint64), 1, file);
-			if(bytes < (int)sizeof(gint64)) {
+			int elements_read = fread(&when, sizeof(gint64), 1, file);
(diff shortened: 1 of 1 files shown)


postprocessing/janus-pp-rec.c
```diff
@@ -671,8 +671,8 @@ int main(int argc, char *argv[])
 		if(data) {
 			/* Things are simpler for data, no reordering is needed: start by the data time */
 			gint64 when = 0;
-			bytes = fread(&when, sizeof(gint64), 1, file);
-			if(bytes < (int)sizeof(gint64)) {
+			int elements_read = fread(&when, sizeof(gint64), 1, file);
+			if(elements_read < 1) {
```
lminiero, line 675: Wouldn't it be easier to just switch `sizeof(gint64)` and `1`? It would make it more consistent with our usage of `fread`, that AFAIK is all based on bytes and not on units.

ricardo-salgado-tekever: I agree.

postprocessing/janus-pp-rec.c
```diff
@@ -671,8 +671,8 @@ int main(int argc, char *argv[])
 		if(data) {
 			/* Things are simpler for data, no reordering is needed: start by the data time */
 			gint64 when = 0;
-			bytes = fread(&when, sizeof(gint64), 1, file);
-			if(bytes < (int)sizeof(gint64)) {
+			int bytes = fread(&when, 1, sizeof(gint64), file);
```
lminiero, line 674: There's actually no need for the `int`, `bytes` is defined at line 341 already.

ricardo-salgado-tekever: Right.

lminiero:
Thanks, merging!

postprocessing/janus-pp-rec.c
```diff
@@ -671,8 +671,8 @@ int main(int argc, char *argv[])
 		if(data) {
 			/* Things are simpler for data, no reordering is needed: start by the data time */
 			gint64 when = 0;
-			bytes = fread(&when, sizeof(gint64), 1, file);
-			if(bytes < (int)sizeof(gint64)) {
+			bytes = fread(&when, 1, sizeof(gint64), file);
+			if(bytes < sizeof(gint64)) {
```
lminiero, line 675: Notice that I've restored the (int) part you removed here, as compilers will complain about the check of different signedness otherwise (`sizeof` returns an unsigned).

PR #2481, opened 2020-12-07
Added support for binary data recordings

lminiero:
This PR expands on the effort made by @ricardo-salgado-tekever in #2468, and adds support for recording binary data channels too, rather than just text data channels. It also fixed a few other things I noticed.

In mjr files, binary recordings are now marked as "binary" (where text-based data channel recordings were already marked as "text"). Since there's no way to know what data is stored in those files and so we can't expect a specific extension for the destination file as we do for other formats, while text based data recordings are still converted to `.srt` files, the postprocessor converts binary recordings to raw files instead: meaning that it will just extract the data from each packet, and append it one after the other in the file name you provide. This should be the easiest and simplest way of handling them, rather than ignoring them altogether.

To test all this, I modified the Record&Play plugin to support saving data recordings as well (Ricardo's patch only added support for playout). To test this, you can add `?data=text` or `?data=binary` as a query string argument to the `recordplaytest.html` demo page, and the demo will show a text area where you can input some text: when you press enter, the message is sent to the plugin and saved to the file in the format you asked for. At the same time, the playout of a recording including data will render the messages in the same box (notice that binary data will simply be rendered as the presence of an ArrayBuffer object).

This seems to be working fine in my setup: if you're using Record&Play in your applications, please test to ensure you don't encounter any regression. I plan to merge soon otherwise.

### DIFF
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -42,7 +42,7 @@
  * 		audio = videoroom-audio.mjr
  * 		video = videoroom-video.mjr
  *
- * Data channel recordings are suppoered via a \c data attribute as well.
+ * Data channel recordings are supported via a \c data attribute as well.
  *
  * \section recplayapi Record&Play API
  *
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -42,6 +42,8 @@
  * 		audio = videoroom-audio.mjr
  * 		video = videoroom-video.mjr
  *
+ * Data channel recordings are suppoered via a \c data attribute as well.
+ *
  * \section recplayapi Record&Play API
  *
  * The Record&Play API supports several requests, some of which are
@@ -83,6 +85,7 @@
 			"date": "<Date of the recording>",
 			"audio": "<Audio rec file, if any; optional>",
 			"video": "<Video rec file, if any; optional>",
+			"data": "<Data rec file, if any; optional>",
 			"audio_codec": "<Audio codec, if any; optional>",
 			"video_codec": "<Video codec, if any; optional>"
 		},
@@ -131,7 +134,8 @@
 	"filename" : "<Base path/name for the file (media type and extension added by the plugin); optional>",
 	"audiocodec" : "<name of the audio codec we prefer for the recording; optional>",
 	"videocodec" : "<name of the video codec we prefer for the recording; optional>",
-	"videoprofile" : "<in case the video codec supports, profile to use (e.g., "2" for VP9, or "42e01f" for H.264); optional>"
+	"videoprofile" : "<in case the video codec supports, profile to use (e.g., "2" for VP9, or "42e01f" for H.264); optional>",
+	"textdata" : "<in case data channels have to be recorded, whether the data will be text (default) or binary; optional>"
 }
 \endverbatim
  *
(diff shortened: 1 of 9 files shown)


plugins/janus_recordplay.c
```diff
@@ -1668,6 +1708,22 @@ static void *janus_recordplay_handler(void *data) {
...
+			data = (janus_sdp_mline_find(offer, JANUS_SDP_APPLICATION) != NULL);
+			if(data) {
+				textdata = dotextdata ? json_is_true(dotextdata) : TRUE;
+				char filename[256];
+				if(filename_text != NULL) {
+					g_snprintf(filename, 256, "%s-data", filename_text);
+				} else {
+					g_snprintf(filename, 256, "rec-%"SCNu64"-data", id);
+				}
+				rec->drc_file = g_strdup(filename);
```
tmatth, line 1721: nit: you could simplify this (and get rid of the intermediate `filename` buffer) with just:
```
if(filename_text != NULL) {
   rec->drc_file = g_strdup_printf("%s-data", filename_text);
} else {
  rec->drc_file = g_strdup_printf("rec-%"SCNu64"-data", id);
}
```

plugins/janus_recordplay.c
```diff
@@ -2073,8 +2130,7 @@ void janus_recordplay_update_recordings_list(void) {
 				*ext = '\0';
 			const char *textcodec = janus_recordplay_parse_codec(recordings_path,
 				rec->drc_file, NULL, sizeof(NULL), NULL);
-			if(textcodec)
-				rec->textdata = TRUE; /* Binary recordings not supported yet */
+			rec->textdata = textcodec && (!strcasecmp(textcodec, "text"));
```
tmatth, line 2133: nit: no need for the extra parentheses around `!strcasecmp()` here

postprocessing/pp-binary.c
```diff
@@ -0,0 +1,92 @@
...
+#endif
+#include <inttypes.h>
+#include <string.h>
+#include <stdlib.h>
+
+#include "pp-binary.h"
+#include "../debug.h"
+
+
+FILE *binary_file = NULL;
```
tmatth, line 28: If requests happen in parallel, won't they clobber this? Maybe put this in some user_data or thread local storage?

plugins/janus_recordplay.c
lminiero: Oh I wasn't aware of `g_strdup_printf`: since we do the same thing in a few other places as well (core and plugins), for the moment I'll keep this as it is, and then once we merge I'll change them all in a single commit. Thanks for the heads up!

plugins/janus_recordplay.c
lminiero: True, I just thought it made it less ambiguous.

postprocessing/pp-binary.c
lminiero: This is part of the postprocessor, not Janus, so there's no requests to handle: each time the postprocessor runs it only handles a single file.

plugins/janus_recordplay.c
```diff
@@ -42,6 +42,8 @@
  * 		audio = videoroom-audio.mjr
  * 		video = videoroom-video.mjr
  *
+ * Data channel recordings are suppoered via a \c data attribute as well.
```
tmatth, line 45: nit: typo on "supported"

lminiero: Fixed :+1:

lminiero:
Merging.

Issue #2484, opened 2020-12-09
Allow MQTT SSL to work with default OpenSSL CA cert, not just explicit


cameronelliott:
This issue is regarding PR #2485 
(should I have just created a PR alone or is a PR+issue the preferred manner?)

In order to use TLS there should be no requirement to specify a CAcert. The normal trust store of the system should be used - but I haven't tested if the library actually use this. This requires some investigation.

As I reflected in the other PR, all versions of the SSL protocol are insecure. Even though our library use "ssl" as a prefix, I think we should promote the prefix "TLS" for all settings related to actually using TLS - since we never support SSL. It's a step forward and require some code, but a "good thing".

Also, the MQTT scheme for TLS enabled MQTT seems to be "mqtts:" in most implementations, even though the paho library for some odd reason use "ssl:".

Will try to compile and test, since I need to steal this code for my MQTT event handler implementation. ;-)

_Originally posted by @oej in https://github.com/meetecho/janus-gateway/issues/1064#issuecomment-347122912_

cameronelliott:
Because I wanted to make sure the janus conversation to AWS was encrypted I used tcpdump to capture the conversation.
It looks like it is encryped with TLS.
I have attached the pcap (inside a zip for anyone that would like to verify)
[janus.mqtt.pcap.zip](https://github.com/meetecho/janus-gateway/files/5665886/janus.mqtt.pcap.zip)

cameronelliott:
I think this can be closed, the PR has been merged.

PR #2485, opened 2020-12-09
Do not require a cacertfile, pass null to openssl

cameronelliott:
This will allow OpenSSL to choose where to find the CACert, or if to use one at all.
See issue #2484 
This can be super helpful for Docker deploys where you don't want to build your own CACert files, and you want OpenSSL to choose how to validate the connection.

### DIFF
--- a/transports/janus_mqtt.c
+++ b/transports/janus_mqtt.c
@@ -377,12 +377,12 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
 
 		ctx->ssl_enabled = TRUE;
 
-		janus_config_item *cacertfile = janus_config_get(config, config_general, janus_config_type_item, "cacertfile");
+		janus_config_item *cacertfile = janus_config_get(config, config_general, janus_config_type_item, "cacertfile");		
 		if(!cacertfile || !cacertfile->value) {
-			JANUS_LOG(LOG_FATAL, "Missing CA certificate for MQTT integration...\n");
-			goto error;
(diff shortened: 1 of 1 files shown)


transports/janus_mqtt.c
```diff
@@ -377,12 +377,12 @@ int janus_mqtt_init(janus_transport_callbacks *callback, const char *config_path
 
 		ctx->ssl_enabled = TRUE;
 
-		janus_config_item *cacertfile = janus_config_get(config, config_general, janus_config_type_item, "cacertfile");
+		janus_config_item *cacertfile = janus_config_get(config, config_general, janus_config_type_item, "cacertfile");
```
lminiero, line 380: Looks like there's an extra tab at the end of this line that should be removed.

lminiero (commented):
Thanks for this patch! I just found a small typo that shouldn't be there, but apart from that it looks fine to me. I'll merge as soon as the typo is fixed.

transports/janus_mqtt.c
cameronelliott: Does that look better now?
Do I need to squash/compress the two commits somehow?

lminiero: No need, I can squash them when merging, thanks!

lminiero:
Merging :+1:

transports/janus_mqtt.c
cameronelliott: Awesome! Thanks for being gentle about my little goofs.

PR #2486, opened 2020-12-09
Replace Travis CI with GitHub Actions.

atoppi:


.github/workflows/janus-ci.yml
```diff
@@ -0,0 +1,135 @@
...
+        make -j$(nproc)
+        sudo make install
+        popd
+    - name: setup lws from sources
+      run: |
+        git clone --depth 1 --quiet -b v4.1-stable https://github.com/warmcat/libwebsockets.git libwebsockets
+        pushd libwebsockets
+        mkdir -p build
+        pushd build
+        cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr -DLWS_WITH_STATIC=OFF -DLWS_WITHOUT_CLIENT=ON -DLWS_WITHOUT_TESTAPPS=ON -DLWS_WITHOUT_TEST_SERVER=ON -DLWS_WITH_HTTP2=OFF ..
```
tmatth, line 103: Would it make sense to build with `-DLWS_MAX_SMP=1` as per the issue linked from the README?

atoppi: Actually `-DLWS_MAX_SMP=1` is the default in recent lws (since 2018 afaik) and I'm checking out a fresh version here.

tmatth: Oh good to know.

atoppi:
If someone has some experience on GitHub Actions I'd be happy to hear some suggestion about how reducing the workflow time.
2-3 minutes per job is imho a bit high, and might quickly exhaust the free quota offered by github.

tmatth:
> If someone has some experience with GitHub Actions I'd be happy to receive some suggestion about reducing the workflow time.
> 2-3 minutes per job is imho a bit high, and might quickly exhaust the free quota offered by github.

The main thing that jumps out at me would be to avoid building dependencies from source (or finding a way of caching these at least). You could push a base image that includes those and use that instead of the stock ubuntu-latest?

atoppi:
@tmatth using e.g. a Docker image to contain all the needed dependencies is a possible optimization (and that was an option also with travis). Anyway freezing dependencies (either apt or sources) to specific versions has a couple of downsides:
1. you must keep track of the versions and update them outside of the CI system (e.g. regularly maintain your image in a docker registry)
2. your CI system is unable to catch building errors coming from broken commits in deps where we use the `master` branch (e.g. usrsctp, where sometimes it happens that a commit completely breaks the build). We'd like to preserve this feature.

What I'll be willing to have is a way to re-use layers (like apt dependencies and sources) within a single batch of jobs.
Example:
- we push a new commit
- the CI starts the concurrent building of the job matrix
- every step about the dependencies will be probably the same in all the jobs (installing packages, building libnice, lws and so on)
- I'd like to reuse the layers of these dependencies for all the active jobs (and not also for future builds).

Using github actions native cache system looks pretty over-complex and "hacky" in our case.

atoppi:
I have added some optimizations to the workflow, basically using apt dependencies in place of sources when executing a "partia"l job (w/o libcurl or w/o datachannels).
I think this PR is now ready to be merged.
Waiting for @lminiero pushing the button :-)

lminiero:
> Waiting for @lminiero to push the button :-)

Pushing! :point_right: :radio_button:

PR #2488, opened 2020-12-11
Playout with seek (again)

ricardo-salgado-tekever:
(Submitted again since the previous PR was inadvertently closed without merging). 
This pull request adds the capability to seek to a particular timestamp in the playout.
The seek capability works by:

changing the `play` message, which supports an optional `start_at` property for the timestamp to start the video at;
adding the seek message, which support a single mandatory `timestamp` property, for the timestamp to which the video should jump to.
For both cases, timestamp is expressed in microsseconds since the start of the recording files. After seeking, all timestamps and sequence numbers are recalculated, so the webRTC client does not perceive the seek as frame repetitions or jumps.

The sample demo was not changed, so to test just use this command to seek to 5.0sec of video on the javascript console:
`recordplay.send({ message : { "request" : "seek", "timestamp" : 5000000 }}, (response) => { console.log(response);})`


lminiero:
Closing as Janus has changed considerably since this was first opened, and there didn't seem to be much interest in the feature. We can consider revisiting this again in the future. Thanks for your contribution anyway!

PR #2489, opened 2020-12-11
Set libmicrohttpd connections limit in http transport configuration.

atoppi:
This PR lets the user define the maximum number of open connection in libmicrohttpd through the `mhd_connection_limit` setting in the http transport configuration.

As explained [in the official docs](https://www.gnu.org/software/libmicrohttpd/manual/libmicrohttpd.html).

> MHD_OPTION_CONNECTION_LIMIT
>Maximum number of concurrent connections to accept (followed by an unsigned int). The default is FD_SETSIZE - 4

(`FD_SETSIZE` is `1024` in most environments)
This should be useful for deployments that make intensive use of http transport and could hit that limit. 
Default value has been preserved to `FD_SETSIZE - 4`.


lminiero:
Merging :+1:

PR #2492, opened 2020-12-14
Close nice agent resources asynchronously

fbellet:
This merge request uses the function nice_agent_close_async() to free the remote reservations that the nice agent may have on remote turn servers, before being disposed. This typically adds a delay of one or two seconds  before disposing the agent (one stun request, and two retransmissions, since these stun requests are not acked).

I tested with both static event loops and per-thread event loop.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1377,6 +1377,19 @@ static void janus_ice_handle_free(const janus_refcount *handle_ref) {
 	g_free(handle);
 }
 
+#ifdef HAVE_CLOSE_ASYNC
+static void janus_ice_cb_agent_closed(GObject *src, GAsyncResult *result, gpointer data) {
+	janus_ice_outgoing_traffic *t = (janus_ice_outgoing_traffic *)data;
+	janus_ice_handle *handle = t->handle;
+
+	JANUS_LOG(LOG_VERB, "[%"SCNu64"] Disposing nice agent %p\n", handle->handle_id, handle->agent);
+	g_object_unref(handle->agent);
+	handle->agent = NULL;
+	g_source_unref((GSource *)t);
+	janus_refcount_decrease(&handle->ref);
+}
+#endif
+
 static void janus_ice_plugin_session_free(const janus_refcount *app_handle_ref) {
 	janus_plugin_session *app_handle = janus_refcount_containerof(app_handle_ref, janus_plugin_session, ref);
 	/* This app handle can be destroyed, free all the resources */
@@ -1436,9 +1449,22 @@ static void janus_ice_webrtc_free(janus_ice_handle *handle) {
 		handle->stream = NULL;
 	}
 	if(handle->agent != NULL) {
+#ifdef HAVE_CLOSE_ASYNC
+		if(G_IS_OBJECT(handle->agent)) {
+			JANUS_LOG(LOG_VERB, "[%"SCNu64"] Removing stream %d from agent %p\n",
+				handle->handle_id, handle->stream_id, handle->agent);
+			nice_agent_remove_stream(handle->agent, handle->stream_id);
+			handle->stream_id = 0;
+			JANUS_LOG(LOG_VERB, "[%"SCNu64"] Closing nice agent %p\n", handle->handle_id, handle->agent);
+			janus_refcount_increase(&handle->ref);
+			g_source_ref(handle->rtp_source);
+			nice_agent_close_async(handle->agent, janus_ice_cb_agent_closed, handle->rtp_source);
+		}
+#else
 		if(G_IS_OBJECT(handle->agent))
 			g_object_unref(handle->agent);
 		handle->agent = NULL;
+#endif
 	}
 	if(handle->pending_trickles) {
 		while(handle->pending_trickles) {
--- a/configure.ac
+++ b/configure.ac
@@ -395,6 +395,12 @@ AC_CHECK_LIB([nice],
              [AC_MSG_NOTICE([libnice version does not support TCP candidates])]
              )
 
+AC_CHECK_LIB([nice],
+             [nice_agent_close_async],
+             [AC_DEFINE(HAVE_CLOSE_ASYNC)],
+             [AC_MSG_NOTICE([libnice version does not have nice_agent_close_async])]
+             )
+
 AC_CHECK_LIB([dl],
              [dlopen],
              [JANUS_MANUAL_LIBS="${JANUS_MANUAL_LIBS} -ldl"],


januscla:
Thanks for your contribution, @fbellet! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks @fbellet! I'll review tomorrow, pinging @atoppi so that he can check the changes too.

ice.c
```diff
@@ -1431,14 +1447,25 @@ static void janus_ice_webrtc_free(janus_ice_handle *handle) {
...
 	if(handle->stream != NULL) {
 		janus_ice_stream_destroy(handle->stream);
 		handle->stream = NULL;
 	}
 	if(handle->agent != NULL) {
+#ifdef HAVE_CLOSE_ASYNC
+		if(G_IS_OBJECT(handle->agent)) {
+			JANUS_LOG(LOG_INFO, "[%"SCNu64"] Closing nice agent %p\n", handle->handle_id, handle->agent);
+			nice_agent_close_async(handle->agent, janus_ice_cb_agent_closed, handle);
+			janus_refcount_increase(&handle->ref);
```
tmatth, line 1462: Would it make sense to increase the refcount before the `nice_agent_close_async` call (to avoid a race), or are we guaranteed that `janus_ice_cb_agent_closed` will not be called before the end of this block (and thus, there is no race)?

lminiero: The handle is already protected by a mutex.

ice.c
```diff
@@ -1431,14 +1447,25 @@ static void janus_ice_webrtc_free(janus_ice_handle *handle) {
 		return;
 	}
 	handle->agent_created = 0;
+#ifdef HAVE_CLOSE_ASYNC
+	nice_agent_attach_recv(handle->agent, handle->stream_id, 1, g_main_loop_get_context(handle->mainloop), NULL, NULL);
+#endif
```
lminiero, line 1452: We actually stopped doing this in #2101, because of [this issue](https://gitlab.freedesktop.org/libnice/libnice/issues/14) in libnice that is apparently still open. As such, I'd rather not reintroduce it here, unless you know it can't cause the issue presented above.

lminiero (commented):
@fbellet I just checked the code and I think there are some wrong assumptions. You apparently exit the loop when the agent is closed, which should not happen: the loop is created when the handle is attached, and is only stopped when the handle itself is detached/destroyed. The confusion may derive from the fact that the source handling the loop is (pobably improperly) called `janus_ice_outgoing_traffic`, but it doesn't handle just that: it actually handles other states as well.

As such, the way the patch works right now, unless static event loops are used the handle would be rendered useless as soon as the first PeerConnection created on it is destroyed, since the loop would disappear (and so would the handle thread).

This also means that preventing `janus_ice_outgoing_traffic_finalize` from closing loop is also probably incorrect. I realize it's done to avoid the last `janus_ice_cb_agent_closed` from being called, though (if we're closing the agent because the whole handle was detached, for instance), so there may need to be a more complex logic there.

I also added an inline note on `nice_agent_attach_recv`.

atoppi (commented):
I agree with @lminiero.
The `janus_ice_cb_agent_closed` callback should not be in charge of quitting the loop because an agent lifecycle is always shorther than an handle lifecycle.
Agents may be destroyed and then recreated any time a new ICE local setup starts in `janus_ice_setup_local`.

fbellet:
Thank you for the review. I'll update my merge request to take your comments into account:

- by adding a ref on `handle->rtp_source`  to prevent the GSource from calling its finalize callback `janus_ice_outgoing_traffic_finalize`  (and from quitting the event loop) before the nice agent is ready to be closed.
- by removing the agent stream instead of setting a null callback in `nice_agent_attach_recv`

atoppi (commented):
Now the PR looks good to me :+1: 
I've made some tests with janus local turn candidates and it seems to work as expected.

ice.c
```diff
@@ -1377,6 +1377,19 @@ static void janus_ice_handle_free(const janus_refcount *handle_ref) {
 	g_free(handle);
 }
 
+#ifdef HAVE_CLOSE_ASYNC
+static void janus_ice_cb_agent_closed(GObject *src, GAsyncResult *result, gpointer data) {
+	janus_ice_outgoing_traffic *t = (janus_ice_outgoing_traffic *)data;
+	janus_ice_handle *handle = t->handle;
+
+	JANUS_LOG(LOG_INFO, "[%"SCNu64"] Disposing nice agent %p\n", handle->handle_id, handle->agent);
```
lminiero, line 1385: This should probably be a `LOG_VERB`, rather than `LOG_INFO`.

ice.c
```diff
@@ -1436,9 +1449,22 @@ static void janus_ice_webrtc_free(janus_ice_handle *handle) {
 		handle->stream = NULL;
 	}
 	if(handle->agent != NULL) {
+#ifdef HAVE_CLOSE_ASYNC
+		if(G_IS_OBJECT(handle->agent)) {
+			JANUS_LOG(LOG_INFO, "[%"SCNu64"] Removing stream %d from agent %p\n",
+				handle->handle_id, handle->stream_id, handle->agent);
```
lminiero, line 1455: Same here, we don't really need all that info displayed by default, while it's useful in verbose mode.

ice.c
```diff
@@ -1436,9 +1449,22 @@ static void janus_ice_webrtc_free(janus_ice_handle *handle) {
...
+			nice_agent_remove_stream(handle->agent, handle->stream_id);
+			handle->stream_id = 0;
+			JANUS_LOG(LOG_INFO, "[%"SCNu64"] Closing nice agent %p\n", handle->handle_id, handle->agent);
```
lminiero, line 1458: Same here.

lminiero (commented):
Apologies if this took a while, I've only recently come back to work after the holidays. I agree the patch looks fine (haven't tested it yet, I will later today), I just added a couple of notes on the possibly excessive verbosity of the new log lines.

ice.c
fbellet: OK, fixed.

lminiero:
Thanks for the fixes and once more for the contribution, merging! :+1:

tgabi333:
is it a performance related PR?

tmatth:
> is it a performance related PR?

AFAICT the motivation here is to not leave TURN allocations lingering on the remote TURN server, see:
https://libnice.freedesktop.org/libnice/NiceAgent.html#nice-agent-close-async

PR #2494, opened 2020-12-14
Add hold parameter media_direction

desarrollo1Calltech:
add hold parameter: media_direction

Added optional parameter "media_direction" , with the different possibilities such as:
media_direction = sendonly
media_direction = inactive
{
         "message": {
             "request": "hold",
             "media_direction": "sendonly"
         }
     };

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -4098,20 +4098,31 @@ static void *janus_sip_handler(void *data) {
 				/* To put the call on-hold, we need to set the direction to recvonly:
 				 * resuming it means resuming the direction we had before */
 				session->media.on_hold = hold;
+				json_t *media_direction = json_object_get(root, "media_direction");
+				const char *media_direction_text = json_string_value(media_direction);
 				janus_sdp_mline *m = janus_sdp_mline_find(session->sdp, JANUS_SDP_AUDIO);
 				if(m) {
 					if(hold) {
 						/* Take note of the original media direction */
 						session->media.pre_hold_audio_dir = m->direction;
 						/* Update the media direction */
-						switch(m->direction) {
-							case JANUS_SDP_DEFAULT:
-							case JANUS_SDP_SENDRECV:
+						if(media_direction_text){
+							if(!strcasecmp(media_direction_text, "sendonly")){
 								m->direction = JANUS_SDP_SENDONLY;
-								break;
-							default:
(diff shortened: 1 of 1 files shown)


plugins/janus_sip.c
```diff
@@ -4098,20 +4098,30 @@ static void *janus_sip_handler(void *data) {
 				/* To put the call on-hold, we need to set the direction to recvonly:
 				 * resuming it means resuming the direction we had before */
 				session->media.on_hold = hold;
+				json_t *media_direction = json_object_get(root, "media_direction");
+				const char *media_direction_text = json_string_value(media_direction);
```
lminiero, line 4102: There should be a new validation parameter here, as other methods to via `JANUS_VALIDATE_JSON_OBJECT`. As it is now, this parameter is never validated anywhere, and is not guaranteed to either exist or be a string at all.

plugins/janus_sip.c
```diff
@@ -4123,14 +4133,22 @@ static void *janus_sip_handler(void *data) {
...
+						} else {
+							switch(m->direction) {
+								case JANUS_SDP_DEFAULT:
+								case JANUS_SDP_SENDRECV:
+									m->direction = JANUS_SDP_SENDONLY;
+									break;
+								default:
+									m->direction = JANUS_SDP_INACTIVE;
+									break;
+							}
```
lminiero, line 4151: Both these changes are overkill, IMHO. They could be left exactly as they are, where only `JANUS_SDP_SENDONLY` is parametrized as a variable. There's no need to do a `strcasecmp` on the argument: there's a method in the SDP utils, called `janus_sdp_parse_mtype`, that can parse it to a `janus_sdp_mtype`, meaning it can be used directly there with minimal changes.

lminiero (commented):
Thanks! But the changes can't be accepted as they are, I think. I added some notes inline.

lminiero:
Any update on my notes? If you don't plan to address them anytime soon I'm afraid I'll have to close the PR.

lminiero:
Closing. We'll add this to the TODO list, since it might be a useful feature.

lminiero:
@desarrollo1Calltech please see the new PR above and let me know if it works for you.

JulianOcampo:
ok @lminiero,  Let me configure this version and I will give you feedback on the tests performed

PR #2502, opened 2020-12-21
devicetest: unused var removed

RSATom:
It looks like someone just forgot remove it on refactoring...

### DIFF
--- a/html/devicetest.js
+++ b/html/devicetest.js
@@ -52,7 +52,6 @@ var janus = null;
 var echotest = null;
 var opaqueId = "devicetest-"+Janus.randomString(12);
 
-var firstTime = true;
 var bitrateTimer = null;
 var spinner = null;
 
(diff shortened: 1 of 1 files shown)


lminiero:
Why do you think it's unused? We use to only perform that action the very first time.

RSATom:
It doesn't matter true it or false `restartCapture()` will be called...

lminiero:
Ah I get what you mean, now: we do a `restartCapture` anyway.

lminiero:
Looks like it's because of 1cae99dd23644695cdac41d6671f3a1093a7be27, which is where we changed the behaviour of the demo to renegotiate rather than hangup and create the session again. Previously, if `firstTime` was `false` we would hangup and then recreate the PR a couple of seconds later. I agree it's not needed anymore.

It's still a bit silly to renegotiate, though: it would make more sense to use `replaceTrack` instead (which we don't support in `janus.js`, though, and so would need to be done manually).

RSATom:
Hm... I've found this when was investigating issue with change source cam with [iosrtc](https://github.com/cordova-rtc/cordova-plugin-iosrtc) on iOS device (it just fails). Maybe it's really related to missing `replaceTrack` in janus.js - I'll investigate. Thank you for good point.

lminiero:
To clarify, `janus.js` will use `replaceTrack` internally, but it will renegotiate as well: there's currently no way to only replace the track and avoid the renegotiation.

PR #2513, opened 2021-01-12
Moderator based muting/unmuting of VideoRoom streams

lminiero:
As the title suggests, this PR adds a new request to the VideoRoom, that allows a moderator (that is, someone aware of the room secret) to moderate participants in a room. By "moderate" I mean decide if they can contribute a medium to the room or not, meaning this can be used, for instance, to forcibly mute a participant: a participant that has been moderated will have packets from the muted stream simply dropped, just as if they disabled the stream themselves. Of course, notice that this feature cannot be used to forcibly _unmute_ someone: removing moderation blocks means the participant can now send that stream, which doesn't necessarily mean they will (e.g., they may be self muted).

The new request, called "moderate", is synchronous, and has a syntax like that:

```
{
	"request" : "moderate",
	"secret" : "<room secret, mandatory if configured>",
	"id" : <unique numeric ID of the participant to moderate>,
	"mute_audio" : <true|false, depending on whether or not audio should be muted by the moderator>,
	"mute_video" : <true|false, depending on whether or not video should be muted by the moderator>,
	"mute_data" : <true|false, depending on whether or not data should be muted by the moderator>,
}
```

As such, sending a "moderate" with `mute_audio: true`, for instance, would mute the participant, while an `mute_audio: false` would remove the constraint; you can definitely pass more properties in a single request, even conflicting ones (e.g., `mute_audio: true, mute_video: false` will mute audio but allow video). Any time a participant is moderated, other participants are notified.

Here's a simple, practical example using a curl snippet:

```
curl -d '{"janus": "message_plugin", "transaction": "123", "admin_secret": "janusoverlord", "plugin": "janus.plugin.videoroom", "request": { "request": "moderate", "room": 1234, "secret": "adminpwd", "id": 5416439326112264, "mute_audio": true, "mute_video": true }}' http://localhost:7088/admin
```

In this snippet we're muting both audio and video of the selected participant, which will send this event to all participants in the room (the moderated one included):

```
{ "videoroom": "event", "room": 1234, id: 5416439326112264, "audio-moderation": "muted", "video-moderation": "muted" }
```

Tested briefly and it seems to be working as expected. Please review and provide feedback, so that this can be merged soon.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -103,8 +103,9 @@ room-<unique room ID>: {
  * (invalid JSON, invalid request) which will always result in a
  * synchronous error response even for asynchronous requests.
  *
- * \c create , \c destroy , \c edit , \c exists, \c list, \c allowed, \c kick
- * and \c listparticipants are synchronous requests, which means you'll
+ * \c create , \c destroy , \c edit , \c exists, \c list, \c allowed,
+ * \c kick , \c moderate , \c enable_recording , \listparticipants
+ * and \c listforwarders are synchronous requests, which means you'll
  * get a response directly within the context of the transaction.
  * \c create allows you to create a new video room dynamically, as an
  * alternative to using the configuration file; \c edit allows you to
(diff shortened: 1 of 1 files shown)


lminiero:
As pointed out on the [group](https://groups.google.com/g/meetecho-janus/c/e9jgVVM1uhI/m/GdjH0ZoZDQAJ), the current syntax for moderating can be a bit ambiguous: as such, `moderate` now expects `mute_audio`/`mute_video`/`mute_data` booleans, rather than `audio`/`video`/`data` as proposed initially. I'll update the intro text accordingly.

lminiero:
Merging.

PR #2518, opened 2021-01-19
Feature/enhance typings

bstuff:
This PR adds `muteAudio` handles typings. Also we can use IAdapter type from webrtc-adapter. Maybe at this point it will be better not to add it to dependencies, but just grab this interface from 7.4.0.
@lminiero do you plan to make janus npm installable? I can help.

### DIFF
--- a/npm/janus.d.ts
+++ b/npm/janus.d.ts
@@ -72,11 +72,7 @@ declare namespace JanusJS {
 		consentDialog?: (on: boolean) => void;
 		webrtcState?: (isConnected: boolean) => void;
 		iceState?: (state: 'connected' | 'failed') => void;
-		mediaState?: (
-			medium: 'audio' | 'video',
-			receiving: boolean,
-			mid?: number
-		) => void;
+		mediaState?: (medium: 'audio' | 'video', receiving: boolean, mid?: number) => void;
 		slowLink?: (state: { uplink: boolean }) => void;
 		onmessage?: (message: Message, jsep?: JSEP) => void;
 		onlocalstream?: (stream: MediaStream) => void;
@@ -143,9 +139,7 @@ declare namespace JanusJS {
 	}
 
 	class Janus {
-		static useDefaultDependencies(
-			deps: Partial<Dependencies>
-		): DependenciesResult;
+		static useDefaultDependencies(deps: Partial<Dependencies>): DependenciesResult;
 		static useOldDependencies(deps: Partial<Dependencies>): DependenciesResult;
 		static init(options: InitOptions): void;
 		static isWebrtcSupported(): boolean;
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @bstuff! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

npm/janus.d.ts
```diff
@@ -20,7 +20,7 @@ declare namespace JanusJS {
 		Debug = 'debug',
 		Log = 'log',
 		Warning = 'warn',
-		Error = 'error'
+		Error = 'error',
```
lminiero, line 23: Why did you add a comma here? It's the last item in the list.

npm/janus.d.ts
```diff
@@ -52,7 +52,7 @@ declare namespace JanusJS {
 		Stopped = 'stopped',
 		SlowLink = 'slow_link',
 		Preparing = 'preparing',
-		Refreshing = 'refreshing'
+		Refreshing = 'refreshing',
```
lminiero, line 55: See above.

npm/package.json
```diff
@@ -11,6 +11,7 @@
 	},
 	"devDependencies": {
 		"rollup": "^0.50.0",
-		"rollup-plugin-replace": "^2.0.0"
+		"rollup-plugin-replace": "^2.0.0",
+		"webrtc-adapter": "^7.7.0"
```
lminiero, line 15: While 7.7.0 is indeed the latest version, the demos still only ship 7.4.0, as that's the latest you can find on cdnjs. Not sure if this is meant for server-side usage of janus.js, or for embedding in other frameworks, but maybe it would be best to stick to 7.4.0 here too? (unless this dependency downloads the right version independently of what the web page does).

lminiero (commented):
Thanks for your contribution! Added a few notes inline.
Next time please do provide a description of what the PR is trying to do, though, and why changes are needed.

npm/janus.d.ts
bstuff: `prettier` did it for me( sorry, I'll fix

npm/janus.d.ts
```diff
@@ -127,6 +127,9 @@ declare namespace JanusJS {
 		handleRemoteJsep(params: { jsep: JSEP }): void;
 		dtmf(params: any): void;
 		data(params: any): void;
+		isAudioMuted(): boolean;
+		muteAudio(): void;
+		unmuteAudio(): void;
```
bstuff, line 132: This is the reason why this pr was created. And I also noticed that adapter has `any` type and I thought it's ok to get actual types from webrtc-adapter.

lminiero:
> @lminiero do you plan to make janus npm installable? I can help.

No, we don't. To be honest, the whole npm folder was accepted as a contribution as there were those who needed an easier way of using `janus.js` in other context, but as you can see we don't pay much attention to it (I never use it, and so it can indeed break or lag behind when there's relevant changes). Any additional effort to make it more complex would mean we'd have to maintain more things here, and honestly we don't want to (I only want to focus on the server side).

About your PR, as soon as you address the notes (extra commas, and whether 7.7.0 or 7.4.0 should be used) I'll merge :+1:

lminiero:
Thanks, merging!

PR #2515, opened 2021-01-13
html: update webrtc-adapter to 7.7.0

tmatth:
N.B. This project is in maintenance mode.

PR #2519, opened 2021-01-19
Update Webpack instruction after webrtc-adapter dependency update

ycherniavskyi:
webrtc-adapter version 7+ (#2515) use ES2015 module, so "For importing the default export of an ES2015 module, you have to specify the default property of module." ([example](https://webpack.js.org/plugins/provide-plugin/#usage-vuejs)).

Without this instead of `Janus.webRTCAdapter.browserDetails` we got `Janus.webRTCAdapter.default.browserDetails` and as result `janus.js` is broken.

### DIFF
--- a/mainpage.dox
+++ b/mainpage.dox
@@ -885,7 +885,7 @@ echotest.createOffer({
      // janus.js does not use 'import' to access to the functionality of webrtc-adapter,
      // instead it expects a global object called 'adapter' for that.
      // Let's make that object available.
-     new webpack.ProvidePlugin({ adapter: 'webrtc-adapter' })
+     new webpack.ProvidePlugin({ adapter: ['webrtc-adapter', 'default'] })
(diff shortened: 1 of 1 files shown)


lminiero:
Is there anything in #2518 that warrants a similar fix?

ycherniavskyi:
Hmm as for me they are unrelated.

As I see currently Janus has two node modules: in root and in `npm` folder.

The first one simply exposes `html/janus.js` and declares some dependency but to use it with Webpack builder [some preparation must be provided](https://janus.conf.meetecho.com/docs/js-modules.html#js-webpack).

The second one builds the module with rollup and even provides DTS file. As I understand it doesn't need any special preparation to use. @bstuff am I right?

lminiero:
Ack, I guess it's safe to merge them both then. Thanks!

PR #2525, opened 2021-01-21
Configurable media direction when putting calls on-hold (SIP plugin)

lminiero:
This PR revisits the effort originally contributed in #2494, and basically allows you to specify which media direction should be set in the SDP when putting a call on-hold in the SIP plugin. Originally, in fact, the SIP plugin would always set `sendonly` by default, but @desarrollo1Calltech explained this may not always be desirable, hence a requirement to make this configurable instead.

To do so, the `hold` request now supports a new `direction` attribute too, which can be either `sendonly` (the default if nothing is provided), `recvonly` or `inactive`; the `unhold` request works exactly as before, restoring the media direction that was in the SDP before putting the call on-hold.

I tested this briefly and it seems to be working for me. Please do test as well, especially if you were using the feature before, so that we can merge this soon.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -378,6 +378,34 @@
  * When a session has been established, there are different requests that
  * you can use to interact with the session.
  *
+ * First of all, you can put a call on-hold with the \c hold request.
+ * By default, this request will send a new INVITE to the peer with a
+ * \c sendonly direction for media, but in case you want to set a
+ * different direction (\c recvonly or \c inactive ) you can do that by
+ * passing a \c direction attribute as well:
+ *
+\verbatim
+{
+	"request" : "hold",
+	"direction" : "<sendonly, recvonly or inactive>"
+}
+\endverbatim
+ *
+ * No WebRTC renegotiation will be involved here on the holder side, as
+ * this will only trigger a re-INVITE on the SIP side. To remove the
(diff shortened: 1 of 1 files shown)


JulianOcampo:
Hello Lorenzo,
thanks for your help,

Once the parameter is sent empty (direction = ""), the call is hung up and the answer is:

**Invalid direction (can only be sendonly, recvonly or inactive)**

Should janus hang up the call?, or should it ignore it? ,or should it set the default parameter direction = sendonly?

desarrollo1Calltech:
Hello Lorenzo,
thanks for your help,

Once the parameter is sent empty (direction = ""), the call is hung up and the answer is:

**Invalid direction (can only be sendonly, recvonly or inactive)**

Should janus hang up the call?, or should it ignore hold request? ,or should it set the default parameter direction = sendonly?

lminiero:
@desarrollo1Calltech that's the demo hanging up the call, because it's configured to do that any time it receives an error (any error) back. The plugin will NOT hang up the call in that case. The error is caused by the fact that an empty direction string is a mistake: if you want the default direction, don't set anything at all. If you add the property, it must be valid.

desarrollo1Calltech:
@lminiero  I get it,
keeping that in mind, sending the "direction" parameter works as expected.

lminiero:
Ack, thanks for the feedback, I'll merge then! :+1:

PR #2527, opened 2021-01-22
Allow marking of RTP extensions in MJR recordings

lminiero:
This is a functional rewrite of #2422 which was initially submitted by @isnumanagic. It's hopefully a simpler integration and implementation than that one was, and should provide the same functionality. As the original PR, I've only modified the Record&Play plugin to support it, and as you'll see from the code the scope is much narrower, since it explicitly only checks for audio-level and video-orientation, ignoring other extensions: this means it's probably less flexible than the original PR, but also much simpler. I've done this on purpose as those are the only two extensions we care about in most plugins anyway. The MJR code itself is flexible, though: if you want to store a different extension from your plugin, you can.

I've reused the same syntax for how extensions are stored in the header, using an `x` object where IDs are the key and the extension namespace the value. This is an example of the post-processor analysing an audio recording which had the audio-level extension negotiated:

```
[lminiero@lminiero mjr-extmaps]$ ./janus-pp-rec -j ~/Work/code/test/mjrextmap/share/janus/recordings/rec-5815531892194331-audio.mjr 
{"t": "a", "c": "opus", "x": {"1": "urn:ietf:params:rtp-hdrext:ssrc-audio-level"}, "s": 1611333015010041, "u": 1611333015071425}
```

```
[lminiero@lminiero mjr-extmaps]$ ./janus-pp-rec -p ~/Work/code/test/mjrextmap/share/janus/recordings/rec-5815531892194331-audio.mjr 
Janus version: 110 (0.10.10)
Janus commit: 46a6c71b0fb7e54cb9acfbec84391c75eb224c13
Compiled on:  Fri 22 Jan 17:41:03 CET 2021

Logging level: 4
RTP silence suppression distance: 100

Source file: /home/lminiero/Work/code/test/mjrextmap/share/janus/recordings/rec-5815531892194331-audio.mjr
  -- Parsing header only

File is 104006 bytes
Pre-parsing file to generate ordered index...
Audio level extension ID: 1
This is an audio recording:
  -- Codec:   opus
  -- Created: 1611333015010041
  -- Written: 1611333015071425
SSRC detected: 632194773
Counted 909 RTP packets
Counted 909 frame packets
Parsing and reordering completed, bye!
```

As you can see, the audio-level extension is automatically recognized, and so there's no need to manually pass it via command line if it's listed in the MJR itself already (and so will be used if necessary).

Tested briefly and it seems to be working, but feedback is welcome of course.

### DIFF
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -411,6 +411,8 @@ typedef struct janus_recordplay_recording {
 	janus_videocodec vcodec;	/* Codec used for video, if available */
 	char *vfmtp;				/* Video fmtp, if any */
 	int video_pt;				/* Payload types to use for audio when playing recordings */
+	guint8 audiolevel_ext_id;	/* Audio level extmap ID */
+	guint8 videoorient_ext_id;	/* Video orientation extmap ID */
 	char *drc_file;				/* Data file name */
 	gboolean textdata;			/* Whether data format is text */
 	char *offer;				/* The SDP offer that will be sent to watchers */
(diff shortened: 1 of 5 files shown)


isnumanagic (approved):
LGTM. Thanks for the refactor @lminiero!

lminiero:
@isnumanagic thanks for reviewing! Did you have a chance to also test this, in case you had a specific requirement for it? The sooner I know it works for more people than just me, the sooner I'll be able to merge.

lminiero:
Merging.

PR #2531, opened 2021-01-23
set webrtc-adapter verstion to 7.4.0

augustblack:
After a clean npm pkg rebuild, 7.4.0 version of webrtc-adapter does appear to play well with janus.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,6 @@
   },
   "homepage": "https://github.com/meetecho/janus-gateway#readme",
   "dependencies": {
-    "webrtc-adapter": "7.7.0"
+    "webrtc-adapter": "7.4.0"
   }
 }


januscla:
Thanks for your contribution, @augustblack! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

tmatth:
LGTM, at least until we figure out why https://cdnjs.cloudflare.com/ajax/libs/webrtc-adapter/7.4.0/adapter.min.js is the last version on cloudflare (cc @fippo)

Alternately we could switch the demos etc. to using unpkg.

lminiero:
Thanks, merging!

fippo:
@tmatth so there is a hint: adapter never provided the minified versions, see e.g. [here](https://github.com/webrtc/adapter/tree/gh-pages/adapter-7.4.0-variants). So this must have been something cloudflare did automagically but no longer does?

PR #2532, opened 2021-01-24
Adding DataChannel to AudioBridge plugin

afshin2003:
Hi Guys!
This is inspired by datachannel support on VideoRoom plugin!

BEST


lminiero:
@afshin2003 we try not to replicate the same features all over in different plugins. The AudioBridge is only meant to provide audio mixing functionality: datachannel relaying is available in the VideoRoom, while data channel "mixing" in the TextRoom. As such, applications can combine the features of different plugins when they need both (as we did ourselves in an audio call + chat application). As such, I'm afraid I'll have to close this. Thanks for the contribution anyway!

afshin2003:
@lminiero I see. My consideration is binary data. So what about to add binary packets to TextRoom? or I need to design a custom plugin for that?

Issue #2533, opened 2021-01-24
Janus doesn't detect failure at binding to a port in websocket correctly


Symbiatch:
If libwebsocket cannot bind to a socket Janus still claims the websocket is correctly initialized. This is due to libwebsocket by default not reporting failure on bind.

> WebSockets server started (port 8188)...

Adding `LWS_SERVER_OPTION_FAIL_UPON_UNABLE_TO_BIND` to `info.options` in websocket creation helps:

> [FATAL] [transports/janus_websockets.c:janus_websockets_init:643] Error creating vhost for WebSockets server...

I can create a PR for this if wanted

Symbiatch:
Fixed by #2534

PR #2534, opened 2021-01-24
Report fail if binding to a socket fails in websockets

Symbiatch:
Fixes issue #2533 when websocket server cannot be created due to a bind failure

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -636,7 +636,7 @@ int janus_websockets_init(janus_transport_callbacks *callback, const char *confi
 			info.ssl_private_key_password = NULL;
 			info.gid = -1;
 			info.uid = -1;
-#if (LWS_LIBRARY_VERSION_MAJOR == 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSIO_MAJOR > 3)
+#if (LWS_LIBRARY_VERSION_MAJOR == 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR > 3)
                         info.options = LWS_SERVER_OPTION_FAIL_UPON_UNABLE_TO_BIND;
 #else
                         info.options = 0;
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @Symbiatch! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

transports/janus_websockets.c
```diff
@@ -636,7 +636,11 @@ int janus_websockets_init(janus_transport_callbacks *callback, const char *confi
 			info.ssl_private_key_password = NULL;
 			info.gid = -1;
 			info.uid = -1;
-			info.options = 0;
+#if (LWS_LIBRARY_VERSION_MAJOR == 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR > 3)
+                        info.options = LWS_SERVER_OPTION_FAIL_UPON_UNABLE_TO_BIND;
+#else
+                        info.options = 0;
+#endif
```
lminiero, line 643: Indentation seems broken? Please use tabs, not spaces (code style nit).
As a side note, no need for the `else` part with the `0` init: we already memset the whole `info` object to `0` when we declare the variable.

transports/janus_websockets.c
```diff
@@ -753,7 +759,11 @@ int janus_websockets_init(janus_transport_callbacks *callback, const char *confi
...
```
lminiero, line 766: Same here.

transports/janus_websockets.c
```diff
@@ -818,7 +828,9 @@ int janus_websockets_init(janus_transport_callbacks *callback, const char *confi
 				info.ssl_cipher_list = ciphers;
 				info.gid = -1;
 				info.uid = -1;
-#if LWS_LIBRARY_VERSION_MAJOR >= 2
+#if (LWS_LIBRARY_VERSION_MAJOR == 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR > 3)
+                                info.options =  LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT | LWS_SERVER_OPTION_FAIL_UPON_UNABLE_TO_BIND;
+#elif LWS_LIBRARY_VERSION_MAJOR >= 2
```
lminiero, line 833: Same here. There's also an extra space after the equal sign that shouldn't be there.

lminiero (commented):
Thanks for taking care of this! I'll make a couple of tests later to check if it works as expected in my setup. I added a couple of notes inline.

Symbiatch:
> 
> 
> Thanks for taking care of this! I'll make a couple of tests later to check if it works as expected in my setup. I added a couple of notes inline.

Thanks, somehow the indentation was mangled, I'll fix it and remove the zeroes, they were there originally so kept them to be sure.

lminiero:
> they were there originally so kept them to be sure

Ah you're right, the diff had confused me! Removing them is still fine, since the memset does indeed take care of it.
Thanks for the quick patch! I'll test and get back to you later.

atoppi:
lgtm :+1:

lminiero:
LGTM too, thanks, merging! :+1:

PR #2535, opened 2021-01-24
Fix race condition in audiobridge plugin changeroom request

JeckLabs:
In `plugins/janus_audiobridge.c` the check for `participant->room == NULL` happends before locking `rooms_mutex`. It may cause a segfault later on `janus_mutex_lock(&old_audiobridge->mutex);`. This pull requests should fix it.


lminiero:
Makes sense, thanks for the fix! Merging :+1:

Issue #1856, opened 2019-11-06
Cannot hangup a ringing call: [ERR] [plugins/janus_sip.c:janus_sip_handler:3225] Wrong state (not established? status=inviting)


LucasFrezarini:
Hey there,

I'm having some trouble with the hang up of calls that hasn't started yet. When the caller dial to some number but give up of the call while it's ringing, the call doesn't hangup properly and keeps ringing to the callee for a while (something close to 10 seconds). I'm sending a `hangup` request to Janus when the caller hangs up, as recommended in the sip plugin [documentation](https://janus.conf.meetecho.com/docs/sipsofia.html):

> hangup can be used to terminate the communication at any time, either to hangup (BYE) an ongoing call or to cancel/decline (CANCEL/BYE) a call that hasn't started yet.

However, checking the SIP traffic, I've found that when the caller hangs up a call while it's ringing, Janus isn't sending a CANCEL/BYE SIP request. Analyzing the logs of Janus, when the hangup request is sent by the caller on a call in the inviting state, the following error is logged:

```
[Mon Nov  4 16:02:36 2019] [ERR] [plugins/janus_sip.c:janus_sip_handler:3225] Wrong state (not established? status=inviting)
```

After finding this out, I dug up the [code](https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_sip.c) related to the sip plugin to investigate what is causing this problem and to verify if I am doing something wrong, and I've found the `if clause` that is preventing the CANCEL request of being made (line 3744):

```C
// [...]
} else if(!strcasecmp(request_text, "hangup")) {
	/* Hangup an ongoing call */
	if(!janus_sip_call_is_established(session)) {
	JANUS_LOG(LOG_ERR, "Wrong state (not established? status=%s)\n", janus_sip_call_status_string(session->status));
	/* Ignore */
	janus_sip_message_free(msg);
	continue;
	//~ g_snprintf(error_cause, 512, "Wrong state (not in a call?)");
	//~ goto error;
}
// [...]
```

Just for testing, I've removed this `if clause` to check up if the hangup would work without this verification, and I've found that after this modification, when a hangup request is send to Janus by the caller on a ringing call, a SIP CANCEL request is send successfully and the call is terminated:

![cancelrequest](https://user-images.githubusercontent.com/16738007/68298670-af93dc80-0078-11ea-8c3f-bd5b5f78d49c.png)

[...]

PR #2536, opened 2021-01-27
Fix hangup on inviting status after #2521

AdrianoMartins:
The merge #2521 is causing a problem in the hangup of the call with status inviting, similar to the problem reported in #1856. I believe that the correct treatment for the situation of #2521, is to do with status different from destroyed.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2560,7 +2560,7 @@ static void janus_sip_hangup_media_internal(janus_plugin_session *handle) {
 		session->media.on_hold = FALSE;
 		janus_sip_call_update_status(session, janus_sip_call_status_closing);
 
-		if(g_atomic_int_get(&session->established))
+		if(!g_atomic_int_get(&session->destroyed))
 			nua_bye(session->stack->s_nh_i, TAG_END());
 		else
 			nua_respond(session->stack->s_nh_i, 480, sip_status_phrase(480), TAG_END());


januscla:
Thanks for your contribution, @AdrianoMartins! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Pinging @zayim as he contributed the patch you mentioned, so that you can both check if this fixed your requirements.

zayim:
Hi! Sorry for late response.

I agree that merge #2521 broke hangup of `inviting` calls. But change you suggested just reverts my PR, since `if` you changed will always be `true` (due to lines 2528-2529).

I suggest adding additional check in `if`, so we send BYE/CANCEL only if session is `established` or status is `inviting`. In all other cases, we would send SIP response instead of BYE, which is correct behaviour.

Something like this: https://github.com/meetecho/janus-gateway/compare/master...zayim:fix-sip-plugin-bye-on-inviting

I reproduced error on `inviting` calls on master, tested this code from my branch and verified that it fixes it.

lminiero:
@zayim I think you moved it too much ahead, since it within the mutex lock now. I can't remember if there were risks of deadlocks there, but if the main purpose is doing something before the state changes, maybe it's simpler to move it before the `janus_sip_call_update_status` call?

lminiero:
@AdrianoMartins ping :slightly_smiling_face:

zayim:
@lminiero I agree, this branch was in a rush just to test my theory. If Adriano agrees, I can edit by branch per your suggestion and make PR.

lminiero:
@zayim sounds good to me, I'll wait for you guys to agree on which solution works best for you both.

AdrianoMartins:
Sorry for the late response. @zayim I agree. Thanks.

lminiero:
@AdrianoMartins thanks for the feedback!

zayim:
Great, I will submit PR in next few hours.

Issue #2509, opened 2020-12-31
crashed Janus with error heap-use-after-free in videoroom plugin


Sensemakers-Nick:
Yesterday one of our servers crashed with the error: **heap-use-after-free** using the videorooms plugin.
Couldn't find updates to this particular part of code in GitHub, nor issues so decided to open a ticket.

gdb pastebin: [https://pastebin.com/6GZFShE0](https://pastebin.com/6GZFShE0)
libasan pastebin: [https://pastebin.com/YRX2j2j8](https://pastebin.com/YRX2j2j8)

The load of the server ramped up pretty badly before crashing.
![Image of the load](https://www.sensemakers.nl/loadspike.png)

Roughly 40 videorooms connected.

atoppi:
Hi @Sensemakers-Nick, thanks for the detailed report.
It seems that a room has been destroyed while the plugin was handling a negotiation.

More specifically the `videoroom->h264_profile` string attribute is being read by `g_ascii_strdown` in `janus_sdp_get_codec_pt_full` in order to craft a SDP answer, but since the room has been freed that string is not valid anymore, hence the libasan trap.

Maybe we need and extra reference on the videoroom while handling the negotiation.

Sensemakers-Nick:
Hi @atoppi,

Thanks for the reply, you're welcome.

atoppi:
@Sensemakers-Nick I'm attaching [a patch](https://github.com/meetecho/janus-gateway/files/5781400/fix-increase-vroom-ref.txt) that tries to address the issue, could you please apply it to your sources and test again?

```bash
git apply fix-increase-vroom-ref.txt
```

then rebuild janus

Sensemakers-Nick:
Thanks @atoppi 
Will apply the patch next week during a maintenance window and keep you posted!

tmatth:
@atoppi we haven't seen this crash but will test the patch and report any issues.

Sensemakers-Nick:
We've succesfully applied the patch, waiting for results, will let you know @atoppi

Sensemakers-Nick:
@atoppi The same error heap-use-after-free in videoroom plugin happened today, but this time during the alteration of allowed participants in a destroyed room.

Should I open a new ticket?

Pastebin: https://pastebin.com/yQDtQVFm

atoppi:
Yes please open a new issue and try to add some more info.
1. enable verbose logging (5)
[...]

PR #2539, opened 2021-01-28
Videoroom race condition fixes (see #2509)

lminiero:
A few fixes to the VideoRoom plugin, to address the issues documented in #2509. Will be merged as soon as we have confirmation the issues are solved.


lminiero:
Merging.

PR #2541, opened 2021-01-29
Add new option to configure ICE nomination mode, if libnice is recent enough

lminiero:
Since 0.1.15, libnice supports configuring the ICE nomination to mode: it defaults to "aggressive" (which I believe was the default for older versions too), but "regular" is supported as well. This PR adds a new option to the `nat` section of `janus.jcfg` to configure what to use in Janus, called `ice_nomination`. This property is obviously ignored if libnice older than 0.1.15.

Valid values for the property are, again, "regular" and "aggressive": if nothing is provided, we default to "regular", since as @fbellet pointed out in a [PR to libnice](https://gitlab.freedesktop.org/libnice/libnice/-/merge_requests/125), that's what [RFC8445](https://tools.ietf.org/html/rfc8445) (the updated RFC for ICE) suggests.

Since this can have an impact on connectivity establishment, please do test in your environments and provide feedback.

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -239,7 +239,10 @@ media: {
 # configured to do full-trickle (Janus also trickles its candidates to
 # users) rather than the default half-trickle (Janus supports trickle
 # candidates from users, but sends its own within the SDP), and whether
-# it should work in ICE-Lite mode (by default it doesn't). Finally,
+# it should work in ICE-Lite mode (by default it doesn't). If libnice is
+# at least 0.1.15, you can choose which ICE nomination mode to use: valid
+# values are "regular" and "aggressive" (the default depends on the libnice
+# version itself; if we can set it, we set regular nomination). Finally,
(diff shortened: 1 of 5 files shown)


lminiero:
As part of the additional configuration capabilities provided by this PR, I added support for libnice's `keepalive-conncheck` too, which you can find documented here: https://libnice.freedesktop.org/libnice/NiceAgent.html#NiceAgent--keepalive-conncheck

atoppi:
**Notice to all testers**
We found out that current libnice master [is breaking connections after 50 seconds](https://gitlab.freedesktop.org/libnice/libnice/-/issues/136) when `keepalive-conncheck` is being used.
So better to stick to `0.1.18` until the issue is not addressed upstream.

lminiero:
Added a warning for @atoppi's remark, and merging.

lminiero:
FYI, I've changed the default back to 'aggressive', as 'regular' seemed too slow in some deployments.

PR #2543, opened 2021-02-03
sdp-testcase: reproducer for pt + profile-level-id parsing bug

tmatth:
To run, simply build janus and:
./sdp-testcase

DO NOT MERGE

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -563,6 +563,24 @@ EXTRA_DIST += conf/janus.plugin.duktape.jcfg.sample.in
 endif
 
 
+bin_PROGRAMS += sdp-testcase
+sdp_testcase_SOURCES = \
+	sdp-testcase.c \
+	log.c \
+	sdp-utils.c \
+	utils.c \
+	$(NULL)
+
+sdp_testcase_CFLAGS = \
+	$(AM_CFLAGS) \
+	$(JANUS_CFLAGS) \
+	$(NULL)
+
+sdp_testcase_LDADD = \
+	$(JANUS_LIBS) \
+	$(JANUS_MANUAL_LIBS) \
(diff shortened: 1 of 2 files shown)


lminiero:
Whoops, sorry, I pressed the wrong button and marked it ready to review. How to make it a draft again?

lminiero:
Done, sorry for the noise :smile:

lminiero:
The issue is that there are multiple `rtpmap` attributes for the same codec, e.g.:

```
a=rtpmap:102 H264/90000
a=rtpmap:127 H264/90000
a=rtpmap:125 H264/90000
a=rtpmap:108 H264/90000
```

Since we parse the attributes in order, any time we find a matching `rtpmap` attribute we update the `pt` property: when looking for a H.264 profile, the last value of `pt` is `108`, so that's the payload type we check when looking for `fmtp` attributes. That's also why you sometimes get the wrong `pt` back (there's indeed such a profile for that payload type) or `-1` (no such profile for that specific payload type).

I'll have to think about the best way to address this. Rather than only keeping track of a single `pt` integer as we do now, we may have to start using a list of some sort (e.g., a `GList` of `GINT_TO_POINTER` elements) in `janus_sdp_get_codec_pt_full', so that if, for instance, H.264 has 4 different payload types, we keep track of them all, and when we have to check `fmtp`, we accept any of those for our validation process. An alternative may be a `GHashTable` but that sounds heavier. I'll play with this a bit and keep you posted: in case I manage to get something that "fixes" your testcase, I'll prepare a PR that can be used to ensure it doesn't mess with anything else that was still working instead.

lminiero:
Closing as I have your tester locally, and this doesn't really need to be merged. Thanks!

PR #2212, opened 2020-06-08
H264 profile fix

groupboard:
Fix janus_sdp_get_codec_pt_full so that it works when fmtp comes before rtpmap in the sdp, for vp9 and h264.

PR #2549, opened 2021-02-03
Fix parsing of SDP to find payload type matching profiles (fixes #2544)

### LINKED ISSUE #2544: videoroom: SDP PT parsing issue if codec profile is set
### Summary
For videoroom publishers that explicitly set the codec profile, it seems that the wrong payload type is being retrieved (specifically the pt number after the one we want, or -1) from the offer SDP. For chrome 88, I hadn't noticed the issue since given the adjacent lines:
```
"a=fmtp:125 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e01f\r\n"
"a=fmtp:108 level-asymmetry-allowed=1;packetization-mode=0;profile-level-id=42e01f\r\n"
```
returning 108 masked the bug (since the profile-level-id matched, even though it has an unsupported packetization-mode that should be excluding it).

However in chrome 90, it's effectively returning -1 and rejecting video:
```
[Tue Feb  2 17:12:54 2021] [WARN] Couldn't find codec we needed (h264) in the offer, rejecting video
```
presumably due to the addition of AV1 (payload type 35).

The line here is suspect since from the logs it seems to be logging that it has the next payload type already, e.g.
```
H.264 profile 42e01f found --> 35
```
where 35 is the payload type for AV1 (new to chrome 90).

https://github.com/meetecho/janus-gateway/blob/213b6c7c63a2fca5450d3dd7672a9472e8235ed1/sdp-utils.c#L731
--
### Reproducer
This should be illustrated by the reproducer here:
https://github.com/meetecho/janus-gateway/pull/2543

which, given some real SDPs (after they've been preprocessed by janus to remove RTX et al) fail at extracting the correct payload types:

```
$ ./sdp-testcase
# expecting 125, 108 is the next line and should be disqualified since it has packetization-mode=0
H.264 profile 42e01f found --> 108
[WARN] Got pt 108 for h264 42e01f

H.264 profile 42e01f found --> 35
[WARN] Got pt -1 for h264 42e01f

VP9 profile 0 found --> 108
[WARN] Got pt -1 for vp9 0

VP9 profile 0 found --> 35
[WARN] Got pt -1 for vp9 0

VP9 profile 2 found --> 108
[WARN] Got pt -1 for vp9 2

VP9 profile 2 found --> 35
[WARN] Got pt -1 for vp9 2
```

Hopefully I'm not missing something obvious and the reproducer makes the problem clear.

### Versions affected
master 213b6c7c63a2fca5450d3dd7672a9472e8235ed1

lminiero:
See #2544 for details, and https://github.com/meetecho/janus-gateway/pull/2543#issuecomment-772447679 for an explanation of this fix. I expanded the test @tmatth provided to validate finding codecs without a profile too, and with this PR it seems to be working as expected in all cases, so I assume this is now working as it should: please test to make sure there are no regressions, though.

This is important because this PR gets rid of a check we had before. In the old check, you could have a `fmtp` attribute before `rtpmap`, and it would still "work": now we always assume `rtpmap` attributes are listed before `fmtp` attributes, instead. This seems to be the case for the SDPs I've seen, but not sure if it's indeed like that in all browser (or non-browser) WebRTC implementations for instance. Again, feedback would help here.

### DIFF
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -700,50 +700,13 @@ int janus_sdp_get_codec_pt_full(janus_sdp *sdp, const char *codec, const char *p
 			ml = ml->next;
 			continue;
 		}
-		/* Look in all rtpmap attributes */
+		/* Look in all rtpmap attributes first */
 		GList *ma = m->attributes;
 		int pt = -1;
 		GList *pts = NULL;
 		while(ma) {
 			janus_sdp_attribute *a = (janus_sdp_attribute *)ma->data;
-			if(profile != NULL && a->name != NULL && a->value != NULL && !strcasecmp(a->name, "fmtp")) {
-				/* Does this match the payload types we're looking for? */
-				pt = atoi(a->value);
-				if(g_list_find(pts, GINT_TO_POINTER(pt)) == NULL) {
-					/* Not what we're looking for */
-					ma = ma->next;
-					continue;
-				}
-				if(vp9) {
-					char profile_id[20];
-					g_snprintf(profile_id, sizeof(profile_id), "profile-id=%s", profile);
-					if(strstr(a->value, profile_id) != NULL) {
-						/* Found */
-						JANUS_LOG(LOG_VERB, "VP9 profile %s found --> %d\n", profile, pt);
-						return pt;
-					}
-				} else if(h264 && strstr(a->value, "packetization-mode=0") == NULL) {
-					/* We only support packetization-mode=1, no matter the profile */
-					char profile_level_id[30];
-					char *profile_lower = g_ascii_strdown(profile, -1);
-					g_snprintf(profile_level_id, sizeof(profile_level_id), "profile-level-id=%s", profile_lower);
-					g_free(profile_lower);
-					if(strstr(a->value, profile_level_id) != NULL) {
-						/* Found */
-						JANUS_LOG(LOG_VERB, "H.264 profile %s found --> %d\n", profile, pt);
-						return pt;
-					}
-					/* Not found, try converting the profile to upper case */
-					char *profile_upper = g_ascii_strup(profile, -1);
-					g_snprintf(profile_level_id, sizeof(profile_level_id), "profile-level-id=%s", profile_upper);
-					g_free(profile_upper);
-					if(strstr(a->value, profile_level_id) != NULL) {
-						/* Found */
-						JANUS_LOG(LOG_VERB, "H.264 profile %s found --> %d\n", profile, pt);
-						return pt;
-					}
-				}
-			} else if(a->name != NULL && a->value != NULL && !strcasecmp(a->name, "rtpmap")) {
+			if(a->name != NULL && a->value != NULL && !strcasecmp(a->name, "rtpmap")) {
 				pt = atoi(a->value);
 				if(pt < 0) {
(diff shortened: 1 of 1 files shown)


tmatth:
@lminiero thanks for the quick fix. Does it still handle cases where the fmtp comes first (see https://github.com/meetecho/janus-gateway/pull/2212)?
I haven't seen such an SDP but in reordering one of the SDPs in my testcase, it didn't succeed.

lminiero:
> Does it still handle cases where the fmtp comes first (see #2212)?

No, as I wrote above it now expects the `rtpmap` attributes to be all before the `fmtp` ones, since there can be more than one. The only way to address this, I think, is doing two iterations: one to only parse `rtpmap` (and build the list), and one to only parse `fmtp` (only needed in case profile parsing is required). More inefficient, but would work.

tmatth:
> > Does it still handle cases where the fmtp comes first (see #2212)?
> 
> No, as I wrote above it now expects the `rtpmap` attributes to be all before the `fmtp` ones, since there can be more than one for the same codec. The only way to address this, I think, is doing two iterations: one to only parse `rtpmap` (and build the list), and one to only parse `fmtp` (only needed in case profile parsing is required). More inefficient, but would work.

Right, although if you short-circuit the passes earlier as soon as each are found it may not end up being too much worse.

lminiero:
@tmatth this new commit parses all `rtpmap` attributes before `fmtp`, and so solves the issue of out of order attributes (verified messing with the SDP in your tester). If no profile is specified, we never parse `fmtp` at all, and so stop sooner.

tmatth:
> @tmatth this new commit parses all `rtpmap` attributes before `fmtp`, and so solves the issue of out of order attributes (verified messing with the SDP in your tester). If no profile is specified, we never parse `fmtp` at all, and so stop sooner.

@lminiero I've cherry-picked this PR and added an artificial testcase with the fmtp lines first here that seems to fail:
https://github.com/tmatth/janus-gateway/commit/4e17a694b451f5f5a50981798abf208c3f004da6

lminiero:
Maybe something wrong with your cherry picking? Just tried again moving all the video rtmap ones after the fmtp ones and it works for me.

tmatth:
> Maybe something wrong with your cherry picking? Just tried again moving all the video rtmap ones after the fmtp ones and it works for me.

It's definitely including your changes:
```
4e17a694 (HEAD -> sdp-test, tmatth_ghub/sdp-test) sdp-testcase: add SDP with video fmtp lines first
fc4c771f Process all rtpmap attributes before fmtp ones
ad4a68a5 Fix parsing of SDP to find payload type matching profiles (fixes #2544)
f4385ed5 sdp-testcase: reproducer for pt + profile-level-id parsing bug
```

In my testcase, I didn't put all the fmtp lines at the top, the opus in is in the middle.

lminiero:
You moved the video fmtp attributes to the audio m-line... :stuck_out_tongue_closed_eyes:

tmatth:
> You moved the video fmtp attributes to the audio m-line... stuck_out_tongue_closed_eyes

lol ok if it's not a valid SDP than I think that's ok to not support.

lminiero:
If you move them around within the same m-line it should work as expected.

tmatth (approved):

lminiero:
Merging.

PR #2554, opened 2021-02-10
if inviting on destroy, send BYE instead of 480 response

zayim:
This is PR as result from discussion here: https://github.com/meetecho/janus-gateway/pull/2536#issuecomment-770772318 with @lminiero and @AdrianoMartins. 

It fixes session destroy in SIP plugin for calls with `inviting` status.


lminiero:
Thanks! If you both can confirm it fixes your issues, I'll merge.

AdrianoMartins:
Tested here, everything ok!

lminiero:
Thank you both, merging then! :+1:

PR #2557, opened 2021-02-16
[janus_sip] Add "call_id" to "calling", "declining", "updatingcall" & "incomingcall" events

nicolasduteil:
As the title mentions, this PR adds an extra `call_id` property (already added in other events such as `hangup`, `missed_call` or `accepted`) to the following events
* `calling`
* `declining`
* `updatingcall`
* `incomingcall`


lminiero:
Sounds reasonable, thanks for the contribution! Merging :+1:

PR #2558, opened 2021-02-17
 Error negotiating bandwidth with multiple line b

aljafiro:
Error negotiating bandwidth.
When receiving a sdp with the following lines b:

    v=0
    o=- ********* ********* IN IP4 **.**.**.**
    s=SBC call
    c=IN IP4 10.89.255.105
    t=0 0
    m=audio 8664 RTP/AVP 18 8 110
    b=AS:41
    b=RS:612
    b=RR:1837
    a=rtpmap:18 G729/8000
    a=rtpmap:8 PCMA/8000
    a=rtpmap:110 telephone-event/8000
    a=fmtp:110 0-15
    a=sendrecv
    a=maxptime:240

The following lines b were received, in this order: AS, RS and RR.
In our case there was a line B with "AS" which made it impossible to enter other lines, even if that line was RS or RR, generating a Multiple m-line b error when the RS arrived.

We propose to advance the check that prevents AS lines and TIAS lines from being mixed together at the beginning of the case, instead of at the end

### DIFF
--- a/sdp-utils.c
+++ b/sdp-utils.c
@@ -507,17 +507,12 @@ janus_sdp *janus_sdp_parse(const char *sdp, char *error, size_t errlen) {
 						break;
 					}
 					case 'b': {
-						if(strcmp(line, "AS") && strcmp(line, "TIAS")) {
-							/* We only support b=AS and b=TIAS, skip */
-							line += 2;
-							break;
-						}
-						if(mline->b_name) {
-							if(error)
-								g_snprintf(error, errlen, "Multiple m-line b= lines: %s", line);
-							success = FALSE;
-							break;
-						}
+						//if(mline->b_name) {
+							//if(error)
+							//	g_snprintf(error, errlen, "Multiple m-line b= lines: %s", line);
+							//success = FALSE;
+							//break;
+						//}
 						line += 2;
 						char *semicolon = strchr(line, ':');
 						if(semicolon == NULL || (*(semicolon+1) == '\0')) {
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @aljafiro! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Moving the check as you did will not do what you want: `line` is not just the value at that point, it's the whole attribute line, so that string compare will always fail.

aljafiro:
By setting the variable "success = FALSE;" when receiving a line b RR or RS, it invalidates any input that comes with a valid AS line

lminiero:
I understood the issue and it makes sense: what I meant is that your patch, as it is, is not the right solution. You moved this line at the beginning:

    if(strcmp(line, "AS") && strcmp(line, "TIAS")) {

but `line`, here, is not something like `AS` or `TIAS` or `RS`; it's something like `b=RS:612` so the check will always fail, breaking support for the `b` attribute in the SDP parser. The `line` pointer is only updated later in the code to prune the extra stuff.

sdp-utils.c
```diff
@@ -507,12 +507,12 @@ janus_sdp *janus_sdp_parse(const char *sdp, char *error, size_t errlen) {
...
-								g_snprintf(error, errlen, "Multiple m-line b= lines: %s", line);
-							success = FALSE;
-							break;
-						}
+						//if(mline->b_name) {
+							//if(error)
+							//	g_snprintf(error, errlen, "Multiple m-line b= lines: %s", line);
+							//success = FALSE;
+							//break;
+						//}
```
lminiero, line 515: Sorry, but this is not a proper fix either... I can try looking into this, but definitely not before a couple of weeks.

lminiero:
@aljafiro I made the multiple b= lines check a warning an not an error, and verified it works with the SDP sample you provided. As such, I'm closing your PR. Thanks!

PR #2563, opened 2021-02-17
add call_id in received sip message

tvildo:
call_id  was missing when you receive sip message during call


lminiero:
Can you add the same to `nua_i_info` as well? Both work pretty much the same way, in the code (and are both only done in-dialog), so it feels incomplete to only "fix" one.

tvildo:
@lminiero  Done

lminiero:
Thanks! Merging then :+1:

PR #2565, opened 2021-02-17
errors: replace strerror with locale-safe and threadsafe g_strerror

### LINKED ISSUE #2561: Use of strerror(errno) pattern in Janus
Currently (as of commit 4f8943adf7c3292a1a4aaa08fd022c1308b9357d) there are 140 invocations of `strerror(errno)`.

The problem is that since `strerror` uses a static buffer, this idiom isn't safe to use in a multithreaded process like Janus. The alternatives are:
- `char *strerror_r(int errnum, char *buf, size_t buflen);` <- caller passes  a buffer and its size, this is guaranteed to be thread-safe
- using the `%m` format string, e.g.
    `JANUS_LOG(LOG_ERR, "Error in sendto... %d (%s)\n", errno, strerror(errno));`
   becomes
   `JANUS_LOG(LOG_ERR, "Error in sendto... %d (%m)\n", errno);` (note the `%m` format string doesn't take an argument).
   
 I'd be happy to submit a PR but this is a lot of churn so I wanted to get feedback first.

tmatth:
Fixes #2561

### DIFF
--- a/events/janus_gelfevh.c
+++ b/events/janus_gelfevh.c
@@ -164,14 +164,14 @@ static int janus_gelfevh_connect(void) {
 				janus_network_address_to_string_buffer(&addr, &addr_buf) != 0) {
 		if(res)
 			freeaddrinfo(res);
-		JANUS_LOG(LOG_ERR, "Could not resolve address (%s): %d (%s)\n", backend, errno, strerror(errno));
+		JANUS_LOG(LOG_ERR, "Could not resolve address (%s): %d (%s)\n", backend, errno, g_strerror(errno));
 		return -1;
 	}
 	char *host = g_strdup(janus_network_address_string_from_buffer(&addr_buf));
 	freeaddrinfo(res);
 
 	if((sockfd = socket(AF_INET, transport, 0)) < 0 ) {
-		JANUS_LOG(LOG_ERR, "Socket creation failed: %d (%s)\n", errno, strerror(errno));
+		JANUS_LOG(LOG_ERR, "Socket creation failed: %d (%s)\n", errno, g_strerror(errno));
 		g_free(host);
 		return -1;
 	}
@@ -206,7 +206,7 @@ static int janus_gelfevh_send(char *message) {
 		while(length > 0) {
 			out_bytes = send(sockfd, buffer, length + 1, 0);
 			if(out_bytes <= 0) {
-				JANUS_LOG(LOG_WARN, "Sending TCP message failed, dropping event: %d (%s)\n", errno, strerror(errno));
+				JANUS_LOG(LOG_WARN, "Sending TCP message failed, dropping event: %d (%s)\n", errno, g_strerror(errno));
 				close(sockfd);
 				return -1;
 			}
@@ -240,7 +240,7 @@ static int janus_gelfevh_send(char *message) {
 		if(total == 1) {
 			int n = send(sockfd, buf, len, 0);
 			if(n < 0) {
-				JANUS_LOG(LOG_WARN, "Sending UDP message failed, dropping event: %d (%s)\n", errno, strerror(errno));
+				JANUS_LOG(LOG_WARN, "Sending UDP message failed, dropping event: %d (%s)\n", errno, g_strerror(errno));
 				return -1;
 			}
 			return 0;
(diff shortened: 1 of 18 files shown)


tmatth:
Reference https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strerror

tmatth:
N.B. I haven't had the chance to test, feedback welcome (especially if using a locale other than `C`).

tmatth:
Also note that `g_strerror` only switched to using `strerror_r` as of 2.53.5, so to reap the full benefit one will have to have a recent enough glib (otherwise the behaviour will be the same as before, but should be no worse):

https://gitlab.gnome.org/GNOME/glib/-/commit/18f8b77c04be373fd4f65560519ab0e09596801e
https://abi-laboratory.pro/?view=changelog&l=glib&v=2.53.6

lminiero:
Thanks! I'll have a look later or in the next few days for some tests. This part from the docs scares me though:

> The value of errno will not be changed by this function. However, it may be changed by intermediate function calls, so you should save its value as soon as the call returns:

as I really hope we won't need to change the code even more to address this as well... of course that's true for how we currently use `strerror` too, not specific to your change: just mentioning as I was reading the documentation you linked.

tmatth:
> Thanks! I'll have a look later or in the next few days for some tests. This part from the docs scares me though:
> 
> > The value of errno will not be changed by this function. However, it may be changed by intermediate function calls, so you should save its value as soon as the call returns:
> 
> as I really hope we won't need to change the code even more to address this as well... of course that's true for how we currently use `strerror` too, not specific to your change: just mentioning as I was reading the documentation you linked.

I think in the existing code it's always used immediately so shouldn't be an issue?

events/janus_gelfevh.c
```diff
@@ -164,14 +164,14 @@ static int janus_gelfevh_connect(void) {
 				janus_network_address_to_string_buffer(&addr, &addr_buf) != 0) {
 		if(res)
 			freeaddrinfo(res);
-		JANUS_LOG(LOG_ERR, "Could not resolve address (%s): %d (%s)\n", backend, errno, strerror(errno));
+		JANUS_LOG(LOG_ERR, "Could not resolve address (%s): %d (%s)\n", backend, errno, g_strerror(errno));
```
tmatth, line 167: Here if `freeaddrinfo` returns `EAI_SYSTEM` it will set errno, and otherwise its error code should be passed to `gai_strerror`. I think that kind of change belongs in a separate PR though.

lminiero:
@tmatth are you done with the changes? I see you're pushing more commits, but can't see what changed, since they're force pushed. No need to do that, since I'll squash before merging anyway.

tmatth:
> @tmatth are you done with the changes? I'd like to start testing, but I see you're pushing more commits, and can't see what changed, since they're force pushed. No need to do that, since I'll squash before merging anyway.

@lminiero I've just been rebasing on top of `master`. Also even with force pushes, in github if you click on the hashes where it says e.g.,:
> tmatth force-pushed the tmatth:bugfix/strerror branch from 4a65396 to c133b41 11 hours ago

you can see what the changes were.

lminiero:
> Also even with force pushes, in github if you click on the hashes where it says e.g.,:
> 
> > tmatth force-pushed the tmatth:bugfix/strerror branch from [4a65396](https://github.com/meetecho/janus-gateway/commit/4a6539684f8e44d0e945e9915b2581fa5b809d07) to [c133b41](https://github.com/meetecho/janus-gateway/commit/c133b41b9a28a85c48836de3bf62671ae0a13ad7) 11 hours ago
> 
> you can see what the changes were.

It always shows the complete list of changes to me, not just the few additional changes you may have done in the meanwhile. Maybe I'm clicking the wrong link, which wouldn't be impossible :hand_over_mouth:

tmatth:
> > Also even with force pushes, in github if you click on the hashes where it says e.g.,:
> > > tmatth force-pushed the tmatth:bugfix/strerror branch from [4a65396](https://github.com/meetecho/janus-gateway/commit/4a6539684f8e44d0e945e9915b2581fa5b809d07) to [c133b41](https://github.com/meetecho/janus-gateway/commit/c133b41b9a28a85c48836de3bf62671ae0a13ad7) 11 hours ago
> > 
> > 
> > you can see what the changes were.
> 
> It always shows the complete list of changes to me, not just the few additional changes you may have done in the meanwhile. Maybe I'm clicking the wrong link, which wouldn't be impossible hand_over_mouth

Oh yeah for reference I haven't modified my changeset at all, just rebased the same patch (and there haven't been conflicts). So really the only thing changing is the hash+timestamp,

lminiero:
Ah ok, thanks for clarifying! I was indeed wondering if you were making other changes I had to be aware of.

lminiero:
This will cause a conflict nightmare in the multistream branch, but no reason to hold on to this... merging :grin:

tmatth:
> This will cause a conflict nightmare in the multistream branch, but no reason to hold on to this... merging grin

Thankfully they are just one-line changes so it shouldn't be too bad.

PR #2567, opened 2021-02-23
Added Content type to SIP message

tijmenNL:
A SIP message has a content type defined. 

If you want to 'reliably' parse a message you can't do this with having the actual content type. An example are messages where the type is 'application/cpim' or 'application/im-iscomposing+xml'.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -411,6 +411,7 @@
 \verbatim
 {
 	"request" : "message",
+	"content_type" : "<content type; optional>"
 	"content" : "<text to send>"
 }
 \endverbatim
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @tijmenNL! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

tijmenNL:
I signed the CLA, however the page tells me I signed, but no details are displayed..

lminiero:
> I signed the CLA, however the page tells me I signed, but no details are displayed..

Yeah, it's a known bug in our handler (that we should fix sooner or later), but I can confirm I see the signature.
Tha patch looks good so I'll merge, thanks for your contribution!

PR #2569, opened 2021-02-23
Don't chain error handler to success handler in Janus.httpAPICall in janus.js

i8-pi:
I ran into a problem in janus.js where multiple HTTP longpoll requests were running simultaneously. This was caused by the success handler and error handler in Janus.httpAPICall for longpoll both being called, which happened because the error handler was chained to the success handler, and the success handler threw an error from running a user supplied callback

This change separates the success handler from the error handler so they won't both be called

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -137,8 +137,12 @@ Janus.useDefaultDependencies = function (deps) {
 				if(response.ok) {
 					if(typeof(options.success) === typeof(Janus.noop)) {
 						return response.json().then(function(parsed) {
-							options.success(parsed);
-						}).catch(function(error) {
+							try {
+								options.success(parsed);
+							} catch(error) {
+								Janus.error('Unhandled httpAPICall success callback error', error);
+							}
+						}, function(error) {
 							return p.reject({message: 'Failed to parse response body', error: error, response: response});
 						});
 					}


lminiero:
Apologies for the delay of this response, we're quite busy on other activities. I can't remember if we ever encountered this issue, but we'll have a look in the next few days and let you know.

html/janus.js
```diff
@@ -137,8 +137,12 @@ Janus.useDefaultDependencies = function (deps) {
 				if(response.ok) {
 					if(typeof(options.success) === typeof(Janus.noop)) {
 						return response.json().then(function(parsed) {
-							options.success(parsed);
-						}).catch(function(error) {
+							try {
+								options.success(parsed);
+							} catch (error) {
```
lminiero, line 142: Code style note: no space between `catch` and bracket.

lminiero (commented):
Apologies again for this late response, we were quite busy with the IETF meeting.

Can you elaborate on what would be the impact of this change on existing applications? It's not clear to me what would happen in case the try/catch there fails: would it break the long poll loop? Since it's only started internally, there is no way for users to restart it externally when that happens. Thanks!

i8-pi:
This change should not break anyone's workflow, since there is an implicit assumption that only either the success or error callback is run.

In the case of longpoll, both the success and error callbacks trigger another longpoll request, so the existing behaviour of calling both is to slowly fork bomb your browser. createSession is another function where running both callbacks leaves you in an inconsistent state. Other places that call httpAPICall have error callbacks that effectively only log messages to the console. So the existing behavior is either broken or inconsequential.

If the try/catch fails, in the case of longpoll the next loop should still be triggered since it happens very early in the success callback and all the code up to that point is in janus.js (i.e. not user supplied callbacks)

lminiero:
Ack, thanks for the clarification! We'll make a few tests and in case we notice no regression we'll merge :+1:

lminiero:
Apologies for the awful delay: I made a few tests, simulating errors in applications, and it seems to do the trick.
Merging then! :+1:

PR #2575, opened 2021-03-04
Using g_malloc0 instead of g_malloc

cppdev-1:
- using g_malloc0 instead of g_malloc
- cleaning up code


januscla:

ice.c
```diff
@@ -3050,23 +3047,19 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 							p->last_retransmit = now;
 							retransmits_cnt++;
 							/* Enqueue it */
-							janus_ice_queued_packet *pkt = g_malloc(sizeof(janus_ice_queued_packet));
-							pkt->data = g_malloc(p->length+SRTP_MAX_TAG_LEN);
+							janus_ice_queued_packet *pkt = g_malloc0(sizeof(janus_ice_queued_packet));
+							pkt->data = g_malloc0(p->length+SRTP_MAX_TAG_LEN);
 							memcpy(pkt->data, p->data, p->length);
```
tmatth, line 3052: In this instance `pkt->data` is immediately written to so I think `g_malloc0` is not needed/worth it.

plugins/janus_streaming.c
```diff
@@ -7911,7 +7877,7 @@ static void *janus_streaming_relay_thread(void *data) {
 							janus_mutex_lock(&source->keyframe.mutex);
 							JANUS_LOG(LOG_HUGE, "[%s] ... other part of keyframe received! ts=%"SCNu32"\n", name, source->keyframe.temp_ts);
 							janus_streaming_rtp_relay_packet *pkt = g_malloc0(sizeof(janus_streaming_rtp_relay_packet));
-							pkt->data = g_malloc(bytes);
+							pkt->data = g_malloc0(bytes);
 							memcpy(pkt->data, buffer, bytes);
```
tmatth, line 7881: same here

plugins/janus_streaming.c
```diff
@@ -7959,7 +7925,7 @@ static void *janus_streaming_relay_thread(void *data) {
 									JANUS_LOG(LOG_HUGE, "[%s] New keyframe received! ts=%"SCNu32"\n", name, source->keyframe.temp_ts);
 									janus_mutex_lock(&source->keyframe.mutex);
 									janus_streaming_rtp_relay_packet *pkt = g_malloc0(sizeof(janus_streaming_rtp_relay_packet));
-									pkt->data = g_malloc(bytes);
+									pkt->data = g_malloc0(bytes);
 									memcpy(pkt->data, buffer, bytes);
```
tmatth, line 7929: and here

plugins/janus_streaming.c
```diff
@@ -8056,7 +8022,7 @@ static void *janus_streaming_relay_thread(void *data) {
 					if(!mountpoint->enabled && !source->drc)
 						continue;
 					/* Copy the data */
-					char *data = g_malloc(bytes);
+					char *data = g_malloc0(bytes);
```
tmatth, line 8025: and here

plugins/janus_streaming.c
```diff
@@ -8071,7 +8037,7 @@ static void *janus_streaming_relay_thread(void *data) {
 						if(source->buffermsg) {
 							janus_mutex_lock(&source->buffermsg_mutex);
 							janus_streaming_rtp_relay_packet *pkt = g_malloc0(sizeof(janus_streaming_rtp_relay_packet));
-							pkt->data = g_malloc(bytes);
+							pkt->data = g_malloc0(bytes);
```
tmatth, line 8040: and here

plugins/janus_streaming.c
```diff
@@ -8522,7 +8488,7 @@ static void janus_streaming_helper_rtprtcp_packet(gpointer data, gpointer user_d
 	}
 	/* Clone the packet and queue it for delivery on the helper thread */
 	janus_streaming_rtp_relay_packet *copy = g_malloc0(sizeof(janus_streaming_rtp_relay_packet));
-	copy->data = g_malloc(packet->length);
+	copy->data = g_malloc0(packet->length);
```
tmatth, line 8491: and here

plugins/janus_textroom.c
```diff
@@ -1389,9 +1385,8 @@ void janus_textroom_incoming_data(janus_plugin_session *handle, janus_plugin_dat
 		janus_refcount_decrease(&session->ref);
 		return;
 	}
-	char *text = g_malloc(len+1);
+	char *text = g_malloc0(len+1);
```
tmatth, line 1388: and here

rtp.h
```diff
@@ -107,7 +107,7 @@ janus_audiocodec janus_audiocodec_from_name(const char *name);
 int janus_audiocodec_pt(janus_audiocodec acodec);
 
 typedef enum janus_videocodec {
-	JANUS_VIDEOCODEC_NONE,
+	JANUS_VIDEOCODEC_NONE = 0,
```
tmatth, line 110: These changes are unrelated and belong in a different commit.

sctp.c
```diff
@@ -69,16 +69,13 @@ typedef struct janus_sctp_pending_message {
 	size_t len;
 } janus_sctp_pending_message;
 static janus_sctp_pending_message *janus_sctp_pending_message_create(uint16_t id, gboolean textdata, char *buf, size_t len) {
-	janus_sctp_pending_message *m = g_malloc(sizeof(janus_sctp_pending_message));
+	janus_sctp_pending_message *m = g_malloc0(sizeof(janus_sctp_pending_message));
 	m->id = id;
 	m->textdata = textdata;
 	if(buf != NULL && len > 0) {
-		m->buf = g_malloc(len);
+		m->buf = g_malloc0(len);
```
tmatth, line 76: Also redundant since it's followed by a memcpy.

sdp-utils.h
```diff
@@ -78,7 +78,7 @@ const char *janus_sdp_mtype_str(janus_sdp_mtype type);
 /*! \brief Helper enumeration to quickly identify m-line directions */
 typedef enum janus_sdp_mdirection {
 	/*! \brief default=sendrecv */
-	JANUS_SDP_DEFAULT,
+	JANUS_SDP_DEFAULT = 0,
```
tmatth, line 81: This also is unrelated

lminiero:
Thanks for the contribution, but as @tmatth pointed out, we did go through this a long time ago, and actually did change many `g_malloc0` to `g_malloc` on purpose, especially when we knew we would be writing/setting everything. As such, we definitely can't accept this PR as it is.

That said, I don't have time right now for a full review, but I also glanced some dangerous and unexplained changes (like removing a destroyed flag set from DTLS? why?). Please elaborate on what exactly this patch is meant to do.

transports/janus_websockets.c
```diff
@@ -1370,7 +1366,7 @@ static int janus_websockets_common_callback(
 			const size_t remaining = lws_remaining_packet_payload(wsi);
 			if(ws_client->incoming == NULL) {
 				JANUS_LOG(LOG_HUGE, "[%s-%p] First fragment: %zu bytes, %zu remaining\n", log_prefix, wsi, len, remaining);
-				ws_client->incoming = g_malloc(len+1);
+				ws_client->incoming = g_malloc0(len+1);
```
tmatth, line 1369: Also redundant

transports/transport.c
```diff
@@ -1,4 +1,4 @@
-/*! \file   transport.h
+/*! \file   transport.c
```
tmatth, line 1: This belongs in a separate commit

tmatth (commented):
I'll let the meetecho team address this generally, but I'd say:
- I think this kind of thing makes sense for initializing structs like sessions that aren't in the hot path and you can see clearly here that it allows some deletion (and also if more future proof as elements get added to these structs)
- I'd be careful about dropping atomic ops (although in context, if the struct has just been initialized it's probably not accessible to other threads yet, so the existing code is probably just that way for consistency)
- Conversely, I'd be careful about doing this for packets/elements in the hot path without benchmarking...it'd probably make sense to do separate commits for the low frequency stuff (sessions, messages, etc.) vs. packets
- As I pointed out, I would not do `g_malloc0` if the existing code is already doing `memcpy` or an equivalent write immediately after allocation, as the zero initialization is never used. `g_malloc0` only makes sense if there's a chance that the code could read uninitialized bytes

utils.c
```diff
@@ -130,7 +130,7 @@ char *janus_random_uuid(void) {
 }
 
 guint64 *janus_uint64_dup(guint64 num) {
-	guint64 *numdup = g_malloc(sizeof(guint64));
+	guint64 *numdup = g_malloc0(sizeof(guint64));
```
tmatth, line 133: This is also redundant (value is overwritten on the next line)

rtp.h
cppdev-1: Do you mean that I should create a separate commit/PR?

tmatth: Yeah, this change is basically cosmetic since `JANUS_VIDEOCODEC_NONE` is already defined as 0 (because of C's rules for enums), and is unrelated to your `g_malloc0` changes, so should be in a separate commit.

cppdev-1:
- Basically, I've changed g_malloc to g_malloc0 just to be sure that if I change something in the future that change won't break anything. For example, if I change pkt->data to pkt->some_structure_data that change might break smth w/o initialisation, Also, sometimes it reduces the amount of code. Yes, right now in some cases it's redundant, but g_malloc0 not slower than g_malloc.
- As for removing a destroyed flag setting it's already done by g_malloc0.

transports/janus_websockets.c
cppdev-1: Because of this code:
`ws_client->incoming = g_malloc(len+1);
memcpy(ws_client->incoming, in, len);
**ws_client->incoming[len] = '\0';`**

BTW, I forgot removing the last line.

sdp-utils.h
cppdev-1: It's just to be 100% sure that JANUS_SDP_DEFAULT will be equal zero.

tmatth: C already guarantees this.

transports/transport.c
cppdev-1: Separate commit/PR?

tmatth: Yes, it's should be in a separate commit as it's unrelated to `g_malloc0` changes.

rtp.h
cppdev-1: It is, but someone can change it to e.g. -2. There are such enums in Janus. I make it start from 0 just to show it must start from 0 and remove code something like `some field = JANUS_VIDEOCODEC_NONE ` because of g_malloc0.

sdp-utils.h
cppdev-1: It is, but someone can change it to e.g. -2. There are such enums in Janus. I make it start from 0 just to show it must start from 0 and remove code something like some field = JANUS_SDP_DEFAULT because of g_malloc0.

plugins/janus_textroom.c
cppdev-1: Because of line `*(text+len) = '\0';`

lminiero:
Sorry, but I'm closing this. As explained by @tmatth, we already updated the usage of `g_malloc` vs `g_malloc0` in the past to address when it was and wasn't needed. Considering we still need to merge the `multistream` branch and this PR would introduce a ton of conflicts for changes that are not, strictly speaking, important, we cannot merge this at this stage. Should you want to provide a similar patch after the multistream branch is merged, we'll consider it, assuming that you separate the different set of changes in separate PRs, rather than all together as you did here.

Thanks!

cppdev-1:
@lminiero Ok, I'll provide modified patch later.

PR #2577, opened 2021-03-09
Added support for admin-protected custom session timeouts

alg:
Adds support for specifying custom session timeouts during session creation. Custom timeouts are protected behind `admin_secret` and not allowed for general purpose scenarios. Default timeout is the one defined globally (currently `60 seconds`).

This is useful for backend-initiated sessions that can't be regularly pinged. At the same time all other client-initiated sessions should respect the global setting.

Example session with 1 hour timeout:

```
{
    "janus": "create",
    "transaction": "trx-create",
    "admin_secret": "janusoverlord",
    "timeout": 3600
}
```

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -124,6 +124,9 @@ static struct janus_json_parameter debug_parameters[] = {
 static struct janus_json_parameter timeout_parameters[] = {
 	{"timeout", JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE}
 };
+static struct janus_json_parameter session_timeout_parameters[] = {
+	{"timeout", JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED}
+};
 static struct janus_json_parameter level_parameters[] = {
 	{"level", JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE}
 };
@@ -275,7 +278,7 @@ static json_t *janus_create_message(const char *status, uint64_t session_id, con
  * incurring in unexpected timeouts (when HTTP is used in janus.js, the
  * long poll is used as a keepalive mechanism). */
 #define DEFAULT_SESSION_TIMEOUT		60
-static uint session_timeout = DEFAULT_SESSION_TIMEOUT;
+static uint global_session_timeout = DEFAULT_SESSION_TIMEOUT;
 
 #define DEFAULT_RECLAIM_SESSION_TIMEOUT		0
 static uint reclaim_session_timeout = DEFAULT_RECLAIM_SESSION_TIMEOUT;
@@ -326,7 +329,7 @@ static json_t *janus_info(const char *transaction) {
 	json_object_set_new(info, "data_channels", json_false());
 #endif
 	json_object_set_new(info, "accepting-new-sessions", accept_new_sessions ? json_true() : json_false());
-	json_object_set_new(info, "session-timeout", json_integer(session_timeout));
+	json_object_set_new(info, "session-timeout", json_integer(global_session_timeout));
 	json_object_set_new(info, "reclaim-session-timeout", json_integer(reclaim_session_timeout));
 	json_object_set_new(info, "candidates-timeout", json_integer(candidates_timeout));
 	json_object_set_new(info, "server-name", json_string(server_name ? server_name : JANUS_SERVER_NAME));
@@ -655,8 +658,6 @@ static void janus_request_unref(janus_request *request) {
 }
 
 static gboolean janus_check_sessions(gpointer user_data) {
-	if(session_timeout < 1 && reclaim_session_timeout < 1)		/* Session timeouts are disabled */
-		return G_SOURCE_CONTINUE;
 	janus_mutex_lock(&sessions_mutex);
 	if(sessions && g_hash_table_size(sessions) > 0) {
 		GHashTableIter iter;
@@ -668,10 +669,15 @@ static gboolean janus_check_sessions(gpointer user_data) {
 				continue;
 			}
 			gint64 now = janus_get_monotonic_time();
-			if ((session_timeout > 0 && (now - session->last_activity >= (gint64)session_timeout * G_USEC_PER_SEC) &&
-					!g_atomic_int_compare_and_exchange(&session->timeout, 0, 1)) ||
+
+			/* Use either session-specific timeout or global. */
+			gint64 timeout = (gint64)session->timeout;
+			if (timeout == -1) timeout = (gint64)global_session_timeout;
+
+			if ((timeout > 0 && (now - session->last_activity >= timeout * G_USEC_PER_SEC) &&
+					!g_atomic_int_compare_and_exchange(&session->timedout, 0, 1)) ||
 					((g_atomic_int_get(&session->transport_gone) && now - session->last_activity >= (gint64)reclaim_session_timeout * G_USEC_PER_SEC) &&
-							!g_atomic_int_compare_and_exchange(&session->timeout, 0, 1))) {
+							!g_atomic_int_compare_and_exchange(&session->timedout, 0, 1))) {
 				JANUS_LOG(LOG_INFO, "Timeout expired for session %"SCNu64"...\n", session->session_id);
 				/* Mark the session as over, we'll deal with it later */
 				janus_session_handles_clear(session);
@@ -739,8 +745,9 @@ janus_session *janus_session_create(guint64 session_id) {
 	session->session_id = session_id;
 	janus_refcount_init(&session->ref, janus_session_free);
 	session->source = NULL;
+	session->timeout = -1; /* Negative means rely on global timeout */
 	g_atomic_int_set(&session->destroyed, 0);
-	g_atomic_int_set(&session->timeout, 0);
+	g_atomic_int_set(&session->timedout, 0);
 	g_atomic_int_set(&session->transport_gone, 0);
 	session->last_activity = janus_get_monotonic_time();
 	session->ice_handles = NULL;
@@ -1063,6 +1070,7 @@ int janus_process_incoming_request(janus_request *request) {
 				goto jsondone;
 			}
 		}
+
 		/* Handle it */
 		session = janus_session_create(session_id);
 		if(session == NULL) {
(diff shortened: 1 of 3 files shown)


januscla:
Thanks for your contribution, @alg! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks for the contribution, but I don't like the idea of mixing Janus and Admin API concepts, in this case the usage of the admin password from the Janus API. The "create" should remain as is, IMO, with possibly a new Admin API method to dynamically change the timeout value for a specific session. This would allow a controlling application to do both calls when needed, and also change the timeout value later rather than at startup (in case it's ever needed).

Currently busy with the IETF meeting so I can't review the code in detail, will try to do this next week :+1:

alg:
@lminiero Sounds equally good to me. I didn't see anything bad with setting the session value there since it's the place it was set to default global anyway and the admin password was just the protection not to let anyone choose the value freely. If you feel a separate call would be a better shot, I can rework it. Maybe give it another thought while you are away. Let me know.

janus.c
```diff
@@ -1063,8 +1062,30 @@ int janus_process_incoming_request(janus_request *request) {
 				goto jsondone;
 			}
 		}
+
+		// Parse session timeout
```
lminiero, line 1066: Please use `/* */` for comments (code style).

lminiero (commented):
I still think it should be a separate request, and only in the Admin API, on existing sessions created the "old" way: it's cleaner in terms of separation of concerns for the APIs, and as anticipated allows for dynamic tweaking of the timeout value also after the fact.

For the rest, I only added a small note on code style for comments. The code changes look fine to me, with the difference of course that now `create` should ensure the new `timeout` property is initialized to the global one: any change to that can be enforced via whatever new request will be added to the Admin API.

alg:
Ok, got it. Will update the PR shortly.

alg:
@lminiero Just pushed the updated version. See if you like it.

lminiero:
Apologies for the late reply @alg. It looks like this is still done as part of the Janus API, though, not the Admin API: the only change I see is that it's done in a separate request and not at session creation time anymore, but it's still the Janus API. As per the exchanges we had, this should be an Admin API method.

alg:
@lminiero Sure, no problem. I think I misunderstood your previous comment. I noticed there's already `set_session_timeout` in admin section that currently sets the global session timeout value. So I created a similar call in session-specific section of admin API and directed both to the same handler that deals with both cases. See if you like it.

janus.c
```diff
@@ -2668,12 +2653,16 @@ int janus_process_incoming_admin_request(janus_request *request) {
 				janus_events_notify_handlers(JANUS_EVENT_TYPE_SESSION, JANUS_EVENT_SUBTYPE_NONE,
 					session_id, "destroyed", NULL);
 			goto jsondone;
-		}
-		/* If this is not a request to destroy a session, it must be a request to list the handles */
-		if(strcasecmp(message_text, "list_handles")) {
+		} else if (!strcasecmp(message_text, "set_session_timeout")) {
+			/* Specific session timeout setting. */
+			ret = janus_process_set_session_timeout(request, session, transaction_text, root);
```
lminiero, line 2658: See above: the action seems relatively simples that it could be done inline, as the other requests handled here.

janus.h
```diff
@@ -105,6 +107,10 @@ void janus_session_handles_clear(janus_session *session);
  * @param[in] session The Janus Core-Client session
  * @returns The JSON array */
 json_t *janus_session_handles_list_json(janus_session *session);
+/*! \brief Method to set session timeout value.
+ * @param[in] session The Janus Core-Client session
+ * @param[in] timeout Timeout value in seconds. */
+void janus_session_set_timeout(janus_session *session, uint timeout);
```
lminiero, line 113: We don't really need the function here, I think. It's probably my fault as initially I put a lot of functions in this header that really didn't need exposing, as they're only used within `janus.c` anyway.

janus.h
```diff
@@ -171,6 +177,14 @@ int janus_process_success(janus_request *request, json_t *payload);
...
  */
 int janus_process_error(janus_request *request, uint64_t session_id, const char *transaction, gint error, const char *format, ...) G_GNUC_PRINTF(5, 6);
+/*! \brief Method that handles session set_timeout request
+ * @param[in] request The request instance and its source
+ * @param[in] session The Janus session
+ * @param[in] transaction The Janus transaction identifier
+ * @param[in] root The request payload root
+ * @returns 0 on success, a negative integer otherwise
+ */
+int janus_process_set_session_timeout(janus_request *request, janus_session *session, const gchar *transaction, json_t *root);
```
lminiero, line 187: See above. I actually think that, besides removing them from the header, we can get rid of both methods and do it more simply inline where the requests are processed, since the update functionality is simple.

lminiero (commented):
Thanks for the quick fixes! I added some inline notes: to summarize, I think we don't need the helper functions, as the action is simple enough to be done inline (which would make the new request fit more with the others as well).

janus.c
```diff
@@ -1999,28 +2005,7 @@ int janus_process_incoming_admin_request(janus_request *request) {
...
-			}
-			session_timeout = timeout_num;
-			/* Prepare JSON reply */
-			json_t *reply = json_object();
-			json_object_set_new(reply, "janus", json_string("success"));
-			json_object_set_new(reply, "transaction", json_string(transaction_text));
-			json_object_set_new(reply, "timeout", json_integer(session_timeout));
-			/* Send the success reply */
-			ret = janus_process_success(request, reply);
+			ret = janus_process_set_session_timeout(request, session, transaction_text, root);
```
lminiero, line 2008: Whoops, looks like this comment didn't make it through with the others. I basically wrote that I don't understand why this needed to change, and that it feels confusing to use the same method for two different things.

alg:
@lminiero Updated.

To tell the truth, I'm a bit overwhelmed by the source code. A file with 5.5k lines is something I've never seen before in my 30 years of software development practice. It's your project and you decide hot to evolve it, but it's already hard to navigate and understand with lots of duplication and all these goto's. I attempted to extract handling of set session timeout action to keep it in one place. Not sure why you told they do radically different things. They both change session timeout (global or specific session), do the same data parsing / checks and return almost the same set of fields. Again, your project, your vision.

lminiero:
I get your point, but right now that's how the code is structured, and it needs to be consistent. While we may change this in the future, it needs to be done for everything and not just parts of it, otherwise it only adds to the confusion.

Thanks for the changes, I'll make some tests. I just realized this may be a breaking change for some, as before when changing the global session timeout this had an impact on all sessions: now this will only have an impact on new sessions created after the change, because existing sessions (even those that still have the default value) will be stuck with the old one, and would need to be updated independently. It may be worthwhile to add a new setting to also update the timeout value of some or all sessions when changing the global one (e.g., "if session->timeout == old timeout, update that one too), but that's something we can do in another PR later on. This breaking change means we need to merge only when we tag a major version change, though: the next one will be 0.11.0 so it should be ok.

alg:
I wonder if it's a wanted contribution at all. If not, I'm fine with it. I can always fork. No need to go through this lengthy routine.

If it is, then instead of writing a global session timeout into a specific session, we can write `-1` showing that there's no local session override and a global timeout value should be used. If there's a non-negative value in the session, it should be used instead.

lminiero:
> I wonder if it's a wanted contribution at all. If not, I'm fine with it. I can always fork. No need to go through this lengthy routine.

I'm sorry you got this from our exchanges, it was not my intention. Of course I do appreciate the contribution, and see value in it. At the same time, as the main maintainer of the project I have to always be extremely careful of the impact contributions can have, as once they're in they're _my_ responsibility: Janus is used in a ton of deployments, and what can be considered breaking changes needs to be taken into account. If you think I was asking _you_ to worry about this, I wasn't: I explicitly said in my previous post it's something we would have taken care of ourselves later on.

> If it is, then instead of writing a global session timeout into a specific session, we can write `-1` showing that there's no local session override and a global timeout value should be used. If there's a non-negative value in the session, it should be used instead.

This might be a good solition, since the "scope" of a session timeout is small enough that we can use a signed integer for it (most of our values are `uint64_t` so that wouldn't be possible otherwise).

alg:
Good to know it's something the project would benefit from.

I changed it as discussed above. See if you like it.

lminiero:
Thanks, from a quick glance it looks good, I'll test it later!
The only thing I noticed is there's still a leftover `janus_session_set_timeout` definition in `janus.h`.

alg:
You are right. Sorry, missed that bit.

janus.c
```diff
@@ -2668,12 +2679,45 @@ int janus_process_incoming_admin_request(janus_request *request) {
...
-		}
-		/* If this is not a request to destroy a session, it must be a request to list the handles */
-		if(strcasecmp(message_text, "list_handles")) {
+		} else if (!strcasecmp(message_text, "set_session_timeout")) {
+			/* Specific session timeout setting. */
+			JANUS_VALIDATE_JSON_OBJECT(root, session_timeout_parameters,
+				error_code, error_cause, FALSE,
+				JANUS_ERROR_MISSING_MANDATORY_ELEMENT, JANUS_ERROR_INVALID_ELEMENT_TYPE);
+			if(error_code != 0) {
+				ret = janus_process_error_string(request, session_id, transaction, error_code, error_cause);
```
lminiero, line 2688: I got a warning here when compiling:

```
janus.c:2688:59: warning: passing argument 3 of ‘janus_process_error_string’ from incompatible pointer type [-Wincompatible-pointer-types]
 2688 |     ret = janus_process_error_string(request, session_id, transaction, error_code, error_cause);
      |                                                           ^~~~~~~~~~~
      |                                                           |
      |                                                           json_t * {aka struct json_t *}
janus.c:89:96: note: expected ‘const char *’ but argument is of type ‘json_t *’ {aka ‘struct json_t *’}
```

This should indeed be `transaction_text` and not `transaction` (which is a jansson object).

janus.c
```diff
@@ -2668,12 +2679,45 @@ int janus_process_incoming_admin_request(janus_request *request) {
...
+				goto jsondone;
+			}
+
+			/* Positive timeout is seconds, 0 is unlimited, -1 is global session timeout */
+			json_t *timeout = json_object_get(root, "timeout");
+			int timeout_num = json_integer_value(timeout);
+			if(timeout_num < -1) {
+				ret = janus_process_error(request, session_id, transaction, JANUS_ERROR_INVALID_ELEMENT_TYPE, "Invalid element type (timeout should be a non-negative integer or -1)");
```
lminiero, line 2696: Same here.

janus.c
```diff
@@ -2668,12 +2679,45 @@ int janus_process_incoming_admin_request(janus_request *request) {
...
+
+			/* Set specific session timeout */
+			janus_mutex_lock(&session->mutex);
+			session->timeout = timeout_num;
+			janus_mutex_unlock(&session->mutex);
+
+			/* Prepare JSON reply */
+			json_t *reply = json_object();
+			json_object_set_new(reply, "janus", json_string("success"));
+			json_object_set_new(reply, "transaction", json_string(transaction));
```
lminiero, line 2708: Same here.

lminiero (commented):
There were some warnings during the compilation, notes inline. I'll fix them manually for my local tests.

lminiero:
Just tested in a few different scenarios and seems to be working as expected :+1: 
Merging, thanks again for your contribution! (and patience :slightly_smiling_face:  )

Issue #2580, opened 2021-03-15
Combined tokens / password access


deep9:
After update to 0.10.10 the streaming connection is not working.

I have api_secret enabled as it allows for token issue, which is then used for user connection.

Before the update it was working fine, now it seems that if the api_secret is set, it needs to be used with the token.

Strange behavior is that the POST request will work, but the GET will return 403.

 I believe the error is located here (https://github.com/meetecho/janus-gateway/blob/46a6c71b0fb7e54cb9acfbec84391c75eb224c13/transports/janus_http.c#L1553), because similar logic is used in here (https://github.com/meetecho/janus-gateway/blob/46a6c71b0fb7e54cb9acfbec84391c75eb224c13/janus.c#L917) and the logic is different (I believe working one, according to the description).

Also, the POST request will eventually end up here (https://github.com/meetecho/janus-gateway/blob/ad54495df09e8b96386df40b96b4212fe36a92b7/janus.c#L899) where the logic is different.

It was "fixed" in this commit (https://github.com/meetecho/janus-gateway/commit/16173af36f9e0644f3cfc7518c05f6a53438d0dc)

I've read the issues and I don't think the fix is valid. 

The proper one should be to still check if the secret/token is needed before setting the value to true.

I don't know C that much to create a PR, but if that is still needed, I can try to do so.

deep9:
Something like this

```
if(!(secret_authorized && gateway->is_api_secret_needed(&janus_http_transport)) && !(token_authorized && gateway->is_auth_token_needed(&janus_http_transport))) {
				response = MHD_create_response_from_buffer(0, NULL, MHD_RESPMEM_PERSISTENT);
				janus_http_add_cors_headers(msg, response);
				ret = MHD_queue_response(connection, MHD_HTTP_FORBIDDEN, response);
				MHD_destroy_response(response);
				goto done;
			}
```

deep9:
So, based on the findings, I think it would maybe make sense to create one validation function in the gateway (janus.c) and call it from those 2 places to prevent errors :)

deep9:
So, I wanted to test it, there is a working version here, not sure if it is correct language wise.
https://github.com/deep9/janus-gateway/tree/hotfix/token-secret-auth

lminiero:
[...]

PR #2524, opened 2021-01-20
Fixed secret authentication on GET requests

remvst:
Follow-up to https://github.com/meetecho/janus-gateway/issues/2520

The secret-based authentication wasn't properly enforced on the long-poll request.

Using this small Node.js client:
```js
import axios from 'axios';

console.log('Testing janus yo');

async function test() {
    console.log('Creating session');

    const sessionResponse = await axios.post('http://localhost:8088/janus', {
        'janus': 'create',
        'transaction': 'createsession',
        'apisecret': 'somesecret',
    });
    const sessionId = sessionResponse.data.data.id;
    console.log(`Session ID is ${sessionId}`);


    console.log('Polling');
    let pollResponse;
    try {
        pollResponse = await axios.get(`http://localhost:8088/janus/${sessionId}`, {
        });
    } catch (err) {
        pollResponse = err.response;
    }

    console.log(pollResponse);

    if (pollResponse.status !== 403) {
        throw new Error('Did not get expected 403');
    }
}

test()
    .then(() => console.log('passed'))
    .catch((err) => console.error(err));
```

Without the fix, I get a keep-alive after 30 seconds (unexpected), with the fix, I get the 403 immediately (expected).

All credit goes to @lminiero for finding the broken part of the code 🙏

PR #2581, opened 2021-03-16
Fix auth when both (token, secret) modes are enabled

deep9:
Fix for #2580

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -919,7 +919,7 @@ static int janus_request_check_secret(janus_request *request, guint64 session_id
 			}
 		}
 		/* We consider a request authorized if either the proper API secret or a valid token has been provided */
-		if(!secret_authorized && !token_authorized)
+		if(!(api_secret != NULL && secret_authorized) && !(janus_auth_is_enabled() && token_authorized))
 			return JANUS_ERROR_UNAUTHORIZED;
 	}
 	return 0;
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -1536,7 +1536,9 @@ static MHD_Result janus_http_handler(void *cls, struct MHD_Connection *connectio
 		const char *secret = MHD_lookup_connection_value(connection, MHD_GET_ARGUMENT_KIND, "apisecret");
 		const char *token = MHD_lookup_connection_value(connection, MHD_GET_ARGUMENT_KIND, "token");
 		gboolean secret_authorized = FALSE, token_authorized = FALSE;
-		if(!gateway->is_api_secret_needed(&janus_http_transport) && !gateway->is_auth_token_needed(&janus_http_transport)) {
+		gboolean is_api_secret_needed = gateway->is_api_secret_needed(&janus_http_transport);
+		gboolean is_auth_token_needed = gateway->is_auth_token_needed(&janus_http_transport);
+		if(!is_api_secret_needed && !is_auth_token_needed) {
 			/* Nothing to check */
 			secret_authorized = TRUE;
 			token_authorized = TRUE;
@@ -1549,8 +1551,8 @@ static MHD_Result janus_http_handler(void *cls, struct MHD_Connection *connectio
 				/* Token is valid or disabled */
 				token_authorized = TRUE;
 			}
-			/* We consider a request authorized if both the token and the API secret are either disabled or valid */
-			if(!secret_authorized || !token_authorized) {
+			/* We consider a request authorized if either the proper API secret or a valid token has been provided */
+			if(!(is_api_secret_needed && secret_authorized) && !(is_auth_token_needed && token_authorized)) {
 				response = MHD_create_response_from_buffer(0, NULL, MHD_RESPMEM_PERSISTENT);
 				janus_http_add_cors_headers(msg, response);
 				ret = MHD_queue_response(connection, MHD_HTTP_FORBIDDEN, response);


janus.c
```diff
@@ -919,7 +919,7 @@ static int janus_request_check_secret(janus_request *request, guint64 session_id
 			}
 		}
 		/* We consider a request authorized if either the proper API secret or a valid token has been provided */
-		if(!secret_authorized && !token_authorized)
+		if(!(api_secret != NULL && secret_authorized) && !(janus_auth_is_enabled() && token_authorized))
```
lminiero, line 922: I'm not sure that's needed. It's easier to change the defaults above, e.g.:

```
gboolean secret_authorized = (api_secret == NULL), token_authorized = !janus_auth_is_enabled();
```

This way the checks are simplified, because what's disabled automatically resolves to `TRUE`. This would also allow us to remove the `Nothing to check` branch entirely and only do something if either mechanism is needed.

transports/janus_http.c
```diff
@@ -1536,7 +1536,9 @@ static MHD_Result janus_http_handler(void *cls, struct MHD_Connection *connectio
 		const char *secret = MHD_lookup_connection_value(connection, MHD_GET_ARGUMENT_KIND, "apisecret");
 		const char *token = MHD_lookup_connection_value(connection, MHD_GET_ARGUMENT_KIND, "token");
 		gboolean secret_authorized = FALSE, token_authorized = FALSE;
-		if(!gateway->is_api_secret_needed(&janus_http_transport) && !gateway->is_auth_token_needed(&janus_http_transport)) {
+		gboolean is_api_secret_needed = gateway->is_api_secret_needed(&janus_http_transport);
+		gboolean is_auth_token_needed = gateway->is_auth_token_needed(&janus_http_transport);
+		if(!is_api_secret_needed && !is_auth_token_needed) {
```
lminiero, line 1541: Same here: if defaults for `secret_authorized` and `token_authorized` are set to `TRUE` if the related mechanism is disabled, then the checks are simpler. In this case, I agree the check below must get back to an `&&` rather than an `||` as in the core.

lminiero (commented):
Thanks, added some notes inline.

janus.c
deep9: I think this will not work, because the problem is in the function `is_api_secret_valid` - if it is disabled, it will return TRUE and we are back where we were before.

So the code might look like this:
```c
	gboolean secret_authorized = (api_secret == NULL), token_authorized = !janus_auth_is_enabled();
	if(secret_authorized && token_authorized) {
		/* Nothing to check */
	} else {
		json_t *root = request->message;
		if(!secret_authorized) {
			/* There's an API secret, check that the client provided it */
			json_t *secret = json_object_get(root, "apisecret");
			if(secret && json_is_string(secret) && janus_strcmp_const_time(json_string_value(secret), api_secret)) {
				secret_authorized = TRUE;
			}
		}
		if(!token_authorized) {
			/* The token based authentication mechanism is enabled, check that the client provided it */
			json_t *token = json_object_get(root, "token");
			if(token && json_is_string(token) && janus_auth_check_token(json_string_value(token))) {
				token_authorized = TRUE;
			}
		}
		/* We consider a request authorized if either the proper API secret or a valid token has been provided */
		if(!secret_authorized && !token_authorized)
			return JANUS_ERROR_UNAUTHORIZED;
	}
```

transports/janus_http.c
deep9: So the code based on the suggestion might look like

```c
		gboolean secret_authorized = !gateway->is_api_secret_needed(&janus_http_transport), token_authorized = gateway->is_auth_token_needed(&janus_http_transport);
		if(secret_authorized && token_authorized) {
			/* Nothing to check */
		} else {
			if(!secret_authorized && gateway->is_api_secret_valid(&janus_http_transport, secret)) {
				/* API secret is valid or disabled */
				secret_authorized = TRUE;
			}
			if(!token_authorized && gateway->is_auth_token_valid(&janus_http_transport, token)) {
				/* Token is valid or disabled */
				token_authorized = TRUE;
			}
			/* We consider a request authorized if either the proper API secret or a valid token has been provided */
			if(!secret_authorized && !token_authorized) {
				response = MHD_create_response_from_buffer(0, NULL, MHD_RESPMEM_PERSISTENT);
				janus_http_add_cors_headers(msg, response);
				ret = MHD_queue_response(connection, MHD_HTTP_FORBIDDEN, response);
				MHD_destroy_response(response);
				goto done;
			}
		}
```

janus.c
lminiero: Good point, makes sense :+1:

lminiero: Actually, thinking about it my notes were just wrong... the defaults should NOT be `TRUE`, because it means that if only one mechanism is enabled, it will always be bypassed by the `TRUE` on the *other*, disabled, mechanism (e.g., tokens in use but no API secret, wrong token still passes the check thanks to `secret_authorized` being `TRUE`).

I think that was the key reason why we did that change in #2524.

deep9: You are right, I did not think about that much when implementing the notes :) The previous solution works fine, even though it is a bit more complicated.

lminiero:
Thanks for the additional fixes! I just restarted the automated checks, as they failed for your build due to issues unrelated to the code. I'll make a few tests to see if they work in the different configurations.

lminiero:
Seems to be working fine :+1: 
@deep9, can you sign our [CLA](https://cla.conf.meetecho.com/janus/)? Apparently the bot didn't ask you for some reason (hope it didn't break), but I don't see your ID in the list. That's a prerequisite for merging, thanks!

lminiero:
Ok, I see the signature now: merging then, thanks!

PR #2584, opened 2021-03-17
fixing race condition in the videoroom plugin

cppdev-1:
The reason: thread tries to clear the mutex whilst another thread is holding it.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1688,6 +1688,7 @@ typedef struct janus_videoroom_subscriber {
 	volatile gint destroyed;
 	janus_refcount ref;
 } janus_videoroom_subscriber;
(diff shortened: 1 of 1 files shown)


januscla:

lminiero:
Mh I don't think this PR can't be accepted, sorry. It introduces a new global mutex for updating stuff, which seems way overkill. If race conditions on critical paths are the issue, they should be solved with careful usages of the existing mutexes (which is what we've tried to do so far).

cppdev-1:
@lminiero In fact, I've tried to use janus_videoroom_session->mutex as a good alternative to global mutex, but I had a deadlock with it. Also, there was possibility of crash in another place. According to my tests there were no significant drawbacks against global mutex. Only session destruction time increased slighly.

lminiero:
Closing as we have a separate fix for the issue. Thanks for bringing it to our attention!

PR #2597, opened 2021-03-22
adding contact header to SIP INVITE request

cppdev-1:
This PR solves the 'missing contact header problem' with the newer versions of libsofia (>=1.13). Here we generate contact header from SIPTAG and add it to SIP INVITE request.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -906,6 +906,13 @@ typedef struct ssip_oper_s ssip_oper_t;
 #undef NUA_HMAGIC_T
 #define NUA_HMAGIC_T	ssip_oper_t
 
+typedef struct janus_event_t
+{
+	gboolean is_ready;
+	janus_mutex mutex;
+	janus_condition cond;
+} janus_event_t;
+
 struct ssip_s {
 	su_home_t s_home[1];
 	su_root_t *s_root;
@@ -914,6 +921,8 @@ struct ssip_s {
 	GHashTable *subscriptions;
 	janus_mutex smutex;
 	struct janus_sip_session *session;
+	const gchar* sip_from_str;
+	janus_event_t event; // event object waiting for nua_r_get_params NUA event
 };
 
 typedef struct janus_sip_transfer {
@@ -1030,6 +1039,70 @@ static GHashTable *masters;
 static GHashTable *transfers;
 static janus_mutex sessions_mutex = JANUS_MUTEX_INITIALIZER;
 
+static void janus_event_init(janus_event_t *event)
+{
+	if (event == NULL)
+		return;
+
+	event->is_ready = FALSE;
+	janus_mutex_init(&event->mutex);
+	janus_condition_init(&event->cond);
+}
+
+static void janus_event_destroy(janus_event_t *event)
+{
+	if (event == NULL)
+		return;
+
+	event->is_ready = FALSE;
+	janus_condition_destroy(&event->cond);
+	janus_mutex_destroy(&event->mutex);
+}
+
+static void janus_event_reset(janus_event_t *event)
+{
+	if (event == NULL)
+		return;
+
+	event->is_ready = FALSE;
+}
+
+static void janus_event_set(janus_event_t *event)
+{
+	if (event == NULL)
+		return;
+
+	janus_mutex_lock(&event->mutex);
+	event->is_ready = TRUE;
+	janus_condition_signal(&event->cond);
+	janus_mutex_unlock(&event->mutex);
+}
+
+static gboolean janus_event_wait_timeout(janus_event_t *event, gint64 timeout)
+{
+	gboolean sts = FALSE;
+	gint64 end_time = 0;
+
+	if (event == NULL || timeout <= 0)
+		return sts;
+
+	janus_mutex_lock(&event->mutex);
+	end_time = g_get_monotonic_time() + timeout * G_TIME_SPAN_SECOND;
+	while (!event->is_ready)
+	{
+		sts = janus_condition_wait_until(&event->cond, &event->mutex, end_time);
+		if (!sts)
+			goto exit; // timeout occured
+	}
+
+	sts = TRUE;
+
+	exit:
+	janus_event_reset(event);
+	janus_mutex_unlock(&event->mutex);
+	return sts;
+}
+
 static void janus_sip_srtp_cleanup(janus_sip_session *session);
 static void janus_sip_media_reset(janus_sip_session *session);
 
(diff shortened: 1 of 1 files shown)


januscla:

plugins/janus_sip.c
```diff
@@ -906,6 +906,13 @@ typedef struct ssip_oper_s ssip_oper_t;
 #undef NUA_HMAGIC_T
 #define NUA_HMAGIC_T	ssip_oper_t
 
+typedef struct janus_event_t
+{
```
lminiero, line 910: Code style: please put the opening bracket on the same line as the struct.
I noticed that in the functions/while/if blocks you added it's done too, so please update them accordingly.

plugins/janus_sip.c
```diff
@@ -2986,37 +3064,54 @@ static void *janus_sip_handler(void *data) {
 			}
 
 			session->account.registration_status = janus_sip_registration_status_registering;
-			if(!refresh && session->stack == NULL) {
+			if (!refresh) {
```
lminiero, line 3067: Why have you removed this check?

plugins/janus_sip.c
```diff
@@ -2986,37 +3064,54 @@ static void *janus_sip_handler(void *data) {
...
 				g_thread_try_new(tname, janus_sip_sofia_thread, session, &error);
 				if(error != NULL) {
-					janus_refcount_decrease(&session->ref);
 					JANUS_LOG(LOG_ERR, "Got error %d (%s) trying to launch the SIP Sofia thread...\n",
 						error->code, error->message ? error->message : "??");
 					error_code = JANUS_SIP_ERROR_UNKNOWN_ERROR;
 					g_snprintf(error_cause, 512, "Got error %d (%s) trying to launch the SIP Sofia thread",
 						error->code, error->message ? error->message : "??");
 					g_error_free(error);
+					janus_refcount_decrease(&session->ref);
```
lminiero, line 3081: Why have you moved this?

plugins/janus_sip.c
```diff
@@ -2986,37 +3064,54 @@ static void *janus_sip_handler(void *data) {
...
-					if(timeout >= 2000000) {
+				while (timeout < 2000000) {
+					janus_mutex_lock(&session->stack->smutex);
+					if (session->stack->s_nua != NULL) {
+						janus_mutex_unlock(&session->stack->smutex);
 						break;
 					}
+					janus_mutex_unlock(&session->stack->smutex);
+					g_usleep(100000);
+					timeout += 100000;
```
lminiero, line 3093: I don't understand the changes above, which seem unrelated to the main purpose of the patch.

plugins/janus_sip.c
```diff
@@ -6653,15 +6795,14 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 		return NULL;
 	}
 	JANUS_LOG(LOG_VERB, "Joining sofia loop thread (%s)...\n", session->account.username);
-	session->stack = g_malloc0(sizeof(ssip_t));
-	su_home_init(session->stack->s_home);
```
lminiero, line 6657: Why have you removed this? AFAIK it's needed.

plugins/janus_sip.c
```diff
@@ -2023,7 +2100,12 @@ void janus_sip_create_session(janus_plugin_session *handle, int *error) {
 	session->account.outbound_proxy = NULL;
 	session->account.registration_status = janus_sip_registration_status_unregistered;
 	session->status = janus_sip_call_status_idle;
-	session->stack = NULL;
+
+	session->stack = g_malloc0(sizeof(ssip_t));
+	su_home_init(session->stack->s_home);
```
lminiero, line 2105: I don't think it's a good idea to do this here. IIRC we did it from the loop thread for a reason.

plugins/janus_sip.c
```diff
@@ -6694,6 +6836,7 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 				SIPTAG_SUPPORTED(NULL),
 				NTATAG_CANCEL_2543(session->account.rfc2543_cancel),
 				TAG_NULL());
+	janus_mutex_unlock(&session->stack->smutex);
```
lminiero, line 6839: Why these mutexes?

plugins/janus_sip.c
```diff
@@ -1030,6 +1039,70 @@ static GHashTable *masters;
...
+			goto exit; // timeout occured
+	}
+
+	sts = TRUE;
+
+	exit:
+	janus_event_reset(event);
+	janus_mutex_unlock(&event->mutex);
+	return sts;
+}
```
lminiero, line 1104: I think that this whole structure + functions are a bit overkill, if the only purpose is waiting for a string to be updated in response to the `nua_get_params`: this feels like something that could be done much more simply with the existing "wait" loop we have + maybe an atomic flag to trigger a state change.

lminiero (commented):
Thanks for the contribution! I didn't know you could use `nua_get_params` to retrieve the contact, but the way it's implemented seems a bit too complex to me. Besides, there are apparently some changes that don't seem related. I've added some notes inline.

plugins/janus_sip.c
cppdev-1: I'll do.

plugins/janus_sip.c
cppdev-1: stack is created in janus_sip_create_session. So, it cannot be NULL here.

plugins/janus_sip.c
cppdev-1: I have a slightly different log line here. I'll return janus_refcount_decrease to the previous place.

plugins/janus_sip.c
cppdev-1: stack is always created in janus_sip_create_session.

plugins/janus_sip.c
cppdev-1: We create session->stack->s_nua in different thread. So, we have to use the mutex here. However, you are right - it's a part of different patch.

plugins/janus_sip.c
cppdev-1: It works fine and also fixes a race.

plugins/janus_sip.c
cppdev-1: Because we access session->stack->s_nua from different thread.

plugins/janus_sip.c
cppdev-1: Yes, atomic flag + loop + sleep also can be used but I don't like "busy" wait.

cppdev-1:
Well, in fact, there are two different patches in this PR:
1. Move stack initialisation into janus_sip_create_session. The purpose of this patch is to fix a race.
2. Add contact header to SIP INVITE request.

I added both because I don't know if second will work w/o first.

cppdev-1:
@lminiero I've fixed the code style.

lminiero:
Sorry for the lack of feedback, I've been busy on other things. I still think this is overly complicated, and introduces a new construct exclusively for this purpose, which for me is overkill. My guess is this might be obtained more simply as a consequence of the fact we first do a REGISTER and then an INVITE. I'll give this some more thought to see if this can be achieved in a simpler way.

cppdev-1:
@lminiero 
> I still think this is overly complicated, and introduces a new construct exclusively for this purpose, which for me is overkill

Do you mean `janus_event_t`? If so, then it can be used in other places where "busy" wait is used.

> My guess is this might be obtained more simply as a consequence of the fact we first do a REGISTER and then an INVITE

I use guest mode and SIP REGISTER isn't used there.

PR #2599, opened 2021-03-25
Prevent race conditions on socket close in SIP and NoSIP plugins

lminiero:
We noticed that in some cases, in misconfigured instances of Janus with high socket contention and high traffic, a race condition could happen where a socket would be closed twice. This is an issue whenever the second time we close a socket it was actually assigned to someone else in the meanwhile: if the socket was used internally by `getifaddrs`, for instance, this could cause that method to lock and never return, and since in Janus that call may be on a critical path, deadlock relevant functionality.

This PR tries to fix this behaviour in SIP and NoSIP plugins, as the SIP plugin is where we noticed this happening, and the NoSIP plugin borrows a lot of the same code. The Streaming plugin doesn't seem to be affected instead, as only a mountpoint thread messes with the struct file descriptor values.

Feedback welcome, especially if you use those plugins a lot in production.

### DIFF
--- a/plugins/janus_nosip.c
+++ b/plugins/janus_nosip.c
@@ -1238,7 +1238,9 @@ static void janus_nosip_hangup_media_internal(janus_plugin_session *handle) {
 	}
 	/* Do cleanup if media thread has not been created */
 	if(!session->media.ready && !session->relayer_thread) {
+		janus_mutex_lock(&session->mutex);
 		janus_nosip_media_cleanup(session);
+		janus_mutex_unlock(&session->mutex);
 	}
 	/* Get rid of the recorders, if available */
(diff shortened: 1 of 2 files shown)


bhakimi:
thanks a ton!!!! @lminiero

lminiero:
@ihusejnovic since I know you guys use the SIP plugin a lot, we'd welcome feedback on this PR: we'd like to merge it soon. Thx!

ihusejnovic:
Sure, we will test it. Can you just sync this branch with master?

ihusejnovic:
We are using this branch last few days and we didn't notice any issues so far.

lminiero:
Merging.

PR #2604, opened 2021-03-26
janus.js - renegotiate with external stream

kmeyerhofer:
This PR adds the ability for an existing PeerConnection to `replaceVideo` with an external `stream` and back again through the `createOffer` renegotiation.

The issue this PR solves is described in this Google Group post [https://groups.google.com/g/meetecho-janus/c/Ucz-U3OORHM/m/mo6lXkdjAwAJ](https://groups.google.com/g/meetecho-janus/c/Ucz-U3OORHM/m/mo6lXkdjAwAJ)
To summarize the issue, a renegotiation `createOffer` sent with a canvas `MediaStream` (external stream) does not replace the track of the existing PeerConnection.

This PR allows the `stream` property sent in a `createOffer` parameter to be an object which can include the `replaceVideo` and `externalStream` properties, similar to the `media` property. The existing stream property behavior has not been modified.

Here is a working example:
```javascript
// Assume Janus is correctly initialized
var canvas = document.getElementById('canvas');
var stream = canvas.captureStream();
janusPlugin.createOffer({
  stream: {
    externalStream: stream,
    replaceVideo: true,
  },
  simulcast: false,
  simulcast2: false,
  success: (jsep) => {
    janusPlugin.send({ message: { request: "configure", audio: true, video: true}, jsep });
  },
  error: (error) => { Janus.error('error', error); },
});
```

Using `createOffer` in this manner replaces a user media (video camera) stream with an external `MediaStream` *and* an external `MediaStream` with another external `MediaStream`.

To renegotiate the same peer connection from an external `MediaStream` back to user media (video camera) utilizes the same `media` property as before. Here is a working example:
```javascript
// Assume Janus is correctly initialized
janusPlugin.createOffer({
  media: {
    video: userMediaVideoDevice,
    replaceVideo: true,
  },
  simulcast: false,
  simulcast2: false,
  success: (jsep) => {
    janusPlugin.send({ message: { request: "configure", audio: true, video: true}, jsep });
  },
  error: (error) => { Janus.error('error', error); },
});
```

In summary this change allows a PeerConnection renegotiation to replace the video in the following three conditions:
1. User media -> external MediaStream
2. External MediaStream -> external MediaStream
3. External MediaStream -> User media

#### Implementation decision making
I was a bit unsure of design implications regarding this note in the `createOffer`/`createAnswer` documentation (emphasis added):

>stream: optional, only to be passed in case you obtained a MediaStream object yourself with a getUserMedia request, and that you want the library to use instead of having it get one by itself **(makes the media property useless, as it won't be read for accessing any device);**

This is why I chose to modify the `stream` object instead of using the already available `{ media: { video: true, replaceVideo: true } }` property.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1741,7 +1741,7 @@ function Janus(gatewayCallbacks) {
 		}
 		// We're now capturing the new stream: check if we're updating or if it's a new thing
 		var addTracks = false;
-		if(!config.myStream || !media.update || config.streamExternal) {
+    if(!config.myStream || !media.update || (config.streamExternal && !media.replaceVideoExternalStream)) {
 			config.myStream = stream;
 			addTracks = true;
 		} else {
@@ -1773,7 +1773,7 @@ function Janus(gatewayCallbacks) {
 					config.pc.addTrack(stream.getAudioTracks()[0], stream);
 				}
 			}
-			if(((!media.update && isVideoSendEnabled(media)) || (media.update && (media.addVideo || media.replaceVideo))) &&
+			if(((!media.update && isVideoSendEnabled(media)) || (media.update && (media.addVideo || media.replaceVideo || media.replaceVideoExternalStream))) &&
 					stream.getVideoTracks() && stream.getVideoTracks().length) {
 				config.myStream.addTrack(stream.getVideoTracks()[0]);
 				if(Janus.unifiedPlan) {
@@ -2079,8 +2079,12 @@ function Janus(gatewayCallbacks) {
 			// can go directly to preparing the new SDP offer or answer
 			if(callbacks.stream) {
 				// External stream: is this the same as the one we were using before?
-				if(callbacks.stream !== config.myStream) {
+        var newStream = (callbacks.stream.externalStream ? callbacks.stream.externalStream : callbacks.stream);
+				if(newStream !== config.myStream) {
 					Janus.log("Renegotiation involves a new external stream");
+          if(callbacks.stream.replaceVideo) {
+            media.replaceVideoExternalStream = true;
+          }
 				}
 			} else {
 				// Check if there are changes on audio
@@ -2232,7 +2236,7 @@ function Janus(gatewayCallbacks) {
 					}
 				}
 			}
-			if(media.removeVideo || media.replaceVideo) {
+			if(media.removeVideo || media.replaceVideo || media.replaceVideoExternalStream) {
 				if(config.myStream && config.myStream.getVideoTracks() && config.myStream.getVideoTracks().length) {
 					var vt = config.myStream.getVideoTracks()[0];
 					Janus.log("Removing video track:", vt);
@@ -2260,12 +2264,12 @@ function Janus(gatewayCallbacks) {
 		}
 		// Was a MediaStream object passed, or do we need to take care of that?
 		if(callbacks.stream) {
-			var stream = callbacks.stream;
+      var stream = callbacks.stream.externalStream ? callbacks.stream.externalStream : callbacks.stream;
 			Janus.log("MediaStream provided by the application");
 			Janus.debug(stream);
 			// If this is an update, let's check if we need to release the previous stream
 			if(media.update) {
-				if(config.myStream && config.myStream !== callbacks.stream && !config.streamExternal) {
+				if(config.myStream && config.myStream !== callbacks.stream && !config.streamExternal && !media.replaceVideoExternalStream) {
 					// We're replacing a stream we captured ourselves with an external one
 					Janus.stopAllTracks(config.myStream);
 					config.myStream = null;
--- a/html/janus.js
+++ b/html/janus.js
@@ -2196,10 +2196,6 @@ function Janus(gatewayCallbacks) {
 				if(media.addData) {
 					media.data = true;
 				}
-				// Reset external stream flag when replacing MediaStream with user media
-				if(config.streamExternal) {
-					config.streamExternal = false;
-				}
 			}
 			// If we're updating and keeping all tracks, let's skip the getUserMedia part
 			if((isAudioSendEnabled(media) && media.keepAudio) &&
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @kmeyerhofer! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

kmeyerhofer:
CLA signed.

lminiero (commented):
Thanks for your contribution! But I feel this needlessly makes things more complicated than they should. My feeling is that this could work just as easily by reusing the existing properties: if `media` contains `replaceAudio` and/or `replaceVideo` and `stream` contains a (new) external MediaStream, then replace the tracks (audio and/or video) with the ones from the new stream and renegotiate. Or am I missing something?

Don't worry too much about what the docs say: for the JS part, most was written at the very beginning, and they may not be exactly up-to-date. The fact `media` is ignored, for instance, is not true, since media direction is taken into account even when using external streams.

kmeyerhofer:
Thanks for your review Lorenzo. I agree with your concern that these changes make things more complicated. Since you've clarified that using `media` `replaceAudio` and `replaceVideo` with a new stream is preferred, I can rewrite these changes to use them.

kmeyerhofer:
Based on your feedback @lminiero, I have simplified the previous changes to utilize the `media` `replaceAudio` and `replaceVideo` properties. 

With the added changes, a `createOffer` or `createAnswer` can now be used like this:

```javascript
// Assume Janus is correctly initialized
var canvas = document.getElementById('canvas');
var stream = canvas.captureStream();
janusPlugin.createOffer({
  stream: stream,
  media: {
    replaceAudio: true,
    replaceVideo: true,
  },
  simulcast: false,
  simulcast2: false,
  success: (jsep) => {
    janusPlugin.send({ message: { request: "configure", audio: true, video: true}, jsep });
  },
  error: (error) => { Janus.error('error', error); },
});
```

lminiero:
Thanks! I'll have a look at the new changes and make some tests when I get the chance.

html/janus.js
```diff
@@ -2206,7 +2210,7 @@ function Janus(gatewayCallbacks) {
 			}
 		}
 		// If we're updating, check if we need to remove/replace one of the tracks
-		if(media.update && !config.streamExternal) {
+		if(media.update && !config.streamExternal || (config.streamExternal && (media.replaceAudio || media.replaceVideo))) {
```
lminiero, line 2213: I think the check as you modified it is ambiguous, since there's an `&&` and an `||` in the same "sentence", so it's not clear what the AND applies to, and what the OR either: did you mean to group the old check in parenthesis as well, and make it an OR between the two groups?

html/janus.js
```diff
@@ -2196,6 +2196,10 @@ function Janus(gatewayCallbacks) {
 				if(media.addData) {
 					media.data = true;
 				}
+				// Reset external stream flag when replacing MediaStream with user media
+				if(config.streamExternal) {
+					config.streamExternal = false;
+				}
```
lminiero, line 2202: But doesn't this mean that the user loses control of the `MediaStream` they provided? Resetting this flag means we'll close the stream when we're done, which the user may not want to happen if they did the `getUserMedia` themselves and just gave us the new stream to use.

lminiero (commented):
Added a few notes inline.

html/janus.js
kmeyerhofer: It is ambiguous, I will make it more clear with an extra set of parenthesis:

`if(media.update && (!config.streamExternal || (config.streamExternal && (media.replaceAudio || media.replaceVideo))))`

html/janus.js
kmeyerhofer: To get to this part of the code, the following conditionals have already been met:
Line 2070 - `!config.pc` is falsey. There is already a PeerConnection.
Line 2080 - `callbacks.stream` is falsey. There isn't an external stream present.

If `config.streamExternal === true` and there isn't an external stream, we'll want to set this flag `false` to clean up.

This flag reset is for the case where the application sends another `createOffer` without an external `stream`, after an initial `createOffer` with an external `stream` _was_ sent. Example: see the second code block in my initial PR description.


I did some testing with this `config.streamExternal = false;` line removed and the renegotiation does work, but it probably should be reset.

I did some tests in my Janus VideoRoom plugin application. We start all participants as publishers with a standard getUserMedia for video and audio and subscribe to any other publisher upon entering the room. One of the participants can use a canvas to replace their video stream. I did some tests where I:

1. Made a `createOffer` with a new external `stream` and `replaceVideo: true`, replacing the userVideo device.
2. Made a `createOffer` with a new external `stream` and `replaceVideo: true`, replacing the previous external `stream`.
3. Made a `createOffer` with user media and `replaceVideo: true`, replacing the external `stream`.

In test case 3, `config.streamExternal === true` at line 2200. It doesn't effect the rest of the flow of the `prepareWebrtc` function, but it seems like it should be cleaned up if it is `true` at that point.

lminiero: Thanks for the explanation, but I still have some doubts. When people provide an external stream, we never mess with it, as the application may be using it for other things as well at the same time, and so it's important we don't remove its tracks for instance, as otherwise any usage the developer may be doing of the same MediaStream somewhere else may be broken as a result. I guess it's probably hard to enforce that, especially with renegotiations, so if you can confirm this patch does the job in the cases you tested it in, it's good enough for me, and we can maybe loosen the assumptions for the future (meaning that if what I highlighted is indeed an issue, we can make this the developer's responsibility to give us a stream we can mess with, if needed).

kmeyerhofer: What I can confirm in my tests is that setting `config.streamExternal = false;` here does not have an effect on what I'm trying to achieve. I do not want this PR to break backwards compatibility, especially if I don't have a solid method of testing it, so I will remove it.

lminiero:
@kmeyerhofer can this be merged? We never use external streams ourselves, so if you tested this and it works for you, it's good enough for me!

kmeyerhofer:
@lminiero Yes this is ready to be merged, thanks!

lminiero:
Merging then, thanks again :+1:

PR #2607, opened 2021-03-30
conf: fix commented out ice nomination default

tmatth:
Since 20dd2db2c60a99fb89a26e2ae53750ebab66da38 the default is actually `aggressive` (if libnice is recent enough).

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -257,7 +257,7 @@ nat: {
 	#stun_port = 3478
 	nice_debug = false
 	#full_trickle = true
-	#ice_nomination = "regular"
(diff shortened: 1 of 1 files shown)


lminiero:
Actually, the text in the config file is correct. As a generic rule (probably not always respected) I put as a commented out sample a value different from the default, as uncommenting and setting the default value would be a non-action in case. Since we now default to `aggressive`, it's normal for the config to show the option to set `regular` by just uncommenting the line. Closing.

lminiero:
As a reference, it's the same we're doing for `full_trickle`, `ice_lite`, etc., which are disabled by default and so the sample shows how to enable them instead.

PR #2609, opened 2021-03-30
Fixed sending responses from Janus for incoming SIP MESSAGE/SIP INFO

tijmenNL:
For SIP MESSAGE and SIP INFO the response is already handled by Sofia sip. If Janus is trying to send the 200 OK, we get :
`[WARN] [tijmen][nua_i_error]: 500 Responding to a Non-Existing Request`

I also removed the 488 for the case of missing payload, because my guess is Sofia SIP will also handle this and send a proper response. If this is the case we simple need to return and do nothing since it is an invalid packet anyway.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -5083,8 +5083,6 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			JANUS_LOG(LOG_VERB, "[%s][%s]: %d %s\n", session->account.username, nua_event_name(event), status, phrase ? phrase : "??");
 			/* We expect a payload */
 			if(!sip->sip_content_type || !sip->sip_content_type->c_type || !sip->sip_payload || !sip->sip_payload->pl_data) {
-				nua_respond(nh, 488, sip_status_phrase(488),
-					NUTAG_WITH_CURRENT(nua), TAG_END());
 				return;
 			}
 			const char *type = sip->sip_content_type->c_type;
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for your contribution!

> I also removed the 488 for the case of missing payload, because my guess is Sofia SIP will also handle this and send a proper response. If this is the case we simple need to return and do nothing since it is an invalid packet anyway.

Have you verified that's indeed the case, though? Otherwise we risk leaving pending dialogs.

tijmenNL:
Yes, you're right about the pending dialogs. But no it did not test it since I can't easily make a wrong packet. 

Since Sofia is auto answering with 200OK, I can't imagine why it would not respond with a failure code if something is wrong. Also if you want to have the application handle this method, it needs to be explicitly specified with NUTAG_APPL_METHOD("MESSAGE"). So my point with this is, that since it handles the 200OK how would you know that you, as application, you need to **only** handle the cases where the packet is wrong.

I will try to send a bogus packet, to see what happens.

tijmenNL:
I managed to send a packet without Content-Type and Content.

I tested without the patch. I hit the part where we send the 488 and got:
```
[tijmen][nua_i_message]: 200 OK
[WARN] [tijmen][nua_i_error]: 500 Responding to a Non-Existing Request
```
What can be observed here is that Sofia always sends a 200OK, even if the request is wrong.

I also tested with the modified version, a saw this behavior. Sofia always sends a 200OK. In the log I now only see:
`[tijmen][nua_i_message]: 200 OK`

So we don't end up with pending dialogs.
So it does not send out the event, which is expected due to the return

lminiero:
Thanks for double checking! Looks good to me then, merging :+1:

PR #2611, opened 2021-03-30
audiobridge: GList leak fixed

RSATom:
According to [docs](https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-get-values) `g_list_free()` should be called for `GList` returned by `g_hash_table_get_values`

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -6663,6 +6663,7 @@ static void *janus_audiobridge_mixer_thread(void *data) {
 				}
 				ps = ps->next;
 			}
+			g_list_free(anncs_list);
 		}
 #endif
 		/* Are we recording the mix? (only do it if there's someone in, though...) */


plugins/janus_audiobridge.c
```diff
@@ -6663,6 +6663,9 @@ static void *janus_audiobridge_mixer_thread(void *data) {
 				}
 				ps = ps->next;
 			}
+
+			g_list_free(anncs_list);
+			anncs_list = NULL;
```
lminiero, line 6668: No need to set the pointer to NULL, since it's a locally defined variable. Please also remove the extra unneeded empty line above.

lminiero (commented):
Good catch! We do indeed free it in `janus.c`, where we also use `g_hash_table_get_values` for a different reason. I just added a small editorial note below.

RSATom:
@lminiero Updated. And btw, I didn't find in docs if it valid to pass NULL value to `g_list_free()`, so maybe it's required add NULL check to https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_audiobridge.c#L6752

tmatth:
> @lminiero Updated. And btw, I didn't find in docs if it valid to pass NULL value to `g_list_free()`, so maybe it's required add NULL check to https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_audiobridge.c#L6752

It's ignored, https://github.com/GNOME/glib/blob/622b31f69e72fdb05b813820e9cd8087e88a47bf/glib/gslice.c#L1201

tmatth:
> @tmatth Yes, but it's implementation details...

Well if they ever change it they'll break a lot of people's code, but sure, a NULL check here doesn't hurt either.

lminiero:
There's no need for a NULL check, it's already in the `if(annc != NULL)` branch.

RSATom:
I was talking about `g_list_free(participants_list);` in the same file.30 мар. 2021 г. 23:10 пользователь Lorenzo Miniero ***@***.***> написал:
There's no need for a NULL check, it's already in the if(annc != NULL) branch.

lminiero:
I don't think we need to worry... anyway, thanks for the quick fix, merging! :+1:

PR #2616, opened 2021-04-01
SIP plugin: SIP MESSAGE out of dialog

ihusejnovic:
Hello,

In this pull request, I added the option to send a SIP MESSAGE "out of dialog". The message will be sent out of dialog if you pass the new parameter called "uri" with the SIP URI of the peer. If the "uri" parameter is not set, the message will be sent in the active dialog as it is in the current implementation. In order to send a message out of dialog, the user must be registered.

I tested it with the Kamailio and it works fine.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -406,14 +406,16 @@
  * and will restore the media direction that was set in the SDP before
  * putting the call on-hold.
  *
- * The \c message request allows you to send a SIP MESSAGE to the peer:
+ * The \c message request allows you to send a SIP MESSAGE to the peer.
+ * By default, it is sent in dialog, during active call.
+ * But, if the user is registered, it might be sent out of dialog also. In that case the uri parameter is required.
  *
 \verbatim
 {
 	"request" : "message",
 	"content_type" : "<content type; optional>"
 	"content" : "<text to send>",
- 	"uri" : "<SIP URI; optional>"
+ 	"uri" : "<SIP URI of the peer; optional; if set, the message will be sent out of dialog>"
 }
 \endverbatim
  *
@@ -4470,8 +4472,8 @@ static void *janus_sip_handler(void *data) {
 		} else if(!strcasecmp(request_text, "message")) {
 			/* Send a SIP MESSAGE request: we'll only need the content and optional payload type */
 			JANUS_VALIDATE_JSON_OBJECT(root, sipmessage_parameters,
-			      	error_code, error_cause, TRUE,
-			      	JANUS_SIP_ERROR_MISSING_ELEMENT, JANUS_SIP_ERROR_INVALID_ELEMENT);
+			      error_code, error_cause, TRUE,
+			      JANUS_SIP_ERROR_MISSING_ELEMENT, JANUS_SIP_ERROR_INVALID_ELEMENT);
 			if(error_code != 0) {
 				goto error;
 			}
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -415,7 +415,8 @@
 	"request" : "message",
 	"content_type" : "<content type; optional>"
 	"content" : "<text to send>",
- 	"uri" : "<SIP URI of the peer; optional; if set, the message will be sent out of dialog>"
+ 	"uri" : "<SIP URI of the peer; optional; if set, the message will be sent out of dialog>",
+ 	"headers" : "<array of key/value objects, to specify custom headers to add to the SIP MESSAGE; optional>"
 }
 \endverbatim
  *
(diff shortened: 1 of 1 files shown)


plugins/janus_sip.c
```diff
@@ -4467,39 +4471,68 @@ static void *janus_sip_handler(void *data) {
...
-				error_code = JANUS_SIP_ERROR_WRONG_STATE;
-				g_snprintf(error_cause, 512, "Wrong state (no callee?)");
-				goto error;
-			}
-			janus_mutex_unlock(&session->mutex);
 			JANUS_VALIDATE_JSON_OBJECT(root, sipmessage_parameters,
-				error_code, error_cause, TRUE,
-				JANUS_SIP_ERROR_MISSING_ELEMENT, JANUS_SIP_ERROR_INVALID_ELEMENT);
+			      	error_code, error_cause, TRUE,
+			      	JANUS_SIP_ERROR_MISSING_ELEMENT, JANUS_SIP_ERROR_INVALID_ELEMENT);
```
lminiero, line 4476: Spaces are used instead of tabs here.

plugins/janus_sip.c
```diff
@@ -4467,39 +4471,68 @@ static void *janus_sip_handler(void *data) {
...
-				SIPTAG_CONTENT_TYPE_STR(content_type),
-				SIPTAG_PAYLOAD_STR(msg_content),
-				TAG_END());
+			if(in_dialog_message) {
+				nua_message(session->stack->s_nh_i,
+					SIPTAG_CONTENT_TYPE_STR(content_type),
+					SIPTAG_PAYLOAD_STR(msg_content),
+					TAG_END());
+			} else {
+				nua_message(session->stack->s_nh_r,
```
lminiero, line 4530: I don't think this will work. First of all, this assumes a dialog for the REGISTER exists for that session, which will not be the case if someone registered as a guest (and maybe if it's a helper too? but I'd have to check). Besides, not sure if it's safe to re-use that dialog anyway. Maybe a new dialog should be created for the job, and torn down when it succeeds?

lminiero (commented):
Thanks! Added a couple of notes inline.

plugins/janus_sip.c
ihusejnovic: _"First of all, this assumes a dialog for the REGISTER exists for that session, which will not be the case if someone registered as a guest (and maybe if it's a helper too? but I'd have to check). "_
You are right. I will update PR.

 _"Maybe a new dialog should be created for the job, and torn down when it succeeds?"_
Do you mean a new nua_handle since SIP dialog is not created for SIP MESSAGE?
Not sure if we really need to destroy nua_handle when the message is sent or we can use the same nua_handle for all SIP MESSAGEs?

lminiero: > Not sure if we really need to destroy nua_handle when the message is sent or we can use the same nua_handle for all SIP MESSAGEs?

That might be an option too, and easier to clean up too :+1:

ihusejnovic: Thanks. I updated the pull request.

plugins/janus_sip.c
```diff
@@ -4521,16 +4522,35 @@ static void *janus_sip_handler(void *data) {
 				content_type = json_string_value(content_type_text);
 
 			const char *msg_content = json_string_value(json_object_get(root, "content"));
+			char custom_headers[2048];
+			janus_sip_parse_custom_headers(root, (char *)&custom_headers, sizeof(custom_headers));
```
lminiero, line 4526: Since you added custom headers support, the `sipmessage_parameters` validation should include that too, otherwise a malformed `headers` property could cause issues.

plugins/janus_sip.c
```diff
@@ -4521,16 +4522,35 @@ static void *janus_sip_handler(void *data) {
...
 			if(in_dialog_message) {
 				nua_message(session->stack->s_nh_i,
 					SIPTAG_CONTENT_TYPE_STR(content_type),
 					SIPTAG_PAYLOAD_STR(msg_content),
+					TAG_IF(strlen(custom_headers) > 0, SIPTAG_HEADER_STR(custom_headers)),
 					TAG_END());
 			} else {
-				nua_message(session->stack->s_nh_r,
+				if(session->stack->s_nh_m == NULL) {
+					janus_mutex_lock(&session->stack->smutex);
```
lminiero, line 4536: I'd put the if check and related actions within the mutex, and not before: otherwise you risk race conditions where two simultaneous `message` requests create different dialogs and one is lost (and leaked).

lminiero (commented):
Thanks! I've added a couple of notes inline.

plugins/janus_sip.c
ihusejnovic: Done.

plugins/janus_sip.c
ihusejnovic: Fixed.

ihusejnovic:
@lminiero I fixed all the things you suggested in the comments. Let me know if there is anything else that needs to be done. Thank you.

lminiero:
Thanks, looks good to me now! Merging :+1:

PR #2623, opened 2021-04-07
mqttevh: added ability use relative to config paths in config

RSATom:
Sometimes it's useful to have ability use relative paths in configs.

Also, if you will find it useful, I can do the same for other plugins.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -296,6 +296,16 @@ int janus_mkdir(const char *dir, mode_t mode) {
 	return 0;
 }
 
+gchar *janus_make_absolute_path(const gchar *base_dir, const gchar *path) {
+	if(!path)
+		return NULL;
+	if(g_path_is_absolute(path))
+		return g_strdup(path);
+	if(!base_dir)
+		return NULL;
+	return g_build_filename(base_dir, path, NULL);
+}
+
 int janus_get_codec_pt(const char *sdp, const char *codec) {
 	if(!sdp || !codec)
 		return -1;
--- a/utils.h
+++ b/utils.h
@@ -143,6 +143,13 @@ gboolean janus_flags_is_set(janus_flags *flags, gsize flag);
  * @note A failure may indicate that creating any of the subdirectories failed: some may still have been created */
 int janus_mkdir(const char *dir, mode_t mode);
 
+/*! \brief Helper to convert \c path relative to \c base_dir to absolute path.
+ *  If \c path already represents absolute path then just g_strdup it.
+ * @param[in] base_dir Path which will be prepended to \c path if it's relative
+ * @param[in] path Some relative or absolute path
+ * @returns g_strdup'ed absolute path. Should be freed with g_free() when no longer needed */
+gchar *janus_make_absolute_path(const gchar *base_dir, const gchar *path);
+
 /*! \brief Ugly and dirty helper to quickly get the payload type associated with a codec in an SDP
  * @param sdp The SDP to parse
  * @param codec The codec to look for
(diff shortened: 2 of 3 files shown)


utils.c
```diff
@@ -296,6 +296,20 @@ int janus_mkdir(const char *dir, mode_t mode) {
 	return 0;
 }
 
+gchar* janus_make_absolute_path(const gchar *base_dir, const gchar *path)
+{
```
lminiero, line 300: Code style: bracket should be on the same line as the function.
Besides, it should be `gchar *` and not `gchar* `.

utils.c
```diff
@@ -296,6 +296,20 @@ int janus_mkdir(const char *dir, mode_t mode) {
...
+	if(!path)
+		return NULL;
+
+	if(g_path_is_absolute(path))
+		return g_strdup(path);
+
+	if(!base_dir)
+		return NULL;
+
+	return g_build_filename(base_dir, path, NULL);
```
lminiero, line 310: For code style as well, I think you can remove all those extra empty lines to separate the checks.

utils.h
```diff
@@ -143,6 +143,13 @@ gboolean janus_flags_is_set(janus_flags *flags, gsize flag);
  * @note A failure may indicate that creating any of the subdirectories failed: some may still have been created */
 int janus_mkdir(const char *dir, mode_t mode);
 
+/*! \brief Helper to convert \a path relative to \a base_dir to absolute path.
+ *  If \a path already represents absolute path then just g_strdup it.
+ * @param[in] base_dir Path wich will be prepended to \a path if it's relative
```
lminiero, line 148: Not sure what `\a` is in Doxygen, but we don't use it anywhere: if it's meant to look like code, we use `\c` instead.

utils.h
```diff
@@ -143,6 +143,13 @@ gboolean janus_flags_is_set(janus_flags *flags, gsize flag);
...
+ * @param[in] path
+ * @returns g_strdup'ed absolute path. Should be freed with g_free() when no longer needed */
+gchar* janus_make_absolute_path(const gchar *base_dir, const gchar *path);
```
lminiero, line 151: See above on `gchar *`.

lminiero (commented):
Thanks! I do see how this could be useful in general, even though applying this on all plugins should probably wait until we merge multistream (I'm afraid this could generate many conflicts othewise).

I'm only wondering if this could cause issues in some edge cases, e.g., where the distinction between relative or absolute paths may be amniguous, or on different systems. But in general it looks fine to me, I just added some marginal notes on the project code style.

RSATom:
@lminiero thank you for review! I'll fix mentioned above issues today evening and will provide updated patches.

RSATom:
@lminiero  I've made requested changes. Please look.

utils.h
```diff
@@ -143,6 +143,13 @@ gboolean janus_flags_is_set(janus_flags *flags, gsize flag);
  * @note A failure may indicate that creating any of the subdirectories failed: some may still have been created */
 int janus_mkdir(const char *dir, mode_t mode);
 
+/*! \brief Helper to convert \c path relative to \c base_dir to absolute path.
+ *  If \c path already represents absolute path then just g_strdup it.
+ * @param[in] base_dir Path wich will be prepended to \c path if it's relative
+ * @param[in] path
```
lminiero, line 149: It looks like `path` is not documented?

lminiero (commented):
Looks good, thanks for the quick fixes! I only found a missing doxygen bit.

utils.h
RSATom: I've just decided `path` meaning is clear enough from other parts of description. But I can add something there if it's required.

lminiero: We try to document everything, so that would help. There's also a typo (wich) in the line above, by the way (sorry for missing it before!)

RSATom: Np, I'll extend it.

P.S. English is not my native language so I surprised there is only one typo there :smile:

RSATom:
I hope now it at least a little bit better...

lminiero:
Looks good, thanks! Merging :+1:

PR #2627, opened 2021-04-08
Fix streaming plugin RTSP sample in config template

lionelnicolas:
This make the example fit with in-template documentation :

https://github.com/meetecho/janus-gateway/blob/4dd379ab6952ccaaa027d5c150da1fbf0fecff16/conf/janus.plugin.streaming.jcfg.sample.in#L83-L90


lminiero:

Issue #2626, opened 2021-04-08
Crash when publish about 1100 streams in multiple rooms


winlinvip:
Janus [v0.10.10](https://github.com/meetecho/janus-gateway/releases/tag/v0.10.10) on CentOS7.

I create some rooms for benchmark by config `janus.plugin.videoroom.jcfg`  https://pastebin.com/qPkB5cxg

> Limit to about 1000 publishers for each room.

Run janus by:

```bash
cd /usr/local && gdb --args `pwd`/bin/janus --debug-level=3
```

> Reduce the detail log by level3.

Then I publish about 2k streams to room by [benchmark](https://github.com/ossrs/srs-bench/tree/feature/rtc#janus):

```bash
./objs/srs_bench -sfu=janus -pr  webrtc://xxx:8088/1000/livestream -sa a.ogg -sv v.h264 -fps 25 -sn 500 -delay 1200 > /dev/null 2>&1 &
./objs/srs_bench -sfu=janus -pr  webrtc://xxx:8088/2000/livestream -sa a.ogg -sv v.h264 -fps 25 -sn 500 -delay 1200 > /dev/null 2>&1 &
./objs/srs_bench -sfu=janus -pr  webrtc://xxx:8088/3000/livestream -sa a.ogg -sv v.h264 -fps 25 -sn 500 -delay 1200 > /dev/null 2>&1 &
./objs/srs_bench -sfu=janus -pr  webrtc://xxx:8088/4000/livestream -sa a.ogg -sv v.h264 -fps 25 -sn 500 -delay 1200 > /dev/null 2>&1 &
```

> We start about 500 publishers for each room(limit to 1000 publishers).

Use the command to show the threads of janus:

```bash
for((;;)); do 
    cat /proc/$(ps aux|grep janus|grep local|grep -v gdb|awk '{print $2}')/status|grep Threads
    sleep 1
done
```

I found that when threads is about 1000~1200, janus crashed, the logs https://pastebin.com/4BeVuR9b

tmatth:
From the line numbers it looks like this isn't the latest Janus, can you try with the latest HEAD? 

Otherwise I'm guessing it's crashing here:
https://github.com/meetecho/janus-gateway/blob/e674f1a808dbd75b762f79de1e1f11080d12e2fd/transports/janus_http.c#L1107

[...]

PR #2629, opened 2021-04-09
transports/http: avoid segfault on JSON error

tmatth:
Refs #2626

### DIFF
--- a/transports/janus_http.c
+++ b/transports/janus_http.c
@@ -1103,6 +1103,11 @@ int janus_http_send_message(janus_transport_session *transport, void *request_id
 		msg->timeout = NULL;
 		char *response_text = json_dumps(message, json_format);
 		json_decref(message);
+		if(!response_text) {
+			JANUS_LOG(LOG_ERR, "Invalid JSON message...\n");
+			janus_refcount_decrease(&msg->ref);
+			return -1;
+		}
 		msg->response = response_text;
 		msg->resplen = strlen(response_text);
 		MHD_resume_connection(msg->connection);


lminiero:
Thanks! This feels like a check that should be added to the other transports (and event handlers?) as well, in case, since they'd suffer from the same issue with overly large objects.

tmatth:
> Thanks! This feels like a check that should be added to the other transports (and event handlers?) as well, in case, since they'd suffer from the same issue with overly large objects.

Yeah that's why I marked as draft, if @winlinvip is able to confirm that this fixes the one crash there are probably a lot of other places where jansson's output needs to be checked for NULL, as unlike glib it won't abort on e.g. a failed allocation (or even invalid JSON).

lminiero:
I doubt we'll get feedback from the OP, and I think these are good checks to have: since they seem straightforward enough, I can take care of them all in a dedicated commit.

lminiero:
The above commit should address them all, so I'll close this PR. Thanks!

winlinvip:
Sorry, I'm a bit quit busy these days, I will check it now.

winlinvip:
I did some benchmarks, NO crash now, thanks~

* For 4CPU, although the load is very high, and some sessions were timeout, but no crash now.
* For 8CPU, the max threads can be about 1800, but lots of timeout and decrease to about 900 threads. No crash now.

It's possible to publish more streams to Janus than before. I got some data but I'm not sure it's ok to show it, but I think the perf and benchmark snapshot is OK, which might be useful for performance improvement.

The bellow is benchmark for 4CPUs, about 900 threads of Janus:

image

atoppi:
@winlinvip thanks for sharing your findings.
I have some observations:
1) could you please inspect the `janus_ice_cb_nice_recv` function with `annotate` to find the hot path ? We suspect that the very high usage is due to nack handling
2) `srtp_aes_encyot` as the second highest function makes me suspect that you are testing with a VM / srtp library without hw acceleration for AES-NI
3) also the high usage from `srtp_aes_encrypt` in a publishers-only test could mean that you are generating **A LOT** of RTCP traffic and that might be explained by **A LOT** of nacks sent by Janus (network loss? the bench tool struggling?)
4) could you share the remaining data that you're unsure to post here? Please drop an email to dev@meetecho.com

atoppi:
Also when expecting such a high load on a server, it would be better to tune the `event_loops` setting in `janus.jcfg` to 
**N x #CORES**
with N >= 1

winlinvip:
@atoppi Thanks for your help~

1. About the performance refine, I'm really sorry, because it takes a lot of time to test and analysis. I think it's better way, if you  use [srs-bench](https://github.com/ossrs/srs-bench/tree/feature/rtc#janus) to benchmark locally, or other tools. 
2. I will try `event_loops` to 8 and give some feedback.

winlinvip:
@atoppi It works! Thanks very much.

For 8 CPU server, janus is 779% and load is 10, and threads is 28. The data is much much better than before.

I will email you the compare data to dev@meetecho.com

Thanks~

PR #2634, opened 2021-04-12
Resolves meetecho/janus-gateway#2624 jansson double referencing

### LINKED ISSUE #2624: Memory loss
We have been running 0.10.10 version lately and noticed quite some mem loss over time:
[
mem
](https://user-images.githubusercontent.com/3499206/113829991-70464f80-9786-11eb-959b-79ab9c3dd974.png)
Running master in Valgrind showed me two places of loss, one initiating videoroom from file which was there before if i remember correctly:
```
==622494== 48 (24 direct, 24 indirect) bytes in 1 blocks are definitely lost in loss record 441 of 797
==622494==    at 0x4839809: malloc (vg_replace_malloc.c:307)
==622494==    by 0x4B4BBB8: g_malloc (in /usr/lib64/libglib-2.0.so.0.6600.8)
==622494==    by 0x4B63481: g_slice_alloc (in /usr/lib64/libglib-2.0.so.0.6600.8)
==622494==    by 0x4B38F37: g_list_append (in /usr/lib64/libglib-2.0.so.0.6600.8)
==622494==    by 0x4210FD: janus_config_get_categories (config.c:507)
==622494==    by 0x718EE51: janus_videoroom_init (janus_videoroom.c:2217)
==622494==    by 0x415077: main (janus.c:5307)
```
but this one is new on incoming requests:
```
==622494== 192 (80 direct, 112 indirect) bytes in 1 blocks are definitely lost in loss record 766 of 797
==622494==    at 0x4839809: malloc (vg_replace_malloc.c:307)
==622494==    by 0x4B4BBB8: g_malloc (in /usr/lib64/libglib-2.0.so.0.6600.8)
==622494==    by 0x451430: janus_session_create (janus.c:743)
==622494==    by 0x454E97: janus_process_incoming_request (janus.c:1075)
==622494==    by 0x45EADD: janus_transport_requests (janus.c:3377)
==622494==    by 0x4B71401: ??? (in /usr/lib64/libglib-2.0.so.0.6600.8)
==622494==    by 0x53923F8: start_thread (in /usr/lib64/libpthread-2.32.so)
==622494==    by 0x5549B52: clone (in /usr/lib64/libc-2.32.so)
```
Let me know if I need to provide more details.
Thanks up front

mirkobrankovic:


events/janus_gelfevh.c
```diff
@@ -589,21 +590,23 @@ static void *janus_gelfevh_handler(void *data) {
...
 
-			if(janus_gelfevh_send(json_dumps(output, json_format)) < 0) {
-				JANUS_LOG(LOG_WARN, "Couldn't send event to GELF, reconnect?, or event was null: %s\n",
-					json_dumps(output, json_format));
+			char *message = json_dumps(output, json_format);
+			if(janus_gelfevh_send(message) < 0) {
+				JANUS_LOG(LOG_WARN, "Couldn't send event to GELF, reconnect?, or event was null: %s\n", message);
 			}
 			json_decref(output);
+			g_free(message);
```
lminiero, line 608: You have to use `free` here, not `g_free`, because `json_dumps` doesn't use `g_malloc`.

mirkobrankovic: thanks :)

events/janus_gelfevh.c
```diff
@@ -589,25 +590,28 @@ static void *janus_gelfevh_handler(void *data) {
...
 
-			if(janus_gelfevh_send(json_dumps(output, json_format)) < 0) {
-				JANUS_LOG(LOG_WARN, "Couldn't send event to GELF, reconnect?, or event was null: %s\n",
-					json_dumps(output, json_format));
+			char *message = json_dumps(output, json_format);
+			if(janus_gelfevh_send(message) < 0) {
+				JANUS_LOG(LOG_WARN, "Couldn't send event to GELF, reconnect?, or event was null: %s\n", message);
 			}
 			json_decref(output);
+			free(message);
```
lminiero, line 608: Apologies for the late correction, but this should also have a check on whether `message` is `NULL`, as while `g_free` checks this automatically, I don't think `free` does. I'll have to check if we do that in other handles or transports, but if we don't we should.

tmatth: `free(NULL)` is safe and well-defined:
> The free function causes the space pointed to by ptr to be deallocated, that is, made available for further allocation. If ptr is a null pointer, no action occurs. [1]

[1] http://www.open-std.org/JTC1/SC22/wg14/www/docs/n1124.pdf

lminiero: I'm having deja-vu, which means I probably said the same silly thing in the past and you corrected me back then as well :smile:  Thanks for the clarification!

mirkobrankovic: hope this should be good to avoid double freeing , thanks

lminiero:
@mirkobrankovic this looks good to me now: if you can confirm you checked it in your own deployments and it's not leaking anymore, it's good to merge :+1:

events/janus_gelfevh.c
```diff
@@ -589,25 +590,30 @@ static void *janus_gelfevh_handler(void *data) {
...
-					json_dumps(output, json_format));
+			char *message = json_dumps(output, json_format);
+			if(janus_gelfevh_send(message) < 0) {
+				JANUS_LOG(LOG_WARN, "Couldn't send event to GELF, reconnect?, or event was null: %s\n", message);
 			}
 			json_decref(output);
+			if(message) {
+				free(message);
+				message = NULL;
+			}
```
tmatth, line 611: This can be simplified as just `free(message)` since as mentioned, `free(NULL)` is safe and also setting message to NULL is unnecessary since message is no longer used.

mirkobrankovic: great thanks @tmatth

lminiero: @mirkobrankovic I think @tmatth meant that the `if` part wasn't needed, since `free` works fine even if you pass `NULL` to it. So your original code was fine (and my mistake for asking you to change it).

tmatth: Yeah I meant the whole simplification you could make to be:
```
free(message);
```
instead of
```
if(message) {
     free(message);
     message = NULL;
}
```

lminiero: @mirkobrankovic as soon as you revert that I'll merge, thanks!

lminiero:
Thanks, merging!

PR #2637, opened 2021-04-16
Fix for race condition between VideoRoom publisher leaving and subscriber hanging up (fixes #2582)

### LINKED ISSUE #2582: Crash in the videoroom plugin
```
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
[Current thread is 1 (Thread 0x7f548e7fc700 (LWP 32457))]
(gdb) bt
#0  0x00007f550b1d27bb in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f550b1bd535 in __GI_abort () at abort.c:79
#2  0x00007f550bec2e0f in g_mutex_clear () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007f55040be9d9 in janus_videoroom_publisher_free (p_ref=0x7f54cc0038d8) at plugins/janus_videoroom.c:1761
#4  0x00007f55040cb5d9 in janus_videoroom_destroy_session (handle=<optimized out>, error=<optimized out>) at plugins/janus_videoroom.c:2731
#5  0x00005625e84af463 in janus_ice_outgoing_traffic_handle (handle=0x7f54f0005650, pkt=0x5625e8523f20 <janus_ice_detach_handle>) at ice.c:4290
#6  0x00005625e84b1ac4 in janus_ice_outgoing_traffic_dispatch (source=0x7f54f0005860, callback=<optimized out>, user_data=<optimized out>) at ice.c:411
#7  0x00007f550be78f2e in g_main_context_dispatch () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#8  0x00007f550be791c8 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#9  0x00007f550be794c2 in g_main_loop_run () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#10 0x00005625e84a33eb in janus_ice_handle_thread (data=0x7f54f0005650) at ice.c:1208
#11 0x00007f550bea1415 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#12 0x00007f550b363fa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#13 0x00007f550b2944cf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

Steps to reproduce
1. The patch reproduce.patch have to be applied to sources;
2. Create a video room with two participants;
3. Start video for them;
4. Close browser for the 2nd participant;
5. Quickly (<3 seconds) stop video for 1st participant;
6. Wait a little for the crash.

[reproduce.zip](https://github.com/meetecho/janus-gateway/files/6149109/reproduce.zip)
[videoroom.zip](https://github.com/meetecho/janus-gateway/files/6149121/videoroom.zip)

Provided patch fixes it. Also it fixes a crash in "switch" request.

lminiero:
As the title says, this is related to #2582, which is an issue @cppdev-1 providing a way to replicate it consistently. A patch was proposed in #2584 to fix this, but it introduced a new global mutex for what are specific resources, and so we didn't consider it acceptable.

This patch should provide a cleaner way of addressing the issue. The main problem was a race caused by changes in the `->feed` property of subscribers: this would cause a publisher reference to get to 0, while briefly still allowing another thread to read the publisher pointer and try to increase its reference to handle a subscription hangup. The result would be the publisher reference going to 0 (and so trigger the free code), while the subscriber would handle it as well (reference increased back to 1, but too late). As a consequence, the subscriber would lock `subscribers_mutex`, and the publisher free code would try to free that mutex at the same time. The patch now ensures `->feed` can only be modified when holding the session mutex: the same mutex is used to access the property when accessing the publisher when hanging up the subscriber.

I tested this briefly in a situation where I could replicate it consistently, and it seems to be working fine for me now. Please do test this more extensively, though, especially if you use the VideoRoom a lot. The sooner you can confirm it doesn't introduce regressions, the sooner I can merge.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -2795,7 +2795,7 @@ void janus_videoroom_destroy_session(janus_plugin_session *handle, int *error) {
 			janus_refcount_increase(&s->ref);
 		session->participant = NULL;
 		janus_mutex_unlock(&session->mutex);
-		if(s->room) {
+		if(s && s->room) {
 			janus_refcount_decrease(&s->room->ref);
 			janus_refcount_decrease(&s->ref);
 		}
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -5629,8 +5629,8 @@ void janus_videoroom_slow_link(janus_plugin_session *handle, int uplink, int vid
 				return;
 			}
 			if(g_atomic_int_get(&subscriber->destroyed)) {
-				janus_refcount_decrease(&session->ref);
 				janus_refcount_decrease(&subscriber->ref);
+				janus_refcount_decrease(&session->ref);
 				return;
 			}
 			/* Send an event on the handle to notify the application: it's
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -2650,6 +2650,24 @@ static janus_videoroom_publisher *janus_videoroom_session_get_publisher_nodebug(
 	return publisher;
 }
 
+static janus_videoroom_subscriber *janus_videoroom_session_get_subscriber(janus_videoroom_session *session) {
+	janus_mutex_lock(&session->mutex);
+	janus_videoroom_subscriber *subscriber = (janus_videoroom_subscriber *)session->participant;
+	if(subscriber)
+		janus_refcount_increase(&subscriber->ref);
+	janus_mutex_unlock(&session->mutex);
+	return subscriber;
+}
+
+static janus_videoroom_subscriber *janus_videoroom_session_get_subscriber_nodebug(janus_videoroom_session *session) {
+	janus_mutex_lock(&session->mutex);
+	janus_videoroom_subscriber *subscriber = (janus_videoroom_subscriber *)session->participant;
+	if(subscriber)
+		janus_refcount_increase_nodebug(&subscriber->ref);
+	janus_mutex_unlock(&session->mutex);
+	return subscriber;
+}
+
 static void janus_videoroom_notify_participants(janus_videoroom_publisher *participant, json_t *msg, gboolean notify_source_participant) {
 	/* participant->room->mutex has to be locked. */
 	if(participant->room == NULL)
@@ -2771,13 +2789,19 @@ void janus_videoroom_destroy_session(janus_plugin_session *handle, int *error) {
 		if(p)
 			janus_refcount_decrease(&p->ref);
 	} else if(session->participant_type == janus_videoroom_p_type_subscriber) {
+		janus_mutex_lock(&session->mutex);
 		janus_videoroom_subscriber *s = (janus_videoroom_subscriber *)session->participant;
+		if(s)
+			janus_refcount_increase(&s->ref);
 		session->participant = NULL;
+		janus_mutex_unlock(&session->mutex);
 		if(s->room) {
 			janus_refcount_decrease(&s->room->ref);
 			janus_refcount_decrease(&s->ref);
 		}
 		janus_videoroom_subscriber_destroy(s);
+		if(s)
+			janus_refcount_decrease(&s->ref);
 	}
 	janus_refcount_decrease(&session->ref);
 	return;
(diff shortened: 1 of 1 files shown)


plugins/janus_videoroom.c
```diff
@@ -2771,13 +2789,19 @@ void janus_videoroom_destroy_session(janus_plugin_session *handle, int *error) {
 		if(p)
 			janus_refcount_decrease(&p->ref);
 	} else if(session->participant_type == janus_videoroom_p_type_subscriber) {
+		janus_mutex_lock(&session->mutex);
 		janus_videoroom_subscriber *s = (janus_videoroom_subscriber *)session->participant;
+		if(s)
+			janus_refcount_increase(&s->ref);
 		session->participant = NULL;
+		janus_mutex_unlock(&session->mutex);
 		if(s->room) {
```
fancycode, line 2798: You're checking ` if(s)` in
https://github.com/meetecho/janus-gateway/pull/2637/files#diff-d96a677f8b8c31f15d44bf3a1d83934e5a6b6bdfab05bd70fd7be6119fb26740R2794
and
https://github.com/meetecho/janus-gateway/pull/2637/files#diff-d96a677f8b8c31f15d44bf3a1d83934e5a6b6bdfab05bd70fd7be6119fb26740R2803

Maybe also change this line to `if(s && s->room) {`?

plugins/janus_videoroom.c
```diff
@@ -5892,22 +5939,24 @@ static void *janus_videoroom_handler(void *data) {
...
 			}
 			if(subscriber->room == NULL) {
+				janus_refcount_decrease(&subscriber->ref);
 				janus_mutex_unlock(&sessions_mutex);
 				JANUS_LOG(LOG_ERR, "No such room\n");
 				error_code = JANUS_VIDEOROOM_ERROR_NO_SUCH_ROOM;
 				g_snprintf(error_cause, 512, "No such room");
 				goto error;
 			}
-			janus_refcount_increase(&subscriber->ref);
```
fancycode, line 5910: Won't this leak a reference to ` subscriber`  if `subscriber->room` is not `NULL`?

fancycode (commented):
Didn't run the test but found two potential issues from reading the code.

plugins/janus_videoroom.c
fancycode: Scratch that, I was misreading `increase` for `decrease` :-/

plugins/janus_videoroom.c
lminiero: Good point, since we only increase the ref if `s` exists in the first place :+1:

plugins/janus_videoroom.c
lminiero: :+1:

cppdev-1:
@lminiero I'd change the janus_refcount_decrease execution order in some situations. For example, I'd prefer
```
if(g_atomic_int_get(&subscriber->destroyed)) {
   janus_refcount_decrease(&subscriber->ref);
   return;
}
```
to

```
if(g_atomic_int_get(&subscriber->destroyed)) {
   janus_refcount_decrease(&subscriber->ref);
   return;
}
```

lminiero:
@cppdev-1 why would you change the order? We usually decrease in opposite order of hierarchy, which is why in the example you mentioned we decrease subscriber first and session next (a subscriber struct is owned by a session one, which is the root in that case).

cppdev-1:
@lminiero In your code session use counter is decreased first.

lminiero:
Ah sorry, I misread, good point then! I thought you wanted to change the first with the second, which confused me. I'll fix the code :+1:

cppdev-1:
@lminiero Another one is here
```
if(g_atomic_int_get(&publisher->destroyed)) {
				janus_refcount_decrease(&publisher->ref);
				return;
			}
```

lminiero:
Ack: I was looking at changes made in this PR, but that was code that was already there so I missed it. I'll fix that one too, please let me know if you find others.

lminiero:
Merging.

PR #2640, opened 2021-04-19
Remove support for framemarking RTP extension

lminiero:
We only supported the RTP framemarking extension because it was the way browsers implemented temporal layers for H.264 simulcasting. That feature has been removed from browsers a long time ago, though, so there's no point in keeping it in here anymore. I'm publishing this as a pull request, rather than committing right away, as it does change the signature of a simulcast preparation method in the core that most plugins use, so this should give you time to prepare for the change if you have a custom plugin that uses it.


lminiero:
Merging.

PR #2641, opened 2021-04-19
Added support for datachannel label/protocol to Lua and Duktape plugins

lminiero:
I realized that, despite labels and subprotocols support had been added to Janus for some time already (in #1551 and #2157 respectively), we still hadn't updated the Lua and Duktape plugins accordingly. This is what this patch does.

In theory, this change is backwards compatible, meaning that if you're using datachannels and don't change anything in your lua and js script, they should just work, but you may want to double check that. Label and subprotocol are now two additional (and again, in theory optional) parameters you can add both to your `incomingTextData`/`incomingBinaryData` callbacks, and the related `relayTextData`/`relayBinaryData` methods. I updated both `echotest.lua` and `echotest.js`, so it should be trivial to see how you can change your scripts.

Feedback welcome, as I plan to merge soon.

### DIFF
--- a/plugins/duktape/echotest.js
+++ b/plugins/duktape/echotest.js
@@ -173,16 +173,16 @@ function hangupMedia(id) {
 	}
 }
 
-function incomingTextData(id, buf, len) {
+function incomingTextData(id, buf, len, label, protocol) {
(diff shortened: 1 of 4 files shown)


lminiero:
Merging.

PR #2645, opened 2021-04-21
Send PLI when starting a paused stream

xadhoom:
We have noticed that sometimes resuming a paused stream takes long (even 10 seconds or more) before getting displayed on the subscriber. This depends on framerate / frame size (eg. on a nHD stream from a webcam with a somewhat high fps takes less than a desktop stream shared in FHD with maybe 3/5 fps), but also on the rtcp feedback from the browser that may not be immediate.

By sending a pli when starting the stream, the restart is more consistent.

This is a wip for the following reasons:
- makes sense to always send a pli request or is better to add a parameter to the start request?
- I'm not sure about the guard for the `janus_videoroom_reqpli` call, the one here is taken from the recording portion and seems ok
-  maybe is better to call it only if the subscriber is paused, so get's triggered only on stream resume?

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -6884,9 +6884,16 @@ static void *janus_videoroom_handler(void *data) {
 			} else if(!strcasecmp(request_text, "start")) {
 				/* Start/restart receiving the publisher streams */
 				if(subscriber->paused && msg->jsep == NULL) {
+					janus_videoroom_publisher *feed = subscriber->feed;
+
 					/* This is just resuming a paused stream, reset the RTP sequence numbers */
 					subscriber->context.a_seq_reset = TRUE;
 					subscriber->context.v_seq_reset = TRUE;
+
+					if(feed && feed->session && g_atomic_int_get(&feed->session->started)) {
+						/* Send a FIR */
+						janus_videoroom_reqpli(feed, "Subscriber start");
+					}
 				}
 				subscriber->paused = FALSE;
 				event = json_object();


januscla:
Thanks for your contribution, @xadhoom! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
I think you're modifying the wrong part of the code, if you want to resume a paused stream. The `start` request is used to send your SDP answer, as a subscriber, to the offer the plugin sent you, and so the negotiation is still ongoing. For startup, we already have a PLI when the PeerConnection goes up, in `janus_videoroom_setup_media`:

```
	janus_videoroom_reqpli(p, "New subscriber available");
```

There already is a PLI also for video subscriptions that had been paused via `configure` requests:

```
	janus_videoroom_reqpli(publisher, "Restoring video for subscriber");
```

lminiero:
Ok, I think I understand the point now: this is a `start` in response to a previous `pause`. Then yes, to answer your question above:

> * maybe is better to call it only if the subscriber is paused, so get's triggered only on stream resume?

I think this should only be done if the PC was up and previously paused, not always (otherwise it will try to always send a PLI during negotiation, before the PC even exists).

xadhoom:
> I think this should only be done if the PC was up and previously paused, not always (otherwise it will try to always send a PLI during negotiation, before the PC even exists).

Ok, moved inside the "paused" block, in order to be triggered only if the stream was previously paused.

plugins/janus_videoroom.c
```diff
@@ -6884,9 +6884,16 @@ static void *janus_videoroom_handler(void *data) {
...
+					janus_videoroom_publisher *feed = subscriber->feed;
+
 					/* This is just resuming a paused stream, reset the RTP sequence numbers */
 					subscriber->context.a_seq_reset = TRUE;
 					subscriber->context.v_seq_reset = TRUE;
+				
+					if(feed && feed->sdp && strstr(feed->sdp, "m=video")) {
+						/* Send a FIR */
+						janus_videoroom_reqpli(feed, "Subscriber start");
+					}
```
lminiero, line 6896: I don't think that works, if the purpose is only ensuring quicker recovery after a resume. It should be something like this instead:

```
if(subscriber->paused && feed) {
	/* Send a FIR */
	janus_videoroom_reqpli(feed, "Subscriber resumed");
}
```

The way the patch is conceived right now, it would _only_ work when first subscribing, which is exactly when the PLI should _not_ be sent (the PeerConnection doesn't exist yet).

lminiero: Ah, scratch that, I'm an idiot, I was looking at the wrong thing...

lminiero: That said, I think the if check can be simplified in

```
if(feed && g_atomic_int_get(&feed->started, 1)) {
```

as all we care about is if the PC is up. If there's no video, the PLI will be ignored by the core.

xadhoom: done

plugins/janus_videoroom.c
```diff
@@ -6882,12 +6882,18 @@ static void *janus_videoroom_handler(void *data) {
...
 				/* Start/restart receiving the publisher streams */
 				if(subscriber->paused && msg->jsep == NULL) {
 					/* This is just resuming a paused stream, reset the RTP sequence numbers */
 					subscriber->context.a_seq_reset = TRUE;
 					subscriber->context.v_seq_reset = TRUE;
 				}
+				if(feed && g_atomic_int_get(&feed->started, 1)) {
+						/* Send a FIR */
+						janus_videoroom_reqpli(feed, "Subscriber start");
+				}
```
lminiero, line 6896: This should be in the `if(subscriber->paused` above, though. Indentation should be fixed too.
As a side note, I'd move the `*feed` initialization right before the check, since it's unused if we're not in `subscriber->paused`.

xadhoom: ok, moved inside the paused if, along with the needed initialization and indentation

plugins/janus_videoroom.c
lminiero: Ah sorry, the atomic check is a typo, my fault, it should be

```
g_atomic_int_get(&feed->started)
```

The extra argument is only needed for set.

plugins/janus_videoroom.c
xadhoom: something is wrong into the `g_atomic_int_get(&feed->started, 1)`, investigating on it

xadhoom: > something is wrong into the `g_atomic_int_get(&feed->started, 1)`, investigating on it

ok, it was `if(feed && feed->session && g_atomic_int_get(&feed->session->started)) {`

plugins/janus_videoroom.c
xadhoom: fixed

plugins/janus_videoroom.c
xadhoom: fixed

lminiero:
Merging, thanks! :+1:

PR #2651, opened 2021-04-29
RabbitMQ Transport Reconnect Logic

chriswiggins:
Hey Janus team! 👋 

As we're running Janus and RabbitMQ in Kubernetes, there are times where our Janus instance will connect to one of our RabbitMQ nodes that then eventually goes down as we're doing maintenance on our hosts. Anyone with a clustered RabbitMQ setup with a load balancer in front will have the same thing happen to them.

Currently, the RabbitMQ Event handler does reconnects, but the Transport does not. This PR brings reconnect logic into the Transport.

It looks like a large change, but realistically most of it is just moving the large connection-based logic into a new function which we call at startup, and again when we notice the connection has dropped. It also brings in RabbitMQ heartbeat functionality.

I'm opening this as a draft to get initial feedback (and while I'm doing tweaks and testing) so any feedback would be awesome 👍

### DIFF
--- a/conf/janus.transport.rabbitmq.jcfg.sample
+++ b/conf/janus.transport.rabbitmq.jcfg.sample
@@ -34,6 +34,7 @@ general: {
 	#queue_durable = false				# Whether or not incoming queue should remain after a RabbitMQ reboot
 	#queue_autodelete = false			# Whether or not incoming queue should autodelete after janus disconnects from RabbitMQ
 	#queue_exclusive = false			# Whether or not incoming queue should only allow one subscriber
+	#heartbeat = 60 				# Defines the seconds without communication that should pass before considering the TCP connection has unreachable.
 
 	#ssl_enabled = false				# Whether ssl support must be enabled
 	#ssl_verify_peer = true				# Whether peer verification must be enabled
--- a/conf/janus.transport.rabbitmq.jcfg.sample
+++ b/conf/janus.transport.rabbitmq.jcfg.sample
@@ -34,7 +34,7 @@ general: {
 	#queue_durable = false				# Whether or not incoming queue should remain after a RabbitMQ reboot
 	#queue_autodelete = false			# Whether or not incoming queue should autodelete after janus disconnects from RabbitMQ
 	#queue_exclusive = false			# Whether or not incoming queue should only allow one subscriber
-	#heartbeat = 60 				# Defines the seconds without communication that should pass before considering the TCP connection has unreachable.
+	#heartbeat = 60 				# Defines the seconds without communication that should pass before considering the TCP connection unreachable.
 
 	#ssl_enabled = false				# Whether ssl support must be enabled
 	#ssl_verify_peer = true				# Whether peer verification must be enabled
--- a/transports/janus_rabbitmq.c
+++ b/transports/janus_rabbitmq.c
@@ -33,10 +33,18 @@
 
 #include "transport.h"
 
+/* Latest RabbitMQ-C library changes the library paths from 0.12.0.0 onwards */
+#ifdef HAVE_RABBITMQ_C_AMQP_H
 #include <rabbitmq-c/amqp.h>
 #include <rabbitmq-c/framing.h>
 #include <rabbitmq-c/tcp_socket.h>
 #include <rabbitmq-c/ssl_socket.h>
+#else
+#include <amqp.h>
+#include <amqp_framing.h>
+#include <amqp_tcp_socket.h>
+#include <amqp_ssl_socket.h>
+#endif
 
 #include "../debug.h"
 #include "../apierror.h"
--- a/transports/janus_rabbitmq.c
+++ b/transports/janus_rabbitmq.c
@@ -148,7 +148,7 @@ typedef struct janus_rabbitmq_client {
 	janus_mutex mutex;						/* Mutex to lock/unlock this session */
 	gint session_timeout:1;					/* Whether a Janus session timeout occurred in the core */
 	gint destroy:1;							/* Flag to trigger a lazy session destruction */
-	gint connected:1;							/* Flag to trigger a lazy session destruction */
+	gint connected:1;							/* Flag to specify whether or not RabbitMQ connection is deemed to be up */
 } janus_rabbitmq_client;
 
 /* RabbitMQ response */
--- a/transports/janus_rabbitmq.c
+++ b/transports/janus_rabbitmq.c
@@ -33,10 +33,10 @@
 
 #include "transport.h"
 
-#include <amqp.h>
-#include <amqp_framing.h>
-#include <amqp_tcp_socket.h>
-#include <amqp_ssl_socket.h>
+#include <rabbitmq-c/amqp.h>
+#include <rabbitmq-c/framing.h>
+#include <rabbitmq-c/tcp_socket.h>
+#include <rabbitmq-c/ssl_socket.h>
 
 #include "../debug.h"
 #include "../apierror.h"
--- a/transports/janus_rabbitmq.c
+++ b/transports/janus_rabbitmq.c
@@ -715,6 +715,8 @@ int janus_rabbitmq_connect(void) {
 
 	rmq_client->connected = 1;
 
+	JANUS_LOG(LOG_INFO, "RabbitMQ Connected successfully");
+
 	return 0;
 }
 
--- a/transports/janus_rabbitmq.c
+++ b/transports/janus_rabbitmq.c
@@ -81,7 +81,7 @@ void janus_rabbitmq_session_claimed(janus_transport_session *transport, guint64 
 json_t *janus_rabbitmq_query_transport(json_t *request);
 
 /* Internal methods */
-int janus_rabbitmq_connect(void);
+static int janus_rabbitmq_connect(void);
 
 /* Transport setup */
 static janus_transport janus_rabbitmq_transport =
@@ -181,13 +181,11 @@ static char *rmqhost = NULL, *vhost = NULL, *username = NULL, *password = NULL,
 	*ssl_cacert_file = NULL, *ssl_cert_file = NULL, *ssl_key_file = NULL,
 	*to_janus = NULL, *from_janus = NULL, *to_janus_admin = NULL, *from_janus_admin = NULL, *janus_exchange = NULL, *janus_exchange_type = NULL,
 	*queue_name = NULL, *queue_name_admin = NULL;
-
 static uint16_t rmqport = AMQP_PROTOCOL_PORT;
 static gboolean ssl_enabled = FALSE, ssl_verify_peer = FALSE, ssl_verify_hostname = FALSE;
 static gboolean declare_outgoing_queue = FALSE, declare_outgoing_queue_admin = FALSE;
-
-amqp_boolean_t queue_durable = 0, queue_exclusive = 0, queue_autodelete = 0, queue_durable_admin = 0, queue_exclusive_admin = 0, queue_autodelete_admin = 0;
-
+amqp_boolean_t queue_durable = 0, queue_exclusive = 0, queue_autodelete = 0,
+	queue_durable_admin = 0, queue_exclusive_admin = 0, queue_autodelete_admin = 0;
 static uint16_t heartbeat = 0;
 
 /* Transport implementation */
@@ -304,7 +302,7 @@ int janus_rabbitmq_init(janus_transport_callbacks *callback, const char *config_
 			ssl_verify_hostname = TRUE;
 	}
 
-	/* heartbeat config */
+	/* Heartbeat config */
 	item = janus_config_get(config, config_general, janus_config_type_item, "heartbeat");
 	if(item && item->value && janus_string_to_uint16(item->value, &heartbeat) < 0) {
 		JANUS_LOG(LOG_ERR, "Invalid heartbeat timeout (%s), falling back to default\n", item->value);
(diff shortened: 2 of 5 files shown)


chriswiggins:
Looks like the checks are failing with the updated RabbitMQ-C imports - I was getting a deprecation when compiling this and using master rabbitmq-c - I will see if I can do a conditional import on those

transports/janus_rabbitmq.c
```diff
@@ -684,6 +521,201 @@ int janus_rabbitmq_init(janus_transport_callbacks *callback, const char *config_
...
 
+int janus_rabbitmq_connect(void) {
+	rmq_client->connected = 0;
+	/* Connect */
+	rmq_client->rmq_conn = amqp_new_connection();
+	amqp_socket_t *socket = NULL;
+	amqp_queue_declare_ok_t *declare = NULL;
+	int status;
+	JANUS_LOG(LOG_VERB, "Creating RabbitMQ socket...\n");
+	if (ssl_enabled) {
```
lminiero, line 532: Code style: no space between `if` and the bracket (that was indeed there already in the existing code, but I may have missed it at the time, so we may as well fix it :grin: )

transports/janus_rabbitmq.c
```diff
@@ -900,8 +932,28 @@ void *janus_rmq_in_thread(void *data) {
 			if(res == AMQP_STATUS_TIMEOUT || res == AMQP_STATUS_SSL_ERROR)
 				continue;
 			JANUS_LOG(LOG_VERB, "Error on amqp_simple_wait_frame_noblock: %d (%s)\n", res, amqp_error_string2(res));
-			break;
+
+			rmq_client->connected = 0;
+
+			/* try and reconnect */
```
lminiero, line 938: See above on code style for comments.

transports/janus_rabbitmq.c
```diff
@@ -974,6 +1026,12 @@ void *janus_rmq_out_thread(void *data) {
 	}
 	JANUS_LOG(LOG_VERB, "Joining RabbitMQ out thread\n");
 	while(!rmq_client->destroy && !g_atomic_int_get(&stopping)) {
+
+		while (!rmq_client->connected) {
+			JANUS_LOG(LOG_WARN, "Waiting for 5s for RabbitMQ to reconnect\n");
+			g_usleep(5000000);
```
lminiero, line 1032: I'm not sure about this big sleep here... wouldn't it make sense to have some sort of exponential backoff, that starts with a much shorter interval instead?

lminiero (commented):
Thanks for the patch! I've added a few notes inline, mostly editorial. I'll let people using the RMQ transport in production get back to you on testing and feedback.

lminiero:
As a side note, please notice one of the CI builds failed:

```
transports/janus_rabbitmq.c:36:10: fatal error: 'rabbitmq-c/amqp.h' file not found
#include <rabbitmq-c/amqp.h>
```

lminiero:
> Looks like the checks are failing with the updated RabbitMQ-C imports - I was getting a deprecation when compiling this and using master rabbitmq-c - I will see if I can do a conditional import on those

Yep, sorry, I answered before seeing this message.

> Any pointers on how I can do that?

Not really, no: maybe some finer check in `configure.ac`?

atoppi:
@chriswiggins as of CI build failure, bear in mind that the test script installs the package `librabbitmq-dev` from `apt`, so maybe the available version is not the expected one.

https://packages.ubuntu.com/focal/librabbitmq-dev

transports/janus_rabbitmq.c
```diff
@@ -684,6 +521,201 @@ int janus_rabbitmq_init(janus_transport_callbacks *callback, const char *config_
...
+		if(ssl_verify_peer) {
+			amqp_ssl_socket_set_verify_peer(socket, 1);
+		} else {
+			amqp_ssl_socket_set_verify_peer(socket, 0);
+		}
+		if(ssl_verify_hostname) {
+			amqp_ssl_socket_set_verify_hostname(socket, 1);
+		} else {
+			amqp_ssl_socket_set_verify_hostname(socket, 0);
+		}
```
tmatth, line 547: Minor style point, these could be simplified to just:
```
amqp_ssl_socket_set_verify_peer(socket, ssl_verify_peer);
amqp_ssl_socket_set_verify_peer(socket, ssl_verify_hostname);
```
as gbooleans are just macros for 0 and 1:
https://developer.gnome.org/glib/stable/glib-Standard-Macros.html#TRUE:CAPS

chriswiggins:
I'm working on the reconnect logic now and over the weekend - more to follow next week. Thanks for the initial feedback!

chriswiggins:
Hey team - want to have a look at that now and see what you think?

transports/janus_rabbitmq.c
```diff
@@ -290,6 +305,13 @@ int janus_rabbitmq_init(janus_transport_callbacks *callback, const char *config_
 			ssl_verify_hostname = TRUE;
 	}
 
+	/* Heartbeat config */
+	item = janus_config_get(config, config_general, janus_config_type_item, "heartbeat");
+	if(item && item->value && janus_string_to_uint16(item->value, &heartbeat) < 0) {
+		JANUS_LOG(LOG_ERR, "Invalid heartbeat timeout (%s), falling back to default\n", item->value);
+		heartbeat = 0;
```
atoppi, line 312: So the default is `0`. Does it mean that heartbeat is being disabled ?

transports/janus_rabbitmq.c
```diff
@@ -945,8 +950,14 @@ void *janus_rmq_in_thread(void *data) {
...
 				int result = janus_rabbitmq_connect();
 				if(result < 0) {
-					JANUS_LOG(LOG_ERR, "Failed to reconnect to RabbitMQ Server. Retrying in 5s...\n");
-					g_usleep(5000000);
+					JANUS_LOG(LOG_WARN, "Failed to reconnect to RabbitMQ Server. Retrying in %fs...\n", (gfloat)rmq_reconnect_backoff/1000000);
+					g_usleep(rmq_reconnect_backoff);
+					rmq_reconnect_backoff *= rmq_reconnect_backoff_multiplier;
+					if (rmq_reconnect_backoff >= rmq_reconnect_backoff_max) {
+						rmq_reconnect_backoff = rmq_reconnect_backoff_max;
+					}
```
lminiero, line 958: Code style: no space between `if` and bracket. Besides, I think you can omit the curly brackets, as the action is simple.

transports/janus_rabbitmq.c
```diff
@@ -1024,13 +1035,19 @@ void *janus_rmq_out_thread(void *data) {
...
 	while(!rmq_client->destroy && !g_atomic_int_get(&stopping)) {
 
 		while (!rmq_client->connected) {
-			JANUS_LOG(LOG_WARN, "Waiting for 5s for RabbitMQ to reconnect\n");
-			g_usleep(5000000);
+			g_usleep(rmq_reconnect_backoff);
+			rmq_reconnect_backoff *= rmq_reconnect_backoff_multiplier;
+			if (rmq_reconnect_backoff >= rmq_reconnect_backoff_max) {
+				rmq_reconnect_backoff = rmq_reconnect_backoff_max;
+			}
```
lminiero, line 1046: See above on code style.

transports/janus_rabbitmq.c
```diff
@@ -973,7 +1035,19 @@ void *janus_rmq_out_thread(void *data) {
 		return NULL;
 	}
 	JANUS_LOG(LOG_VERB, "Joining RabbitMQ out thread\n");
+	guint rmq_reconnect_backoff = rmq_reconnect_backoff_initial;
 	while(!rmq_client->destroy && !g_atomic_int_get(&stopping)) {
+
+		while (!rmq_client->connected) {
```
atoppi, line 1041: the thread will be blocked forever in the loop when trying to destroy the transport

transports/janus_rabbitmq.c
```diff
@@ -1024,13 +1035,19 @@ void *janus_rmq_out_thread(void *data) {
 		return NULL;
 	}
 	JANUS_LOG(LOG_VERB, "Joining RabbitMQ out thread\n");
+	guint rmq_reconnect_backoff = rmq_reconnect_backoff_initial;
 	while(!rmq_client->destroy && !g_atomic_int_get(&stopping)) {
 
 		while (!rmq_client->connected) {
```
lminiero, line 1041: This should take into account whether the server is shutting down or not. The `while` outside does it, but the inner `while` doesn't, and so won't break if in between `g_usleep` calls the server has been closed. Maybe it's simpler to turn the inner `while` to an `if`, and do a `continue` after the sleep and the reconnection timer update, a bit like you do in `janus_rmq_in_thread`?

lminiero (commented):
Thanks, added a few notes inline :+1:

transports/janus_rabbitmq.c
```diff
@@ -900,8 +936,34 @@ void *janus_rmq_in_thread(void *data) {
...
 			JANUS_LOG(LOG_VERB, "Error on amqp_simple_wait_frame_noblock: %d (%s)\n", res, amqp_error_string2(res));
-			break;
+
+			rmq_client->connected = 0;
+
+			/* Try and reconnect */
+			if(rmq_client->rmq_conn && rmq_client->rmq_channel) {
+				amqp_channel_close(rmq_client->rmq_conn, rmq_client->rmq_channel, AMQP_REPLY_SUCCESS);
+				amqp_connection_close(rmq_client->rmq_conn, AMQP_REPLY_SUCCESS);
+				amqp_destroy_connection(rmq_client->rmq_conn);
```
atoppi, line 946: according to the docs, destroying the connection will also close the same connection and every channel

transports/janus_rabbitmq.c
```diff
@@ -891,6 +925,8 @@ void *janus_rmq_in_thread(void *data) {
 	timeout.tv_sec = 0;
 	timeout.tv_usec = 20000;
 	amqp_frame_t frame;
+	guint rmq_reconnect_backoff = rmq_reconnect_backoff_initial;
+
 	while(!rmq_client->destroy && !g_atomic_int_get(&stopping)) {
 		amqp_maybe_release_buffers(rmq_client->rmq_conn);
```
atoppi, line 931: what happens if connection is getting destroyed (either implicitly or explicitly) and then we re-use the `rmq_conn` ?

atoppi (commented):
thanks for the effort @chriswiggins , I've left some comments

transports/janus_rabbitmq.c
chriswiggins: Good point!

conf/janus.transport.rabbitmq.jcfg.sample
```diff
@@ -34,6 +34,7 @@ general: {
 	#queue_durable = false				# Whether or not incoming queue should remain after a RabbitMQ reboot
 	#queue_autodelete = false			# Whether or not incoming queue should autodelete after janus disconnects from RabbitMQ
 	#queue_exclusive = false			# Whether or not incoming queue should only allow one subscriber
+	#heartbeat = 60 				# Defines the seconds without communication that should pass before considering the TCP connection has unreachable.
```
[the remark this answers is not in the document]
chriswiggins: This was copied from the event handler, so I've fixed it there too

transports/janus_rabbitmq.c
chriswiggins: Yes - I've updated the log output

transports/janus_rabbitmq.c
chriswiggins: Fixed - I copied the logic from the event handler so I've updated the logic there too

transports/janus_rabbitmq.c
chriswiggins: I'm not sure I understand what you're asking here? Surely we have this covered with the destroy functionality?

atoppi: scratch that, I was worried that the same `rmq_conn` could be reused between retries, but I forgot that `janus_rabbitmq_connect` creates a new connection at every attempt

transports/janus_rabbitmq.c
```diff
@@ -900,8 +934,31 @@ void *janus_rmq_in_thread(void *data) {
 			if(res == AMQP_STATUS_TIMEOUT || res == AMQP_STATUS_SSL_ERROR)
 				continue;
 			JANUS_LOG(LOG_VERB, "Error on amqp_simple_wait_frame_noblock: %d (%s)\n", res, amqp_error_string2(res));
-			break;
+
+			rmq_client->connected = 0;
+
+			/* Try and reconnect */
+			if(rmq_client->rmq_conn && rmq_client->rmq_channel) {
```
atoppi, line 941: I don't think that checking `rmq_channel` is needed here

chriswiggins: 👍

lminiero:
@chriswiggins thanks for the quick fixes! This looks fine to me, so unless you wanted to do any additional changes, this is good to merge for me.

chriswiggins:
I've just done some extra testing and am now happy with the changes - I made a final couple of tweaks, mainly around the event handler logging so it's easier to see what relates to that vs the transport plugin. 

If you're happy with those then I'm happy for you to merge!

lminiero:
Ack, thanks! Just FYI, I'll probably change the "connected successfully" verbosity you added recently from LOG_INFO to LOG_VERB.

PR #2652, opened 2021-04-30
Bugfix: make sure that every destroyed plugin handle is 'detached'

SamyCookie:
Sometimes, when sessions are destroyed, the `detached` property is not correctly set in plugin handle.
This patch attempts to fix this bad behaviour

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -721,7 +721,6 @@ function Janus(gatewayCallbacks) {
 				// Don't warn here because destroyHandle causes this situation.
 				return;
 			}
-			pluginHandle.detached = true;
 			pluginHandle.ondetached();
 			pluginHandle.detach();
 		} else if(json["janus"] === "media") {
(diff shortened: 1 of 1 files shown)


lminiero:
That's not the proper fix. It assumes the client is always the one detaching the handle, which is not true: plugins can be detached programmatically by plugins, meaning users will receive a "detached" event without sending a detach themselves.

lminiero:
Edit: checking the code, though, `plugin.detach()` in the event will invoke `destroyHandle`, so we should be all set in that case too. Sorry forthe noise!

SamyCookie:
> Edit: checking the code, though, `plugin.detach()` in the event will invoke `destroyHandle`, so we should be all set in that case too. Sorry forthe noise!

Indeed, my patch extends the `detached` behaviour in the detach/destroy client side, but keep the previous behaviour in server side as well (via the event).

SamyCookie:
Also, I think that the `.ondetached()` callback associated to the `detached` property should be moved in `destroyHandle` at the same place, but it was not my bug...

lminiero:
Merging, thanks :+1:

SamyCookie:
the same ;) Thank you for merging this !

PR #2661, opened 2021-05-11
feat: support for SUBSCRIBE expiry (Expires header) in sip plugin

nicolasduteil:
This PR adds following behaviour

* define default `SUBSCRIBE` expiry to `3600` seconds (same as default `REGISTER` expiry)
* support for a new optional `subscribe_ttl` parameter in `subscribe` request (naming consistent with `register_ttl` used in `register` request)
* add an `expires` property to `subscribe_succeeded` event (if `Expires` header exists in SIP response)

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -122,8 +122,9 @@
 	"headers" : "<array of key/value objects, to specify custom headers to add to the SIP REGISTER; optional>",
 	"contact_params" : "<array of key/value objects, to specify custom Contact URI params to add to the SIP REGISTER; optional>",
 	"incoming_header_prefixes" : "<array of strings, to specify custom (non-standard) headers to read on incoming SIP events; optional>",
-	"refresh" : <true|false; if true, only uses the SIP REGISTER as an update and not a new registration; optional>",
-	"master_id" : <ID of an already registered account, if this is an helper for multiple calls (more on that later); optional>
+	"refresh" : "<true|false; if true, only uses the SIP REGISTER as an update and not a new registration; optional>",
+	"master_id" : "<ID of an already registered account, if this is an helper for multiple calls (more on that later); optional>",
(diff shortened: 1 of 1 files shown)


plugins/janus_sip.c
```diff
@@ -3162,6 +3164,17 @@ static void *janus_sip_handler(void *data) {
 				to = session->account.identity;
 			const char *event_type = json_string_value(json_object_get(root, "event"));
 			const char *accept = json_string_value(json_object_get(root, "accept"));
+
+			/* TTL */
+			int ttl = subscribe_ttl;
+			json_t *sub_ttl = json_object_get(root, "subscribe_ttl");
```
lminiero, line 3170: New parameters need to be added to the validation object, in this case `subscribe_parameters`.

plugins/janus_sip.c
```diff
@@ -5778,6 +5792,9 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 					json_t *headers = janus_sip_get_incoming_headers(sip, session);
 					json_object_set_new(result, "headers", headers);
 				}
+				if (sip->sip_expires) {
+					json_object_set_new(result, "expires", json_integer(sip->sip_expires->ex_delta));
+				}
```
lminiero, line 5797: No need for the curly brackets here, the action is simple.

lminiero (commented):
Thanks! I added a couple of comments.

nicolasduteil:
Should be good now. I previously wrongly assumed that third parameter of JANUS_VALIDATE_JSON_OBJECT was a default value, instead of an error code.

nicolasduteil:
I also noticed that `register_ttl` is not declared of `register_parameters`. I can include an update of `register_parameters` for consistency. What do you think  ?

nicolasduteil:
Done. I also updated the doc section to indicate support for `register_ttl` & `subscribe_ttl`

lminiero:
Looks good, thanks, merging!

PR #2662, opened 2021-05-12
janus-pp-rec: support HEVC AP(aggregation packet)

nu774:
Added support for HEVC Aggregation Packets (AP), described in rfc7798.

### DIFF
--- a/postprocessing/pp-h265.c
+++ b/postprocessing/pp-h265.c
@@ -338,6 +338,62 @@ int janus_pp_h265_preprocess(FILE *file, janus_pp_frame_packet *list) {
 		} else if(type == 34) {
 			/* PPS */
 			JANUS_LOG(LOG_HUGE, "[PPS] %u/%u/%u/%u\n", fbit, type, lid, tid);
+		} else if(type == 48) {
+			/* AP */
+			JANUS_LOG(LOG_HUGE, "[AP] %u/%u/%u/%u\n", fbit, type, lid, tid);
+			uint8_t *end = (uint8_t*)prebuffer + len;
+			uint16_t payload_len;
+			for (uint8_t *p = (uint8_t*)prebuffer + 2; p < end - 2; p += payload_len) {
+				payload_len = (p[0] << 8 | p[1]);
+				p += 2;
+
+				uint16_t unit = 0;
+				memcpy(&unit, p, sizeof(uint16_t));
+				unit = ntohs(unit);
+				uint8_t fbit = (unit & 0x8000) >> 15;
+				uint8_t type = (unit & 0x7E00) >> 9;
+				uint8_t lid = (unit & 0x01F8) >> 3;
+				uint8_t tid = (unit & 0x0007);
+
+				switch(type) {
+					case 32:
+						JANUS_LOG(LOG_HUGE, "[VPS] %u/%u/%u/%u\n", fbit, type, lid, tid);
+						break;
+					case 33:
+						JANUS_LOG(LOG_HUGE, "[SPS] %u/%u/%u/%u\n", fbit, type, lid, tid);
+						/* Get rid of the Emulation Prevention code, if present */
+						int i = 0, j = 0, zeros = 0;
+						for(i=0; i<payload_len; i++) {
+							if(zeros == 2 && p[i] == 0x03) {
+								/* Found, get rid of it */
+								i++;
+								zeros = 0;
+							}
+							/* Update the content of the buffer as we go along */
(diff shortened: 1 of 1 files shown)


postprocessing/pp-h265.c
```diff
@@ -338,6 +338,62 @@ int janus_pp_h265_preprocess(FILE *file, janus_pp_frame_packet *list) {
...
+				uint16_t unit = 0;
+				memcpy(&unit, p, sizeof(uint16_t));
+				unit = ntohs(unit);
+				uint8_t fbit = (unit & 0x8000) >> 15;
+				uint8_t type = (unit & 0x7E00) >> 9;
+				uint8_t lid = (unit & 0x01F8) >> 3;
+				uint8_t tid = (unit & 0x0007);
+
+				switch(type) {
+				case 32:
```
lminiero, line 359: Code style: `case` and `default` parts should all be indented, not at the same level as `switch` (the same applies to both the `switch` statements you added).

postprocessing/pp-h265.c
```diff
@@ -451,7 +507,37 @@ int janus_pp_h265_process(FILE *file, janus_pp_frame_packet *list, int *working)
 				memset(temp + 2, 0x01, 1);
 				frameLen += 3;
 			}
-			if(type == 49) {
+			else if(type == 48) {
```
lminiero, line 510: Code style: `else` should be on the same line as the previous closing bracket.

postprocessing/pp-h265.c
```diff
@@ -451,7 +507,37 @@ int janus_pp_h265_process(FILE *file, janus_pp_frame_packet *list, int *working)
...
+				for (uint8_t *p = buffer + 2; p < end - 2; p += payload_len) {
+					payload_len = (p[0] << 8 | p[1]);
+					p += 2;
+					uint8_t *temp = received_frame + frameLen;
+					temp[0] = 0;
+					temp[1] = 0;
+					temp[2] = 1;
+					memcpy(temp + 3, p, payload_len);
+					switch(p[0] >> 1) {
+					case 32: case 33: case 34:
```
lminiero, line 523: Code style: apart from indentation (see above), the three `case` should each have their own line.

postprocessing/pp-h265.c
```diff
@@ -451,7 +507,37 @@ int janus_pp_h265_process(FILE *file, janus_pp_frame_packet *list, int *working)
...
+						break;
+					}
+					frameLen += 3 + payload_len;
+				}
+				if(tmp->next == NULL || tmp->next->ts > tmp->ts)
+					break;
+				tmp = tmp->next;
+				continue;
+			}
+			else if(type == 49) {
```
lminiero, line 540: See above on `else` on the same line as the bracket.

lminiero (commented):
Thanks! Is there any specific reason for the patch? Mostly I'm interested in understanding if there are other WebRTC clients that do H.265 out there besides Safari which this patch is needed for. If so, I guess whether the `janus_h265_is_keyframe` helper in `utils.c` needs to be updated as well, as otherwise we won't detect keyframes in the cases you're extending here.

I just added a few notes inline on code style.

januscla:
Thanks for your contribution, @nu774! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

nu774:
Fixed code style.

> Is there any specific reason for the patch? Mostly I'm interested in understanding if there are other WebRTC clients that do H.265 out there besides Safari which this patch is needed for. If so, I wonder whether the janus_h265_is_keyframe helper in utils.c needs to be updated as well, as otherwise we won't detect keyframes in the cases you're extending here.

Well actually, I'm trying to push streams from ffmpeg via simple RTSP-WebRTC gateway built with https://github.com/pion/webrtc and https://github.com/aler9/gortsplib .
In case of H264 it seems to work, but in case of HEVC it doesn't.
I'm trying to investigate why, and I noticed I cannot check recorded bitstream with janus-pp-rec due to lack of HEVC AP support. That's why I created this patch.

In other word, it doesn't have something to do with an existing implementation.
AP support is needed in my case because ffmpeg generates APs when re-packetizing HEVC short NAL units for RTP transport.

IMHO, RTP senders don't necessarily have to implement APs (it just becomes more redundant without them), but receivers should be able to handle them.

lminiero:
Thanks for the clarification! I'll take note we may have to change the `janus_h265_is_keyframe` code in the future, but for now this looks good to merge for me :+1:

PR #2664, opened 2021-05-14
feat: support for custom call-id in subscribe request + add 'call_id' property to subscribe & notify related events

nicolasduteil:
This PR adds following behaviour to SIP plugin :
* allow to pass a custom `call_id` in `subscribe` request
* add `call_id` property to `subscribing`, `subscribe_succeeded` & `subscribe_failed` events
* add `call_id` property to `notify` event

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -474,6 +474,7 @@
 \verbatim
 {
 	"request" : "subscribe",
+	"call_id" : "<user-defined value of Call-ID SIP header used in all SIP requests throughout the subscription; optional>",
 	"event" : "<the event to subscribe to, e.g., 'message-summary'; mandatory>",
 	"accept" : "<what should be put in the Accept header; optional>",
 	"to" : "<who should be the SUBSCRIBE addressed to; optional, will use the user's identity if missing>",
@@ -488,6 +489,7 @@
 \verbatim
 {
 	"sip" : "event",
+	"call_id" : "<value of SIP Call-ID header for related subscription>",
 	"result" : {
 		"event" : "notify",
 		"notify" : "<name of the event that the user is subscribed to, e.g., 'message-summary'>",
@@ -756,7 +758,8 @@ static struct janus_json_parameter subscribe_parameters[] = {
 	{"to", JSON_STRING, 0},
 	{"event", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"accept", JSON_STRING, 0},
-	{"subscribe_ttl", JANUS_JSON_INTEGER, 0}
+	{"subscribe_ttl", JANUS_JSON_INTEGER, 0},
+	{"call_id", JANUS_JSON_STRING, 0}
 };
 static struct janus_json_parameter proxy_parameters[] = {
 	{"proxy", JSON_STRING, 0},
@@ -3179,6 +3182,12 @@ static void *janus_sip_handler(void *data) {
 			char ttl_text[20];
 			g_snprintf(ttl_text, sizeof(ttl_text), "%d", ttl);
 
+			/* Take call-id from request, if it exists */
+			char *callid = NULL;
+			json_t *request_callid = json_object_get(root, "call_id");
+			if (request_callid)
+				callid = json_string_value(request_callid);
+
 			/* Do we have a handle for this subscription already? */
 			janus_mutex_lock(&session->stack->smutex);
 			nua_handle_t *nh = NULL;
@@ -3225,6 +3234,7 @@ static void *janus_sip_handler(void *data) {
 			nua_subscribe(nh,
 				SIPTAG_TO_STR(to),
 				SIPTAG_EVENT_STR(event_type),
+				SIPTAG_CALL_ID_STR(callid),
 				SIPTAG_ACCEPT_STR(accept),
 				SIPTAG_EXPIRES_STR(ttl_text),
 				NUTAG_PROXY(session->helper && session->master ?
@@ -3232,6 +3242,8 @@ static void *janus_sip_handler(void *data) {
 				TAG_END());
 			result = json_object();
 			json_object_set_new(result, "event", json_string("subscribing"));
+			if (callid)
+				json_object_set_new(result, "call_id", json_string(callid));
 		} else if(!strcasecmp(request_text, "unsubscribe")) {
 			/* Unsubscribe from some SIP events */
 			JANUS_VALIDATE_JSON_OBJECT(root, subscribe_parameters,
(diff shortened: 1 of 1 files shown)


plugins/janus_sip.c
```diff
@@ -5252,6 +5264,7 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			/* Notify the application */
 			json_t *notify = json_object();
 			json_object_set_new(notify, "sip", json_string("event"));
+			json_object_set_new(notify, "call_id", json_string(sip->sip_call_id->i_id));
```
lminiero, line 5267: Is a call-ID always guaranteed to be there even for out of dialog SUBSCRIBE transactions? (probably yes, as part of the SIP standard, but as I rarely use SIP I do forget things sometimes!)

plugins/janus_sip.c
```diff
@@ -5860,6 +5874,7 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 				JANUS_LOG(LOG_WARN, "[%s] SUBSCRIBE failed: %d %s\n", session->account.username, status, phrase ? phrase : "");
 				json_t *event = json_object();
 				json_object_set_new(event, "sip", json_string("event"));
+				json_object_set_new(event, "call_id", json_string(sip->sip_call_id->i_id));
```
lminiero, line 5877: See above.

lminiero (commented):
Thanks, I added a few notes inline.

This also made me think we're probably not sending some messages via event handlers, as the call-id is what we use to identify who an outgoing message belongs to, and for some out-of-dialog transactions (subscribe, message, info) we don't know it: but this isn't something that needs to be solved here, so consider this me thinking out loud.

plugins/janus_sip.c
```diff
@@ -3225,13 +3234,16 @@ static void *janus_sip_handler(void *data) {
 			nua_subscribe(nh,
 				SIPTAG_TO_STR(to),
 				SIPTAG_EVENT_STR(event_type),
+				SIPTAG_CALL_ID_STR(callid),
```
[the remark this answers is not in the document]
nicolasduteil: As per [sofia-sip documentation](http://sofia-sip.sourceforge.net/refdocs/sip/group__sip__call__id.html#ga9f2290f1715815a68e50e009f6909426), it is possible to pass a `NULL` value 

> s pointer to a string containing Call-ID header value, or NULL.

plugins/janus_sip.c
nicolasduteil: It is mandatory in all SIP requests, as per [RFC3261](https://datatracker.ietf.org/doc/html/rfc3261)

> A valid SIP request formulated by a UAC MUST, at a minimum, contain
   the following header fields: To, From, CSeq, Call-ID, Max-Forwards,
   and Via; all of these header fields are mandatory in all SIP
   requests

I first thought about adding safe-guards, but discard the idea since header is supposed to be mandatory

lminiero: Yep, thought so, thanks for the clarification :+1: 
I was just wondering of whether Sofia SIP does indeed enforce those ckecks, e.g., if it's possible for it to receive a message without call-ID and make it available at the application level above.

nicolasduteil: That's what I'm trying to confirmed :grinning: by checking the code of sofia-sip-1.12.11.

nicolasduteil: [SIP parser method `sip_complete_message`](https://github.com/freeswitch/sofia-sip/blob/master/libsofia-sip-ua/sip/sip_parser.c#L657) specifically [checks for mandatory headers](https://github.com/freeswitch/sofia-sip/blob/master/libsofia-sip-ua/sip/sip_parser.c#L709)

nicolasduteil: This is also done in method [sip_sanity_check](http://sofia-sip.sourceforge.net/refdocs/sip/sip__util_8h.html#a82c4aab1711baed02335ddee2f36aad9) 

```
sip_sanity_check(sip_t const *sip)
{
  if (!sip ||
      !((sip->sip_request != NULL) ^ (sip->sip_status != NULL)) ||
      !sip->sip_to ||
      !sip->sip_from ||
      !sip->sip_call_id ||
      !sip->sip_cseq ||
      !sip->sip_via ||
      (sip->sip_flags & MSG_FLG_TRUNC))
    return -1;  /* Bad request */
```

nicolasduteil: `sip_sanity_check` is called in both [agent_recv_request](https://github.com/freeswitch/sofia-sip/blob/master/libsofia-sip-ua/nta/nta.c#L2949) & [agent_recv_response](https://github.com/freeswitch/sofia-sip/blob/master/libsofia-sip-ua/nta/nta.c#L3345)

lminiero: Ack, thanks for checking!

nicolasduteil:
> Thanks, I added a few notes inline.
> 
> This also made me think we're probably not sending some messages via event handlers, as the call-id is what we use to identify who an outgoing message belongs to, and for some out-of-dialog transactions (subscribe, message, info) we don't know it: but this isn't something that needs to be solved here, so consider this me thinking out loud.

Could you clarify what you meant by saying *this isn't something that needs to be solved here, so consider this me thinking out loud.* ? Is it related to support for custom *call-id* in the subscribe request ?

lminiero:
It's not something caused by your PR, but something I noticed only now. Sofia SIP doesn't have a way to intercept outgoing messages, which is something we need for event handlers (e.g., to notify monitoring frameworks like Homer about the SIP traffic we do); to overcome that, we basically hichjack the logging Sofia SIP does, to extract messages from there. Since we're parsing text, though, we need ways to figure out who among our sessions sent that message, which is something we do via call-IDs. This works for generic calls, but I think that since we started having separate dialogs for some messages, for those messages we may not be able to find the "owner". Your PR may make it easier to do for SUBSCRIBE, but I'll have to check: that's why it was mostly a memo for myself, not a note to you (apologies for the confusion).

nicolasduteil:
> It's not something caused by your PR, but something I noticed only now. Sofia SIP doesn't have a way to intercept outgoing messages, which is something we need for event handlers (e.g., to notify monitoring frameworks like Homer about the SIP traffic we do); to overcome that, we basically hichjack the logging Sofia SIP does, to extract messages from there. Since we're parsing text, though, we need ways to figure out who among our sessions sent that message, which is something we do via call-IDs. This works for generic calls, but I think that since we started having separate dialogs for some messages, for those messages we may not be able to find the "owner". Your PR may make it easier to do for SUBSCRIBE, but I'll have to check: that's why it was mostly a memo for myself, not a note to you (apologies for the confusion).

nicolasduteil:
Thanks for the explanation.

I also noticed that current implementation seems to support a single dialog per *event type*. While it's fine for MWI / voicemail subscription, it might prevent implementing [rfc4235](https://datatracker.ietf.org/doc/html/rfc4235) (event dialog) where you have a separate dialog for each extension which is being monitored. Anyway, it's outside of the scope of this PR anyway and might not even be in the scope of the project :grimacing:

lminiero:
Fair point but as you pointed out, out of scope here. Besides, it would require a complete refactoring of how subscriptions are implemented, as we use the event type as key for the dialogs, and anything more fine grained than that would require something way more complex (and to be honest, something that's very low priority for me).

lminiero:
Looks good to me, thanks again: merging! :+1:

Issue #2658, opened 2021-05-06
Output opus (optionally) in mka


lu-zero:
Since the preprocessor is already muxing to webm would be fairly simple to support this format as well.

If you are ok with adding this feature, I can send a patch myself to implement it.

lminiero:
Do you mean the postprocessor? Sure, please submit the patch as a pull request and we'll have a look, thanks!

lminiero:
As a side note, the Opus output currently uses libogg directly, and not ffmpeg, so the changes may not be trivial; the webm output is only there for VP8 and VP9.

PR #2665, opened 2021-05-15
Cleanup avformat-based preprocessors

lu-zero:
If you like the set I can convert the other modules likewise and further factor out the common parts.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -585,6 +585,7 @@ janus_pp_rec_SOURCES = \
 	postprocessing/pp-h264.h \
 	postprocessing/pp-av1.c \
 	postprocessing/pp-av1.h \
+	postprocessing/pp-avformat.h \
 	postprocessing/pp-h265.c \
 	postprocessing/pp-h265.h \
 	postprocessing/pp-opus.c \
--- a/Makefile.am
+++ b/Makefile.am
@@ -585,6 +585,7 @@ janus_pp_rec_SOURCES = \
 	postprocessing/pp-h264.h \
 	postprocessing/pp-av1.c \
 	postprocessing/pp-av1.h \
+	postprocessing/pp-avformat.c \
 	postprocessing/pp-avformat.h \
 	postprocessing/pp-h265.c \
 	postprocessing/pp-h265.h \
--- a/postprocessing/pp-av1.c
+++ b/postprocessing/pp-av1.c
@@ -386,6 +386,7 @@ int janus_pp_av1_process(FILE *file, janus_pp_frame_packet *list, int *working) 
 	int len = 0, frameLen = 0, total = 0, dataLen = 0;
 	int keyFrame = 0;
 	gboolean keyframe_found = FALSE;
+	AVPacket *packet = av_packet_alloc();
 
 	while(*working && tmp != NULL) {
 		keyFrame = 0;
@@ -506,21 +507,20 @@ int janus_pp_av1_process(FILE *file, janus_pp_frame_packet *list, int *working) 
 			total += frameLen;
 			JANUS_LOG(LOG_HUGE, "[%"SCNu64"] Saving frame: %d (tot=%d)\n", tmp->ts, frameLen, total);
 
-			AVPacket packet;
-			av_init_packet(&packet);
-			packet.stream_index = 0;
-			packet.data = received_frame;
-			packet.size = frameLen;
+			av_packet_unref(packet);
+			packet->stream_index = 0;
+			packet->data = received_frame;
+			packet->size = frameLen;
 			if(keyFrame)
-				packet.flags |= AV_PKT_FLAG_KEY;
+				packet->flags |= AV_PKT_FLAG_KEY;
 
 			/* First we save to the file... */
-			packet.dts = tmp->ts-list->ts;
-			packet.pts = tmp->ts-list->ts;
+			packet->dts = tmp->ts-list->ts;
+			packet->pts = tmp->ts-list->ts;
 			JANUS_LOG(LOG_HUGE, "%"SCNu64" - %"SCNu64" --> %"SCNu64"\n",
-				tmp->ts, list->ts, packet.pts);
+				tmp->ts, list->ts, packet->pts);
 			if(fctx) {
-				int res = av_write_frame(fctx, &packet);
+				int res = av_write_frame(fctx, packet);
 				if(res < 0) {
 					JANUS_LOG(LOG_ERR, "Error writing video frame to file... (error %d)\n", res);
 				}
@@ -528,6 +528,7 @@ int janus_pp_av1_process(FILE *file, janus_pp_frame_packet *list, int *working) 
 		}
 		tmp = tmp->next;
 	}
+	av_packet_free(&packet);
 	g_free(received_frame);
 	g_free(obu_data);
 	g_free(start);
--- a/postprocessing/pp-av1.c
+++ b/postprocessing/pp-av1.c
@@ -19,35 +19,9 @@
 #include <string.h>
 #include <stdlib.h>
 
-#include <libavcodec/avcodec.h>
-#include <libavformat/avformat.h>
-
+#include "pp-avformat.h"
 #include "pp-av1.h"
 #include "../debug.h"
-
-
-#define LIBAVCODEC_VER_AT_LEAST(major, minor) \
-	(LIBAVCODEC_VERSION_MAJOR > major || \
-	 (LIBAVCODEC_VERSION_MAJOR == major && \
-	  LIBAVCODEC_VERSION_MINOR >= minor))
-
-#if LIBAVCODEC_VER_AT_LEAST(51, 42)
-#define PIX_FMT_YUV420P AV_PIX_FMT_YUV420P
-#endif
-
-#if LIBAVCODEC_VER_AT_LEAST(56, 56)
-#ifndef CODEC_FLAG_GLOBAL_HEADER
-#define CODEC_FLAG_GLOBAL_HEADER AV_CODEC_FLAG_GLOBAL_HEADER
-#endif
-#ifndef FF_INPUT_BUFFER_PADDING_SIZE
-#define FF_INPUT_BUFFER_PADDING_SIZE AV_INPUT_BUFFER_PADDING_SIZE
-#endif
-#endif
-
-#if LIBAVCODEC_VER_AT_LEAST(57, 14)
-#define USE_CODECPAR
-#endif
-
 
 /* MP4 output */
 static AVFormatContext *fctx;
@@ -66,17 +40,7 @@ int janus_pp_av1_create(char *destination, char *metadata, gboolean faststart) {
 	JANUS_LOG(LOG_ERR, "This version of libavcodec doesn't support AV1...\n");
 	return -1;
 #endif
-	/* Setup FFmpeg */
-#if ( LIBAVFORMAT_VERSION_INT < AV_VERSION_INT(58,9,100) )
-	av_register_all();
-#endif
-	/* Adjust logging to match the postprocessor's */
-	av_log_set_level(janus_log_level <= LOG_NONE ? AV_LOG_QUIET :
-		(janus_log_level == LOG_FATAL ? AV_LOG_FATAL :
-			(janus_log_level == LOG_ERR ? AV_LOG_ERROR :
-				(janus_log_level == LOG_WARN ? AV_LOG_WARNING :
-					(janus_log_level == LOG_INFO ? AV_LOG_INFO :
-						(janus_log_level == LOG_VERB ? AV_LOG_VERBOSE : AV_LOG_DEBUG))))));
+	janus_pp_setup_avformat();
 	/* MP4 output */
 	fctx = avformat_alloc_context();
 	if(fctx == NULL) {
--- a/postprocessing/pp-av1.c
+++ b/postprocessing/pp-av1.c
@@ -26,9 +26,6 @@
 /* MP4 output */
 static AVFormatContext *fctx;
 static AVStream *vStream;
-#ifdef USE_CODECPAR
-static AVCodecContext *vEncoder;
-#endif
 static uint16_t max_width = 0, max_height = 0;
 int fps = 0;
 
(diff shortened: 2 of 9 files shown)


januscla:
Thanks for your contribution, @lu-zero! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks for the contribution @lu-zero! I definitely see value in the more compact approach you're suggesting, since we had indeed many similar sections that could be streamlined.

I only have one doubt related to the changes you made on AVPacket management, probably related to my limited knowledge of its memory management internals: I see the packet is allocated once before the loop is started, and then at each loop there's a `av_packet_unref` call that precedes its usage. Won't that first unref actually get rid of the packet before we get a chance of using it? First iteration in the loop aside, it's also not clear whether `av_write_frame` will add a reference to the packet when using it, as the documentation says it does not take ownership.

lu-zero:
`packet_unref()` does the same work as the former `init_packet()`. Since we do not use av_buffer, there is no actual refcounting on the payload.

```
void av_packet_unref(AVPacket *pkt)
{
    av_packet_free_side_data(pkt); // NO-OP we do not have side data
    av_buffer_unref(&pkt->buf); // NO-OP we do not use the buffer
    av_init_packet(pkt); // same as the deprecated function
    pkt->data = NULL; // basic sanitization
    pkt->size = 0;
}
```

lminiero:
Ack, thanks for the clarification :+1:

lu-zero:
@lminiero do you have test samples I can use to make sure everything works after the changes? I have samples for opus and vp8 but not for the others.

lminiero:
@lu-zero I don't have any recording in the other formats, but they should be quite easy to generate. If you have a Janus instance you control somewhere, you can open the echotest demo with a custom codec preference, e.g.:

    https://.../echotest.html?vcodec=h264

to force a specific video codec to be used (`acodec` does the same for audio), and then on the JS console send a `record: true` command to start recording the session:

    echotest.send({ message: { record: true }});

This will create a different mjr file for each medium (in EchoTest audio, video and data). Sending `record: false` or stopping the demo will finalize the files, and allow you to use them.

lminiero:
@lu-zero please let me know when you think the refactoring is done, so that I know when I can do a review :+1: 
I'm currently working on a separate branch that, due to some new features, will touch the way some of the packets are processed in the pprec tool as well, so I'll also need to know what I'll have to take into account in terms of refactoring there.

lu-zero:
Do you have a known use for the silence packets in opus? I just commented it out since locally it seems to cause more problems that it solve. If you are happy with what's in there I'd say it could land.

lminiero:
> Do you have a known use for the silence packets in opus? I just commented it out since locally it seems to cause more problems that it solves. If you are happy with what's in there I'd say it could land.

Actually yes, that's quite important. When we detect a packet loss, we have to inject a silence packet, otherwise the duration of the processed stream will be broken. That's because, at least from my understanding of libogg, packets can only be added to the format with monotonically increasing packet IDs, and so skipping missing packets reduces the duration of the stream (as the 20ms that were missed are not reflected in the audio recording). That might not be important in case of very few losses, but if there's many of them, this could lead to problems, like out-of-sync audio and video streams if they need to be combined ex-post.

Makefile.am
```diff
@@ -585,6 +585,8 @@ janus_pp_rec_SOURCES = \
 	postprocessing/pp-h264.h \
 	postprocessing/pp-av1.c \
 	postprocessing/pp-av1.h \
+	postprocessing/pp-avformat.h \
+	postprocessing/pp-avformat.c \
```
lminiero, line 589: Just a small nit, can you invert the order of .c and .h file here to match the others in the list?

postprocessing/pp-av1.c
```diff
@@ -65,98 +36,30 @@ int janus_pp_av1_create(char *destination, char *metadata, gboolean faststart) {
...
-	if(avcodec_open2(vEncoder, codec, NULL) < 0) {
-		/* Error opening video codec */
-		JANUS_LOG(LOG_ERR, "Encoder error\n");
-		return -1;
-	}
-	avcodec_parameters_from_context(vStream->codecpar, vEncoder);
-#else
-	vStream = avformat_new_stream(fctx, 0);
+
+	vStream = janus_pp_new_video_avstream(fctx, AV_CODEC_ID_AV1, max_width, max_height);;
```
lminiero, line 47: Small nit: there's a double semicolon at the end.

postprocessing/pp-avformat.c
```diff
@@ -0,0 +1,92 @@
...
+AVFormatContext *janus_pp_create_avformatcontext(const char *format, const char *metadata, const char *destination) {
+	janus_pp_setup_avformat();
+
+	AVFormatContext *ctx = avformat_alloc_context();
+	if(!ctx)
+		return NULL;
+
+	/* We save the metadata part as a comment (see #1189) */
+        if(metadata)
+                av_dict_set(&ctx->metadata, "comment", metadata, 0);
```
lminiero, line 34: Broken indentation in the above lines: per the project code style, we use tabs, not spaces.

postprocessing/pp-avformat.c
```diff
@@ -0,0 +1,92 @@
...
+
+	ctx->oformat = av_guess_format(format, NULL, NULL);
+        if(ctx->oformat == NULL) {
+                JANUS_LOG(LOG_ERR, "Error guessing format\n");
+		avformat_free_context(ctx);
+                return NULL;
+        }
```
lminiero, line 41: Same indentation issues for the above lines.

postprocessing/pp-opus.c
```diff
@@ -19,48 +19,42 @@
...
-ogg_packet *op_from_pkt(const unsigned char *pkt, int len);
-void op_free(ogg_packet *op);
-int ogg_write(void);
-int ogg_flush(void);
+int janus_pp_opus_create(char *destination, char *metadata) {
+	if(destination == NULL)
+		return -1;
 
+	/* WebM output */
+	fctx = janus_pp_create_avformatcontext("ogg", metadata, destination);
```
lminiero, line 42: I guess that if we wanted to optionally allow people to output the processing to `.mka`, as you said in #2658, we'd only need to make the "ogg" part a configurable string where we can put a different format target, correct? Same for videos, I guess, where VP8/VP9 could be outputted to `.mkv` rather than `.webm` as we do now.

postprocessing/pp-opus.c
```diff
@@ -128,176 +129,29 @@ int janus_pp_opus_process(FILE *file, janus_pp_frame_packet *list, int *working)
...
 			pos, bytes, tmp->len, tmp->seq, diff, tmp->ts, (tmp->ts-list->ts)/48000);
-		op->granulepos = 960*(pos); /* FIXME: get this from the toc byte */
-		ogg_stream_packetin(stream, op);
-		g_free(op);
-		ogg_write();
-		ogg_flush();
+		av_packet_unref(pkt);
+		pkt->stream_index = 0;
+		pkt->data = buffer;
+                pkt->size = bytes;
```
lminiero, line 137: Broken indentation (spaces).

postprocessing/pp-opus.c
```diff
@@ -128,176 +129,29 @@ int janus_pp_opus_process(FILE *file, janus_pp_frame_packet *list, int *working)
...
+		pkt->pts = pkt->dts = av_rescale_q(tmp->ts - list->ts, timebase, fctx->streams[0]->time_base);
```
lminiero, line 138: Is this maybe why you were getting issues with the silence packets being inserted manually as we did when using libogg directly? Does the ogg format target in ffmpeg automatically insert silence audio packets looking at what the pts is, or is still increased monotonically even in case of gaps?

postprocessing/pp-webm.c
```diff
@@ -75,92 +47,30 @@ int janus_pp_webm_create(char *destination, char *metadata, gboolean vp8) {
...
-#endif
+
+	int codec_id;
 #if LIBAVCODEC_VER_AT_LEAST(54, 25)
 	#if LIBAVCODEC_VERSION_MAJOR >= 55
-	vStream->codec->codec_id = vp8 ? AV_CODEC_ID_VP8 : AV_CODEC_ID_VP9;
+		codec_id = vp8 ? AV_CODEC_ID_VP8 : AV_CODEC_ID_VP9;
 	#else
-	vStream->codec->codec_id = AV_CODEC_ID_VP8;
+		codec_id = AV_CODEC_ID_VP8;
```
lminiero, line 62: The two `codec_id` lines don't need to be indented (the one below isn't, for instance), as macro `ifs` for us are not like "regular" ifs.

lminiero (commented):
Just had a look at the code, and it looks like a really useful refactoring, thanks! I just added a few minor editorial comments inline (mostly related to the project code style), and a note on the Opus silence ingestion we discussed a few days ago. I haven't tested this yet, but plan to do it soon.

postprocessing/pp-opus.c
lu-zero: That's the plan, ideally should be easy to select the output format (mp4 or mkv or webm).

I'll try to make the postprocessor able to mux both audio and video in a single pass later probably. One step at time first though :)

postprocessing/pp-opus.c
lu-zero: Looks like in one of the machines I tested I do not have editorconfig support. Do you happen to have a clang-format setup I can use?

postprocessing/pp-opus.c
lu-zero: The sample I used to test has the audio lasting much longer than the video, I thought the problem was in the audio, but probably it is the video having problems.

I still need to get a local setup to get better samples.

Inserting the silence seems fine with both ogg and mkv output.

postprocessing/pp-opus.c
lminiero: > I'll try to make the postprocessor able to mux both audio and video in a single pass later probably. One step at time first though :)

That's actually not needed, since our `.mjr` files only store one media stream anyway (audio, video or data). Any muxing after that should be done separately, e.g., using ffmpeg itself, and possibly using timing info from the `.mjr` header to sort out sync issues.

postprocessing/pp-opus.c
lminiero: > Inserting the silence seems fine with both ogg and mkv output.

Ack, thanks!

> I still need to get a local setup to get better samples.

There are a couple of old .mjr files you can play with in [this folder of the repo](https://github.com/meetecho/janus-gateway/tree/master/plugins/recordings): they use the old format, but should be ok if you want to test audio and video streams that have the same length.

lu-zero:
The formatting problems should be addressed now, I hope :)

postprocessing/pp-opus.c
```diff
@@ -19,48 +19,42 @@
...
+static AVStream *vStream;
 
-/* OGG/Opus helpers */
-FILE *ogg_file = NULL;
-ogg_stream_state *stream = NULL;
+static const uint8_t opus_extradata[19] = {
+    'O', 'p', 'u', 's', 'H', 'e', 'a', 'd',
+    1, 2, 56, 1, 128, 187,
+    0, 0, 0, 0, 0,
+};
```
lminiero, line 35: Whoops looks like I missed this space-indentation in the previous review, sorry...

lminiero:
> The formatting problems should be addressed now, I hope :)

Thanks! I just found one I missed before (I added a comment above), but the code changes do look fine to me. I plan to make some tests either later today or tomorrow morning, and in case they work fine for me it will be good to merge. Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus) before we get to that, though, as we won't be able to merge otherwise.

lu-zero:
I think I already signed the CLA, it does not show on your side?

lminiero:
Mh no, it's not there. Did you fill the form at the end of the page? Maybe you only did the Github authentication on top? (the page is indeed a bit confusing in that regard)

lu-zero:
I completed the form at least now.

lminiero:
Thanks, I do see it now! :v:

postprocessing/pp-g722.c
```diff
@@ -192,18 +170,17 @@ int janus_pp_g722_process(FILE *file, janus_pp_frame_packet *list, int *working)
 		JANUS_LOG(LOG_VERB, "Writing %d bytes out of %d (seq=%"SCNu16", step=%"SCNu16", ts=%"SCNu64", time=%"SCNu64"s)\n",
 			bytes, tmp->len, tmp->seq, diff, tmp->ts, (tmp->ts-list->ts)/8000);
 		/* Decode and save to wav */
-		AVPacket avpacket;
-		av_init_packet(&avpacket);
-		avpacket.data = (uint8_t *)buffer;
-		avpacket.size = bytes;
+		AVPacket *avpacket = av_packet_alloc();
```
lminiero, line 173: @lu-zero any reason why this allocation is inside the loop? I see for all other codecs you just do it once before the `while`, and just unref/initialize the packet at each loop, while for G.722 this is happening in the `while` instead. Is it because we're actually decoding something here, rather than just saving to a container?

lu-zero:
Yes, it felt simpler that way.

postprocessing/pp-opus.c
```diff
@@ -72,30 +66,37 @@ int janus_pp_opus_process(FILE *file, janus_pp_frame_packet *list, int *working)
...
 			int i=0;
 			for(i=0; i<silence_count; i++) {
-				pos = (tmp->prev->ts - list->ts) / 48 / 20 + i + 1;
+				pos = tmp->prev->ts - list->ts;
 				if(tmp->next != NULL)
-					nextPos = (tmp->next->ts - list->ts) / 48 / 20;
+					nextPos = tmp->next->ts - list->ts;
 				if(pos >= nextPos) {
-					JANUS_LOG(LOG_WARN, "[SKIP] pos: %06" SCNu64 ", skipping remaining silence\n", pos);
+					JANUS_LOG(LOG_WARN, "[SKIP] pos: %06" SCNu64 ", skipping remaining silence\n", pos / 48 / 20 + 1);
```
lminiero, line 86: Mh looking at these changes again (with fresh eyes from @atoppi) there seems to be something broken in the way `pos` is updated and handled now. I see how you decided to normalize both `pos` and `nextPos`, but before `pos` depended on `i`, while now it doesn't: since `tmp` never changes in the `for` loop, neither will the value of `pos`, making the checks broken. Considering `pos` is used as a basis for the `av_rescale_q` below, this would likely not insert all the silences we need, in case multiple packets were lost in a row.

lu-zero: Oh, right. Updated to increase the pos by `20 * 48`.

atoppi:
hi @lu-zero ,
gave this PR a run with an opus recording sample ([attached](https://github.com/meetecho/janus-gateway/files/6539716/rename_to_audio.mjr.txt)).
This specific sample has many 1-packet gaps and a single large gap of consecutive packets.
The new post-processor correctly filled all the gaps, so everything looks fine.

Anyway I have observed some differences in the `ffprobe` output:
```
Input #0, ogg, from 'old-pp.opus':
  Duration: 00:00:34.98, start: 0.000000, bitrate: 39 kb/s
    Stream #0:0: Audio: opus, 48000 Hz, stereo, fltp
    Metadata:
      comment         : {"t": "a", "c": "opus", "s": 1621947438860716, "u": 1621947438863033}
```
```
Input #0, ogg, from 'new-pp.opus':
  Duration: 00:00:34.96, start: -0.020000, bitrate: 30 kb/s
    Stream #0:0: Audio: opus, 48000 Hz, stereo, fltp
    Metadata:
      comment         : {"t": "a", "c": "opus", "s": 1621947438860716, "u": 1621947438863033}
      encoder         : Lavf58.29.100
```

As you can see the opus file from the PR (`new-pp.opus`):
1) has a shorter duration (20 ms) but a negative start time that seems to "compensate"
2) has a lower bitrate (30kbps vs 39kbps), indeed the file size is smaller (130K vs 171K)

What is the reason of those differences?

lu-zero:
lavf adds the [pre-skip](https://datatracker.ietf.org/doc/html/rfc7845#section-4.2), and has a different muxing strategy apparently, the same amount of packets is stored. 

`ffprobe -show_packets` shows that all the packets are stored that, I can see using vimdiff.

atoppi:
Thanks for clarifying, according to the rfc:
```
A 'pre-skip' field in the ID header (see Section 5.1) signals the
number of samples that SHOULD be skipped (decoded but discarded) at
the beginning of the stream
```
AFAIU that means that a player will basically skip (e.g. not play) the first 20 ms (1 packet) of the opus file, while the current pp-rec is generating a file that will play since the very first packet.
Still I don't understand if the two files are "synchronized", I mean if they start playing in the same point, have same duration and there is any "loss" of information in the new one (e.g. will that pre-skip packet played somehow? why drop it?).
Sorry for the questions, I'm not exactly a codec guy.

As of my other point

> has a lower bitrate (30kbps vs 39kbps), indeed the file size is smaller (130K vs 171K)

do you have any comment ?

lu-zero:
It is an estimate

```
    if (ic->pb && (filesize = avio_size(ic->pb)) > 0 && ic->duration > 0) {
        /* compute the bitrate */
        double bitrate = (double) filesize * 8.0 * AV_TIME_BASE /
                         (double) ic->duration;
        if (bitrate >= 0 && bitrate <= INT64_MAX)
            ic->bit_rate = bitrate;
    }
```

lminiero:
@lu-zero I think what @atoppi is mentioning are two separate things, that may or may not be an issue, and that's what we're trying to figure out:

1. If the pre-skip is causing the first 20ms of audio to be ignored by players (and media processors), that's a problem, as that may cause some desync problems when we're going to use the resulting media file in other contexts, e.g., to mux it with the corresponding video file. Is this something that libavformat always does, or is there a way to disable it? Or is it related to the pts/dts generated by `av_rescale_q`, that may generate negative values at startup?
2. The average bitstream is not just a matter of how it's calculated, but seems to be a consequence of files that actually are of a different size: the same MJR file (with some packet losses to test the silence ingestion) was converted to a 170KB file by the current processor, while it was converted to a 130KB file by the new one. This means that there are 40KB unaccounted for, one way or another, and it's not clear if it's because of differences in how silence is ingested. We'd expect the size to be more or less the same, as while there may be differences in how the ogg file header is created when using libogg difrecly vs. libavformat, we're manually adding a static silent audio packet, and packets that weren't lost all are expected to have the same in both files as well. We suspected it might be because of some transcoding happening somewhere (which would indeed result in a different Opus bitstream with a different bitrate), but looking at the code it does seem to be just saving the data to the container as we did for video streams already. Do you know what may be causing such a big difference instead?

Thanks!

lu-zero:
> @lu-zero I think what @atoppi is mentioning are two separate things, that may or may not be an issue, and that's what we're trying to figure out:
> 
>     1. If the pre-skip is causing the first 20ms of audio to be ignored by players (and media processors), that's a problem, as that may cause some desync problems when we're going to use the resulting media file in other contexts, e.g., to mux it with the corresponding video file. Is this something that libavformat always does, or is there a way to disable it? Or is it related to the pts/dts generated by `av_rescale_q`, that may generate negative values at startup?

I set explicitly the packet duration. The original code assumed 20ms packets, I made it explicit. Now the timestamps are identical.
 
>     2. The average bitstream is not just a matter of how it's calculated, but seems to be a consequence of files that actually are of a different size: the same MJR file (with some packet losses to test the silence ingestion) was converted to a 170KB file by the current processor, while it was converted to a 130KB file by the new one. This means that there are 40KB unaccounted for, one way or another, and it's not clear if it's because of differences in how silence is ingested. We'd expect the size to be more or less the same, as while there may be differences in how the ogg file header is created when using libogg difrecly vs. libavformat, we're manually adding a static silent audio packet, and packets that weren't lost all are expected to have the same in both files as well. We suspected it might be because of some transcoding happening somewhere (which would indeed result in a different Opus bitstream with a different bitrate), but looking at the code it does seem to be just saving the data to the container as we did for video streams already. Do you know what may be causing such a big difference instead?
> 

It is the muxing overhead, you can use `ffprobe -show_packets ${file} | grep size | cut -d '=' -f 2 | paste -sd+ | bc` to check.

Your sample has `128947 bytes` of actual data.

You can remux using `ffmpeg -i old.opus -c copy new.opus` and get
```
172K	old.opus
132K	new.opus
```

lminiero:
Oh I wasn't aware of muxing overhead and that it could be so high, thanks for the clarification and for the helpful snippet!

> I set explicitly the packet duration. The original code assumed 20ms packets, I made it explicit. Now the timestamps are identical.

Do you mean in a new commit? I don't see it yet.

lu-zero:
The push happened a bit later than the message :)

atoppi:
@lu-zero tried with another recording sample of my voice and got the  following with the new pp-rec
```
[ogg @ 0x61b000000080] Timestamps are unset in a packet for stream 0. This is deprecated and will stop working in the future. Fix your code to set the timestamps properly
[ogg @ 0x61b000000080] Encoder did not produce proper pts, making some up.
audio-new.opus is 70993 bytes
Bye!
```
Current version does not show any warning.
The generated file seems ok, though.

lu-zero:
Pull and try again please.

atoppi:
thanks, now everything looks good to me :+1:

lminiero:
We can merge then! :v: 
Thanks again for your contribution @lu-zero! (and for your patience going through our silly questions :hand_over_mouth: )

lu-zero:
Thank you for the review :)

PR #2671, opened 2021-05-21
fix caller and callee both simulcast or just one side simucast fail and fix the substream status show error on the webpage

lucylu-star:
1. fix caller and callee both simulcast fail.
2. support just callee  simulcast ok.
3. fix the substream status show error on the webpage.

### DIFF
--- a/plugins/janus_videocall.c
+++ b/plugins/janus_videocall.c
@@ -786,7 +786,7 @@ void janus_videocall_incoming_rtp(janus_plugin_session *handle, janus_plugin_rtp
 				json_t *result = json_object();
 				json_object_set_new(result, "event", json_string("simulcast"));
 				json_object_set_new(result, "videocodec", json_string(janus_videocodec_name(session->vcodec)));
-				json_object_set_new(result, "substream", json_integer(session->sim_context.substream));
+				json_object_set_new(result, "substream", json_integer(peer->sim_context.substream));
 				json_object_set_new(event, "result", result);
 				gateway->push_event(peer->handle, &janus_videocall_plugin, NULL, event, NULL);
 				json_decref(event);
@@ -798,7 +798,7 @@ void janus_videocall_incoming_rtp(janus_plugin_session *handle, janus_plugin_rtp
 				json_t *result = json_object();
 				json_object_set_new(result, "event", json_string("simulcast"));
 				json_object_set_new(result, "videocodec", json_string(janus_videocodec_name(session->vcodec)));
-				json_object_set_new(result, "temporal", json_integer(session->sim_context.templayer));
+				json_object_set_new(result, "temporal", json_integer(peer->sim_context.templayer));
 				json_object_set_new(event, "result", result);
 				gateway->push_event(peer->handle, &janus_videocall_plugin, NULL, event, NULL);
 				json_decref(event);
@@ -1358,20 +1358,15 @@ static void *janus_videocall_handler(void *data) {
 			json_t *msg_simulcast = json_object_get(msg->jsep, "simulcast");
 			if(msg_simulcast && janus_get_codec_pt(msg_sdp, "vp8") > 0) {
 				JANUS_LOG(LOG_VERB, "VideoCall callee (%s) is going to do simulcasting\n", session->username);
-				session->ssrc[0] = json_integer_value(json_object_get(msg_simulcast, "ssrc-0"));
-				session->ssrc[1] = json_integer_value(json_object_get(msg_simulcast, "ssrc-1"));
-				session->ssrc[2] = json_integer_value(json_object_get(msg_simulcast, "ssrc-2"));
+		        int rid_ext_id = -1;
+				janus_rtp_simulcasting_prepare(msg_simulcast, &rid_ext_id, session->ssrc, session->rid);
+				session->sim_context.rid_ext_id = rid_ext_id;
 			} else {
 				int i=0;
 				for(i=0; i<3; i++) {
 					session->ssrc[i] = 0;
 					g_free(session->rid[0]);
 					session->rid[0] = NULL;
-					if(peer) {
-						peer->ssrc[i] = 0;
-						g_free(peer->rid[0]);
-						peer->rid[0] = NULL;
-					}
 				}
 			}
 			/* Check which codecs we ended up using */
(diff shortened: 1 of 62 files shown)


lminiero:
Simulcast cannot be enabled by the receiving party, as far as I know, because the simulcast envelope is created when the transceiver is created. Since the transceiver is created by the PeerConnection and not by the browser, in case of a callee, simulcast cannot be added there.

lucylu-star:
As you said, callee cannot simulcast.  I revert some.
But, now have the bug that is caller simulcast failed and substream level status shows error when videocall.
I think the new pull request fix the issue.

lucylu-star:
Only modify  janus_videocall.c

plugins/janus_videocall.c
```diff
@@ -786,7 +786,7 @@ void janus_videocall_incoming_rtp(janus_plugin_session *handle, janus_plugin_rtp
 				json_t *result = json_object();
 				json_object_set_new(result, "event", json_string("simulcast"));
 				json_object_set_new(result, "videocodec", json_string(janus_videocodec_name(session->vcodec)));
-				json_object_set_new(result, "substream", json_integer(session->sim_context.substream));
+				json_object_set_new(result, "substream", json_integer(peer->sim_context.substream));
```
lminiero, line 789: This is wrong. The substream returned here is the substream the user (so `session`) wants from the peer. Substream and temporal info are always receiver related.

plugins/janus_videocall.c
```diff
@@ -798,7 +798,7 @@ void janus_videocall_incoming_rtp(janus_plugin_session *handle, janus_plugin_rtp
 				json_t *result = json_object();
 				json_object_set_new(result, "event", json_string("simulcast"));
 				json_object_set_new(result, "videocodec", json_string(janus_videocodec_name(session->vcodec)));
-				json_object_set_new(result, "temporal", json_integer(session->sim_context.templayer));
+				json_object_set_new(result, "temporal", json_integer(peer->sim_context.templayer));
```
lminiero, line 801: Same here.

plugins/janus_videocall.c
```diff
@@ -1358,20 +1358,15 @@ static void *janus_videocall_handler(void *data) {
 			json_t *msg_simulcast = json_object_get(msg->jsep, "simulcast");
 			if(msg_simulcast && janus_get_codec_pt(msg_sdp, "vp8") > 0) {
 				JANUS_LOG(LOG_VERB, "VideoCall callee (%s) is going to do simulcasting\n", session->username);
-				session->ssrc[0] = json_integer_value(json_object_get(msg_simulcast, "ssrc-0"));
-				session->ssrc[1] = json_integer_value(json_object_get(msg_simulcast, "ssrc-1"));
-				session->ssrc[2] = json_integer_value(json_object_get(msg_simulcast, "ssrc-2"));
+		        int rid_ext_id = -1;
+				janus_rtp_simulcasting_prepare(msg_simulcast, &rid_ext_id, session->ssrc, session->rid);
+				session->sim_context.rid_ext_id = rid_ext_id;
```
lminiero, line 1363: As I pointed out, callees cannot do simulcast. It's likely the whole code above (the one already in the repo) is redundant and should be removed.

plugins/janus_videocall.c
```diff
@@ -1358,20 +1358,15 @@ static void *janus_videocall_handler(void *data) {
...
 				int i=0;
 				for(i=0; i<3; i++) {
 					session->ssrc[i] = 0;
 					g_free(session->rid[0]);
 					session->rid[0] = NULL;
-					if(peer) {
-						peer->ssrc[i] = 0;
-						g_free(peer->rid[0]);
-						peer->rid[0] = NULL;
-					}
```
lminiero, line 1374: This is the only fix that I think does belong here: in fact, it was incorrectly disabling the previously configured simulcast settings for the peer (the caller in this instance).

lminiero (commented):
Added some new notes inline. Thanks!

plugins/janus_videocall.c
lucylu-star: disagree. 
 I think peer->sim_context.substream is right.   you can check it seeing the substream status on videocall webpage.  I have verified it.   Using peer, the showing is ok.  if using session, the status showing is abnormal.
Peer is the callee.   The substream returned here is what substream the callee wants. It is showed on webpage.

plugins/janus_videocall.c
lucylu-star: yes, I will remove it.

plugins/janus_videocall.c
lucylu-star: yes

plugins/janus_videocall.c
lminiero: Ah you're rght, sorry: this is an event we're sending to `peer`, not `session` as I thought :+1:   The `session->vcodec` above threw me off.

plugins/janus_videocall.c
```diff
@@ -1358,20 +1358,15 @@ static void *janus_videocall_handler(void *data) {
 			json_t *msg_simulcast = json_object_get(msg->jsep, "simulcast");
 			if(msg_simulcast && janus_get_codec_pt(msg_sdp, "vp8") > 0) {
 				JANUS_LOG(LOG_VERB, "VideoCall callee (%s) is going to do simulcasting\n", session->username);
-				session->ssrc[0] = json_integer_value(json_object_get(msg_simulcast, "ssrc-0"));
-				session->ssrc[1] = json_integer_value(json_object_get(msg_simulcast, "ssrc-1"));
-				session->ssrc[2] = json_integer_value(json_object_get(msg_simulcast, "ssrc-2"));
+		        int rid_ext_id = -1;
```
lminiero, line 1361: Indentation is broken here: it should be tabs, not spaces.

plugins/janus_videocall.c
lminiero: As soon as you remove tis part here, it's good to merge for me :+1:

lminiero: Note: I can get rid of the whole existing (and unneeded) simulcast check for callees myself later.

lucylu-star:
just remove several unneeded lines. Why auto check failed?

lminiero:
It looks like an issue pulling libnice from gitlab, so not caused by your patch. Hopefully it was a temporary error, I'll try to restart the automated build later.

lminiero:
Thanks, merging! :+1:

PR #2673, opened 2021-05-26
support sdp ssrc-group parse when SIM line is NULL in sdp

lucylu-star:
When the sdp is created by some android device and not by chrome ,  the created sdp have not the line of 'a=ssrc-group:SIM' and just have the line of 'a=ssrc-group:FID'.   The parse function can only get one ssrc.
So fix it to parse and get all SSRC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,22 @@
 All notable changes to this project will be documented in this file.
 
 
+## [v0.11.2] - 2021-05-03
+
+- Added support for relative paths in config files, currently only in MQTT event handler (thanks @RSATom!) [[PR-2623](#2623)]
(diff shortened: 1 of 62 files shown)


lminiero:
Sorry but this can't be accepted. The old way of doing simulcast (emphasis on OLD), that is what Chrome used to do years ago and still supports, needs the SIM attribute: the new one (the standard approach) used the rid attributes instead, and includes no sim attribute at all. This is explained in [this blog post](https://www.meetecho.com/blog/simulcast-janus-ssrc/). Not sure what that Android implementation you're mentioning is doing, but if it's trying to do old-style simulcast without using sim, then it's broken, and I'm not going to break our SDP parser just for that. You should fix the SDP those devices send yourself, before sending it to Janus.

PR #2674, opened 2021-05-26
added enable_recording api for audiobridge plugin.

rajneeshksoni:
- Added support in audio bridge plugin to start/stop recording in the middle of call.
- This is using the similar structure as "enable_recording" of videoroom plugin.
- WAV file is created with start timestamp so that audiobridge output can be combined with mrj dumps from videoroom
- i think it still have issue, when the 'janus_audiobridge_mixer_thread' is not closed, we probably need to update the wav header during the audioroom destroy call.

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -244,6 +244,7 @@ room-<unique room ID>: {
 	"room" : <unique numeric ID of the room>,
 	"secret" : "<room secret; mandatory if configured>"
 	"record" : <true|false, whether this room should be automatically recorded or not>,
+	"record_file" : "<file where audio recording is saved (optional)>"
 }
 \endverbatim 
  *
@@ -3298,6 +3299,7 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 		if(error_code != 0)
 			goto prepare_response;
 		json_t *record = json_object_get(root, "record");
+		json_t *recfile = json_object_get(root, "record_file");
 		gboolean recording_active = json_is_true(record);
 		/* Lookup room */
 		janus_mutex_lock(&rooms_mutex);
@@ -3315,6 +3317,11 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 			/* Room recording state has changed */
 			JANUS_LOG(LOG_VERB, "Recording status changed: prev=%d, curr = %d\n", g_atomic_int_get(&audiobridge->record), recording_active);
 			g_atomic_int_set(&audiobridge->record, room_prev_recording_active);
+			if(recfile && recording_active) {
+				g_free(audiobridge->record_file);
+				audiobridge->record_file = g_strdup(json_string_value(recfile));
+				JANUS_LOG(LOG_VERB, "Recording file path: %s\n", audiobridge->record_file);
+			}
 		}
 		response = json_object();
 		json_object_set_new(response, "audiobridge", json_string("success"));
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -234,6 +234,20 @@ room-<unique room ID>: {
 }
 \endverbatim
  *
+ * To enable or disable recording of mixed audio stream while the conference
+ * is in progress, you can make use of the \c enable_recording request,
+ * which has to be formatted as follows:
+ *
+\verbatim
+{
+	"request" : "enable_recording",
+	"room" : <unique numeric ID of the room>,
+	"secret" : "<room secret; mandatory if configured>"
+	"record" : <true|false, whether participants in this room should be automatically recorded or not>,
+}
+\endverbatim 
+ *
+ *
  * You can check whether a room exists using the \c exists request,
  * which has to be formatted as follows:
  *
@@ -1147,12 +1161,12 @@ typedef struct janus_audiobridge_room {
 	uint default_prebuffering;	/* Number of packets to buffer before decoding each participant */
 	int audio_active_packets;	/* Amount of packets with audio level for checkup */
 	int audio_level_average;	/* Average audio level */
-	gboolean record;			/* Whether this room has to be recorded or not */
+	volatile gint record;		/* Whether this room has to be recorded or not */
 	gchar *record_file;			/* Path of the recording file */
 	FILE *recording;			/* File to record the room into */
 	gint64 record_lastupdate;	/* Time when we last updated the wav header */
-	gboolean wav_header_added;  /* if wav header is added in recording file */
-	gint64 rec_start_time;		/* time when recording started for generating file name */
+	volatile gint wav_header_added;	/* If wav header is added in recording file */
+	gint64 rec_start_time;		/* Time when recording started for generating file name */
 	gboolean allow_plainrtp;	/* Whether plain RTP participants are allowed*/
 	gboolean destroy;			/* Value to flag the room for destruction */
 	GHashTable *participants;	/* Map of participants */
@@ -2339,9 +2353,9 @@ int janus_audiobridge_init(janus_callbacks *callback, const char *config_path) {
 			if(pin != NULL && pin->value != NULL) {
 				audiobridge->room_pin = g_strdup(pin->value);
 			}
-			audiobridge->record = FALSE;
+			g_atomic_int_set(&audiobridge->record, 0);
 			if(record && record->value && janus_is_true(record->value))
-				audiobridge->record = TRUE;
+				g_atomic_int_set(&audiobridge->record, 1);
 			if(recfile && recfile->value)
 				audiobridge->record_file = g_strdup(recfile->value);
 			audiobridge->recording = NULL;
@@ -2406,7 +2420,7 @@ int janus_audiobridge_init(janus_callbacks *callback, const char *config_path) {
 	while(g_hash_table_iter_next(&iter, NULL, &value)) {
 		janus_audiobridge_room *ar = value;
 		JANUS_LOG(LOG_VERB, "  ::: [%s][%s] %"SCNu32" (%s be recorded)\n",
-			ar->room_id_str, ar->room_name, ar->sampling_rate, ar->record ? "will" : "will NOT");
+			ar->room_id_str, ar->room_name, ar->sampling_rate, g_atomic_int_get(&ar->record) ? "will" : "will NOT");
 	}
 	janus_mutex_unlock(&rooms_mutex);
 
@@ -2656,7 +2670,6 @@ static int janus_audiobridge_access_room(json_t *root, gboolean check_modify, ja
 			g_snprintf(error_cause, error_cause_size, "No such room (%s)", room_id_str);
 		return error_code;
 	}
-	// FIXME:: it should check for recording lock
 	if(check_modify) {
 		char error_cause2[100];
 		JANUS_CHECK_SECRET((*audiobridge)->room_secret, root, "secret", error_code, error_cause2,
@@ -2870,9 +2883,9 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 				goto prepare_response;
 		}
 		audiobridge->room_ssrc = janus_random_uint32();
-		audiobridge->record = FALSE;
+		g_atomic_int_set(&audiobridge->record, 0);
 		if(record && json_is_true(record))
-			audiobridge->record = TRUE;
+			g_atomic_int_set(&audiobridge->record, 1);
 		if(recfile)
 			audiobridge->record_file = g_strdup(json_string_value(recfile));
 		audiobridge->recording = NULL;
@@ -3000,39 +3013,7 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 			g_free(room_id_str);
 		janus_mutex_unlock(&rooms_mutex);
 		goto prepare_response;
-	} else if(!strcasecmp(request_text, "enable_recording")) {
-		JANUS_VALIDATE_JSON_OBJECT(root, record_parameters,
-			error_code, error_cause, TRUE,
-			JANUS_AUDIOBRIDGE_ERROR_MISSING_ELEMENT, JANUS_AUDIOBRIDGE_ERROR_INVALID_ELEMENT);
-		if(error_code != 0)
-			goto prepare_response;
-		json_t *record = json_object_get(root, "record");
-		gboolean recording_active = json_is_true(record);
-		JANUS_LOG(LOG_VERB, "Enable Recording : %d \n", (recording_active ? 1 : 0));
-		/* Lookup room */
-		janus_mutex_lock(&rooms_mutex);
-		janus_audiobridge_room *audiobridge = NULL;
-		error_code = janus_audiobridge_access_room(root, TRUE, &audiobridge, error_cause, sizeof(error_cause));
-		if(error_code != 0) {
-			JANUS_LOG(LOG_ERR, "Failed to access audiobridge\n");
-			janus_mutex_unlock(&rooms_mutex);
-			goto prepare_response;
-		}
-		janus_mutex_lock(&audiobridge->mutex);
-		janus_mutex_unlock(&rooms_mutex);
-		/* Set recording status */
-		gboolean room_prev_recording_active = recording_active;
-		if (room_prev_recording_active != audiobridge->record) {
-			/* Room recording state has changed */
-			JANUS_LOG(LOG_VERB, "recording status changed : prev = %d curr = %d \n",audiobridge->record,recording_active);
-			audiobridge->record = room_prev_recording_active;
-        }
-		janus_mutex_unlock(&audiobridge->mutex);
-		response = json_object();
-		json_object_set_new(response, "audiobridge", json_string("success"));
-		json_object_set_new(response, "record", json_boolean(audiobridge->record));
-		goto prepare_response;
-	}	else if(!strcasecmp(request_text, "edit")) {
+	} else if(!strcasecmp(request_text, "edit")) {
 		JANUS_LOG(LOG_VERB, "Attempt to edit an existing AudioBridge room\n");
 		JANUS_VALIDATE_JSON_OBJECT(root, edit_parameters,
 			error_code, error_cause, TRUE,
@@ -3310,6 +3291,36 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 		json_object_set_new(response, "permanent", save ? json_true() : json_false());
 		JANUS_LOG(LOG_VERB, "Audiobridge room destroyed\n");
 		goto prepare_response;
+	} else if(!strcasecmp(request_text, "enable_recording")) {
+		JANUS_VALIDATE_JSON_OBJECT(root, record_parameters,
+			error_code, error_cause, TRUE,
+			JANUS_AUDIOBRIDGE_ERROR_MISSING_ELEMENT, JANUS_AUDIOBRIDGE_ERROR_INVALID_ELEMENT);
+		if(error_code != 0)
+			goto prepare_response;
+		json_t *record = json_object_get(root, "record");
+		gboolean recording_active = json_is_true(record);
+		JANUS_LOG(LOG_VERB, "Enable Recording : %d \n", (recording_active ? 1 : 0));
+		/* Lookup room */
+		janus_mutex_lock(&rooms_mutex);
+		janus_audiobridge_room *audiobridge = NULL;
+		error_code = janus_audiobridge_access_room(root, TRUE, &audiobridge, error_cause, sizeof(error_cause));
+		if(error_code != 0) {
+			JANUS_LOG(LOG_ERR, "Failed to access audiobridge\n");
+			janus_mutex_unlock(&rooms_mutex);
+			goto prepare_response;
+		}
+		janus_mutex_unlock(&rooms_mutex);
+		/* Set recording status */
+		gint room_prev_recording_active = recording_active ? 1 : 0;
+		if (room_prev_recording_active != g_atomic_int_get(&audiobridge->record)) {
+			/* Room recording state has changed */
+			JANUS_LOG(LOG_VERB, "Recording status changed : prev = %d curr = %d \n", g_atomic_int_get(&audiobridge->record), recording_active);
+			g_atomic_int_set(&audiobridge->record, room_prev_recording_active);
+		}
+		response = json_object();
+		json_object_set_new(response, "audiobridge", json_string("success"));
+		json_object_set_new(response, "record", json_boolean(g_atomic_int_get(&audiobridge->record)));
+		goto prepare_response;
 	} else if(!strcasecmp(request_text, "list")) {
 		/* List all rooms (but private ones) and their details (except for the secret, of course...) */
 		JANUS_LOG(LOG_VERB, "Request for the list for all audiobridge rooms\n");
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @rajneeshksoni! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

plugins/janus_audiobridge.c
```diff
@@ -1138,6 +1141,8 @@ typedef struct janus_audiobridge_room {
 	gchar *record_file;			/* Path of the recording file */
 	FILE *recording;			/* File to record the room into */
 	gint64 record_lastupdate;	/* Time when we last updated the wav header */
+	gboolean wav_header_added;  /* if wav header is added in recording file */
+	gint64 rec_start_time;		/* time when recording started for generating file name */
```
lminiero, line 1145: Code style: comments should be separated by the code via tabs, not spaces. Besides, all comments should start with a capitalized letter.

plugins/janus_audiobridge.c
```diff
@@ -2605,6 +2610,49 @@ json_t *janus_audiobridge_query_session(janus_plugin_session *handle) {
...
+		return error_code;
+	}
+	if(g_atomic_int_get(&((*audiobridge)->destroyed))) {
+		JANUS_LOG(LOG_ERR, "No such room (%s)\n", room_id_str);
+		error_code = JANUS_AUDIOBRIDGE_ERROR_NO_SUCH_ROOM;
+		if(error_cause)
+			g_snprintf(error_cause, error_cause_size, "No such room (%s)", room_id_str);
+		return error_code;
+	}
+	// FIXME:: it should check for recording lock
```
lminiero, line 2642: Code style: comments must always be in `/* */` form, not `//`.
Not sure why the comment says that, though: if this method was copied from the VideoRoom, then it can be used for other things as well, not just the new recording API you added.

plugins/janus_audiobridge.c
```diff
@@ -2931,7 +2979,39 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 			g_free(room_id_str);
 		janus_mutex_unlock(&rooms_mutex);
 		goto prepare_response;
-	} else if(!strcasecmp(request_text, "edit")) {
```
lminiero, line 2934: This should be *after* both `edit` and `destroy`, not before.

plugins/janus_audiobridge.c
```diff
@@ -2931,7 +2979,39 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+		janus_mutex_lock(&rooms_mutex);
+		janus_audiobridge_room *audiobridge = NULL;
+		error_code = janus_audiobridge_access_room(root, TRUE, &audiobridge, error_cause, sizeof(error_cause));
+		if(error_code != 0) {
+			JANUS_LOG(LOG_ERR, "Failed to access audiobridge\n");
+			janus_mutex_unlock(&rooms_mutex);
+			goto prepare_response;
+		}
+		janus_mutex_lock(&audiobridge->mutex);
+		janus_mutex_unlock(&rooms_mutex);
```
lminiero, line 3001: This is dangerous, and can lead to locks. In the VideoRoom we mostly use access_room to add a reference to the room before unlocking the global mutex. If you then need to lock the room, you can do that outside of the global mutex lock. That said, I think it's easier not to use a lock at all, and turn the `audiobridge->record` property into a `volatile int` instead and use atomic methods on it.

plugins/janus_audiobridge.c
```diff
@@ -2931,7 +2979,39 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+		/* Set recording status */
+		gboolean room_prev_recording_active = recording_active;
+		if (room_prev_recording_active != audiobridge->record) {
+			/* Room recording state has changed */
+			JANUS_LOG(LOG_VERB, "recording status changed : prev = %d curr = %d \n",audiobridge->record,recording_active);
```
lminiero, line 3006: Code style: messages should always start with a capitalized letter.
Arguments should be separated by a comma *and* a space.

plugins/janus_audiobridge.c
```diff
@@ -2931,7 +2979,39 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+			audiobridge->record = room_prev_recording_active;
+        }
```
lminiero, line 3008: Broken indentation: use tabs, not spaces.

plugins/janus_audiobridge.c
```diff
@@ -6740,6 +6820,101 @@ static void *janus_audiobridge_handler(void *data) {
...
 
+static void *janus_audiobridge_rec_add_wav_header(janus_audiobridge_room *audiobridge) {
+
+	/* Do we need to record the mix? */
+	char filename[255];
+	gint64 now = janus_get_real_time();
+	audiobridge->rec_start_time = now;
+	if(audiobridge->record_file) {
+		g_snprintf(filename, 255, "%s-%"SCNi64"%s%s", audiobridge->record_file,
+			now,rec_tempext ? "." : "", rec_tempext ? rec_tempext : "");
```
lminiero, line 6831: Code style: missing space after `now,`.

plugins/janus_audiobridge.c
```diff
@@ -6740,6 +6820,101 @@ static void *janus_audiobridge_handler(void *data) {
...
+	} else {
+		g_snprintf(filename, 255, "janus-audioroom-%s-%"SCNi64".wav%s%s", audiobridge->room_id_str,
+			now,rec_tempext ? "." : "", rec_tempext ? rec_tempext : "");
```
lminiero, line 6834: See above.

plugins/janus_audiobridge.c
```diff
@@ -6740,6 +6820,101 @@ static void *janus_audiobridge_handler(void *data) {
 	return NULL;
 }
 
+static void *janus_audiobridge_rec_add_wav_header(janus_audiobridge_room *audiobridge) {
+
```
lminiero, line 6824: This empty line is unneeded.

plugins/janus_audiobridge.c
```diff
@@ -6740,6 +6820,101 @@ static void *janus_audiobridge_handler(void *data) {
...
+			JANUS_LOG(LOG_ERR, "Error writing WAV header...\n");
+		}
+		fflush(audiobridge->recording);
+		audiobridge->record_lastupdate = janus_get_monotonic_time();
+		audiobridge->wav_header_added = TRUE;
+	}
+}
+
+static void *janus_audiobridge_update_wav_header(janus_audiobridge_room *audiobridge) {
+
```
lminiero, line 6867: This empty line is unneeded.

plugins/janus_audiobridge.c
```diff
@@ -6993,6 +7132,14 @@ static void *janus_audiobridge_mixer_thread(void *data) {
 			g_list_free_full(anncs_list, (GDestroyNotify)janus_audiobridge_participant_unref);
 		}
 #endif
+		if(audiobridge->record == TRUE && audiobridge->wav_header_added == FALSE) {
```
lminiero, line 7135: This can be simplified in

    if(audiobridge->record && !audiobridge->wav_header_added) {

plugins/janus_audiobridge.c
```diff
@@ -6993,6 +7132,14 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+			JANUS_LOG(LOG_VERB, "Add wave header for recording %s (%s)...\n", audiobridge->room_id_str, audiobridge->room_name);
+			janus_audiobridge_rec_add_wav_header(audiobridge);
+		}
+		if(audiobridge->record == FALSE && audiobridge->wav_header_added == TRUE) {
```
lminiero, line 7139: See above.

plugins/janus_audiobridge.c
```diff
@@ -6993,6 +7132,14 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+			JANUS_LOG(LOG_VERB, "update wave header for recording %s (%s)...\n", audiobridge->room_id_str, audiobridge->room_name);
```
lminiero, line 7140: See above (capitalized first message letter).

plugins/janus_audiobridge.c
```diff
@@ -7187,50 +7334,11 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
-		/* Also notify event handlers */
-		if(notify_events && gateway->events_is_enabled()) {
-			json_t *info = json_object();
-			json_object_set_new(info, "event", json_string("recordingdone"));
-			json_object_set_new(info, "room",
-				string_ids ? json_string(audiobridge->room_id_str) : json_integer(audiobridge->room_id));
-			json_object_set_new(info, "record_file", json_string(filename));
-			gateway->notify_event(&janus_audiobridge_plugin, NULL, info);
-		}
+	if(audiobridge->recording != NULL && audiobridge->wav_header_added == TRUE) {
```
lminiero, line 7337: See above on streamlining check.

plugins/janus_audiobridge.c
```diff
@@ -7187,50 +7334,11 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+		JANUS_LOG(LOG_VERB, "update wave header for recording %s (%s)...\n", audiobridge->room_id_str, audiobridge->room_name);
```
lminiero, line 7338: See above (capitlized first letter).

plugins/janus_audiobridge.c
```diff
@@ -7187,50 +7334,11 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+		janus_audiobridge_update_wav_header(audiobridge);
 	}
+	// close the recording file here
```
lminiero, line 7341: Comments should always use `/* */`.

plugins/janus_audiobridge.c
```diff
@@ -1138,6 +1141,8 @@ typedef struct janus_audiobridge_room {
 	gchar *record_file;			/* Path of the recording file */
 	FILE *recording;			/* File to record the room into */
 	gint64 record_lastupdate;	/* Time when we last updated the wav header */
+	gboolean wav_header_added;  /* if wav header is added in recording file */
```
lminiero, line 1144: This property could be a `volatile int` and updated/read using GLib atomic utilities.

lminiero (commented):
Thanks! We don't use the recording to WAV, so I can't say I care much about the feature, but I understand how you and others may actually benefit from this.

I added a few notes inline, mostly editorial.

lminiero:
@rajneeshksoni any update on my notes above? As a side note, please notice that there's apparently conflicts now that we've merged some new things in master. I suspect it may impact some code you moved around, since part of the changes are optional support for stereo mixing.

rajneeshksoni:
@itamayo I'll fix the conflicts, sorry for the delay.
I have observed one more issue about janus_audiobridge_mixer_thread, this thread is started at the time of room creation but i am not seeing any code which close this thread at the time of room destroy.

rajneeshksoni:
@lminiero Sorry for the delay, i'll fix the conflict and update as per review comments.

plugins/janus_audiobridge.c
rajneeshksoni: @lminiero in most of the code i see empty line between end of function and start of next function. sometimes it is missing. What is the exact convention

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: resolved by adding audiobridge->record to atomic method.

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
rajneeshksoni: done

plugins/janus_audiobridge.c
lminiero: Fair point! think in the majority of cases it's no-empty line at the beginning, at least in the code I write myself.

rajneeshksoni:
@lminiero can you check the fixes and provide your feedback.

lminiero:
@rajneeshksoni I already have a review planned, but I'll need time to do that and go into the details, especially since there's recent merges to take into account. I'll get back to you as soon as I do that.

plugins/janus_audiobridge.c
```diff
@@ -233,6 +233,20 @@ room-<unique room ID>: {
...
+ * To enable or disable recording of mixed audio stream while the conference
+ * is in progress, you can make use of the \c enable_recording request,
+ * which has to be formatted as follows:
+ *
+\verbatim
+{
+	"request" : "enable_recording",
+	"room" : <unique numeric ID of the room>,
+	"secret" : "<room secret; mandatory if configured>"
+	"record" : <true|false, whether participants in this room should be automatically recorded or not>,
```
lminiero, line 246: I'd change this sentence to "whether this room should be automatically recorded or not", or otherwise someone reading the documentation might think this will record each participant individually to a separate .mjr (which the AudioBridge plugin does support as a feature). I guess this was copied from the VideoRoom docs, where that's indeed what's done.

plugins/janus_audiobridge.c
```diff
@@ -3230,9 +3291,39 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
 		goto prepare_response;
+	} else if(!strcasecmp(request_text, "enable_recording")) {
+		JANUS_VALIDATE_JSON_OBJECT(root, record_parameters,
+			error_code, error_cause, TRUE,
+			JANUS_AUDIOBRIDGE_ERROR_MISSING_ELEMENT, JANUS_AUDIOBRIDGE_ERROR_INVALID_ELEMENT);
+		if(error_code != 0)
+			goto prepare_response;
+		json_t *record = json_object_get(root, "record");
+		gboolean recording_active = json_is_true(record);
+		JANUS_LOG(LOG_VERB, "Enable Recording : %d \n", (recording_active ? 1 : 0));
```
lminiero, line 3302: I think this line here is not needed, or at the very least it's useless here, since it doesn't even say which room this refers to. The line you added below is enough.

plugins/janus_audiobridge.c
```diff
@@ -3230,9 +3291,39 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+			JANUS_LOG(LOG_ERR, "Failed to access audiobridge\n");
+			janus_mutex_unlock(&rooms_mutex);
+			goto prepare_response;
+		}
+		janus_mutex_unlock(&rooms_mutex);
+		/* Set recording status */
+		gint room_prev_recording_active = recording_active ? 1 : 0;
+		if (room_prev_recording_active != g_atomic_int_get(&audiobridge->record)) {
+			/* Room recording state has changed */
+			JANUS_LOG(LOG_VERB, "Recording status changed : prev = %d curr = %d \n", g_atomic_int_get(&audiobridge->record), recording_active);
```
lminiero, line 3317: Debug style: there should be no space before the semicolon (`changed: `), or empty spaces at the end of a sentence. The spaces around the `=` are unneeded as well, and values should be comma separated (`prev=%d, curr=%d\n`).

lminiero (commented):
Added a few minor notes inline. I'll try to make some tests later today :+1:

lminiero:
@rajneeshksoni I tested it and it seems to be working as expected :+1: 

I think there's something missing in the API, though, that is the filename you want to save to. At the moment, you can only specify a filename when creating the room, but that assumes your willingness to start recording something right away. With your new functionality, you may want to only start recording later, which means you didn't have the option of specifying where to save and with with filename base, but only an on/off trigger: as a result, all files were created in the folder I started Janus from. I think the `enable_recording` should be extended with a `record_file` property too (to only be used when starting a recording, of course).

rajneeshksoni:
@lminiero regarding "record_file" parameter in enable_recording, i have change the recording file naming format. it also contain the time stamp, e.g. for filename 'audiobridge', file is currently saved as  audiobridge-<timestamp>.
so i think we should just have parameter 'record_dir' where file will be saved as record_dir/audiobridge-<timestamp>.wav
pls confirm

rajneeshksoni:
@lminiero Since i have added "-timestamp" as well in the name, so i am but confused how 'record_file' should handle appending timestamp in the filename. Appending timestamp in filename is important for later combining with video.
if user provide filename as /opt/recording/recfile.wav  should i create recording file as - /opt/recording/recfile-'timestamp'.wav or /opt/recording/recfile.wav-'timestamp'.

lminiero:
That's already what your code is trying to do, which means it would be broken if people passed `record_file` when creating a room:

```
	if(audiobridge->record_file) {
		g_snprintf(filename, 255, "%s-%"SCNi64"%s%s", audiobridge->record_file,
			now, rec_tempext ? "." : "", rec_tempext ? rec_tempext : "");
	} else {
		g_snprintf(filename, 255, "janus-audioroom-%s-%"SCNi64".wav%s%s", audiobridge->room_id_str,
			now, rec_tempext ? "." : "", rec_tempext ? rec_tempext : "");
	}
```

If `record_file` is provided, you just need to pass `%s` as the original code did and nothing else. It's the user who has control on filenames, and their responsibility to use different names when starting the recording multiple times.

plugins/janus_audiobridge.c
```diff
@@ -1047,6 +1062,9 @@ static struct janus_json_parameter join_parameters[] = {
 	{"rtp", JSON_OBJECT, 0},
 	{"secret", JSON_STRING, 0}
 };
+static struct janus_json_parameter record_parameters[] = {
+	{"record", JANUS_JSON_BOOL, JANUS_JSON_PARAM_REQUIRED}
```
lminiero, line 1066: Since there's `record_file` too, now, it should be validates as well.

plugins/janus_audiobridge.c
```diff
@@ -6819,6 +6916,99 @@ static void *janus_audiobridge_handler(void *data) {
...
+		fwrite(&size, sizeof(uint32_t), 1, audiobridge->recording);
+		fflush(audiobridge->recording);
+	}
+	fclose(audiobridge->recording);
+	audiobridge->recording = NULL;
+	g_atomic_int_set(&audiobridge->wav_header_added, 0);
+
+	char filename[255];
+	if(audiobridge->record_file) {
+		g_snprintf(filename, 255, "%s-%"SCNi64, audiobridge->record_file,audiobridge->rec_start_time);
```
lminiero, line 6981: This should just be `"%s"`, not `"%s-%"SCNi64`: `record_file` contains the `.wav` extension already, which means you'd be appending the time _after_ the extension, breaking it. There's no need to add the current time to the filename when it's provided by the end user: as explained in another comment, in that case it's the user's responsibility to provide unique filenames, or risk overwriting previous recordings otherwise.

plugins/janus_audiobridge.c
```diff
@@ -6819,6 +6916,99 @@ static void *janus_audiobridge_handler(void *data) {
...
+	} else {
+		g_snprintf(filename, 255, "janus-audioroom-%s-%"SCNi64".wav", audiobridge->room_id_str,
+			audiobridge->rec_start_time);
+	}
+	if(rec_tempext) {
+		/* We need to rename the file, to remove the temporary extension */
+		char extfilename[255];
+		if(audiobridge->record_file) {
+			g_snprintf(extfilename, 255, "%s-%"SCNi64".%s", audiobridge->record_file, 
+				audiobridge->rec_start_time,rec_tempext);
```
lminiero, line 6991: Same here, no need for the start time. You can refer to how the code was before, to see how we handled that case (`"%s%s%s"`).

lminiero (commented):
Thanks for the changes! I still see some problems with the user-provided filenames, though: I've added a few comments inline.

plugins/janus_audiobridge.c
rajneeshksoni: I wanted to record the timestamp in filename so that it can be used for audio video synchronisation, i fear if timestamp is added in with the API call, the delay between API call may not reflect the exact time.
Can we add ".wav" after the timestamp ?

plugins/janus_audiobridge.c
rajneeshksoni: record_file is optional parameter, if user do not supply parameter we are using default name. Do we still need to add it for validation ?

plugins/janus_audiobridge.c
lminiero: Again, that's not something YOU have to worry about. If people provide `record_file`, than they may put the timestamp in the filename yourself. If you don't provide it, then your pattern does the job.

>  Can we add ".wav" after the timestamp ?

No, it would be a breaking change (since yout changes impact how recordings are started after `create` too).

plugins/janus_audiobridge.c
lminiero: ALL parameters need to be validated, optional or not. You'll see that we have a lot of parameters defined without `JANUS_JSON_PARAM_REQUIRED`, for instance.

lminiero:
I've made some tests and it looks like the feature doesn't work if there's no one in the room: in that case, if I stop the recording and then start it again, when a new person joins it will keep on recording media to the old file, and will probably result in a broken file when we try to update the header (because the filename we were keeping track of will be different than before). I suspect that's because of how the mixer thread works, which does close to nothing when the room is empty: as such, when the recording is stopped and then started again, when it reaches your code it finds the recording status the same as before, and so thinks nothing has changed.

That said, I think this is something I can address myself. To speed things up, I'll address my notes above as well (missing parameter, recording file without timestamp), and maybe think of an alternative `record_dir` to add in the future. As such, I'll merge this as it is, and reference it in the upcoming commits. Thanks!

rajneeshksoni:
@lminiero I'll analyze the above issue and create separate PR for it.
Thanks for merging.

lminiero:
No need, I'm looking at it right now :+1: 
Hopefully I should have a fix soon.

lminiero:
The above commit fixes it for me: it was enough to move the WAV init/update before the `continue` for empty rooms. This way, the recording is always updated no matter whether the mixer will do anything or not. This has the side effect of immediately creating an empty `.wav` for empty rooms, where before we'd only start creating a file at the first join, but I think that's not an issue (we do the same with `.mjr` files).

rajneeshksoni:
working for me.

PR #2679, opened 2021-06-01
fix: [janus_sip] Fix "call_id" property in "missed_call" events

nicolasduteil:
This PR fixes the way `call_id` property is retrieved in sip plugin when emitting `missed_call` events

In current master, when a session is busy and no other helper is available, the retrieved *call-id* is the one from the previously established session instead of the one from the new received *INVITE*


lminiero:
Good catch, I guess we were putting the current Call-ID in there instead. Merging :+1:

PR #2680, opened 2021-06-01
Additional target formats for some recorded codecs (fixes #2658)

### LINKED ISSUE #2658: Output opus (optionally) in mka

lminiero:
This patch takes advantage of the refactoring @lu-zero performed recently in #2665, to add support for additional target formats to some recordings. Specifically, this patch now allows you to convert Opus .mjr files to `.ogg` and `.mka` as well (which addreses #2658), while VP8/VP9 recordings can now also be converted to `.mkv`. In the future we may extend other codec processors to support more formats as well (e.g., H.264/H.265/AV1 in webm or mkv too, if they're supported?), but this is just a way to start changing the code a bit to make it possible.


chriswiggins:
h.264,h.265 and av1 can definitely go in an MKV container - happy to add that functionality in but don't have the `.mjr` files to test with unless there are some handy?

lminiero:
Thanks for the heads up, I'll give those a try :+1: 
On the mjrs, they should be easy to generate if you want to test yourself, e.g., by passing `?vcodec=av1` to the Record&Play demo page url to force AV1.

lminiero:
Mh I tried using matroska as a target container for an H.264 recording, and it resulted in a 18m video instead of the 12s it lasts, with video moving very slowly. Something broken with how the framerate is written to the file, I guess.

lminiero:
Ok, using `av_rescale_q` instead of my ugly old code fixed it :hand_over_mouth:

lminiero:
Merging.

PR #2684, opened 2021-06-04
Configurable mechanism for manually setting static event loop to use for new handles (see #2450)

lminiero:
Same as #2450, but for master. Please refer to that PR for more info.
I don't have time to test. If you care about this, please provide feedback.

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -105,6 +105,18 @@ general: {
 									# As such, if you want to use this you should
 									# provision the correct value according to the
 									# available resources (e.g., CPUs available).
+	#allow_loop_indication = true	# In case a static number of event loops is
+									# configured as explained above, by default
+									# new handles will be allocated on one loop or
+									# another by the Janus core itself. In some cases
+									# it may be helpful to manually tell the Janus
+									# core which loop a handle should be added to,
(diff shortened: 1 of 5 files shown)


fancycode:
Thanks, would it be possible to include this setting in the "info" object, similar to what is already done with `static-event-loops`?
https://github.com/meetecho/janus-gateway/blob/16e2cc310c6081c645b63b1cc9dbb80b8d55ff26/janus.c#L376

That way, clients could detect if handles can be assigned to loops or not.

I'm happy to do a follow-up PR myself once this has landed.

lminiero:
Done (not tested).

lminiero:
Any update on this?

lminiero:
This has been around two months and I got no feedback. If I don't get any feedback on Monday, I'll close the PR, as I'll assume no one needs this anymore (we currently don't).

fancycode:
Sorry for not replying earlier, other things got in the way. This is working fine for us and improves CPU usage a little bit in some scenarios if we pin the subscribers to the same loop than the publisher is running on.

Thanks for splitting this up in a separate PR, would be great if you could merge it. Thanks!

lminiero:
Thanks for the feedback, and apologies for the late reply (just got back from vacation). Merging then :+1:

PR #2685, opened 2021-06-04
Initial integration of RED for audio

lminiero:
This effort implements the foundation for RED support in Janus. You can find more information on what this means in this [informative blog post](https://webrtchacks.com/red-improving-audio-quality-with-redundancy/) by @fippo, and I'll add a blog post going more in detail on this patch soon as well.

At the moment, this provides RED support to:

* SDP negotiation (obviously)
* recordings and post-processor
* EchoTest, VideoCall, SIP, NoSIP, Record&Play and (partially) Streaming plugins

No integration yet in AudioBridge and VideoRoom, but the "bricks" are there. As you'll read in the blog post, there are some things to take into account before this can be actually used, but if you're interested in this effort, please do make some tests, especially in problematic networks, to check if you notice audio improvements.


lminiero:
FYI, there's a first integration of RED in the VideoRoom plugin as well: notice that you'll want not to enforce too low bitrate caps in the room, as RED takes a lot of bandwidth itself that would otherwise mostly be used for video under "normal" circumstances. The way it works is that you can configure a room to accept RED, when you create a new one: then, if a publisher negotiates RED, it becomes a RED-enabled publisher; subscribers can choose whether to use RED themselves too, or fallback to plain RTP (RED→RED and RED→nonRED). We didn't add the option of having RED subscribers for nonRED publishers, even though this may come later.

I think I'll change the way the RED integration works in the Streaming plugin, instead. At the moment, that's exactly what it does (nonRED→RED), but I don't think it makes much sense. I think a better option would be the ability to receive a RED stream externally, and then forward it via RED or nonRED to subscribers (a bit like we do with the VideoRoom). This would also facilitate the integration with RTP forwarders, the moment we update those as well. I do see some challenges on that, though, mostly related to payload types: at the moment, it doesn't really matter what payload type the RTP packets we receive have, as we overwrite it anyway with whatever we negotiated; the moment we have to support RED on the way in too, though, it means that the payload type would be an additional mechanism for decapsulation ("did I receive an Opus packet, or a RED one?"), as we'd need to know how to process an incoming packet. This might be more constraining than how mountpoints are right now.

fippo:
note: libwebrtc will soon require the fmtp line in the sdp

lminiero:
> note: libwebrtc will soon require the fmtp line in the sdp

@fippo thanks for the heads up! Do you have a reference link for the change, so that I can check what needs to be updated in our code?

lminiero:
PS: is the number of redundant packets also going to change officially? I seem to remember you told they were experimenting with 1 instead of 2 by default.

fippo:
the fmtp change got reverted since it broke an arcane "downstream" test that I don't have access to even. You should still announce it in the SDP ;-)

Distance 1 is the default as of https://bugs.chromium.org/p/webrtc/issues/detail?id=11640#c46

lminiero:
@shripchenko thanks for the detailed feedback! And no offense taken, this is indeed very experimental and mostly a first attempt at integrating this somehow :wink: 

You make good points on how the current code doesn't really take advantage of the redundant info when depacketizing stuff, especially in the VideoRoom plugin. I don't have a clear idea of how to solve this, especially in the RED-to-nonRED scenario, as the plugin would need to start keeping track of what it sent to a subscriber already, and packets may be coming in out of order. I'll have to think about this.

Making this a core-only-thing, and keeping it transparent to plugins is an interesting concept, even though I see some problems with this approach:

1. it would make the core work harder to perform, as it would need to keep track of which packet it already passed to a plugin and which wasn't passed yet (something it doesn't have to do now), since we would of course not want duplicate packets to be managed by plugins (this never happens now, thanks to SRTP);
2. plugins would have no control on when to use RED and when not to use it, on a session that negotiated it, which might be helpful to do in a dynamic way; if the core is in control of that, the plugin won't have a word in it;
3. it would be much less efficient in one-to-many scenarios, because if it's the core that does RED packetization, then the same audio stream sent to 200 people means doing the same RED packetization 200 times for the same packet, rather than just once as we do now.

These are just the first issues I see, but there may be more. I guess possibly a hybrid approach might be taken, e.g., where we keep this in the plugins responsibility as it is now, but the core does depacketize RED before sending a RED packet, just to be able to NACK an inner packet if it wasn't received (which as you pointed out wouldn't be possible otherwise).

lminiero:
Since this has been around for a while, already, I've decided to limit the scope of this PR, in order to be able to merge it right away. As such, I removed any of the changes I made to Streaming and VideoRoom plugins, since they were more invasive, and quite incomplete: it made more sense to me to start merging the relevant parts we implemented (negotiation support, parsing/building RED blocks) and the simpler integrations (EchoTest, Record&Play, VideoCall, SIP, NoSIP), leaving more complex integrations for future efforts, where we can focus on those parts specifically (AudioBridge, Streaming and VideoRoom will each require a dedicated effort for sure).

That said, this does already provide some value, especially for initial testing: besides, the Record&Play plugin does make use of the new RED utilities already.

lminiero:
You'll need to open a new issue, and be much more detailed than that (Janus version, plugins used, how, etc.) for us to be able to help. RED integration in Janus is currently superficial, and most plugins won't actually use it.

PR #2687, opened 2021-06-06
minor adjustment to the audiobridge docs

augustblack:


lminiero:
Good catch, since that's tied to the [Opus complexity](https://wiki.xiph.org/Opus_Recommended_Settings#Trading_Coding_Efficiency_with_CPU_Time) settings, so that was indeed a weird typo on our end. Merging :+1:

PR #2688, opened 2021-06-07
Fix SIP plugin unhold request docs typo

ycherniavskyi:
It seems just copy-paste typo.


lminiero:

PR #2689, opened 2021-06-08
Optionally allow IPv6 link-local addresses to be gathered too

lminiero:
Addresses discussion made on the [group](https://groups.google.com/g/meetecho-janus/c/TcVt9-iaZQ8), where apparently our filtering of link-local addresses causes issues sometimes. This patch adds a configuration setting (and command-line flag) to allow them: default remains `FALSE` to be backwards compatible.


mdevaev:
It seems it's working! Many thanks.

lminiero:
Ack, merging then.

PR #2690, opened 2021-06-10
Fix streaming plugin mutex unlock when disabling mountpoint

lionelnicolas:
When trying to debug a different issue (see #2691) in the streaming plugin, I've made a review of locks/unlocks in that plugin and found this (probable?) issue.


lminiero:
Yes, that's definitely an issue, and yours is the right fix :+1: 
Thanks, merging!

PR #2692, opened 2021-06-10
Fixed broken switching when using different payload types in Streaming plugin

lminiero:
This patch fixes a broken behaviour I discovered in the Streaming plugin when using the switching functionality. The plugin allows you to dynamically "switch" to a different mountpoint you subscribed to initially, assuming the same codecs are used: this can be used to easily "change channel" when using the Streaming plugin, without using a new PeerConnection or renegotiating the existing one (as far as the browser is concerned, it's the same stream, as the plugin will fix RTP headers to make the transition seamless).

This usually works nicely in both Streaming plugin and VideoRoom, but in the Streaming plugin this breaks if the mountpoint you're leaving and the one you're moving to don't use the same payload types. In that case, if e.g. you're expecting payload type X for VP8 because that's what subscribing to the first mountpoint offered you in the SDP, if the mountpoint you switch to uses payload type Y for VP8, then you'll start receiving RTP packets with payload type Y, and the browser will obviously ignore them and video will stop working, even though it's still VP8. This patch fixes this behaviour, by basically keeping track of which payload types were assigned to a subscriber PeerConnection, and overwrite payload types accordingly when sending that subscriber RTP packets no matter the source (old or new mountpoint).

Notice that this is only an issue in the Streaming plugin. The VideoRoom implements the feature differently (payload types are _always_ overwritten, and are always the same for the same codec), and so is not affected.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -1017,6 +1017,7 @@ typedef struct janus_streaming_rtp_relay_packet {
 	uint32_t ssrc[3];
 	janus_videocodec codec;
 	int substream;
+	int ptype;
 	uint32_t timestamp;
 	uint16_t seq_number;
 	/* The following are only relevant for VP9 SVC*/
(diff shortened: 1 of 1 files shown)


lminiero:
Merging.

PR #2693, opened 2021-06-10
Added missing curly brace

0xdw:


januscla:
Thanks for your contribution, @0xdw! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Good catch :+1:  I see we have a similar typo in the example after that, I'll fix it myself.

0xdw:
> Good catch 👍 I see we have a similar typo in the example after that, I'll fix it myself.

Yeah, there are more. I was thinking to fix them out today night.

PR #2699, opened 2021-06-12
Fix memory leak on aborted RTSP connection

lionelnicolas:
Two return statements without cleaning up `libcurl` data


lminiero:
Thanks, merging :+1:

Issue #2691, opened 2021-06-10
streaming plugin: deadlock when handling API messages


lionelnicolas:
I think one of our janus servers experienced a deadlock in the streaming plugin. The websocket API became unresponsive and I was not able to create/destroy mountpoint anymore (no response from the streaming plugin API)

I think the deadlock happens on the `mountpoints_mutex` mutex.

To determine that, I use the admin API to communicate with the streaming plugin.

If sending :
```
{
   "request": "info"
}
```

I receive the reply a reply from the streaming plugin :

```
{
    "janus": "success",
    "transaction": "A0WGq3Tkgnaz",
    "response": {
        "streaming": "event",
        "error_code": 453,
        "error": "Missing mandatory element (id)"
    }
}
```

Which is handled by https://github.com/meetecho/janus-gateway/blob/2d83e9669d0c6509b82d3b9c1d0c0d247f0026d5/plugins/janus_streaming.c#L2489-L2499 , before the mutex is locked.

But, if trying to request info on a non-existing endpoint :

```
{
   "request": "info",
   "id": "123"
}
```

In that case the request never completes, and no log messages are shown (I was expecting `No such mountpoint/stream`), so I guess the query is stuck here :

https://github.com/meetecho/janus-gateway/blob/2d83e9669d0c6509b82d3b9c1d0c0d247f0026d5/plugins/janus_streaming.c#L2510

I wasn't able to reproduce the issue yet, but I'll update the ticket when I have some news.

Notes:
 * this was on 9eeeb38c71098d7bed7cf951c2d1503d0c4872ec (3 commits from master as of today)
 * not sure if that's a coincidence, but if looks like this happened few seconds after 6-7 mountpoint destroy were done in a short period of time
 * existing streaming plugin RTSP mountpoints are still running (I can see them looping on RTSP retries because the source is gone, but I can't destroy those mountpoints)
 * the videoroom plugin is also running on the same server, and continue to work even if the streaming plugin API seems dead

lminiero:
If you want to debug locks, you can use our lock debugging feature, which will print any attempt to lock/unlock and the file/line where it happened. Ideally you'll want it to start from the beginning (enabled in janus.jcfg), but it can be enabled/disabled dynamically via Admin API as well. Of course, you should ensure it's enabled before the issue starts, or you may be missing info. Notice it might also greatly increase the log size.

lionelnicolas:
Ok thanks, I'm planning to use that feature as soon as I find a reproducer.

PR #2700, opened 2021-06-12
Fix deadlock on mountpoint destroy during RTSP reconnect

lionelnicolas:
This fixes a deadlock which occurs under some conditions when an RTSP mountpoint destroy is requested from the API, while the mountpoint is reconnecting to the RTSP server.

When a `destroy` message is handled, the `mountpoints_mutex` is locked (https://github.com/meetecho/janus-gateway/blob/c621bb553a207ffb7843101d304564a31f4ec6e2/plugins/janus_streaming.c#L3777), then the mountpoint is removed from the hash table (https://github.com/meetecho/janus-gateway/blob/c621bb553a207ffb7843101d304564a31f4ec6e2/plugins/janus_streaming.c#L3798). The removal triggers `janus_streaming_mountpoint_destroy()` callback.

`janus_streaming_mountpoint_destroy()` interrupt the poll then join the relay thread to wait for it to exit (https://github.com/meetecho/janus-gateway/blob/c621bb553a207ffb7843101d304564a31f4ec6e2/plugins/janus_streaming.c#L1284).

But, if the mountpoint was trying to reconnect at the same moment (https://github.com/meetecho/janus-gateway/blob/c621bb553a207ffb7843101d304564a31f4ec6e2/plugins/janus_streaming.c#L7643) and `mountpoint->destroyed` not set to `1` yet, then the relay thread will also try to lock `mountpoints_mutex` (https://github.com/meetecho/janus-gateway/blob/c621bb553a207ffb7843101d304564a31f4ec6e2/plugins/janus_streaming.c#L6485). This results in a deadlock if the lock is currently held by the caller of `join()` on that mountpoint thread.

As the value of `mountpoint->destroyed` could change while waiting for the lock, the solution was to use `g_mutex_trylock()` to be able to abort the lock attempt when `destroyed` is eventually set to `1`.

I've added a `janus_mutex_unlock()` macro, with no trailing `;` to be able to use the macro in `if` or `while` statement (otherwise the compiler was [not happy](https://www.includehelp.com/c-programs/expected-bracket-before-semicolon-token-error-in-c.aspx)). I've also removed the semicolon in other `janus_mutex_*` definitions for consistency.

This is based on top of #2699. So I'll rebase this PR once the other one is merged. I've chosen to open two separated PRs as they are fixing two different issues.

Fixes: #2691

### DIFF
--- a/mutex.h
+++ b/mutex.h
@@ -28,17 +28,17 @@ typedef pthread_mutex_t janus_mutex;
 /*! \brief Janus mutex destruction */
 #define janus_mutex_destroy(a) pthread_mutex_destroy(a)
 /*! \brief Janus mutex lock without debug */
-#define janus_mutex_lock_nodebug(a) pthread_mutex_lock(a);
+#define janus_mutex_lock_nodebug(a) pthread_mutex_lock(a)
 /*! \brief Janus mutex lock with debug (prints the line that locked a mutex) */
-#define janus_mutex_lock_debug(a) { JANUS_PRINT("[%s:%s:%d:lock] %p\n", __FILE__, __FUNCTION__, __LINE__, a); pthread_mutex_lock(a); };
+#define janus_mutex_lock_debug(a) { JANUS_PRINT("[%s:%s:%d:lock] %p\n", __FILE__, __FUNCTION__, __LINE__, a); pthread_mutex_lock(a); }
 /*! \brief Janus mutex lock wrapper (selective locking debug) */
-#define janus_mutex_lock(a) { if(!lock_debug) { janus_mutex_lock_nodebug(a); } else { janus_mutex_lock_debug(a); } };
+#define janus_mutex_lock(a) { if(!lock_debug) { janus_mutex_lock_nodebug(a); } else { janus_mutex_lock_debug(a); } }
 /*! \brief Janus mutex unlock without debug */
-#define janus_mutex_unlock_nodebug(a) pthread_mutex_unlock(a);
+#define janus_mutex_unlock_nodebug(a) pthread_mutex_unlock(a)
 /*! \brief Janus mutex unlock with debug (prints the line that unlocked a mutex) */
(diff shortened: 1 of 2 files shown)


lminiero:
The changes you made should be done on both definitions of the macros: how you've implemented it now, it only works if GMutex is used, and would trigger the compiler errors you mentioned if using pthread mutex instead.

That said, I'm not convinced by this trylock stuff. If the cause of the issue is `janus_streaming_rtsp_connect_to_server` trying to lock the mutex because `destroyed` was not `1` yet, wouldn't it be easier to move the `destroyed` check within the mutex? e.g.:

	janus_mutex_lock(&mountpoints_mutex);
	if(g_atomic_int_get(&mp->destroyed)) {
		janus_mutex_unlock(&mountpoints_mutex);
		curl_easy_cleanup(curl);
		g_free(curldata->buffer);
		g_free(curldata);
		return -8;
	}

	/* Parse both video and audio first before proceed to setup as curldata will be reused */

lionelnicolas:
> The changes you made should be done on both definitions of the macros: how you've implemented it now, it only works if GMutex is used, and would trigger the compiler errors you mentioned if using pthread mutex instead.

Ok yeah I missed that `#ifdef`. Fixed.

> Edit: probably not, since this would _always_ perform the lock, which is what causes the issue now when `destroyed` is still 0... I'll have to think about this.

Yes exactly. Using `trylock` make that lock attempt cancellable depending on the `destroyed` value, if the thread is stuck in that `while` for a while. On a very loaded streaming plugin API, I realized that acquiring the lock on this mutex can take some time (I already observed ~ 200 threads stuck on waiting for that lock, which was increasing the probability of having the value of `destroyed` modified during that wait time.

Using that `trylock` logic, I wasn't able to reproduce the deadlock, and I saw `Destroying mountpoint while trying to reconnect, aborting` log appearing multiple times, so this code was avoiding the deadlock.

lminiero:
Makes sense, thanks for the explaination! I tried compiling both with GMutex and pthread mutex and so no problems, so this is good to merge for me :+1:

PR #2702, opened 2021-06-14
demos: update webrtc adapter 8.0.0

tmatth:
The issue with webrtc-adapter tags not being published to cdnjs has been resolved, see:
https://github.com/cdnjs/packages/pull/746

Tested locally (note: I don't have edge).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,6 +33,6 @@
     "tests"
   ],
   "dependencies": {
-    "webrtc-adapter": "7.7.0"
+    "webrtc-adapter": "8.0.0"
   }
 }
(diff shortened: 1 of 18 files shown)


bower.json
```diff
@@ -33,6 +33,6 @@
     "tests"
   ],
   "dependencies": {
-    "webrtc-adapter": "7.7.0"
+    "webrtc-adapter": "8.0.0"
```
lminiero, line 36: Did you test this too? I remember bower having issues because the version available online wasn't available there too.

lminiero (commented):
Thanks! I'll have to make some tests to make sure nothing break in the demos, before merging.

bower.json
tmatth: Just running `npm install` within the janus directory, but worked as expected.

lminiero:
Ok, we finally managed to test with Edge, which worked (at least what we tried) so I'll merge, thanks :+1:

lminiero:
I should clarify that we tested with the _new_ Chromium-based Edge, not the old one: but since I don't really care about that one, I think we can safely ignore it.

PR #2703, opened 2021-06-14
Fix SIP plugin missed_call event call_id field name typo

ycherniavskyi:
In all events except `missed_call` used `call_id` field name for SIP Call-ID. And even docs of `missed_call` points `call_id` field name. So `callid` seems to be a typo here.


lminiero:
Fair point, it looks like the name of the property was changed in #2679 (my fault for not catching that). Mentioning @nicolasduteil who authored the PR, and so may need to be aware the property name is being changed (in case he uses it somewhere).

nicolasduteil:
Yep, my mistake. Sorry about that

lminiero:
No worries! Just wanted you to be aware of the new change.

PR #2706, opened 2021-06-15
Fix missing secret check on room enable_recording

soudhaf:
enable_recoding should check against check_modify permission not check_join permission


januscla:
Thanks for your contribution, @soudhaf! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:

PR #2707, opened 2021-06-16
Add ability to specify recordings folder in AudioBridge

lminiero:
This extends the `enable_recording` feature added by @rajneeshksoni to the AudioBridge plugin in #2674, by adding an additional property, called `record_dir`, where you can specify where the `.wav` recordings should be stored. It's available both in room creation (API or jcfg file) and when using `enable_recording`: when provided, it makes `record_file` a relative path. Notice this only impacts `.wav` recordings, and not the individual recording via `.mjr` that we do support in the AudioBridge: it may make sense to enforce it for those too, but maybe later.

Tested briefly but it seems to be working for me, feedback welcome.

### DIFF
--- a/conf/janus.plugin.audiobridge.jcfg.sample
+++ b/conf/janus.plugin.audiobridge.jcfg.sample
@@ -13,6 +13,7 @@
 # default_prebuffering = number of packets to buffer before decoding each particiant (default=6)
 # record = true|false (whether this room should be recorded, default=false)
 # record_file = "/path/to/recording.wav" (where to save the recording)
+# record_dir = "/path/to/" (path to save the recording to, makes record_file a relative path if provided)
 # allow_rtp_participants = true|false (whether participants should be allowed to join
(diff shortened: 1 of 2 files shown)


rajneeshksoni:
I like function with "record_dir" this makes it much easier to use.
I tested it and found a issue if during enable_recoding only "record_dir" is provided and "record_file" is not provided, API return success, record_dir directory is created but audio file is generated in the location where janus is launched.
so either api should return error or we should handle default file with user specified record_dir.

lminiero:
Good catch: `record_dir` with no `record_file` is definitely allowed, it must create the filename randomly using your code, but place it in that folder. I'll fix it shortly. Thanks for the feedback!

lminiero:
The commit above shoud fix it (haven't tested though).

rajneeshksoni:
Tested, now it is working fine.

lminiero:
Ack, merging then.

PR #2708, opened 2021-06-22
Manually add SIP Contact header for calls when using Sofia >= 1.13 (fixes #2439, replaces #2597)

### LINKED ISSUE #2439: missing contact header in sip invite
We use janus-gateway master with websockets and SIP plugin enabled. 
We use share/janus/demos to test sip registeration and sip call.
After we disable sips and force using of udp by modifying siptest.js, sip registration succeeds.
But sip call fails with the following error:  Missing Contact Header

Has anyone experienced this? I'm not sure what additional info to provide here but can provide more if needed.

We use centos 7.4 and the sofia version:
sofia-sip.x86_64                            1.13.0-14.d10a3d268c     @freeswitch
sofia-sip-devel.x86_64                      1.13.0-14.d10a3d268c     @freeswitch

Here is console output:
nta.c:3305 agent_recv_response() nta: received 400 Missing Contact Header for INVITE (28286682)
nta.c:3372 agent_recv_response() nta: 400 Missing Contact Header is going to a transaction

Below is the sip invite message captured by freeswitch, the Contact header is NOT found. 

<details><summary>SDP</summary>

```
INVITE sip:1001@sbc03.onecall.vip:16385 SIP/2.0
Via: SIP/2.0/UDP 172.16.0.232:36673;rport;branch=z9hG4bKS79HeK1gt6mNr
Max-Forwards: 70
From: "1000" <sip:1000@121.196.44.143>;tag=jZ5Xt3y5cpUSc
To: <sip:1001@sbc03.onecall.vip:16385>
Call-ID: ILwMIAZNwDveNJ45aIWDRJH
CSeq: 28285324 INVITE
User-Agent: Cool WebRTC Gateway
Allow: INVITE, ACK, BYE, CANCEL, OPTIONS, UPDATE, REFER, MESSAGE, INFO, NOTIFY
Supported: replaces
Content-Type: application/sdp
Content-Disposition: session
Content-Length: 958

v=0
o=- 3731379636021817265 2 IN IP4 1.1.1.1
s=-
t=0 0
m=audio 40676 RTP/AVP 111 103 104 9 0 8 106 105 13 110 112 113 126
a=rtpmap:104 ISAC/32000
a=rtpmap:8 PCMA/8000
a=rtpmap:106 CN/32000
a=rtpmap:105 CN/16000
a=rtpmap:13 CN/8000
a=rtpmap:110 telephone-event/48000
a=rtpmap:112 telephone-event/32000
a=rtpmap:113 telephone-event/16000
a=rtpmap:126 telephone-event/8000
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=rtpmap:0 PCMU/8000
a=rtpmap:8 PCMA/8000
a=rtpmap:106 CN/32000
a=rtpmap:105 CN/16000
a=rtpmap:13 CN/8000
a=rtpmap:110 telephone-event/48000
a=rtpmap:112 telephone-event/32000
a=rtpmap:113 telephone-event/16000
a=rtpmap:126 telephone-event/8000
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
a=rtcp-fb:111 transport-cc
```
</details>

lminiero:
Apparently, newer versions of Sofia SIP as maintained by Freeswitch (>= 1.13) don't add a Contact header anymore for calls that are not associated with a previous registration: this is explained in #2439. A first patch was contributed in #2597, which IMHO was a bit too complex. This PR tries to address it in a "simpler" way.

First of all, we check the Sofia SIP version at startup: if it's 1.12.x or older, we do nothing at all (as the Contact header is always there for me). Otherwise, rather than invoking `nua_get_params` at every call as #2597 did, we only do it once when creating the stack. Then, we conditionally add the Contact header _only_ for guest calls, and not all of them as the original PR did.

Not sure if we should do the same for other requests as well (e.g., INFO, MESSAGE, etc.), or if my assumption that only guest users are impacted is correct. I don't have 1.13.x installed so I can't test myself: if you care about this, please test in different conditions (including, if possible, helper sessions) and let us know if this addresses the issue.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -657,6 +657,7 @@
 #include <sofia-sip/url.h>
 #include <sofia-sip/tport_tag.h>
 #include <sofia-sip/su_log.h>
+#include <sofia-sip/sofia_features.h>
 
 #include "../debug.h"
 #include "../apierror.h"
@@ -837,6 +838,8 @@ static uint16_t rtp_range_max = 60000;
 static int dscp_audio_rtp = 0;
 static int dscp_video_rtp = 0;
 
+static gboolean query_contact_header = FALSE;
+
 static GThread *handler_thread;
 static void *janus_sip_handler(void *data);
 static void janus_sip_hangup_media_internal(janus_plugin_session *handle);
@@ -928,6 +931,7 @@ struct ssip_s {
 	su_root_t *s_root;
 	nua_t *s_nua;
 	nua_handle_t *s_nh_r, *s_nh_i, *s_nh_m;
+	char *contact_header;	/* Only needed for Sofia SIP >= 1.13 */
 	GHashTable *subscriptions;
 	janus_mutex smutex;
 	struct janus_sip_session *session;
@@ -1088,6 +1092,7 @@ static void janus_sip_session_free(const janus_refcount *session_ref) {
 			g_hash_table_unref(session->stack->subscriptions);
 		session->stack->subscriptions = NULL;
 		janus_mutex_unlock(&session->stack->smutex);
+		g_free(session->stack->contact_header);
 		g_free(session->stack);
 		session->stack = NULL;
 	}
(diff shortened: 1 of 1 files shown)


flaviogrossi:
i can confirm this fixes the issue for us, tested compiling with sofia 1.13 from freeswitch

Thanks for looking into this, i can confirm that including this patch the contact header is correctly added in the INVITE and other requests, like MESSAGE or responses like the 200 OK after a OPTIONS request from the server

However the issue impacts also registered users, not only guest. But after removing the guest check in this PR, everything works fine even when registering to the server before placing the call.

I did not (yet) test helper sessions though, will update you on this

flaviogrossi:
I quickly tested helper sessions using the demo in the html/ directory and everything seems to work

lminiero:
@flaviogrossi thanks for the tests! I'll remove the `guest` check then.

lminiero:
Thanks to Flavio's tests, I think this can be merged now :+1:

PR #2709, opened 2021-06-23
SIP plugin: Fix sending BYE

ihusejnovic:
Hello,
I noticed that we sometimes receive the following error from sofia sip when the call should be terminated.
`[nua_i_error]: 500 Responding to a Non-Existing Request`

After further investigation, I found out that this is happening when ICE or DTLS negotiation fails. The problem was in this condition if BYE or 480 should be sent. It should use method `janus_sip_call_is_established` because `session->established` is set to `TRUE` only when ICE and DTLS negotiation finish. 

I already tested this and it's working fine but the feedback is welcome.

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -2584,7 +2584,7 @@ static void janus_sip_hangup_media_internal(janus_plugin_session *handle) {
 		session->media.on_hold = FALSE;
 
 		/* Send a BYE or respond with 480 */
-		if(g_atomic_int_get(&session->established) || session->status == janus_sip_call_status_inviting)
+		if(janus_sip_call_is_established(session) || session->status == janus_sip_call_status_inviting)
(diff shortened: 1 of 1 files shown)


lminiero:
This feels like deja-vù: haven't we discussed this already in the past? IIRC there were a couple of "conflicting" PRs that tried to address the same thing in different ways.

lminiero:
It may be #2554.

ihusejnovic:
https://github.com/meetecho/janus-gateway/pull/2554 addressed the issue when the session is destroyed in "sip inviting" status. This one is fixing the issue when the SIP call is established by SIP protocol (INVITE sent and 200 OK received or INVITE received and 200 OK sent), but the webrtc side is not. In this case, SIP BYE should be sent too.

lminiero:
Makes sense :+1: 
Merging then, thanks!

PR #2714, opened 2021-06-24
makefile: Use PKG_CONFIG variable to call system dependent pkg-config

### LINKED ISSUE #2713: Cross-compilation fails on Debian because `pkg-build` is used directly to to create `libnice_version_string`
This is the line in `Makefile.am`:

https://github.com/meetecho/janus-gateway/blob/23535f02ac15d8b5bf6b4a7dfb51c7cadc4bad07/Makefile.am#L211

Not sure why but at least on Debian the `PKG_CONFIG_PATH` is not set however it provides wrappers like `aarch64-linux-gnu-pkg-config` setup which are set by `PKG_CONFIG` variable.

I think changing that line to use `"$(PKG_CONFIG)"` fixes the issue. I tested ARM64 cross-compile on Debian and AMD64 on Arch Linux and it seems to work.

igo95862:
Fixes cross-compilation on Debian and seems to not break
compilation on Arch Linux

Closes #2713


januscla:
Thanks for your contribution, @igo95862! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

igo95862:
Can't be bothered with CLA for a simple fix. It is described in issue anyway.

PR #2715, opened 2021-06-25
Videoroom API "list" endpoint: update docs, add is_private

jmfotokite:
The videorooms "list" endpoint API docs have gotten quite out of sync with reality (in the C source code).

Additionally, add "is_private" to the object returned, because since commit adb77a10e, it is not always clear if a room returned in the list is private or not.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -362,8 +362,7 @@ room-<unique room ID>: {
  *
 \verbatim
 {
-	"request" : "list",
-	"admin_key" : "..."
+	"request" : "list"
 }
 \endverbatim
  *
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -357,14 +357,13 @@ room-<unique room ID>: {
 \endverbatim
  *
  * To get a list of the available rooms you can make use of the \c list request.
- *
  * \c admin_key is optional. If included and correct, rooms configured/created
  * as private will be included in the list as well.
  *
 \verbatim
 {
 	"request" : "list",
-	"admin_key" : "..." // optional
+	"admin_key" : "..."
 }
 \endverbatim
  *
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @jmfotokite! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

plugins/janus_videoroom.c
```diff
@@ -356,13 +356,15 @@ room-<unique room ID>: {
 }
 \endverbatim
  *
- * To get a list of the available rooms (excluded those configured or
- * created as private rooms) you can make use of the \c list request,
- * which has to be formatted as follows:
+ * To get a list of the available rooms you can make use of the \c list request.
+ *
```
lminiero, line 360: Please remove this extra line, or doxygen will create a new paragraph. Since the text below is a continuation/expansion to the intro to `list` itself, it makes sense to keep in contiguous.

plugins/janus_videoroom.c
```diff
@@ -356,13 +356,15 @@ room-<unique room ID>: {
...
+ * \c admin_key is optional. If included and correct, rooms configured/created
+ * as private will be included in the list as well.
  *
 \verbatim
 {
-	"request" : "list"
+	"request" : "list",
+	"admin_key" : "..." // optional
```
lminiero, line 367: I'm not sure this is required, since you already explain it in the text above. There are other requests where `admin_key` can be used, and we don't list it there either.

plugins/janus_videoroom.c
```diff
@@ -376,16 +378,29 @@ room-<unique room ID>: {
 			"room" : <unique numeric ID>,
 			"description" : "<Name of the room>",
 			"pin_required" : <true|false, whether a PIN is required to join this room>,
+			"is_private" : <whether this room is 'private' (as in hidden) or not>,
```
lminiero, line 381: This is missing `true|false, ` to be consistent with others (like the one above).

plugins/janus_videoroom.c
```diff
@@ -376,16 +378,29 @@ room-<unique room ID>: {
...
+			"rec_dir" : "<if recording, the path where the .mjr files are being saved>",
 			"lock_record" : <true|false, whether the room recording state can only be changed providing the secret>,
 			"num_participants" : <count of the participants (publishers, active or not; not subscribers)>
+			"audiolevel_ext": <true|false whether the ssrc-audio-level extension must be negotiated or not for new publishers>,
+			"audiolevel_event": <true|false whether to emit event to other users about audiolevel>,
+			"audio_active_packets": <amount of packets with audio level for checkup (optional, only if audiolevel_event is true)>,
+			"audio_level_average": <average audio level (optional, only if audiolevel_event is true)>,
+			"videoorient_ext": <true|false whether the video-orientation extension must be negotiated or not for new publishers>,
+			"playoutdelay_ext": <true|false whether the playout-delay extension must be negotiated or not for new publishers>,
+			"transport_wide_cc_ext": <true|false whether the transport wide cc extension must be negotiated or not for new publishers>
```
lminiero, line 403: Most of the lines you added are missing a comma after `true|false`, which should be there for consistency with the other fields.

lminiero (commented):
Thanks, I added some comments inline :+1: 
As a side note, please don't forget to sign the CLA, or we won't be able to merge this.

plugins/janus_videoroom.c
jmfotokite: done

plugins/janus_videoroom.c
jmfotokite: removed

plugins/janus_videoroom.c
jmfotokite: done

plugins/janus_videoroom.c
jmfotokite: added

jmfotokite:
Have addressed the comments. For the CLA, I'll need a little while more.

plugins/janus_videoroom.c
```diff
@@ -376,16 +377,29 @@ room-<unique room ID>: {
 			"room" : <unique numeric ID>,
 			"description" : "<Name of the room>",
 			"pin_required" : <true|false, whether a PIN is required to join this room>,
+			"is_private" : <true|false, whether this room is 'private' (as in hidden) or not>,
 			"max_publishers" : <how many publishers can actually publish via WebRTC at the same time>,
 			"bitrate" : <bitrate cap that should be forced (via REMB) on all publishers by default>,
-			"bitrate_cap" : <true|false, whether the above cap should act as a limit to dynamic bitrate changes by publishers>,
+			"bitrate_cap" : <true|false, whether the above cap should act as a limit to dynamic bitrate changes by publishers (optional)>,
```
lminiero, line 383: Not sure why you added the optional there: basically all the other propeerties are as well. Please remove this.

plugins/janus_videoroom.c
```diff
@@ -356,13 +356,14 @@ room-<unique room ID>: {
...
- * which has to be formatted as follows:
+ * To get a list of the available rooms you can make use of the \c list request.
+ * \c admin_key is optional. If included and correct, rooms configured/created
+ * as private will be included in the list as well.
  *
 \verbatim
 {
-	"request" : "list"
+	"request" : "list",
+	"admin_key" : "..."
```
lminiero, line 366: As said in my previous review, for consistency this shouldn't be in the snippet. The text above already explains it can be added.

plugins/janus_videoroom.c
lminiero: I see some of the other properties have it too, possibly because of copy/pasting from the config documentation. I think they should all be removed, since they incorrectly suggest other properties are mandatory instead, many of which aren't.

lminiero (commented):
Thanks, added some other comments inline.

plugins/janus_videoroom.c
jmfotokite: sorry, this was overlooked. Is gone now.

plugins/janus_videoroom.c
jmfotokite: I can remove the "optional" part if you want.

That said, if you compare to the actual code here [here](https://github.com/meetecho/janus-gateway/blob/36b85ee61806c92caa267d39a64094375007b4c7/plugins/janus_videoroom.c#L3764), you will notice that all the fields which I have not marked with `(optional)` are always set, so the 

> since they incorrectly suggest other properties are mandatory instead, many of which aren't.

part is not true, unless there is some logic somewhere which removes empty fields in returned JSON which I did not catch.

lminiero: Fair point, I got confused with `create` :slightly_smiling_face:

lminiero:
Thanks for the fixes! Any update on the CLA?

jmfotokite:
> Thanks for the fixes! Any update on the CLA?

Sorry for the delay, expect it next week or so.

lminiero:
Thanks, merging then :v:

PR #2724, opened 2021-07-05
Add pause/resume recording functionality to Record&Play and SIP plugins

isnumanagic:
This PR adds pause/resume functionality to Janus recorder, and implements signaling for it in Record&Play and SIP plugins.

This functionality is useful when relaying sensitive information during recorded meeting. It is implemented by skipping writing of RTP packets during the pause duration. On resume we request PLI to fix any video artifacts that can occur due to skipped frames in recording.

### DIFF
--- a/html/recordplaytest.html
+++ b/html/recordplaytest.html
@@ -92,8 +92,7 @@
 								<h3 class="panel-title">
 									<span id="videotitle">Remote Video</span>
 									<button class="btn-xs btn-danger pull-right" autocomplete="off" id="stop">Stop</button>
-									<button class="btn-xs btn-primary pull-right" hidden autocomplete="off" id="resume">Resume</button>
-									<button class="btn-xs btn-primary pull-right" autocomplete="off" id="pause">Pause</button>
+									<button class="btn-xs btn-primary pull-right" autocomplete="off" id="pause-resume">Pause</button>
 								</h3>
 							</div>
 							<div class="panel-body" id="videobox"></div>
--- a/html/recordplaytest.html
+++ b/html/recordplaytest.html
@@ -89,7 +89,12 @@
 					<div class="col-md-6 hide" id="video">
 						<div class="panel panel-default">
 							<div class="panel-heading">
-								<h3 class="panel-title"><span id="videotitle">Remote Video</span> <button class="btn-xs btn-danger pull-right" autocomplete="off" id="stop">Stop</button></h3>
+								<h3 class="panel-title">
+									<span id="videotitle">Remote Video</span>
+									<button class="btn-xs btn-danger pull-right" autocomplete="off" id="stop">Stop</button>
+									<button class="btn-xs btn-primary pull-right" hidden autocomplete="off" id="resume">Resume</button>
+									<button class="btn-xs btn-primary pull-right" autocomplete="off" id="pause">Pause</button>
+								</h3>
 							</div>
 							<div class="panel-body" id="videobox"></div>
 						</div>
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -497,18 +497,15 @@ function startRecording() {
 					recordplay.hangup();
 				}
 			});
-
-			$('#pause').on('click', () => {
+		$('#pause-resume').unbind('click').on('click', function() {
+			if($(this).text() === 'Pause') {
 				recordplay.send({message: {request: 'pause'}});
-				$('#pause').attr('hidden', true);
-				$('#resume').attr('hidden', false);
-			});
-
-			$('#resume').on('click', () => {
+				$(this).text('Resume');
+			} else {
 				recordplay.send({message: {request: 'resume'}});
-				$('#pause').attr('hidden', false);
-				$('#resume').attr('hidden', true);
-			});
+				$(this).text('Pause');
+			}
+		});
 	});
 }
 
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -453,6 +453,7 @@ function startRecording() {
 		$('#list').unbind('click').attr('disabled', true);
 		$('#recset').attr('disabled', true);
 		$('#recslist').attr('disabled', true);
+		$('#pause-resume').removeClass('hide');
 
 		// bitrate and keyframe interval can be set at any time:
 		// before, after, during recording
@@ -524,6 +525,7 @@ function startPlayout() {
 	$('#list').unbind('click').attr('disabled', true);
 	$('#recset').attr('disabled', true);
 	$('#recslist').attr('disabled', true);
+	$('#pause-resume').addClass('hide');
 	var play = { request: "play", id: parseInt(selectedRecording) };
 	recordplay.send({ message: play });
 }
--- a/html/recordplaytest.js
+++ b/html/recordplaytest.js
@@ -497,6 +497,18 @@ function startRecording() {
 					recordplay.hangup();
 				}
 			});
+
+			$('#pause').on('click', () => {
+				recordplay.send({message: {request: 'pause'}});
+				$('#pause').attr('hidden', true);
+				$('#resume').attr('hidden', false);
+			});
+
+			$('#resume').on('click', () => {
+				recordplay.send({message: {request: 'resume'}});
+				$('#pause').attr('hidden', false);
+				$('#resume').attr('hidden', true);
+			});
 	});
 }
 
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -2014,17 +2014,17 @@ playdone:
 			JANUS_LOG(LOG_VERB, "Record&Play: Got pause/resume request\n");
 			if(session->recording) {
 				gboolean pause = !strcasecmp(request_text, "pause");
+				result = json_object();
+				json_object_set_new(result, "status", json_string(pause ? "paused" : "resumed"));
+				json_object_set_new(result, "id", json_integer(session->recording->id));
+				/* Also notify event handlers */
+				if(notify_events && gateway->events_is_enabled()) {
+					json_t *info = json_object();
+					json_object_set_new(info, "event", json_string(pause ? "paused" : "resumed"));
+					json_object_set_new(info, "id", json_integer(session->recording->id));
+					gateway->notify_event(&janus_recordplay_plugin, session->handle, info);
+				}
 				if (g_atomic_int_compare_and_exchange(&session->recording->paused, !pause, pause)) {
-					result = json_object();
-					json_object_set_new(result, "status", json_string(pause ? "paused" : "resumed"));
-					json_object_set_new(result, "id", json_integer(session->recording->id));
-					/* Also notify event handlers */
-					if(notify_events && gateway->events_is_enabled()) {
-						json_t *info = json_object();
-						json_object_set_new(info, "event", json_string(pause ? "paused" : "resumed"));
-						json_object_set_new(info, "id", json_integer(session->recording->id));
-						gateway->notify_event(&janus_recordplay_plugin, session->handle, info);
-					}
 					if(pause) {
 						janus_recorder_pause(session->arc);
 						janus_recorder_pause(session->vrc);
@@ -2035,8 +2035,6 @@ playdone:
 						janus_recorder_resume(session->drc);
 						gateway->send_pli(session->handle);
 					}
-				} else {
-					JANUS_LOG(LOG_VERB, "Record&Play: Tried to pause/resume recording of same state, ignoring\n");
 				}
 			} else {
 				JANUS_LOG(LOG_VERB, "Record&Play: Not recording, ignoring pause/resume request\n");
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -1104,7 +1104,8 @@ struct janus_plugin_result *janus_recordplay_handle_message(janus_plugin_session
 		json_object_set_new(response, "settings", settings);
 		goto plugin_response;
 	} else if(!strcasecmp(request_text, "record") || !strcasecmp(request_text, "play")
-			|| !strcasecmp(request_text, "start") || !strcasecmp(request_text, "stop")) {
+			|| !strcasecmp(request_text, "start") || !strcasecmp(request_text, "stop")
+			|| !strcasecmp(request_text, "pause") || !strcasecmp(request_text, "resume")) {
 		/* These messages are handled asynchronously */
 		janus_recordplay_message *msg = g_malloc(sizeof(janus_recordplay_message));
 		msg->handle = handle;
@@ -2007,6 +2008,32 @@ playdone:
 			}
 			/* Tell the core to tear down the PeerConnection, hangup_media will do the rest */
 			gateway->close_pc(session->handle);
+		} else if (!strcasecmp(request_text, "pause") || !strcasecmp(request_text, "resume")) {
+			JANUS_LOG(LOG_VERB, "Record&Play: Got pause/resume request\n");
+			int pause = !strcasecmp(request_text, "pause");
+			result = json_object();
+			json_object_set_new(result, "status", json_string(pause ? "paused" : "resumed"));
+			if(session->recording) {
+				json_object_set_new(result, "id", json_integer(session->recording->id));
+				/* Also notify event handlers */
+				if(notify_events && gateway->events_is_enabled()) {
+					json_t *info = json_object();
+					json_object_set_new(info, "event", json_string(pause ? "paused" : "resumed"));
+					if(session->recording)
+						json_object_set_new(info, "id", json_integer(session->recording->id));
+					gateway->notify_event(&janus_recordplay_plugin, session->handle, info);
+				}
+				if(pause) {
+					janus_recorder_pause(session->arc);
+					janus_recorder_pause(session->vrc);
+					janus_recorder_pause(session->drc);
+				} else {
+					janus_recorder_resume(session->arc);
+					janus_recorder_resume(session->vrc);
+					janus_recorder_resume(session->drc);
+					gateway->send_pli(session->handle);
+				}
+			}
 		} else {
 			JANUS_LOG(LOG_ERR, "Unknown request '%s'\n", request_text);
 			error_code = JANUS_RECORDPLAY_ERROR_INVALID_REQUEST;
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -2010,17 +2010,16 @@ playdone:
 			gateway->close_pc(session->handle);
 		} else if (!strcasecmp(request_text, "pause") || !strcasecmp(request_text, "resume")) {
 			JANUS_LOG(LOG_VERB, "Record&Play: Got pause/resume request\n");
-			int pause = !strcasecmp(request_text, "pause");
-			result = json_object();
-			json_object_set_new(result, "status", json_string(pause ? "paused" : "resumed"));
 			if(session->recording) {
+				gboolean pause = !strcasecmp(request_text, "pause");
+				result = json_object();
+				json_object_set_new(result, "status", json_string(pause ? "paused" : "resumed"));
 				json_object_set_new(result, "id", json_integer(session->recording->id));
 				/* Also notify event handlers */
 				if(notify_events && gateway->events_is_enabled()) {
 					json_t *info = json_object();
 					json_object_set_new(info, "event", json_string(pause ? "paused" : "resumed"));
-					if(session->recording)
-						json_object_set_new(info, "id", json_integer(session->recording->id));
+					json_object_set_new(info, "id", json_integer(session->recording->id));
 					gateway->notify_event(&janus_recordplay_plugin, session->handle, info);
 				}
 				if(pause) {
@@ -2033,6 +2032,8 @@ playdone:
 					janus_recorder_resume(session->drc);
 					gateway->send_pli(session->handle);
 				}
+			} else {
+				JANUS_LOG(LOG_VERB, "Record&Play: Not recording, ignoring pause/resume request\n");
 			}
 		} else {
 			JANUS_LOG(LOG_ERR, "Unknown request '%s'\n", request_text);
--- a/plugins/janus_recordplay.c
+++ b/plugins/janus_recordplay.c
@@ -418,6 +418,7 @@ typedef struct janus_recordplay_recording {
 	char *offer;				/* The SDP offer that will be sent to watchers */
 	gboolean e2ee;				/* Whether media in the recording is encrypted, e.g., using Insertable Streams */
 	GList *viewers;				/* List of users watching this recording */
+	volatile gint paused;		/* Whether this recording is paused */
 	volatile gint completed;	/* Whether this recording was completed or still going on */
 	volatile gint destroyed;	/* Whether this recording has been marked as destroyed */
 	janus_refcount ref;			/* Reference counter */
@@ -1635,6 +1636,7 @@ static void *janus_recordplay_handler(void *data) {
 			rec->acodec = JANUS_AUDIOCODEC_NONE;
 			rec->vcodec = JANUS_VIDEOCODEC_NONE;
 			rec->e2ee = e2ee;
+			g_atomic_int_set(&rec->paused, 0);
 			g_atomic_int_set(&rec->destroyed, 0);
 			g_atomic_int_set(&rec->completed, 0);
 			janus_refcount_init(&rec->ref, janus_recordplay_recording_free);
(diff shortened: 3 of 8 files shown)


lminiero:
I remember this was attempted already in #918, which I had several doubts on. I think your patch has a similar issue in how recordings are handled, because the way you're doing it in SIP and Record&Play this will simply introduce big holes in the recording when postprocessing: when you pause and resume after a while, packets will have huge timestamp/seq number changes, which will translate to either a ton of silence, or missing video.

isnumanagic:
Yes, that is by design.

isnumanagic:
Perhaps this can be ammended by writing silence timespans into the recording header, so that after or during post processing those segments can be skipped?

lminiero:
> Yes, that is by design.

I don't think that's correct. A pause/resume should be seamless in the recording too.

> Perhaps this can be ammended by writing silence timespans into the recording header, so that after or during post processing those segments can be skipped?

MJR headers can't be edited after they're written, and I don't like the idea of making a lot of changes to janus-pp-rec to accomodate this.

isnumanagic:
I see the packet timestamps are calculated by Janus, then the solution can be to track the total paused time and subtracting it when calculating timestamps?

lminiero:
The solution is to write packets to the recorder that have consistent RTP headers, as we do in other plugins e.g. for source switching and RTP forwarders.

isnumanagic:
Going through the code i see you use `janus_rtp_switching_context` struct for tracking timestamps and seq-numbers, and `janus_rtp_header_update` method to make changes to header. I'll try to do the same here and will update the PR.

lminiero:
My guess is that, once you add the context in the mix, you can just trigger its behaviour by changing its SSRC before a resume (which will trick it into thinking the source changed and so headers need to be updated).

isnumanagic:
Updated PR with requested changes. Just changing the ssrc did not work - I had to add time offset information.

lminiero:
> Updated PR with requested changes. Just changing the ssrc did not work - I had to add time offset information.

From a quick glance to the code I don't think you're using the RTP context correctly. Timestamp offset management is already done automatically by RTP contexts, when the SSRC changes: it looks to me like you're just resetting the sequence number instead, which is why you're forced to manually fix the timestamp. When recording is resumed, it's the SSRCs in the context you have to change (`a_last_ssrc`, `v_last_ssrc`), not the one in the packet: this way, even if the packet is using the same SSRC it always used, the context will think it changed, and it will trigger the rewriting logic.

isnumanagic:
I tried to just reset the ssrc by changing the one in the context, but as I said it does not work - the silence during the pause duration remained in the recording. This is because the base timestamp is reset, but the offset is then calcualted and applied (denoted by the `How much time since the last audio RTP packet? We compute an offset accordingly` comment). That is why i reset the seq numbering instead of entire ssrc and adjust the timestamp.

lminiero:
Ah I think I understand what you mean now: the SSRC change forces sequence number and timestamp to be updated, but the timestamp will take into account the time that passed, and so a lot of silence will be inserted, which shouldn't indeed happen in this case. That said, I don't think the way you're updating the timestamp offset works, and I don't like the idea of adding new properties to an already crowded structure. It should be possible to do this with the properties already there.

isnumanagic:
I have updated CR as requested. I have managed to remove the silence between pauses by overwriting `a/v_last_time` properties. I have removed `a/v_time_offset` properties I have added to switching context, but kept `a/v_ts_reset` flags as it is more descriptive to do `seq_reset` and `ts_reset` on resume than to change the ssrc, but I can roll back the flags as well if needed.

lminiero:
Thanks! I'll need time to review, since your changes modifies the code of the context update method as well, which can have a huge impact (we use it pretty much everywhere).

isnumanagic:
The context update method has not been functionally changed, I just extracted the `seq_reset` part (which was already there actually, i just removed duplicated code) and `ts_reset` part of ssrc reset into separate if branches for finer control.

isnumanagic:
So basically:
```c
if(ssrc != context->a_last_ssrc) {
  ...
  // ts_reset code
  // seq_reset code
  ...
}
```
became:
```c
if(ssrc != context->a_last_ssrc) {
  ...
  context->a_ts_reset = TRUE;
  context->a_seq_reset = TRUE;
  ...
}
if(context->a_ts_reset) {
  // ts_reset code
}
if(context->a_seq_reset) {
  // seq_reset code
}
```

isnumanagic:
Hey @lminiero, if you have the time please take a look.

lminiero:
It's IETF week so not sure when I'll be able to do a deep dive. I'll try to add some comments to the PR when I can.

html/recordplaytest.js
```diff
@@ -497,6 +497,18 @@ function startRecording() {
...
+				recordplay.send({message: {request: 'pause'}});
+				$('#pause').attr('hidden', true);
+				$('#resume').attr('hidden', false);
+			});
+
+			$('#resume').on('click', () => {
+				recordplay.send({message: {request: 'resume'}});
+				$('#pause').attr('hidden', false);
+				$('#resume').attr('hidden', true);
+			});
```
lminiero, line 511: No need for the empty lines above. As a side note, the way you implemented this won't work for consecutive recordings: new click handlers will be added any time you start one, which means that any time you click the button it will send the same request multiple times. The way we handle this in other demos is via something like:

    .unbind('click').click(function() {

As a side note, I'd prefer the pause/resume to be a single one with a toggle (e.g., like the audio/video on/off buttons in the EchoTest demo, and others), but that's not important.

lminiero: PS: we use the bootstrap `hide` class to hide/show elements. Please use that instead of the manual `hidden` attribute.

plugins/janus_recordplay.c
```diff
@@ -2007,6 +2008,32 @@ static void *janus_recordplay_handler(void *data) {
 			}
 			/* Tell the core to tear down the PeerConnection, hangup_media will do the rest */
 			gateway->close_pc(session->handle);
+		} else if (!strcasecmp(request_text, "pause") || !strcasecmp(request_text, "resume")) {
+			JANUS_LOG(LOG_VERB, "Record&Play: Got pause/resume request\n");
+			int pause = !strcasecmp(request_text, "pause");
```
lminiero, line 2013: Why not `gboolean`?

plugins/janus_recordplay.c
```diff
@@ -2007,6 +2008,32 @@ static void *janus_recordplay_handler(void *data) {
...
+			result = json_object();
+			json_object_set_new(result, "status", json_string(pause ? "paused" : "resumed"));
```
lminiero, line 2015: Shouldn't this check if the current session is in recording mode? It may be idle, or playing something instead: if playing, the session object will still have a `recording` object set, and you'd be sending broken events.

plugins/janus_recordplay.c
```diff
@@ -2007,6 +2008,32 @@ static void *janus_recordplay_handler(void *data) {
...
+			if(session->recording) {
+				json_object_set_new(result, "id", json_integer(session->recording->id));
+				/* Also notify event handlers */
+				if(notify_events && gateway->events_is_enabled()) {
+					json_t *info = json_object();
+					json_object_set_new(info, "event", json_string(pause ? "paused" : "resumed"));
+					if(session->recording)
+						json_object_set_new(info, "id", json_integer(session->recording->id));
+					gateway->notify_event(&janus_recordplay_plugin, session->handle, info);
```
lminiero, line 2024: See above: this would trigger for playback too, which is wrong.

plugins/janus_sip.c
```diff
@@ -4357,10 +4357,11 @@ static void *janus_sip_handler(void *data) {
 				goto error;
 			json_t *action = json_object_get(root, "action");
 			const char *action_text = json_string_value(action);
-			if(strcasecmp(action_text, "start") && strcasecmp(action_text, "stop")) {
-				JANUS_LOG(LOG_ERR, "Invalid action (should be start|stop)\n");
+			if(strcasecmp(action_text, "start") && strcasecmp(action_text, "stop") &&
+				strcasecmp(action_text, "pause") && strcasecmp(action_text, "resume")) {
```
lminiero, line 4361: This needs another tab of indentation, since it's a continuation of the `if` above and is confused with the action below.

plugins/janus_sip.c
```diff
@@ -4502,6 +4503,35 @@ static void *janus_sip_handler(void *data) {
...
+				}
+				if(record_peer_audio) {
+					janus_recorder_resume(&session->arc_peer);
+				}
+				if(record_peer_video) {
+					janus_recorder_resume(&session->vrc_peer);
+				}
+				if(record_video || record_peer_video) {
+					gateway->send_pli(&session->handle);
+				}
```
lminiero, line 4534: All actions above are simple, so the brackets are unneeded.

record.c
```diff
@@ -255,6 +256,33 @@ janus_recorder *janus_recorder_create_full(const char *dir, const char *codec, c
 	return rc;
 }
 
+int janus_recorder_pause(janus_recorder *recorder) {
+	if(!recorder)
+		return -1;
+	janus_mutex_lock_nodebug(&recorder->mutex);
+	g_atomic_int_set(&recorder->paused, TRUE);
+	janus_mutex_unlock_nodebug(&recorder->mutex);
```
lminiero, line 264: There's no need for a mutex, here: it's an atomic operation. Besides, the value should be `1` and not `TRUE`.

record.c
```diff
@@ -255,6 +256,33 @@ janus_recorder *janus_recorder_create_full(const char *dir, const char *codec, c
...
+	if(recorder->type == JANUS_RECORDER_AUDIO) {
+		recorder->context.a_ts_reset = TRUE;
+		recorder->context.a_seq_reset = TRUE;
+		recorder->context.a_last_time = janus_get_monotonic_time();
+	} else if(recorder->type == JANUS_RECORDER_VIDEO) {
+		recorder->context.v_ts_reset = TRUE;
+		recorder->context.v_seq_reset = TRUE;
+		recorder->context.v_last_time = janus_get_monotonic_time();
+	}
+	g_atomic_int_set(&recorder->paused, FALSE);
```
lminiero, line 281: `FALSE` --> `0`.

record.c
```diff
@@ -384,17 +416,33 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
 				res, sizeof(gint64), g_strerror(errno));
 		}
 	}
+	/* Edit packet header if needed */
+	rtp_header *header = (rtp_header *)buffer;
```
lminiero, line 420: `janus_rtp_header`: `rtp_header` is only there for backwards compatibility, as that's how we called it originally, which we want to get rid of.

record.c
```diff
@@ -384,17 +416,33 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
...
+	uint32_t ssrc = ntohl(header->ssrc);
+	uint16_t seq = ntohs(header->seq_number);
+	timestamp = ntohl(header->timestamp);
+	if(recorder->type != JANUS_RECORDER_DATA) {
+		janus_rtp_header_update(header, &recorder->context, recorder->type == JANUS_RECORDER_VIDEO, 0);
+	}
```
lminiero, line 426: Simple action, no need for brackets.

record.c
```diff
@@ -384,17 +416,33 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
 				res, sizeof(gint64), g_strerror(errno));
 		}
 	}
+	/* Edit packet header if needed */
+	rtp_header *header = (rtp_header *)buffer;
+	uint32_t ssrc = ntohl(header->ssrc);
+	uint16_t seq = ntohs(header->seq_number);
+	timestamp = ntohl(header->timestamp);
+	if(recorder->type != JANUS_RECORDER_DATA) {
```
lminiero, line 424: Accessing RTP fields should be done ONLY if it's not a data recorder, so this check needs to be moved up. The way it is, if you're recording a small data message (<12 bytes) this will crash.

record.c
```diff
@@ -384,17 +416,33 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
...
 	/* Save packet on file */
 	int temp = 0, tot = length;
 	while(tot > 0) {
 		temp = fwrite(buffer+length-tot, sizeof(char), tot, recorder->file);
 		if(temp <= 0) {
 			JANUS_LOG(LOG_ERR, "Error saving frame...\n");
+			/* Restore packet header data */
+			header->ssrc = htonl(ssrc);
+			header->seq_number = htons(seq);
+			header->timestamp = htonl(timestamp);
```
lminiero, line 436: See above: only for non-data.

record.c
```diff
@@ -384,17 +416,33 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
...
 			janus_mutex_unlock_nodebug(&recorder->mutex);
 			return -6;
 		}
 		tot -= temp;
 	}
+	/* Restore packet header data */
+	header->ssrc = htonl(ssrc);
+	header->seq_number = htons(seq);
+	header->timestamp = htonl(timestamp);
```
lminiero, line 445: See above: only for non-data.

record.h
```diff
@@ -60,6 +61,10 @@ typedef struct janus_recorder {
 	volatile int header;
 	/*! \brief Whether this recorder instance can be used for writing or not */
 	volatile int writable;
+	/*! \brief Whether writing RTP packets is paused */
+	volatile int paused;
+	/*! \brief RTP switching context for rewriting RTP headers */
+	janus_rtp_switching_context context;
```
lminiero, line 67: I'm not fond of the idea of having a context as part of the recorder itself. Our plugins all have a dedicated context which they use for recordings too, even though I guess in this case we need something that's specific to the recording itself. I'll need to think about this.

record.h
```diff
@@ -60,6 +61,10 @@ typedef struct janus_recorder {
 	volatile int header;
 	/*! \brief Whether this recorder instance can be used for writing or not */
 	volatile int writable;
+	/*! \brief Whether writing RTP packets is paused */
```
lminiero, line 64: s/RTP packets/data

rtp.c
```diff
@@ -601,10 +601,17 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
...
 				context->v_last_ssrc, ssrc);
 			context->v_last_ssrc = ssrc;
+			context->v_ts_reset = TRUE;
+			context->v_seq_reset = TRUE;
+			/* Reset skew compensation data */
+			context->v_new_ssrc = TRUE;
+		}
+		if(context->v_ts_reset) {
+			/* Video timestamp was paused for a while */
+			JANUS_LOG(LOG_VERB, "Video RTP timestamp reset requested");
```
lminiero, line 611: This feels too verbose, maybe `LOG_HUGE` is better.

rtp.c
```diff
@@ -615,11 +622,10 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
 				context->v_last_ts += (guint32)time_diff;
 				JANUS_LOG(LOG_VERB, "Computed offset for video RTP timestamp: %"SCNu32"\n", (guint32)time_diff);
 			}
-			/* Reset skew compensation data */
-			context->v_new_ssrc = TRUE;
 		}
 		if(context->v_seq_reset) {
-			/* Video sequence number was paused for a while: just update that */
+			/* Video sequence number was paused for a while */
+			JANUS_LOG(LOG_VERB, "Video RTP sequence number reset requested");
```
lminiero, line 628: See above.

rtp.c
```diff
@@ -640,10 +646,17 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
...
 				context->a_last_ssrc, ssrc);
 			context->a_last_ssrc = ssrc;
+			context->a_ts_reset = TRUE;
+			context->a_seq_reset = TRUE;
+			/* Reset skew compensation data */
+			context->a_new_ssrc = TRUE;
+		}
+		if(context->a_ts_reset) {
+			/* Audio timestamp was paused for a while */
+			JANUS_LOG(LOG_VERB, "Audio RTP timestamp reset requested");
```
lminiero, line 656: See above.

rtp.c
```diff
@@ -658,11 +671,10 @@ void janus_rtp_header_update(janus_rtp_header *header, janus_rtp_switching_conte
 				context->a_last_ts += (guint32)time_diff;
 				JANUS_LOG(LOG_VERB, "Computed offset for audio RTP timestamp: %"SCNu32"\n", (guint32)time_diff);
 			}
-			/* Reset skew compensation data */
-			context->a_new_ssrc = TRUE;
 		}
 		if(context->a_seq_reset) {
-			/* Audio sequence number was paused for a while: just update that */
+			/* Audio sequence number was paused for a while */
+			JANUS_LOG(LOG_VERB, "Audio RTP sequence number reset requested");
```
lminiero, line 677: See above.

lminiero (commented):
I had another look at the changes: please notice I haven't tested this yet, which will probably not happen before next week.

isnumanagic:
Updated PR with requested changes. If you

isnumanagic:
Hey @lminiero, any updates?

lminiero:
@isnumanagic apologies, I went on vacation and forgot about this :hand_over_mouth: 
I'll have a look at the fixes you made at the time and let you know. Since this has an impact on how mjr files are created (the extra RTP context that is part of the code is a bug change) we'll need time to do proper testing anyway, as we use recordings a lot.

plugins/janus_recordplay.c
```diff
@@ -2007,6 +2008,33 @@ static void *janus_recordplay_handler(void *data) {
 			}
 			/* Tell the core to tear down the PeerConnection, hangup_media will do the rest */
 			gateway->close_pc(session->handle);
+		} else if (!strcasecmp(request_text, "pause") || !strcasecmp(request_text, "resume")) {
+			JANUS_LOG(LOG_VERB, "Record&Play: Got pause/resume request\n");
+			if(session->recording) {
+				gboolean pause = !strcasecmp(request_text, "pause");
```
lminiero, line 2014: I think this should probably check what the state of the recording is, before doing something. As it is, a "pause" on a paused recording will set the paused property again, and will send a new event too, which seems incorrect. I think this is particolarly important due to the way "resume" is implemented in record.c, where the last time is reset every time the method is called.

I guess the right way to do that would be a new property on the `session->recording` object: maybe a volatile int to check and tweak via `g_atomic_int_compare_and_exchange` as we do in a few other places.

plugins/janus_sip.c
```diff
@@ -4502,6 +4503,26 @@ static void *janus_sip_handler(void *data) {
...
+				if(record_audio)
+					janus_recorder_resume(session->arc);
+				if(record_video)
+					janus_recorder_resume(session->vrc);
+				if(record_peer_audio)
+					janus_recorder_resume(session->arc_peer);
+				if(record_peer_video)
+					janus_recorder_resume(session->vrc_peer);
+				if(record_video || record_peer_video)
+					gateway->send_pli(session->handle);
```
lminiero, line 4525: Same here, I guess, even though the property should be part of the SIP plugin session object.

record.c
```diff
@@ -255,6 +256,31 @@ janus_recorder *janus_recorder_create_full(const char *dir, const char *codec, c
...
+		recorder->context.a_ts_reset = TRUE;
+		recorder->context.a_seq_reset = TRUE;
+		recorder->context.a_last_time = janus_get_monotonic_time();
+	} else if(recorder->type == JANUS_RECORDER_VIDEO) {
+		recorder->context.v_ts_reset = TRUE;
+		recorder->context.v_seq_reset = TRUE;
+		recorder->context.v_last_time = janus_get_monotonic_time();
+	}
+	g_atomic_int_set(&recorder->paused, 0);
+	janus_mutex_unlock_nodebug(&recorder->mutex);
```
lminiero, line 280: See discussion above: maybe, besides checks at the plugin level, this code should check the status of the recorder too, and not reset these properties if the recorder is not paused already.

record.c
```diff
@@ -384,17 +414,39 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
 				res, sizeof(gint64), g_strerror(errno));
 		}
 	}
+	/* Edit packet header if needed */
+	janus_rtp_header *header = (janus_rtp_header *)buffer;
+	uint32_t ssrc;
+	uint16_t seq;
```
lminiero, line 420: These two integers should be initialized.

html/recordplaytest.html
```diff
@@ -89,7 +89,11 @@ <h3 class="panel-title">Recorder/Playout</h3>
 					<div class="col-md-6 hide" id="video">
 						<div class="panel panel-default">
 							<div class="panel-heading">
-								<h3 class="panel-title"><span id="videotitle">Remote Video</span> <button class="btn-xs btn-danger pull-right" autocomplete="off" id="stop">Stop</button></h3>
+								<h3 class="panel-title">
+									<span id="videotitle">Remote Video</span>
+									<button class="btn-xs btn-danger pull-right" autocomplete="off" id="stop">Stop</button>
+									<button class="btn-xs btn-primary pull-right" autocomplete="off" id="pause-resume">Pause</button>
```
lminiero, line 95: This button is currently visible also when playing back recordings, which shouldn't happen (especially since it does nothing in that case).

lminiero (commented):
Took me a while, but I found some time to test this: it seemed to work as expected for the brief test I made, even though I didn't push it much. I did notice some UI things, though (pause button still visible when replaying recordings), and some messaging/state things related to how pause/resume are handled that may need fixing (I'll try to make some tests for that too).

isnumanagic:
Uodated PR with requested changes

plugins/janus_recordplay.c
```diff
@@ -2012,25 +2014,29 @@ static void *janus_recordplay_handler(void *data) {
...
+						janus_recorder_resume(session->vrc);
+						janus_recorder_resume(session->drc);
+						gateway->send_pli(session->handle);
+					}
 				} else {
-					janus_recorder_resume(session->arc);
-					janus_recorder_resume(session->vrc);
-					janus_recorder_resume(session->drc);
-					gateway->send_pli(session->handle);
+					JANUS_LOG(LOG_VERB, "Record&Play: Tried to pause/resume recording of same state, ignoring\n");
```
lminiero, line 2039: I'd send the event back anyway, even if we didn't have to change the state, as there may be applications expecting an event back after sending the request.

plugins/janus_sip.c
```diff
@@ -4374,12 +4374,29 @@ static void *janus_sip_handler(void *data) {
...
 				goto error;
 			}
+			gint recorder_bitmap =
+				g_atomic_int_get(&session->arc->paused) |
+				g_atomic_int_get(&session->vrc->paused) << 1 |
+				g_atomic_int_get(&session->arc_peer->paused) << 2 |
+				g_atomic_int_get(&session->vrc_peer->paused) << 3;
+			if ((!strcasecmp(action_text, "pause") && !(~recorder_bitmap & update_bitmap)) ||
+				(!strcasecmp(action_text, "resume") && !(recorder_bitmap & update_bitmap))) {
+				JANUS_LOG(LOG_ERR, "Invalid pause/resume request (at least one of audio, video, peer_audio and peer_video should change recorder state)\n");
```
lminiero, line 4395: I don't think this bitmap approach works, which feels a bit convoluted. First all, this will crash Janus if you're not recording everything (e.g., just the user, just the peer, or video is not active), as in that case some recordings object would be `NULL`. Besides, I suspect this check would break if the provided requests are not aligned with the state, and go in an unrecoverable state.

Considering the state enhancement you added to `record.c` that does make the pause/resume methods idempotent, I think we can keep the code as it was in the SIP plugin for the sake of simplicity, but only send the PLI if we're really resuming (so only adding a single additional check below).

lminiero (commented):
Thanks for the quick fixes! I added a couple more notes: I think we're close to merging this :v:

isnumanagic:
Updated PR with requested changes

plugins/janus_sip.c
```diff
@@ -4521,26 +4504,25 @@ static void *janus_sip_handler(void *data) {
...
-				if(resume_bitmap & 4)
+				if(record_video && !janus_recorder_resume(session->vrc))
+					send_pli = TRUE;
+				if(record_peer_audio)
 					janus_recorder_resume(session->arc_peer);
-				if(resume_bitmap & 8)
-					janus_recorder_resume(session->vrc_peer);
-				if(resume_bitmap & 10)
+				if(record_peer_video && !janus_recorder_resume(session->vrc_peer))
+					send_pli = TRUE;
```
lminiero, line 4524: You don't need `send_pli` here: if we're recording the peer, we need a PLI from the SIP user, not the WebRTC user. I think you can just do a

	if(record_video && !janus_recorder_resume(session->vrc))
		gateway->send_pli(session->handle);

without the additional `send_pli` boolean definition.

lminiero:
LGTM now, thanks for all the fixes! Pinging @atoppi to give a review as well (in case I missed anything), and @alexamirante to check if this might cause issues with the way we handle recordings in production.

atoppi:
this looks good, I just don't get why we need the `ts_reset` flag, is this just a design choice?

atoppi (commented):
left my only question as a separate comment

isnumanagic:
@atoppi It was a design choice. It's more intuitive to set ts_reset and seq_reset flags when resuming recording, than to manually change SSRC, expecting it to reset timestamp and sequence number.

atoppi:
ok, thanks or clarifying, LGTM :+1:

lminiero:
Ack, merging then :+1:  Thanks for the patience!

lminiero:
@isnumanagic you created a monster: I'm way too often using the pause/resume feature just to do funny faces! I see a potential for doing TikTok like videos with WebRTC :rofl:

PR #2729, opened 2021-07-07
Set new remote credentials after restarting ICE (fixes #2672)

### LINKED ISSUE #2672: ice restart fail due to empty credential on libnice master
I've do some test to discover the reason why ice restart fail in janus 0.11.2 (nosip plugin) with libnice master on centos7.

Libnice log, after ice restart , always report that have empty credential (ufrag/pwd) when start connectivity check and the reason is on method [nice_agent_restart](https://gitlab.freedesktop.org/libnice/libnice/-/blob/master/agent/agent.c#L5525) that [reset credential](https://gitlab.freedesktop.org/libnice/libnice/-/blob/master/agent/stream.c#L143)

The reason that libnice have empty credential is janus call [nice_agent_set_remote_credentials](https://github.com/meetecho/janus-gateway/blob/4294f20606050316209bb6d20d8bef1a66272e4d/janus.c#L1523) with new credential before [nice_agent_restart](https://github.com/meetecho/janus-gateway/blob/4294f20606050316209bb6d20d8bef1a66272e4d/janus.c#L1528)

By reverse the order when calling this methods ice restart always goes on

lminiero:
As explained in #2672, apparently new versions of libnice reset the remote credentials after invoking `nice_agent_restart`: since when receiving updated offers that include a restart we first reset the credentials, and then trigger the restart, this actually results in broken sessions, since the restart ends up not having access to remote credentials for the new STUN exchanges. This patch should address that. Notice this only changes the "browser offering the restart" scenario, since when it's Janus offering a restart, we already perform `nice_agent_restart` first, and then get the new remote credentials in an answer later on. That said, it might be a good idea to make sure that still does indeed work even for newer versions of libnice.

Despite the very small change, it's a PR and not a commit since I want to make sure this doesn't introduce regressions for restarts in current versions of libnice, so please test and provide feedback.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -1520,16 +1520,16 @@ int janus_process_incoming_request(janus_request *request) {
 				}
 				if(janus_flags_is_set(&handle->webrtc_flags, JANUS_ICE_HANDLE_WEBRTC_ICE_RESTART)) {
 					JANUS_LOG(LOG_INFO, "[%"SCNu64"] Restarting ICE...\n", handle->handle_id);
-					/* Update remote credentials for ICE */
-					if(handle->stream) {
-						nice_agent_set_remote_credentials(handle->agent, handle->stream->stream_id,
-							handle->stream->ruser, handle->stream->rpass);
-					}
 					/* FIXME We only need to do that for offers: if it's an answer, we did that already */
 					if(offer) {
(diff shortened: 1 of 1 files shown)


lminiero:
Merging.

PR #2730, opened 2021-07-08
[janus-pp-rec] activate keyframe logic for VP9

matthieu77000:
Hi, 

When post-processing vp9 stream into a webm file, it happens that the generated file miss keyframes flag, and make the file difficult to seek for some tools (tested with chrome 91).

Activating it allow the file to be seekable.

Regard,

Matthieu


januscla:
Thanks for your contribution, @matthieu77000! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Good catch, we were indeed just changing the flag for "first keyframe detected", but not setting the variable for "new keyframe detected" as we do for VP8. Merging :+1:

PR #2732, opened 2021-07-11
Fix SIP plugin doxygen command for section reference

ycherniavskyi:
`sipmc` here differently must be reference instead of code.


lminiero:
Good catch, thanks! :+1:

Issue #2734, opened 2021-07-13
Eventing Websocket connection time if destination is not available


JanFellner:
Hi @lminiero 
hi MeetEcho team

as you were complaining about not receiving PRs (At least @fippo told me so ;] ) from our side i am finally stepping into the code.
Actually i wanted to improve something super trivial but in the end my lack of knowledge in libwebsocket hinders me to really make a PR out of it but i provide the code.

We are using Janus as SFU and have an own signalling server upfront.
In order to get everything up and running is thus start janus in advance the the signalling server.
The problem is that it then takes time until janus is able to setup the outbound eventing websocket connection to our signalling server as in the initial startup the connect fails. In the end leading that leads to a total delay of additional 30 seconds to get a new deployed server up an running.

Furthermore if the connection attempt fails janus is always doubling the time for a reconnect which makes it kind of ugly to develop. If the signalling server is not running for 10 minutes i need to wait several minutes for janus to reconnect or i restart janus.

Thus the code tries to solve two things:
- A configurable parameter to define how long the janus should maximum delay a reconnect attempt (max_reconnect_retry_timeout)
- Improve that the reconnect acts faster.

https://github.com/meetecho/janus-gateway/compare/master...JanFellner:feature/CC-2166-make-eventing-websocket-reconnect-time-configurable

The issue i have is that lws_service always blocks for about 25 seconds. I added diagnostic code that shows the issue:

* [Tue Jul 13 16:05:03 2021] woke up (return from lws_service)
* [Tue Jul 13 16:05:03 2021] going to sleep (diving into lws_service) 
* [Tue Jul 13 16:05:03 2021] [ERR] [events/janus_wsevh.c:janus_wsevh_callback:729] Reconnecting failed (reason: Unable to connect), next retry in 1s... 
* [Tue Jul 13 16:05:28 2021] woke up (return from lws_service)
* [Tue Jul 13 16:05:28 2021] [WARN] Reconnecting WebSocketsEventHandler event handler to 10.20.3.220:3000...
* [Tue Jul 13 16:05:28 2021] going to sleep (diving into lws_service)
* [Tue Jul 13 16:05:33 2021] woke up (return from lws_service)
* [Tue Jul 13 16:05:33 2021] going to sleep (diving into lws_service)
* [Tue Jul 13 16:05:33 2021] [ERR] [events/janus_wsevh.c:janus_wsevh_callback:729] Reconnecting failed (reason: Unable to connect), next retry in 2s...

[...]

PR #2736, opened 2021-07-15
Feature/cc 2166 speedup eventing websocket reconnect

JanFellner:
As followup from https://github.com/meetecho/janus-gateway/issues/2734

Changed the implementation for the reconnect from the handling inside the lws_service thread to the scheduler approach as shown in the libsocket example. (https://libwebsockets.org/git/libwebsockets/tree/minimal-examples/ws-client/minimal-ws-client-tx/minimal-ws-client.c)

The current approach is not working properly if the target is not localhost. The connection failed callback is triggered 5 seconds after the connect however the lws_service continues to sleep for 25 seconds. 

Using the lws_sul_schedule the time when the next reconnect has to happen is exactly specified, so no need to do that in the lws_service loop and the loop just does the lws_service loop.

(!) Not yet tested with wsl 3.1 want to do that now but a short feedback in the mean time would be nice.

### DIFF
--- a/events/janus_wsevh.c
+++ b/events/janus_wsevh.c
@@ -710,6 +710,7 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
 			reconnect = FALSE;
 			janus_mutex_init(&ws_client->mutex);
 			lws_callback_on_writable(wsi);
+			JANUS_LOG(LOG_INFO, "WebSocketsEventHandler connected\n");
 			return 0;
 		}
 		case LWS_CALLBACK_CLIENT_CONNECTION_ERROR: {
@@ -806,11 +807,12 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
 #else
 		case LWS_CALLBACK_CLOSED: {
 #endif
-			JANUS_LOG(LOG_INFO, "WebSockets event handler connection closed\n");
+			reconnect_delay = 1;
+			JANUS_LOG(LOG_INFO, "Connection to WebSocketsEventHandler backend closed (next connection attempt in %ds)\n", reconnect_delay);
 			if(ws_client != NULL) {
 				/* Cleanup */
 				janus_mutex_lock(&ws_client->mutex);
-				JANUS_LOG(LOG_INFO, "Destroying WebSockets event handler client\n");
+				JANUS_LOG(LOG_INFO, "Destroying WebSocketsEventHandler client\n");
 				ws_client->wsi = NULL;
 				/* Free the shared buffers */
 				g_free(ws_client->buffer);
@@ -820,8 +822,6 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
 				ws_client->bufoffset = 0;
 				janus_mutex_unlock(&ws_client->mutex);
 			}
-			reconnect_delay = 1;
-			JANUS_LOG(LOG_INFO, "Connection to WebSockets event handler backend closed (next connection attempt in %ds)\n", reconnect_delay);
 			/* Check if we should reconnect */
 			ws_client = NULL;
 			wsi = NULL;
--- a/events/janus_wsevh.c
+++ b/events/janus_wsevh.c
@@ -390,7 +390,6 @@ int janus_wsevh_init(const char *config_path) {
 	info.protocols = protocols;
 	info.gid = -1;
 	info.uid = -1;
-	info.timeout_secs = 10;
 	if(secure)
 		info.options |= LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT;
 	context = lws_create_context(&info);
--- a/events/janus_wsevh.c
+++ b/events/janus_wsevh.c
@@ -46,7 +46,11 @@ json_t *janus_wsevh_handle_request(json_t *request);
 /* Forward declarations */
 void janus_wsevh_schedule_connect_attempt(void);
 void janus_wsevh_calculate_reconnect_delay_on_fail(void);
-static void janus_wsevh_connect_attempt(void* sul);
+// lws_sorted_usec_list_t is defined starting with lws 3.2
+#if !(((LWS_LIBRARY_VERSION_MAJOR == 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || LWS_LIBRARY_VERSION_MAJOR >= 4))
+	#define lws_sorted_usec_list_t void
+#endif
+static void janus_wsevh_connect_attempt(lws_sorted_usec_list_t* sul);
 
 /* Event handler setup */
 static janus_eventhandler janus_wsevh =
@@ -571,27 +575,27 @@ plugin_response:
 		}
 }
 
-#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR > 2) || (LWS_LIBRARY_VERSION_MAJOR > 3)
+#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR >= 4)
 	/*
 	* Websocket thread loop for websocket library newer thant 3.2
-	* The reconnect is handled in a dedicated wsl scheduler janus_wsevh_schedule_connect_attempt
+	* The reconnect is handled in a dedicated lws scheduler janus_wsevh_schedule_connect_attempt
 	*/
 	static void *janus_wsevh_thread(void *data) {
-	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (wsl>=3.2) client thread\n");
+		JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (lws>=3.2) client thread\n");
 		int n = 0;
 		while(n >= 0 && g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping))
 				n = lws_service(context, 0);
 		lws_context_destroy(context);
-		JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (wsl>=3.2) client thread\n");
+		JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (lws>=3.2) client thread\n");
 		return NULL;
 	}
 #else
 	/*
 	* Websocket thread loop for websocket library prior to (less than) 3.2
-	* The reconnect is handled in the loop for wsl < 3.2
+	* The reconnect is handled in the loop for lws < 3.2
 	*/
 	static void *janus_wsevh_thread(void *data) {
-		JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (wsl<3.2) client thread\n");
+		JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (lws<3.2) client thread\n");
 		while(g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping)) {
 			/* Loop until we have to stop */
 			if(!reconnect) {
@@ -609,12 +613,16 @@ plugin_response:
 				}
 				ws_client = NULL;
 				janus_wsevh_connect_attempt(NULL);
-				if (!wsi)
+				if (!wsi) {
 					janus_wsevh_calculate_reconnect_delay_on_fail();
+					disconnected = janus_get_monotonic_time();
+					reconnect = TRUE;
+					JANUS_LOG(LOG_WARN, "WebSocketsEventHandler: Error attempting connection... (next retry in %ds)\n", reconnect_delay);
+				}
 			}
 		}
 		lws_context_destroy(context);
-		JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (wsl<3.2) client thread\n");
+		JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (lws<3.2) client thread\n");
 		return NULL;
 	}
 #endif
@@ -712,6 +720,9 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
 			JANUS_LOG(LOG_ERR, "Error connecting to backend (%s)\n",
 				in ? (char *)in : "unknown error");
 			janus_wsevh_calculate_reconnect_delay_on_fail();
+			disconnected = janus_get_monotonic_time();
+			reconnect = TRUE;
+			JANUS_LOG(LOG_WARN, "WebSocketsEventHandler: Error attempting connection... (next retry in %ds)\n", reconnect_delay);
 			janus_wsevh_schedule_connect_attempt();
 			return 1;
 		}
@@ -833,7 +844,7 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
  * Implements the connecting attempt to the upstream websocket server
  * sets the connection result (lws_client_connect_info) to static wsi
  */
-static void janus_wsevh_connect_attempt(void* sul) {
+static void janus_wsevh_connect_attempt(lws_sorted_usec_list_t* sul) {
 	struct lws_client_connect_info i = { 0 };
 	i.host = address;
 	i.origin = address;
@@ -865,16 +876,13 @@ void janus_wsevh_calculate_reconnect_delay_on_fail(void) {
 		if (reconnect_delay > JANUS_WSEVH_MAX_RETRY_SECS)
 			reconnect_delay = JANUS_WSEVH_MAX_RETRY_SECS;
 	}
-	disconnected = janus_get_monotonic_time();
-	reconnect = TRUE;
-	JANUS_LOG(LOG_WARN, "WebSocketsEventHandler: Error attempting connection... (next retry in %ds)\n", reconnect_delay);
 }
 
 /**
  * Schedules a connect attempt using the lws scheduler as 
  */
 void janus_wsevh_schedule_connect_attempt(void) {
-	#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR > 2) || (LWS_LIBRARY_VERSION_MAJOR > 3)
+	#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR >= 4)
 		lws_sul_schedule(context, 0, &sul_stagger, janus_wsevh_connect_attempt, reconnect_delay * LWS_US_PER_SEC);
 	#endif
 }
--- a/events/janus_wsevh.c
+++ b/events/janus_wsevh.c
@@ -46,7 +46,7 @@ json_t *janus_wsevh_handle_request(json_t *request);
 /* Forward declarations */
 void janus_wsevh_schedule_connect_attempt(void);
 void janus_wsevh_calculate_reconnect_delay_on_fail(void);
-static void janus_wsevh_connect_attempt(struct lws_sorted_usec_list* sul);
+static void janus_wsevh_connect_attempt(void* sul);
 
 /* Event handler setup */
 static janus_eventhandler janus_wsevh =
@@ -155,7 +155,9 @@ static const char *protocol = NULL, *address = NULL;
 static char path[256];
 static int port = 0;
 static struct lws_context *context = NULL;
-static lws_sorted_usec_list_t sul_stagger = { 0 };
+#if ((LWS_LIBRARY_VERSION_MAJOR == 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || LWS_LIBRARY_VERSION_MAJOR >= 4)
+	static lws_sorted_usec_list_t sul_stagger = { 0 };
+#endif
 static gint64 disconnected = 0;
 static gboolean reconnect = FALSE;
 static int reconnect_delay = 0;
@@ -384,6 +386,7 @@ int janus_wsevh_init(const char *config_path) {
 	info.protocols = protocols;
 	info.gid = -1;
 	info.uid = -1;
+	info.timeout_secs = 10;
 	if(secure)
 		info.options |= LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT;
 	context = lws_create_context(&info);
@@ -828,7 +831,7 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
  * Implements the connecting attempt to the upstream websocket server
  * sets the connection result (lws_client_connect_info) to static wsi
  */
-static void janus_wsevh_connect_attempt(struct lws_sorted_usec_list* sul) {
+static void janus_wsevh_connect_attempt(void* sul) {
 	struct lws_client_connect_info i = { 0 };
 	i.host = address;
 	i.origin = address;
@@ -845,6 +848,7 @@ static void janus_wsevh_connect_attempt(struct lws_sorted_usec_list* sul) {
 	wsi = lws_client_connect_via_info(&i);
 	if(!wsi) // As we specified a callback pointer in the context the NULL result is unlikely to happen -> just logging it here in case we see it in the field
 		JANUS_LOG(LOG_ERR, "WebSocketsEventHandler: Connecting to upstream websocket server %s:%d failed\n", address, port);
+	reconnect = FALSE;
 }
 
 /**
--- a/events/janus_wsevh.c
+++ b/events/janus_wsevh.c
@@ -574,7 +574,7 @@ plugin_response:
 	* The reconnect is handled in a dedicated wsl scheduler janus_wsevh_schedule_connect_attempt
 	*/
 	static void *janus_wsevh_thread(void *data) {
-	JANUS_LOG(LOG_INFO, "Joining WebSocketsEventHandler (wsl>=3.2) client thread\n");
+	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (wsl>=3.2) client thread\n");
 		int n = 0;
 		while(n >= 0 && g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping))
 				n = lws_service(context, 0);
@@ -588,7 +588,7 @@ plugin_response:
 	* The reconnect is handled in the loop for wsl < 3.2
 	*/
 	static void *janus_wsevh_thread(void *data) {
-		JANUS_LOG(LOG_INFO, "Joining WebSocketsEventHandler (wsl<3.2) client thread\n");
+		JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (wsl<3.2) client thread\n");
 		while(g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping)) {
 			/* Loop until we have to stop */
 			if(!reconnect) {
--- a/events/janus_wsevh.c
+++ b/events/janus_wsevh.c
@@ -582,7 +582,7 @@ static void *janus_wsevh_thread(void *data) {
 	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (lws>=3.2) client thread\n");
 	int n = 0;
 	while(n >= 0 && g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping))
-			n = lws_service(context, 0);
+		n = lws_service(context, 0);
 	lws_context_destroy(context);
 	JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (lws>=3.2) client thread\n");
 	return NULL;
@@ -862,6 +862,7 @@ static void janus_wsevh_connect_attempt(lws_sorted_usec_list_t* sul) {
 		disconnected = janus_get_monotonic_time();
 		reconnect = TRUE;
 		JANUS_LOG(LOG_ERR, "WebSocketsEventHandler: Connecting to upstream websocket server %s:%d failed\n", address, port);
+		return;
 	}
 	reconnect = FALSE;
 }
--- a/events/janus_wsevh.c
+++ b/events/janus_wsevh.c
@@ -624,7 +624,7 @@ static void *janus_wsevh_thread(void *data) {
 	JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (lws<3.2) client thread\n");
 	return NULL;
 }
-#endif 
+#endif
 
 /* Thread to handle incoming events */
 static void *janus_wsevh_handler(void *data) {
--- a/events/janus_wsevh.c
+++ b/events/janus_wsevh.c
@@ -43,6 +43,11 @@ const char *janus_wsevh_get_package(void);
 void janus_wsevh_incoming_event(json_t *event);
 json_t *janus_wsevh_handle_request(json_t *request);
 
+/* Forward declarations */
+void janus_wsevh_schedule_connect_attempt(void);
+void janus_wsevh_calculate_reconnect_delay_on_fail(void);
+static void janus_wsevh_connect_attempt(struct lws_sorted_usec_list* sul);
+
 /* Event handler setup */
 static janus_eventhandler janus_wsevh =
 	JANUS_EVENTHANDLER_INIT (
@@ -150,9 +155,11 @@ static const char *protocol = NULL, *address = NULL;
 static char path[256];
 static int port = 0;
 static struct lws_context *context = NULL;
+static lws_sorted_usec_list_t sul_stagger = { 0 };
 static gint64 disconnected = 0;
 static gboolean reconnect = FALSE;
-static int reconnect_retries = 0;
+static int reconnect_delay = 0;
+#define JANUS_WSEVH_MAX_RETRY_SECS	8
 
 typedef struct janus_wsevh_client {
 	struct lws *wsi;		/* The libwebsockets client instance */
@@ -371,7 +378,6 @@ int janus_wsevh_init(const char *config_path) {
 		protocols[0].name = g_strdup(item->value);
 
 	/* Connect */
-	JANUS_LOG(LOG_VERB, "WebSocketsEventHandler: Connecting to WebSockets server...\n");
 	gboolean secure = !strcasecmp(protocol, "wss");
 	struct lws_context_creation_info info = { 0 };
 	info.port = CONTEXT_PORT_NO_LISTEN;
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @JanFellner! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
As I said in the issue, the schedule is a no go, because it's only available on newer versions of the library. Even doing two different implementations up accommodate both I think will unneededly complicate things, especially considering your changes are a bit refactoring already. I'd rather try and fix the one we have now for the issue you identified.

JanFellner:
The code is aligned with what libwebsocket suggest for >= 3.2 (no own grown code, just the plain sample)

I understand your concerns, tried my best documenting the code, using readable function names.
The difference are two different thread method implementations.

I did not find a solution to get it working the lws_service with 3.2 just sleeps and i see no way in waking it up.

(Testing with 3.1 in the mean time, loosen to what happens with the PR)

JanFellner:
Updated PR to work with lws 3.1 (now building with 3.1 and 4.2)
For lws 3.1 the behaviour is absolutly the same, lws_service wakes up every 25 seconds.
Specified a timeout of 10 seconds but that does not change the behavior

JanFellner:
CLA signed...
i am out now (received two pinball machines from italy need to check what i received :D)

events/janus_wsevh.c
```diff
@@ -372,33 +380,21 @@ int janus_wsevh_init(const char *config_path) {
...
 
 	/* Connect */
-	JANUS_LOG(LOG_VERB, "WebSocketsEventHandler: Connecting to WebSockets server...\n");
 	gboolean secure = !strcasecmp(protocol, "wss");
 	struct lws_context_creation_info info = { 0 };
 	info.port = CONTEXT_PORT_NO_LISTEN;
 	info.protocols = protocols;
 	info.gid = -1;
 	info.uid = -1;
+	info.timeout_secs = 10;
```
atoppi, line 389: have you also checked out `connect_timeout_secs` ?

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
-		} else {
-			/* We should reconnect, get rid of the previous context */
-			if(reconnect_retries > 0) {
-				/* Wait a few seconds before retrying */
-				gint64 now = janus_get_monotonic_time();
-				if((now-disconnected) < (gint64)reconnect_retries*G_USEC_PER_SEC) {
-					/* Try again later */
-					g_usleep(100000);
-					continue;
+#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR > 2) || (LWS_LIBRARY_VERSION_MAJOR > 3)
```
atoppi, line 574: I'd replace `LWS_LIBRARY_VERSION_MAJOR > 3` with `LWS_LIBRARY_VERSION_MAJOR >= 4` to uniform with other similar checks

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+	/*
+	* Websocket thread loop for websocket library newer thant 3.2
+	* The reconnect is handled in a dedicated wsl scheduler janus_wsevh_schedule_connect_attempt
```
atoppi, line 577: I guess _wsl_ is a typo ?

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+	*/
+	static void *janus_wsevh_thread(void *data) {
+	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (wsl>=3.2) client thread\n");
```
atoppi, line 580: fix indentation and _wsl_

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+		int n = 0;
+		while(n >= 0 && g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping))
+				n = lws_service(context, 0);
+		lws_context_destroy(context);
+		JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (wsl>=3.2) client thread\n");
```
atoppi, line 585: _wsl_

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR > 2) || (LWS_LIBRARY_VERSION_MAJOR > 3)
+	/*
+	* Websocket thread loop for websocket library newer thant 3.2
+	* The reconnect is handled in a dedicated wsl scheduler janus_wsevh_schedule_connect_attempt
+	*/
+	static void *janus_wsevh_thread(void *data) {
+	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (wsl>=3.2) client thread\n");
+		int n = 0;
+		while(n >= 0 && g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping))
+				n = lws_service(context, 0);
```
atoppi, line 583: broken indentation

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+		lws_context_destroy(context);
+		JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (wsl>=3.2) client thread\n");
+		return NULL;
+	}
+#else
+	/*
+	* Websocket thread loop for websocket library prior to (less than) 3.2
+	* The reconnect is handled in the loop for wsl < 3.2
```
atoppi, line 591: _wsl_

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+	*/
+	static void *janus_wsevh_thread(void *data) {
+		JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (wsl<3.2) client thread\n");
```
atoppi, line 594: _wsl_

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
-			i.method = NULL;
-			wsi = lws_client_connect_via_info(&i);
-			if(wsi == NULL) {
-				JANUS_LOG(LOG_WARN, "Error attempting reconnection...\n");
-				continue;
-			}
-			reconnect = FALSE;
 		}
+		lws_context_destroy(context);
+		JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (wsl<3.2) client thread\n");
```
atoppi, line 617: _wsl_

events/janus_wsevh.c
```diff
@@ -43,6 +43,11 @@ const char *janus_wsevh_get_package(void);
 void janus_wsevh_incoming_event(json_t *event);
 json_t *janus_wsevh_handle_request(json_t *request);
 
+/* Forward declarations */
+void janus_wsevh_schedule_connect_attempt(void);
+void janus_wsevh_calculate_reconnect_delay_on_fail(void);
+static void janus_wsevh_connect_attempt(void* sul);
```
atoppi, line 49: the _sul_ parameter seems to be unused

events/janus_wsevh.c
```diff
@@ -839,3 +828,53 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
 	}
 	return 0;
 }
+
+/**
+ * Implements the connecting attempt to the upstream websocket server
+ * sets the connection result (lws_client_connect_info) to static wsi
+ */
+static void janus_wsevh_connect_attempt(void* sul) {
```
atoppi, line 836: unused param?

events/janus_wsevh.c
```diff
@@ -839,3 +828,53 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
...
+	disconnected = janus_get_monotonic_time();
+	reconnect = TRUE;
+	JANUS_LOG(LOG_WARN, "WebSocketsEventHandler: Error attempting connection... (next retry in %ds)\n", reconnect_delay);
+}
+
+/**
+ * Schedules a connect attempt using the lws scheduler as 
+ */
+void janus_wsevh_schedule_connect_attempt(void) {
+	#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR > 2) || (LWS_LIBRARY_VERSION_MAJOR > 3)
```
atoppi, line 877: LWS_LIBRARY_VERSION_MAJOR >= 4

events/janus_wsevh.c
```diff
@@ -839,3 +828,53 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
...
+	if(reconnect_delay == 0)
+		reconnect_delay = 1;
+	else if (reconnect_delay < JANUS_WSEVH_MAX_RETRY_SECS) {
+		reconnect_delay += reconnect_delay;
+		if (reconnect_delay > JANUS_WSEVH_MAX_RETRY_SECS)
+			reconnect_delay = JANUS_WSEVH_MAX_RETRY_SECS;
+	}
+	disconnected = janus_get_monotonic_time();
+	reconnect = TRUE;
+	JANUS_LOG(LOG_WARN, "WebSocketsEventHandler: Error attempting connection... (next retry in %ds)\n", reconnect_delay);
```
atoppi, line 870: imho it's not intuitive to change connection status in a method used te calculate the re-connection delay
I'd move `disconnected` assignment, `reconnect` assignment and the log about the error outside of the method (sometimes duplicating the code makes it more readable)

events/janus_wsevh.c
```diff
@@ -372,33 +380,21 @@ int janus_wsevh_init(const char *config_path) {
...
-	i.port = port;
-	i.path = path;
-	i.context = context;
-	if(secure)
-		i.ssl_connection = 1;
-	i.ietf_version_or_minus_one = -1;
-	i.client_exts = exts;
-	i.protocol = protocols[0].name;
-	wsi = lws_client_connect_via_info(&i);
+	janus_wsevh_connect_attempt(NULL);
```
atoppi, line 397: unused param

events/janus_wsevh.c
```diff
@@ -575,60 +571,53 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+					/* Wait a few seconds before retrying */
+					gint64 now = janus_get_monotonic_time();
+					if((now-disconnected) < (gint64)reconnect_delay*G_USEC_PER_SEC) {
+						/* Try again later */
+						g_usleep(100000);
+						continue;
+					}
 				}
+				ws_client = NULL;
+				janus_wsevh_connect_attempt(NULL);
```
atoppi, line 611: unused param

atoppi (commented):
Thanks @JanFellner , I think this is a good enhancement.
You missed the handling of re-connection in case of `LWS_CALLBACK_CLOSED`.

I'm leaving some other notes that should be addressed.

JanFellner:
@atoppi are you using any kind for code formatting with rules? (using the same env as you vscode)
Could not find any clang rules in the repo but i guess some sort of automation is needed that you don´t need to comment PR by coding style (which i don´t care i just use any ruleset available).
If not i will just follow your comments 😉

events/janus_wsevh.c
JanFellner: is only available in >= 3.2 so i just left this one here as it is available in all versions of lws.
But as mentioned before here or in the other ticket i played with so many timing options in such a wide manner but did not get the lws_service to wake up... It´s the same behaviour actually also in the earlier versions so i don´t know where the reconnect in the service loop comes from (i did not blame it :)). But i guess that the eventing websocket is not widley used as the doubling of the time on connect fails each time is such a blocker to use it productive. I would not fix it for wsl earlier than 3.2, i just ensure that it compiles but the behaviour for the old one is the same. Wakeup every 25 seconds.

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: yeah :) and copy pasted, fixing it everywhere

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: i need to have that pointer there for the lws_sul_schedule call. The compiler is otherwise complaining about it.
As i don´t need it and the type is unknown in 3.2 i just use void* (I will add the appropriate comment to the forward decl and the actual impl to point out why)

events/janus_wsevh.c
JanFellner: See https://github.com/meetecho/janus-gateway/pull/2736#discussion_r671037322

events/janus_wsevh.c
JanFellner: See https://github.com/meetecho/janus-gateway/pull/2736#discussion_r671037322

events/janus_wsevh.c
JanFellner: Change it once more to have it better readable.
I use the required lws_sorted_usec_list_t (*) and if not available define it to void (*)

events/janus_wsevh.c
JanFellner: i duplicated it and placed it accordingly

events/janus_wsevh.c
atoppi: I guess that eventing websocket is not widely used _with an unreachable host_.

atoppi:
@JanFellner No I'm not applying any kind of template in the IDE, just following the style in the rest of code.

events/janus_wsevh.c
```diff
@@ -150,9 +159,12 @@ static const char *protocol = NULL, *address = NULL;
 static char path[256];
 static int port = 0;
 static struct lws_context *context = NULL;
+#if ((LWS_LIBRARY_VERSION_MAJOR == 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || LWS_LIBRARY_VERSION_MAJOR >= 4)
+	static lws_sorted_usec_list_t sul_stagger = { 0 };
```
atoppi, line 163: fix indentation ;-)

events/janus_wsevh.c
```diff
@@ -43,6 +43,15 @@ const char *janus_wsevh_get_package(void);
 void janus_wsevh_incoming_event(json_t *event);
 json_t *janus_wsevh_handle_request(json_t *request);
 
+/* Forward declarations */
+void janus_wsevh_schedule_connect_attempt(void);
+void janus_wsevh_calculate_reconnect_delay_on_fail(void);
```
atoppi, line 48: This group of declarations should be made _static_.
Also IMHO it should be moved under this block
```c
/* Useful stuff */
static volatile gint initialized = 0, stopping = 0;
static GThread *ws_thread, *handler_thread;
static void *janus_wsevh_thread(void *data);
static void *janus_wsevh_handler(void *data);
```
and the comment should be more informative like e.g.
```c
/* Connection helpers */
```

events/janus_wsevh.c
```diff
@@ -43,6 +43,15 @@ const char *janus_wsevh_get_package(void);
...
+// lws_sorted_usec_list_t is defined starting with lws 3.2
```
atoppi, line 49: please use `/*` syntax for comments

events/janus_wsevh.c
```diff
@@ -575,60 +575,57 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
-			if(reconnect_retries > 0) {
-				/* Wait a few seconds before retrying */
-				gint64 now = janus_get_monotonic_time();
-				if((now-disconnected) < (gint64)reconnect_retries*G_USEC_PER_SEC) {
-					/* Try again later */
-					g_usleep(100000);
-					continue;
+#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR >= 4)
+	/*
+	* Websocket thread loop for websocket library newer thant 3.2
```
atoppi, line 580: typo _thant_

events/janus_wsevh.c
```diff
@@ -575,60 +575,57 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
-		} else {
-			/* We should reconnect, get rid of the previous context */
-			if(reconnect_retries > 0) {
-				/* Wait a few seconds before retrying */
-				gint64 now = janus_get_monotonic_time();
-				if((now-disconnected) < (gint64)reconnect_retries*G_USEC_PER_SEC) {
-					/* Try again later */
-					g_usleep(100000);
-					continue;
+#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR >= 4)
```
atoppi, line 578: code following preprocessor declarations should not be intended, please fix.

events/janus_wsevh.c
```diff
@@ -575,60 +575,57 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+						g_usleep(100000);
+						continue;
+					}
+				}
+				ws_client = NULL;
+				janus_wsevh_connect_attempt(NULL);
+				if (!wsi) {
+					janus_wsevh_calculate_reconnect_delay_on_fail();
+					disconnected = janus_get_monotonic_time();
+					reconnect = TRUE;
```
atoppi, line 619: I guess it is ok to set `disconnected` and `reconnect` inside `janus_wsevh_connect_attempt`, just keeping the delay calculation and the log here.

atoppi (commented):
The PR is still **missing the reconnection schedule in case** of `LWS_CALLBACK_CLOSED`.
I've left other notes to address.

JanFellner:
While working on the code changes i lost the closed in mind, will address it

events/janus_wsevh.c
JanFellner: k got it now :]
i always intended the code when i was mainly developing c++ but thats definetly user perspective... will address the others as well..

events/janus_wsevh.c
JanFellner: like /* Useful stuff */ 😂 adopting and moving code

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: inside the connect attempt i set the reconnect to false as i am currently inbetween the connect and the result.
The main interaction is the callback that gives me the final async result of the operation.
It looks like there were old wsl implementations that used to give a synchronous connection feedback of the connect attempt but i never saw that. the result object is always filled

so i would actually prefer to get rid of that pointer check as to me the handling should be connect initiates something, the event tells me what happend and how to proceed.

(maybe thats grown code, it does not make sense to me... i just adopted it the way it was, the null result somehow was at some point associated with failed but i never saw it happening that way)

JanFellner:
Hope i did now cover everything :)

events/janus_wsevh.c
JanFellner: I adopted it but i guess the code will anyway never run and should be removed as everything comes through the callback anyway.

events/janus_wsevh.c
```diff
@@ -575,60 +573,55 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+/*
+* Websocket thread loop for websocket library newer than 3.2
+* The reconnect is handled in a dedicated lws scheduler janus_wsevh_schedule_connect_attempt
+*/
 static void *janus_wsevh_thread(void *data) {
-	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler client thread\n");
+	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (lws>=3.2) client thread\n");
+	int n = 0;
+	while(n >= 0 && g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping))
+			n = lws_service(context, 0);
```
atoppi, line 585: fix indentation

events/janus_wsevh.c
```diff
@@ -839,3 +837,54 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
...
+	i.ietf_version_or_minus_one = -1;
+	i.client_exts = exts;
+	i.protocol = protocols[0].name;
+	JANUS_LOG(LOG_INFO, "WebSocketsEventHandler: Connecting to upstream websocket server %s:%d...\n", address, port);
+	wsi = lws_client_connect_via_info(&i);
+	if(!wsi) {
+		/* As we specified a callback pointer in the context the NULL result is unlikely to happen */
+		disconnected = janus_get_monotonic_time();
+		reconnect = TRUE;
+		JANUS_LOG(LOG_ERR, "WebSocketsEventHandler: Connecting to upstream websocket server %s:%d failed\n", address, port);
```
atoppi, line 864: missing return

JanFellner:
As much of the comments were styling related, i wonder if you ever discussed using code formatters.
I did not until 2 years ago and i regret not having used them before.
It just releases you from the burden of having to communicate how code should look like. The rule just formats it.
(We started with just specifying what is common in the team and not with rules the have conflict potential like line length limits)

events/janus_wsevh.c
JanFellner: fixed

events/janus_wsevh.c
JanFellner: fixed

atoppi:
@JanFellner nope, never discussed about this.
Are you thinking about a simple text document describing the common rules to follow or a specific file that auto-format your code in the IDE (like VSC clang-format) ?
If the latter, consider that @lminiero afaik is using _geany_ to write code, so I'm not sure how he could use or validate it.

JanFellner:
I am a friend about automation.
If you specify it in a standard you can also apply automated tools that validated stuff e.g. in a PR. so no unformatted code comes in the repo. So a clang file would for me be a starting point but it´s something you need to discuss.
The main task in the beginning is to setup the tooling, in vscode it´s just the plugin that you need to enable. That finds the file and does the magic for you. Other editors other ways...

atoppi:
It seems to work as intended on lws v4.2 :-)

However when specifying an invalid address, the timeout on my machine is 15 seconds.
I guess, if not specified, it is falling back to a default environment specific value.
I confirm that setting the timeout (e.g. to 5) the connection goes in failed state quicker.

JanFellner:
Added a connected log entry and removed a duplicate disconnected log entry

JanFellner:
> However when specifying an invalid address, the timeout on my machine is 15 seconds.
> I guess, if not specified, it is falling back to a default environment specific value.
> I confirm that setting the timeout (e.g. to 5) the connection goes in failed state quicker.

I added that value with d914fa4  for lws>=4.2

I used it now the whole day and it looks good to me.

atoppi:
@JanFellner lgtm :+1: 
waiting for @lminiero feedback before merging.

events/janus_wsevh.c
```diff
@@ -575,60 +576,55 @@ json_t *janus_wsevh_handle_request(json_t *request) {
 		}
 }
 
-/* Thread to implement the WebSockets loop */
+#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR >= 4)
+/*
+* Websocket thread loop for websocket library newer than 3.2
+* The reconnect is handled in a dedicated lws scheduler janus_wsevh_schedule_connect_attempt
+*/
```
lminiero, line 583: Code style: comments should be in the style below

```
/* Bla bla
 * bla bla */
```

events/janus_wsevh.c
```diff
@@ -575,60 +576,55 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
 static void *janus_wsevh_thread(void *data) {
-	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler client thread\n");
+	JANUS_LOG(LOG_VERB, "Joining WebSocketsEventHandler (lws>=3.2) client thread\n");
+	int n = 0;
+	while(n >= 0 && g_atomic_int_get(&initialized) && !g_atomic_int_get(&stopping))
+		n = lws_service(context, 0);
```
lminiero, line 588: What can cause `lws_service` to return a negative integer? I'm thinking of intermittent errors that could cause this loop to break before Janus is shutdown.

events/janus_wsevh.c
```diff
@@ -575,60 +576,55 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
+	lws_context_destroy(context);
+	JANUS_LOG(LOG_VERB, "Leaving WebSocketsEventHandler (lws>=3.2) client thread\n");
+	return NULL;
+}
+#else
+/*
+* Websocket thread loop for websocket library prior to (less than) 3.2
+* The reconnect is handled in the loop for lws < 3.2
+*/
```
lminiero, line 597: See above on comments style.

events/janus_wsevh.c
```diff
@@ -575,60 +576,55 @@ json_t *janus_wsevh_handle_request(json_t *request) {
...
-			i.ietf_version_or_minus_one = -1;
-			i.client_exts = exts;
-			i.protocol = protocols[0].name;
-			i.method = NULL;
-			wsi = lws_client_connect_via_info(&i);
-			if(wsi == NULL) {
-				JANUS_LOG(LOG_WARN, "Error attempting reconnection...\n");
-				continue;
+			janus_wsevh_connect_attempt(NULL);
+			if (!wsi) {
```
lminiero, line 617: Code style: no space between 'if` and the bracket.

events/janus_wsevh.c
```diff
@@ -713,17 +709,21 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
...
 			return 0;
 		}
 		case LWS_CALLBACK_CLIENT_CONNECTION_ERROR: {
 			JANUS_LOG(LOG_ERR, "Error connecting to backend (%s)\n",
 				in ? (char *)in : "unknown error");
-			/* Should we reconnect? */
+			janus_wsevh_calculate_reconnect_delay_on_fail();
 			disconnected = janus_get_monotonic_time();
 			reconnect = TRUE;
+			JANUS_LOG(LOG_WARN, "WebSocketsEventHandler: Error attempting connection... (next retry in %ds)\n", reconnect_delay);
```
lminiero, line 725: We already have a log error above: please add the retry info there.

events/janus_wsevh.c
```diff
@@ -839,3 +840,55 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
 	}
 	return 0;
 }
+
+/**
+ * Implements the connecting attempt to the upstream websocket server
+ * sets the connection result (lws_client_connect_info) to static wsi
+ */
```
lminiero, line 847: See above on comments style.

events/janus_wsevh.c
```diff
@@ -839,3 +840,55 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
...
+		JANUS_LOG(LOG_ERR, "WebSocketsEventHandler: Connecting to upstream websocket server %s:%d failed\n", address, port);
+		return;
+	}
+	reconnect = FALSE;
+}
+
+/**
+ * Adopts the reconnect_delay value in case of an error
+ * Increases the value up to JANUS_WSEVH_MAX_RETRY_SECS
+ */
```
lminiero, line 876: See above on comments style.

events/janus_wsevh.c
```diff
@@ -839,3 +840,55 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
...
+static void janus_wsevh_calculate_reconnect_delay_on_fail(void) {
+	if(reconnect_delay == 0)
+		reconnect_delay = 1;
+	else if (reconnect_delay < JANUS_WSEVH_MAX_RETRY_SECS) {
+		reconnect_delay += reconnect_delay;
+		if (reconnect_delay > JANUS_WSEVH_MAX_RETRY_SECS)
```
lminiero, line 882: The two `if` above have an extra space that per code style shouldn't be there.

events/janus_wsevh.c
```diff
@@ -839,3 +840,55 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
...
+			reconnect_delay = JANUS_WSEVH_MAX_RETRY_SECS;
+	}
+}
+
+/**
+ * Schedules a connect attempt using the lws scheduler as 
+ */
```
lminiero, line 889: See above on comments style.

events/janus_wsevh.c
```diff
@@ -839,3 +840,55 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
...
+static void janus_wsevh_schedule_connect_attempt(void) {
+	#if (LWS_LIBRARY_VERSION_MAJOR >= 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || (LWS_LIBRARY_VERSION_MAJOR >= 4)
+		lws_sul_schedule(context, 0, &sul_stagger, janus_wsevh_connect_attempt, reconnect_delay * LWS_US_PER_SEC);
+	#endif
+}
```
lminiero, line 894: There's a missing empty line at the end.

events/janus_wsevh.c
```diff
@@ -69,13 +69,21 @@ janus_eventhandler *create(void) {
...
 static void *janus_wsevh_handler(void *data);
 
+/* Connection related helper methods */
+static void janus_wsevh_schedule_connect_attempt(void);
+static void janus_wsevh_calculate_reconnect_delay_on_fail(void);
+/* lws_sorted_usec_list_t is defined starting with lws 3.2 */
+#if !(((LWS_LIBRARY_VERSION_MAJOR == 3 && LWS_LIBRARY_VERSION_MINOR >= 2) || LWS_LIBRARY_VERSION_MAJOR >= 4))
+	#define lws_sorted_usec_list_t void
+#endif
+static void janus_wsevh_connect_attempt(lws_sorted_usec_list_t* sul);
```
lminiero, line 85: Per code style, the `*` should be next to the pointer, not the type:

```
lws_sorted_usec_list_t *sul
```

events/janus_wsevh.c
```diff
@@ -839,3 +840,55 @@ static int janus_wsevh_callback(struct lws *wsi, enum lws_callback_reasons reaso
 	}
 	return 0;
 }
+
+/**
+ * Implements the connecting attempt to the upstream websocket server
+ * sets the connection result (lws_client_connect_info) to static wsi
+ */
+static void janus_wsevh_connect_attempt(lws_sorted_usec_list_t* sul) {
```
lminiero, line 848: Same here (`*` next to the pointer, not type).

lminiero (commented):
Thanks! Added some editorial notes and a question inline.

JanFellner:
Thanks @lminiero for the comments, will check those now...

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: added

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: ack

events/janus_wsevh.c
JanFellner: It´s copy pasted from the sample actually.

https://libwebsockets.org/git/libwebsockets/tree/minimal-examples/ws-client/minimal-ws-client/minimal-ws-client.c#n210

I will check inside libwebsocket if it returns a different value and what might be conditions

JanFellner: If you follow the functions through the lib there are some places where the function returns -1,
I did not validate them all but to me it looks like those are scenarios where the context is not properly settled and therefore further execution does not make sense...

* lws_service() returns 1 or lws_plat_service()
* lws_plat_service() returns _lws_plat_service_tsi()
* _lws_plat_service_tsi() returns on seldom cases -1 but the main negative returns are here lws_service_fd_tsi()

https://libwebsockets.org/git/libwebsockets/tree/lib/core-net/service.c?h=v4.2-stable#n624

The question is whats better in the end.
A 100% thread or a thread termination :)

JanFellner:
Pushed

Concerning the styling stuff please do also check https://github.com/meetecho/janus-gateway/pull/2736#issuecomment-881380705 and the following two posts as suggestion...
(Even if you don´t use it it would help others to use the style you are using without having to comment them on PRs :) )

events/janus_wsevh.c
lminiero: The problem is that this could result in a fatal scenario due to a temporary issue: as a first integration of the return value, I'd just print the negative value in case of errors, but I'd move on with the loop (so no `n>=0` in the `while`). This would give us some more information from people's logs, in case it ever happens.

JanFellner: I can add it but i fear i causes more trouble than it solves.
What if the lws_service just returns instantly without sleeping? 
That would just generate endless log entries with one core just writing log entries...

JanFellner:
Logging lws_service result if it gets negative and just in case if it reverts back from negativ (or changes) as well.
Both classified as errors just to have them both in the log and not just one of them

lminiero:
Looks good to me too, now! Thanks for the quick feedback and fixes, merging :v:

lminiero:
PS: just noticed the automated builds weren't started, I'll do those before merging just to be sure everything's fine.

PR #2738, opened 2021-07-20
Use crypto safe random numbers

jmfotokite:
This is a continuation of the discussion in https://groups.google.com/g/meetecho-janus/c/dIVe_v1xaxM.

It is not yet a ready-to-merge PR, more of a suggestion.

Some considerations:
* creates a new janus_random_uint64_javacript_safe() method to make it explicit that those numbers are safe to use in JS
* but do use janus_random_uint64() in the other places (like when generating an UUID)
* make "crypto safe random" optional? This might sound like a good idea at first, because it can theoretically fail (unlike the pseudo random g_random_int()). BUT:
    * it is easier to have only one behavior in the code
    * pretty much any system in use nowadays depends on having working crypto safe random numbers, a lot stuff would not work otherwise (SSL etc)
    * the same is true for Janus, e.g. the DTLS implementation
* in terms of CPU usage, I doubt that this is an issue. We are not using random stuff at a great rate, and a quick test shows that, on my Thinkpad, I can generate "good" random data quickly enough: `cat /dev/urandom | pv > /dev/null` --> `0:00:05 [ 197MiB/s] [                    <=> ]`


lminiero:
Thanks! A few comments:

* I don't think we need a `janus_random_uint64_javacript_safe`, as it's a breaking signature name change for something that won't be really used: you just updated the core, but we use random uint64 values all over the plugins; besides, instead than modifying the call name EVERYWHERE, I'd rather the method defaults to capping at 2^53-1, and maybe have another method (with a different name) with no cap instead, in case it's needed by anything (nothing that comes to mind)
* the automated build failed a couple of times due to a missing `RAND_bytes()` availability: this means that there are build configurations we support where this may happen

lminiero:
> the automated build failed a couple of times due to a missing `RAND_bytes()` availability: this means that there are build configurations we support where this may happen

Look like the problem is `mjr2pcap` (and possibly other postprocessing tools that are compiled later), that uses these random generators too, is not linked to ssl/crypto, and so fails.

atoppi:
Thanks for the effort @jmfotokite 
1) What if we make crpyto-safe RNG an option when configuring janus, not changing the signatures all over like @lminiero mentioned
2) I agree on the low impact in terms of CPU. This code will be executed only on the signaling path.
3) We need to double-check if the fuzzers are still working after this change

jmfotokite:
> I don't think we need a janus_random_uint64_javacript_safe

OK. We could also name it janus_random_uint52() or something. I just though it was very wrong to have a janus_random_uint64() method which returns only 52 bite. In the end, it is your call of course. I can understand if you do not want to change the signature name because 3rd party plugins depend on it or similar. What should we call the "real" rand64 method then?

> What if we make crpyto-safe RNG an option when configuring janus, not changing the signatures all over like lminiero mentioned

Those are two orthogonal things. The crypto-safe RNG has nothing to do with the signature change.

lminiero:
> What should we call the "real" rand64 method then?

No need to be overly creative, I guess something like `janus_random_uint64_full()` works :hand_over_mouth: 

> Those are two orthogonal things. The crypto-safe RNG has nothing to do with the signature change.

I agree, it would just change their implementation: we do the same capping to 2^53-1 in the current version as well. In case this configure-level thing is added, though, it would mean not add a dependency to the crypto stuff where it's not needed in some cases though (e.g., mjr2pcap when not using the crypto-safe version).

Makefile.am
```diff
@@ -637,10 +637,12 @@ mjr2pcap_SOURCES = \
 mjr2pcap_CFLAGS = \
 	$(AM_CFLAGS) \
 	$(POST_PROCESSING_CFLAGS) \
+	$(BORINGSSL_CFLAGS) \
```
lminiero, line 640: Why BORINGSSL? That's not wat we use in configure as a variable for the crypto stuff.

jmfotokite: uh, that was just my naive attempt to fix the build after grepping for "ssl" and mjr2pcap... where do I have to fix this for real?

atoppi:
My idea was to use a macro defined in configure.
In the (default) case the macro is not defined everything is exactly like it is now.
Conversely, should the macro be defined, the random_uint methods will use the crypto stuff.
In any case the signatures of the involved methods do not change and, as we are already discussing, we could just introduce a new method that will give a full uint64 random in case something would need it.

Makefile.am
lminiero: It's probably easier to add the ssl and/or crypto checks to the `POST_PROCESSING` definition in `configure.ac`, which will make it available to all post-processing tools. The `BORINGSSL_CFLAGS` and `BORINGSSL_LIBS` variables are still needed in case BoringSSL is used: you can refer to how `janus_CFLAGS` and `janus_LDADD` use them.

jmfotokite: Sorry, it seems I cannot get it to work. I added `libssl >= $ssl_version` to `POST_PROCESSING`, but the build still fails. I don't know what else I have to do. I have used CMake, but not autotools before :grimacing:

jmfotokite:
The signatures are reverted now.

> My idea was to use a macro defined in configure.
> ...
> Conversely, should the macro be defined, the random_uint methods will use the crypto stuff.

So, should I add a `--enable-crypto-safe-random` option to the build config (and adapt the docstrings)?

atoppi:
@jmfotokite I was just trying to collect feedback.
If @lminiero agrees to moving to crypto-safe implementation, then I guess we do not need the configuration.

atoppi:
@jmfotokite the linking with `libcrypto` is missing.
Please apply the attached patch. It should fix compilation for post-processor and fuzzers.

[fix_crypto_rng_pr.txt](https://github.com/meetecho/janus-gateway/files/6863397/fix_crypto_rng_pr.txt)

jmfotokite:
Thanks! Let's hope it works now :pray:

atoppi:
For sake of completeness, here is a small benchmark that compares the two approaches
```c
#include <stdio.h>
#include <inttypes.h>
#include <time.h>
#include <glib.h>
#include <openssl/rand.h>

static guint64 janus_random_uint64(void);
static guint64 janus_random_uint64_crypto(void);

guint64 janus_random_uint64(void) {
	guint64 num = g_random_int() & 0x1FFFFF;
	num = (num << 32) | g_random_int();
	return num;
}

guint64 janus_random_uint64_crypto(void) {
	guint64 ret = 0;
	if (RAND_bytes((void *)&ret, sizeof(ret)) != 1) {
		printf("\tOpenSSL RAND_bytes() failed\n");
		exit(1);
	}
	ret = ret & 0x1FFFFFFFFFFFFF;
	return ret;
}

int main(void) 
{	
	const int ITERS=1000000;
	
	clock_t tic = clock();
	for(int i=0; i<ITERS; i++) {
		janus_random_uint64();
		//printf("random1 = %" PRIu64 "\n", val);
	}
	clock_t toc = clock();
	printf("[\tg_random_int\t] : %d iterations took %f seconds\n", ITERS, (double)(toc - tic) / CLOCKS_PER_SEC);
	
	tic = clock();
	for(int j=0; j<ITERS; j++) {
		janus_random_uint64_crypto();
		//printf("random2 = %" PRIu64 "\n", val);
	}
	toc = clock();
	printf("[\tRAND_bytes\t] : %d iterations took %f seconds\n", ITERS, (double)(toc - tic) / CLOCKS_PER_SEC);
} 

```

Iterating the RNG functions 100,000 times:
```
[	g_random_int	] : 1000000 iterations took 0.042884 seconds
[	RAND_bytes	] : 1000000 iterations took 1.255055 seconds
```
So yeah, the crypto-safe functions have more impact on the CPU, but the time needed to sample a value is still very low (0,0125 ms on average).

jmfotokite:
Wow, I did not expect a 30x slowdown... but yes, we are actually not using janus_random_uint64() a lot, so that's fine.

atoppi:
I've made a some tests with echotest, videoroom and streaming.
Everything seems to work as expected.

configure.ac
```diff
@@ -954,6 +954,8 @@ AS_IF([test "x$enable_post_processing" = "xyes"],
                          [
                            glib-2.0 >= $glib_version
                            jansson >= $jansson_version
+                           libssl >= $ssl_version
+                           libcrypto
```
lminiero, line 958: If you add these here, you'll need to add `$(BORINGSSL_LIBS)` and `$(BORINGSSL_CFLAGS)` as we do for `janus_LDADD` and `janus_CFLAGS` to all the postprocessor tools as well. That's needed because if people are configuring Janus to use BoringSSL, the lack of those variables will cause the wrong crypto library to be used (or one not to be found at all).

Just see how those variables are currently used in `Makefile.am`: there's no problem adding them even when BoringSSL is not used, as they default to empty strings and are only set if BoringSSL is enabled when configuring Janus.

janus.c
```diff
@@ -4996,6 +4996,11 @@ gint main(int argc, char *argv[])
 	SSL_library_init();
 	SSL_load_error_strings();
 	OpenSSL_add_all_algorithms();
+	/* check if random pool looks ok (this does not give any guarantees for later, though) */
```
lminiero, line 4999: Just a nit on code syle: comments should start with a capitized letter.

janus.c
```diff
@@ -4996,6 +4996,11 @@ gint main(int argc, char *argv[])
...
+	if(RAND_status() != 1) {
+		JANUS_LOG(LOG_FATAL, "\tOpenSSL PRNG is not properly seeded, cannot generate random numbers\n");
```
lminiero, line 5001: I don't think the `\t` is needed here. I'd also remove the explicit reference to OpenSSL, since we support LibreSSL and BoringSSL as well.

plugins/janus_nosip.c
```diff
@@ -786,11 +786,6 @@ int janus_nosip_init(janus_callbacks *callback, const char *config_path) {
 	}
 	JANUS_LOG(LOG_VERB, "Local IP set to %s\n", local_ip);
 
-#ifdef HAVE_SRTP_2
-	/* Init randomizer (for randum numbers in SRTP) */
-	RAND_poll();
-#endif
-
```
lminiero, line 793: We should check if removing the randomizers init here (and the SIP plugin) is ok, since plugins are loaded with `dlopen` and we only do the init in the core.

utils.c
```diff
@@ -71,23 +72,25 @@ gboolean janus_strcmp_const_time(const void *str1, const void *str2) {
 }
 
 guint32 janus_random_uint32(void) {
-	return g_random_int();
+	guint32 ret = 0;
+	if (RAND_bytes((void *)&ret, sizeof(ret)) != 1) {
```
lminiero, line 76: Code style: there shouldn't be a space between `if` and the bracket.

utils.c
```diff
@@ -71,23 +72,25 @@ gboolean janus_strcmp_const_time(const void *str1, const void *str2) {
...
+		JANUS_LOG(LOG_FATAL, "\tOpenSSL RAND_bytes() failed\n");
+		exit(1);
```
lminiero, line 78: I don't like the `exit(1)` here at all... This will cause the program to just exit while in use, which should never happen. I'd much rather fallback to the unsafe glib random-int implementation when `RAND_bytes` fails (maybe with a warning printed out rather than a fatal error).

utils.c
```diff
@@ -71,23 +72,25 @@ gboolean janus_strcmp_const_time(const void *str1, const void *str2) {
...
+	}
+	return ret;
+}
+
+guint64 janus_random_uint64_full(void) {
+	guint64 ret = 0;
+	if (RAND_bytes((void *)&ret, sizeof(ret)) != 1) {
+		JANUS_LOG(LOG_FATAL, "\tOpenSSL RAND_bytes() failed\n");
+		exit(1);
```
lminiero, line 87: See above on code style and exit.

utils.h
```diff
@@ -62,15 +62,29 @@ gboolean janus_strcmp_const_time(const void *str1, const void *str2);
...
+/*! \brief Helper to generate random 64-bit unsigned integers
+ * @note Unlike janus_random_uint64(), which actually only generates 52 bits, this
+ * generates the full 64 bits. See the janus_random_uint64() docstring for details.
  * @returns A random 64-bit unsigned integer */
+guint64 janus_random_uint64_full(void);
+
+/*! \brief Helper to generate random 52 bit unsigned integers
+ * @note The reason for 52 instead of 64 bits: Javascript does not have real integers,
+ * its builtin "number" type is a float64. Thus, only integer values up to
+ * Number.MAX_SAFE_INTEGER == 2^53 - 1 == 9007199254740991
```
lminiero, line 77: You may want to wrap this line in `<code>Number.MAX...</code>` so that Doxygen will format it accordingly.

lminiero (commented):
Just added some notes inline, thanks again!

janus.c
jmfotokite: done

janus.c
jmfotokite: done

plugins/janus_nosip.c
jmfotokite: Calling RAND_poll() it not necessary at all: https://www.openssl.org/docs/manmaster/man3/RAND_poll.html

>  These functions can be used to seed the random generator and to check its seeded state. In general, manual (re-)seeding of the default OpenSSL random generator (RAND_OpenSSL(3)) is not necessary (but allowed), since it does (re-)seed itself automatically using trusted system entropy sources. This holds unless the default RAND_METHOD has been replaced or OpenSSL was built with automatic reseeding disabled, see RAND(7) for more details.

utils.c
jmfotokite: done

utils.c
jmfotokite: done

utils.h
jmfotokite: done

utils.c
jmfotokite: ok

configure.ac
jmfotokite: done, please check again

utils.c
```diff
@@ -71,23 +72,25 @@ gboolean janus_strcmp_const_time(const void *str1, const void *str2) {
...
+		JANUS_LOG(LOG_ERR, "Safe RAND_bytes() failed, falling back to unsafe PRNG\n");
+		return g_random_int();
+	}
+	return ret;
+}
+
+guint64 janus_random_uint64_full(void) {
+	guint64 ret = 0;
+	if(RAND_bytes((void *)&ret, sizeof(ret)) != 1) {
+		JANUS_LOG(LOG_ERR, "Safe RAND_bytes() failed, falling back to unsafe PRNG\n");
```
lminiero, line 86: The two `LOG_ERR` above should be a `LOG_WARN`.

lminiero (commented):
Looks good, thanks! I just added a note on logging, and then it's good to go for me.

jmfotokite:
(obviously this dirty commit history should be squash-merged)

lminiero:
Thanks, this looks good to me now! Merging then, thanks for the huge work :+1:

PR #2740, opened 2021-07-22
CC-2216 show libwebsocket version in admin dependencies

JanFellner:
As the libwebsocket version is currently missing in the admin dependencies i thought i just quickly add it here
Want to mention that the libwebsocket does not provide a machine parseable version "4.2.0" but only a "v4.2.0"
If we want it machine parseable a dedicated string with the major minor patch would be required.
I am just using the version string from the lib

{
"glib2":"2.64.6",
"jansson":"2.12",
"libnice":"0.1.18",
"libsrtp":"libsrtp2 2.3.0",
"libcurl":"7.68.0",
"libwebsocket":"v4.2.0",
"crypto":"OpenSSL 1.1.1 (compatible; BoringSSL)"
}

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -28,6 +28,7 @@
 #ifdef HAVE_TURNRESTAPI
 #include <curl/curl.h>
 #endif
+#include <libwebsockets.h>
 
 #include "janus.h"
 #include "version.h"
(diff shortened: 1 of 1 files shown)


lminiero:
Sorry, but this cannot be accepted. We only list dependencies in the core, there, and you just added a hardcoded include to a library that may not even be there (all plugins, including the WebSocket transport and event handler, are optional).

lminiero:
We have a ton of different transports, and you don't just talk to Janus from a browser (which could use HTTP instead of WS, incidentally). There's people using RabbitMQ to control Janus, for instance.

JanFellner:
k so a possible approach would then be to list dependencies plugin related?

PR #2743, opened 2021-07-24
Fix sample event handler behaviour

i8-pi:
This pull request makes some breaking changes but the new behaviour is arguably more correct

### Remove a trailing null in gzipped messages using janus_gzip_compress

The trailing null in a string is not logically part of the string contents. It should not be included in gzip message body.

The only two places in the code that use this function are the sample and gelf event handlers

### Use standard HTTP headers in messages sent by the sample event handler

- The `Accept` header is not relevant since we ignore the HTTP response
- The standard method to signal a gzipped message body is not to modify `Content-Type` but to use `Content-Encoding: gzip`
- charsets is not a standard HTTP header. utf8 is implied by JSON


lminiero:
Looks good to me, and thanks for the clarifications! Merging :+1:

PR #2744, opened 2021-07-27
Update README.md

danjenkins:
Stun server cli param wasn't a filename but a url (not a uri because it doesnt include the `stun:` protocol prefix

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ or on the command line:
 	-c, --cert-pem=filename       DTLS certificate
 	-k, --cert-key=filename       DTLS certificate key
 	-K, --cert-pwd=text           DTLS certificate key passphrase (if needed)
-	-S, --stun-server=filename    STUN server(:port) to use, if needed (e.g.,
+	-S, --stun-server=url    STUN server(:port) to use, if needed (e.g.,
 								  Janus behind NAT, default=none)
 	-1, --nat-1-1=ip              Public IP to put in all host candidates,
                                   assuming a 1:1 NAT is in place (e.g., Amazon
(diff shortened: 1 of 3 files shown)


januscla:
Thanks for your contribution, @danjenkins! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Good catch! But this should be updated in `janus.ggo` and `janus.1`, as the former is what created the command line arguments (of which the README is just a summary) and the latter provides the man documentation.

lminiero:
Then the README is indeed outdated, and it should report what `janus --help` prints:

```
  -S, --stun-server=ip:port     STUN server(:port) to use, if needed (e.g.,
                                  Janus behind NAT, default=none)
```

lminiero:
On ip, it makes sense: maybe `address:port`? In that case, all three files should be updated accordingly.

README.md
```diff
@@ -264,8 +264,9 @@ or on the command line:
 	-c, --cert-pem=filename       DTLS certificate
 	-k, --cert-key=filename       DTLS certificate key
 	-K, --cert-pwd=text           DTLS certificate key passphrase (if needed)
-	-S, --stun-server=url    STUN server(:port) to use, if needed (e.g.,
-								  Janus behind NAT, default=none)
+	-S, --stun-server=address:port
+                                  STUN server(:port) to use, if needed (e.g.,
```
danjenkins, line 268: I've followed how the formatting was done for `-m, --reclaim-session-timeout=number` as that was another parameter that forced the start of the description onto the next line

lminiero: It should be enough to recompile and launch Janus with the `--help` property, and copy it in the README, and then just indent  the first line with a tab instead of spaces (for the other lines you can keep what is generated).

lminiero: I mean just copying the updated property, of course, not all of them.

danjenkins: It seems it doesnt want to compile on mac for me... so I'll have to sort this out tomorrow.

lminiero: No need, I can fix it myself after merging :+1:

PR #2746, opened 2021-07-30
CMake implementation

SamuelMarks:
Just started CMake for you.

Do you want it?

If so, next I'll add 3rd-party library support with [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html) and [`target_link_libraries`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,114 @@
+cmake_minimum_required(VERSION 3.19)
+
+# set the project name and version
+project(janus_gateway VERSION 0.11.4 LANGUAGES C)
+string(TOLOWER "${PROJECT_NAME}" PROJECT_LOWER_NAME)
+
+set(CMAKE_C_STANDARD 90)
+set(CMAKE_VERBOSE_MAKEFILE ON)
+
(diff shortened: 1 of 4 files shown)


januscla:
Thanks for your contribution, @SamuelMarks! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
While I appreciate the (huge!) effort, we're not looking for a cmake replacement for autoconf/automake at the moment. Our setup is quite complex, with a lot of optional modules and applications, so it's not just Janus itself. Even with a perfect drop-in replacement (and from a quick glance this is just a portion of the work), considering we're not familiar with cmake (or at least I'm definitely not) this would make it much harder for us to mantain or update when new stuff is added. Besides, we're about to merge the `multistream` branch, and so any refactoring of tooling of these proportions would need to wait until that happens anyway.

SamuelMarks:
Ok maybe just tag me when multistream is merged and I'll see what I can do

lminiero:
Ack, closing this for now.

PR #2747, opened 2021-07-30
fix ssrc header of transport-cc

magiclin99:
If peerconnection contains one send-only video stream and one or more recv-only video stream, for example:

- mid-0 = sendrecv audio
- mid-1 = sendonly video
- mid-2 = recvonly video

`pc->media_bytype` only hold mid-0 and mid-2 medium, and `g_hash_table_lookup(pc->media_bytype, GINT_TO_POINTER(JANUS_MEDIA_VIDEO))` will return mid-2 medium. This cause rtcp report sending with ssrc=0.

Remote peer will slow down RTP transmission because no correct rtcp report send back to it.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3617,7 +3617,21 @@ static gint rtcp_transport_wide_cc_stats_comparator(gconstpointer item1, gconstp
 static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data) {
 	janus_ice_handle *handle = (janus_ice_handle *)user_data;
 	janus_ice_peerconnection *pc = handle->pc;
-	janus_ice_peerconnection_medium *medium = pc ? g_hash_table_lookup(pc->media_bytype, GINT_TO_POINTER(JANUS_MEDIA_VIDEO)) : NULL;
+
+	janus_ice_peerconnection_medium *medium;
+	if (pc) {
+		/* Find inbound video medium */
+		GHashTableIter iter;
+		gpointer value;
+		g_hash_table_iter_init(&iter, pc->media_bymid);
+		while (g_hash_table_iter_next(&iter, NULL, &value)) {
+			janus_ice_peerconnection_medium *m = value;
+			if (m->type == JANUS_MEDIA_VIDEO && m->recv) {
+				medium = m;
+			}
+		}
+	}
+
 	if(pc && pc->do_transport_wide_cc && medium) {
 		/* Create a transport wide feedback message */
 		size_t size = 1300;
--- a/ice.c
+++ b/ice.c
@@ -3618,16 +3618,17 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
 	janus_ice_handle *handle = (janus_ice_handle *)user_data;
 	janus_ice_peerconnection *pc = handle->pc;
 
-	janus_ice_peerconnection_medium *medium;
-	if (pc) {
+	janus_ice_peerconnection_medium *medium = NULL;
+	if(pc) {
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @magiclin99! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

ice.c
```diff
@@ -3617,7 +3617,21 @@ static gint rtcp_transport_wide_cc_stats_comparator(gconstpointer item1, gconstp
 static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data) {
 	janus_ice_handle *handle = (janus_ice_handle *)user_data;
 	janus_ice_peerconnection *pc = handle->pc;
-	janus_ice_peerconnection_medium *medium = pc ? g_hash_table_lookup(pc->media_bytype, GINT_TO_POINTER(JANUS_MEDIA_VIDEO)) : NULL;
+
+	janus_ice_peerconnection_medium *medium;
```
lminiero, line 3621: `medium` should be initialized to `NULL`, since there's no guarantee the check you added will find something (while the `g_hash_table_lookup` we had can return `NULL` in that case).

ice.c
```diff
@@ -3617,7 +3617,21 @@ static gint rtcp_transport_wide_cc_stats_comparator(gconstpointer item1, gconstp
...
+	if (pc) {
```
lminiero, line 3622: Per code stype, there should be no space between `if` and bracket.

ice.c
```diff
@@ -3617,7 +3617,21 @@ static gint rtcp_transport_wide_cc_stats_comparator(gconstpointer item1, gconstp
...
+		/* Find inbound video medium */
+		GHashTableIter iter;
+		gpointer value;
+		g_hash_table_iter_init(&iter, pc->media_bymid);
+		while (g_hash_table_iter_next(&iter, NULL, &value)) {
```
lminiero, line 3627: Same as the `if` above.

ice.c
```diff
@@ -3617,7 +3617,21 @@ static gint rtcp_transport_wide_cc_stats_comparator(gconstpointer item1, gconstp
...
+			janus_ice_peerconnection_medium *m = value;
+			if (m->type == JANUS_MEDIA_VIDEO && m->recv) {
```
lminiero, line 3629: Same as the `if` above. Besides, you should also check if `m` is a valid pointer: as it is, you're dereferencing the pointer right away, which would crash if there's no value.

ice.c
```diff
@@ -3617,7 +3617,21 @@ static gint rtcp_transport_wide_cc_stats_comparator(gconstpointer item1, gconstp
...
+				medium = m;
```
lminiero, line 3630: There should be a `break;` here: no point in going through all the other m-lines after you've found one.

lminiero (commented):
I'm not exactly sure your fix is the way to go. While I understand the issue, we do the same lookup in other areas that may be similarly affected, so it may make more sense to have a different table for this kind of scenarios, or update how we initialize the existing one. Anyway, I've added some notes inline on things that should be fixed, in case we go on with this.

ice.c
magiclin99: fixed

ice.c
magiclin99: fixed

lminiero:
Apologies for the lack of feedback, still catching up with post-vacation activities. I'll get back to you ASAP.

lminiero:
Sorry for the delay, this looks good to me :+1: 
Thanks, merging!

PR #2763, opened 2021-09-01
Fix partial/broken ACL support in TextRoom plugin

lminiero:
As pointed out in a [post on the group](https://groups.google.com/g/meetecho-janus/c/h7kBK9dvBfQ), the TextRoom plugin allows you to configure an ACL (Access Control List) via API, but then it has no way to enforce it. This is indeed a bug in the plugin, probably due to some copy/paste leftovers when we first implemented the feature for multiple plugins in #745.

This PR fixes this broken behaviour. Planning to merge soon, as soon as we have feedback that it works as expected, so if you're using the plugin you may want to double check it doesn't introduce regressions for you (it shouldn't).


techwipes:
Good day! 
I am the creator of the post. I compiled a docker image from the current branch and tested it on my project. The fixes helped. Thanks!

lminiero:
Ack, thanks for confirming, I'll merge then :+1:

PR #2764, opened 2021-09-03
Fix streaming mountpoint name generation when starting with 0

lionelnicolas:
Strings starting with `0` and have a following non-zero digits were ok, but not the one with `0` + letters.

```
atoi("0123abc") = 123
atoi("004abc") = 4
atoi("000abc") = 0     <--- as a result this one was missing the mp- prefix
```


lminiero:
Good catch, thanks, merging! :+1:

PR #2768, opened 2021-09-14
handle host being in ignore list and enforced list

danjenkins:
If you add a eth adapter to enforced of say `eth0` and it has 3 IPs attached to it, but you want to remove one of them, you'd expect to be able to add that IP (or partial IP) to the ignore list and have it removed. Currently it doesnt work like that and doesnt allow us the flexibility over eth adapters with multiple IPs you'd expect

By making this change we get the flexibility of being able to pass in these options 

```
    --ice-enforce-list=eth0
    --ice-ignore-list=10.16.0.
```

eth0 has 3 IPs attached to it, two ipv4 and one ipv6 - one of the ipv4's being a private address. Now I can remove that private IP from being offered as a candidate

The change means I get 2 candidates instead of 3.

More control over candidates presented = win.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3630,10 +3630,11 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 			if(janus_ice_enforce_list != NULL) {
 				if(ifa->ifa_name != NULL && !janus_ice_is_enforced(ifa->ifa_name) && !janus_ice_is_enforced(host))
 					continue;
-			} else {
-				if(janus_ice_is_ignored(host))
-					continue;
 			}
+			
+			if(janus_ice_is_ignored(host))
+				continue;
+			
 			/* Ok, add interface to the ICE agent */
 			JANUS_LOG(LOG_VERB, "[%"SCNu64"] Adding %s to the addresses to gather candidates for\n", handle->handle_id, host);
 			NiceAddress addr_local;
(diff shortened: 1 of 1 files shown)


ice.c
```diff
@@ -3630,10 +3630,11 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
 			if(janus_ice_enforce_list != NULL) {
 				if(ifa->ifa_name != NULL && !janus_ice_is_enforced(ifa->ifa_name) && !janus_ice_is_enforced(host))
 					continue;
-			} else {
-				if(janus_ice_is_ignored(host))
-					continue;
 			}
+
```
atoppi, line 3634: please avoid empty lines

ice.c
```diff
@@ -3630,10 +3630,11 @@ int janus_ice_setup_local(janus_ice_handle *handle, int offer, int audio, int vi
...
 				if(ifa->ifa_name != NULL && !janus_ice_is_enforced(ifa->ifa_name) && !janus_ice_is_enforced(host))
 					continue;
-			} else {
-				if(janus_ice_is_ignored(host))
-					continue;
 			}
+			
+			if(janus_ice_is_ignored(host))
+				continue;
+
```
atoppi, line 3637: empty line

atoppi (commented):
Thanks @danjenkins, I have left a couple of notes.

I'd change the comment
```c
/* Check if this IP address is in the ignore/enforce list, now: the enforce list has the precedence */
```
by making clear that enforce list has the preference on the ifaces but the ignore list can still discard the host candidate.

atoppi:
lgtm @danjenkins, merging.
Thanks for the contribute!

PR #2769, opened 2021-09-16
REMB based simulcast layer switching on a subscriber peerConnection

JanFellner:
What janus currently sadly lacks in is tried to be covered with this PR.
It's a first idea and definitely needs fine tuning and further improvements (more later)

The idea:
On the publisher side the sending browser controls which simulcast layers it shall send based on the available bit rate it sees using transport-cc. Transport-cc is pretty complex but @fippo came up with janus is missing some header to support REMB.
(If you are not familiar with either transport-cc or REMB. Both try to gather the available bit rate on a PeerConnection.
Transport-cc is would be done on the sender side (for the subscriptions janus) where REMB is done on the client side (browser for the subscriptions) 
REMB is slower than Transport-cc, currently lacks in probing (therefore just estimates based on the data currently being send) but its a starting point
(As it estimates, without probing, based on the data being send the value on REMB does not really see whats possible.
e.g. If you transport 150kBit and have no technical limit on the PC the REMB estimation goes up to ~280kBit.

If we know the bandwidth we can try to map the available to the bit rates we need to the different simulcast layers.
We currently do not map that to the real bit rates of the layers janus receives on the publisher PeerConnection.
We map it to what we told the browser to use for the different layers. In order to use this implementation you need to let janus know what the browser will be configured for the different layers.
We do that with adding a JSON array with the bit rates [L,M,H] in either the configure or the join as publisher command

We now build a ramp for going to higher or lower bit rates. We switch not only the substream but also the temporal layers as we otherwise have blind areas where the REMB does not reach a value that allows us to switch to a higher layer.
The ramp consists of 6 positions L15fps, L30fps, M15fsp, M30fps, H15fps, H30fps.
These ramp position are associated with a certain bit rate we need to see to be able to switch to it.

If the REMB value now goes above the required bit rate for the next layer or falls below the bit rate of the current layer.
A helper variable is increased or decreased (this ensures that we do not switch instantly but if the new bit rate has settled).
If the variable falls to -20 we ramp down, if it goes to 20 we ramp up.

The logic takes in charge what the client wanted to receive on a subscription.
Thus we expose if we are in a limit situation through the eventing.
The videoroom subscription events announce if we were forced to limit the substream or temporal layers where the client actually wanted to receive higher layers.
(substream_remb_limited_to / temporal_remb_limited_to)

Whats missing:
* Expose the currently seen REMB value in the general peerConnection statistics (just missing for completeness as other pc related values are exposed and helpful for external debugging)
* REMB probing on the subscription peerConnection (would speed up the REMB detection and would allow a switching without changing the temporal layers)
* transport-cc which would dramatically improve the probing (hard to implement especially in C, so nothing we target now)
* Gather the currently really used bit rates directly from the publishers peerConnection and not from the configure/join values
* REMB trends sometimes to signal bit rates which are actually no available so we need to ensure that we do not cycle through the ramp all the time
* Support for VP9 (the current approach only targets VP8)

I am still working on that topic but wanted to receive some feedback as i know have a working solution.


JanFellner:
It´s pretty much code but only functional if you set the mentioned simulcast bitrates on the publishers peerConnection.
Otherwise the whole code is dormant.

Tried to comment it as best as possible to have it understandable for everyone interested in that PR

roisasson:
Hello Jan. Thank you for taking this on. It’s a great initiative and an important addition to simulcast support in Janus. I skimmed through the code to understand your approach and it makes sense. I hope to have some time soon to set it up and try it out in practice.
 
One minor comment. I think that in Janus_videoroom.c lines 8342-8343 the if statement might be wrong
```
   else if(remb > higher_bitrate) // If we are just below the lower bitrate
      pRembContext->substream_switch_layer_counter -= 1;
```
Maybe I am just missing your logic – can we have (remb < current_bitrate) and (remb > higher_bitrate) at the same time?

JanFellner:
Hi @roisasson.

Great it someone is catching up and has the same thinking.
I am still working on it and currently investigating into cycling issue.
As the remb based bitrate is currently not supporting probing i need to detect if we do ramp up but the available bitrate in the end is not fit the need, so we drop again. It´s a bit tricky but i hope to have that solved today or tomorrow.

Concerning your finding i need to check if the current code is still making use of it :). Will check and keep you posted.

@lminiero i would like to completely separate the remb based stuff from the videoroom as the file is already huge.
To do so it would be neccessary to split the code into declaration and implementation. As all the plugins currently follow a single file approach (just the c file) i would like to get your personal feeling wether you would support if we split decl and impl into to files to have this businesslogic in a separated file. (I would then do another PR just for the splitting of the file)

Jan

lminiero:
@JanFellner first of all, thanks for this huge effort! I haven't looked at it yet because it's quite big, touches a lot of parts, and I don't have the time to review it right now.

From a quick glance, my first feedback would be that I'd definitelty move all the code you added to the VideoRoom out of the plugin. It's not just a matter of having other code in a monolithic .c file, as you said, but that you're adding features other plugins can benefit from too. Take the simulcast implementation we have, for instance: it's all in `utils.c`/`utils.h`, as that way the VideoRoom can use it, but so can EchoTest, Record&Play, Lua/Duktape and others. This would also greatly simplify the integration in the multistream branch, in the future. As it is now, it would cause a sh*tstorm of conflicts, and so I would never merge it.

JanFellner:
@lminiero Thanks for the short feedback. It is still some work to be done. As soon as others also hop on the train (Start to use it) hopefully will also face other improvements cause what i currently do comes just from the way i do it right now and others might also have great ideas on improving it.
So i would put it in a dedicated vp8_remb_simulcast file so in further iteration we would only have the changes condensed in that one file.
Will check if i get it working without splitting the plugin into decl and impl. 
*STAY* tuned 🧐

JanFellner:
I have a good feeling but invite others to jump on the train cause velocity is currently pretty good and i would like to keep things moving...

@lminiero i separated the logic as proposed into a dedicated file which makes following what actually changes in the implementation pretty easy.
I need help from your side for certain pieces that are hard to solve for me. e.g.

- Do you see any chance to implement probing? 
According to @fippo it´s the only way to get proper remb values and especially to speed up the ramping up as otherwise it takes pretty long
- Can you guide me (or maybe it´s even easier for you if you take a look) on how to get the remb values into the statistics. I use the statistics janus provides me to understand whats going on in the real world. The remb value is mandatory to understand why streams have switch but i don´t find the proper path to get the value into the stats. (Also the stats currently separate audio and video but the remb value would cover both...)

⚠Whats now missing:
- Currently the videoroom starts to send the layers as requested by the API. Then REMB comes into place and reduces the layers as the reported REMB value is not fullfilling the needs so it is forced to switch.

✅Options to address this:
- We start with a layer that is fullfilled by the first REMB values we receive thus are not forced to reduce the bitrate (e.g. Ramp Position 1 (Low quality, 30fps)
- Or we use the % factor the remb value is changing and do not reduce 

Everybody, if you think or hope this topic might help please leave a thumb up :)

roisasson:
Maybe probing can be implemented by having the plugin relay some RTP packets multiple times? I believe that will drive up the REMB estimates without impacting the receivers, as they should just discard redundant payloads. I am not familiar enough with the code so this idea might not make much sense.

OxleyS:
Thank you for this work! The issue of automatic simulcast layer selection just cropped up in our application too, so it's great to see this.

On current `master`, each subscriber peer connection in a videoroom is associated with at most one video stream, but on the `multistream` branch, a subscriber connection can have an arbitrary number of video streams. In that case, if the bandwidth estimate indicates that Janus is sending too much data, how should we choose which streams to switch to a lower quality?

We could, for example, simply drop all the video streams down a layer, or we could have a greedy algorithm that picks streams based on how much bandwidth it would save, or we could simply arbitrarily choose streams until we're within budget again.

That might be a difficult question to answer in a way that all applications would be happy with, though. For my application's purposes, I would be happy with just having the estimated bandwidth in stats, so that the app can make its own decision. That's not to say that Videoroom can't have default behavior we could opt-out of, though.

JanFellner:
@OxleyS seeing the remb value in the stats would also allow me to do stuff on the client side but i did currently decide not to impl it on the client side as i would then have to deal with it in different products and impl in the end... :)
Already asked @lminiero #https://github.com/meetecho/janus-gateway/pull/2769#issuecomment-923952775 that having the remb bitrate estimates in the stats would improve and help in understanding whats going on but i do not really find a way (or maybe i am just too blind) in getting them into the stats. Also the stats may need an own element to encapsulate PC related data loosen from the media that it transports...

JanFellner:
Concerning the multistreaming it needs another iteration cause likely you want to leverage the stream quality for the different streams and optionally exclude some (e.g. you have some speaker representation that needs larger streams where you have others as tiny tiles beside) so that depends on the use case scenario.
I wanted to start with this idea as an example. In the end the multistreaming will need further iterations on adopting it....

JanFellner:
Last thing todo now is to be able to control how the subscription shall behave when i start or configure it:

- enforce a certain higher layer to start with (if i am using chromium based or safari, that works out of the box) **(START HIGH)**
- tell janus todo the ramp up (start with a limit and let janus increase that limit if the remb value is high enough) **(RAMP UP)**
- tell janus not to do and remb adoption for this subscription **(IGNORE)**

That has to be added to the configure or subscribe command and handled accordingly.

JanFellner:
@roisasson According to @fippo the way others do it is to use an unuses SSRC and send dummy data in certain time frames. But i have no clue how to finally get this going.... I would not duplicate elements as the wireshark trace might be pretty ugly in the end 💣

lminiero:
Probing is a completely separate effort, that we might investigate in the future (don't expect anything short/medium term though).

roisasson:
> @roisasson According to @fippo the way others do it is to use an unuses SSRC and send dummy data in certain time frames. But i have no clue how to finally get this going.... I would not duplicate elements as the wireshark trace might be pretty ugly in the end 💣

Yes, that would be the the more common method. I was just throwing out an idea that may be easier to implement in the plugin without extensive changes to the core. As you rightfully point out it has its downsides.

JanFellner:
@lminiero  can you also just shortly comment on the questions about the REMB value in the statistics? Its also loosen from that one here but in the end really needed to understand from the external what janus is doing and why. I did not find the right code path to get the REMB value into the stats so i cannot really make a PR out of it.
Either you give me a short guidance or shall we discuss that in google groups or shall i make an FR issue  :) out of it?

lminiero:
I don't understand why you need REMB from the stats, the plugin already has access to the REMB sent by subscribers. I'll comment when I've had a chance to review the code.

JanFellner:
Its about putting the remb value into the stats janus is exposing either through the eventing websocket or the admin api.

JanFellner:
It´s feature complete and looking good.
We will now start to use it in our dev environment to see how it behaves in the deployed development environment.
Therefore i change this PR from WIP to ready for review.

Happy to receive comments on it.

lminiero:
Hi @JanFellner, thanks for the update! Looking forward to news about whether or not it works as expected in your internal tests. Due to the volume of the patch and a few other activities I'm taking care of right now, it may take me a bit to find time to a first review, but I'll try to do that ASAP, possibly by early next week.

JanFellner:
We shifted the test as i need to have the client reported REMB value in the statistics, otherwise interpreting whats happening is quite difficult. (i see that something is happening but i cannot really interpret if the business logic is doing what it is supposed to)
So i am trying to find the way to expose them in the stats (fiddeling around in ice.c and rtcp.c to find the way how to get them in the stats...) (will create a dedicated PR for it...)

JanFellner:
Added a first PR to get media statistic events combined...
https://github.com/meetecho/janus-gateway/pull/2785

Will now try to find a way how to add this REMB value to the combined stats.

lminiero:
> We shifted the test as i need to have the client reported REMB value in the statistics, otherwise interpreting whats happening is quite difficult.

That's something I can add easily, I think, both to the event handler stats and the Admin API info. Normally we only read the REMB feedback in plugins, but the method to read that info is available as a core function, so it's easy to use there too. Let me know if you want me to add that piece of info.

JanFellner:
> That's something I can add easily, I think, both to the event handler stats and the Admin API info. Normally we only read the
> REMB feedback in plugins, but the method to read that info is available as a core function, so it's easy to use there too. Let me  know if you want me to add that piece of info.

That would be really awesome cause for me it is really hard to find the right path between the rtcp layer that fetches the stuff  and the ice layer that reports it for the eventing and the admin api.
There is (at least i did not find it) no general rtcp info object that covers information that is not related to the audio or video media (e.g. rtt would be something that is related to the pc)?

I would see those elements on a different level in the statistic events but that will be hard to change now but maybe for the multistreaming there needs to be general information loosen from the track anyway so things can be adopted there...

lminiero:
@JanFellner I added the REMB info in this commit here: https://github.com/meetecho/janus-gateway/commit/0bb2064dd63428cdc99733a45adee1cce3274193
Notice that, in its current version, it's only added to the base video layer (so `0` in case of simulcast), since REMB is a global PeerConnection value and so it made little sense to repeat the same info in audio or other video reports. It's in a different property in the Admin API instead. I'll think about how to report it in multistream, where there could be multiple video streams instead.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
+/* New methods for an REMB based simulcast layer switching on a subscriber peerConnection
```
lminiero, line 1: Please use the same templating we have in other files for the initial comments part. That's used by Doxygen to automatically create pages out of source files, and the related detailed information that describes its contents.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+void janus_vp8_remb_simulcast_get_simulcast_current_debug(char *szBuffer, gsize bufSize, janus_rtp_simulcasting_context *pSim);
+
+/* Reads the remb_adoption property from the handed over json property and sets it
+ * accordingliy in the simulcasting_remb_context context, returns false if the property has had
+ * an invalid value
+ * @param config_property - the property that contains the remb_adoption config value
+ * @param pRemb_context - struct where we set the config_property into
+ * @returns false in case the property contained an invalid value, true in any other case (not set or valid configured)
+ */
+gboolean janus_vp8_remb_simulcast_get_remb_adoption_config(json_t *config_property, janus_rtp_simulcasting_remb_context *pRemb_context) {
```
lminiero, line 97: Camel case.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+const char* janus_vp8_remb_simulcast_get_debug_temporal_to_text(uint32_t temporal);
+void janus_vp8_remb_simulcast_get_simulcast_requested_debug(char *szBuffer, gsize bufSize, janus_rtp_simulcasting_context *pSim);
+void janus_vp8_remb_simulcast_get_simulcast_current_debug(char *szBuffer, gsize bufSize, janus_rtp_simulcasting_context *pSim);
+
+/* Reads the remb_adoption property from the handed over json property and sets it
+ * accordingliy in the simulcasting_remb_context context, returns false if the property has had
+ * an invalid value
+ * @param config_property - the property that contains the remb_adoption config value
+ * @param pRemb_context - struct where we set the config_property into
+ * @returns false in case the property contained an invalid value, true in any other case (not set or valid configured)
```
lminiero, line 95: The `@param` and `@returns` are only needed in headers, as they're used by Doxygen to automatically build documentation for functions. The description of functions in the `.c` files should be limited to a one-liner.

As a side note, the vast majority of functions in here seem to be private to this file, since only `janus_vp8_remb_simulcast_based_subscriber_simulcast_switching` and `janus_vp8_remb_simulcast_get_remb_adoption_config` are exposed. As such, all the others should be defined as `static` instead, since they'll be confined in here.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+/* Forward declarations */
+uint32_t janus_vp8_remb_simulcast_get_client_requested_ramp_position(janus_rtp_simulcasting_context *pSim);
+uint32_t janus_vp8_remb_simulcast_get_current_ramp_position(janus_rtp_simulcasting_context *pSim);
+uint32_t janus_vp8_remb_simulcast_get_bitrate_for_ramp_position(janus_rtp_simulcasting_remb_context *pRembContext, uint32_t rampPos);
+void janus_vp8_remb_simulcast_get_neighbour_ramp_positions(janus_rtp_simulcasting_context *pSim, uint32_t rampPos, uint32_t *pNextLower, uint32_t *pNextUpper);
+double janus_vp8_remb_simulcast_calculate_bitrate_change(janus_vp8_remb_subscriber *pSubscriber, uint32_t remb);
+void janus_vp8_remb_simulcast_update_last_bitrate_change(janus_rtp_simulcasting_remb_context *pRempContext, double change);
+gboolean janus_vp8_remb_simulcast_set_flags_for_ramp_position(janus_rtp_simulcasting_remb_context *pRembContext, uint32_t position);
+const char* janus_vp8_remb_simulcast_get_debug_substream_to_text(uint32_t substream);
+const char* janus_vp8_remb_simulcast_get_debug_temporal_to_text(uint32_t temporal);
```
lminiero, line 86: Code style: I've noticed this in a few other areas of the code as well, our code style for pointers is

    type *pointer

not

    type* pointer

and the same for functions that return a pointer, so

    type *function(..)

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+
+/* Reads the remb_adoption property from the handed over json property and sets it
+ * accordingliy in the simulcasting_remb_context context, returns false if the property has had
+ * an invalid value
+ * @param config_property - the property that contains the remb_adoption config value
+ * @param pRemb_context - struct where we set the config_property into
+ * @returns false in case the property contained an invalid value, true in any other case (not set or valid configured)
+ */
+gboolean janus_vp8_remb_simulcast_get_remb_adoption_config(json_t *config_property, janus_rtp_simulcasting_remb_context *pRemb_context) {
+	// Set the default
```
lminiero, line 98: Code style: comments.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+			pRemb_context->remb_adoption = janus_vp8_remb_adoption_ignore;
+		} else if(!strcasecmp(szRembAdoption, "ramp_up")) {
+			pRemb_context->remb_adoption = janus_vp8_remb_adoption_ramp_up;
+		} else if(!strcasecmp(szRembAdoption, "start_high")) {
+			pRemb_context->remb_adoption = janus_vp8_remb_adoption_start_high;
+		} else {
+			return FALSE;
+		}
+		if(pRemb_context->remb_adoption == janus_vp8_remb_adoption_ramp_up) {
+			// We start with a low quality stream and 15fps
```
lminiero, line 112: Code style: comments.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+	}
+
+	return TRUE;
+}
+
+/* This method implements the remb based subscriber simulcast switching as described above
+ * @param pSubscriber - the subscriber object we are handling
+ * @param bitrate - the received REMB bitrate value on the peerConnection
+ */
+void janus_vp8_remb_simulcast_based_subscriber_simulcast_switching(janus_vp8_remb_subscriber *pSubscriber, uint32_t bitrate) {
```
lminiero, line 125: There should be a `NULL` check for on the pointer we receive.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+
+	/* Subscriber did not configure remb adoption -> nothing todo */
+	if(pRembContext->remb_adoption == janus_vp8_remb_adoption_ignore)
+		return;
+
+	/* Did the publisher announce layer bitrates via the api? */
+	if(bitrate && pRembContext->publisher_simulcast_layer_count && pSim->substream != -1) {
+		/*! Calculate the % change from the last to the current remb values */
+		double dbBitrateChange = janus_vp8_remb_simulcast_calculate_bitrate_change(pSubscriber, bitrate);
+		/*! Whats the ramp position we are currently sending and the one the client requested? */
```
lminiero, line 138: See note above on unneeded `!` in comments.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+        uint32_t currentRampPos = janus_vp8_remb_simulcast_get_current_ramp_position(pSim);
```
lminiero, line 139: Broken indentation (spaces instead of tabs).
It actually looks like the whole rest of the function has broken indentation too?

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+		uint32_t requestedRampPos = janus_vp8_remb_simulcast_get_client_requested_ramp_position(pSim);
+
+        if((int)requestedRampPos != pRembContext->last_requested_ramp_position) {
+            // What the client requested has changed...
```
lminiero, line 143: Code style: comments (applies to the whole method, many occurrences).

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+    		if(currentRampPos > 0 && bitrate < current_bitrate) {
+                /* We are above layer 0 and the bitrate is currently lower than we need for the current layer */
+    			if(bitrate < current_bitrate * 0.8 && compareValue < 0.2)	// If we are 20% below the lower bitrate and the bitself itself is not increasing by at least 0.2% between two remb messages
+    				pRembContext->substream_switch_layer_counter -= 4;
+    			else if(bitrate < current_bitrate * 0.9 && compareValue < 0.5) // If we are 10% below the lower bitrate and the bitself itself is not increasing by at least 0.5% between two remb messages
+    				pRembContext->substream_switch_layer_counter -= 2;
+    			else if(compareValue < 1) // If we are just below the lower bitrate and the bitrate is not growing by 1% between two remb messages
+    				pRembContext->substream_switch_layer_counter -= 1;
+    		}
+    		else if(bitrate > current_bitrate) {
```
lminiero, line 185: Code style: `else` should be on the same line as the closing bracket.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+    					pRembContext->substream_switch_layer_counter += 2;
+    				else  // If we are just above the higher bitrate
+    					pRembContext->substream_switch_layer_counter += 1;
+    			}
+    		}
+        }
+
+		double db_time_since_last_ramp_up = 0;
+		if(pRembContext->tm_last_ramp_up) {
+			db_time_since_last_ramp_up = (double)(g_get_monotonic_time() - pRembContext->tm_last_ramp_up) / 1000000;
```
lminiero, line 204: You can use `G_USEC_PER_SEC` instead of `1000000` which will ensure no risk of typos.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+			if(pRembContext->failed_highest_ramp_pos >= 0 && db_time_since_last_ramp_up > TIME_RAMP_UP_FAILED && (int)currentRampPos >= pRembContext->failed_highest_ramp_pos) {
+				/* Ramp up succeeded -> cleanup failing information */
+				pRembContext->failed_highest_ramp_pos = -1;
+				pRembContext->failed_counter = 0;
+			}
+		}
+
+#ifdef REMB_CONSOLE_DEBUGGING
```
lminiero, line 212: I assume this is temporary and just for your internal debugging, correct?

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+}
+
+/* Calculates the % delta between the current and the last remb messages.
+ *
+ * @param pSubscriber - the subscriber object we are handling
+ * @param bitrate - the received REMB bitrate value on the peerConnection
+ * @returns the delta between the last and the current remb value as signed % value (-100 <-> +100)
+ */
+double janus_vp8_remb_simulcast_calculate_bitrate_change(janus_vp8_remb_subscriber *pSubscriber, uint32_t bitrate) {
+	double change = 0;
```
lminiero, line 287: Missing `NULL` check on provided pointer.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+/* Updates the % remb change value in the janus_rtp_simulcasting_remb_context
+ * the value always covers the median change over the last two remb values
+ * (add the current value to the last_bitrate_change, if the value was set before divide it by 2)
+ * due to this mathematic approach it is a dragged/towed value that gives you a feeling about the change in the last requests
+ *
+ * @param pRembContext - the currently used REMB simulcasting context
+ * @param change - the % change value as calculated above
+ */
+void janus_vp8_remb_simulcast_update_last_bitrate_change(janus_rtp_simulcasting_remb_context *pRempContext, double change) {
+	if(pRempContext->last_bitrate_change) {
```
lminiero, line 302: Missing `NULL` check on provided pointer.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+	}
+}
+
+/* Retrieve the ramp position the client has chosen to receive
+ *
+ * @param pSim - the currently used simulcasting context
+ * @returns the current ramp position based on the currently send substream and temporal layer
+ */
+uint32_t janus_vp8_remb_simulcast_get_client_requested_ramp_position(janus_rtp_simulcasting_context *pSim) {
+	uint32_t rampPos = 0;
```
lminiero, line 316: Missing `NULL` check on provided pointer.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+			break;
+		default:
+			break;
+	}
+
+	// We had a failing in ramping up, so we adopt the bitrate of the failing layer by 5% for each failing
+	// 1. failing we require 5% more
+	// 2. failing we require 10% more ...
+	if(pRembContext->failed_highest_ramp_pos != -1 && pRembContext->failed_highest_ramp_pos == (int)rampPos) {
+		next *= 1 + pRembContext->failed_counter * REQUIRED_BITRATE_INCREASE_PER_FAIL;
```
lminiero, line 398: I'd suggest using parentheses here to make it more readable and less ambiguous.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+	int nextHigher = rampPos;
+
+	if(nextLower > 0) {
+		/* Decrease the ramp Position */
+		nextLower--;
+
+		/* Check that a preselected temporal from the client does not conflict our ramping up */
+		if(nextLower % 2) {
+			/* 1,3,5 are the layers where we switch to 30fps, */
+			if (pSim->templayer_target == 0 || pSim->templayer_target == 1) {
```
lminiero, line 425: Code style: `if (`

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+	}
+
+	if(nextHigher < 5) {
+		/* Increase the ramp Position */
+		nextHigher++;
+
+		/* Check that a preselected temporal from the client does not conflict our ramping up */
+		if(nextHigher % 2) {
+			/* 1,3,5 are the layers where we switch to 30fps, */
+			if (pSim->templayer_target == 0 || pSim->templayer_target == 1) {
```
lminiero, line 443: See above.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+		default:
+			return FALSE;
+	}
+}
+
+/*
+ * Retrieve diagnostic log text for a substream layer (L,M,H)
+ *
+ * @param substream - the substream layer we want to have the text for
+ * @returns - L,M,H for the different substreams or unknown if the value could not get converted
```
lminiero, line 511: To match the style we use in the core (e.g., in Admin API), these should be "l", "m" and "h", not the capitalized versions.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+}
+
+/*
+ * Creates diagnostic log for the requested simulcast layers
+ *
+ * @param szBuffer - the buffer to copy the log to
+ * @param bufSize - the buffer size in bytes
+ * @param pSim - the simulcasting context
+ */
+void janus_vp8_remb_simulcast_get_simulcast_requested_debug(char *szBuffer, gsize bufSize, janus_rtp_simulcasting_context *pSim) {
```
lminiero, line 554: Missing `NULL` check on pointers.

vp8_remb_simulcast.c
```diff
@@ -0,0 +1,571 @@
...
+}
+
+/*
+ * Creates diagnostic log for the currently transported simulcast layers
+ *
+ * @param szBuffer - the buffer to copy the log to
+ * @param bufSize - the buffer size in bytes
+ * @param pSim - the simulcasting context
+ */
+void janus_vp8_remb_simulcast_get_simulcast_current_debug(char *szBuffer, gsize bufSize, janus_rtp_simulcasting_context *pSim) {
```
lminiero, line 567: Missing `NULL` check on pointers.

vp8_remb_simulcast.h
```diff
@@ -0,0 +1,43 @@
+#include <glib.h>
+#include "rtp.h"
+#include "mutex.h"
+
+/* This is a helper object that is used to pass publisher related data to the methods of the vp8_remb_simulcast implementation */
```
lminiero, line 5: The struct and properties need the `! \brief` prefix to be detected by Doxygen.

vp8_remb_simulcast.h
```diff
@@ -0,0 +1,43 @@
...
+	/* Display name (just for fun) */
+	gchar *display;
+ 	/* Pointer to the original publisher object, to be handed over in the callback isMultiCasting */
+	void *pOriginalPublisherObject;
+ 	/* Callback that is called inside a mutex context to verify if the publisher is simulcasting */
+	gboolean (*isMultiCasting)(void *pOriginalPublisherObject);
+	/* Callback to request a PLI / FIR on the subscribers peerConnection */
+	void (*sendPLI)(void *pOriginalPublisherObject, const char *reason);
+
+} janus_vp8_remb_publisher;
```
lminiero, line 18: Unneeded empty line above.

vp8_remb_simulcast.h
```diff
@@ -0,0 +1,43 @@
...
+
+/* This is a helper object that is used to pass subscriber related data to the methods of the vp8_remb_simulcast implementation */
```
lminiero, line 20: See above.

vp8_remb_simulcast.h
```diff
@@ -0,0 +1,43 @@
...
+	/* Simulcasting context of the subscribers peer Connection */
+	janus_rtp_simulcasting_context *pSimContext;
+	/* Mutex one has to lock to safely access the publishers feed */
+	janus_mutex* pFeedMutex;
+	/* Pointer to the publisher of this subscription */
+	janus_vp8_remb_publisher *pFeed;
+} janus_vp8_remb_subscriber;
+
+/* Entry point to allow remb based switching simulcast layers
+   This method needs to be called when the remb value has been aquired from the rtcp data */
```
lminiero, line 37: Missing `! \brief` and the `@param` you added in the `.c` instead.

vp8_remb_simulcast.h
```diff
@@ -0,0 +1,43 @@
...
+void janus_vp8_remb_simulcast_based_subscriber_simulcast_switching(janus_vp8_remb_subscriber *pSubscriber, uint32_t bitrate);
+
+/* Reads the remb_adoption property from the handed over json property and sets it
+   accordingliy in the simulcasting_remb_context context, returns false if the property has had
+   an invalid value */
+gboolean janus_vp8_remb_simulcast_get_remb_adoption_config(json_t *config_property, janus_rtp_simulcasting_remb_context *remb_context);
```
lminiero, line 43: See above. Besides, the file is missing an empty line at the end.

vp8_remb_simulcast.h
JanFellner: Adopted

vp8_remb_simulcast.h
JanFellner: Adopted

vp8_remb_simulcast.h
JanFellner: Adopted

JanFellner:
I adopted all coding style related recommendations.
If you dislike the p_ for pointers let me know. I find it way better readable but thats i guess personal feeling.

I am sadly still working on other stuff to get it deployed in our dev landscape...
(Just improved the statistic backend to see whats going on with the remb values and the limits janus is signalling when there are limits)

lminiero:
@JanFellner is this effort still relevant, or have you stopped using it when migrating to multistream? I'm trying to understand if we should close this, or if then plan is to align it again in the future (it seems it conflicts with `0.x` at the moment) and share some of the results you collected using it in the past. Thanks!

JanFellner:
Let´s close it. Booked the effort on learning janus insides...

lminiero:
Ack, closing then. Thanks for working on this, and hope it provided enough info to reboot it on multistream someday!

PR #2772, opened 2021-09-20
Make janus.js pass linter

davel:
Hello,

I have attempted to get `janus.js` to pass a JavaScript linter. The changes mostly relate to variable scope one way or the other, I hope they are improvements. I have also enabled `"use strict";` as the code appears to be safe with regards to that. Testing with our application does not find any problems with these changes.

I have also configured GitHub actions to automatically run eslint along with the rest of the CI process.


lminiero:
I have no idea what linter is, so I'll leave considerations to younger folks like @atoppi :slightly_smiling_face:

.eslintrc.js
```diff
@@ -0,0 +1,267 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es6": true
+    },
+    "extends": "eslint:recommended",
+    "parserOptions": {
+        "ecmaVersion": 2018
```
atoppi, line 8: is this really needed ?

.eslintrc.js
```diff
@@ -0,0 +1,267 @@
...
+    },
+    "rules": {
```
atoppi, line 10: Can you explain where does this huge set of rules come from ?

package.json
```diff
@@ -14,6 +14,7 @@
   },
   "homepage": "https://github.com/meetecho/janus-gateway#readme",
   "dependencies": {
-    "webrtc-adapter": "8.0.0"
+    "webrtc-adapter": "8.0.0",
+    "eslint": "5.0.1"
```
atoppi, line 18: please move `eslint` to `devDependencies` since it is only needed for developing and testing purposes

html/janus.js
```diff
@@ -95,6 +97,7 @@ Janus.useDefaultDependencies = function (deps) {
 		newWebSocket: function(server, proto) { return new socketCls(server, proto); },
 		extension: (deps && deps.extension) || defaultExtension,
 		isArray: function(arr) { return Array.isArray(arr); },
+		/*global adapter */
```
atoppi, line 100: imho I'd prefer adding the globals to eslint config instead of polluting the script with comments like this

html/janus.js
```diff
@@ -2384,17 +2381,19 @@ function Janus(gatewayCallbacks) {
 							});
 						return;
 					}
+					// FIXME what is the correct scope of
+					// callbackUserMedia and getScreenMedia?
```
atoppi, line 2385: the scope is the one where the arguments of `streamsDone` are visibile.

atoppi (commented):
Hi @davel thanks for the effort.
I've left some notes, my only concern is about the rule set you are using in eslint.
What is the source / rationale of that ? It is way too large to check if any rule is really needed.

html/janus.js
```diff
@@ -2301,46 +2306,38 @@ function Janus(gatewayCallbacks) {
 						videoSupport = media.video;
 					} else {
 						var width = 0;
-						var height = 0, maxHeight = 0;
```
atoppi, line 2304: why `maxHeight` has been deleted ?

.eslintrc.js
davel: The rules were auto-generated by eslint. However, it looks like we only need three exceptions from the "recommend" defaults, I shall revise this.

package.json
davel: Will do.

html/janus.js
davel: Will do.

html/janus.js
davel: `maxHeight` does not appear to be used by anything.

html/janus.js
davel: I shall strip this out. Nesting a function below the top level of another function fires this warning, https://eslint.org/docs/rules/no-inner-declarations

html/janus.js
```diff
@@ -230,8 +232,8 @@ Janus.init = function(options) {
 		// Already initialized
 		options.callback();
 	} else {
-		if(typeof console == "undefined" || typeof console.log == "undefined") {
-			console = { log: function() {} };
+		if(typeof console.log == "undefined") {
```
lminiero, line 235: Are you 100% sure `console` is always defined?

html/janus.js
```diff
@@ -2386,15 +2380,15 @@ function Janus(gatewayCallbacks) {
 					}
 					// We're going to try and use the extension for Chrome 34+, the old approach
 					// for older versions of Chrome, or the experimental support in Firefox 33+
-					function callbackUserMedia (error, stream) {
+					function callbackUserMedia (error, stream) { // eslint-disable-line no-inner-declarations
```
lminiero, line 2383: I don't like these comments unrelated to the code lingering here. Is there any alternative?

lminiero (commented):
Added a few (clueless) comments.

.github/workflows/janus-ci.yml
```diff
@@ -152,3 +152,11 @@ jobs:
         ./configure $JANUS_CONFIG_COMMON $JANUS_CONFIG_OPTS
          make -j$(nproc)
          make check-fuzzers
+  javascript-lint:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install modules
+      run: npm install --only=dev
```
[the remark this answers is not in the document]
atoppi: I have tested the workflow locally and it seems to work as intended.
Github Actions images [are shipped with the latest LTS version](https://github.com/actions/virtual-environments/issues/1953).

html/janus.js
```diff
@@ -1,3 +1,5 @@
+"use strict";
```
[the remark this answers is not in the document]
atoppi: I'm not aware of any modern browser / runtime not supporting strict mode.
What will be impacted is the way you'll write the code though ;-)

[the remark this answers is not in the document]
Frenzie: 'use strict' itself has been supported by all browsers since 2011/2012 (so even if supporting, say, IE10 were a goal it'd work); of course that doesn't necessarily mean it'll Just Work™ with the code as is. ^_^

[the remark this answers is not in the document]
lminiero: What I want to ensure is that demos won't start breaking because of this, as I've written them the same way at the time.

html/janus.js
Frenzie: in the form maxHeight it's for old Chromia, but note that you can still pass min and max (which doesn't make much sense if they're exactly the same, of course)

https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia

.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es6": true,
+        "jquery": true
```
Frenzie, line 5: Is that really necessary?

davel: I believe so. We're using browser objects, ECMAScript 6 features (such as `const`), and the `jQuery` object.

lminiero: janus.js doesn't use jQuery, only the demos do.

Frenzie: To be fair, it'd make sense to extend the linting to all *.js files in a follow-up. Personally I'd leave this line out of this specific PR though since it seems to serve no purpose.

lminiero: I would 100% object linting the demos. The demos are not canon, they're examples: `janus.js` is the only thing that can end up in people's projects, and even then it's a sample implementation of the Janus API, and there are several (often better) JS stacks out there.

davel: janus.js refers to jQuery inside `Janus.useOldDependencies`.

html/janus.js
davel: Yes, `console.log` should always be defined (so we could potentially remove the entire block). https://developer.mozilla.org/en-US/docs/Web/API/Console/log#browser_compatibility

lminiero: Again, this is not only used in browsers though.

.eslintrc.js
Frenzie: I completely understand if you have no interest in doing that yourself and such a follow-up may never even happen, but I find it surprising that you would actively object if someone's willing to put in the effort.

lminiero: What I meant is that I wouldn't want this in the project, so no point in anyone putting the effort. Whatever people contribute and gets merged, becomes something we need to maintain. This linting on janus.js will already be quite a burden on me (I'm not a JS developer), as any change I make I'll need to make sure it fits: while I understand why it may be needed for janus.js (which will be used in several projects) I see no point in doing the same for demos, where I want to be much more free to experiment and be "sloppy" since they're meant to be dumb example, and people should never base their products on those anyway.

html/janus.js
atoppi: @davel I think you should revert this change as it is not related to the linter at all

html/janus.js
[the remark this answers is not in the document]
davel: I have been through all the demos and can confirm that they still (at least superficially) work in Firefox and do not log any exceptions. I have also added another build step to make npm produce a CommonJS file so as to confirm that works. Our application also continues to work.

I have not spotted the code doing anything forbidden by `use strict`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

html/janus.js
davel: The assignment to `console` on line 234 is picked up as an error. Would you prefer me to remove the entire block, it should never be executed?

The `console` object should also exist as a global variable in Node.js too, https://nodejs.org/api/console.html

atoppi: I see thanks for clarifying.
I guess we can leave as it is in the PR then.

.eslintrc.js
atoppi: I guess this is not the right place to discuss the linting of js plugins.
Setting jquery globals in the linter does not sound like a big issue to me, so I'm happy to accept it.

.github/workflows/janus-ci.yml
```diff
@@ -152,3 +152,24 @@ jobs:
...
+    - name: Run ESLint
+      run: node_modules/.bin/eslint html/janus.js
+  javascript-rollup:
```
atoppi, line 163: imho I don't get the rationale of this.
Could you please elaborate ?

html/janus.js
atoppi: @davel so will you change this or not ?

davel: @atoppi I think I already have, you may need to refresh? I changed it into a closure as mentioned.

.github/workflows/janus-ci.yml
davel: It was asked whether the code would still build into a module. This confirms that rollup process still completes (although does not check that the resulting file works.)

html/janus.js
atoppi: you're right I'm looking at the outdated code, sorry!

.github/workflows/janus-ci.yml
atoppi: I totally forgot about the `npm` folder we have in the repo, my bad.
Anyway now that I have noticed it, I guess we should also:
- enable strict mode [in the rollup config](https://github.com/meetecho/janus-gateway/blob/master/npm/rollup.config.js#L8) 
- not sure, but maybe we can now [enable the linter again](https://github.com/meetecho/janus-gateway/blob/master/npm/module.js#L1) in the generated module ?

davel: Strict mode is now enabled (by removing the config disabling it.) We end up with `use strict` twice in the CJS file now, but I believe that is harmless.

The generated module has this comment, ` * Since this counts as 'autogenerated' code, ESLint is instructed to ignore the contents of this file when linting your project.` which I guess means you're not expected to re-lint rolled-up modules.

atoppi: That comment is already there before the module generation.

atoppi: I'm not sure what is the best practice here honestly

atoppi: > We end up with use strict twice in the CJS file now, but I believe that is harmless
Oh right... sorry for the noise can you revert this, I guess we can proceed with the merge then.

davel: I've reverted the change to `config.rollup.js`.

.github/workflows/janus-ci.yml
```diff
@@ -152,3 +152,24 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Rollup
+      run: |
+        cd npm
+        npm install
+        npm run rollup -- --o janus.cjs.js --f cjs
+    - uses: actions/upload-artifact@v2
```
atoppi, line 172: sorry one more question, why are you uploading the module ?

davel: It's so I could see the output to check it was sensible. GitHub will delete it automatically.

atoppi:
All right, LGTM :+1:

lminiero:
I'm not knowledgeable enough on the matter to have an opinion, so I'll trust @atoppi's review and notes. I do know that I'll start insulting you all if this translates to a ton of conflicts to resolve on the multistream branch, though :rofl:

atoppi:
_merging..._
image

PR #2773, opened 2021-09-20
add the token used to join the room to the info sent to event handlers

moezF:
all in  the title, use cases can be similar to the one mentioned in [this google grp dicussion](https://groups.google.com/g/meetecho-janus/c/9maXG4mjNdw)

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -6366,15 +6366,16 @@ static void *janus_videoroom_handler(void *data) {
 				/* Also notify event handlers */
 				if(notify_events && gateway->events_is_enabled()) {
 					json_t *info = json_object();
-					json_t *token = json_object_get(root, "token");
 					json_object_set_new(info, "event", json_string("joined"));
 					json_object_set_new(info, "room", string_ids ? json_string(publisher->room->room_id_str) :
 					json_integer(publisher->room->room_id));
 					json_object_set_new(info, "id", string_ids ? json_string(user_id_str) : json_integer(user_id));
 					json_object_set_new(info, "private_id", json_integer(publisher->pvt_id));
-                    if (publisher->room->check_allowed)
-    					json_object_set_new(info, "token", token);
-					if(display_text != NULL)
+                    if(publisher->room->check_allowed) {
+                        const char *token = json_string_value(json_object_get(root, "token"));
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @moezF! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

moezF:
@lminiero when there is no token it just sends the other info fields and omits the token : 
Screenshot from 2021-09-21 16-10-28

you are right, when a room doesn't require a token and the user did send a token it will be transferred to the event handler even if it's not required, i though there is no harm with that (to be honest, when implementing i was only thinking of my use case :see_no_evil: ) but if you think it might cause problems i can add the necessary checks to send it only when it's required :wink:

moezF:
@lminiero i've just updated the change to check whether the room requires a token before adding to info

plugins/janus_videoroom.c
```diff
@@ -6366,11 +6366,14 @@ static void *janus_videoroom_handler(void *data) {
...
 				if(notify_events && gateway->events_is_enabled()) {
 					json_t *info = json_object();
+					json_t *token = json_object_get(root, "token");
 					json_object_set_new(info, "event", json_string("joined"));
 					json_object_set_new(info, "room", string_ids ? json_string(publisher->room->room_id_str) :
 					json_integer(publisher->room->room_id));
 					json_object_set_new(info, "id", string_ids ? json_string(user_id_str) : json_integer(user_id));
 					json_object_set_new(info, "private_id", json_integer(publisher->pvt_id));
+                    if (publisher->room->check_allowed)
+    					json_object_set_new(info, "token", token);
```
lminiero, line 6376: Code style: no space between `if` and bracket.

I think something may go wrong with the Jansson refcount mechanism the way you're taking the object from one and adding it to another. Better to read the constant string and set that. On a related note, I'd only read `token` if the check succeeds, so something like:

```
if(publisher->room->check_allowed) {
	const char *token = json_string_value(json_object_get(root, "token"));
	json_object_set_new(info, "token", json_string(token));
}
```

moezF: thx for having a look into this, i just updated the pr

lminiero:
Thanks, looks good now, merging! :+1:

PR #2774, opened 2021-09-21
Add API to force Janus to use TURN

lminiero:
In case you configured Janus to use a TURN server (which you almost always [will NOT](https://github.com/meetecho/janus-gateway/pull/new/force-relay) want to do!), you may also sometimes want to actually force Janus to use TURN: this is what this patch allows you to do, and can be considered the Janus equivalent of the `iceTransportPolicy:"relay"` you can use in browsers, to force them to use TURN instead. Enabling it is very easy, since all you need to do is add a `force_relay:true` attribute to the JSEP object you send to Janus, whether it's an offer or answer.

Notice this only works for libnice >= 0.1.14, as that's the version the [property we use](https://libnice.freedesktop.org/libnice/NiceAgent.html#NiceAgent--force-relay) was added to. Besides, notice that this WILL cause PeerConnections to fail if: a. you didn't set any TURN server in Janus, b. Janus couldn't get a relay candidate from that server, or c. that candidate can't talk to the peer for any reason.

**VERY IMPORTANT**: Notice that you should **NOT** use this if you want **USERS** connected to Janus to use TURN. For **USERS**, you use `iceTransportPolicy:"relay"` on the client side: Janus will automatically talk to the TURN server, and does **NOT** need to be configured with TURN support itself. Janus **ONLY** needs TURN support in very specific cases, like when Janus itself is deployed behind a restrictive firewall (e.g., Janus acting as an endpoint in your home), and that's the **ONLY** time you'll need to actually be aware ths new feature exists.

I wrote this in all caps multiple times as it is a VERY common thing people get wrong all the time (main reason why we have a FAQ item for that), so I pretty much felt I had to repeat it all over again :stuck_out_tongue_winking_eye:

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -304,14 +304,19 @@ nat: {
 	# You can configure a TURN server in two different ways: specifying a
 	# statically configured TURN server, and thus provide the address of the
 	# TURN server, the transport (udp/tcp/tls) to use, and a set of valid
-	# credentials to authenticate...
+	# credentials to authenticate. Notice that you should NEVER configure
+	# a TURN server for Janus unless it's really what you want! If you want
+	# *users* to use TURN, then you need to configure that on the client
+	# side, and NOT in Janus. The following TURN configuration should ONLY
+	# be enabled when Janus itself is sitting behind a restrictive firewall
+	# (e.g., it's part of a service installed on a box in a private home).
 	#turn_server = "myturnserver.com"
 	#turn_port = 3478
 	#turn_type = "udp"
 	#turn_user = "myuser"
 	#turn_pwd = "mypassword"
 
-	# ... or you can make use of the TURN REST API to get info on one or more
+	# You can also make use of the TURN REST API to get info on one or more
 	# TURN services dynamically. This makes use of the proposed standard of
 	# such an API (https://tools.ietf.org/html/draft-uberti-behave-turn-rest-00)
 	# which is currently available in both rfc5766-turn-server and coturn.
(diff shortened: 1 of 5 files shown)


lminiero:
Just added a config property to allow/prevent this new functionality to be used in the Janus API: the default is `false`, meaning that trying to pass `force_relay` will be ignored and a warning be showed in the logs. If you want to use this feature, allow it globally in `janus.jcfg` first. A warning is also now printed in the logs at startup when the functionality is allowed, as it may not be something they want.

lminiero:
Thanks for the feedback! I'll merge then :+1:

B-R-Bender:
@lminiero @danjenkins  hi guys, can anyone please tell me how to configure janus js in such case?
I want to try to deploy janus itself behind a firewall
and I'm confused a bit
what should I put into server option as server obviously has no public ip or domain name
`const janus = new Janus({server: "...?"});`
thank you

lminiero:
It's explained in the first post.

B-R-Bender:
I'm sorry Lorenzo, but I can't get it.
I've got my local network, with janus deployed. I've also got coturn installed on a computer with public ip
I've configure janus to use TURN as described [here](https://github.com/meetecho/janus-gateway/blob/878555abf751a4306f6dd1aa6dae786be162ad13/conf/janus.jcfg.sample.in#L306)
but I'm really can't understand what to set as a janus server @lminiero

oej:
If you need further support please use the mailing list where many users can help you. Lorenzo is trying to be on holiday...

B-R-Bender:
yeah, got it
thanks guys anyway and happy new year!

PR #2776, opened 2021-09-24
Use prepend with reverse list in the sdp parsing (Credit to OSS-Fuzz)

atoppi:
This PR comes after an oss-fuzz timeout report.
When parsing the SDP janus makes large use of `g_list_append`. Unfortunately this specific implementation needs to traverse the entire list for every append (ref. [glib sources](https://github.com/GNOME/glib/blob/GLIB_2_16_6/glib/glist.c#L94)).
As suggested by glib docs, we replaced the _append_ with _prepend_ + _reverse_ operation in tail (so the list gets traversed only once).
The timeout is due to SDPs containing a lot of m-lines or attributes.


lminiero:
LGTM :+1:

PR #2777, opened 2021-09-26
Extending ACL feature for the videoroom plugin

AriartWt:
Extending ACL feature for the videoroom plugin

The purpose of this PR is to extend ACL features of the videoroom plugin to be able to grant special permissions to each token as, for now, the current implementation give any handle that provide a valid token the possibility to publish or subscribe without any way to restrict it.

For now, the extended ACL allow for two permissions : `publish` and `subscribe`. I plan to add more (like `publish/recieve_audio`, `publish/recieve_video` for example)

The current PR introduce no breaking change in the way the current ACL works.

# Why extend default ACL ?

In some situations, you may want to distinguish room users between those who can publish and those who can subscribe to only watch actual publishers. 

With the current ACL, it's not possible : the only solution implies to use rtp_forwarding to a streaming mountpoint. In some situations (especially with a variable number of participants in a room), this solution can add significant overhead to the developers, since we must maintain a state or a connection between the Janus server, our own UI and our backends. In scenarios where you would have a stateless and distributed architecture (multiple NodeJS instance on one or several servers for example), it becomes very difficult to enforce ACL policies on videoroom subscribers and to take care of all streaming mountpoints. Managing the desynchronization between janus and our infrastructure can be even more challenging.

With the extended ACL provided by this PR, any actor that can manage a room can now provide a token to a room user that will not allow him to publish and/or subscribe any feed if the token have restricted permissions. This scenario allow us, for example, to design a WebComponents that will be compliant with any VideoRoom in any scenario, since we can work in all cases with the same Janus API.

# Summary

1. [Permissions](#permissions)
2. [`max_publishers` vs `max_subscribers`](#acl-mpublishers-m-subscribers)
3. [ACL configuration persistence](#acl-conf-persistence)
4. [Impacts on Janus/videoroom APIs](#impact-on-apis)
   1. [The `join` request](#join-request)
   2. [The `joinandconfigure` request](#joinandconfigure-request)
   3. [The `allowed` request](#allowed-request)
   4. [The `create` request](#create-request)
   5. [The `edit` request](#edit-request)
5. [Implementation choices](#implementation-choices)
   1. [Storing the `join` token in the participants struct](#join-struct) 
   2. [Usage of bit flags for ACL permissions](#bit-flags)
   3. [Closing `PeerConnection` when a `consifugre` request is rejected](#closing-pc)
   4. [Creation of function `janus_videoroom_write_videoroom_config`](#write-config)
   5. [Misc](#misc)
6. [Documentation update](#doc-update) 

# Solution

<div id="permissions"></div>

## Permissions

The solution I choosed to implements will assign permissions to tokens. For now, there are only two :

- `publish`
- `subscribe`

Obviously, permissions can be compound to fit any ACL scenario with the videoroom plugin.

Adding permissions and/or revoking them is pretty straight forward and will be exposed in the section [Impact on Janus/videoroom APIs](#impact-on-apis)

<div id="acl-mpublishers-m-subscribers"></div>

## max_publishers & max_subscribers

To follow this new distinction between room user's permissions, we may need to enforce some maximum number of participants that only have a `subscribe` permission through their token, and since we may want to have a different maximum for token that can `publish`, I introduced a new room configuration : `subscribers`.

- `publishers` now only count for participants that have the `publish` permission.
- `subscribers` now only count for participants that have `subscribe` permission but **NOT** `publish` permission. A handle that has both `subscribe` and `publish` permissions is counted as a publisher.

Please note that if ACL is disabled, only the `publishers` will be accurate since we won't be able to distinguish any permission for any subscriber. 

So, if we want a room to have 4 publishers and 25 viewers, we must set the room as : 

```js
{
    publishers : 4,
    subscribers : 100 //4 feeds to subscribe to by viewer
}
```

To not setting any limit to the number of subscribers, you can either set it to `-1` or not set it at all.

<div id="acl-conf-persistence"></div>

## ACL configuration persistence

Since we want to be able to persist the ACL state if rooms are permanent (always to avoid having to maintain a state/ a connection to a Janus instance to detect an eventual reboot),
some optional properties are now recognized in the videoroom room config file: 

```
 {
    check_allowed = "no"; # ACL disabled by default. Can be enabled with an 'allowed' request with action 'enable'
    check_allowed_to_subscribe = "no"; # ACL disabled to subscribe by default. Can be enabled with an 'allowed' request with action 'enable_to_subscribe'
    subscribers = -1; #by default, no limitation. Can be set with an 'edit' request, setting the property 'new_subscribers' of the room, or providing 'subscribers' at room creation time.
    allowed = []; # strings, unrestricted tokens allowed to publish and subscribe
    allowed_subscribers = []; # strings, tokens only allowed to subscribe
    allowed_publishers = []; # strings, tokens only allowed to publish
 }
```

<div id="impact-on-apis"></div>

## Impact on Janus/videoroom APIs

Obviously, the Janus core API is not impacted in any way.

The Janus Javascript API is not impacted in any way too : users that wants to join the room just have to provide a token, as for the current videoroom ACL.

The only API changes are for the `videoroom` plugin itself.

<div id="join-request"></div>

### The `join` request

The `join` request API doesn't change and it's behavior is **not** modified : If ACL is enabled, a `join` request without token or with a wrong token will be rejected (as in the current ACL room).

The reason why there is a section about this there is because if you `join` with a token that have only the `subscribe` permission, the `join` request will **not** be rejected. 

The rejection will only occur if you send a `configure` request to try to `publish` a feed, so I wanted to address this point first.

<div id="joinandconfigure-request"></div>

### The `joinandconfigure` request

Regarding the `joinandconfigure` request, its API doesn't change **BUT** it's behavior is lightly modified. If ACL is enabled, a `joinandconfigure` request without token or with a wrong token will be rejected (as in the current ACL room).

But if you `joinandconfigure` with a token that has only a `subscribe` permission, the `joinandconfigure` request will be rejected.

For those requests, I had two choices with the new ACL system : 

1. Rejecting the request if ACL permissions doesn't allow us to `publish`
2. Letting them join but ignoring the configure part of the settings they added through the query.

For now, I chose to reject, since it can be a little confusing for a user to not get any feedback on why the `configure` part was ignored while the request succeed. It's a point that may be discussed.

<div id="allowed-request"></div>

### The `allowed` request

The videoroom `allowed` request has two new optional properties `allowed_publishers` and `allowed_subscribers`, similar to `allowed` : 

```js
{
    "request" : "allowed",
    "secret" : "<room secret, mandatory if configured>",
    "room" : "<unique numeric ID of the room to update>",
    "action" : "enable|disable|add|remove|enable_to_subscribe|disable_to_subscribe",
    "allowed" : [ /* string[] : tokens allowed to publish and subscribe */ ],
    "allowed_publishers" : [ /* string[] : tokens allowed to publish */ ],
    "allowed_subscribers" : [ /* string[] : tokens allowed to subscribe */ ]
}
```

Please note that two actions have been added to this request API : `enable_to_subscribe` and `disable_to_subscribe`. They allow us to enable/disable mandatory ACL to **subscribe** to available feeds.

It's disabled by default and must be enabled explicitly, either by sending and `allowed` request, by editing the videoroom config file by adding `check_allowed_to_subscribe = "yes"` or by creating the room with the setting `check_allowed_to_subscribe` to `true`.

<div id="create-request"></div>

### The `create` request

The videoroom `create` request has 6 new optional parameters to be able to configure ACL at room creation time : 

```js
{
    check_allowed: false; //boolean
    check_allowed_to_subscribe = false; //boolean
    subscribers = -1; // Integer; Negative means no limit.
    allowed = []; // string[], tokens allowed to publish and subscribe
    allowed_subscribers = []; // string[], tokens allowed to subscribe
    allowed_publishers = []; // string[], tokens allowed to publish
}
```

<div id="edit-request"></div>

### The `edit` request

The videoroom `edit` request receive 1 new optional parameter to be able to update the max_subscribers of a room with an edit request (as for publishers) :

```js
{
    "new_subscribers" : -1 //Integer
}
```

<div id="implementation-choices"></div>

## Implementation choices

In this section I'll try to highlight some decisions I took while extending the ACL and that may be subject to discussions.

<div id="join-struct"></div>

### Storing the `join` token in the participants’ struct

Since we want to give permissions to room users, we store the provided token with `join` request in `janus_videoroom_publisher` and `janus_videoroom_subscriber` structs to be able to check the handle permissions each time the handle try to publish/subscribe to one or more feeds.

It have been done this way because it offers more flexibility than storing the ACL result, since the token may be get granted new permissions (or revoked) later.

The second reason it have been done like this is because I would like to add a way for the videoroom plugin to close all `PeerConnection` that is related to a revoked token or a revoked permission in a near future. I think it could be a good idea to have an option to enable this behavior from the room config (maybe something like `acl_auto_close_peer_connections`).

<div id="bit-flags"></div>

### Usage of bit flags for ACL permissions

I know that it could be considered as bad practice to use bit flags. But it seems the simplest (and the more flexible) way to check and compose ACL permissions. Since I foresee to add some other permissions like `publish/recieve_audio`, `publish/recieve_video` for example, it may be a good system to allow for more features later. 

<div id="closing-pc"></div>

### Closing `PeerConnection` when a `configure` request is rejected

When I tried to reject a `configure` request for a handle that doesn't have the `publish` permission, the `PeerConnection` initiated with the `configure` request always ended in an inconsistent state. 

All further `configure` request ended up in a `490` error : `Error setting ICE localy` that was unrecoverable. I had to drop the handle each times it occured, I'm unsure why. The only solution I have been able to come up with was to close the PeerConnection before `goto error;`: 

```c
janus_videoroom_hangup_media(session->handle);
```

It can make sense since the videoroom plugin respond with a `512` error : `Unauthorized, permission denied to publish` but since I'm not very sure about this, I prefer to address it there. 

<div id="write-config"></div>

### janus_videoroom_write_videoroom_config

Since I needed to write room config after an `allowed` request on permanent rooms, I refactorised the write config routine in a `janus_videoroom_write_videoroom_config` function.

It may or may not be a good idea. I'm not familiar enough with the language nor the codebase to be sure about this, so I prefer to signal it too.

<div id="misc"></div>

### Misc

I tried my best to follow the code style and naming conventions, but I'm often too verbose on functions naming. Feel free to address this too.

**Note that I'm far to be a competent C developer. So please, as you review, pay extra attention to the resources freeing aspect of what I've done. I'm afraid to introduce a memory leak by mistake.**

<div id="doc-update"></div>

## Documentation update

I didn't updated documentation-related stuff yet. I will if the PR is on the road of acceptation and, if it is, once the whole thing would be reviewed/corrected/changed.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1280,7 +1280,13 @@ static struct janus_json_parameter create_parameters[] = {
 	{"lock_record", JANUS_JSON_BOOL, 0},
 	{"permanent", JANUS_JSON_BOOL, 0},
 	{"notify_joining", JANUS_JSON_BOOL, 0},
-	{"require_e2ee", JANUS_JSON_BOOL, 0}
+	{"require_e2ee", JANUS_JSON_BOOL, 0},
+	{"allowed", JSON_ARRAY, 0},
+	{"allowed_subscribers", JSON_ARRAY, 0},
+	{"allowed_publishers", JSON_ARRAY, 0},
+	{"subscribers", JSON_INTEGER, 0},
+	{"check_allowed", JANUS_JSON_BOOL, 0},
+	{"check_allowed_to_subscribe", JANUS_JSON_BOOL, 0},
 };
 static struct janus_json_parameter edit_parameters[] = {
 	{"secret", JSON_STRING, 0},
@@ -1292,6 +1298,7 @@ static struct janus_json_parameter edit_parameters[] = {
 	{"new_bitrate", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"new_fir_freq", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"new_publishers", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
+	{"new_subscribers", JSON_INTEGER, 0},
 	{"permanent", JANUS_JSON_BOOL, 0}
 };
 static struct janus_json_parameter room_parameters[] = {
@@ -1336,7 +1343,9 @@ static struct janus_json_parameter destroy_parameters[] = {
 static struct janus_json_parameter allowed_parameters[] = {
 	{"secret", JSON_STRING, 0},
 	{"action", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
-	{"allowed", JSON_ARRAY, 0}
+	{"allowed", JSON_ARRAY, 0},
+	{"allowed_subscribers", JSON_ARRAY, 0},
+	{"allowed_publishers", JSON_ARRAY, 0}
 };
 static struct janus_json_parameter kick_parameters[] = {
 	{"secret", JSON_STRING, 0}
@@ -1505,11 +1514,14 @@ typedef struct janus_videoroom {
 	GHashTable *participants;	/* Map of potential publishers (we get subscribers from them) */
 	GHashTable *private_ids;	/* Map of existing private IDs */
 	volatile gint destroyed;	/* Whether this room has been destroyed */
-	gboolean check_allowed;		/* Whether to check tokens when participants join (see below) */
-	GHashTable *allowed;		/* Map of participants (as tokens) allowed to join */
 	gboolean notify_joining;	/* Whether an event is sent to notify all participants if a new participant joins the room */
 	janus_mutex mutex;			/* Mutex to lock this room instance */
 	janus_refcount ref;			/* Reference counter for this room */
+	gboolean permanent;         /* Whether this room is permanent or not */
+	int max_subscribers;        /* Maximum number of publishers marked as subscriber_only */
+	gboolean check_allowed;		/* Whether to check tokens when participants join (see below) */
+	gboolean check_allowed_to_subscribe; /* Wether to check token when a subscriber try to join */
+	GHashTable *allowed;		/* Map of participants (as tokens) allowed to join */
 } janus_videoroom;
 static GHashTable *rooms;
 static janus_mutex rooms_mutex = JANUS_MUTEX_INITIALIZER;
@@ -1674,6 +1686,7 @@ typedef struct janus_videoroom_publisher {
 	gboolean e2ee;		/* If media from this publisher is end-to-end encrypted */
 	volatile gint destroyed;
 	janus_refcount ref;
+	char *token;    /* The token used by this publisher to join if any */
 } janus_videoroom_publisher;
 static guint32 janus_videoroom_rtp_forwarder_add_helper(janus_videoroom_publisher *p,
 	const gchar *host, int port, int rtcp_port, int pt, uint32_t ssrc,
@@ -1705,6 +1718,7 @@ typedef struct janus_videoroom_subscriber {
 	gboolean e2ee;		/* If media for this subscriber is end-to-end encrypted */
 	volatile gint destroyed;
 	janus_refcount ref;
+	char *token;        /* The token this publisher used to subscribe if any */
 } janus_videoroom_subscriber;
 
 typedef struct janus_videoroom_rtp_relay_packet {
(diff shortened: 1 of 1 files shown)


lminiero:
Closing as Janus has changed considerably since this was first opened, and there didn't seem to be much interest in the feature. We can consider revisiting this again in the future. Thanks for your contribution anyway!

PR #2782, opened 2021-09-29
Ensure we access room uniformly and safely across all the VideoRoom code

shripchenko:
Hi everyone.

Could you please review my attempt to correct accessing room everywhere in VideoRoom code to:
1. under rooms_mutex
2. before unlocking rooms_mutex -> always incref(room)
3. lock room->mutex after unlocking rooms_mutex (if it was intended previously)

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -2727,11 +2727,13 @@ static void janus_videoroom_leave_or_unpublish(janus_videoroom_publisher *partic
 		janus_mutex_unlock(&rooms_mutex);
 		return;
 	}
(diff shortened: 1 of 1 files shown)


lminiero:
The changes seem reasonable to me, and in line with what we do in most of the rest of the code (increasing a reference before releasing the mutex that protects something). I'll do a better review later, and in case merge :+1:

lminiero:

lminiero:
I think we have the same approach in other plugins (AudioBridge, TextRoom?), so I'll have to check if I have to apply your fixes to those too.

PR #2784, opened 2021-09-30
ifdef around optional defines (build process related / not necessarily defined in local dev)

JanFellner:
There are some defines which are defined in the build process… but not necessarily in local dev. So we should only use them if they are defined cause the config defines them in the end.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -4128,8 +4128,10 @@ gint main(int argc, char *argv[])
 	}
 	if(args_info.configs_folder_given) {
 		configs_folder = g_strdup(args_info.configs_folder_arg);
+#ifdef CONFDIR
 	} else {
 		configs_folder = g_strdup(CONFDIR);
+#endif
 	}
 	if(config_file == NULL) {
 		char file[255];
@@ -4304,7 +4306,9 @@ gint main(int argc, char *argv[])
 	DIR *dir = NULL;
 	/* External loggers are usually disabled by default: they need to be enabled in the configuration */
 	gchar **disabled_loggers = NULL;
+#ifdef LOGGERDIR
 	path = LOGGERDIR;
+#endif
 	item = janus_config_get(config, config_general, janus_config_type_item, "loggers_folder");
 	if(item && item->value)
 		path = (char *)item->value;
@@ -5171,7 +5175,9 @@ gint main(int argc, char *argv[])
 		JANUS_LOG(LOG_INFO, "Event handlers support disabled\n");
 	} else {
 		gchar **disabled_eventhandlers = NULL;
+#ifdef EVENTDIR
 		path = EVENTDIR;
+#endif
 		item = janus_config_get(config, config_general, janus_config_type_item, "events_folder");
 		if(item && item->value)
 			path = (char *)item->value;
@@ -5312,7 +5318,9 @@ gint main(int argc, char *argv[])
 	}
 
 	/* Load plugins */
+#ifdef PLUGINDIR
 	path = PLUGINDIR;
+#endif
 	item = janus_config_get(config, config_general, janus_config_type_item, "plugins_folder");
 	if(item && item->value)
 		path = (char *)item->value;
@@ -5431,7 +5439,9 @@ gint main(int argc, char *argv[])
 
 	/* Load transports */
 	gboolean janus_api_enabled = FALSE, admin_api_enabled = FALSE;
+#ifdef TRANSPORTDIR
 	path = TRANSPORTDIR;
+#endif
 	item = janus_config_get(config, config_general, janus_config_type_item, "transports_folder");
 	if(item && item->value)
 		path = (char *)item->value;


lminiero:
Is there a specific use case for this? There's currently no way to build Janus without going through the configure, so those variables should always be defined. Or am I missing something?

atoppi:
By adding the `ifdef` we are basically suggesting to the IDE that those parts are optional and the IDE could skip syntax checking for "unreachable" code.
In VSC if I change the code to:
```
#ifdef CONFDIR
   configs_folder = g_strdup(CONFDIR);
   blah blah broken syntax
#endif
```
No syntax error is being detected.

My solution has always been to configure my dev environment to include the macros that will be added by the build chain.
For instance, this is my `c_cpp_properties.json` in VSC:
```
{
    "configurations": [
        {
            "name": "Linux",
            "includePath": [
                "${default}",
                "/usr/include/sofia-sip-1.12"
            ],
            "defines": [
                "CONFDIR=\"\"",
                "EVENTDIR=\"\"",
                "HAVE_CLOSE_ASYNC=1",
                "HAVE_LIBCURL=1",
                "HAVE_LIBOGG=1",
                "HAVE_LIBOPUS=1",
                "HAVE_PORTRANGE=1",
                "HAVE_SCTP=1",
                "HAVE_SRTP_2=1",
                "LOGGERDIR=\"\"",
                "PLUGINDIR=\"\"",
                "TRANSPORTDIR=\"\""
            ]
        }
    ],
    "version": 4
}
```

JanFellner:
It´s exactly about syntax checking as the IDE complains about elements that are just optional default values but overwritten by the config at startup and the build sets them anyway. So there is no need to configure them at all for local dev.

OT:
(Also i still think that way more people would work on janus if the startup hurdle would be lower, so if you would provide for e.g. vscode the config one can use to just start right away..., can also be any other ide but that would get more people inwolved. Getting it up and running took me a couple of days...)

lminiero:
But if IDEs are the problem (they aren't for me, I'm happy with Geany), then Alessandro's answer is the solution, not adding artificial ifdefs to the code IMHO (which as Alessandro pointed out would cause the IDE to incorrectly miss some checks).

JanFellner:
I am happy with adding them to the config.

(I just emphasize that having an IDE will get you more people into the project cause the initial setup is way easier...)

Please close it...

lminiero:
We don't plan to add any IDE configuration to the project, as there's too many IDEs out there and I'd rather not start chasing changes there either... If you think a page in the documentation with a sample config like the one @atoppi suggested makes sense, we can definitely do that instead. Closing :+1:

JanFellner:
@atoppi if you add your config to the doc that would really help.
You know that i have been asking about that initial setup and how to get things going like half a year ago in the google groups and that really helped me a lot. Don´t know why the directories got lost or were missing there... my config currently looks like this:

`{
    "env": {
      "myDefaultIncludePath": [
        "${workspaceFolder}",
        "/usr/include/glib-2.0/",
        "/usr/lib/x86_64-linux-gnu/glib-2.0/include",
        "/usr/include/nice",
        "/usr/include/srtp2",
        "/usr/include/x86_64-linux-gnu/"
      ],
      "myCompilerPath": "/usr/bin/gcc"
    },
    "configurations": [
      {
        "name": "J",
        "intelliSenseMode": "gcc-x64",
        "includePath": ["${myDefaultIncludePath}"],
        "defines": [
          "HAS_DTLS_WINDOW_SIZE=1",
          "HAVE_SRTP_AESGCM=1",
          "HAVE_BORINGSSL=1",
          "HAVE_DTLS_SETTIMEOUT=1",
          "HAVE_PORTRANGE=1",
          "HAVE_LIBNICE_TCP=1",
          "HAVE_CLOSE_ASYNC=1",
          "HAVE_ICE_NOMINATION=1",
          "HAVE_SRTP_2=1",
          "HAVE_LIBCURL=1",
          "HAVE_TURNRESTAPI=1",
          "HAVE_WEBSOCKETS=1",
          "HAVE_LIBWEBSOCKETS_PEER_SIMPLE=1",
          "HAVE_WSEVH=1",
          "HAVE_LIBOGG=1",
          "CONFDIR=\"\"",
          "EVENTDIR=\"\"",
          "LOGGERDIR=\"\"",
          "PLUGINDIR=\"\"",
          "TRANSPORTDIR=\"\"",
          "_GNU_SOURCE"
        ],
        "compilerPath": "${myCompilerPath}",
        "cStandard": "c17",
        "cppStandard": "c++20",
        "browse": {
          "path": ["${workspaceFolder}"],
          "limitSymbolsToIncludedHeaders": true,
          "databaseFilename": ""
        }
      }
    ],
    "version": 4
  }
`

And thats something you will definelty not want to write from scratch if you start with the project...

atoppi:
Adding here the whole configuration (user + workspace + folder) for future reference (we'll soon add this to the docs).
This is valid for Ubuntu 20.04.
```
{
    "configurations": [
        {
            "name": "Linux",
            "compilerPath": "/usr/bin/gcc",
            "cStandard": "gnu11",
            "cppStandard": "gnu++17",
            "includePath": [
                "${workspaceFolder}/**",
                "/usr/lib/x86_64-linux-gnu/glib-2.0/include",
                "/usr/include/glib-2.0",
                "/usr/include/nice",
                "/usr/include/sofia-sip-1.12"
            ],
            "intelliSenseMode": "linux-gcc-x64",
            "systemIncludePath": [
                "/usr/lib/gcc/x86_64-linux-gnu/9/include",
                "/usr/local/include",
                "/usr/include/x86_64-linux-gnu",
                "/usr/include"
            ],
            "defines": [
                "CONFDIR=\"\"",
                "EVENTDIR=\"\"",
                "HAVE_CLOSE_ASYNC=1",
                "HAS_DTLS_WINDOW_SIZE=1",
                "HAVE_ICE_NOMINATION=1",
                "HAVE_LIBCURL=1",
                "HAVE_LIBNICE_TCP=1",
                "HAVE_LIBOGG=1",
                "HAVE_LIBOPUS=1",
                "HAVE_LIBWEBSOCKETS_PEER_SIMPLE=1",
                "HAVE_PORTRANGE=1",
                "HAVE_SCTP=1",
                "HAVE_SRTP_2=1",
                "LOGGERDIR=\"\"",
                "PLUGINDIR=\"\"",
                "TRANSPORTDIR=\"\""
            ]
        }
    ],
    "version": 4
}
```

PR #2786, opened 2021-10-02
plugins/janus_sip.c: MESSAGE Authentication and Deliver Status Report

thetechpanda:
CHANGE: sets MESSAGE Call-ID if not defined in the message event
ADD: SIP Authentication for 401 and 407 reponse to Message
ADD: messagedelivery event to report delivery status back to the peer

### DIFF
--- a/html/siptest.html
+++ b/html/siptest.html
@@ -108,8 +108,8 @@
 								<span class="input-group-addon"><i class="fa fa-phone fa-fw"></i></span>
 								<input class="form-control" type="text" placeholder="SIP URI to call (e.g., sip:1000@example.com)" autocomplete="off" id="peer" onkeypress="return checkEnter(this, event);" />
 							</div>
-							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video						</div>
-					</div>
+                            <button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video
+						</div>
 				</div>
 				<div id="videos" class="hide">
 					<div class="col-md-6">
--- a/html/siptest.html
+++ b/html/siptest.html
@@ -108,8 +108,9 @@
 								<span class="input-group-addon"><i class="fa fa-phone fa-fw"></i></span>
 								<input class="form-control" type="text" placeholder="SIP URI to call (e.g., sip:1000@example.com)" autocomplete="off" id="peer" onkeypress="return checkEnter(this, event);" />
 							</div>
-                            <button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video
+							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video
 						</div>
+					</div>
 				</div>
 				<div id="videos" class="hide">
 					<div class="col-md-6">
--- a/html/siptest.html
+++ b/html/siptest.html
@@ -108,10 +108,12 @@
 								<span class="input-group-addon"><i class="fa fa-phone fa-fw"></i></span>
 								<input class="form-control" type="text" placeholder="SIP URI to call (e.g., sip:1000@example.com)" autocomplete="off" id="peer" onkeypress="return checkEnter(this, event);" />
 							</div>
-							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video
+							<input class="form-control" type="text" placeholder="any text message" autocomplete="off" id="message" ></input>
+							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video<br />
+							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="sendmessage" onclick="return doMessage();" >Message</button>
 						</div>
 					</div>
-				<div/>
+				</div>
 				<div id="videos" class="hide">
 					<div class="col-md-6">
 						<div class="panel panel-default">
--- a/html/siptest.html
+++ b/html/siptest.html
@@ -108,10 +108,7 @@
 								<span class="input-group-addon"><i class="fa fa-phone fa-fw"></i></span>
 								<input class="form-control" type="text" placeholder="SIP URI to call (e.g., sip:1000@example.com)" autocomplete="off" id="peer" onkeypress="return checkEnter(this, event);" />
 							</div>
-							<input class="form-control" type="text" placeholder="any text message" autocomplete="off" id="message" ></input>
-							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video<br />
-							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="sendmessage" onclick="return doMessage();" >Message</button>
-						</div>
+							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video						</div>
 					</div>
 				</div>
 				<div id="videos" class="hide">
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -446,6 +446,16 @@ $(document).ready(function() {
 											$('#call').removeAttr('disabled').html('Call')
 												.removeClass("btn-danger").addClass("btn-success")
 												.unbind('click').click(doCall);
+										} else if(event === 'messagedelivery') {
+											// message delivery status
+											let reason = result["reason"];
+											let code = result["code"];
+											let callid = msg['call_id'];
+											if (code == 200) {
+												toastr.success(`${callid} Delivery Status: ${code} ${reason}`);
+											} else {
+												toastr.error(`${callid} Delivery Status: ${code} ${reason}`);
+											}
 										}
 									}
 								},
@@ -961,10 +971,12 @@ function addHelper(helperCreated) {
 		'					<span class="input-group-addon"><i class="fa fa-phone fa-fw"></i></span>' +
 		'					<input disabled class="form-control" type="text" placeholder="SIP URI to call (e.g., sip:1000@example.com)" autocomplete="off" id="peer' + helperId + '" onkeypress="return checkEnter(this, event, ' + helperId + ');"></input>' +
 		'				</div>' +
-		'				<button disabled class="btn btn-success margin-bottom-sm" autocomplete="off" id="call' + helperId + '">Call</button> <input autocomplete="off" id="dovideo' + helperId + '" type="checkbox">Use Video</input>' +
+		'				<input class="form-control" type="text" placeholder="any text message" autocomplete="off" id="message' + helperId + '" ></input>' +
+		'				<button disabled class="btn btn-success margin-bottom-sm" autocomplete="off" id="call' + helperId + '">Call</button> <input autocomplete="off" id="dovideo' + helperId + '" type="checkbox">Use Video</input><br />' +
+		'				<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="sendmessage' + helperId + '" onclick="return doMessage('+helperId+');">Message</button>' +
 		'			</div>' +
 		'		</div>' +
-		'	<div/>' +
+		'	</div>' +
 		'	<div id="videos' + helperId + '" class="hide">' +
 		'		<div class="col-md-6">' +
 		'			<div class="panel panel-default">' +
@@ -1313,6 +1325,16 @@ function addHelper(helperCreated) {
 						$('#call' + helperId).removeAttr('disabled').html('Call')
 							.removeClass("btn-danger").addClass("btn-success")
 							.unbind('click').click(doCall);
+					} else if(event === 'messagedelivery') {
+						// message delivery status
+						let reason = result["reason"];
+						let code = result["code"];
+						let callid = msg['call_id'];
+						if (code == 200) {
+							toastr.success(`${callid}/${helperId} Delivery Status: ${code} ${reason}`);
+						} else {
+							toastr.error(`${callid}/${helperId} Delivery Status: ${code} ${reason}`);
+						}
 					}
 				}
 			},
@@ -1518,4 +1540,34 @@ function removeHelper(helperId) {
 		// Remove the related UI too
 		$('#sipcall'+helperId).remove();
 	}
+}
+/**
+ * send message function
+ *
+ * @param {number} suffix helper id
+ */
+function doMessage(suffix) {
+	if (suffix === undefined) {
+		suffix = '';
+	}
+	$('#peer' + suffix).attr('disabled', true);
+	$('#message' + suffix).attr('disabled', true);
+	$('#sendmessage' + suffix).attr('disabled', true);
+	let to = $('#peer' + suffix).val();
+	let body = $('#message' + suffix).val();
+
+	if(to === "" || to.indexOf("sip:") != 0 || to.indexOf("@") < 0) {
+		bootbox.alert('Please insert a valid SIP address (e.g., sip:pluto@example.com)');
+	} else {
+		let handle = sipcall;
+		if(suffix && helpers[suffix] && helpers[suffix]) {
+			handle = helpers[suffix].sipcall;
+		}
+		var msg = { request: "message", uri: to, content: body };
+		handle.send({ message: msg });
+	}
+	$('#peer' + suffix).removeAttr('disabled').val("");
+	$('#message' + suffix).removeAttr('disabled').val("");
+	$('#sendmessage' + suffix).removeAttr('disabled');
+	return;
 }
--- a/html/siptest.js
+++ b/html/siptest.js
@@ -853,14 +853,14 @@ function doCall(ev) {
 		bootbox.alert('Please insert a valid SIP address (e.g., sip:pluto@example.com)');
 		$('#peer' + suffix).removeAttr('disabled');
 		$('#dovideo' + suffix).removeAttr('disabled');
-		$('#call' + suffix).removeAttr('disabled').click(function() { doCall(helperId); });
+		$('#call' + suffix).removeAttr('disabled').click(function(ev) { doCall(ev); });
 		return;
 	}
 	if(username.indexOf("sip:") != 0 || username.indexOf("@") < 0) {
 		bootbox.alert('Please insert a valid SIP address (e.g., sip:pluto@example.com)');
 		$('#peer' + suffix).removeAttr('disabled').val("");
 		$('#dovideo' + suffix).removeAttr('disabled').val("");
-		$('#call' + suffix).removeAttr('disabled').click(function() { doCall(helperId); });
+		$('#call' + suffix).removeAttr('disabled').click(function(ev) { doCall(ev); });
 		return;
 	}
 	// Call this URI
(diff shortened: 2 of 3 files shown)


plugins/janus_sip.c
```diff
@@ -5419,7 +5439,78 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
...
+					TAG_END());
+			}  else {
+				/* message response, notify the application */
+				json_t *result = json_object();
+				/* sip code and reason */
+				json_object_set_new(result, "event", json_string("messagedelivery"));
+				json_object_set_new(result, "code", json_integer(status));
+				json_object_set_new(result, "reason", json_string(phrase));
+				/* 200 OK, message delivered */
+				json_object_set_new(result, "delivered", (status == 200 ? json_true() : json_false()));
```
lminiero, line 5502: Is this property really needed? We're already sending the `code` back, so this info is redundant.

lminiero (commented):
Thanks! Added some notes inline.

plugins/janus_sip.c
```diff
@@ -5419,7 +5439,78 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
...
+				json_object_set_new(result, "call_id", json_string(sip->sip_call_id->i_id));
+				/* builds the delivery receipt */
+				json_t *dr = json_object();
+				json_object_set_new(dr, "sip", json_string("event"));
+				json_object_set_new(dr, "result", result);
+				json_object_set_new(dr, "call_id", json_string(sip->sip_call_id->i_id));
```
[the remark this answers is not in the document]
thetechpanda: https://datatracker.ietf.org/doc/html/rfc7329
   The SIP [RFC3261] Call-ID header value is a globally unique
   identifier, which is mandatory in all requests/responses and
   identifies SIP messages belonging to the same dialog or registration.
   It provides a portion of the SIP message dialog-matching criteria and
   is used in such things as "Replaces" headers [RFC3891] and dialog-
   event packages [RFC4235] for matching to dialogs, and in SIP Identity
   [RFC4474] and Connected Identity [RFC4916] as one of the inputs for
   signing.

I don't know about Sofia (not much experience with the library) but unless a result of a bug in the library exchanged SIP messages should always have a Call-ID. Do you still feel there should be a check?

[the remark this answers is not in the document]
lminiero: Actually I remembered the same doubts came out in #2664, so I think we indeed don't have to worry about that :+1:

plugins/janus_sip.c
```diff
@@ -4656,18 +4659,35 @@ static void *janus_sip_handler(void *data) {
...
+					SIPTAG_CALL_ID_STR(message_callid),
 					TAG_END());
 			}
 			/* Notify the operation */
 			result = json_object();
 			json_object_set_new(result, "event", json_string("messagesent"));
+			/* message_callid will be later used to return the delivery status of the message */
+			json_object_set_new(result, "call_id", json_string(message_callid));
+			if (session->callid == NULL) {
+				session->callid = g_strdup(message_callid);
```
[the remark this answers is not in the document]
thetechpanda: I see your point, in most implementation I've seen the SIP MESSAGEs are exchanged outside of a dialog even when a call is active between the peers. I will research a bit more but I think an hashtable for the SIP MESSAGEs may be unnecessary. What would be your preference?

plugins/janus_sip.c
thetechpanda: No is not, the reason I've put it is because I saw the event as an TCP/IP packet, the response is encapsulated in the result so that you don't need the whole frame. I've used in my client side implementation of the feature, I will remove the redundant call_id and the delivered flag and resubmit my code.

plugins/janus_sip.c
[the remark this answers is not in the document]
lminiero: Sorry, missed this message before.

One one end, I think you maybe don't need to track it at all, since when you get a response in `nua_r_message` you just read the Call-ID from the message you received, and pass that back as an event. On the other end, though, the way you're sending the event right now won't always work when using session helpers, as you may be sending a message from helper n.2, but the delivered event would be sent to the session master (the session owning the loop).

As such, there may indeed be a need to keep a hashtable in the session, or even just a list: this way, when you get a `nua_r_message` back, you start from the master session (the one the loop notifies about) and try to find the message Call-ID there; if it can't be found, then iterate on helper sessions until you find the Call-ID, and if you find it that's the session you send the event to.

Of course, if there's a master session and helper sessions, they all belong to the same application, so we may even just keep it simple and say that, whatever session sent the message, the delivered event will always be notified back to the master session. In that case, nothing else you need to do, except maybe add some text in the documentation to explain this and get rid of the

			if (session->callid == NULL) {
				session->callid = g_strdup(message_callid);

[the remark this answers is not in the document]
lminiero: > If my understanding is correct Helpers are used to offer concurrency for sip calls on the same account, something that IMHO is not needed for messages as messages live and die within the server response (success or error).

That depends. If I'm using helpers to have concurrent calls with you (master session), Alice (helper 1) and Bob (helper 2), sending an in-dialog MESSAGE to Alice will happen on a helper session, but the "delivered" event would be sent to the master session, which may be seen as incorrect from a signalling flow perspective. This can be problematic depending how the application is written, which is why I said that if we say this is ok, it needs to be documented, because it means that it's up to the application (e.g., the JavaScript code) to keep a global map of MESSAGE session IDs to know which call the acknowledgement actually refers to.

> Additionally, on a personal note, I am not too confident I can perform such a structural change as my skills have dropped a lot since I've last developed in C. C is not like riding a bicycle (ahah).

That's absolutely fine, we can indeed revisit this later (after it's merged) if it turns out to be a problem in practice.

plugins/janus_sip.c
lminiero: Ack, make sure you remove the `call_id` from inside `result`, as in other events we generate `call_id` is always at the root of the event.

lminiero:
@thetechpanda any update on this? I'm planning to do a new release soon, so I was wondering if we should wait for this too, or leave it for the next update. Thanks!

thetechpanda:
Hi Lorenzo,

I’ve done the changes, including the change so that the helpers will
receive the responses.

Unfortunately I don’t have a testing rig anymore, my SIP service provider
has disabled messages exchange in the latest major software release. I was
planning to run Kamailio so that I could check the change was working as
expected, but I am in a very busy schedule at work and leaves me little
time to spend on personal projects.

I don’t like to push code that I didn’t test, but if are ok with it, I
believe is ready.

Kind regards,

thetechpanda:
I pushed the changes they should be ok. let me know if you need me to fix anything else. cheers!

plugins/janus_sip.c
```diff
@@ -4656,18 +4663,34 @@ static void *janus_sip_handler(void *data) {
...
-			/* Notify the operation */
+			/* Notify the application */
 			result = json_object();
 			json_object_set_new(result, "event", json_string("messagesent"));
+			json_object_set_new(result, "call_id", json_string(message_callid));
+			/* Store message id and session */
+			janus_mutex_lock(&sessions_mutex);
+			g_hash_table_insert(messageids, message_callid, session);
+			janus_mutex_unlock(&sessions_mutex);
+			g_free(message_callid);
```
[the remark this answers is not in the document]
thetechpanda: converted to g_hash_table_new_full, I've used the original `janus_sip_session_destroy`, I've added a function to free the key as well `janus_sip_message_key_destroy`.

plugins/janus_sip.c
```diff
@@ -5419,7 +5442,88 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
...
+				nua_authenticate(nh,
+					NUTAG_AUTH(auth),
+					TAG_END());
+			}  else {
+				/* Find session associated with the message */
+				janus_mutex_lock(&sessions_mutex);
+				janus_sip_session *message_session = g_hash_table_lookup(messageids, sip->sip_call_id->i_id);
+				janus_mutex_unlock(&sessions_mutex);
+				if (!message_session) {
+					message_session = session
```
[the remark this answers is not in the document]
thetechpanda: I've tested and compiled the code, no missing semicolon this time.

plugins/janus_sip.c
```diff
@@ -5419,7 +5442,88 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
...
+					JANUS_LOG(LOG_VERB, "Message (%s) not associated with any session, event will be reported to master\n", sip->sip_call_id->i_id);
+				} else {
+					/* Session was found in messageids, removes from hashtable */
+					janus_mutex_lock(&sessions_mutex);
+					g_hash_table_remove(messageids, sip->sip_call_id->i_id);
```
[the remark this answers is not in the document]
thetechpanda: I've increased the refcount before inserting in the `messageids` hashtable, `g_hash_table_remove` should now correctly decrease the refcount and free the `messageid` key.

plugins/janus_sip.c
```diff
@@ -4656,18 +4663,34 @@ static void *janus_sip_handler(void *data) {
...
 					TAG_END());
 			}
-			/* Notify the operation */
+			/* Notify the application */
 			result = json_object();
 			json_object_set_new(result, "event", json_string("messagesent"));
+			json_object_set_new(result, "call_id", json_string(message_callid));
+			/* Store message id and session */
+			janus_mutex_lock(&sessions_mutex);
+			g_hash_table_insert(messageids, message_callid, session);
```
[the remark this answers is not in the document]
thetechpanda: I've increased the refcount then added to `messageids`. Assuming that the SIP server will always provide a response for the message we sent refcount will be decreased after the "messagedelivery" event.

thetechpanda:
Hello,

I've rebuilt the test bench, using kamailio. I've tested using http and websocket transports and done multiple tests sending and receiving messages. I've debugged as extensively as I can, I've not seen any unexpected behaviour and I've carefully reviewed the locks, hope this time the patch is worthy or your time.

I've also added some documentation at the top of the plugin to cater for the new behaviour: optional `call_id` for `message` requests and `messagedelivery` event documentation.

I've noticed something, sending messages via Helpers works as expected and the `messagedelivery` event is correctly sent to the master or the helper. However inbound messages are always notified to the master session - I don't see an issue with it.

Let me know, 

Kind Regards,

plugins/janus_sip.c
```diff
@@ -5419,7 +5483,88 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
...
+					scheme,
+					realm,
+					authuser,
+					session->account.secret_type == janus_sip_secret_type_hashed ? "HA1+" : "",
+					secret);
+				JANUS_LOG(LOG_VERB, "\t%s\n", auth);
+				/* Authenticate */
+				nua_authenticate(nh,
+					NUTAG_AUTH(auth),
+					TAG_END());
```
ycherniavskyi, line 5537: Just noticed that this bloc of code repeated the third time. Could we move it to a separate function?

I am implementing `publish` command and will repeat it a fourth time. For sure I could do it in my PR, just curious is it the right way of refactoring, or is there some other sense in repeating the same code several times?

thetechpanda: It could be a function or could be before the main `switch(event)` in `janus_sip_sofia_callback`. Auth and ProxyAuth are not used outside of this function, IMHO it is better to split code into functions only if used elsewhere.

lminiero:
> Hello,
> 
> I've rebuilt the test bench, using kamailio. I've tested using http and websocket transports and done multiple tests sending and receiving messages. I've debugged as extensively as I can, I've not seen any unexpected behaviour and I've carefully reviewed the locks, hope this time the patch is worthy or your time.

Contributions are always worthy of my time! :relaxed: 
I'll try to make some tests, too, e.g., to check if there can be minor leaks :+1: 

> I've also added some documentation at the top of the plugin to cater for the new behaviour: optional `call_id` for `message` requests and `messagedelivery` event documentation.

Ack, thx!

> I've noticed something, sending messages via Helpers works as expected and the `messagedelivery` event is correctly sent to the master or the helper. However inbound messages are always notified to the master session - I don't see an issue with it.

I guess this depends on what kind of message this is. If this is an out-of-dialog MESSAGE, then yes, it will be delivered to the owner of SIP stack, so the master session: if it's an in-dialog MESSAGE, though, as part of an ongoing call, it should go to the session that's actually handling that call. I guess you tested the former? We'll probably also want to make some tests with in-dialog MESSAGE transactions, just to make sure nothing breaks there.

plugins/janus_sip.c
lminiero: We do re-use the same auth code for different SIP responses, true. Anyway, this isn't something I'd do here: this is a change I'd do separately in another PR, devoted just to that, so I can look into this once we merge this effort.

html/siptest.html
```diff
@@ -108,10 +108,12 @@ <h3>Demo details</h3>
 								<span class="input-group-addon"><i class="fa fa-phone fa-fw"></i></span>
 								<input class="form-control" type="text" placeholder="SIP URI to call (e.g., sip:1000@example.com)" autocomplete="off" id="peer" onkeypress="return checkEnter(this, event);" />
 							</div>
-							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video
+							<input class="form-control" type="text" placeholder="any text message" autocomplete="off" id="message" ></input>
+							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="call">Call</button> <input autocomplete="off" id="dovideo" type="checkbox" />Use Video<br />
+							<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="sendmessage" onclick="return doMessage();" >Message</button>
```
lminiero, line 113: Not sure what these are for? We already automatically add buttons for MESSAGE and INFO to calls, e.g.

https://github.com/meetecho/janus-gateway/blob/master/html/siptest.js#L500

html/siptest.js
```diff
@@ -961,10 +971,12 @@ function addHelper(helperCreated) {
 		'					<span class="input-group-addon"><i class="fa fa-phone fa-fw"></i></span>' +
 		'					<input disabled class="form-control" type="text" placeholder="SIP URI to call (e.g., sip:1000@example.com)" autocomplete="off" id="peer' + helperId + '" onkeypress="return checkEnter(this, event, ' + helperId + ');"></input>' +
 		'				</div>' +
-		'				<button disabled class="btn btn-success margin-bottom-sm" autocomplete="off" id="call' + helperId + '">Call</button> <input autocomplete="off" id="dovideo' + helperId + '" type="checkbox">Use Video</input>' +
+		'				<input class="form-control" type="text" placeholder="any text message" autocomplete="off" id="message' + helperId + '" ></input>' +
+		'				<button disabled class="btn btn-success margin-bottom-sm" autocomplete="off" id="call' + helperId + '">Call</button> <input autocomplete="off" id="dovideo' + helperId + '" type="checkbox">Use Video</input><br />' +
+		'				<button class="btn btn-success margin-bottom-sm" autocomplete="off" id="sendmessage' + helperId + '" onclick="return doMessage('+helperId+');">Message</button>' +
```
lminiero, line 976: Same here.

html/siptest.js
```diff
@@ -1519,3 +1541,33 @@ function removeHelper(helperId) {
 		$('#sipcall'+helperId).remove();
 	}
 }
+/**
+ * send message function
+ *
+ * @param {number} suffix helper id
+ */
+function doMessage(suffix) {
```
lminiero, line 1549: Doxygen style documentation for JavaScript functions is not needed, a brief one-liner comment is more than enough.

plugins/janus_sip.c
```diff
@@ -1077,6 +1093,11 @@ static void janus_sip_session_destroy(janus_sip_session *session) {
 		janus_refcount_decrease(&session->ref);
 }
 
+static void janus_sip_message_key_destroy(char *key) {
+	if (key)
+		g_free(key);
+}
```
lminiero, line 1099: This is actually not needed: you can use `g_free` as a `GDestroyNotify` callback directly.

plugins/janus_sip.c
```diff
@@ -1937,7 +1958,7 @@ int janus_sip_init(janus_callbacks *callback, const char *config_path) {
 	sessions = g_hash_table_new_full(NULL, NULL, NULL, (GDestroyNotify)janus_sip_session_destroy);
 	identities = g_hash_table_new(g_str_hash, g_str_equal);
 	callids = g_hash_table_new(g_str_hash, g_str_equal);
-	messageids = g_hash_table_new(g_str_hash, g_str_equal);
+	messageids = g_hash_table_new_full(NULL, NULL, (GDestroyNotify)janus_sip_message_key_destroy, (GDestroyNotify)janus_sip_session_destroy);
```
lminiero, line 1961: No, this should not be `janus_sip_session_destroy`: this will mark the whole session as destroyed, and prevent the SIP stack to be used further. Besides, it will cause leaks, because `janus_sip_session_destroy` is already used as the callback to use for actually getting rid of a session, and since it can only be invoked once, a reference will remain dangling. You should instead add a new function that just dereferences a session, e.g., like we do here for publishers in the VideoRoom:

https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_videoroom.c#L1750

so basically something like a `janus_sip_session_dereference` function that you can use as a `GDestroyNotify` callback to remove the reference you added previously.

lminiero (commented):
Just checked your changes, and added some more notes inline  :+1:

html/siptest.html
thetechpanda: They are for out-of-dialog messages. There is no way to send messages without making a call first. Do you want me to revert these?

html/siptest.js
thetechpanda: will do

plugins/janus_sip.c
thetechpanda: ack

plugins/janus_sip.c
thetechpanda: ack

html/siptest.js
```diff
@@ -843,14 +853,14 @@ function doCall(ev) {
 		bootbox.alert('Please insert a valid SIP address (e.g., sip:pluto@example.com)');
 		$('#peer' + suffix).removeAttr('disabled');
 		$('#dovideo' + suffix).removeAttr('disabled');
-		$('#call' + suffix).removeAttr('disabled').click(function() { doCall(helperId); });
+		$('#call' + suffix).removeAttr('disabled').click(function(ev) { doCall(ev); });
```
thetechpanda, line 856: I think this was a bug, after a failed attempt to call a sip uri (due invalid sip uri), `doCall` was receiving the helperId instead of the event. Can you double check this?

thetechpanda:
I've done the requested changes, I'm not too sure about the function name `janus_sip_session_dereference` as it is used only by the messageids hashtable.

I've tested in-dialog and out-of-dialog messages and seem to work as expected. I've also fixed an issue I was having with siptest.js where the helpers could not make the call after trying to dial an invalid SIP uri.

If you still feel we don't need the out-of-dialog message UI, let me know and I'll revert the changes, if you want me to change the ui I am more than happy to do further changes. 

Apologies for the long time this is taking and thank you for taking the time to review my code.

html/siptest.js
lminiero: Yeah it's definitely a leftover from a previous version of that function, as `doCall` does expect an object and not an ID now. Your change seems fine, since `ev` here is whatever jQuery adds as the argument to `click` (so the button info IIUC), and not the `ev` that was passed to `doCall` in the first place.

lminiero:
Apologies for the delay, but I've been quite busy last week. Your latest changes do look fine to me now! To address your notes:

> I'm not too sure about the function name `janus_sip_session_dereference` as it is used only by the messageids hashtable.

That's fine: it's a good function to have in case we'll need it again for other hashtables or lists in the future.

> If you still feel we don't need the out-of-dialog message UI, let me know and I'll revert the changes, if you want me to change the ui I am more than happy to do further changes.

Yeah, I personally don't think we need this new UI, which would complicate and make the UI even more confusing than it currently is (which is my fault!). For now, I think it's enough to have it documented in Doxygen that "message" can be used both in-dialog and out-of-dialog. We can always consider changes to the UI in separate PRs in the future, if it's needed.

> Apologies for the long time this is taking and thank you for taking the time to review my code.

No problem at all, no one's in a hurry! I always appreciate contributions, and I know people are busy, so I never expect anyone to do things any faster than they can/want: that's how open source works :slightly_smiling_face:

thetechpanda:
For some reason VSCode kept changing the spacing for my changes after I'd save, hope you don't mind the two styling commits in the latest code push.

I've removed the `out-of-dialog` ui, anything else you'd like me to address?

lminiero:
I think that the two `</div>` fixes are actually correct, and a typo on my end originally.
Looks good now, thanks again for the effort and the patience! Merging :+1:

thetechpanda:
yay!

PR #2787, opened 2021-10-06
Fix admin unix sockets

thatsmydoing:
The `admin_ws_unix` and `admin_wss_unix` options were mistakenly being taken from the `general` section instead of the `admin`section.

Additionally, `admin_ws_unix` did not actually set the `LWS_SERVER_OPTION_UNIX_SOCK` so it would not work anyway.

I decided to just factor out the whole websocket server creation so that any further changes here don't need to be repeated for each variation of the server.

### DIFF
--- a/transports/janus_websockets.c
+++ b/transports/janus_websockets.c
@@ -819,7 +819,7 @@ int janus_websockets_init(janus_transport_callbacks *callback, const char *confi
 				}
 				ip = iface;
 			}
-			item = janus_config_get(config, config_general, janus_config_type_item, "admin_ws_unix");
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @thatsmydoing! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero (commented):
Thanks for spotting the issue! I think the refactoring you did is quite helpful, even though from a quick glance it will need fixing (see below).

thatsmydoing:
I've fixed the issue and signed the CLA

lminiero:
Looks good, merging :v:

PR #2789, opened 2021-10-12
Initial support for DTX (EchoTest, VideoRoom)

lminiero:
Small patch to make EchoTest and VideoRoom aware of DTX in negotiation, so that it can be enabled in case. Not tested extensively yet, which will need to be done to see if there can be an impact on recordings or audio quality in rooms. Since this patch mostly touched the SDP negotiation those plugins do, I took advantage of this to also change something else, e.g., allow the VideoRoom to negotiate stereo on subscribers if the publisher offers it.

Feedback welcome!


lminiero:
Merging.

PR #2792, opened 2021-10-28
utils: add janus_strlcat helper

tmatth:
This will log an error if the output is truncated.

### DIFF
--- a/postprocessing/janus-pp-rec.c
+++ b/postprocessing/janus-pp-rec.c
@@ -119,6 +119,7 @@ Usage: janus-pp-rec [OPTIONS] source.mjr
 #include <jansson.h>
 
 #include "../debug.h"
+#include "../utils.h"
 #include "../version.h"
 #include "pp-cmdline.h"
 #include "pp-rtp.h"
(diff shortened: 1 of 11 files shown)


utils.c
```diff
@@ -1217,3 +1217,14 @@ size_t janus_gzip_compress(int compression, char *text, size_t tlen, char *compr
 	return zs.total_out;
 }
 #endif
+
+
+gsize janus_strlcat(gchar *dest, const gchar *src, gsize dest_size)
+{
```
lminiero, line 1223: Just a code style nit: please put opening bracket on same line as function.

postprocessing/janus-pp-rec.c
```diff
@@ -117,6 +117,7 @@ Usage: janus-pp-rec [OPTIONS] source.mjr
 #include <jansson.h>
 
 #include "../debug.h"
+#include "../utils.h"
```
lminiero, line 120: I'll have to check if this means we need to add `utils.o` as an object when building `janus-pp-rec`. I see the automated Github Actions checks have all succeeded, so this shouldn't be an issue.

lminiero (commented):
Apart from a tiny code-style nit (I know, I'm the worst :laughing: ), this looks like a good fence to me. I already know it will cause me more than a little headache when re-aligning multistream (it will likely cause many conflicts), but that's indeed needed. I'll make a couple of tests, then merge and address the multistream conflicts so that we can start addressing the SDP size there.

postprocessing/janus-pp-rec.c
tmatth: `utils.c` is already listead as part of `janus_pp_rec_SOURCES` in `Makefile.am` so I think there's no change really.

lminiero:
Thanks for the quick fix! Merging then, and heading down the multistream merge conflicts hole...

PR #2797, opened 2021-11-04
Grow buffer as needed when generating SDPs (fixes #2791, see #2793)

### LINKED ISSUE #2791: Empty (broken?) candidate in janus SDP anwer on publisher joining
I am back again working on https://github.com/meetecho/janus-gateway/pull/2769 and merged the latest master into my feature branch.

I now notice that my publishing chrome browser has issues handling the SDP answer from janus as it contains an empty 😲 candidate. I never saw that and it looks pretty much like a bug to me. (I cannot tell if that is now master merge related, the dependencies have not been touched...)

Here just a snap of the candidates. The last one in video is always (reproducible) empty.

...
a=candidate:21 1 tcp 1010830335 fd10::d2d:4b3c:2118:ed68 21399 typ host tcptype passive
a=candidate:22 1 udp 2015366399 fd10::c501:a7c6:cbdf:c8d3 27558 typ host
⁉ a=candidate:23

Full offer answer is here:
https://pastebin.com/tKvkHuDH

Any hint how to nail that down?

Jan

lminiero:
This patch tries to address #2791 the same way we handled it in the `multistream` branch, that is by using static buffers just for the m-line, and growing the whole SDP as needed. While less flexible than #2793, this should be a simpler fix that requires no new method, and so no changes to the signatures of the methods we use.

### DIFF
--- a/janus.h
+++ b/janus.h
@@ -37,8 +37,6 @@
 #include "plugins/plugin.h"
 
 
-#define JANUS_BUFSIZE	8192
-
 /*! \brief Helper to address requests and their sources (e.g., a specific HTTP connection, websocket, RabbitMQ or others) */
 typedef struct janus_request janus_request;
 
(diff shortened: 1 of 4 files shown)


JanFellner:
Looking good. Please merge :)

lminiero:
Made some tests and it seems to be ok, merging :v:

PR #2802, opened 2021-11-09
Add configurable expected loss to AudioBridge to actually send FEC

lminiero:
While we've been supporting Opus FEC in the AudioBridge for a while, we were not actually using it properly on the way out, that is, on streams encoded by the AudioBridge itself. In fact, the libopus documentation explains that, even when you tell the encoder to use FEC, no FEC will actually be added to packets out of the box: in order for that to happen, you also need to tell the encoder the percentage of packets you expect to be lost after they're sent, so that the encoder can add enough info to help compensate such a loss.

This patch tries to address exactly that. More specifically, it adds new settings you can configure for participants (at join time and dynamically, using "configure") to modify this expected loss property. If FEC is negotiated and `expected_loss` for a participant is set, e.g., to `10`, then the encoder for that participant will indeed add FEC info to packets. I tested this by checking the stats in webrtc-internals, which shows the number of FEC packets that have been received and used: as expected, under normal circumstancies (no losses) I saw many FEC packets being received, and all of them discarded since they were not needed; simulating some loss with iptables did show some of them being used, and I could hear the audio sounding "better".

The PR also adds a per-room default expected loss property: if you don't set anything, it will be 0 (so no FEC added even if negotiated), but you can set it to something higher than that so that by default participants can be configured with some FEC without the need for a per-participant setting. This per-room default is also automatically used for RTP forwarders, which helps when you're actually broadcasting a mixer room somewhere else (e.g., via the Streaming plugin), as that way you'll have some FEC info the subscribers can take advantage of too. Notice we don't have a per-forwarder FEC/loss setting since encoders are shared when forwarders are used, which means we can't set expected loss X on one and Y on another.

I only tested this briefly, so I'm looking forward to feedback from you in case you decide to experiment with this (which you should if you'd like better audio quality in the AudioBridge).

### DIFF
--- a/conf/janus.plugin.audiobridge.jcfg.sample
+++ b/conf/janus.plugin.audiobridge.jcfg.sample
@@ -11,6 +11,7 @@
 # audio_active_packets = 100 (number of packets with audio level, default=100, 2 seconds)
 # audio_level_average = 25 (average value of audio level, 127=muted, 0='too loud', default=25)
 # default_prebuffering = number of packets to buffer before decoding each particiant (default=6)
+# default_expectedloss = percent of packets we expect participants may miss, to help with FEC (default=0)
 # record = true|false (whether this room should be recorded, default=false)
 # record_file = "/path/to/recording.wav" (where to save the recording)
 # record_dir = "/path/to/" (path to save the recording to, makes record_file a relative path if provided)
--- a/conf/janus.plugin.audiobridge.jcfg.sample
+++ b/conf/janus.plugin.audiobridge.jcfg.sample
@@ -12,6 +12,7 @@
 # audio_level_average = 25 (average value of audio level, 127=muted, 0='too loud', default=25)
 # default_prebuffering = number of packets to buffer before decoding each particiant (default=6)
 # default_expectedloss = percent of packets we expect participants may miss, to help with FEC (default=0, max=20; automatically used for forwarders too)
+# default_bitrate = default bitrate in bps to use for the all participants (default=auto, libopus decides; automatically used for forwarders too)
 # record = true|false (whether this room should be recorded, default=false)
 # record_file = "/path/to/recording.wav" (where to save the recording)
 # record_dir = "/path/to/" (path to save the recording to, makes record_file a relative path if provided)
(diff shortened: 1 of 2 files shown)


atoppi:
Some interesting things that I discovered while tinkering with the PR:
- FEC packets **are not** produced if there is only 1 participant in the audiobridge
- FEC packets **are not** produced if mic volume is 0 in the OS settings
- Sampling rate configuration seems to play an important role
  - with 16kHz sampling and expected loss < 6%, the mixer does not produce any FEC data
  - using higher sampling rate (e.g. 48 kHz) the mixer will produce FEC data also with low expected loss (e.g. 2%)

I'm not an expert of opus  and libOpus at all, so maybe someone with more context could come up with an explanation.

fippo:
You might want to cap the loss at 20% -- this is what WebRTC does [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/audio_coding/codecs/opus/audio_encoder_opus.cc;l=728).

Also mind you that the fec bitrate gets subtracted from the target bitrate ([some pointers here](https://webrtchacks.com/red-improving-audio-quality-with-redundancy/)) so you might want to advise a higher target bitrate.

lminiero:
> You might want to cap the loss at 20% -- this is what WebRTC does [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/audio_coding/codecs/opus/audio_encoder_opus.cc;l=728).

Ah that's helpful to know, thanks!

> Also mind you that the fec bitrate gets subtracted from the target bitrate ([some pointers here](https://webrtchacks.com/red-improving-audio-quality-with-redundancy/)) so you might want to advise a higher target bitrate.

Yep, I was aware FEC "stole" room from the actual packets, even if it's not clear to me by how much. We're not using `OPUS_SET_BITRATE` anywhere in the AudioBridge code, which means the library is probably deciding a reasonable bitrate by itself depending on the sampling rate we configure. IIRC for 16khz (which is what we use most of the times) the bitrate is ~20k. I agree it does make sense to also allow users to configure the bitrate to use for a connection, so I'll add this to this patch.

Thanks for the feedback!

lminiero:
Made a commit that does both, I'll give it a try later this afternoon.

lminiero:
Yep, seems to be working, and to address Alessandro's notes, as soon as I started increasing the bitrate, libopus started to honour lower expected loss values too: for a 16khz sampling rate stream, for instance, libopus defaulted to ~14kbps; setting a bitrate of ~24kbps made libopus start sending FEC also with an expected_loss of 1.

lminiero:
Merging.

PR #2803, opened 2021-11-10
Disable slowlink events by default

lminiero:
This PR changes the default setting for the slowlink threshold, that is the value we use to decide how many lost packets per second should trigger a slowlink event via Janus API (and the related callback in plugins). Previously this value was `4`, meaning that if in or out we detected at least 4 lost packets, we'd trigger a slow link event, and would notify plugins.

We did this originally as a way to provide applications with some info they could use, e.g., to react somehow, but in practice the way it currently works just means an excessively verbose API usage (events spam), which is actually useless and redundant, since clients that have access to client stats can get the same information (actually more) and faster on their own, e,g., via getStats. As such, we decided to disable the feature by default: of course, it can still be enabled in the settings, or dynamically via Admin API, if you need it for anything.

I'm publishing this as a PR and not committing directly as it may have an impact on existing applications, e.g., people actually relying on the event for anything. This should give them time to become aware of this, and explicitly set the value in `janus.jcfg` to whatever they currently use if they weren't doing it already. I plan to merge soon anyway, so in case you have feedback on this please let us know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,7 @@ or on the command line:
 	-W, --slowlink-threshold=number
                                   Number of lost packets (per s) that should
                                   trigger a 'slowlink' Janus API event to users
+                                  (default=0, feature disabled)
 	-r, --rtp-port-range=min-max  Port range to use for RTP/RTCP (only available
 								  if the installed libnice supports it)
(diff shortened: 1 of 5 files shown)


lminiero:
Merging.

PR #2805, opened 2021-11-11
Obtaining subscriber's feed under session mutex in order to avoid races

cppdev-1:
There are places where subscriber->feed is accessed. Sometimes it is done w/o any synchronisation mechanism. In this PR session->mutex is used when subscriber->feed is accessed in order to avoid possible races.

I'll add a second commit with janus_videoroom_relay_rtp_packet refactoring.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1729,6 +1729,33 @@ typedef struct janus_videoroom_rtp_relay_packet {
 	gboolean textdata;
 } janus_videoroom_rtp_relay_packet;
 
+static janus_videoroom_publisher *janus_videoroom_subscriber_feed_acquire(janus_videoroom_subscriber *subscriber) {
+	janus_videoroom_publisher *p = NULL;
+
+	if (subscriber == NULL || g_atomic_int_get(&subscriber->destroyed) || subscriber->session == NULL)
+		goto exit;
+
+	janus_mutex_lock(&subscriber->session->mutex);
+	p = subscriber->feed;
+	if (p && !g_atomic_int_get(&p->destroyed) && p->session) {
+		janus_refcount_increase(&p->ref);
+		janus_refcount_increase(&p->session->ref);
+	}
+	else
+		p = NULL;
+	janus_mutex_unlock(&subscriber->session->mutex);
+
+	exit:
+	return p;
+}
+
+static void janus_videoroom_subscriber_feed_release(janus_videoroom_publisher *p) {
+	if (p) {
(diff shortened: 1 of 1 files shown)


lminiero:
While I do see value in this, it changes the internals of the plugins too much for me to be comfortable merging this before multistream, as it would cause changes in the plugin there that would be hard to revert. In case, we can keep this on hold until multistream is merged, and then merge this into the `0.x` branch we'll create to keep the existing master alive.

On `janus_videoroom_relay_rtp_packet`, I'm not sure we need any change there, considering it's already protected by a mutex:

	janus_mutex_lock_nodebug(&participant->subscribers_mutex);
	g_slist_foreach(participant->subscribers, janus_videoroom_relay_rtp_packet, &packet);
	janus_mutex_unlock_nodebug(&participant->subscribers_mutex);

cppdev-1:
Ok then.

As for `janus_videoroom_relay_rtp_packet`, yes, it is protected by `publisher->subscribers_mutex` and `session->mutex` like here
```
...
janus_mutex_lock(&prev_feed->subscribers_mutex);
prev_feed->subscribers = g_slist_remove(prev_feed->subscribers, subscriber);
janus_mutex_unlock(&prev_feed->subscribers_mutex);
...
g_clear_pointer(&subscriber->feed, janus_videoroom_publisher_dereference);
janus_mutex_unlock(&session->mutex);
...
```
My idea is to do this
```
...
typedef struct janus_videoroom_relay_rtp_packet_params {
	janus_videoroom_rtp_relay_packet *packet;
	janus_videoroom_publisher *publisher;
} janus_videoroom_relay_rtp_packet_params;
...
void janus_videoroom_incoming_rtp(janus_plugin_session *handle, janus_plugin_rtp *pkt) {
...
janus_videoroom_relay_rtp_packet_params relay_params = { .packet = &packet, .publisher = participant };
janus_mutex_lock_nodebug(&participant->subscribers_mutex);
g_slist_foreach(participant->subscribers, janus_videoroom_relay_rtp_packet, &relay_params);
janus_mutex_unlock_nodebug(&participant->subscribers_mutex);
...
```
Maybe this is for a separate PR.

cppdev-1:
@lminiero What about this PR?

cppdev-1:
@lminiero Well, this PR is about using mutex when accessing subscriber's feed and as far as I understood, you had agreed to merge it with the `0.x` branch. I still think changes introduced in this PR are important.
As for another PR I mentioned, it is about refactored verson of `janus_videoroom_relay_rtp_packet_params`.

PR #2806, opened 2021-11-12
add support for custom datachannel options in janus.js

sqxieshuai:


lminiero:
Thanks! I'll have to check if this can cause issues with Janus, as IIRC we do make some assumptions (e.g., ordered=true) in the Janus core too (sctp.c). Have you checked if the custom datachannel options work as expected for you when you use them?

sqxieshuai:
@lminiero In my business, I use datachannel to send some data that does not depend on reliability, such as heartbeat and speed measurement. After using it for a period of time, no serious problems are found.

lminiero:
Thanks, merging then! :v:

PR #2808, opened 2021-11-15
Solve the media state flag is missing in case of fast renegotiations

JanFellner:
This is an idea how to retransmit the media state in case the connection is renegotiating.
I don´t know where the sdp is handshaked so how it would be possible to set the flag to unknown.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -667,22 +667,39 @@ static int janus_seq_in_range(guint16 seqn, guint16 start, guint16 len) {
 	return (s <= n && n < e) || (s <= nh && nh < e);
 }
 
-/* Checks wether based on the conditions in the stats an update to the mediastate needs to get dispatched
+/* Checks whether based on the conditions in the stats an update to the mediastate needs to get dispatched
  * stats - the stats object we are inspecting
  * now - the current monotonic time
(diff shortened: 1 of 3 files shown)


JanFellner:
This PR fixes the issue that after an SDP exchange the (inactive, send_ony) on a publisher connection janus is not transmitting a media receiving event.

What was changed to achive this:
* The janus_ice_stats_info now contains a *last_notified* enum that expresses which media state janus has dispatched before
* In case of a SDP handshake that enables or disables media, this value is set to **JANUS_ICE_MEDIA_STATE_UPDATE_PENDIG** and a timestamp **send_update_after** is stored.
* Either the state now changes to UP or DOWN by the regular operation or the time stored in **send_update_after** expires and janus simply dispatches the current state
* The logic that checks wether the media state has to get signalled was relocated into a dedicated method that handles it.
* Documenation in that method was added to express which conditions is hit under which circumstance
* The notified_lastsec boolean was renamed to temporarily_down to better express the purpose of the variable

Comments welcome

JanFellner:
will get addressed by https://github.com/meetecho/janus-gateway/pull/2824

PR #2809, opened 2021-11-16
Multistream html fixes

aet:
I was doing some testing against https://github.com/meetecho/janus-gateway/pull/2211 and browsed around the differences between master/multistream JS code examples. I noticed that mediaState arguments on some of the examples are out of order

Follow-up notes/questions:
- html/nosiptest.js has onlocalstream callback and onremotetrack, but didn't update onlocalstream variant into onlocaltrack
- Janus Admin page is showing [object Object] in few places when JSON pretty printing is enabled IIRC - due to streams moved into JSON media arrays :)
- mediaState/slowLink are the only callbacks where mid parameter is not the first one after handle, should this be unified?
- There's no question that once multistream is in, there will be breaking changes, but alternatively should the mid parameters be the last one on all callbacks instead and try to preserve existing arguments order when possible?

### DIFF
--- a/html/audiobridgetest.js
+++ b/html/audiobridgetest.js
@@ -133,7 +133,7 @@ $(document).ready(function() {
 								iceState: function(state) {
 									Janus.log("ICE state changed to " + state);
 								},
-								mediaState: function(medium, mid, on) {
+								mediaState: function(medium, on, mid) {
 									Janus.log("Janus " + (on ? "started" : "stopped") + " receiving our " + medium + " (mid=" + mid + ")");
 								},
 								webrtcState: function(on) {
--- a/html/canvas.js
+++ b/html/canvas.js
@@ -152,7 +152,7 @@ $(document).ready(function() {
 								iceState: function(state) {
 									Janus.log("ICE state changed to " + state);
 								},
-								mediaState: function(medium, mid, on) {
+								mediaState: function(medium, on, mid) {
 									Janus.log("Janus " + (on ? "started" : "stopped") + " receiving our " + medium + " (mid=" + mid + ")");
 								},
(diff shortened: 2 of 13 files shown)


lminiero:
Thanks for looking into this! Answering the different points:

> html/nosiptest.js has onlocalstream callback and onremotetrack, but didn't update onlocalstream variant into onlocaltrack

Probably a leftover, should be easy to fix in case, I'll have a look later.

>     * Janus Admin page is showing [object Object] in few places when JSON pretty printing is enabled IIRC - due to streams moved into JSON media arrays :)

To be honest I was planning to get rid of the "pretty print" part entirely, since it needs to be aware of what can be there, and that can change (as it did here). Besides, it's actually less pretty than the JSON ostensibly :laughing: 


> mediaState/slowLink are the only callbacks where mid parameter is not the first one after handle, should this be unified?

I think that was done to preserve a semblance of backwards compatibility, especially in case you only do a single audio and video stream anyway: in that case, you don't really care about the mid. The new onlocaltrack and onremotetrack are brand new callbacks, instead, so we could use any signature we wanted.

>     * There's no question that once multistream is in, there will be breaking changes, but alternatively should the mid parameters be the last one on all callbacks instead and try to preserve existing arguments order when possible?

Actually I think that semantically it's more correct for mid to appear first, since it's what really identifies a specific track.

aet:
I agree mid argument appearing first, but might as well asking about alternatives :)

Equally fine with either mid appearing as first argument or last for mediaState and slowLink, but for consistency I would go with mid appearing first on them too. Currently only mvideoroom.js is declaring the mediaState signature the way janus.js is using it.

lminiero:
Merging, thanks!

PR #2812, opened 2021-11-23
Fix signed token auth not work on join to the videoroom #2810

### LINKED ISSUE #2810: Signed token authentication doesn't work well on join to the room.
The issue is present on the latest master branch.

Cases with `token_auth = true` and `token_auth_secret = "janus871625371523"` in janus.cfg.

Scenario
1) Generate two tokens for all requests and for specail videoroom plugin:
    a) `1637606957,janus,janus.plugin.videoroom:/rWSlAurLtqWhVgRQ/x/UEO+IzI=`
    b) `1637606957,janus.plugin.videoroom,room=3edb0cdd-7173-481d-82f0-5eb31937d20f:XZQHOyCKF02PA48PIeUftAZjwC0=`
2) Connect to Websocket and use (a) token - all is fine. (without token we receive not authorized - all good).
3) Join the room `3edb0cdd-7173-481d-82f0-5eb31937d20f`:
3.1) **Actual behavior**: Try to join the room `3edb0cdd-7173-481d-82f0-5eb31937d20f` without token - we enter the room. Also with token (b) we will enter the room.
3.2) **Expected behavior**: Try to join the room `3edb0cdd-7173-481d-82f0-5eb31937d20f` without token - not authorized error. With token (b) - enter the room.

There are an issues in `plugins/janus_videoroom.c` (janus_videoroom_access_room function) and in `auth.c` (janus_auth_check_signature_contains function).

I can fix the bug.

timsolov:
Fixes #2810

### DIFF
--- a/auth.c
+++ b/auth.c
@@ -121,8 +121,9 @@ fail:
 }
 
 gboolean janus_auth_check_signature_contains(const char *token, const char *realm, const char *desc) {
-	if (!auth_enabled || auth_secret == NULL)
-		return FALSE;
+	if (!auth_enabled || auth_secret == NULL) {
+		return TRUE;
+	}
 	gchar **parts = g_strsplit(token, ":", 2);
 	gchar **data = NULL;
 	/* Token should have exactly one data and one hash part */
--- a/auth.h
+++ b/auth.h
@@ -42,7 +42,7 @@ gboolean janus_auth_check_signature(const char *token, const char *realm);
  * @param[in] token The token to validate
  * @param[in] realm The token realm
  * @param[in] desc The descriptor to search for
- * @returns TRUE if the token is valid, not expired and contains the descriptor, FALSE otherwise */
+ * @returns When the token based authentication is enabled: TRUE if  the token is valid, not expired and contains the descriptor, FALSE otherwise. When the token based authentication is disabled: always TRUE. */
 gboolean janus_auth_check_signature_contains(const char *token, const char *realm, const char *desc);
 
 /*! \brief Method to add a new valid token for authenticating
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1162,6 +1162,7 @@ room-<unique room ID>: {
 
 #include "plugin.h"
 
+#include <string.h>
 #include <jansson.h>
 #include <netdb.h>
 
@@ -2997,7 +2998,7 @@ static int janus_videoroom_access_room(json_t *root, gboolean check_modify, gboo
 		/* signed tokens bypass pin validation */
 		json_t *token = json_object_get(root, "token");
 		if(token) {
-			char room_descriptor[26];
+			char room_descriptor[100];
 			g_snprintf(room_descriptor, sizeof(room_descriptor), "room=%s", room_id_str);
 			if(gateway->auth_signature_contains(&janus_videoroom_plugin, json_string_value(token), room_descriptor))
 				return 0;
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -1162,7 +1162,6 @@ room-<unique room ID>: {
 
 #include "plugin.h"
 
-#include <string.h>
 #include <jansson.h>
 #include <netdb.h>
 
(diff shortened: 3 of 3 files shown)


januscla:
Thanks for your contribution, @timsolov! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

timsolov:
> Thanks for your contribution, @timsolov! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

signed

auth.c
```diff
@@ -121,8 +121,9 @@ gboolean janus_auth_check_signature(const char *token, const char *realm) {
 }
 
 gboolean janus_auth_check_signature_contains(const char *token, const char *realm, const char *desc) {
-	if (!auth_enabled || auth_secret == NULL)
-		return FALSE;
+	if (!auth_enabled || auth_secret == NULL) {
+		return TRUE;
```
lminiero, line 125: Good catch, as if this is disabled returning `FALSE` might trigger unwanted errors.

auth.h
```diff
@@ -42,7 +42,7 @@ gboolean janus_auth_check_signature(const char *token, const char *realm);
  * @param[in] token The token to validate
  * @param[in] realm The token realm
  * @param[in] desc The descriptor to search for
- * @returns TRUE if the token is valid, not expired and contains the descriptor, FALSE otherwise */
+ * @returns When the token based authentication enabled: TRUE if  the token is valid, not expired and contains the descriptor, FALSE otherwise. When the token based authentication disabled: always TRUE. */
```
lminiero, line 45: I'd say "is enabled", rather than just "enabled", in both additions.

plugins/janus_videoroom.c
```diff
@@ -1162,6 +1162,7 @@ room-<unique room ID>: {
 
 #include "plugin.h"
 
+#include <string.h>
```
lminiero, line 1165: I don't think this is necessary.

plugins/janus_videoroom.c
```diff
@@ -2997,10 +2998,13 @@ static int janus_videoroom_access_room(json_t *root, gboolean check_modify, gboo
 		/* signed tokens bypass pin validation */
 		json_t *token = json_object_get(root, "token");
 		if(token) {
-			char room_descriptor[26];
+			char room_descriptor[100];
```
lminiero, line 3001: Was this the actual cause of the issue? A string that was too small, and so was only containing a partial representation of the required token causing comparisons and checks to fail?

plugins/janus_videoroom.c
```diff
@@ -2997,10 +2998,13 @@ static int janus_videoroom_access_room(json_t *root, gboolean check_modify, gboo
...
 			g_snprintf(room_descriptor, sizeof(room_descriptor), "room=%s", room_id_str);
-			if(gateway->auth_signature_contains(&janus_videoroom_plugin, json_string_value(token), room_descriptor))
-				return 0;
+			if(!gateway->auth_signature_contains(&janus_videoroom_plugin, json_string_value(token), room_descriptor)) {
+				g_snprintf(error_cause, error_cause_size, "Unauthorized (wrong token %s)", token);
```
lminiero, line 3004: This is missing the `if(error_cause)` the other error management lines of code have, as `error_cause` may be `NULL`.

lminiero (commented):

plugins/janus_videoroom.c
timsolov: This affects on checking rooms with id in UUID format, the description have to be: `room=3edb0cdd-7173-481d-82f0-5eb31937d20f` but when it has only 26 characters it's looks like: `room=3edb0cdd-7173-481d-82` and when we do strcmp we have not equal result.

auth.c
timsolov: thank you

auth.h
timsolov: fixed

plugins/janus_videoroom.c
timsolov: removed

plugins/janus_videoroom.c
timsolov: done

lminiero:
Thanks for the quick fixes! This looks good to me, so I'll merge :+1:

PR #2814, opened 2021-11-23
Added basic history support to TextRoom plugin

lminiero:
As the title says, this PR adds basic support for history: if you configure a new room to hold a history of X messages, then incoming messages and announcements will be stored; when the queue is full, older messages are discarded. New users joining are then sent those messages as soon as they join (unless they ask not to).

I have done some basic testing and it seems to be working as expected, but you may of course want to test this more extensively if it's something you care about.


lminiero:
Merging.

PR #2816, opened 2021-11-25
Audiobridge: added ability play announcement only to specific participant

RSATom:
or exclude some participant from announcement.

@lminiero I've tested this patch some months on older Janus version, and plan to do the same for the latest stable/master so don't want you merge it right now, but I just would like to ask Is there any chance something like this could be merged? 

Thanks in advance!

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -1188,7 +1188,9 @@ static struct janus_json_parameter play_file_parameters[] = {
 	{"filename", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"file_id", JSON_STRING, 0},
 	{"group", JSON_STRING, 0},
-	{"loop", JANUS_JSON_BOOL, 0}
+	{"loop", JANUS_JSON_BOOL, 0},
+	{"to", JSON_STRING, 0},
+	{"exclude", JSON_STRING, 0}
 };
 static struct janus_json_parameter checkstop_file_parameters[] = {
 	{"file_id", JSON_STRING, JANUS_JSON_PARAM_REQUIRED}
@@ -1308,6 +1310,8 @@ typedef struct janus_audiobridge_file {
 	char *oggbuf;
 	gboolean started, loop;
 	gint state, headers;
+	char *to;
+	char *exclude;
 } janus_audiobridge_file;
 /* Helper method to open an Opus file, and make sure it's valid */
 static int janus_audiobridge_file_init(janus_audiobridge_file *ctx) {
@@ -1443,6 +1447,8 @@ static void janus_audiobridge_file_free(janus_audiobridge_file *ctx) {
 	if(ctx->headers > 0)
 		ogg_stream_clear(&ctx->stream);
 	ogg_sync_clear(&ctx->sync);
+	g_free(ctx->to);
+	g_free(ctx->exclude);
 	g_free(ctx);
 }
 #endif
@@ -4786,6 +4792,13 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 			g_snprintf(error_cause, 512, "Error creating Opus decoder");
 			goto prepare_response;
 		}
+		const char *to = json_string_value(json_object_get(root, "to"));
+		if(to)
+			p->annc->to = g_strdup(to);
+		const char *exclude = json_string_value(json_object_get(root, "exclude"));
+		if(exclude)
+			p->annc->exclude = g_strdup(exclude);
+
 		/* We're done, add the announcement to the room */
 		g_hash_table_insert(audiobridge->anncs, g_strdup(p->user_id_str), p);
 		janus_mutex_unlock(&audiobridge->mutex);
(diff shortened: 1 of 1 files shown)


plugins/janus_audiobridge.c
```diff
@@ -7756,7 +7769,8 @@ static void *janus_audiobridge_mixer_thread(void *data) {
 							buffer[i] += (resampled[i]*p->volume_gain)/100;
 						}
 					}
-				} else {
+				}
+				if(groups_num != 0) {
```
lminiero, line 7773: Why isn't this `} else if(groups_num != 0) {`? And why doesn't this have the to check as well? Not sure why exclude isn't part of this check.

plugins/janus_audiobridge.c
```diff
@@ -7767,9 +7781,24 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+
+					janus_mutex_lock(&p->qmutex);
+					/* Insert packets sorting by sequence number */
+					p->inbuf = g_list_prepend(p->inbuf, pkt);
+					janus_mutex_unlock(&p->qmutex);
+				}
+
 				ps = ps->next;
 			}
-			g_list_free_full(anncs_list, (GDestroyNotify)janus_audiobridge_participant_unref);
```
lminiero, line 7772: Why have you removed this?

plugins/janus_audiobridge.c
```diff
@@ -7852,6 +7881,47 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+
+#if HAVE_LIBOGG
+			for(GList *annc = anncs_list; annc; annc = annc->next) {
+				janus_audiobridge_participant *anncp = (janus_audiobridge_participant *)annc->data;
+
+				if(anncp->annc == NULL || g_atomic_int_get(&anncp->destroyed))
+					continue;
+
+				if(!anncp->annc->to && !anncp->annc->exclude)
+					continue; // already mixed
```
lminiero, line 7893: Code style: comments should be `/*  */`

plugins/janus_audiobridge.c
```diff
@@ -7852,6 +7881,47 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+
+				GList *first = g_list_first(anncp->inbuf);
+				janus_audiobridge_rtp_relay_packet *pkt = (janus_audiobridge_rtp_relay_packet *)(first ? first->data : NULL);
+				curBuffer = (opus_int16 *)((pkt && pkt->length && !pkt->silence) ? pkt->data : NULL);
+				if(!curBuffer)
+					continue;
+
+				const char* to = anncp->annc->to;
+				const char* exclude = anncp->annc->exclude;
```
lminiero, line 7902: See above on code style for pointers.

plugins/janus_audiobridge.c
```diff
@@ -7852,6 +7881,47 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+				if(to) {
+					if(0 == g_strcmp0(p->display, to)) {
```
lminiero, line 7904: So `to` and `exclude` are display names? I think this is very fragile. Display names are _not_ unique, and can be changed dynamically by people in the room. If you want this to be reliable and consistent, it should be user IDs instead (which means making the same checks we do on whether they're numbers or string IDs).

lminiero (commented):
Thanks for your contribution! To be honest, I'm not 100% sold on the idea and how it's implemented. Using display names is fragile, and it seems you're adding additional cycles to an already burdened loop: I feel like this could possibly be done in a lighter way, even though I haven't spent much time thinking about how. I don't fully understand why a single to and a single exclude, if not possibly to do the traditional "You're now in the room" and "Pluto has entered the room"?

plugins/janus_audiobridge.c
```diff
@@ -1308,6 +1310,8 @@ typedef struct janus_audiobridge_file {
 	char *oggbuf;
 	gboolean started, loop;
 	gint state, headers;
+	char* to;
+	char* exclude;
```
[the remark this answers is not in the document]
RSATom: Sorry, just used to a little bit different code style...

plugins/janus_audiobridge.c
RSATom: > Why isn't this } else if(groups_num != 0) {?
Just was thinking it's a little bit more clear... will fix.

> And why doesn't this have the to check as well? 
`to` handling a little bit later by add samples directly to buffers for specific user.
similar with `exclude` - it handling a little bit later by subtracting samples from buffers for specific user

P.S. I think there is another issue is there - I should take into account group notification - will fix...

plugins/janus_audiobridge.c
RSATom: `anncs_list` is required later and moved below

plugins/janus_audiobridge.c
RSATom: Good point. I need to think if it suitable in our case. Thanks!

RSATom:
> I don't fully understand why a single to and a single exclude, if not possibly to do the traditional "You're now in the room" and "Pluto has entered the room"?

`to` was added just as counterpart to `exclude`, just to have symmetry in API.
What about `exclude` - it's for a little bit more complex task. It's using it for realtime text-to-speach (with Google TTS API) for messages in chat (and for text-to-speach translated to different languages) - and user could want not hear it's own messages.

PR #2817, opened 2021-11-25
Fix potential Cross-site Scripting (XSS) exploits in demos

lminiero:
This patch addresses an issue in the TextRoom demo found by @SoufElhabti (thanks!). Specifically, while we were escaping incoming messages before adding them to the page, we weren't doing the same for display names: since they would be added to the HTML as-is as well, this could cause bad things to happen. See CVE-2021-4020 for details.

This patch fixes that vulnerability in the demo, but it also got me thinking about other demos, where we do similar things with display names, descriptions, etc.  In some demos we have checks on what you can put in a display name (e.g., by enforcing alphanumeric values), but those don't protect you from users joining via other means (e.g,, a webpage accessing the same VideoRoom but without the check). As such, I've modified the other demos as well to escape those values that may need it.

I've already updated the demos online with the fix, so it shouldn't be an issue anymore for people using those. I'm planning to merge this patch soon, so please let me know if I missed some file/value.

### DIFF
--- a/html/audiobridgetest.js
+++ b/html/audiobridgetest.js
@@ -178,7 +178,7 @@ $(document).ready(function() {
 												Janus.debug("Got a list of participants:", list);
 												for(var f in list) {
 													var id = list[f]["id"];
-													var display = list[f]["display"];
+													var display = escapeXmlTags(list[f]["display"]);
 													var setup = list[f]["setup"];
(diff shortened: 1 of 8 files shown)


atoppi (commented):
lgtm

lminiero:
Merging :+1:

PR #2823, opened 2021-11-29
Protect removal from hashtables when destroying SIP sessions (fixes #2818)

### LINKED ISSUE #2818: Race conditions when accessing callids and identities hash tables
callids and identities hash tables is not always protected by a mutex. For example,

```
static void janus_sip_session_free(const janus_refcount *session_ref) {
...
if(session->master == NULL && session->account.identity) {
		g_hash_table_remove(identities, session->account.identity);
		g_free(session->account.identity);
		session->account.identity = NULL;
}
...
if(session->callid) {
		g_hash_table_remove(callids, session->callid);
		g_free(session->callid);
		session->callid = NULL;
}
...
```
That leads to race conditions when accessing them.

lminiero:
See #2818 for details.


lminiero:
Merging.

Issue #2807, opened 2021-11-15
Missing media receiving after removing and immediate readding video


JanFellner:
I have an issue where janus is missing to send me a media receiving event in the videoroom.
I maybe just need a hint where and how to solve it. This ticket as is reporting the core issue.

I have a settled peerConnection containing audio and video. webrtcup and media receiving received for both the audio and the videotrack.

If i now disable and enable the video i would expect janus to send me the media receiving flag as the media actually has changed but janus seems to have stored the flag internally and not have reset it as i remove the video (by a=inactive).

Thus i do not reliably get the information that media is transported.

original SDP https://pastebin.com/eEq2Rf97 (after that everything is settled)
inactivate video SDP: https://pastebin.com/92khiwXx
reenabled video SDP: https://pastebin.com/mV5GA9je

JanFellner:
reopened as i the original ticket did not contain the description...

JanFellner:
The issue in the businesslogic is that when disabling i reset my internal status as i know there won´t be any media any longer. I do not wait for any media receiving false from janus.
But on the other hand after reenabling i would need that event.

So i understand that janus is caching that info to not send it all the time but the cache should get cleared once sdp is exchanged.

lminiero:
Janus has timers to detect missing media: I think for the `receiving:false` event we need to have one second without receiving any media. This means that if the renegotiation is quick enough, you won't get it.

JanFellner:
Do you think its feasable to intercept the sdp handshake here and to enforce a resend when sdp for that media has changed?

I have seen that issue in the past and tried to overcome it with checking the stats but that tries to tackle something in a roundabout fashion... (Cause on the one side you have the media flag and on the other the stats that will always delay everything)

JanFellner:
The PR fixes the issue. Check comments in the PR for what has changed

lminiero:
To be honest, I still don't understand exactly what the problem is, so it's hard for me to review your PR, or know what to review. It seems to be you may be expecting too much from an event that is not meant to be immediate nor particularly informative: it's actually quite slow, since it will typically only notify you about a second later or so. As a consequence, from a quick glance I feel the PR changes too much for me to be really comfortable with it, especially if this can be solved with less changes and a more trivial approach.

Can you please elaborate on exactly what happens, what changes, and what you expect to happen, to see if I can replicate it somehow?

JanFellner:
[...]

PR #2824, opened 2021-11-29
Send receiving:false notifications right after renegotiating (see #2807 and #2808)

lminiero:
This PR should address #2807, and is an alternative approach to #2808. Specifically, it tries to ensure that any time Janus is aware of a renegotiation that disables incoming media, a `receiving:false` notification for that medium is sent right away (a bit like we do when we start receiving media after a while, on the first incoming packet). As such, it also implicitly addresses the missing `receiving:true` notification when a further renegotiation adds incoming media back in (triggered again by the first incoming packet). This should also be easier to port to multistream, since it makes use of a generic event the loop can intercept and operate upon.

I tested this with the EchoTest plugin, by replicating a quick removeVideo/addVideo transition, and it seems to be working as expected. That said, further testing should be done. Pinging @JanFellner as the main reporter of both issue and original PR.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -387,6 +387,7 @@ static janus_ice_queued_packet
 	janus_ice_start_gathering,
 	janus_ice_add_candidates,
 	janus_ice_dtls_handshake,
+	janus_ice_media_stopped,
 	janus_ice_hangup_peerconnection,
 	janus_ice_detach_handle,
 	janus_ice_data_ready;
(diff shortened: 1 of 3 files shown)


JanFellner:
Thanks Lorenzo, will merge it into my remb adoption branch and give you feedback either today or tomorrow. (Depends on the meetingitis level (Amount of meetings) today :)

lminiero:
> Depends on the meetingitis level (Amount of meetings) today :)

:rofl:

JanFellner:
Checked, looks good to me.
Event about receiving comes super fast after renegotiation.
👍👍👍 three thumbs up for solving an issue which bothered me a year

JanFellner:
Closed my PR. Waiting for the merge :)

lminiero:
Ack, I'll merge then :+1:

PR #2825, opened 2021-11-30
Allow VideoRoom to choose whether signed tokens should be used

lminiero:
After a recent fix contributed in #2812, some users experienced issues when using the VideoRoom with tokens. Specifically, when Janus was used in this specific configuration:

* signed tokens used in the core
* regular tokens used in the VideoRoom (ACL)

access would always fail, because the plugin was hardcoded to assume signed tokens would be used there too when enabled in the core. This patch tries to address that, by adding a new core callback to check if signed-mode is active, and adding a new property when creating VideoRoom to choose whether they should be used in the plugin too. By default, signed tokens are _never_ used, which means you need to set `signed_tokens:true` when creating a room yourself if you do want them instead. Of course, trying to enable signed tokens when they're not active in the core will not work, and a warning will be displayed.

I haven't tested this yet, but I'd ask @timsolov to make sure I didn't make any change that invalidates the PR he contributed (noting again that `signed_token` must be manuallyset to `true` when creating the room or they won't be used).

### DIFF
--- a/auth.c
+++ b/auth.c
@@ -70,6 +70,10 @@ gboolean janus_auth_is_stored_mode(void) {
 	return auth_enabled && tokens != NULL;
 }
 
+gboolean janus_auth_is_signed_mode(void) {
+	return auth_enabled && auth_secret != NULL;
+}
+
 void janus_auth_deinit(void) {
 	janus_mutex_lock(&mutex);
 	if(tokens != NULL)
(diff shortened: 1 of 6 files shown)


lminiero:
Merging.

timsolov:
Hello, I think this is good behavior. I just checked it and it works fine except `signed_tokens` field doesn't save to `janus.plugin.videoroom.jcfg` config file when room is permanent. We need to fix it.

PR #2828, opened 2021-12-03
conf: add exit_on_dl_error option

tmatth:
This makes dlopen or dlsym errors **optionally** fatal, which is especially useful for CI.

### DIFF
--- a/janus.c
+++ b/janus.c
@@ -305,6 +305,9 @@ static uint candidates_timeout = DEFAULT_CANDIDATES_TIMEOUT;
 /* By default we list dependencies details, but some may prefer not to */
 static gboolean hide_dependencies = FALSE;
 
+/* By default we do not exit if a shared library cannot be loaded or is missing an expected symbol */
+static gboolean exit_on_dl_error = FALSE;
+
 /* WebRTC encryption is obviously enabled by default. In the rare cases
  * you want to disable it for debugging purposes, though, you can do
  * that either via command line (-w) or in the main configuration file */
@@ -4298,6 +4301,11 @@ gint main(int argc, char *argv[])
 		server_name = g_strdup(item->value);
 	}
 
+	/* Check if we should exit immediately on dlopen or dlsym errors */
+	item = janus_config_get(config, config_general, janus_config_type_item, "exit_on_dl_error");
+	if(item && item->value && janus_is_true(item->value))
+		exit_on_dl_error = TRUE;
+
 	/* Initialize logger */
 	if(janus_log_init(daemonize, use_stdout, logfile) < 0)
 		exit(1);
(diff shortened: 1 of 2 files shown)


tmatth:
For reference I added this option as I've been burned a few times by plugins, transports, etc. not loading at runtime due to undefined symbols. The latest example was trying to use the websocket transport with libwebsockets built with `LWS_WITHOUT_CLIENT=ON`, which fails at runtime with:
```
Couldn't load event handler plugin 'libjanus_wsevh.so': /opt/app/lib/janus/events/libjanus_wsevh.so: undefined symbol: lws_client_connect_via_info
```
As an aside, does it make sense to set that option in CI here?
https://github.com/meetecho/janus-gateway/blame/1d3f68d094df200d7246a3a349ec50bed1e79985/.github/workflows/janus-ci.yml#L129

janus.c
```diff
@@ -4357,12 +4360,16 @@ gint main(int argc, char *argv[])
 			g_snprintf(eventpath, 1024, "%s/%s", path, eventent->d_name);
 			void *event = dlopen(eventpath, RTLD_NOW | RTLD_GLOBAL);
 			if (!event) {
-				JANUS_LOG(LOG_ERR, "\tCouldn't load logger plugin '%s': %s\n", eventent->d_name, dlerror());
+				JANUS_LOG(exit_on_dl_error ? LOG_FATAL : LOG_ERR, "\tCouldn't load logger plugin '%s': %s\n", eventent->d_name, dlerror());
+				if (exit_on_dl_error)
```
lminiero, line 4364: I don't think this is used, since the property is only read from the config file later.

janus.c
```diff
@@ -4357,12 +4360,16 @@ gint main(int argc, char *argv[])
...
+					exit(1);
 			} else {
 				create_l *create = (create_l*) dlsym(event, "create");
 				const char *dlsym_error = dlerror();
 				if (dlsym_error) {
-					JANUS_LOG(LOG_ERR, "\tCouldn't load symbol 'create': %s\n", dlsym_error);
+					JANUS_LOG(exit_on_dl_error ? LOG_FATAL : LOG_ERR, "\tCouldn't load symbol 'create': %s\n", dlsym_error);
+					if (exit_on_dl_error)
```
lminiero, line 4371: Same here.

lminiero (commented):
This does sound useful, thanks! Added a couple of notes inline. I'll leave it to @atoppi to comment on whether it might be useful in CI, but in that case my guess is we'd need to add a command line argument to that: if this can only be enabled via config file, not sure we can automate that easily.

janus.c
tmatth: Good catch I moved the block that fetches this from the config file earlier.

lminiero:
> As an aside, does it make sense to set that option in CI here?

@tmatth I just realized you were talking of something entirely different, i.e., the libwebsockets option in the CI build. Again, letting @atoppi address that, though :laughing: 

For the rest, this looks good to me, so I'll merge :+1:

atoppi:
@tmatth sorry for being late to the party
Indeed this is an interesting feature and I can see its appliance.
I agree with Lorenzo about the CLI argument, that will be easier for any automation tool in future.

As of
> As an aside, does it make sense to set that option in CI here?
https://github.com/meetecho/janus-gateway/blame/1d3f68d094df200d7246a3a349ec50bed1e79985/.github/workflows/janus-ci.yml#L129

The rationale here is to speed-up the building of lws.
We never actually starts janus in the CI workflow (maybe in future we could add this?) since we only check that compilation completes successfully, hence a symbol failure after `dlopen` can never be triggered.

PR #2829, opened 2021-12-06
AudioBridge Recording dir decoupled from recording setting

rjnpnfigueiredo:
Allows for a permanent Recording Dir setting in the audiobridge plugin so that any participant recording started during actual usage can generate recordings in the defined folder rather than the default folder

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -3223,9 +3223,9 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 			if(audiobridge->record_file) {
 				janus_config_add(config, c, janus_config_item_create("record", "yes"));
 				janus_config_add(config, c, janus_config_item_create("record_file", audiobridge->record_file));
-				if(audiobridge->record_dir)
-					janus_config_add(config, c, janus_config_item_create("record_dir", audiobridge->record_dir));
 			}
+			if(audiobridge->record_dir)
+				janus_config_add(config, c, janus_config_item_create("record_dir", audiobridge->record_dir));
 			if(audiobridge->spatial_audio)
 				janus_config_add(config, c, janus_config_item_create("spatial_audio", "yes"));
 			/* Save modified configuration */
@@ -3386,9 +3386,9 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 			if(audiobridge->record_file) {
 				janus_config_add(config, c, janus_config_item_create("record", "yes"));
 				janus_config_add(config, c, janus_config_item_create("record_file", audiobridge->record_file));
-				if(audiobridge->record_dir)
-					janus_config_add(config, c, janus_config_item_create("record_dir", audiobridge->record_dir));
 			}
+			if(audiobridge->record_dir)
+				janus_config_add(config, c, janus_config_item_create("record_dir", audiobridge->record_dir));
 			if(audiobridge->spatial_audio)
 				janus_config_add(config, c, janus_config_item_create("spatial_audio", "yes"));
 			/* Save modified configuration */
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @rjnpnfigueiredo! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

plugins/janus_audiobridge.c
```diff
@@ -6120,7 +6120,11 @@ static void *janus_audiobridge_handler(void *data) {
 						if(recording_base) {
 							/* Use the filename and path we have been provided */
 							g_snprintf(filename, 255, "%s-audio", recording_base);
-							participant->arc = janus_recorder_create(NULL, "opus", filename);
+							if(audiobridge->record_dir == NULL) {
+								participant->arc = janus_recorder_create(NULL, "opus", filename);
+							} else {
+								participant->arc = janus_recorder_create(audiobridge->record_dir, "opus", filename);
+							}
```
lminiero, line 6127: There's no need for the if: just always pass `audiobridge->record_dir`, which if `NULL` means no dir.

plugins/janus_audiobridge.c
```diff
@@ -6129,7 +6133,11 @@ static void *janus_audiobridge_handler(void *data) {
 							/* Build a filename */
 							g_snprintf(filename, 255, "audiobridge-%s-%s-%"SCNi64"-audio",
 								participant->room->room_id_str, participant->user_id_str, now);
-							participant->arc = janus_recorder_create(NULL, "opus", filename);
+							if(audiobridge->record_dir == NULL) {
+								participant->arc = janus_recorder_create(NULL, "opus", filename);
+							} else {
+								participant->arc = janus_recorder_create(audiobridge->record_dir, "opus", filename);
+							}
```
lminiero, line 6140: Same here.

plugins/janus_audiobridge.c
```diff
@@ -6501,10 +6509,15 @@ static void *janus_audiobridge_handler(void *data) {
...
+						janus_audiobridge_room *audiobridge = participant->room;
 						if(recording_base) {
 							/* Use the filename and path we have been provided */
 							g_snprintf(filename, 255, "%s-audio", recording_base);
-							participant->arc = janus_recorder_create(NULL, "opus", filename);
+							if(audiobridge == NULL || (audiobridge != NULL && audiobridge->record_dir == NULL)) {
+								participant->arc = janus_recorder_create(NULL, "opus", filename);
+							} else {
+								participant->arc = janus_recorder_create(audiobridge->record_dir, "opus", filename);
+							}
```
lminiero, line 6520: Same here.

plugins/janus_audiobridge.c
```diff
@@ -6513,7 +6526,11 @@ static void *janus_audiobridge_handler(void *data) {
 							/* Build a filename */
 							g_snprintf(filename, 255, "audiobridge-%s-%s-%"SCNi64"-audio",
 								participant->room->room_id_str, participant->user_id_str, now);
-							participant->arc = janus_recorder_create(NULL, "opus", filename);
+							if(audiobridge == NULL || (audiobridge != NULL && audiobridge->record_dir == NULL)) {
+								participant->arc = janus_recorder_create(NULL, "opus", filename);
+							} else {
+								participant->arc = janus_recorder_create(audiobridge->record_dir, "opus", filename);
+							}
```
lminiero, line 6533: Same here.

lminiero (commented):
This is an interesting feature, but you're also introducing a potentially breaking change. So far `record_dir` only impacted the recording of `.wav` files, but your patch now makes it act on the individual `.mjr` recordings as well. With this change, it means people cannot save wavs to `/home/path1` and mjrs to `/home/path2`, because if `record_dir` is `/home/path1` then it will be the base directory for all mjrs as well, which I think is wrong.

I'd limit this patch to only allowing the plugin to keep track of `record_dir` even when not recording (yet). Then maybe this can be extended (now or later on) with a different dir property for mjr files to avoid the ambiguity.

rjnpnfigueiredo:
Thanks for the feedback, I'll have another go at the code to include the use case of users that might be expecting audio bridge mjrs to go to the home folder, I had not considered that.

lminiero:
@rjnpnfigueiredo can you check if PR #2862 addresses your original requirement? I added a separate folder for storing MJR files, and I also made sure that, as your PR did, now the recording folder is read independently of whether recording is active or not (for both approaches).

rjnpnfigueiredo:
Hi, thanks for that. If I'm looking at the code correctly, it is also possible to create the AudioBridge with the mjr recording to a specific folder for all participants, which is something that will simplify the intended workflow. This should fulfill the purpose of my PR.

lminiero:
Thanks for the feedback! I'll close this PR then, and address it there. Thanks again for raising this feature to my attention!

PR #2831, opened 2021-12-08
Move from linux/ip.h to netinet/ip.h

gitautas:
Hiya, currently postprocessing/pcap2mjr.c doesn't build on macOS because it depends on a linux specific struct for IP headers, even though netinet/ip.h is already included.

This just moves the struct to one that's available on all UNIX systems.


januscla:
Thanks for your contribution, @gitautas! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks for the fix, just tested on my machine and all still works as expected so I'll merge! :+1:

PR #2835, opened 2021-12-14
Add strlcat variant that uses memccpy for writing SDPs

lminiero:
We recently found out that the process of creating/writing SDPs could be slow, due to how `g_strlcat` works. This is an attempt to speed that part up by adding a `janus_strlcat` variant that uses `memccpy` instead of `g_strlcat`. It seems to be working as expected in some tests when using SDP fuzzing with a huge (~20k mlines) SDP, as it's basically 10x faster:

```
[lminiero@lminiero janus-gateway]$ time ./fuzzers/out/sdp_fuzzer ~/Downloads/giganto.sdp 
Running: /home/lminiero/Downloads/giganto.sdp
Done:    /home/lminiero/Downloads/giganto.sdp: (410058 bytes)

real	0m3.556s
user	0m3.523s
sys	0m0.022s
```

```
[lminiero@lminiero memccpy]$ time ./fuzzers/out/sdp_fuzzer ~/Downloads/giganto.sdp 
Running: /home/lminiero/Downloads/giganto.sdp
Done:    /home/lminiero/Downloads/giganto.sdp: (410058 bytes)

real	0m0.340s
user	0m0.324s
sys	0m0.016s
```

and a simple test with the EchoTest demo worked flawlessly too, but of course we need to ensure this doesn't break anything in production, hence why this is still a PR and not in master yet, so feedback welcome! Once this is confirmed to be ok, I'll update the multistream branch as well where this enhancement would definitely have more of an impact.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -275,22 +275,22 @@ char *janus_string_replace(char *message, const char *old_string, const char *ne
 size_t janus_strlcat(char *dest, const char *src, size_t dest_size) {
 	size_t ret = g_strlcat(dest, src, dest_size);
 	if(ret >= dest_size)
-		JANUS_LOG(LOG_ERR, "janus_strlcat: truncation occurred, %lu >= %lu\n", ret, dest_size);
+		JANUS_LOG(LOG_ERR, "Truncation occurred, %lu >= %lu\n", ret, dest_size);
 	return ret;
 }
 
 int janus_strlcat_fast(char *dest, const char *src, size_t dest_size, size_t *offset) {
 	if(dest == NULL || src == NULL || offset == NULL) {
-		JANUS_LOG(LOG_ERR, "janus_strlcat_fast: invalid arguments\n");
+		JANUS_LOG(LOG_ERR, "Invalid arguments\n");
 		return -1;
 	}
 	if(*offset >= dest_size) {
-		JANUS_LOG(LOG_ERR, "janus_strlcat_fast: overflow\n");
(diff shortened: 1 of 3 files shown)


utils.c
```diff
@@ -279,6 +279,27 @@ size_t janus_strlcat(char *dest, const char *src, size_t dest_size) {
...
+		JANUS_LOG(LOG_ERR, "janus_strlcat_fast: invalid arguments\n");
+		return -1;
+	}
+	if(*offset >= dest_size) {
+		JANUS_LOG(LOG_ERR, "janus_strlcat_fast: overflow\n");
+		return -2;
+	}
+	char *p = memccpy(dest + *offset, src, 0, dest_size - *offset);
+	if(p == NULL) {
+		JANUS_LOG(LOG_ERR, "janus_strlcat: truncation occurred, %lu >= %lu\n",
```
tmatth, line 293: nit: to avoid confusion, probably better to log `"janus_strlcat_fast:...`

lminiero: Ah that's a copy/paste typo, thanks for catching! Incidentally, since this is a `LOG_ERR`, I think the name of the function is actually redundant, since it will appear in the prefix anyway, e.g.:

    [ERR] [utils.c:janus_strlcat_fast:288] janus_strlcat_fast: overflow

so I'll just remove it from both methods.

lminiero:
Merging.

PR #2838, opened 2021-12-18
Janus pp rec add ext json

adnanel:
Setup a "report for extensions" object which is allocated only when the -e flag is supplied.

When the report is not-null, extension information (for now only video rotations and audio level) are supposed to be added in the object.

Finally, the object is converted to JSON and printed out, without any further file processing (similar to -p).

Example usage on my local development machine:
```

root@7c4cd447baff:/tmp# janus-pp-rec -e user4.mjr
{"rotations": [{"ts": 0.0, "rotation": 90}, {"ts": 12.366, "rotation": 180}, {"ts": 20.931000000000001, "rotation": 90}], "audio_level": -1}

```

Which, when prettified, looks like this:
```
{
	"rotations": [{
		"ts": 0.0,
		"rotation": 90
	}, {
		"ts": 12.366,
		"rotation": 180
	}, {
		"ts": 20.931000000000001,
		"rotation": 90
	}],
	"audio_level": -1
}
```

This enables us to apply the video rotations when processing the MJR files.


lminiero:
Thanks! While I see value in this, I'm not sure this is how I'd do it.

First of all, I don't think the individual codec processors need to be modified: they don't need the report, and have nothing to add themselves, so if you're just concerned about the fact that they shouldn't print anything, the application can leave before getting there just as when we just parse or just print the JSON object for instance. Then, I don't think audio levels make sense: video orientation does because it's occasional and useful, but audio levels are sent in ALL packets (~50 per second) and so you'd just end up with a huge useless piece of JSON info you'd have no use of.

I'd personally stick to just rotations as you did in your first commits, at least in this effort: if other things come to mind later on, this can be expanded. But taking into account this extensibility, I wouldn't call this a "print extensions" feature, since it could later on be extended to print other generic info (e.g., video resolution changes). 

That said, per code style all new functions should be prefixed by `janus_pp_`.

adnanel:
Hello

I will apply your suggestions, but I'm unsure about this part:
How should the separate codecs be configured to _not_ print out anything? Perhaps this feature should use a separate log level that's above LOG_INFO?

lminiero:
By returning early, as other properties do. That said, I guess you also refer to the `_preprocess` methods that are called too. You can just set the log level to `LOG_NONE`, then, and use `g_print()` for only printing the JSON output.

adnanel:
Ah, good point, I missed the fact that the codecs don't have to be invoked at all! I will do it promptly.
Thanks

adnanel:
This is WIP, should I close the PR until I'm done?

lminiero:
No need, I converted this to draft: once you think this is done we can turn it back to a regular PR.
Notice the automated builds are failing though.

adnanel:
Yes, my internal build pipeline requires me to push before I can test.

Okay, now it should be fine, what do you think?

postprocessing/janus-pp-rec.1
```diff
@@ -24,6 +24,9 @@ Only print the supported target file extensions per codec  (default=off)
 .BR \-j ", " \-\-json
 Only print JSON header  (default=off)
 .TP
+.BR \-e ", " \-\-ext
+Only print extensions  (default=off)
```
lminiero, line 28: As discussed in the comments, I't call this something like "extended report", rather than "extensions". In the future we may not only have the extension in there.

postprocessing/janus-pp-rec.c
```diff
@@ -54,6 +54,7 @@ Usage: janus-pp-rec [OPTIONS] source.mjr
   -F, --file-extensions         Only print the supported target file extensions
                                   per codec  (default=off)
   -j, --json                    Only print JSON header  (default=off)
+  -e, --ext                     Only print extensions (default=off)
```
lminiero, line 57: Same here.

postprocessing/janus-pp-rec.ggo
```diff
@@ -2,6 +2,7 @@
 usage "janus-pp-rec [OPTIONS] source.mjr [destination.[opus|ogg|mka|wav|webm|mkv|h264|srt]]"
 option "file-extensions" F "Only print the supported target file extensions per codec" flag off
 option "json" j "Only print JSON header" flag off
+option "ext" e "Only print extensions" flag off
```
lminiero, line 5: Same here.

postprocessing/janus-pp-rec.c
```diff
@@ -255,12 +257,18 @@ int main(int argc, char *argv[])
...
+	janus_pp_extension_report *report = NULL;
+
 	if(args_info.json_given)
 		jsonheader_only = TRUE;
 	if(args_info.header_given && !jsonheader_only)
 		header_only = TRUE;
-	if(args_info.parse_given && !jsonheader_only && !header_only)
+	if(args_info.ext_given) {
+		ext_only = TRUE;
+		report = janus_pp_create_ext_report();
```
lminiero, line 269: Not sure why you're creating this here, considering it will only be used much later. Besides, there are points in the code where we might exit sooner (e.g., in case of errors), so this would leak.

postprocessing/janus-pp-rec.c
```diff
@@ -811,6 +823,7 @@ int main(int argc, char *argv[])
 	int audiolevel = 0, rotation = 0, last_rotation = -1, rotated = -1;
 	uint16_t rtp_header_len, rtp_read_n;
 	/* Start loop */
+
```
lminiero, line 826: I don't understand why you added this empty line here.

postprocessing/janus-pp-rec.c
```diff
@@ -949,8 +962,9 @@ int main(int argc, char *argv[])
 			} else {
 				rtp_header_len += rtp_read_n;
 			}
-			if(audio_level_extmap_id > 0)
+			if(audio_level_extmap_id > 0) {
 				janus_pp_rtp_header_extension_parse_audio_level(prebuffer, len, audio_level_extmap_id, &audiolevel);
+			}
```
lminiero, line 967: I don't understand why you added these brackets here. Please only make changes relevant to your contribution.

postprocessing/janus-pp-rec.c
```diff
@@ -1152,12 +1166,14 @@ int main(int argc, char *argv[])
 		offset += len;
 		count++;
 	}
+
```
lminiero, line 1169: Same here: unneeded empty line, and irrelevant.

postprocessing/janus-pp-rec.c
```diff
@@ -1171,15 +1187,28 @@ int main(int argc, char *argv[])
...
-		if(rotated == 0 && last_rotation != 0) {
-			JANUS_LOG(LOG_INFO, "The video is rotated\n");
-		} else if(rotated > 0) {
-			JANUS_LOG(LOG_INFO, "The video changed orientation %d times\n", rotated);
+	if(!ext_only) {
+		JANUS_LOG(LOG_INFO, "Counted %"SCNu32" frame packets\n", count);
+		if (rotated != -1) {
+			if (rotated == 0 && last_rotation != 0) {
+				JANUS_LOG(LOG_INFO, "The video is rotated\n");
+			} else if (rotated > 0) {
```
lminiero, line 1195: You changed the code style when you moved the code, please don't do that. Per our code style, it's `if(` and not `if (` (no space).

postprocessing/janus-pp-rec.c
```diff
@@ -1171,15 +1187,28 @@ int main(int argc, char *argv[])
...
+				JANUS_LOG(LOG_INFO, "The video changed orientation %d times\n", rotated);
+			}
 		}
 	}
 
+	if (ext_only) {
+		/* Look for rotations changes */
+		janus_pp_detect_rotation_changes(report, list);
```
lminiero, line 1203: I don't understand this. You're looping on the whole list again just to populate the report. Why not populate it _while_ we're parsing it, since we know the extensions? Is it to account for potentially out of order packets?

postprocessing/janus-pp-rec.c
```diff
@@ -1171,15 +1187,28 @@ int main(int argc, char *argv[])
...
+
+
```
lminiero, line 1205: Please remove these empty lines.

postprocessing/pp-ext-report.c
```diff
@@ -0,0 +1,87 @@
+#include "pp-ext-report.h"
+
+#include <jansson.h>
+
+#include "stdlib.h"
+#include "glib.h"
```
lminiero, line 6:
```
#include <stdlib.h>
#include <glib.h>
```

postprocessing/pp-ext-report.c
```diff
@@ -0,0 +1,87 @@
...
+
+#include "../debug.h"
+
+void janus_pp_print_ext_report(janus_pp_extension_report* report) {
+	json_t *obj = json_object();
+
+
```
lminiero, line 13: Two empty lines are way too much, please only add one.

postprocessing/pp-ext-report.c
```diff
@@ -0,0 +1,87 @@
...
+	janus_pp_extension_report_rotation *rot = report->rotations;
+
+	/* add rotations to json */
```
lminiero, line 16: Code style: comments should start with a capitalized letter.

postprocessing/pp-ext-report.c
```diff
@@ -0,0 +1,87 @@
...
+	JANUS_PRINT("%s\n", str);
+	free(str);
+
+	json_decref(obj);
+}
+
+void janus_pp_free_ext_report(janus_pp_extension_report* report) {
+	if(!report) {
+		return;
+	}
```
lminiero, line 39: No need for brackets, action is simple.

postprocessing/pp-ext-report.c
```diff
@@ -0,0 +1,87 @@
...
+
+	janus_pp_extension_report_rotation *current = report->rotations;
+	while(current) {
+		janus_pp_extension_report_rotation* next = current->next;
```
lminiero, line 43: Code style: asterisk of pointer should go next to the variable, not the type.

postprocessing/pp-ext-report.h
```diff
@@ -0,0 +1,26 @@
...
+#define JANUS_GATEWAY_PP_EXT_REPORT_H
+
+#include <inttypes.h>
+
+#include "pp-rtp.h"
+
+typedef struct janus_pp_extension_report_rotation {
+	int rotation;
+	double timestamp;
+	struct janus_pp_extension_report_rotation* next;
```
lminiero, line 11: This seems overkill. You can define `janus_pp_extension_report_rotation` without that and just rely on GLib's `GList` or `GSlist` for the job.

lminiero (commented):
Added a few notes inline, thanks!

postprocessing/pp-ext-report.h
```diff
@@ -0,0 +1,26 @@
...
+} janus_pp_extension_report_rotation;
+
+typedef struct janus_pp_extension_report {
+	struct janus_pp_extension_report_rotation* rotations;
+} janus_pp_extension_report;
```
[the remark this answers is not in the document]
adnanel: i have added the type to keep it extensible, but i will remove it for now.

adnanel:
Hi, i think i catched all your remarks and have applied them, sorry for the many mistakes. i have yet to get used to this codestyle, i will put more effort honoring it in the future.

adnanel:
@lminiero any chance to get an ETA on this?

lminiero:
I'm still on vacation, maybe Monday. Apologies for the delay.

lminiero:
@adnanel apologies for the late feedback: after having a new look at your changes, I came up with a slightly different approach, that I've made in the PR you can see referenced above. Please let me know if it makes sense to you.

lminiero:
Closing as I merged that other PR. Thanks for the suggestion and the first draft!

PR #2843, opened 2021-12-28
fix saving signed_tokens field when room is permanent

timsolov:
#2825
fix saving room's configuration to config file when room is permanent.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -3445,6 +3445,8 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 				janus_config_add(config, c, janus_config_item_create("is_private", "yes"));
 			if(videoroom->require_pvtid)
 				janus_config_add(config, c, janus_config_item_create("require_pvtid", "yes"));
+			if(videoroom->signed_tokens)
+				janus_config_add(config, c, janus_config_item_create("signed_tokens", "yes"));
 			if(videoroom->require_e2ee)
 				janus_config_add(config, c, janus_config_item_create("require_e2ee", "yes"));
 			g_snprintf(value, BUFSIZ, "%"SCNu32, videoroom->bitrate);
@@ -3629,6 +3631,8 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 				janus_config_add(config, c, janus_config_item_create("is_private", "yes"));
 			if(videoroom->require_pvtid)
 				janus_config_add(config, c, janus_config_item_create("require_pvtid", "yes"));
+			if(videoroom->signed_tokens)
+				janus_config_add(config, c, janus_config_item_create("signed_tokens", "yes"));
 			if(videoroom->require_e2ee)
 				janus_config_add(config, c, janus_config_item_create("require_e2ee", "yes"));
 			g_snprintf(value, BUFSIZ, "%"SCNu32, videoroom->bitrate);


lminiero:
This patch goes beyond just saving the property to the config file, when saving rooms permanently. It also adds an option to "edit" to change the behaviour dynamically within a room. Please only do the former in this patch: the latter is something that is not there yet, and I'm not sure it should be, and so needs a separate discussion.

timsolov:
@lminiero updated

plugins/janus_videoroom.c
```diff
@@ -3445,6 +3445,8 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 				janus_config_add(config, c, janus_config_item_create("is_private", "yes"));
 			if(videoroom->require_pvtid)
 				janus_config_add(config, c, janus_config_item_create("require_pvtid", "yes"));
+			if(videoroom->signed_tokens)
+				janus_config_add(config, c, janus_config_item_create("signed_tokens", "yes"));
```
lminiero, line 3449: Shouldn't this be in the "edit" part too? Even if "edit" doesn't expose a way to modify that property, if you don't have lines to save the value there, a permanent "edit" will result in not saving the `signed_tokens` property, which would be lost as a consequence.

timsolov: Yeh, you're right. I forgot about `edit`.

timsolov: @lminiero done

plugins/janus_videoroom.c
```diff
@@ -2328,11 +2328,14 @@ int janus_videoroom_init(janus_callbacks *callback, const char *config_path) {
 			videoroom->is_private = priv && priv->value && janus_is_true(priv->value);
 			videoroom->require_pvtid = req_pvtid && req_pvtid->value && janus_is_true(req_pvtid->value);
 			if(signed_tokens && signed_tokens->value && janus_is_true(signed_tokens->value)) {
+				JANUS_LOG(LOG_INFO, "Room id %s signed_tokens value: %s\n", videoroom->room_id_str, signed_tokens->value);
```
lminiero, line 2331: This is new, please remove.

plugins/janus_videoroom.c
```diff
@@ -2328,11 +2328,14 @@ int janus_videoroom_init(janus_callbacks *callback, const char *config_path) {
...
 				if(!gateway->auth_is_signed()) {
 					JANUS_LOG(LOG_WARN, "Can't enforce signed tokens for this room, signed-mode not in use in the core\n");
 				} else {
 					videoroom->signed_tokens = TRUE;
 				}
+			} else {
+				JANUS_LOG(LOG_INFO, "Room id %s signed_tokens value: false\n", videoroom->room_id_str);
```
lminiero, line 2338: Please remove.

plugins/janus_videoroom.c
```diff
@@ -3005,6 +3008,7 @@ static int janus_videoroom_access_room(json_t *root, gboolean check_modify, gboo
 		}
 	}
 	if(check_join) {
+		JANUS_LOG(LOG_INFO, "check_join = true (%s)\n", room_id_str);
```
lminiero, line 3011: This looks like a debug line that was left there?

plugins/janus_videoroom.c
```diff
@@ -3024,6 +3028,8 @@ static int janus_videoroom_access_room(json_t *root, gboolean check_modify, gboo
 			g_strlcpy(error_cause, error_cause2, error_cause_size);
 			return error_code;
 		}
+	} else {
+		JANUS_LOG(LOG_INFO, "check_join = false (%s)\n", room_id_str);
```
lminiero, line 3032: See above.

lminiero (commented):
Thanks for the quick fixes! But I see some new lines were added that weren't there before: please only keep the permanent saving of the property.

plugins/janus_videoroom.c
timsolov: removed

plugins/janus_videoroom.c
timsolov: removed

plugins/janus_videoroom.c
timsolov: removed

lminiero:
Thanks, merging then!

PR #2849, opened 2022-01-07
Handle sdp in 180 sip message same as 183

lejlasolak:
183 is not always with SDP and 180 is not always without SDP. 
In general, 180 with SDP is used for the distinctive ring tone and 183 with SDP for announcement. 
We need to treat 180 and 183 the same and only make a difference whether there is SDP (early media) or not.

Nice writeup here: https://thanhloi2603.wordpress.com/2017/08/05/sip-180-vs-183-vs-early-media/

According to https://datatracker.ietf.org/doc/html/rfc3261#section-21.1.5:
“The Reason-Phrase, header fields, or message body MAY be used to convey more details about the call progress.”

This RFC sums up how to handle 180 (and 183): https://datatracker.ietf.org/doc/html/rfc3960

### DIFF
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -5604,8 +5604,8 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 
 			gboolean in_progress = FALSE;
 			if(status < 200) {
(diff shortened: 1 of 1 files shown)


lminiero:
Makes sense to me, and it looks good to merge, thanks! :+1:

PR #2836, opened 2021-12-16
Ext orientation video recording sip

adnanel:
Similar to what was done in https://github.com/meetecho/janus-gateway/pull/2527

We add the orientation extension id in the MJR header so it can be read & applied when processing the video afterwards.

PR #2852, opened 2022-01-07
Multistream add orient extmap

adnanel:
Similar to what was done in #2836 and #2527.

For videoroom on multistream branch, we add the orientation extension id in the MJR header so it can be read & applied when processing the video afterwards.


lminiero:
Looks good, merging :+1:

PR #2858, opened 2022-01-21
Added option to print extended header in janus-pp-rec (inspired by #2838)

lminiero:
This PR is inspired by the excellent work @adnanel did in #2838, but using a slightly different approach, and broadening the scope a bit. Specifically, it adds a new command line option to `janus-pp-rec` (`-e` or `--extended-json`) which, if set, only prints the JSON header, but enriched with information it parses from the MJR file itself. An example (with artificial intendation, since the tool prints compact JSON) is the following:

```
{
   "t": "v",
   "c": "vp8",
   "x": {
      "13": "urn:3gpp:video-orientation"
   },
   "s": 1612369719662379,
   "u": 1612369719904395,
   "extended": {
      "packets": 1427,
      "duration": 29.850000000000001,
      "rotations": [
         {
            "ts": 0.0,
            "rotation": 0
         }
      ],
      "codec": {
         "resolution": [
            {
               "ts": 0.0,
               "width": 480,
               "height": 270
            },
            {
               "ts": 13.85,
               "width": 720,
               "height": 405
            },
            {
               "ts": 17.850000000000001,
               "width": 960,
               "height": 540
            },
            {
               "ts": 25.901,
               "width": 1440,
               "height": 810
            }
         ]
      }
   }
}
```
where as you can see, besides the "usual" fields already present in the header, there's a new `extended` object that contains additional information. At the moment, the information I've added are the number of packets in the MJR file, the estimated duration of the recording, the rotations we detected (which was the main purpose of #2838 itself) and, for video files, the resolution changes we detected as well. In the future we may add other information, here, depending on what people think might be useful.

Please let me know what you think about this, as I plan to merge soon.

### DIFF
--- a/postprocessing/janus-pp-rec.1
+++ b/postprocessing/janus-pp-rec.1
@@ -30,6 +30,9 @@ Only parse .mjr header  (default=off)
 .BR \-p ", " \-\-parse
 Only parse and re-order packets  (default=off)
 .TP
+.BR \-e ", " \-\-extended-json
+Only print extended JSON report (automatically enables --json)  (default=off)
+.TP
 .BR \-m ", " \-\-metadata=metadata
 Save this metadata string in the target file
 .TP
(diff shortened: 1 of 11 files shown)


adnanel:
So basically this would cover all use-cases for `-j` and provide even more information. Perhaps think about deprecating that?

In any case, in our specific case this would actually merge 2 janus-pp-ext (one for `-j` and the other being the one I proposed) invocations into one, so I'm all for it.

Thanks for the addition!

lminiero:
I don't think `-j` needs to be deprecated, since that one stops much sooner and returns the exact header that is in the file, so if all you need is there (e.g., codec) then there's no need to waste CPU cycles for the extended report. Thanks for the feedback!

lminiero:
Merging.

PR #2860, opened 2022-01-23
Make rtp port range variables static in ice to prevent clashes with plugins

jimmymic:
Our custom plugin was not using a storage class specifier on rtp_range_min and rtp_range_max. This led to a clash with the ice handling code which is also not using a storage class specifier, hence defaulting to extern.

This change is to explicitly add a static specifier hence removing potential clashes for other plugin developers.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -332,8 +332,8 @@ int janus_ice_get_event_stats_period(void) {
 
 
 /* RTP/RTCP port range */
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @jimmymic! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

PR #2862, opened 2022-01-24
Add new API to bulk start/stop MJR-based recordings in AudioBridge

lminiero:
As the title says, this adds a feature that was missing in the AudioBridge, namely the ability to start/stop the recording of all participants to MJR files as we do have in the VideoRoom. In fact, while AudioBridge does have an `enable_recording` request just as the VideoRoom does, it uses that request to trigger the recorded mix, which is the default mechanism for recording a room in the AudioBridge: since the AudioBridge plugin can also be instructed to save the contributions of each participant to a separate MJR file as well, though (e.g., for external mixing and processing), then a request to enable/disable it for all participants in a room at the same time was helpful as well.

The syntax for the new request is basically the same as the existing `enable_recording`, with the difference that it is named differently, and references a different variable for the target folder. In fact, you can specify where MJR files should be saved by default by specifying an `mjrs_dir` when creating the room; the folder can be changed (for new recordings, not old or ongoing ones) using `edit` as well, or in the request that follows. Then, you can start/stop a bulk recording to MJR using the following request:

```
{
	"request" : "enable_mjrs",
	"secret" : "<room secret; mandatory if configured>"
	"mjrs" : <true|false, whether all participants in the room should be individually recorded to mjr files or not>,
	"mjrs_dir" : "<path where all MJR files should be saved to (optional)>"
}
```

I plan to merge soon, so feedback welcome!

### DIFF
--- a/conf/janus.plugin.audiobridge.jcfg.sample
+++ b/conf/janus.plugin.audiobridge.jcfg.sample
@@ -16,6 +16,8 @@
 # record = true|false (whether this room should be recorded, default=false)
 # record_file = "/path/to/recording.wav" (where to save the recording)
 # record_dir = "/path/to/" (path to save the recording to, makes record_file a relative path if provided)
+# mjrs = true|false (whether all participants in the room should be individually recorded to mjr files, default=false)
+# mjrs_dir = "/path/to/" (path to save the mjr files to)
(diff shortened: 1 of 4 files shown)


lminiero:
Merging :v:

Issue #2840, opened 2021-12-20
Janus crashes with latest release


dnish:
Hey,
since the last updates, Janus crashes every 10 minutes on our main server (Ubuntu 20.04).

Version: 0.11.7 and Master

Error: 
```Dec 20 20:10:36 ns541136 kernel: [983376.078081] hloop 475389242[5625]: segfault at f8 ip 00005619fda729e8 sp 00007f0d5f6f5750 error 4 in janus[5619fda24000+b8000]```.

We are using the video call plugin (not the video room plugin).

lminiero:
That's not enough information to debug. Please check the debugging instructions in the docs to collect something we can look at.

dnish:
Yeah, we are having a .crash file in ```/var/crash/_opt_janus_bin_janus.0.crash```. But when using <code>gdb /opt/janus/bin/janus /var/crash/_opt_janus_bin_janus.0.crash</code> I get the following error:


> Reading symbols from /opt/janus/bin/janus...done.
> "/var/crash/_opt_janus_bin_janus.0.crash" is not a core dump: File format not recognized

direritcom:
Hello, have you found a solution to the problem?
We have exactly the same situation on Ubuntu 20.04.
We can't fix the error so far to make a dump.

> Hey, since the last updates, Janus crashes every 10 minutes on our main server (Ubuntu 20.04).
> 
> Version: 0.11.7 and Master
> 
> Error: `Dec 20 20:10:36 ns541136 kernel: [983376.078081] hloop 475389242[5625]: segfault at f8 ip 00005619fda729e8 sp 00007f0d5f6f5750 error 4 in janus[5619fda24000+b8000]`.
> 
> We are using the video call plugin (not the video room plugin).

lminiero:
There's literally _nothing_ we can do until you share a meaningful dump as I've said my previous message.

dnish:
Hey,#
nope, I'm still waiting for an answer from @lminiero how I can read the crash file.

lminiero:
I have no idea how you can open that dump: I don't know your system. Please do some more research on your own.

dnish:
> I have no idea how you can open that dump: I don't know your system. Please do some more research on your own.

Well, it is written in the issue (Ubuntu 20.04). I've already searched for the gdb error and didn't find anything that fixed the issue.

direritcom:
Didn't you have this problem in this version?

> Meanwhile you can revert back to 0.11.6.

[...]

PR #2869, opened 2022-02-02
Update RTP extensions for outgoing packets in the PeerConnection loop (fixes #2867)

### LINKED ISSUE #2867: [0.x] Janus crashes (heap-use-after-free)
0.11.8

Master of Git

**Was this working before?**
We couldn't detect the issue on version 0.11.6. The regularly crashing started with version 0.11.7.

https://pastebin.com/NMbpKEj7

lminiero:
This PR should fix the issue reported in #2867, which as explained in https://github.com/meetecho/janus-gateway/issues/2840#issuecomment-1026129348 was apparently caused by a race condition between `janus_ice_webrtc_free` and a plugin calling `relay_rtp`: specifically, the crash was caused by an attempt to access one of the stream `mid` attributes to update the RTP extensions on the way out, which we need to do in the core.

As I discussed in https://github.com/meetecho/janus-gateway/issues/2867#issuecomment-1027041948, one of the potential solutions was to move the code that prunes/updates/sets RTP extensions on the way out in a function that is part of the main PeerConnection loop instead, which would ensure no conflict occurs when accessing the fields (since that loop is also the one responsible for incoming traffic, the PC life cycle, and getting rid of its resources).

The impact should be neglectible: actually, while we may be doing some small reallocs when updating extensions, we're now also not iterating on a couple of extensions we always only set when sending packets (abs-send-time, twcc), which means we're doing less work there in this patch (those extensions are filled in right away).

I tested this briefly with a few plugins and it seems to be working as expected, but please let me know if you encounter any regression, especially if you were relying on extensions for anything.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -374,6 +374,7 @@ typedef struct janus_ice_queued_packet {
 	char *data;
 	char *label;
 	char *protocol;
+	janus_plugin_rtp_extensions extensions;
 	gint length;
 	gint type;
 	gboolean control;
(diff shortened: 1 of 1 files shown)


atoppi (commented):
lgtm :+1:

dnish:
@lminiero I've updated our version to the latest rebased version of your branch. The server is running now more stable, but after 4 hours, we got the next crash:

https://pastebin.com/r7nDHqWJ

lminiero:
Just rebased the PR with your fix. Fingers crossed that we're done now! :crossed_fingers:

dnish:
Hey,
thank you for the update. Just synced the repo, now I get the following crash after 2 hours:

https://pastebin.com/nthih03H

tmatth:
> Hey, thank you for the update. Just synced the repo, now I get the following crash after 2 hours:
> 
> https://pastebin.com/nthih03H

This one's a little less straightforward, @lminiero would it make sense to increment the ice_handle reference count for the duration of this function (as we're checking it throughout?).

lminiero:
I was going to say that I don't think they're using the latest code, as it's crashing on a point that was fixed already, but if it's a race condition, it may be that it's triggering even in between a check and its action. Not sure refcounting is the way to go here, as if we're processing an SDP while the session is being torn down we're still going to end up with weird sessions. Maybe we should just put the whole method in the handle mutex: there shouldn't be risks of deadlocks since we're already using that mutex in a couple of parts.

lminiero:
For sure I'm wondering what the heck @dnish is doing with a Janus instance to cause all those race conditions to appear in the first place :stuck_out_tongue_closed_eyes:

dnish:
> I was going to say that I don't think they're using the latest code, as it's crashing on a point that was fixed already, but if it's a race condition, it may be that it's triggering even in between a check and its action. Not sure refcounting is the way to go here, as if we're processing an SDP while the session is being torn down we're still going to end up with weird sessions. Maybe we should just put the whole method in the handle mutex: there shouldn't be risks of deadlocks since we're already using that mutex in a couple of parts.

I can confirm that we did a <code>git checkout rtpext-loop</code> and were on commit <code>2a0066b20dc26c9221a61708b6c7ae2329d67da2</code> before I did a reinstall with <code>make clean</code> and the option mentioned in the debug section of the docs.

dnish:
> For sure I'm wondering what the heck @dnish is doing with a Janus instance to cause all those race conditions to appear in the first place 😝

We are running a video chat instance with about 150-200 concurrent online users. So I guess the amount of users causes this.

lminiero:
I doubt the number of users has an impact: our virtual event platform runs way more than that concurrently for instance. I suspect it's simply the pattern that is less explored, since from what I can see in the logs you're using the VideoCall plugin that is a less commonly used plugin that, e.g., the VideoRoom. With the VideoCall it's probably easier to end up in some scenarios like that (caller or callee hanging up has a direct impact on the other), since fewer people use it and so there have been less chances to fix hidden issues.

That said, I don't see how 0.11.6 can be more stable, since we addressed the differences (new negotiated properties), and 0.11.6 uses the exact same approach for dealing with SDPs from plugins. It's more likely a matter of chance.

dnish:
@lminiero Yeah I understand, I have also no explanation for this. But 0.11.6 is really running without any issues. I've checked it over the last week. Using a commit after that release always ends with a crash after 2-3 hours, on 0.11.6 we can't notice any crashes even after 2 days.

lminiero:
I'll think of a way to address the last crash you reported. In case I think it requires some particular change, I'll merge this PR first (since it's not related to the problem it was created to fix, and helps already), and prepare a different one. This is more likely to happen next week, though, as it's almost EOD here.

dnish:
> I'll think of a way to address the last crash you reported. In case I think it requires some particular change, I'll merge this PR first (since it's not related to the problem it was created to fix, and helps already), and prepare a different one. This is more likely to happen next week, though, as it's almost EOD here.

Thank you very much. No problem. Meanwhile I can check again if 0.11.6 is really running without any crashes over that time. Maybe someone has an idea what has changed after that release that could cause this issue.

dnish:
So the instance is running for over a day now without any crashes. I don't think that it is a coincidence that 0.11.7 crashes after 1-2 hours, while 0.11.6 is running for days. I'll now upgrade commit by commit to check at which point the crashes begin.

lminiero:
I'll merge this PR in the meanwhile, since it's unrelated to the new problem. Please open a new issue with info on that other crash.

PR #2872, opened 2022-02-03
janus: add yet another missing NULL check for opusred

tmatth:
This is analogous to the fixes in cba15bb and a832c5d.


lminiero:
Thanks, merging! I'll rebase the PR accordingly.

PR #2875, opened 2022-02-08
fix PCMA/PCMU RTP forwarding in audiobridge - incorrect RTP header of…

AlexYaremchuk:
Fix: audiobridge plugin creates incorrect RTP packets during forwarding of PCMA/PCMU. 12 header bytes offset was moved forward and intersects with RTP payload . Wireshark shows [Malformed Packet] for such packets.


januscla:
Thanks for your contribution, @AlexYaremchuk! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Good catch, the +12 is where the buffer is. Merging, thanks!

PR #2877, opened 2022-02-09
Unlock mutex before goto

Bug-Fairy:
Just a missing mutex unlock before goto.


lminiero:
Good catch! Just let me double check if we have the same issue in master, as in case I'd need to apply the fix there first, and then re-align multistream.

PR #2880, opened 2022-02-10
Add opus DTX detection to the post-processor

atoppi:
This PR adds to the post-processor the missing code to handle [opus DTX](https://datatracker.ietf.org/doc/html/rfc6716#section-2.1.9) detection.
To enable the feature please use the already existing `-X` / `--dtx` flag.
Notice that when DTX detection is enabled the silence suppression detection will be disabled regardless of the actual audio codec (this also was the previous behavior that I did not want to change).


atoppi:
Closing since the [commit f51524eb](https://github.com/meetecho/janus-gateway/commit/f51524eb9ed0b5302a65bce05d8ee64a2af3e790)  has the same effect on post processing.

PR #2883, opened 2022-02-14
Repaired link to issue templates

tacerus:


lminiero:
Good catch, thanks: this changed when we started using the issue template generator in GitHub itself, and I didn't remember we had a link in the README. Merging :+1:

lminiero:
Actually, looks like we should link to this file instead:

https://github.com/meetecho/janus-gateway/blob/master/.github/CONTRIBUTING.md

since the issue template is now a folder with little human readable info. I'll change this shortly.

tacerus:
That'd make more sense indeed. I can change it.

tacerus:
Ah, you already did. Very nice! :)

PR #2886, opened 2022-02-14
Link to Duktape engine as a library

lminiero:
When we first added support for Duktape in #1264, we put the whole Duktape engine in the `duktape-deps` folder, so that it could be built and embedded directly in the target Janus Duktape plugin. This works, but it takes longer to compile (not much, but still longer), and it shows a ton of warnings we can do nothing about due to our compilation flags. This PR changes that so that we now look for the Duktape engine as a library instead, which should be available on most, if not all, repos (it definitely is on Fedora and Ubuntu, for instance). The module and console extensions are still in `duktape-deps` instead, since they're indeed extensions to the main engine (and we customuzed the console logger to make it fit our logging).

If you're using the Duktape plugin, make sure to test this change to check whether or not it breaks anything for you. Notice that this is a PR based on the `src-folder` (#2885) branch, not master.


lminiero:
Pinging @voicenter in particular as I remember they were using the Duktape engine some time ago.

lminiero:
Merging.

PR #2887, opened 2022-02-15
Remove distinction between simulcast and simulcast2 in janus.js

lminiero:
For a long time `janus.js` has supported two different ways to enable simulcast:

1. `simulcast: true`, which on Chrome enabled simulcast using the old SDP munging approach, and on Firefox its custom rid approach;
2. `simulcast2: true`, which on Chrome enabled simulcast using the standard rid approach.

This was initially done to account for the different ways simulcast could be enabled in browsers, as explained in this [blog post](https://www.meetecho.com/blog/simulcast-janus-ssrc/) I wrote at the time.

This patch gets rid of this distinction, and assumes `simulcast: true` is all you need to enable simulcast in a web page: `janus.js` then automatically enables the right mechanism depending on the browser (munging on very old versions of Chrome, custom-rid for Firefox, standard-rid otherwise). `simulcast2: true` is still accepted for backwards compatibility, but it's considered an alias of `simulcast: true` internally. Notice there's no way to force simulcast via SDP munging now: I don't see a reason for having it, since `janus.js` is only useful if you're talking to Janus, and Janus is perfectly capable of handling both mechanisms just fine.

Planning to merge soon, so feedback welcome before I hit the button.

### DIFF
--- a/html/canvas.js
+++ b/html/canvas.js
@@ -60,7 +60,6 @@ var audioenabled = false;
 var videoenabled = false;
 
 var doSimulcast = (getQueryStringValue("simulcast") === "yes" || getQueryStringValue("simulcast") === "true");
-var doSimulcast2 = (getQueryStringValue("simulcast2") === "yes" || getQueryStringValue("simulcast2") === "true");
 var acodec = (getQueryStringValue("acodec") !== "" ? getQueryStringValue("acodec") : null);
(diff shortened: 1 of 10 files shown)


lminiero:
Adding the multistream label since that's where I've made the change, but I'll backport to the `0.x` branch as well.

PR #2888, opened 2022-02-15
Keep track of when simulcast substreams are disabled in SDP (fixes #2879)

### LINKED ISSUE #2879: [1.x] Simulcast inactive layer (~ syntax) is not being recognized by Janus and ~ is not returned in the answer
Simulcast supports sending inactive layer through addTransceiver method. 
```
[
    { rid: 'h', active: false, maxBitrate: bitrates.high },
    ....................
]
```

This is translated to the tilde syntax (~) in the SDP: **a=simulcast:send ~h;m;l**
Currently Janus SDP answer is: **a=simulcast:recv h;m;l**, which unfortunately reenables the simulcast layer.

Expected behaviour of the answer should be: **a=simulcast:recv ~h;m;l**

Currently the workaround is to hack the SDP answer locally in the browser to add the ~ sign.
Other hack is to call **.active = false** on the layer again after the answer.

The tilde (~) syntax is explained in https://datatracker.ietf.org/doc/html/rfc8853

lminiero:
As the title says, see #2879 for details.


lminiero:
Merging.

PR #2890, opened 2022-02-17
Fix error message when creating a session

rcmuniz1994:
On the createSession function, the error callback is called like this:

```js
  callbacks.error(textStatus + ": " + errorThrown);
```

But errorThrown isn't a string. So if I try do something like:

```js
 new Janus({
          server: 'http://localhost:8088/janus',
          iceServers: [
            {
              urls: 'stun:stun.l.google.com:19302',
            },
          ],
          success: () => {
              ...
          },
          error: (error) => {
            console.warn(error);
          },
        });
      },
    });

```

If the connection fails for any reason, the message could be something like this:

`Probably a network error, is the server down?: [object Object]`

I believe it happens on another parts of the code. But I couldn't  test them yet.

PS: Sorry about my rusty English.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -968,7 +968,7 @@ function Janus(gatewayCallbacks) {
 				if(errorThrown === "")
 					callbacks.error(textStatus + ": Is the server down?");
 				else if (errorThrown && errorThrown.error)
-					callbacks.error(textStatus + ": " + errorThrown.error.stack);
+					callbacks.error(textStatus + ": " + errorThrown.error.message);
 				else
 					callbacks.error(textStatus + ": " + errorThrown);
 			}
--- a/html/janus.js
+++ b/html/janus.js
@@ -967,8 +967,10 @@ function Janus(gatewayCallbacks) {
 				}
 				if(errorThrown === "")
 					callbacks.error(textStatus + ": Is the server down?");
-				else
+				else if (errorThrown && errorThrown.error)
 					callbacks.error(textStatus + ": " + errorThrown.error.stack);
+				else
+					callbacks.error(textStatus + ": " + errorThrown);
 			}
 		});
 	}
--- a/html/janus.js
+++ b/html/janus.js
@@ -968,7 +968,7 @@ function Janus(gatewayCallbacks) {
 				if(errorThrown === "")
 					callbacks.error(textStatus + ": Is the server down?");
 				else
-					callbacks.error(textStatus + ": " + errorThrown);
+					callbacks.error(textStatus + ": " + errorThrown.error.stack);
 			}
 		});
 	}


januscla:
Thanks for your contribution, @rcmuniz1994! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

html/janus.js
```diff
@@ -968,7 +968,7 @@ function Janus(gatewayCallbacks) {
 				if(errorThrown === "")
 					callbacks.error(textStatus + ": Is the server down?");
 				else
-					callbacks.error(textStatus + ": " + errorThrown);
+					callbacks.error(textStatus + ": " + errorThrown.error.stack);
```
lminiero, line 971: Is this guaranteed to always exist? There are different implementations of `Janus.httpAPICall` in `janus.js`, and people can provide their own: while the default (using `Janus.useDefaultDependencies`) uses `fetch`, there's a version (using `Janus.useOldDependencies`) that uses jQuery's XHR stack instead.

From how I see it, in case `errorThrown.error` is `null` or `undefined`, this will cause an error while trying to report an actual problem.

lminiero: @rcmuniz1994 ping :slightly_smiling_face:

rcmuniz1994: @lminiero Hi! I see. In case errorThrown.error is null or undefined I think something like this will avoid the error:

```js
...
else if (errorThrown && errorThrown.error) {
   callbacks.error(textStatus + ": " + errorThrown.error.stack);
}
```

I'm going to test this solution and I'll try another way in case this one cause an error too.

rcmuniz1994: I checked the Janus.useDefaultDependencies and Janus.useOldDependencies. Now I totally get what you said, there are many error implementations.

lminiero: Thanks! I'll have a look tomorrow :+1:

lminiero:
I just checked but I think we should return `error.message`, not `error.stack`: first of all, because apparently `stack` is an empty string on Firefox; and besides, it prints a full stack on Chrome which makes for a very ugly dialog message. IMO `error.message` looks like a more compact and readable reason to return for display purposes: `error.stack` can be printed as an error on the log, in case.

rcmuniz1994:
Make sense. I'll have a look and test it. Thanks :smile:

lminiero:
Hi @rcmuniz1994, did you have a chance to look into that? Thanks!

rcmuniz1994:
Hi @lminiero, I'm going to check it this week. Thanks. XD

rcmuniz1994:
Hi, @lminiero ! I check what you suggested and it does really look more friendly. So I made the change. Could you check it, please?

lminiero:
Thanks, much better now! Merging :partying_face:

lminiero:
I also backported this to the `janus.js` in the `0.x` branch.

Issue #2896, opened 2022-02-21
[1.x] Protection fault 


amnonbb:
**What version of Janus is this happening on?**
Latest git ("commit-hash": "e7fec8294704b5f034dd03546b85bb201724d496")

**Was this working before?**
It's actually was worked but yesterday it's happend like boom, every 10-15 minutes crash on every server. I did change on client, but nothing that touch to streaming plugin

**Is there a gdb or libasan trace of the issue?**
https://pastebin.com/AFyNUyHy

atoppi:
hi @amnonbb it seems glib is reporting a failed assertion, not clear which one since the symbols for glib are not enabled.
Is there something on the janus log about the assertion?
How your client is interacting with the streaming plugin? Any chance of "risky" concurrent tasks (like i.e. mountpoint destruction while in use)?

atoppi:
My guess is that the `session->streams_byid` hashtable is being un-reffed and freed by another thread through `janus_streaming_hangup_media_internal`.

You're probably using and old version of glib that [includes an assertion ](https://github.com/GNOME/glib/blob/2.46.2/glib/ghash.c#L373)when looking up nodes, hence the **SIGABRT**.

Very likely this is the same problem of #2882

amnonbb:
Thanks for fast feedback.
Mountpoints does not change, but possible that someone try to get something from died sessions

amnonbb:
> You're probably using and old version of glib that [includes an assertion ]

We use latest CentOS (Rocky) that, as known, not include latest version libs.
What glib version does not include includes an assertion?

atoppi:
glib (or its version) is not to blame here, the assertion (or whatever equivalent action in recent versions) is just making the program abort for "safety" reason because the hashtable memory in that point is not valid anymore.
We need to investigate the issue and let you know if further data is needed.

atoppi:
@amnonbb can you please check the janus logs?
> Is there something on the janus log about the assertion?

amnonbb:
In log nothing about, but systemd journal got some info:
https://pastebin.com/vCF39u9m

atoppi:
yep I can spot both

```GLib:ERROR:ghash.c:378:g_hash_table_lookup_node: assertion failed: (hash_table->ref_count > 0)```

and

```
g_hash_table_lookup: assertion 'hash_table != NULL' failed
```

this seems to confirm my previous guess and also that #2882 refers to the same issue.

lminiero:
[...]

PR #2897, opened 2022-02-21
Attempt to fix race condition in Streaming plugin (see #2896)

lminiero:
As the title says, this is an attempt to fix the race condition identified in #2896, where the `streams_byid` hashtable may be destroyed by a core loop while a mountpoint thread accesses the same table to service packets.


lminiero:
Merging.

PR #2901, opened 2022-02-24
Use the same version number for npm package

siebertm:
I noticed that the package.json still includes version `0.7.5`, while the main repo is at `0.11.8` right now. I feel those should be matching? (I may be wrong, but there's only one way to find out!)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "janus-gateway",
-  "version": "0.7.5",
+  "version": "1.0.0",
   "description": "A javascript library for interacting with the C based Janus WebRTC Server",
   "main": "html/janus.js",
   "repository": {


januscla:
Thanks for your contribution, @siebertm! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Good catch: while I usually update several files every time we do a release, I don't think we ever updated that one (probably because I even forgot it exists: I most definitely never use that file).

That said, the version number you committed is incorrect: `master` is `1.0.0` now. The `0.11.8` version should be put on the `0.x` branch, where the previous version of Janus lives on.

siebertm:
Thanks for the quick feedback! I am a strong believer of Changelog files and tags, and both state 0.11.8 on master... Where's the "real" version number noted?

lminiero:
The changelog is only updated when we freeze a release. Master is version 1.0.0:

https://github.com/meetecho/janus-gateway/blob/master/configure.ac#L1
https://groups.google.com/g/meetecho-janus/c/t181VvXl0Es/m/qgBkfo77BQAJ

lminiero:
As a side note, `0.x` is currently `0.12.0`, which will be the next version we'll tag for the 0.x branch.
https://github.com/meetecho/janus-gateway/blob/0.x/configure.ac#L1

As such, the `package.json` for that will be `0.12.0`.

siebertm:
Alright, noted! Also, the version number is updated!

lminiero:
Ack, thanks! I'll merge then, and update `0.x` accordingly.

siebertm:
Thanks a bunch!

PR #2904, opened 2022-02-25
Update typings

siebertm:
While implementing our solution basing on janus.js, our team did not notice that janus ships with (some) typings at all, so they rolled their own. This is due to using the git checkout directly, whose package.json does not mention the `types` option.

This was added now, so the types are picked up. Additionally, I updated the types with the more refined version from our team, plus changes for the multistream version of janus.js (mainly `onremotetrack` etc).

Those typings make it through our typechecks, but I reserve the option to add more stuff there as we go further!

Any feedback is valued!

### DIFF
--- a/npm/janus.d.ts
+++ b/npm/janus.d.ts
@@ -23,7 +23,13 @@ declare namespace JanusJS {
 		Error = 'error'
 	}
 
-	interface JSEP {}
+	interface JSEP {
+		ee2e?: boolean;
+		sdp?: string;
+		type?: string;
+		rid_order?: "hml" | "lmh";
+		force_relay?: boolean;
+	}
 
 	interface InitOptions {
 		debug?: boolean | 'all' | DebugLevel[];
@@ -44,18 +50,23 @@ declare namespace JanusJS {
 		error?: (error: any) => void;
 		destroyed?: Function;
 	}
-	
+
 	interface ReconnectOptions {
 		success?: Function;
-		error?: (error: any) => void;
+		error?: (error: string) => void;
 	}
 
 	interface DestroyOptions {
-		cleanupHandles?: boolean
-		notifyDestroyed?: boolean
-		unload?: boolean
-		success?: () => void
-		error?: (error: Error | unknown) => void
+		cleanupHandles?: boolean;
+		notifyDestroyed?: boolean;
+		unload?: boolean;
+		success?: () => void;
+		error?: (error: string) => void;
+	}
+
+	interface GetInfoOptions {
+		success?: (info: any) => void;
+		error?: (error: string) => void;
 	}
 
 	enum MessageType {
@@ -74,27 +85,31 @@ declare namespace JanusJS {
 			id?: string;
 			uplink?: number;
 		};
-		error?: Error;
+		error?: string;
+		[key: string]: any;
 	}
 
-	interface PluginOptions {
-		plugin: string;
-		opaqueId?: string;
+	interface PluginCallbacks {
 		dataChannelOptions?: RTCDataChannelInit;
 		success?: (handle: PluginHandle) => void;
-		error?: (error: any) => void;
+		error?: (error: string) => void;
 		consentDialog?: (on: boolean) => void;
 		webrtcState?: (isConnected: boolean) => void;
 		iceState?: (state: 'connected' | 'failed' | 'disconnected' | 'closed') => void;
 		mediaState?: (medium: 'audio' | 'video', receiving: boolean, mid?: number) => void;
-		slowLink?: (uplink: boolean, lost: number) => void;
+		slowLink?: (uplink: boolean, lost: number, mid: string) => void;
 		onmessage?: (message: Message, jsep?: JSEP) => void;
-		onlocalstream?: (stream: MediaStream) => void;
-		onremotestream?: (stream: MediaStream) => void;
+		onlocaltrack?: (track: MediaStreamTrack, on: boolean) => void;
+		onremotetrack?: (track: MediaStreamTrack, mid: string, on: boolean) => void;
 		ondataopen?: Function;
 		ondata?: Function;
 		oncleanup?: Function;
-		detached?: Function;
+		ondetached?: Function;
+	}
+
+	interface PluginOptions extends PluginCallbacks {
+		plugin: string;
+		opaqueId?: string;
 	}
 
 	interface OfferParams {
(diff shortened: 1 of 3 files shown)


lminiero:
We never use those files ourselves, so I guess it was to be expected they'd be left behind when we merged the multistream branch, that indeed changed some things in `janus.js`. We'll review later, thanks!

npm/janus.d.ts
```diff
@@ -44,18 +50,18 @@ declare namespace JanusJS {
...
-		cleanupHandles?: boolean
-		notifyDestroyed?: boolean
-		unload?: boolean
-		success?: () => void
-		error?: (error: Error | unknown) => void
+		cleanupHandles?: boolean;
+		notifyDestroyed?: boolean;
+		unload?: boolean;
+		success?: () => void;
+		error?: (error: Error | unknown) => void;
```
atoppi, line 64: the other `error` cb are declared as 
```
(error: any) => void
```
or
```
(error: string) => void
```

npm/janus.d.ts
```diff
@@ -130,33 +141,65 @@ declare namespace JanusJS {
 			[otherProps: string]: any;
 		};
 		jsep?: JSEP;
-		success?: Function;
-		error?: (error: any) => void;
+		success?: (data?: any) => void;
+		error?: (error: string) => void;
+	}
+	interface Publisher {
```
atoppi, line 147: this declaration should not be here, since it is part of the videoroom plugin

npm/janus.d.ts
```diff
@@ -74,29 +80,34 @@ declare namespace JanusJS {
...
+		slowLink?: (uplink: boolean, lost: number, mid: string) => void;
 		onmessage?: (message: Message, jsep?: JSEP) => void;
-		onlocalstream?: (stream: MediaStream) => void;
-		onremotestream?: (stream: MediaStream) => void;
+		onlocaltrack?: (track: MediaStreamTrack, on: boolean) => void;
+		onremotetrack?: (track: MediaStreamTrack, mid: string, on: boolean) => void;
 		ondataopen?: Function;
 		ondata?: Function;
 		oncleanup?: Function;
 		detached?: Function;
```
atoppi, line 103: this should be `ondetached`

npm/janus.d.ts
```diff
@@ -130,33 +141,65 @@ declare namespace JanusJS {
...
+	}
+	interface DetachOptions {
+		success?: () => void;
+		error?: (error: string) => void;
+		noRequest?: boolean;
+	}
 	interface PluginHandle {
 		plugin: string;
-		id: string;
+		opaqueId?: string;
```
atoppi, line 199: I can't find this parameter in the code

npm/janus.d.ts
```diff
@@ -130,33 +141,65 @@ declare namespace JanusJS {
...
+		};
+	}
+	interface DetachOptions {
+		success?: () => void;
+		error?: (error: string) => void;
+		noRequest?: boolean;
+	}
 	interface PluginHandle {
 		plugin: string;
-		id: string;
```
atoppi, line 139: This seems to be still part of PluginHandle

npm/janus.d.ts
```diff
@@ -193,12 +241,14 @@ declare namespace JanusJS {
 
 		constructor(options: ConstructorOptions);
 
-		getServer(): string;
-		isConnected(): boolean;
-		getSessionId(): string;
 		attach(options: PluginOptions): void;
-		reconnect(options: ReconnectOptions): void;
-		destroy(options: DestroyOptions): void;
```
atoppi, line 201: any reason this bunch of declarations has been removed?

atoppi (commented):
Thanks for the effort, I left some comments.

npm/janus.d.ts
lminiero: This was already there, though, which means it will need to be fixed in the `0.x` branch too.

npm/janus.d.ts
lminiero: It isn't, in fact: it can be passed to `createHandle`, but isn't stored anywhere, since it's passed to Janus right away when doing the `attach`.

siebertm:
Lorenzo, sorry for not replying, I'm knee-deep in simulcast streams right now. I haven't forgotten this, I'll fix & update the MR this week!

lminiero:
> Lorenzo, sorry for not replying, I'm knee-deep in simulcast streams right now. I haven't forgotten this, I'll fix & update the MR this week!

No problem and no hurry! Just wanted to be sure you were still around :smile:

npm/janus.d.ts
siebertm: I agree, this is a bit out of place here. But, where else would you see those plugin-specific stuff? (it's also an option to not have this at all or "later")

npm/janus.d.ts
siebertm: Ah, so this was mixed up with `id`. :+1:

npm/janus.d.ts
siebertm: mhhhh.... "stupid developer"? :-D

siebertm (commented):
Thanks for your feedback, I addressed this now. Please check the types again, in case I missed anything. Especially the VideoRoom stuff @atoppi mentioned, I did not touch that yet.

npm/janus.d.ts
atoppi: These declarations should be addressed in separate `d.ts` files, not something to discuss here I guess.

atoppi (commented):
I left some comments.

There are some major points that I still want to address:
1) went through `janus.js` again (and it gets "funnier" every time), I noticed there is a bunch of `callbacks.error` calls with an Object parameter (either an Error / Exception or generic Object), so maybe we should double check the error types for any error callbacks or simply switch everything to 

```
error?(error: any): void;
```

2) functions are sometimes declared with the arrow syntax, I guess we could switch to the compact colon syntax, e.g.

```js
//success?: () => void;
success?(): void;
```

3) I'm not sure declarations with generic `Function` type are correct at all, e.g.
```
callback?: Function;
```

lminiero:
@siebertm is there any update on this? I see Alessandro added some notes but I don't think they were addressed eventually. Thanks!

npm/janus.d.ts
```diff
@@ -173,32 +227,53 @@ declare namespace JanusJS {
...
 		static randomString(length: number): string;
-		static attachMediaStream(element: HTMLMediaElement, stream: MediaStream): void;
-		static reattachMediaStream(to: HTMLMediaElement, from: HTMLMediaElement): void;
+		static attachMediaStream(
+			element: HTMLMediaElement,
+			stream: MediaStream
+		): void;
+		static reattachMediaStream(
+			to: HTMLMediaElement,
+			from: HTMLMediaElement
```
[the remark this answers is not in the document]
siebertm: you're not a fan of line breaks, right? :-)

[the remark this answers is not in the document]
atoppi: I'm not the one to blame here 😅

siebertm:
>     1. went through `janus.js` again (and it gets "funnier" every time), I noticed there is a bunch of `callbacks.error` calls with an Object parameter (either an Error / Exception or generic Object), so maybe we should double check the error types for any error callbacks or simply switch everything to
>  
> ```
> error?(error: any): void;
> ```

Mhhh.... janus.js indeed is Javascript code that does work. I don't want to offend anybody here, because we all know how hard WebRTC can be, so the code is indeed of great value. But it is visible that it's not 100% up to par with how I'd personally write it **if I'd do such a thing today** (Oh, believe me, I've got tons of crappy code written in my life - like all of us!)

The error handling is also part of what can be handled in a better way (at least unified). I'd not touch this right now, in this PR, to limit the scope, if you ask me.

>     2. functions are sometimes declared with the arrow syntax, I guess we could switch to the compact colon syntax, e.g.

Yes, those could be unified in one or the other way, as they're mostly equivalent. In our code, we usually use the `success: () => void` to say "this is a parameter, to be passed by the user, that is named `success` and should be a function returning `void`.

Versus `success(): void` being a method provided by the system to the user, returning `void`.

But, :shrug:
Would not mix those code style things into this non-style PR.

>     3. I'm not sure declarations with generic `Function` type are correct at all, e.g.

Those are equivalent to `(...args: any[]) => any`, so totally fine.


@lminiero sorry again. need more hours per day. :/

npm/janus.d.ts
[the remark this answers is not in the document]
siebertm: Who am I to blame anybody here?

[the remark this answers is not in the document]
lminiero: No worries, he's blaming me as I'm the one usually talking about code style :hand_over_mouth:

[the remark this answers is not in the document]
siebertm: For JS/TS: `yarn add -D prettier`. No debates anymore. If prettier decides it's that way, it is that way. Solved tons of discussions on our side. I wish there was prettier for every single programming language. Would save tons of flame wars.

lminiero:
No problem, no offense taken! We do know `janus.js` is not perfect at all, as it's basically a Frankenstein piece of code that started from a non-JS developer like me, and so comes with a lot of baggage. Ugly as it is, though, there are people using it in their applications, so any change to some key things (e.g., how errors are reported) can lead to breaking changes.

We now always clarify that `janus.js` is only meant as an example of how you can talk to Janus, and so it really shouldn't be used for anything more than a demo. A better way to write a new application (web or nodejs) is using our [Janode SDK](https://github.com/meetecho/janode/) instead.

I'm fine with the PR as it is, so unless @atoppi thinks it needs some further tweaking, we can merge :+1:

lminiero:
> so you're saying that for "production contributions", that would be the better way to spend our time?

I think so, as Janode is what we use ourselves as well in our own production services (even though we use it for server side controllers, rather than in the browser itself), and so it's what we focus on the most. At the moment it still doesn't fully support multistream, but that will come sooner or later.

atoppi:
Thanks for the changes @siebertm.
As a side note: kind of a personal request, but reviewing a force-pushed branch after a month feels like reviewing the whole thing again ;-) I would prefer checking a diff. We can squash the commits when merging.

> The error handling is also part of what can be handled in a better way (at least unified). I'd not touch this right now, in this PR, to limit the scope, if you ask me.

Ok, I agree on limiting the scope.

> > 2. functions are sometimes declared with the arrow syntax, I guess we could switch to the compact colon syntax, e.g.
> > ```
> 
> Yes, those could be unified in one or the other way, as they're mostly equivalent. In our code, we usually use the `success: () => void` to say "this is a parameter, to be passed by the user, that is named `success` and should be a function returning `void`.
> 
> Versus `success(): void` being a method provided by the system to the user, returning `void`.
> 
> But, shrug Would not mix those code style things into this non-style PR.

Ok.

> 
> > ```
> > 3. I'm not sure declarations with generic `Function` type are correct at all, e.g.
> > ```
> 
> Those are equivalent to `(...args: any[]) => any`, so totally fine.

Oh good to know thanks.
I quickly went through TS docs without finding anything relevant, maybe it is a kind of legacy syntax?

atoppi (approved):
lgtm thanks again

siebertm:
> I quickly went through TS docs without finding anything relevant, maybe it is a kind of legacy syntax?

`Function` is essentially a "type" here, so it accepts everything whose prototype is `Function`, that is, all ... functions. So that's part of Javascript, actually: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function.

lminiero:
Merging then :+1:

Issue #2899, opened 2022-02-22
[1.x] Videoroom relays packets from two simulcast substreams


OxleyS:
**What version of Janus is this happening on?**
1.x, commit `ee313ac2d855dd33a9ebc402386397340c9914c1`

In seemingly very particular circumstances, Janus routes packets from two simulcast SSRCs to a single subscriber. When this happens, in `LOG_VERB` the log is spammed with
```
SSRC changed, 1619259369 --> 2561054186
Computed offset for RTP timestamp: 2
SSRC changed, 2561054186 --> 1619259369
Computed offset for RTP timestamp: 539
```
for several seconds before calming down. During this period, the subscriber receives corrupted video.

The publisher's video sender is configured with three `sendEncodings`, with RIDs `h,m,l` in that order, but due to the video resolution being 768x432, and the bitrate constraints set on the room, Chrome seems to be silently disabling the encoding with RID `l`.

I briefly looked into the code - when this happens, at [this point in the code](https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_videoroom.c#L6604), `ps->vssrc` contains `[0, 1619259369, 2561054186]`. That causes [this case](https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_videoroom.c#L10673) to be hit when relaying packets from both SSRCs, causing packets to be unconditionally relayed.

A Node project reproducing the problem can be found [here](https://github.com/OxleyS/janus-ssrc-switch-bug-repro). It's a modified version of the repro project for https://github.com/meetecho/janus-gateway/issues/2775, so the setup should be all the same.

lminiero:
I guess the `} else if(packet->ssrc[0] != 0) {` check should really be a `} else if(ps->simulcast) {` check, exactly because you're in a scenario where index `0` is currently empty but simulcast is in use anyway and so should be processed accordingly. Can you check if that fixes it?

lminiero:
If it does, I guess we should add a boolean `simulcast` property to `janus_videoroom_rtp_relay_packet` like the `svc` we have already, since checking `ps->simulcast` may be frail (according to the checks we have before, `ps` could be `NULL`, even though it shouldn't).

lminiero:
> There does seem to be a separate problem in this scenario where the video starts to progressively lag behind farther and farther, but I'll investigate that more and file a separate issue if I find anything.

[...]

PR #2908, opened 2022-03-01
Fix transport-wide CC feedback when simulcast SSRCs are missing

OxleyS:
When a publisher never sends video frames for a simulcast layer (whether due to its own heuristics or the layer being disabled by the user), the SSRC for that layer will remain unknown. The TWCC feedback mechanism was always picking the first layer's SSRC, causing the report to default to an SSRC of zero if the layer never got any data.

As I understand it, being transport-wide, any SSRC going over the transport will do, so this simply goes through each simulcast layer and picks the first SSRC that exists.

The bug this PR fixes turned out to be the underlying cause of the behavior described [here](https://github.com/meetecho/janus-gateway/issues/2899#issuecomment-1047912315). In short, the lack of feedback reports with known SSRCs caused the publisher to continually downgrade the BWE for the connection, until it got low enough that the VP8 encoder fell over, causing the timestamps on frames to fall out of pace.

### DIFF
--- a/src/ice.c
+++ b/src/ice.c
@@ -3994,22 +3994,22 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
 				/* Pick the first valid SSRC we find across all considered mediums */
 				int i = 0;
 				for(i = 0; i < 3; i++) {
-					if(m->ssrc_peer[i] != 0)
+					if(m->ssrc_peer[i] != 0) {
+						ssrc_peer = m->ssrc_peer[i];
+						medium = m;
 						break;
+					}
 				}
 
 				/* Stop if we found a valid SSRC/medium to use */
-				if(i < 3) {
-					ssrc_peer = m->ssrc_peer[i];
-					medium = m;
+				if(medium && ssrc_peer)
 					break;
-				}
 			}
 		}
 		janus_mutex_unlock(&handle->mutex);
 	}
 
-	if(!medium) {
+	if(medium == NULL) {
 		JANUS_LOG(LOG_HUGE, "No medium with a valid peer SSRC found for transport-wide CC feedback\n");
 		return G_SOURCE_CONTINUE;
 	}
--- a/src/ice.c
+++ b/src/ice.c
@@ -3997,6 +3997,20 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
 	}
 
 	if(pc && pc->do_transport_wide_cc && medium) {
+		/* Simulcast layers that have not received data will not have a recorded peer SSRC. */
+		/* Pick the first layer that has a peer SSRC */
+		guint32 ssrc_peer = 0;
+		for(int i = 0; i < 3; i++) {
+			if(medium->ssrc_peer[i] != 0) {
+				ssrc_peer = medium->ssrc_peer[i];
+				break;
+			}
+		}
+		if (ssrc_peer == 0) {
+			JANUS_LOG(LOG_WARN, "No peer SSRC, cannot send transport-wide CC feedback\n");
+			return G_SOURCE_CONTINUE;
+		}
+
 		/* Create a transport wide feedback message */
 		size_t size = 1300;
 		char rtcpbuf[1300];
@@ -4065,7 +4079,7 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
 			guint8 feedback_packet_count = pc->transport_wide_cc_feedback_count++;
 			/* Create RTCP packet */
 			int len = janus_rtcp_transport_wide_cc_feedback(rtcpbuf, size,
-				medium->ssrc, medium->ssrc_peer[0], feedback_packet_count, packets_to_process);
+				medium->ssrc, ssrc_peer, feedback_packet_count, packets_to_process);
 			/* Enqueue it, we'll send it later */
 			if(len > 0) {
 				janus_plugin_rtcp rtcp = { .mindex = medium->mindex, .video = TRUE, .buffer = rtcpbuf, .length = len };
--- a/src/ice.c
+++ b/src/ice.c
@@ -3979,6 +3979,7 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
 	janus_ice_handle *handle = (janus_ice_handle *)user_data;
 	janus_ice_peerconnection *pc = handle->pc;
 
+	guint32 ssrc_peer = 0;
 	janus_ice_peerconnection_medium *medium = NULL;
 	if(pc) {
 		/* Find inbound video medium */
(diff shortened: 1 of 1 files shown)


atoppi:
Thanks @OxleyS for the effort.
I did suspect indeed something weird happening on the RTCP but I just focused on the SR/RR ignoring TWCC and BWE related issues.
The proposed changes look good in general.

Did you try what happens with Firefox when different layers SSRCs are being used for twcc media ?

lminiero:
> Did you try what happens with Firefox when different layers SSRCs are being used for twcc media ?

Firefox always advertises SSRCs so it's not an issue I guess. It's Chrome that doesn't when rid-based simulcast is used.

src/ice.c
```diff
@@ -3997,6 +3997,20 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
 	}
 
 	if(pc && pc->do_transport_wide_cc && medium) {
+		/* Simulcast layers that have not received data will not have a recorded peer SSRC. */
+		/* Pick the first layer that has a peer SSRC */
+		guint32 ssrc_peer = 0;
+		for(int i = 0; i < 3; i++) {
```
lminiero, line 4003: Per code style, we don't define variables inside the `for` loop, so please define `i` outside.

src/ice.c
```diff
@@ -3997,6 +3997,20 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
...
+			if(medium->ssrc_peer[i] != 0) {
+				ssrc_peer = medium->ssrc_peer[i];
+				break;
+			}
+		}
+		if (ssrc_peer == 0) {
```
lminiero, line 4009: Code style: no space between `if` and bracket.

src/ice.c
```diff
@@ -3997,6 +3997,20 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
...
+			JANUS_LOG(LOG_WARN, "No peer SSRC, cannot send transport-wide CC feedback\n");
```
lminiero, line 4010: I wouldn't display a warning here: in case it happens, this is going to spam the logs a lot, since TWCC feedback is very frequent (5 to 10 times per second by default). I'd make it a `LOG_HUGE` instead.

lminiero (commented):
I added a few notes inline. That said, I think the approach may need to be changed, since you're applying that logic only to a single medium instance, the one we choose initially in `janus_ice_outgoing_transport_wide_cc_feedback` where we say `/* Find inbound video medium */`. Under normal circumstances that's ok, but a PeerConnection may contain more than one video stream, so why should only one of them be analysed for this process, especially considering you may end up without a SSRC? (you currently have a warning for that).

You may want to move your SSRC lookup login in the previous while instead:

	while (g_hash_table_iter_next(&iter, NULL, &value)) {
		janus_ice_peerconnection_medium *m = value;
		if(m && m->type == JANUS_MEDIA_VIDEO && m->recv) {
			medium = m;
			break;
		}
	}

so that you can always move to another medium candidate in case the previous one doesn't contain an SSRC that might be of use.

I'm also wondering if we might extend the selection to audio m-lines as well: I can't remember if we negotiate TWCC for audio streams as well, even though that may be problematic (especially if browsers do not take those into account instead).

OxleyS:
Thanks for the review - I'll hopefully be able to work on the points you brought up later today.

lminiero:
> As I understand it, we have to take bundling into consideration. If audio/video are being bundled, then we can and should include audio SSRCs in the SSRC selection process

Media is always bundled in Janus. Not all media lines may be part of the TWCC process, though, as it depends whether or not the extension (and RTCP feedback) is negotiated on all of them or not. IIRC that's only true for video streams in browsers, but I may be remembering things wrong.

> VideoRoom, meanwhile, only inserts it into SDP (both offer and answer) for video m-lines. I'm wondering if this could cause issues down the line if, say, a connection first negotiates audio-only, then adds video later.

Shouldn't be an issue, as we update our references to RTP extensions any time an update takes place.

OxleyS:
> IIRC that's only true for video streams in browsers, but I may be remembering things wrong

Tested this just now, it seems Chrome negotiates it for both, at least in the latest version. Firefox still does video-only. If we want to use this to gate which SSRCs we consider for the feedback report, then we might want to move the `do_transport_wide_cc` flag from PeerConnection-level to medium-level.

That's an increase in scope, though, so for this PR maybe we can just expand to all video SSRCs as you suggested, and take care of audio in a followup if so desired.

> Shouldn't be an issue, as we update our references to RTP extensions any time an update takes place.

I was more worried about whether browsers would do the right thing, although I tested that too with Chrome/Firefox and it seems they handle it fine.

atoppi:
> That's an increase in scope, though, so for this PR maybe we can just expand to all video SSRCs as you suggested, and take care of audio in a followup if so desired.

From my point of view, even extending the search to all video m-lines is out of scope here (the issue was about missing rid layers in simulcast). The proposed change is still an improvement over the existing code and fixes the issue.

lminiero:
Sure, we can merge as is (removing the LOG_WARN at least though), but this would need to be addressed in another commit nevertheless.

src/ice.c
```diff
@@ -3989,28 +3990,31 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
...
+					if(m->ssrc_peer[i] != 0)
+						break;
+				}
+
+				/* Stop if we found a valid SSRC/medium to use */
+				if(i < 3) {
+					ssrc_peer = m->ssrc_peer[i];
+					medium = m;
+					break;
+				}
```
lminiero, line 4006: I think this could be simplified as

```
int i = 0;
for(i=0; i<3; i++) {
	if(m->ssrc_peer[i] > 0) {
		medium = m;
		ssrc_peer = m->ssrc_peer[i];
		break;
	}
}
if(medium && ssrc_peer)
	break;
```

as I feel the second check is a bit more "awkward" (and possibly error prone should we decide to increase the number of substreams in the future), and the whole block could be made more compact.

src/ice.c
```diff
@@ -3989,28 +3990,31 @@ static gboolean janus_ice_outgoing_transport_wide_cc_feedback(gpointer user_data
...
-			if(medium->ssrc_peer[i] != 0) {
-				ssrc_peer = medium->ssrc_peer[i];
-				break;
-			}
-		}
-		if (ssrc_peer == 0) {
-			JANUS_LOG(LOG_WARN, "No peer SSRC, cannot send transport-wide CC feedback\n");
-			return G_SOURCE_CONTINUE;
-		}
+	if(!medium) {
```
lminiero, line 4012: Just as a small nit and personal preference in terms of code style, we typically use `if(medium == NULL)` for checks like that.

lminiero (commented):
Thanks! Added a couple of editorial notes (the changes look fine to me otherwise).

src/ice.c
OxleyS: I found instances of both styles in the code base and wasn't sure which to use. Is there a coding style document I could refer to in the future?

lminiero: You're 100% right, and apologies for the confusion: that's mainly because of the many contributions we've received so far (and of which I'm grateful for of course), where we haven't always enforced a specific code style. Unfortunately we don't have a reference for that, and I'm not really sure we'll prepare one: even @atoppi and I write things a bit differently :laughing:  These are small nits that can be ascribed to my excessive pedantry on the matter, so no need to concern too much on that!

OxleyS:
Yeah, I've been checking it using the project where I originally noticed the behavior.

lminiero:
Just tested and I can't replicate the delayed/stuttering video anymore, so I'll merge. Thanks!

PR #2909, opened 2022-03-01
Add quirk for RTSP servers

jp-bennett:
Some RTSP Servers expect only the path, not the full url, for DESCRIBE, SETUP, and PLAY commands.

### DIFF
--- a/plugins/janus_streaming.c
+++ b/plugins/janus_streaming.c
@@ -144,6 +144,8 @@ The following options are only valid for the 'rtsp' type:
 url = RTSP stream URL
 rtsp_user = RTSP authorization username, if needed
 rtsp_pwd = RTSP authorization password, if needed
+rtsp_quirk = Some RTSP servers offer the stream using only the path, instead of the fully qualified URL.
+	If set true, this boolean informs Janus that we should try a path-only DESCRIBE request if the initial request returns 404.
 rtsp_failcheck = whether an error should be returned if connecting to the RTSP server fails (default=true)
 rtspiface = network interface IP address or device name to listen on when receiving RTSP streams
 rtsp_reconnect_delay = after n seconds passed and no media assumed, the RTSP server has gone and schedule a reconnect (default=5s)
@@ -1096,6 +1098,8 @@ typedef struct janus_streaming_rtp_source {
 	janus_streaming_buffer *curldata;
 	char *rtsp_url;
 	char *rtsp_username, *rtsp_password;
+	char *rtsp_stream_uri;
+	gboolean rtsp_quirk;
 	gint64 ka_timeout;
 	char *rtsp_ahost, *rtsp_vhost;
 	gboolean reconnecting;
@@ -1221,7 +1225,7 @@ janus_streaming_mountpoint *janus_streaming_create_file_source(
 janus_streaming_mountpoint *janus_streaming_create_rtsp_source(
 		uint64_t id, char *id_str, char *name, char *desc, char *metadata,
 		char *url, char *username, char *password,
-		gboolean doaudio, int audiopt, char *artpmap, char *afmtp,
+		gboolean quirk, gboolean doaudio, int audiopt, char *artpmap, char *afmtp,
 		gboolean dovideo, int videopt, char *vrtpmap, char *vfmtp, gboolean bufferkf,
 		const janus_network_address *iface, int threads,
 		gint64 reconnect_delay, gint64 session_timeout, int rtsp_timeout, int rtsp_conn_timeout,
@@ -2132,6 +2136,7 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 				janus_config_item *file = janus_config_get(config, cat, janus_config_type_item, "url");
 				janus_config_item *username = janus_config_get(config, cat, janus_config_type_item, "rtsp_user");
 				janus_config_item *password = janus_config_get(config, cat, janus_config_type_item, "rtsp_pwd");
+				janus_config_item *quirk = janus_config_get(config, cat, janus_config_type_item, "rtsp_quirk");
 				janus_config_item *audio = janus_config_get(config, cat, janus_config_type_item, "audio");
 				janus_config_item *artpmap = janus_config_get(config, cat, janus_config_type_item, "audiortpmap");
 				janus_config_item *acodec = janus_config_get(config, cat, janus_config_type_item, "audiopt");
@@ -2155,6 +2160,7 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 					continue;
 				}
 				gboolean is_private = priv && priv->value && janus_is_true(priv->value);
+				gboolean rtsp_quirk = quirk && quirk->value && janus_is_true(priv->value);
 				gboolean doaudio = audio && audio->value && janus_is_true(audio->value);
 				gboolean dovideo = video && video->value && janus_is_true(video->value);
 				gboolean bufferkf = video && vkf && vkf->value && janus_is_true(vkf->value);
@@ -2189,6 +2195,7 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 						(char *)file->value,
 						username ? (char *)username->value : NULL,
 						password ? (char *)password->value : NULL,
+						rtsp_quirk,
 						doaudio,
 						(acodec && acodec->value) ? atoi(acodec->value) : -1,
 						artpmap ? (char *)artpmap->value : NULL,
@@ -2616,6 +2623,8 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 						json_object_set_new(ml, "rtsp_user", json_string(source->rtsp_username));
 					if(source->rtsp_password)
 						json_object_set_new(ml, "rtsp_pwd", json_string(source->rtsp_password));
+					if(source->rtsp_quirk)
+						json_object_set_new(ml, "rtsp_quirk", json_true());
 				}
 			}
 #endif
@@ -3220,6 +3229,7 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 			json_t *url = json_object_get(root, "url");
 			json_t *username = json_object_get(root, "rtsp_user");
 			json_t *password = json_object_get(root, "rtsp_pwd");
+			json_t *quirk = json_object_get(root, "rtsp_quirk");
 			json_t *iface = json_object_get(root, "rtspiface");
 			json_t *threads = json_object_get(root, "threads");
 			json_t *failerr = json_object_get(root, "rtsp_failcheck");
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @jp-bennett! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

plugins/janus_streaming.c
```diff
@@ -1096,6 +1096,7 @@ typedef struct janus_streaming_rtp_source {
 	janus_streaming_buffer *curldata;
 	char *rtsp_url;
 	char *rtsp_username, *rtsp_password;
+        char *rtsp_stream_uri;
```
lminiero, line 1099: Code style: please use tabs and not spaces for indentation.

plugins/janus_streaming.c
```diff
@@ -6519,6 +6521,23 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 	}
 	long code = 0;
 	res = curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &code);
+#if CURL_AT_LEAST_VERSION(7, 62, 0)
+	if(code == 404) { //Possibly a quirk in the RTSP server, where the DESCRIBE request expects a path only.
```
lminiero, line 6525: Just a nit, per code style we only use `/* */` for comments: I'd also put it as the first line inside the if, rather than on the same line as the if itself (just for the sake of readability).

plugins/janus_streaming.c
```diff
@@ -6519,6 +6521,23 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
...
+		CURLU *curl_u;
+		char *path;
+		curl_u = curl_url();
```
lminiero, line 6528: I'd collapse the definition as `CURLU *curl_u = curl_url();`, no need for pre-definitions. Besides, `path` should be initialized to `NULL`, since otherwise it may point to random garbage and the `&path` may be dangerous.

plugins/janus_streaming.c
```diff
@@ -6519,6 +6521,23 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
...
+		if(!(curl_url_set(curl_u, CURLUPART_URL, source->rtsp_url, 0))) {
+			if(!(curl_url_get(curl_u, CURLUPART_PATH, &path, 0))) {
+				curl_easy_setopt(curl, CURLOPT_RTSP_STREAM_URI, path);
+				res = curl_easy_perform(curl);
+                                source->rtsp_stream_uri = path;
```
lminiero, line 6533: Indentation broken here (spaces instead of tabs). That said, I'd do a `g_strdup()` for `rtsp_stream_uri`, so you can use the `curl_free(path)` you commented below.

plugins/janus_streaming.c
```diff
@@ -5795,6 +5796,7 @@ static void janus_streaming_rtp_source_free(janus_streaming_rtp_source *source)
 	g_free(source->rtsp_url);
 	g_free(source->rtsp_username);
 	g_free(source->rtsp_password);
+	curl_free(source->rtsp_stream_uri);
```
lminiero, line 5799: If we use `g_strdup`, this should be `g_free`, which will make it consistent with all other strings we keep track of.

plugins/janus_streaming.c
```diff
@@ -7073,7 +7092,10 @@ static int janus_streaming_rtsp_play(janus_streaming_rtp_source *source) {
 	source->curldata->buffer = g_malloc0(1);
 	source->curldata->size = 0;
 	JANUS_LOG(LOG_VERB, "Sending PLAY request...\n");
-	curl_easy_setopt(source->curl, CURLOPT_RTSP_STREAM_URI, source->rtsp_url);
+	if(source->rtsp_stream_uri)
+		curl_easy_setopt(source->curl, CURLOPT_RTSP_STREAM_URI, source->rtsp_stream_uri);
+	else
+		curl_easy_setopt(source->curl, CURLOPT_RTSP_STREAM_URI, source->rtsp_url);
```
lminiero, line 7098: This might be simplified with a single line using a ternary operator:

    curl_easy_setopt(source->curl, CURLOPT_RTSP_STREAM_URI, source->rtsp_stream_uri ? source->rtsp_stream_uri : source->rtsp_url);

lminiero (commented):
Thanks for the patch! I added a few notes below, mostly editorial for now, since the feature itself seems mostly straightforward.

lminiero:
I just noticed this is a PR for `0.x`, not master. I'll port the changes to `master` as well if this is merged, since the RTSP code has seen little change, and so would apply there too.

plugins/janus_streaming.c
jp-bennett: Lots of little code style things -- I blame muscle memory. I'll fix these up, and add the config option, hopefully tonight.

jp-bennett:
@lminiero The second and third commits are compile tested only. I've temporarily lost access to the camera that demonstrated this quirk. Let me know if you see anything obviously wrong, and I'll test this in the real world once I'm able.

plugins/janus_streaming.c
```diff
@@ -3264,6 +3274,7 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 					(char *)json_string_value(url),
 					username ? (char *)json_string_value(username) : NULL,
 					password ? (char *)json_string_value(password) : NULL,
+					quirk ? TRUE : FALSE,
```
lminiero, line 3277: Same mistake as above. If you check how we deal with boolean properties in JSON requests, we do something like this:

    gboolean dovideo = video ? json_is_true(video) : FALSE;

so you should do something similar with your new feature, and pass the boolean to `janus_streaming_create_rtsp_source`.

plugins/janus_streaming.c
```diff
@@ -6519,6 +6533,23 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
...
+#if CURL_AT_LEAST_VERSION(7, 62, 0)
+	if(source->rtsp_quirk && code == 404) {
+		/*Possibly a quirk in the RTSP server, where the DESCRIBE request expects a path only.*/
+		CURLU *curl_u = curl_url();
+		char *path = NULL;
+		if(!(curl_url_set(curl_u, CURLUPART_URL, source->rtsp_url, 0))) {
+			if(!(curl_url_get(curl_u, CURLUPART_PATH, &path, 0))) {
+				curl_easy_setopt(curl, CURLOPT_RTSP_STREAM_URI, path);
+				res = curl_easy_perform(curl);
+				source->rtsp_stream_uri = g_strdup(path);
```
lminiero, line 6545: Shouldn't we check what `curl_easy_perform` and/or `curl_easy_getinfo` returned, before saving this property? If we got a `404` or a different error again, then the quirk didn't help. Besides, `path` may still be `NULL` maybe?

plugins/janus_streaming.c
```diff
@@ -7166,6 +7197,8 @@ janus_streaming_mountpoint *janus_streaming_create_rtsp_source(
 	live_rtsp_source->rtsp_url = g_strdup(url);
 	live_rtsp_source->rtsp_username = username ? g_strdup(username) : NULL;
 	live_rtsp_source->rtsp_password = password ? g_strdup(password) : NULL;
+	live_rtsp_source->rtsp_stream_uri = NULL;
+	live_rtsp_source->rtsp_quirk = FALSE;
```
lminiero, line 7201: Shouldn't this be equal to the `quirk` that's been passed to `janus_streaming_create_rtsp_source`?

plugins/janus_streaming.c
jp-bennett: I don't think path can be null. The curl documentation seems to indicate that it will be set to "/" at the worst. (https://curl.se/libcurl/c/curl_url_get.html)

jp-bennett:
@lminiero Next revision, with your comments addressed. Went ahead and squashed to a single commit

plugins/janus_streaming.c
```diff
@@ -2155,6 +2160,7 @@ int janus_streaming_init(janus_callbacks *callback, const char *config_path) {
 					continue;
 				}
 				gboolean is_private = priv && priv->value && janus_is_true(priv->value);
+				gboolean rtsp_quirk = quirk && quirk->value && janus_is_true(priv->value);
```
lminiero, line 2163: This should be `janus_is_true(quirk->value)` (my typo in the original message, sorry)

plugins/janus_streaming.c
```diff
@@ -6519,6 +6536,25 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
...
+		CURLU *curl_u = curl_url();
+		char *path = NULL;
+		if(!(curl_url_set(curl_u, CURLUPART_URL, source->rtsp_url, 0))) {
+			if(!(curl_url_get(curl_u, CURLUPART_PATH, &path, 0))) {
+				curl_easy_setopt(curl, CURLOPT_RTSP_STREAM_URI, path);
+				curl_easy_perform(curl);
+				res = curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &code);
+				if((res == CURLE_OK) && (code != 404)) {
+					source->rtsp_stream_uri = g_strdup(path);
+					curl_free(path);
```
lminiero, line 6551: The `curl_free` should be outside of the `if`, as otherwise the `path` string will leak if an error occurred.

lminiero:
> Went ahead and squashed to a single commit

Thanks but no need to do that, as I squash when I merge anyway.

plugins/janus_streaming.c
jp-bennett: This feels less like black magic, with the correct value in there. =)

plugins/janus_streaming.c
jp-bennett: Derp. Yes.

lminiero:
Thanks! This looks good to me now. If you can make a test with those quirky cameras and confirm it works as expected for you, this is good to merge for me. Before we can do that, though, we'd need you to sign the [CLA](https://cla.conf.meetecho.com/janus).

plugins/janus_streaming.c
```diff
@@ -3417,6 +3429,8 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 					janus_config_add(config, c, janus_config_item_create("rtsp_user", source->rtsp_username));
 				if(source->rtsp_password)
 					janus_config_add(config, c, janus_config_item_create("rtsp_pwd", source->rtsp_password));
+				if(source->rtsp_password)
```
lminiero, line 3432: Just noticed a typo: this should be `if(rtsp_quirk)`.

plugins/janus_streaming.c
```diff
@@ -3631,6 +3645,8 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 						janus_config_add(config, c, janus_config_item_create("rtsp_user", source->rtsp_username));
 					if(source->rtsp_password)
 						janus_config_add(config, c, janus_config_item_create("rtsp_pwd", source->rtsp_password));
+					if(source->rtsp_password)
```
lminiero, line 3648: Same here.

lminiero:
I'll merge and fix those two small typos myself. Thanks again for your contribution!

PR #2910, opened 2022-03-02
Add clang-format file

atoppi:
This **does not exactly** match the code style.
Even if not perfect, the configuration could help contributors and avoid useless code style reviews.


.clang-format
```diff
@@ -0,0 +1,40 @@
...
+SpaceBeforeSquareBrackets: false
+SpaceInEmptyBlock: false
+SpaceInEmptyParentheses: false
+SpacesInCStyleCastParentheses: false
+SpacesInConditionalStatement: false
+SpacesInContainerLiterals: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+TabWidth: 4
+UseTab: Always
```
lminiero, line 40: Should there be an empty line at the end here?

AleXoundOS:
I would be glad if sources get auto-formatted in repository. Some git-blame info might be lost, but maybe it's not so crucial.

PR #2916, opened 2022-03-07
Check if IPv6 is disabled to avoid failure when creating forwarder socket (fixes #2915)

### LINKED ISSUE #2915: [1.x] VideoRoom RTP Forwarder Doesn't Work When IPv6 is Disabled
Hey,

I've seen it on v0.10.5 but I believe is relevant to v1.0.0  as well

No, mainly due to other dependencies in the system.

No trace, as there's no crash. Just an error message in the logs:
```
Could not open UDP socket for RTP stream for publisher
```

**Additional context**
Looking at the [sources](https://github.com/meetecho/janus-gateway/blob/71c5c070a0d32e415533833ee561df2d5cbfac2e/src/plugins/janus_videoroom.c#L5097), the socket created for forwarding RTP packets is always created as IPv6 and then marked for IPv4 as well. On systems where IPv6 is disabled, e.g., when adding `ipv6.disable=1` to the kernel command line, the call to `socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);` fails. Although the `rtp_forward` command does allow to specify the host family, it seems that this is only used for DNS resolving.

I'm not sure what would be the best way to over come this but perhaps if all the forward stream requests are IPv4, we can create the socket that way in the first place? Though that would only be relevant for the first time the socket is created. Following calls to `rtp_forward` might add a new IPv6 host.

Thanks in advance!

lminiero:
As explained in #2915, if IPv6 is disabled (e.g., via kernel), creating IPv6 sockets will fail. This is a problem with RTP forwarders in the AudioBridge and VideoRoom, since we always create an IPv6 socket to send to both IPv4 and IPv6 (a single socket is created per publisher, which means the same socket can send to different families). As a result, forwarders will always fail in a system with those constraints.

This patch tries to address the problem by making a check when the plugin is started: we try and create a socket like the one we'd use for forwarders, and if that fails, we consider IPv6 unavailable. As a result, in that case we only create IPv4 sockets, and return an error if there's an attempt to forward to an IPv6 address.

I tested this briefly and it seems to work as expected, but of course you may want to make sure this doesn't break anything for you, especially if you use forwarders a lot. Notice that this PR is on master, but when this is merged I'll make sure to backport the changes to `0.x` as well.

### DIFF
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -1251,6 +1251,7 @@ static janus_mutex config_mutex = JANUS_MUTEX_INITIALIZER;
 static volatile gint initialized = 0, stopping = 0;
 static gboolean notify_events = TRUE;
 static gboolean string_ids = FALSE;
+static gboolean ipv6_disabled = FALSE;
 static janus_callbacks *gateway = NULL;
 static GThread *handler_thread;
 static void *janus_audiobridge_handler(void *data);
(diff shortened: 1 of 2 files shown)


lminiero:
Merging.

PR #2922, opened 2022-03-14
send a PLI when a subscriber is resuming a previously paused stream

flaviogrossi:
Hi, i think i found a small issue since the multistream merge:

in the current codebase, a PLI is sent when a subscriber issues a `configure` request to start a previously paused stream, see [here](https://github.com/meetecho/janus-gateway/blob/03db74a81f6ab294cd3164211a985ed2a11dae37/src/plugins/janus_videoroom.c#L9268).

However this is only done when using the deprecated `video` parameter, and not when using `send`, causing a delay to the receiver (the stream is "frozen" until the next complete frame).

This PR is to add the PLI request also in latter case.

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -9259,7 +9259,7 @@ static void *janus_videoroom_handler(void *data) {
 						gboolean oldvideo = stream->send;
 						gboolean newvideo = json_is_true(video);
 						if(!oldvideo && newvideo) {
-							/* Audio just resumed, reset the RTP sequence numbers */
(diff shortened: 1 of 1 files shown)


flaviogrossi:
Great, thanks!

Fiddling with the sources i noticed too that the `mid` parameter is ignored when using the old parameter `video`, causing all of the streams to be paused.

Out of curiosity, is this intentional? I guess it's to be absolutely retrocompatible, because the mid could be used anyway, if found.

lminiero:
Yes, if you pass `video` you implicitly ask for them all.

PR #2924, opened 2022-03-21
add data and peer_data to recording for sip plugin

fjoseph13:


januscla:
Thanks for your contribution, @fjoseph13! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
We don't negotiate datachannels in the SIP plugin, so those properties would be useless. Besides, adding datachannels support to the SIP plugin is not planned, because we might need to use those to bridge to other technologies via SIP (e.g., as we did as a proof of concept in #1898).

lminiero:
Closing for the reason explained in the above comment. Thanks for the contribution anyway!

PR #2928, opened 2022-03-25
fix: streaming plugin shows invalid stream age

RouquinBlanc:
Fixing a typo causing invalid age to be shown on streams

This is probably a regression from the multi-stream branch, as we were not getting this issue on 0.11.3

Tested against RTP/H264 stream


lminiero:
Good catch! It must have been a typo we introduced refactoring the code. Merging, thanks!

PR #2930, opened 2022-03-28
Set MTU size to play stream in NordVPN/Chrome incognito mode setup

davidengelmaier:
In case NordVPN(and possibly other VPNs) is used in combination with the Chrome incognito mode or Safari private mode and the streaming plugin, we can see a significant packet loss:

image

This can be reproduced in the official demo at https://janus.conf.meetecho.com/streamingtest.html (OPUS/VP8 live stream (sent by gstreamer script)(live), you can hear the audio, but not see the video.

You can see it working when the mtu=1360 is used at https://streaming.totodon.com/streamingtest.html by choosing the color 1280x720 (live) stream

### DIFF
--- a/src/plugins/streams/test_gstreamer1.sh
+++ b/src/plugins/streams/test_gstreamer1.sh
@@ -7,5 +7,5 @@ gst-launch-1.0 \
   videotestsrc ! \
     video/x-raw,width=320,height=240,framerate=15/1 ! \
     videoscale ! videorate ! videoconvert ! timeoverlay ! \
-    vp8enc error-resilient=1 ! \
-      rtpvp8pay ! udpsink host=127.0.0.1 port=5004
+    vp8enc keyframe-max-dist=7 error-resilient=1 ! \
+      rtpvp8pay mtu=1360 ! udpsink host=127.0.0.1 port=5004
--- a/src/plugins/streams/test_gstreamer1.sh
+++ b/src/plugins/streams/test_gstreamer1.sh
@@ -7,5 +7,5 @@ gst-launch-1.0 \
   videotestsrc ! \
     video/x-raw,width=320,height=240,framerate=15/1 ! \
     videoscale ! videorate ! videoconvert ! timeoverlay ! \
-    vp8enc keyframe-max-dist=7 error-resilient=1 ! \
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @davidengelmaier! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

atoppi:
`keyframe-max-dist` to 7 is an absolute overkill.

davidengelmaier:
> 1360 is too high: in case IPv6 is used, the MTU would be 1280 anyway. 1200 is a bette value (and the one we usually use ourselves). That said, why the `keyframe-max-dist` property? That's not in the original example, and I don't see why it should be: this is not meant to be a comprehensive and complete example, but just a quick way to test something, so I'll oppose anything that goes too much in detail.

1360 is the highest working value, i'll set it to 1200 as adviced.

The `keyframe-max-dist` is only in place to make the stream start asap, if not in place it takes a couple of seconds for the first frame to show up

davidengelmaier:
> `keyframe-max-dist` to 7 is an absolute overkill.

cca halve a second interval, that's what we use to make the time to first frame as low as possible

lminiero:
> The `keyframe-max-dist` is only in place to make the stream start asap, if not in place it takes a couple of seconds for the first frame to show up

That's normal and expected. Frequent keyframes are very bad for real-time video.

lminiero:
> ok, changed. I'm coming from a business case where immediate start of the stream is required, hence the setting.

You'll get immediate start but the performance will be bad. Frequent keyframes means basically almost every frame is a keyframe, which take more room, and are more fragile and sensitive to loss. This will make the video encoder process very inefficient. When you need more frequent keyframes, once every 5 seconds is usually a reasonable tradeoff.

That said, it looks like you only added the mtu to one of the video elements in the multistream example, though? There's two video streams in there, and they might both be affected.

lminiero:
Thanks! As soon as the CLA is signed we'll merge :+1:

davidengelmaier:
> Thanks! As soon as the CLA is signed we'll merge 👍

signed

lminiero:
Thanks, merging then :v:

PR #2935, opened 2022-03-31
Additionally return reason header protocol and cause

ajsa-terko:
Current syntax for reporting on SIP code, reason, and reason header looks as: 

```
{
        "sip" : "event",
        "result" : {
                "event" : "<name of the error event>",
                "code" : <SIP error code>,
                "reason" : "<SIP error reason>",
                "reason_header" : "<SIP reason header; optional>"
        }
}
```

reason_header, although its name suggests differently, returns only the value of the text parameter from the reason header, (correspond to the value of the text parameter returned by Sofia, [re_text](http://sofia-sip.sourceforge.net/refdocs/sip/structsip__reason__s.html#af1df8d948057c9a3251edbf114eed4c4)). Looking at the example below, the current response would return "reason_header": "Terminated".  
> Reason: Q.850;cause=16;**text=”Terminated”**

As an additional two parameters from the common syntax of the header could provide more context to what has happened (and enhance reason header mappings on the user side), the pull request suggests adding two new optional fields, reason_header_protocol, and reason_header_cause. Both of these are contained in Sofia's sip reason structure [sip_reason_s](http://sofia-sip.sourceforge.net/refdocs/sip/structsip__reason__s.html) as [re_protocol](http://sofia-sip.sourceforge.net/refdocs/sip/structsip__reason__s.html#aa8ce0397718dd57f127fbd2d73312c4d) and [re_cause](http://sofia-sip.sourceforge.net/refdocs/sip/structsip__reason__s.html#a13f0bca88f9c73a63d5dae8a78c7ea2f), and are read and returned in the same way as the current reason_header. 

After the update, the new syntax would be as:
```
{
        "sip" : "event",
        "result" : {
                "event" : "<name of the error event>",
                "code" : <SIP error code>,
                "reason" : "<SIP error reason>",
		"reason_header" : "<Reason header; optional>",
		"reason_header_protocol" : "<Reason header protocol; optional>",
		"reason_header_cause" : "<Reason header cause code; optional>"
        }
}
```

I would appreciate feedback on naming both of these fields, as reason_header may be misleading having protocol and cause separately, but cannot be simply renamed not to introduce breaking change.


src/plugins/janus_sip.c
```diff
@@ -7157,4 +7195,4 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 	JANUS_LOG(LOG_VERB, "Leaving sofia loop thread...\n");
 	g_thread_unref(g_thread_self());
 	return NULL;
-}
+}
```
lminiero, line 7198: Looks like your editor ate the last newline?

lminiero (commented):
I don't use the SIP plugin much, but I think it makes sense, especially in case those properties are indeed of use to users/developers. I agree that renaming them all would be better (possibly grouped in an object) but we do indeed have to preserve backwards compatibility, so I have no opinion on the names of the properties.

src/plugins/janus_sip.c
lminiero: Thanks!

PR #2936, opened 2022-03-31
0.x: videoroom: always default {min,max}_delay to -1

tmatth:
This was defaulting to 0 since publishers are allocated with g_malloc0.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -6302,6 +6302,8 @@ static void *janus_videoroom_handler(void *data) {
 				publisher->audio_level_extmap_id = 0;
 				publisher->video_orient_extmap_id = 0;
 				publisher->playout_delay_extmap_id = 0;
+				publisher->min_delay = -1;	/* We'll deal with this later */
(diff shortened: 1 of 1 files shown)


tmatth:
For context, we noticed a regression where subscribers were unable to watch a stream (for whatever reason, specifically with our iOS client). From bisecting, we tracked it down to https://github.com/meetecho/janus-gateway/commit/a5eee9f4c9e3bf453be0acce1365839521e7512a and I realized that e.g., this code:

https://github.com/meetecho/janus-gateway/blob/e111be2796fed4cc3734b4bab3a5f8bdb6189334/plugins/janus_videoroom.c#L6978

is only setting min/max_delay to 0 if it looked up the json value and it was < 0. If we used the config defaults (where it's not set at all), code that assumed uninitialized == -1 would break.

lminiero:
Looks good to me, merging! I'll check if the same is true in multistream too, and in case fix it there as well.

lminiero:
> I'll check if the same is true in multistream too, and in case fix it there as well.

Looks like we always initialize both properties there, when allocating publisher or subscriber streams, so no need to fix multistream too.

PR #2939, opened 2022-04-07
Fixed: the track argument of onlocaltrack() is undefined when MediaStreamTrack ended event fires.

panzy:


januscla:
Thanks for your contribution, @panzy! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
I'm not sure this works without breaking the demos: the `onlocaltrack` callback in all demos expects the track, while I'm not sure what this `target` is, and if all browsers do the same thing. Can you elaborate on why you think this patch is necessary, and if you tested this in all browsers too?

panzy:
ev.target refers to the track, and ev.track in the original code is
actually undefined.

The patch is necessary because without it the onlocaltrack callback gets
arguments (undefined, false)  when track ended happens. It addresses
exactly this FIXME issue:

> // FIXME What does this event contain? Is there a reference to the track?
>

Event.target is documented here: Event - Web APIs | MDN (mozilla.org)
<https://developer.mozilla.org/en-US/docs/Web/API/Event>

> A reference to the object to which the event was originally dispatched.


Possible causes of ended event are documented here: MediaStreamTrack: ended
event - Web APIs | MDN (mozilla.org)
<https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/ended_event>
:

> ended events fire when the media stream track's source permanently stops
> sending data on the stream. There are various ways this can happen,
> including:
>
>    - There is no more data left to send.
>    - The user revoked the permissions needed for the data to be sent.
>    - The hardware generating the source data has been removed or ejected.
>    - A remote peer has permanently stopped sending data; pausing media
>    does not generate an ended event.
>
> Below is an example event object captured from Edge on Windows:

> isTrusted: true
> bubbles: false
> cancelBubble: false
> cancelable: false
> composed: false
> currentTarget: MediaStreamTrack {kind: 'video', id:
> '50af7415-6c55-456c-acbb-815a895f31b6', label: 'USB2.0 HD UVC WebCam
> (13d3:56a8)', enabled: true, muted: false, …}
> defaultPrevented: false
> eventPhase: 0
> path: []
> returnValue: true
> srcElement: MediaStreamTrack {kind: 'video', id:
> '50af7415-6c55-456c-acbb-815a895f31b6', label: 'USB2.0 HD UVC WebCam
> (13d3:56a8)', enabled: true, muted: false, …}
> target: MediaStreamTrack {kind: 'video', id:
> '50af7415-6c55-456c-acbb-815a895f31b6', label: 'USB2.0 HD UVC WebCam
> (13d3:56a8)', enabled: true, muted: false, …}
> timeStamp: 66082.30000001192
> type: "ended"


And this is an example from Firefox on Linux:

> bubbles: false
> cancelBubble: false
> cancelable: false
> composed: false
> currentTarget: null
> defaultPrevented: false
> eventPhase: 0
> explicitOriginalTarget: MediaStreamTrack { kind: "video", id:
> "{832745a5-02bb-4a76-ad6c-9d42095d89a9}", label: "My Fake Webcam", … }
> isTrusted: true
> originalTarget: MediaStreamTrack { kind: "video", id:
> "{832745a5-02bb-4a76-ad6c-9d42095d89a9}", label: "My Fake Webcam", … }
> returnValue: true
> srcElement: MediaStreamTrack { kind: "video", id:
> "{832745a5-02bb-4a76-ad6c-9d42095d89a9}", label: "My Fake Webcam", … }
> target: MediaStreamTrack { kind: "video", id:
> "{832745a5-02bb-4a76-ad6c-9d42095d89a9}", label: "My Fake Webcam", … }
> timeStamp: 0
> type: "ended"
>

The patch was tested on Edge/Windows, Edge/macOS and Chrome/Linux. I didn't
find a way to trigger ended event on Safari/macOS.

lminiero:
Thanks for the detailed answer! I'll test when I get back to the office on Monday and merge if all's good :v:

panzy:
Sure!

FYI, one easy way to trigger ended event is to revoke the media access permission.

lminiero:
> Thanks for your contribution, @panzy! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

Could you sign the CLA so that we can merge this? Thanks!

lminiero:
Sounds good, I'll do that then, thanks for the heads up. Hopefully this will be sorted out in the future!

PR #2940, opened 2022-04-08
Don't clear buffer before passing to "SSL_read".

fancycode:
This is not necessary as "SSL_read" will fill parts of the buffer and return the number of bytes filled.


lminiero:
I can't remember if we added that for a reason, to be honest. We also support variants of OpenSSL like BoringSSL and LibreSSL, and I don't know if they may have expectations on the buffer. Your patch sounds reasonable, but I just want to be sure we're not removing something that could cause issues.

fancycode:
Of course it's your call, but I don't think any of these libraries expect any sensible data (i.e. `0`) in the buffer to be filled.

lminiero:
Makes sense, merging then :+1:

PR #2941, opened 2022-04-08
Added new demos: WebAudio and Virtual Backgrounds

lminiero:
As te title says, this adds nothing to Janus, but only adds a couple of demos focused entirely on the client side. Specifically, I added a couple of very basic demos:

1. `webaudio.html/:js` shows how to use [Web Audio](https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API) to do some processing of what comes from the microphone, before sending it to Janus; it's very basic, since on the capture side it only adds a compressor, where you can dynamically tweak the parameters via some form inputs; the audio is then sent to the EchoTest plugin, which sends it back, and so WebAudio is used again just to visualize the output as well (which should help see the effect of the compressor).
2. `virtualbg.html/.js` instead shows how you can use [MediaPipe](https://google.github.io/mediapipe/) to add a virtual background to the local camera before sending it to Janus: since the EchoTest is used here as well, you can see that what you send is exactly what you receive.

Both demos are quite basic and are not meant to provide something production ready: there's much more you can do with Web Audio than that, for instance, and the same could be said for virtual backgrounds (where you may want to blur, instead of using a background, or use a green screen like [this excellent webrtcHacks blog post](https://webrtchacks.com/how-to-make-virtual-backgrounds-transparent-in-webrtc/) describes).

That said, hopefully they'll provide useful to people that want to tinker with features like this, as I wanted to stress out these are things you typically do on the client side, and have nothing to do with Janus.


lminiero:
Merging.

PR #2945, opened 2022-04-12
Add stream description

adnanel:
Initial PoC for the idea [discussed here](https://groups.google.com/g/meetecho-janus/c/zRtRr4HZ_yY).

The stream description is stored inside the MJR JSON header.

Example:
root@d686bae6dcd2:/# janus-pp-rec --json videoroom-1234-user-7961364863991430-1649775467729201-video-1.mjr 
{"t": "v", "c": "vp8", "desc": "screenshare", "x": {"13": "urn:3gpp:video-orientation"}, "s": 1649775467729302, "u": 1649775469347109}

I'm unsure about the fact that the stream description is now saved in 2 places, the `janus_recorder` instance (newly added), and the owning `janus_videoroom_publisher_stream` (where it was located before). 

Open to feedback

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -7104,7 +7104,7 @@ static void janus_videoroom_recorder_create(janus_videoroom_publisher_stream *ps
 			}
 		}
 		/* If the stream has a description, store it in the recording */
-		if(ps->description && rc)
+		if(ps->description)
 			janus_recorder_description(rc, ps->description);
 		/* If the video-orientation extension has been negotiated, mark it in the recording */
 		if(ps->video_orient_extmap_id > 0)
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -7103,6 +7103,9 @@ static void janus_videoroom_recorder_create(janus_videoroom_publisher_stream *ps
 					janus_videoroom_media_str(ps->type));
 			}
 		}
+		/* If the stream has a description, store it in the recording */
+		if(ps->description && rc)
+			rc->description = ps->description;
 		/* If the video-orientation extension has been negotiated, mark it in the recording */
 		if(ps->video_orient_extmap_id > 0)
 			janus_recorder_add_extmap(rc, ps->video_orient_extmap_id, JANUS_RTP_EXTMAP_VIDEO_ORIENTATION);
@@ -10410,6 +10413,7 @@ static void *janus_videoroom_handler(void *data) {
 					while(temp) {
 						janus_videoroom_publisher_stream *ps = (janus_videoroom_publisher_stream *)temp->data;
 						janus_videoroom_recorder_create(ps);
+						AAA
 						temp = temp->next;
 					}
 					participant->recording_active = TRUE;
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -7105,7 +7105,7 @@ static void janus_videoroom_recorder_create(janus_videoroom_publisher_stream *ps
 		}
 		/* If the stream has a description, store it in the recording */
 		if(ps->description && rc)
-			rc->description = ps->description;
+			rc->description = g_strdup(ps->description);
 		/* If the video-orientation extension has been negotiated, mark it in the recording */
 		if(ps->video_orient_extmap_id > 0)
 			janus_recorder_add_extmap(rc, ps->video_orient_extmap_id, JANUS_RTP_EXTMAP_VIDEO_ORIENTATION);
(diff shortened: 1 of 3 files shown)


src/plugins/janus_videoroom.c
```diff
@@ -7103,6 +7103,9 @@ static void janus_videoroom_recorder_create(janus_videoroom_publisher_stream *ps
 					janus_videoroom_media_str(ps->type));
 			}
 		}
+		/* If the stream has a description, store it in the recording */
+		if(ps->description && rc)
+			rc->description = g_strdup(ps->description);
```
lminiero, line 7108: I'd prefer there to be a setter in `record.h` for the description, just as in other properties. I don't like the idea of plugins messing with the properties of the `janus_recorder` directly, as a getter can use a mutex for instance.

src/plugins/janus_videoroom.c
```diff
@@ -8617,6 +8620,11 @@ static void *janus_videoroom_handler(void *data) {
 								desc_updated = TRUE;
 								g_free(ps->description);
 								ps->description = g_strdup(d_desc);
+								if(ps->rc) {
+									if(ps->rc->description)
+										g_free(ps->rc->description);
+									ps->rc->description = g_strdup(d_desc);
+								}
```
lminiero, line 8627: See above: just use a setter and you don't have to worry about freeing and duping. It's important to point out that this call would be useless when invoked after the mjr header has been written and the recorder is working, since we only write it down before all that happens.

src/record.c
```diff
@@ -348,6 +349,8 @@ int janus_recorder_save_frame(janus_recorder *recorder, char *buffer, uint lengt
 		json_object_set_new(info, "c", json_string(recorder->codec));					/* Media codec */
 		if(recorder->fmtp)
 			json_object_set_new(info, "f", json_string(recorder->fmtp));				/* Codec-specific info */
+		if(recorder->description)
+			json_object_set_new(info, "desc", json_string(recorder->description));		/* Stream description */
```
lminiero, line 353: All other properties have a one-char attribute name, so I'd follow the same "rule": maybe `d` if it's not used yet?

lminiero (commented):

src/record.c
adnanel: Fixed, there is another value under the name "or", which is why I didn't realize there was a naming convention here.

src/plugins/janus_videoroom.c
adnanel: Fixed

src/plugins/janus_videoroom.c
adnanel: I turned this into a setter. Now that you mention it, I think this can be omitted. Since the value is copied when the recorder is instantiated, and that's enough, I believe?

src/plugins/janus_videoroom.c
```diff
@@ -7103,6 +7103,9 @@ static void janus_videoroom_recorder_create(janus_videoroom_publisher_stream *ps
 					janus_videoroom_media_str(ps->type));
 			}
 		}
+		/* If the stream has a description, store it in the recording */
+		if(ps->description && rc)
```
lminiero, line 7107: I think the `&& rc` part is unneeded, since the setter can do the `NULL` check internally.

src/record.c
```diff
@@ -292,6 +293,15 @@ int janus_recorder_add_extmap(janus_recorder *recorder, int id, const char *extm
 	return 0;
 }
 
+int janus_recorder_description(janus_recorder *recorder, const char *description) {
```
lminiero, line 296: You need a

	if(!recorder)
		return -1;

as the first check here.

src/plugins/janus_videoroom.c
adnanel: I removed this part, and it (still) works as expected.

src/record.c
```diff
@@ -292,6 +293,15 @@ int janus_recorder_add_extmap(janus_recorder *recorder, int id, const char *extm
...
+	if (!description)
+		return -1;
```
lminiero, line 298: I'd add a

	janus_mutex_lock_nodebug(&recorder->mutex);

here (to unlock at the end before returning), and before replacing the string I'd also add a

	if(g_atomic_int_get(&recorder->header))

to return immediately `0` in case the header was already written (as in that case the description was already used and there's no point updating it).

lminiero (commented):
Thanks! Added a few additional notes.

src/plugins/janus_videoroom.c
adnanel: Fixed

src/record.c
adnanel: Fixed

src/record.c
adnanel: done

lminiero:
Looks good, thanks: merging :+1:

PR #2952, opened 2022-04-19
Fix segmentation fault

adnanel:
We are seeing segmentation faults around this event.

Looking at the code, this seems like obvious culprit.

Haven't tested it yet, to be honest, because it's rather hard to reproduce for us, but it seems obvious enough?

In the meantime I'm also deploying this branch on our test environment and we'll see if it happens again.


lminiero:
I don't think that would fix it, since in both changes there's a dereference to `ice_handle` in a log line just a couple of lines above, meaning it would crash there anyway.

atoppi:
on top of that, I'm wondering how that pointer can be null, since we already check that in 
https://github.com/meetecho/janus-gateway/pull/2952/files#diff-f522ef48ad1b0f18f8ecc595a9d4ea827db81de4972fecc96497ff2fe2d277dbL3549

The same function contains a lot of (wrong?) checks for that ptr.

lminiero:
It's not NULL, in fact: maybe an early free? But without debugging information no idea.

atoppi:
@adnanel please share a libasan report, my guess is that the pointer is not NULL but simply freed

lminiero:
Please share it in a new issue, though, not this PR.

PR #2954, opened 2022-04-21
Making janus.js a modern library

dj-foxxy:
Hi, we use Janus extensively with browser clients with `html/janus.js`. We'd like to extend the JS so that our application can manage `MediaStream`s (it's an in-browser game engine). Rather than do the work separately and have our `janus.js` diverge from yours, I would really like to work with you to have improvements go back into the main repo.

As a first step, I've converted it to TypeScript. It's the absolute minimum required: I just changed the extension to `.ts`. Running `npm install && npm run build` from `npm` creates `janus.js` and a sym-link in `html` makes sure everything still works exactly as it did before.

If you're happy with this direction, I'll start typing the internals of  `janus.ts` in preparation for larger changes. I want to maintain 100% compatible with current API so that anything I do can be easily adopted.

Let me know what you think or if there's a better place to talk about this.

Thanks, Pete.


januscla:
Thanks for your contribution, @dj-foxxy! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Sorry, but we're not going to accept major changes to `janus.js`: if you want a modern library to use, please check [Janode](https://github.com/meetecho/janode/) instead, as that's where we'll put the efforts on a client side SDK. The demos and `janus.js` will stay as they are, and only act as basic (and dumb) examples, rather than references for production quality code.

PR #2963, opened 2022-04-29
0.x: Honor "audio", "data", "video" flag changes in subscriber updates.

fancycode:
If a `configure` request is sent for a subscriber with `update: true` and `video: false`, the flags are evaluated in https://github.com/fancycode/janus-gateway/blob/79f0b172be0fc1d9e45e6fa77935b336444ffda4/plugins/janus_videoroom.c#L7234-L7257 but then later might get overridden again.

### DIFF
--- a/plugins/janus_videoroom.c
+++ b/plugins/janus_videoroom.c
@@ -7454,9 +7454,9 @@ static void *janus_videoroom_handler(void *data) {
 						json_decref(jsep);
 						g_free(newsdp);
 						/* Any update in the media directions? */
-						subscriber->audio = publisher->audio && subscriber->audio_offered;
-						subscriber->video = publisher->video && subscriber->video_offered;
-						subscriber->data = publisher->data && subscriber->data_offered;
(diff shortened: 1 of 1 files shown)


fancycode:
The original `&&` could set `subscriber->audio` back to `TRUE` even if it was requested to be `FALSE` (and set to this value at https://github.com/fancycode/janus-gateway/blob/79f0b172be0fc1d9e45e6fa77935b336444ffda4/plugins/janus_videoroom.c#L7241) with the `update: true, audio: false` configure request. Same for `video` and/or `data`. Hope that makes it clearer...

lminiero:
Thanks for confirming that was the intention: it looks good to me, so I'll merge! :+1:

fancycode:
Thanks! I haven't checked `master`, but this probably needs to be changed there, too.

lminiero:
I don't think it matters there: while we support the legacy API there, it won't be forever, and the new API doesn't seem to be affected.

fancycode:
Perfect, thanks for confirming.

PR #2966, opened 2022-05-02
Handle publishing of multiple streams from same publisher

ngstwr:
It used to save the single published stream in config.myStream, and config.remoteStream for remote stream.

Now both has been changed to config.myStreams and config.remoteStreams respectively to hold an array of MediaStream.

In case of media.update offer, file will look for media.streamId to fetch the correct stream from config.myStreams array to update. Usually this stream ID will belong to an external stream which was already published earlier. 

In media.update request, if media.streamId is not provide then file will automatically take the first MediaStream inside config.myStreams array to update.

Useful Hint for Frontend Developers:
In order to provide the stream ID for the update offer, an object map of stream ID can be stored while publishing the external stream earlier.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1208,7 +1208,6 @@ function Janus(gatewayCallbacks) {
 						detached : false,
 						webrtcStuff : {
 							started : false,
-							myStream : null,
 							myStreams : [],
 							streamExternal : false,
 							remoteStreams : [],
--- a/html/janus.js
+++ b/html/janus.js
@@ -1209,8 +1209,9 @@ function Janus(gatewayCallbacks) {
 						webrtcStuff : {
 							started : false,
 							myStream : null,
+							myStreams : [],
 							streamExternal : false,
-							remoteStream : null,
+							remoteStreams : [],
 							mySdp : null,
 							mediaConstraints : null,
 							pc : null,
@@ -1283,9 +1284,9 @@ function Janus(gatewayCallbacks) {
 						detached : false,
 						webrtcStuff : {
 							started : false,
-							myStream : null,
+							myStreams : [],
 							streamExternal : false,
-							remoteStream : null,
+							remoteStreams : [],
 							mySdp : null,
 							mediaConstraints : null,
 							pc : null,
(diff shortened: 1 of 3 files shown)


januscla:
Thanks for your contribution, @ngstwr! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

html/janus.js
```diff
@@ -1731,16 +1731,31 @@ function Janus(gatewayCallbacks) {
...
 			Janus.debug("  -- Video tracks:", stream.getVideoTracks());
 		}
+
 		// We're now capturing the new stream: check if we're updating or if it's a new thing
 		var addTracks = false;
-		if(!config.myStream || !media.update || (config.streamExternal && !media.replaceAudio && !media.replaceVideo)) {
-			config.myStream = stream;
+
+		if(!config.myStreams.length || !media.update || (config.streamExternal && !media.replaceAudio && !media.replaceVideo)) {
+            if(stream) {
```
ycherniavskyi, line 1739: For sure here are some formatting issues. It seems that for intent you use spaces but the original code use tabs.

lminiero:
I'll need some time to review this, since as anticipated I'm also planning for a way to refactor janus.js to be able to push multiple streams rather than the more constraining approach we have now. I'm not sure the approach you're choosing now would fit what I'm designing, so I'll need a bit to better understand what it does. That said, as @ycherniavskyi you'll definitely need to use tabs and not spaces.

html/janus.js
ngstwr: Oh apologies, I will fix it as soon as I get time! 
Thanks for your feedback!

ngstwr:
@lminiero Oh that's great news! 
What I have done can be a temporary approach for now if the new version will take some time.
I would be happy to contribute on your ideas for Janus.js too if you would allow. I really like this project! 
About the indentation, I will fix it as soon as I get time tomorrow. 
Thanks for the feedback!

ngstwr:
@ycherniavskyi @lminiero Done! Fixed the indentation issue!

lminiero:
@ngstwr apologies for the radio silence on this, but as anticipated I've been working on a slight refactoring of `janus.js` that will actually change the way you can ask for things to be published, allowing for a much easier way of adding, replacing and removing streams and tracks. I hope a first version will be ready for testing in one or two weeks, so I'll make sure to mention it here, since I'll guess you'll be interested in the feature!

lminiero:
@ngstwr as promised, I just published a draft PR with the proposed changes. I'd appreciate some feedback, if you have time to look into this and check if it would address your needs!

lminiero:
Closing as this is being replaced by #3003.

PR #2972, opened 2022-05-10
Make SIP timer T1X64 configurable

oriol-c:
This PR adds the possibility to specify via configuration file the SIP timer T1x64 value used by sofia-sip (used for timers B, F, H, and J (UDP)). The default value of the timer is 32 seconds. 
It can be useful to lower that value in order to reduce the failover time when using a DNS SRV record with multiple priorities in proxy/outbound proxy parameters.


lminiero:
Oh I wasn't even aware that Sofia had this configuration knob: I agree it can be quite useful, so thanks for contributing an easy way to tweak that! I'll merge and backport it to the `0.x` branch as well :v:

PR #2977, opened 2022-05-11
videoroom: fix "not in a room" error code

tmatth:
JANUS_VIDEOROOM_ERROR_ALREADY_JOINED is misleading for this error case.

Reported by @chrisfillmore


lminiero:
We discussed this internally at the time, when we changed the message, and "already joined" seemed like a better choice, especially since the error message explicitly tells you to create a new handle: a "join first" error would be even more confusing, IMO, since a "join" is exactly what causes the error.

tmatth:
> We discussed this internally at the time, when we changed the message, and "already joined" seemed like a better choice, especially since the error message explicitly tells you to create a new handle: a "join first" error would be even more confusing, IMO, since a "join" is exactly what causes the error.

Oh sorry for the noise, I remembered the commit in question but was looking at the wrong thing.

PR #2978, opened 2022-05-11
Add kick_all requests and possibility to remove PIN code to both Audiobridge and Streaming plugins

mikaelnousiainen:
This PR adds the following features to both the Audiobridge and the Streaming plugins:

* A new `kick_all` request, which kicks and disconnects all participants in an Audiobridge room and disconnects all viewers/listeners in a Streaming plugin endpoint
* Option to remove an existing PIN code (by setting it to an empty string) from either an Audiobridge room or a Streaming plugin endpoint

Both of these features are very useful if one needs to be able to control the set of participants/viewers in these plugins for long-lived mountpoints and/or rooms.

My specific application is using these features for providing exclusive (reserved) access to audio streams by following this process:

1) Set a new PIN code to a Streaming plugin endpoint (or an Audiobridge room)
2) Distribute the new PIN code to the participants/viewers allowed to access these streams
3) Use the new `kick_all` requests to guarantee that there are no participants/viewers left after changing the PIN code
4) Allow the participants with the new PIN code to join/view
5) Once the exclusive access period is over, the PIN code is removed (set to empty string)

### DIFF
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -4508,21 +4508,15 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 			json_t *event = json_object();
 			json_object_set_new(event, "audiobridge", json_string("event"));
 			json_object_set_new(event, "room", string_ids ? json_string(room_id_str) : json_integer(room_id));
-			json_object_set_new(event, "kicked", string_ids ? json_string(user_id_str) : json_integer(user_id));
-			GHashTableIter iter;
-			gpointer value;
-			g_hash_table_iter_init(&iter, audiobridge->participants);
-			while(g_hash_table_iter_next(&iter, NULL, &value)) {
-				janus_audiobridge_participant *p = value;
-				JANUS_LOG(LOG_VERB, "Notifying participant %s (%s)\n", p->user_id_str, p->display ? p->display : "??");
-				int ret = gateway->push_event(p->session->handle, &janus_audiobridge_plugin, NULL, event, NULL);
-				JANUS_LOG(LOG_VERB, "  >> %d (%s)\n", ret, janus_get_api_error(ret));
-			}
+			json_object_set_new(event, "kicked_all", string_ids ? json_string(user_id_str) : json_integer(user_id));
+			JANUS_LOG(LOG_VERB, "Notifying participant %s (%s)\n", participant->user_id_str, participant->display ? participant->display : "??");
+			int ret = gateway->push_event(participant->session->handle, &janus_audiobridge_plugin, NULL, event, NULL);
+			JANUS_LOG(LOG_VERB, "  >> %d (%s)\n", ret, janus_get_api_error(ret));
 			json_decref(event);
 			/* Also notify event handlers */
 			if(notify_events && gateway->events_is_enabled()) {
 				json_t *info = json_object();
-				json_object_set_new(info, "event", json_string("kicked"));
+				json_object_set_new(info, "event", json_string("kicked_all"));
 				json_object_set_new(info, "room", string_ids ? json_string(room_id_str) : json_integer(room_id));
 				json_object_set_new(info, "id", string_ids ? json_string(user_id_str) : json_integer(user_id));
 				gateway->notify_event(&janus_audiobridge_plugin, session ? session->handle : NULL, info);
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -202,7 +202,7 @@ room-<unique room ID>: {
 	"secret" : "<room secret, mandatory if configured>",
 	"new_description" : "<new pretty name of the room, optional>",
 	"new_secret" : "<new password required to edit/destroy the room, optional>",
-	"new_pin" : "<new password required to join the room, optional>",
+	"new_pin" : "<new PIN required to join the room, PIN will be removed if set to an empty string, optional>",
 	"new_is_private" : <true|false, whether the room should appear in a list request>,
 	"new_record_dir" : "<new path where new recording files should be saved>",
 	"new_mjrs_dir" : "<new path where new MJR files should be saved>",
@@ -345,6 +345,29 @@ room-<unique room ID>: {
 	"secret" : "<room secret, mandatory if configured>",
 	"room" : <unique numeric ID of the room>,
 	"id" : <unique numeric ID of the participant to kick>
+}
+\endverbatim
+ *
+ * A successful request will result in a \c success response:
+ *
+\verbatim
+{
+	"audiobridge" : "success",
+}
+\endverbatim
+ *
+ * If you're the administrator of a room (that is, you created it and have access
+ * to the secret) you can kick all participants using the \c kick_all request. Notice
+ * that this only kicks all users out of the room, but does not prevent them from
+ * re-joining: to ban them, you need to first remove them from the list of
+ * authorized users (see \c allowed request) and then perform \c kick_all.
+ * The \c kick_all request has to be formatted as follows:
+ *
+\verbatim
+{
+	"request" : "kick_all",
+	"secret" : "<room secret, mandatory if configured>",
+	"room" : <unique numeric ID of the room>
 }
 \endverbatim
  *
@@ -3476,10 +3499,14 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
 			audiobridge->room_secret = new_secret;
 			g_free(old_secret);
 		}
-		if(pin && strlen(json_string_value(pin)) > 0) {
+		if(pin) {
 			char *old_pin = audiobridge->room_pin;
-			char *new_pin = g_strdup(json_string_value(pin));
-			audiobridge->room_pin = new_pin;
+			if(strlen(json_string_value(pin)) > 0) {
+				char *new_pin = g_strdup(json_string_value(pin));
+				audiobridge->room_pin = new_pin;
+			} else {
+				audiobridge->room_pin = NULL;
+			}
 			g_free(old_pin);
 		}
 		if(recdir) {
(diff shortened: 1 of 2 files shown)


src/plugins/janus_audiobridge.c
```diff
@@ -4412,6 +4439,106 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+			}
+			/* Notify all participants about the kick */
+			json_t *event = json_object();
+			json_object_set_new(event, "audiobridge", json_string("event"));
+			json_object_set_new(event, "room", string_ids ? json_string(room_id_str) : json_integer(room_id));
+			json_object_set_new(event, "kicked", string_ids ? json_string(user_id_str) : json_integer(user_id));
+			GHashTableIter iter;
+			gpointer value;
+			g_hash_table_iter_init(&iter, audiobridge->participants);
+			while(g_hash_table_iter_next(&iter, NULL, &value)) {
```
lminiero, line 4515: I think this is overkill. You're notifying everyone about every single kick, meaning an exponential number of notifications. Just send each kicked participant a notification that everyone has been kicked (e.g., a `kicked_all` event), without the while here.

src/plugins/janus_streaming.c
```diff
@@ -548,6 +548,26 @@ multistream-test: {
 {
 	"streaming" : "ok"
 }
+\endverbatim
+ *
+ * You can kick all viewers from a mountpoint using the \c kick_all request. Notice
+ * that this only removes all viewers, but does not prevent them from starting to watch
+ * the mountpoint again. The \c kick_all request has to be formatted as follows:
```
lminiero, line 555: I think this request is unneeded: we already have a `disable` request that can be used for this, and maybe just needs to be extended to close the PeerConnections when we're disabling a mountpoint. This gives you time to change the PIN and then re-enable the mountpoint.

lminiero (commented):
Thanks! I think this may be useful in VideoRoom and TextRoom as well, so in case I'll work on a similar enhancement for those too should this be merged.

That said, I think this patch needs work. First of all, I don't think this new functionality is needed at all in the Streaming plugin, since we have a request called "disable" that is very similar, and so maybe just needs to be extended a bit if it doesn't do exactly what you need. For the AudioBridge, as I pointed out in the review below, I'd definitely cut down the potentially huge number of notifications you're generating.

I'll do a deeper code review after the patch is updated.

src/plugins/janus_audiobridge.c
mikaelnousiainen: Will fix this. The code was copied and adapted from the `destroy` request.

src/plugins/janus_streaming.c
mikaelnousiainen: Currently, the `disable` request seems to be for a totally different use case. As far as I understand, it does not explicitly close the peer connections, but just disables the stream -- though I have to say I didn't test what really happens at that point to the viewers.

Anyway, with this use case, I really would not like to disable the endpoint at all.

Also, from the developer's viewpoint, having symmetric/similar requests in plugins is helpful IMO. And when thinking about the naming the meaning of the requests, having "kick all" as an option of a `disable` request sounds a bit artificial.

As a summary, I feel like `disable` and `kick_all` are two entirely different pieces of functionality and they complement each other rather than being similar.

Thoughts?

lminiero: The Streaming plugin is completely different from AudioBridge, VideoRoom and TextRoom: it doesn't have a concept of room at all, so I think "kick" makes little sense. Since the context is completely different, symmetric requests are less important IMO.

It makes much more sense to me to disable a mountpoint and have that feature optionally close PeerConnections (which is something a few asked for in the past, I can't remember why that didn't happen eventually): since what you're trying to do is kicking everyone out and preventing them access until you change something, "disable" is actually a much better fit.

mikaelnousiainen: Ok, so could the `disable` request then contain options for:

1) closing all connections (not calling it kicking anymore), e.g. `close_connections` - defaults to `false` to preserve existing behavior
2) have an option NOT to disable the stream, e.g. `disable_stream` - defaults to `true`, disabling the stream to preserve existing behavior

Because in many use cases that would follow what I'm doing, disabling the stream is completely unnecessary, as the PIN is already set. Then I'd need to re-enable the stream, which is again an unnecessary request.

lminiero: > Because in many use cases that would follow what I'm doing, disabling the stream is completely unnecessary, as the PIN is already set. Then I'd need to re-enable the stream, which is again an unnecessary request.

Ah yes, I hadn't considered this. In that case, I guess a request that just kicks viewers makes sense. You still need to clarofy in both the documentation and code that this won't work for all types of mountpoints, though: on demand mountpoints for instance don't have a group of viewers, but spawn a different thread for each of them, since each has a separate context of the same source media.

mikaelnousiainen: Or altervatively `disable_stream` could be called `close_only`, defaulting to `false` which would indicate that connections should be closed only.

What I'm after is forcing users to perform unnecessary operations here. Doesn't really matter what they're called :)

lminiero: No, that wouldn't be a `disable` then, it would be highjacking a different request to then only do something else. I guess kick is fine :smile:

src/plugins/janus_audiobridge.c
mikaelnousiainen: I've removed the inner loop now, please have a look.

src/plugins/janus_streaming.c
mikaelnousiainen: Ok, so I'll be keeping the separate `kick_all` request after all now :)

Can you expand a bit more on what to put into the documentation? I have to say I'm not aware how the plugin works in these different cases. Or alternatively, could we make `kick_all` work for on-demand mountpoints too?

lminiero: I'd simply say that the request works with all mountpoint types except the ondemand ones. I wouldn't waste time trying to get the method to work with those too, since they don't use the viewers list as the others do, and don't keep track of who's receiving it: each viewer has their own thread and their own playback context.

mikaelnousiainen: :+1: Done!

src/plugins/janus_audiobridge.c
```diff
@@ -4412,6 +4439,100 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+		if(audiobridge == NULL) {
+			janus_mutex_unlock(&rooms_mutex);
+			error_code = JANUS_AUDIOBRIDGE_ERROR_NO_SUCH_ROOM;
+			JANUS_LOG(LOG_ERR, "No such room (%s)\n", room_id_str);
+			g_snprintf(error_cause, 512, "No such room (%s)", room_id_str);
+			goto prepare_response;
+		}
+		janus_refcount_increase(&audiobridge->ref);
+		janus_mutex_lock(&audiobridge->mutex);
+		janus_mutex_unlock(&rooms_mutex);
```
lminiero, line 4482: I'm not convinced by the room lock, followed by the global unlock. I see the same happens in "kick", so I guess it was copied from there, but it sounds dangerous to me, since it's not what we do in other places, and I wonder if it could cause some reverse-order locks in some specific circumstances that might lead to a deadlock. I'd change both "kick" and "kick_all" so that the room lock is done after the global unlock and not before (the global lock is only needed so that we can ref the audiobridge instance).

src/plugins/janus_streaming.c
```diff
@@ -548,6 +548,27 @@ multistream-test: {
...
+	"id" : <unique ID of the mountpoint to disable; mandatory>,
+	"secret" : "<mountpoint secret; mandatory if configured>",
+}
+\endverbatim
+ *
+ * If successful, a generic \c ok is returned:
+ *
+\verbatim
+{
+	"streaming" : "ok",
```
lminiero, line 570: I'd rather this be called "kicked", to be more in line with our response to, e.g., "create", "edit" and "destroy". I know we return "ok" in a few other synchronous requests, but I'm not fond of those, so for new requests we can try do something nicer :grin:

src/plugins/janus_streaming.c
```diff
@@ -4404,6 +4429,119 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
...
+				mp->viewers = g_list_remove_all(mp->viewers, s);
+				viewer = g_list_first(mp->viewers);
+				continue;
+			}
+            janus_mutex_lock(&s->mutex);
+			if(s->mountpoint != mp) {
+				mp->viewers = g_list_remove_all(mp->viewers, s);
+				viewer = g_list_first(mp->viewers);
+                janus_mutex_unlock(&s->mutex);
+				continue;
```
lminiero, line 4496: Looks like indentation is broken in all the block above. Please notice we use tabs, not spaces. Not sure if the same is happening in other changes as well, since they're not easy to spot from the diff view in Github.

src/plugins/janus_streaming.c
```diff
@@ -4404,6 +4429,119 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
...
+							break;
+						}
+						janus_mutex_unlock(&ht->mutex);
+						l = l->next;
+					}
+				}
+			}
+			mp->viewers = g_list_remove_all(mp->viewers, s);
+			viewer = g_list_first(mp->viewers);
+            janus_mutex_unlock(&s->mutex);
```
lminiero, line 4528: Broken indentation again.

src/plugins/janus_streaming.c
```diff
@@ -4404,6 +4429,119 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
...
+			if(s->mountpoint != mp) {
+				mp->viewers = g_list_remove_all(mp->viewers, s);
+				viewer = g_list_first(mp->viewers);
+                janus_mutex_unlock(&s->mutex);
+				continue;
+			}
+			g_atomic_int_set(&s->stopping, 1);
+			g_atomic_int_set(&s->started, 0);
+			g_atomic_int_set(&s->paused, 0);
+			s->mountpoint = NULL;
```
lminiero, line 4501: This is where the lock on `s->mutex` should be released, it's not needed anymore after that.

src/plugins/janus_streaming.c
```diff
@@ -4404,6 +4429,119 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
...
+			/* Tell the core to tear down the PeerConnection, hangup_media will do the rest */
+			gateway->push_event(s->handle, &janus_streaming_plugin, NULL, event, NULL);
+			gateway->close_pc(s->handle);
+			janus_refcount_decrease(&s->ref);
```
lminiero, line 4505: This decrease should happen later, because we're still using it in the following block when we update helper threads, and there's an unlock after that which dereferences it. I see you copied that block from the "destroy" code, but it's not a problem there because it doesn't have to extra unlock on `s->mutex` after that (which I think is unneeded anyway).

src/plugins/janus_streaming.c
```diff
@@ -4404,6 +4429,119 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
...
+		if(notify_events && gateway->events_is_enabled()) {
+			json_t *info = json_object();
+			json_object_set_new(info, "event", json_string("kicked_all"));
+			json_object_set_new(info, "id", string_ids ? json_string(id_value_str) : json_integer(id_value));
+			gateway->notify_event(&janus_streaming_plugin, session ? session->handle : NULL, info);
+		}
+		janus_mutex_unlock(&mountpoints_mutex);
+		/* Send info back */
+		response = json_object();
+		json_object_set_new(response, "streaming", json_string("ok"));
```
lminiero, line 4543: See above on response name.

lminiero (commented):

mikaelnousiainen:
@lminiero Changes done, please have a look at the refcounts and mutexes, as I'm not 100% sure about the ordering.

Also, in Streaming plugin `destroy` event handling, there is mutex locking on `session->mutex` too, like this: https://github.com/meetecho/janus-gateway/pull/2978/files#diff-3d84c5c61736914d38e005f37baaecfa324295ff90a45962773a0cb90520f2beL4482

Is this needed with `kick_all`? I'm simply locking each individual session for every viewer, as they are modified.

lminiero:
Just as a quick note, I haven't forgotten about this! I'll have to review the use of mutexes as you suggested, and I'll do it in the next few days.

src/plugins/janus_streaming.c
```diff
@@ -4404,6 +4429,119 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
...
+			if(s == NULL) {
+				mp->viewers = g_list_remove_all(mp->viewers, s);
+				viewer = g_list_first(mp->viewers);
+				continue;
+			}
+			janus_mutex_lock(&s->mutex);
+			if(s->mountpoint != mp) {
+				mp->viewers = g_list_remove_all(mp->viewers, s);
+				viewer = g_list_first(mp->viewers);
+				janus_mutex_unlock(&s->mutex);
```
lminiero, line 4495: This has made me realize that in the management of `"destroy"` we're doing the wrong thing, as we're locking on `session` instead of `s` as we should. I'll have to fix that, since `session` may be `NULL` (e.g., in case of synchronous requests coming from the Admin API).

lminiero (commented):
Thanks for the changes! They all look fine to me. The note you see below is a memo to me, so something I have to fix in the existing code myself.

I'll make a couple of tests later, and in case I don't see issues, I'll merge :+1:

src/plugins/janus_streaming.c
lminiero: @mikaelnousiainen it looks like the fix I mentioned above (that I just pushed) caused a conflict in your PR, could you fix that? Sorry for the inconvenience!

mikaelnousiainen: @lminiero Ok, I'll check and try to fix the conflict -- noticed your message only now :)

mikaelnousiainen: Also, just a note, I've been using this code in my own tests and it seems to work very well!

mikaelnousiainen: @lminiero Conflicts fixed, I think this should be good to merge!

lminiero: Thanks, I'll review and in case merge later today :+1:

src/plugins/janus_streaming.c
```diff
@@ -4516,7 +4652,9 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 			}
 			mp->viewers = g_list_remove_all(mp->viewers, s);
 			viewer = g_list_first(mp->viewers);
-			janus_mutex_unlock(&s->mutex);
+			janus_mutex_unlock(&session->mutex);
```
lminiero, line 4655: This should be `s`, not `session`. I fixed this in a commit recently.

mikaelnousiainen: Good catch! Fixed.

PR #2986, opened 2022-05-25
Support for batched configure requests in VideoRoom

lminiero:
The `configure` request in the VideoRoom plugin can be used for different purposes, e.g., pausing/resuming streams, choosing simulcast layers, etc. So far, since we merged multistream, this was a drilldown request, meaning you had to specify the `mid` to act upon: this also meant that if you wanted to make multiple changes on different m-lines at the same time, you had to send different `configure` requests.

This PR addresses that by adding a batched mode to `configure` instead, where you pass a stream called `arrays` with all the individual changes you want to make, e.g.:

    subscription.send({message: {request: "configure", streams: [{mid: "0", send: true}, {mid: "1", send: true}]}})

I've tested this briefly and it seems to work as expected, but of course more feedback would be helpful. Pinging @mirkobrankovic in particular as he expressed interested in something like this a few days ago.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,20 +7,20 @@ All notable changes to this project will be documented in this file.
 
 - Abort DTLS handshake if DTLSv1_handle_timeout returns an error
 - Fixed rtx not being offered on Janus originated PeerConnections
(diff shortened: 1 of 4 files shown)


lminiero:
Merging.

PR #2989, opened 2022-05-30
Added suspend/resume participant API to AudioBridge

lminiero:
This patch adds a simple functionality to the AudioBridge, called `suspend`, where participants can be temporarily "unhooked" from a mixer instance, while still remaining in the room and keeping the PeerConnection active: the end result is that, when suspended, incoming audio from the participant is discarded, and no audio is sent back to them either. The `resume` request can then re-activate their PeerConnection. The main aim of this patch is allowing for a quick and easy way to have inactive participants not weigh on the CPU, and quickly bring them back in when needed without a costly (in time) new PeerConnection.

This is currently based on `0.x` but I'll port it to `master` as well as soon as it is merged.


jarsj:
@lminiero any timelines on when this can be merged to 0.x ?

lminiero:
@jarsj there's no timeline, since we received to feedback :smile: 
Should you decide to give it a try, we'd appreciate feedback on whether it works as expected for you, and more importantly if you notice any regression in existing functionality. I'd like to be able to merge soon but that won't happen unless other people beside me see value in this and test it.

lminiero:
Merging.

PR #2994, opened 2022-06-07
Make Streaming plugin use SDP utils, and codecs instead of rtpmaps

lminiero:
The Streaming plugin was kind of an anomaly in the plugins we ship out of the box: in fact, while we've had SDP utilities for a long time, the Streaming plugin was still the only one still crafting SDPs manually appending strings (well the TextRoom too, but that doesn't really count, since it's pretty static SDP-wise). This was obviously quite suboptimal and error prone, so I decided to make sure the Streaming plugin would rely on the SDP utils as well, to ensure some consistency in how we deal with SDPs.

This required an additional change, though. In fact, since the very beginning the Streaming plugin has required you to provide rtpmaps when creating mountpoints (e.g., "opus/48000/2" or "VP8/90000"), since those are used in the SDP and so they were easier to just past in the manually crafter SDP. While this "works", it's very error prone, since it's easy to miswrite an rtpmap (e.g., "VP8/9000" with one less 0) or to try and offer a codec that Janus doesn't support, which means it's easy to end up with broken mountpoints that can't be used. Besides, the SDP utils in Janus work on codec names, since they enforce the right rtpmaps automatically. As a consequence, this patch also modifies the Streaming plugin API so that codec names can be passed instead, which solves both issues. Of course, to preserve backwards compatibility, mountpoints can still be created by passing an rtpmap as before: in that case, the plugin automatically translates them to known codecs.

I have made some tests and it seems to work nicely, even though I haven't tested everything yet: I haven't checked if this causes any regression in RTSP, for instance, so that will need to be tested as well before merging. If you care about the Streaming plugin and use it a lot, please do test these changes and let us know if there's anything that needs fixing. A code review might help as well, since there are many small changes that could be affected by some typos that I may have overlooked: I checked many already, but a pair of fresh eyes will definitely help.


lminiero:
Merging.

PR #2995, opened 2022-06-07
Videoroom: Fix RID memory leak

OxleyS:
When processing an offer from a publisher using RID-based simulcast, Videoroom calls `janus_rtp_simulcast_prepare()` for the stream associated with every m-line in the offer, even if that m-line had already appeared in a previous offer. This means the RID array is repopulated every time, but the existing values were not being freed before being overwritten. This leads to a memory leak for every successive renegotiation.

Here I just unconditionally free the previous values beforehand, assuming they need to be completely repopulated anyway. While we're at it here, should we also clear `rid_ext_id` or the SSRC array somehow? (Although they're not leaking memory, obviously)

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -2142,7 +2142,7 @@ typedef struct janus_videoroom_publisher_stream {
 	gboolean opusstereo;					/* Whether this publisher is doing stereo Opus */
 	gboolean simulcast, svc;				/* Whether this stream uses simulcast or VP9 SVC */
 	uint32_t vssrc[3];						/* Only needed in case VP8 (or H.264) simulcasting is involved */
-	janus_mutex rid_mutex;    /* Protects access to the rid array and the extmap ID  */
+	janus_mutex rid_mutex;		/* Protects access to the rid array and the extmap ID  */
 	char *rid[3];							/* Only needed if simulcasting is rid-based */
 	int rid_extmap_id;						/* rid extmap ID */
 	/* RTP extensions, if negotiated */
@@ -2713,9 +2713,7 @@ static janus_videoroom_rtp_forwarder *janus_videoroom_rtp_forwarder_add_helper(j
 		forward->simulcast = TRUE;
 		janus_rtp_switching_context_reset(&forward->context);
 		janus_rtp_simulcasting_context_reset(&forward->sim_context);
-		janus_mutex_lock(&ps->rid_mutex);
 		forward->sim_context.rid_ext_id = ps->rid_extmap_id;
-		janus_mutex_unlock(&ps->rid_mutex);
 		forward->sim_context.substream_target = 2;
 		forward->sim_context.templayer_target = 2;
 	}
@@ -3036,9 +3034,7 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(
 	janus_refcount_init(&stream->ref, janus_videoroom_subscriber_stream_free);
 	janus_refcount_increase(&stream->ref);	/* This is for the mid-indexed hashtable */
 	janus_rtp_simulcasting_context_reset(&stream->sim_context);
-	janus_mutex_lock(&ps->rid_mutex);
 	stream->sim_context.rid_ext_id = ps->rid_extmap_id;
-	janus_mutex_unlock(&ps->rid_mutex);
 	stream->sim_context.substream_target = 2;
 	stream->sim_context.templayer_target = 2;
 	janus_vp8_simulcast_context_reset(&stream->vp8_context);
@@ -3091,9 +3087,7 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add_
 				stream->opusfec = ps->opusfec;
 				janus_rtp_simulcasting_context_reset(&stream->sim_context);
 				if(ps->simulcast) {
-					janus_mutex_lock(&ps->rid_mutex);
 					stream->sim_context.rid_ext_id = ps->rid_extmap_id;
-					janus_mutex_unlock(&ps->rid_mutex);
 					stream->sim_context.substream_target = 2;
 					stream->sim_context.templayer_target = 2;
 				}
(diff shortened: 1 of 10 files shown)


lminiero (commented):
I'm not sure that's something that should be done there: `janus_rtp_simulcasting_prepare` doesn't own the rid array, which is passed to it by someone who calls the method (e.g., plugins). For all the method knows, those might not be malloc-ed, but allocated in the heap, which as a result would crash Janus. On the other end, the method does do a `g_strdup` when a `rid` is found, meaning it's allocating stuff on behalf of the caller, so it might indeed make sense.

I've added a small editorial note inline. On your question, I guess that if we reset rids, it does indeed make sense to reset `rid_ext_id` to `-1` as well.

src/rtp.c
```diff
@@ -1020,11 +1020,18 @@ void janus_rtp_simulcasting_context_reset(janus_rtp_simulcasting_context *contex
 void janus_rtp_simulcasting_prepare(json_t *simulcast, int *rid_ext_id, uint32_t *ssrcs, char **rids) {
 	if(simulcast == NULL)
 		return;
+	/* Clear existing RIDs in case this array was reused from a previous call */
+	size_t i = 0;
+	for(i=0; i<3; i++) {
```
lminiero, line 1025: Your new commit made me realize that you should have an `if(rids)` for this block as well, since it may be `NULL`.

OxleyS: Ah, right, sorry, good catch.

OxleyS:
I agree that there is some ambiguity here with regards to who owns the memory, as currently the caller is also responsible for freeing the RID array contents when e.g. the session is destroyed. Probably the most unambiguous and least code-duplicating resolution would to have the `janus_rtp_simulcast` function family handle the entire lifetime of both the RID array and its contents, but that would be a much larger change.

OxleyS:
Hmm, that is indeed a possible race. We could mitigate/maybe-solve the issue by not touching the output RIDs array in `janus_rtp_simulcasting_prepare` if the set of RIDs has not changed between calls. As I understand it, both libwebrtc and browser APIs will not let you change the RIDs nor simulcast layer count for a transceiver after they have been set anyway, so this may be a relatively safe assumption to make (we could even make it an error case?).

lminiero:
I'm not sure we can ensure that: rids can be added (because you add substreams), and if you add a lower substream than the ones you had before, the existing rids would be shifted. Besides, m-lines can be recycled (simulcast video published, then video removed, then another simulcast video published), which means there can be bigger changes.

lminiero:
> The second one is tricky because if we simply protect all access to the RID array via a plugin-level mutex, then we're forced to either take the mutex on every incoming simulcast RTP packet, which would hurt performance, or pass that mutex into `janus_rtp_simulcasting_context_process_rtp` and have it lock it only when needed, which is not pretty.

I agree it's not pretty but I think it's not that ugly either... and would ensure we only use the mutex when we really have to, which wouldn't be often. It would indeed mean a change in the signature of the function, though, so a breaking change for custom plugins that rely on it, but if we update all our plugins to use it I don't particularly care. It would be important to bump the plugin version in `plugin.h`, though, so that older plugins have to be recompiled and so need to be aware of the change.

OxleyS:
Alright, I'll give that a shot.

OxleyS:
I've been looking through the other plugins, and it seems the assumption that the RIDs array can be safely read from unprotected is fairly deep-set into them. I don't have confidence that this is something I can address without either being lock-happy and hurting performance, or risking breaking something. Some plugins (namely, Duktape and Lua) I don't have the means to properly test in any case.

In light of this, I believe it's best for me to back out of freeing the RIDs in `janus_rtp_simulcasting_prepare()`, and to instead do so at the plugin level just before calling this function. This means the memory leak will be fixed in Videoroom but remain in other plugins, which will have to be fixed further down the road.

I have, however, implemented the change to `janus_rtp_simulcasting_context_process_rtp()` as discussed, where it takes a new mutex argument that it locks when it has to read from the RIDs array. I've accordingly bumped the plugin version and updated non-Videoroom plugins to just pass `NULL` for the argument, preserving existing behavior.

lminiero:
> Maybe an easier way to address this might be to change the `janus_rtp_simulcasting_context` struct so that it hosts the SSRC and rid array itself

Mh I don't think that will work, mostly because the ssrc and rid array have nothing to do with the `janus_rtp_simulcasting_context` struct: the fact the function is called `janus_rtp_simulcasting_prepare` is just because the "topic" is the same, but they're not related at all. This becomes evident when the same simulcast source is shared among multiple recipients (e.g., VideoRoom), and so where the same arrays are used for all recipients, each with their own `janus_rtp_simulcasting_context` struct, while the publisher doesn't have one.

I'll add a few notes inline to your latest changes shortly.

src/plugins/janus_videoroom.c
```diff
@@ -2711,7 +2713,9 @@ static janus_videoroom_rtp_forwarder *janus_videoroom_rtp_forwarder_add_helper(j
 		forward->simulcast = TRUE;
 		janus_rtp_switching_context_reset(&forward->context);
 		janus_rtp_simulcasting_context_reset(&forward->sim_context);
+		janus_mutex_lock(&ps->rid_mutex);
 		forward->sim_context.rid_ext_id = ps->rid_extmap_id;
+		janus_mutex_unlock(&ps->rid_mutex);
```
lminiero, line 2718: I don't think you need a lock here, it feels overkill to me.

src/plugins/janus_videoroom.c
```diff
@@ -6991,21 +7001,25 @@ void janus_videoroom_incoming_rtp(janus_plugin_session *handle, janus_plugin_rtp
 				sc = 1;
 			else if(ssrc == ps->vssrc[2])
 				sc = 2;
-			else if(ps->rid_extmap_id > 0) {
+			else {
```
lminiero, line 7004: For the reason above, I don't think you need to split the if that way to protect the ID, it's enough to just protect the "parse rid" method.

lminiero (commented):
Added a few notes, mostly editorial, but also on where I don't think we need a mutex. I'll have another look at the finalized changes after that, so that I can check what I'll need to change in the other plugins. I think I'll probably add a helper for cleaning up rids, as it feels like something we can make a function of (a bit like the prepare one).

OxleyS:
Thanks for looking at this again, I've pushed a commit that addresses all the points above.

lminiero:
Thanks! This looks good to me: I'll merge and start updating the other plugins too then.

lminiero:
@OxleyS I should have done everything in the above commit, which now includes a new `janus_rtp_simulcasting_cleanup` function to selectively reset some properties, with or without a mutex. If you can give it a glance just to make sure I didn't mess anything up it would be really appreciated!

PR #3000, opened 2022-06-10
Send an "edited" event to all viewers when an "edit" event occurred in the streaming plugin

amoizard:
When someone sends an "edit" request to modify a streaming plugin mountpoint, viewers are not aware of the "edit".

This PR is used within our deployments so that all viewers will receive an "edited" event containing the id, permanent and metadata parameters. There are 2 purposes: being aware of a change for the mountpoint definition and retrieving immediately the updated metadata, if any.

> {
>         "streaming" : "edited",
>         "id" : <unique ID of the just edited mountpoint>,
>         "permanent" : <true|false, depending on whether the changes were saved to configuration file or not>
>         "metadata" : "<updated metadata for the mountpoint; optional>",
> }

Feel free to send any comments.
Kind Regards,

### DIFF
--- a/src/plugins/janus_streaming.c
+++ b/src/plugins/janus_streaming.c
@@ -467,7 +467,7 @@ multistream-test: {
 	"new_pin" : "<new PIN for the mountpoint, PIN will be removed if set to an empty string; optional>",
 	"new_is_private" : <true|false, depending on whether the mountpoint should be now listable; optional>,
 	"permanent" : <true|false, whether the mountpoint should be saved to configuration file or not; false by default>,
-	"edited_event" : <true|false, wheter janus will send an edited event to all viewers when metadata is updated; false by default>
+	"edited_event" : <true|false, whether an event will be sent to all viewers when metadata is updated; false by default>
 }
 \endverbatim
  *
@@ -481,7 +481,8 @@ multistream-test: {
 }
 \endverbatim
  *
- * A successful \c edit will also result in an \c edited event sent to all viewers when the metadata has changed:
+ * In case \c edited_event was set to \c true in the request, a successful \c edit will
+ * also result in an \c edited event sent to all viewers when the metadata has changed:
  *
 \verbatim
 {
@@ -1006,10 +1007,12 @@ static struct janus_json_parameter adminkey_parameters[] = {
 };
 static struct janus_json_parameter edit_parameters[] = {
 	{"new_description", JSON_STRING, 0},
+	{"new_metadata", JSON_STRING, 0},
 	{"new_secret", JSON_STRING, 0},
 	{"new_pin", JSON_STRING, 0},
 	{"new_is_private", JANUS_JSON_BOOL, 0},
-	{"permanent", JANUS_JSON_BOOL, 0}
+	{"permanent", JANUS_JSON_BOOL, 0},
+	{"edited_event", JANUS_JSON_BOOL, 0}
 };
 static struct janus_json_parameter create_parameters[] = {
 	{"name", JSON_STRING, 0},
@@ -4239,14 +4242,8 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 			char *old_metadata = mp->metadata;
 			char *new_metadata = g_strdup(json_string_value(md));
 			mp->metadata = new_metadata;
-			if (send_edited_event == TRUE) {
-				if (old_metadata == NULL && new_metadata == NULL) {
-					send_edited_event = FALSE;
-				}
-				if (old_metadata != NULL && new_metadata != NULL && !strcmp(old_metadata, new_metadata)) {
-					send_edited_event = FALSE;
-				}
-			}
+			if(send_edited_event == TRUE && ((old_metadata == NULL && new_metadata == NULL) || (old_metadata != NULL && new_metadata != NULL && !strcmp(old_metadata, new_metadata))))
+				send_edited_event = FALSE;
 			g_free(old_metadata);
 		}
 		if(is_private)
@@ -4446,7 +4443,7 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 		}
 
 		/* Also notify viewers */
-		if (send_edited_event) {
+		if(send_edited_event) {
 			json_t *info = json_object();
 			json_object_set_new(info, "event", json_string("edited"));
 			json_object_set_new(info, "id", string_ids ? json_string(mp->id_str) : json_integer(mp->id));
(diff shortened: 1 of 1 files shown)


amoizard:
Hi Lorenzo,

Following your advise, I pushed another commit:

- I have added an optional **edited_event** (false by default) in the **edit** request.
- I have also made sure that the **edited** event will not be sent if there is no update of the metadata parameter.
- I have fixed the documentation by removing the extra **permanent** parameter.
- I finally updated the documentation with the new **edited_event** parameter for the **edit** request.

Thanks!

src/plugins/janus_streaming.c
```diff
@@ -466,7 +466,8 @@ multistream-test: {
 	"new_secret" : "<new secret for the mountpoint; optional>",
 	"new_pin" : "<new PIN for the mountpoint, PIN will be removed if set to an empty string; optional>",
 	"new_is_private" : <true|false, depending on whether the mountpoint should be now listable; optional>,
-	"permanent" : <true|false, whether the mountpoint should be saved to configuration file or not; false by default>
+	"permanent" : <true|false, whether the mountpoint should be saved to configuration file or not; false by default>,
+	"edited_event" : <true|false, wheter janus will send an edited event to all viewers when metadata is updated; false by default>
```
lminiero, line 470: There's a typo on "whether". And it's not "janus" that sends the event: I'd just generically say

> whether an event will be sent...

src/plugins/janus_streaming.c
```diff
@@ -478,6 +479,16 @@ multistream-test: {
 	"id" : <unique ID of the just edited mountpoint>,
 	"permanent" : <true|false, depending on whether the changes were saved to configuration file or not>
 }
+\endverbatim
+ *
+ * A successful \c edit will also result in an \c edited event sent to all viewers when the metadata has changed:
```
lminiero, line 484: This is missing a prefix like

> In case \c edited_event was set to \c true in the request, a successful...

As it is right now, the text suggests that the only required precondition is metadata being edited.

src/plugins/janus_streaming.c
```diff
@@ -4178,6 +4189,8 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 		json_t *pin = json_object_get(root, "new_pin");
 		json_t *is_private = json_object_get(root, "new_is_private");
 		json_t *permanent = json_object_get(root, "permanent");
+		json_t *edited_event = json_object_get(root, "edited_event");
+		gboolean send_edited_event = edited_event ? json_is_true(edited_event) : FALSE;
```
lminiero, line 4193: Since you're adding a new property, it must be added to the object that performs the automated valitation checks on incoming JSON payloads. Look for `static struct janus_json_parameter edit_parameters[]` in the code, where other existing parameters are defined already.

src/plugins/janus_streaming.c
```diff
@@ -4226,6 +4239,14 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
 			char *old_metadata = mp->metadata;
 			char *new_metadata = g_strdup(json_string_value(md));
 			mp->metadata = new_metadata;
+			if (send_edited_event == TRUE) {
```
lminiero, line 4242: Code style not: please remove the space between `if` and bracket (same for other occurrences below).

src/plugins/janus_streaming.c
```diff
@@ -4226,6 +4239,14 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
...
+				if (old_metadata == NULL && new_metadata == NULL) {
+					send_edited_event = FALSE;
+				}
+				if (old_metadata != NULL && new_metadata != NULL && !strcmp(old_metadata, new_metadata)) {
+					send_edited_event = FALSE;
+				}
```
lminiero, line 4248: Code style nit: both checks above have a simple one-line action, so we usually don't put curly brackets in that case. It looks like you can combine the two ifs in a single one with an OR, though, e.g.:

	if ((old_metadata == NULL && new_metadata == NULL) ||
			(old_metadata != NULL && new_metadata != NULL && !strcmp(old_metadata, new_metadata)))
		send_edited_event = FALSE;

src/plugins/janus_streaming.c
```diff
@@ -4423,6 +4444,30 @@ static json_t *janus_streaming_process_synchronous_request(janus_streaming_sessi
...
+		/* Also notify viewers */
+		if (send_edited_event) {
+			json_t *info = json_object();
+			json_object_set_new(info, "event", json_string("edited"));
+			json_object_set_new(info, "id", string_ids ? json_string(mp->id_str) : json_integer(mp->id));
+			if(mp->metadata)
+				json_object_set_new(info, "metadata", json_string(mp->metadata));
+			GList *viewer = g_list_first(mp->viewers);
+			while(viewer) {
+				janus_streaming_session *session = (janus_streaming_session *)viewer->data;
```
lminiero, line 4457: I think you're shadowing the existing `session` variable in `janus_streaming_process_synchronous_request` by calling it the same way. Better to use a different name, e.g., `janus_streaming_session *s`, which is what we do in the management of other synchronous requests that result in notifications.

lminiero (commented):
Thanks for the fixes! I've noticed a few other things, which you can find below. Mostly are editorial, so nothing big!

src/plugins/janus_streaming.c
amoizard: Typo fixed!

src/plugins/janus_streaming.c
amoizard: I have updated the text.

src/plugins/janus_streaming.c
amoizard: Should be ok now.

src/plugins/janus_streaming.c
amoizard: Variable renamed.

src/plugins/janus_streaming.c
amoizard: I have combined all the 3 ifs... I'm unsure about coding style here... Any URL for coding style?

amoizard:
The newer commit should takes into account all your comments.

I also added new_metadata to the object that performs the automated valitation checks on incoming JSON payloads.
It seems it was also missing there?

src/plugins/janus_streaming.c
amoizard: This is resolved. I have also added the new_metadata parameter which seems to be also missing.

src/plugins/janus_streaming.c
lminiero: Fair point :smile:  We don't have a specific set of rules, even if @atoppi tried to codify them in #2910. In general it's not a big deal for us anyway, you'll see several pieces of code that don't comply for instance: I just happen to point out those that I notice in order to try and provide some consistency.

lminiero:
> I also added new_metadata to the object that performs the automated validation checks on incoming JSON payloads.
> It seems it was also missing there?

Thanks and good catch!

lminiero:
Thanks again, merging then :+1:

PR #3012, opened 2022-06-29
enable recording on session start

adnanel:
Currently, Janus starts recording for a participant in 3 scenarios:

- When configuring a new participant
    - In our case, recording didn’t start here because the room hasn’t started recording yet.
- When processing an SDP
    - In our case, recording didn’t start here because the room hasn’t started recording yet.
- When enable_recording is requested.
    - In our case, recording didn’t start here because the session hasn’t started yet (no WEBRTC media)

A moment later, WebRTC media is available and the session is marked as active. However, even though the room has the flag set that it should be recorded, and media is being published, no recording is being generated for the participant in question.

In this PR we simply add a check when the session starts, to also start the recording, if the room has recording enabled.

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -6873,7 +6873,7 @@ void janus_videoroom_setup_media(janus_plugin_session *handle) {
 				GList *temp = participant->streams;
 				while(temp) {
 					janus_videoroom_publisher_stream *ps = (janus_videoroom_publisher_stream *)temp->data;
-					janus_videoroom_recorder_create(participant, ps);
+					janus_videoroom_recorder_create(ps);
 					temp = temp->next;
 				}
 				participant->recording_active = TRUE;


src/plugins/janus_videoroom.c
```diff
@@ -6866,6 +6866,19 @@ void janus_videoroom_setup_media(janus_plugin_session *handle) {
 			/* Notify all other participants that there's a new boy in town */
 			janus_videoroom_notify_about_publisher(participant, FALSE);
 			janus_refcount_decrease(&participant->ref);
```
lminiero, line 6868: You should move all your new code before this line, that is before we decrease the participant's ref. For all we know, the participant may be destroyed after this point.

adnanel:
Thanks for the feedback - I applied all suggestions and it should be OK now.

As for the case:
> the recording is enabled in the time between we prepare the answer, and the PeerConnection actually goes up.

I believe so. Judging by the flags we logged when troubleshooting this, it seems that "enable_recording" was sent after offer/answer, but before any media is actually transmitted (session is not marked as started).

adnanel:
For some reason my most recent commit isn't being added in the PR, I will re-create the PR and continue the discussion there.

PR #3013, opened 2022-06-29
enable recording on session start

adnanel:
(Continuation of #3012)

Currently, Janus starts recording for a participant in 3 scenarios:

- When configuring a new participant
    - In our case, recording didn’t start here because the room hasn’t started recording yet.
- When processing an SDP
    - In our case, recording didn’t start here because the room hasn’t started recording yet.
- When enable_recording is requested.
    - In our case, recording didn’t start here because the session hasn’t started yet (no WEBRTC media)

A moment later, WebRTC media is available and the session is marked as active. However, even though the room has the flag set that it should be recorded, and media is being published, no recording is being generated for the participant in question.

In this PR we simply add a check when the session starts, to also start the recording, if the room has recording enabled.

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -6866,6 +6866,19 @@ void janus_videoroom_setup_media(janus_plugin_session *handle) {
 			/* Notify all other participants that there's a new boy in town */
 			janus_videoroom_notify_about_publisher(participant, FALSE);
 			janus_refcount_decrease(&participant->ref);
+
+			/* Check if we need to start recording */
+			janus_mutex_lock(&participant->rec_mutex);
+			if(participant->room->record) {
+				GList *temp = participant->streams;
+				while(temp) {
+					janus_videoroom_publisher_stream *ps = (janus_videoroom_publisher_stream *)temp->data;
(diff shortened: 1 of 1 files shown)


src/plugins/janus_videoroom.c
```diff
@@ -6865,6 +6865,18 @@ void janus_videoroom_setup_media(janus_plugin_session *handle) {
 			janus_videoroom_publisher *participant = janus_videoroom_session_get_publisher(session);
 			/* Notify all other participants that there's a new boy in town */
 			janus_videoroom_notify_about_publisher(participant, FALSE);
+			/* Check if we need to start recording */
+			janus_mutex_lock(&participant->rec_mutex);
+			if(participant->room && participant->room->record) {
```
lminiero, line 6870: I guess there should be an OR with `participant->recording_active` here, since we have a similar check in the other places where we auomatically start recording (like the already mentioned post-answer check).

adnanel: ah, i seem to have misunderstood the meaning of `participant->recording_active`, thanks.

lminiero:
Thanks for the quick fix! Merging :+1:

PR #3018, opened 2022-07-05
Set appropriate channel type when make DATA_CHANNEL_REQUEST message

ktyu:
When Janus Server open a data channel first, I found out "unordered" parameter doesn't work. (always set true)

I think in ```janus_sctp_send_open_request_message()```, ```uint8_t unordered``` parameter is never used.
So I changed the channel_type when make DATA_CHANNEL_REQUEST message, according to "unordered" parameter.

```
// sctp.h

#define DATA_CHANNEL_RELIABLE							0x00
#define DATA_CHANNEL_RELIABLE_UNORDERED					0x80
#define DATA_CHANNEL_PARTIAL_RELIABLE_REXMIT			0x01
#define DATA_CHANNEL_PARTIAL_RELIABLE_REXMIT_UNORDERED	0x81
#define DATA_CHANNEL_PARTIAL_RELIABLE_TIMED				0x02
#define DATA_CHANNEL_PARTIAL_RELIABLE_TIMED_UNORDERED	0x82
```

### DIFF
--- a/src/sctp.c
+++ b/src/sctp.c
@@ -602,14 +602,14 @@ int janus_sctp_send_open_request_message(struct socket *sock, uint16_t stream, c
 	switch (pr_policy) {
 		case SCTP_PR_SCTP_NONE:
 			/* XXX: What about DATA_CHANNEL_RELIABLE_STREAM */
-			req->channel_type = DATA_CHANNEL_RELIABLE;
+			req->channel_type = unordered ? DATA_CHANNEL_RELIABLE_UNORDERED : DATA_CHANNEL_RELIABLE;
 			break;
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @ktyu! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks @ktyu! That does look like a useful addition, especially since we never tinkered much with different kind of datachannels. Have you verified that this way you can indeed get both ordered and unordered data channels to work with this patch?

lminiero:
Thanks, merging then! I'll backport the same changes to `0.x`, as the datachannel code there is the same.

Issue #1393, opened 2018-10-03
pp-rec segment fault with h264


3nk:
Hi,
First of all thanks for your great work on janus !

My application works fine with janus (recordplay) over almost all browsers (Firefox, Chrome, Edge) 
but on Safari since v12 when i try to process the generated MJR video file, the result is:

```
/opt/janus/bin/janus-pp-rec /opt/janus/share/janus/recordings/rec-1538557294902-video.mjr /opt/janus/share/janus/recordings/rec-1538557294902-video.mp4
Janus version: 45 (0.4.5)
Janus commit: 319a67308925e5258ed7032281ccf1d082a642e1
Compiled on:  Tue Oct  2 14:14:23 UTC 2018

/opt/janus/share/janus/recordings/rec-1538557294902-video.mjr --> /opt/janus/share/janus/recordings/rec-1538557294902-video.mp4
File is 706206 bytes
Pre-parsing file to generate ordered index...
[WARN] New .mjr header format
This is a video recording:
  -- Codec:   h264
  -- Created: 1538557294920720
  -- Written: 1538557296669300
SSRC detected: 4173138992
Counted 861 RTP packets
Counted 861 frame packets
[WARN] Profile is not baseline (100 != 66)
[WARN] Profile is not baseline (100 != 66)
  -- 32x16 (fps [90,17280] ~ 1000)
First keyframe: 0
First keyframe: 0
First keyframe: 0
First keyframe: 0
Segmentation fault

```

And sometimes:

```
/opt/janus/bin/janus-pp-rec /opt/janus/share/janus/recordings/rec-1538557294902-video.mjr /opt/janus/share/janus/recordings/rec-1538557294902-video.mp4
Janus version: 45 (0.4.5)
Janus commit: 319a67308925e5258ed7032281ccf1d082a642e1
Compiled on:  Tue Oct  2 14:14:23 UTC 2018

/opt/janus/share/janus/recordings/rec-1538557294902-video.mjr --> /opt/janus/share/janus/recordings/rec-1538557294902-video.mp4
File is 706206 bytes
Pre-parsing file to generate ordered index...
[WARN] New .mjr header format
This is a video recording:
  -- Codec:   h264
  -- Created: 1538557294920720
  -- Written: 1538557296669300
SSRC detected: 4173138992
Counted 861 RTP packets
Counted 861 frame packets
[WARN] Profile is not baseline (100 != 66)
[WARN] Profile is not baseline (100 != 66)
[...]

PR #3019, opened 2022-07-06
pp-h264: add skip helper for ignored bits

tmatth:
This PR adds a helper to simply skip unused exponential golomb codes rather than calculating them all

### DIFF
--- a/src/postprocessing/pp-h264.c
+++ b/src/postprocessing/pp-h264.c
@@ -134,6 +134,14 @@ static uint32_t janus_pp_h264_eg_getbit(uint8_t *base, uint32_t offset) {
 	return ((*(base + (offset >> 0x3))) >> (0x7 - (offset & 0x7))) & 0x1;
 }
 
+static void janus_pp_h264_eg_skip(uint8_t *base, uint32_t *offset) {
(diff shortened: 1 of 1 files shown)


tmatth:
Tested against: https://github.com/meetecho/janus-gateway/issues/1393#issuecomment-426598678

lminiero:
@tmatth is there any particular reason for this? Any considerable performance/speed gain from skipping rather than decoding? I'm not competent enough about golomb to know if these changes can cause issues parsing SPS/PPS in some cases.

tmatth:
> @tmatth is there any particular reason for this? Any considerable performance/speed gain from skipping rather than decoding? I'm not competent enough about golomb to know if these changes can cause issues parsing SPS/PPS in some cases.

I wish I could say I had a benchmark but it was more just trying to simplify/do less work in this function which is ultimately just extracting dimensions. Since the returned values are ignored it's possible the compiler is already optimizing away some of this (or could be made to), I will dig into this a bit further.

 I can mark as draft until I have more info for you.

lminiero:
Ack, thanks anyway! Should you find some more info we can come back to this.

PR #3026, opened 2022-07-14
janus.js: remove DtlsSrtpKeyAgreement constraint

fippo:
which is not even possible to configure otherwise in recent Chrome versions
and has been the default for the better part of a decade

(maybe remove googIPv6 too? It is going away in [M108](https://groups.google.com/g/discuss-webrtc/c/85e-f_siCws/m/4yWismzCAQAJ?)...)

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1800,9 +1800,7 @@ function Janus(gatewayCallbacks) {
 				// For Chrome versions before 72, we force a plan-b semantic, and unified-plan otherwise
 				pc_config["sdpSemantics"] = (Janus.webRTCAdapter.browserDetails.version < 72) ? "plan-b" : "unified-plan";
 			}
-			var pc_constraints = {
-				"optional": [{"DtlsSrtpKeyAgreement": true}]
-			};
+			var pc_constraints = {};
 			if(ipv6Support) {
 				pc_constraints.optional.push({"googIPv6":true});
 			}


html/janus.js
```diff
@@ -1800,9 +1800,7 @@ function Janus(gatewayCallbacks) {
 				// For Chrome versions before 72, we force a plan-b semantic, and unified-plan otherwise
 				pc_config["sdpSemantics"] = (Janus.webRTCAdapter.browserDetails.version < 72) ? "plan-b" : "unified-plan";
 			}
-			var pc_constraints = {
-				"optional": [{"DtlsSrtpKeyAgreement": true}]
-			};
+			var pc_constraints = {};
 			if(ipv6Support) {
 				pc_constraints.optional.push({"googIPv6":true});
```
henbos, line 1805: Yes, delete this? googIPv6:true has been a no-op because IPv6 has been enabled by default for several years. So the removal of googIPv6 in M108 would only affect googIPv6:false use cases

lminiero: Sure, no problem, I'm ok with getting rid of that too! I think it's been there from the very start when it ad a purpose, and we forgot it there...

html/janus.js
```diff
@@ -1800,9 +1800,7 @@ function Janus(gatewayCallbacks) {
...
 			}
```
henbos, line 1806: The code below, "Any custom constraints to add?", is also unnecessary now that there are no other constraints available than the two deprecated ones above: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/peerconnection/goog_media_constraints.idl

The rest of the constraints have already been removed and are necessarily NO-OPs on up-to-date versions of Chromium. The correct way to construct an RTCPeerConnection is to not provide a second argument (pc_constraints)

henbos: @lminiero FYI

henbos: Awesome! (You can delete the entire pc_constraints variable when you do this)

Issue #3024, opened 2022-07-13
Unable to identify multiple subscriptions to the same feed


JanFellner:
I am posting it here as i see this kind of a lack in the current subscription implementation.

Context:
- In a meeting scenario (videoroom) i have differnt styles of streams.
- In a side bar is see small streams (simulcast low substream)
- In the center area i see larger streams (simulcast m/h substream)

In certain scenarios both streams may show the same feed.

As subscription updates may come async i have no context if a change is related to a subscribe / unsubscribe operation.
So i need to be able to identify the subscribed streams in the updated event by other parameters.

The issue in this context is that i cannot differ the streams as they have no own unique identifier.

Solution to solve it would be to have an option to add an opaque_id to the subscribe request.
In see no need to be able to unsubscribe or configure based on that opaque_id.
The only purpose would be to have an own id to match the streams from the updated event with the effective subscription in the internal logic.

Further complexity would come if i use switch on the larger streams, then i have absolutly no clue which stream in the updated belongs to what.

Any thoughts about this? Don´t think that making a PR for that should be such a big issue.
Actually only the subscribe needs to read it, put it into the internal store and on every updated we add it as parameter to the streams list.

lminiero:
I'm not sure I understand the issue. For each subscribed stream you do have a unique ID, the `mid` in the subscription; if the problem is you want to know who that's from in a subscription update, you can couple that with the `feed_id` and `feed_mid` which uniquely identify the source. This way, even if you're subscribed twice to the same stream, those three identifiers together will tell you what's what.

I don't understand why it matters whether a change comes from subscribe or unsubscribe either, especially considering we now have a request called "update" where you can do both things at the same time. No matter the request, you always get an updated snapshot of all the subscribed streams in each event, meaning you can reconstruct your state out of that, and using the three IDs I mentioned as a way to figure out what's new and what went away.

An opaque_id wouldn't solve anything, since as we already discussed you have no guarantee that each subscribe/unsubscribe/update request will generate an event, since some of them may be grouped in a single one if we're still processing a previous update.

JanFellner:
The issue is the asynchronity.

I do a subscribe or unsubscribe but the response tells me updating.

Later i get an updated event and i have no property in there that allows me to distinguish the two subscriptions on the same feed (one small, the other large one).

Let´s say i subscribe the same feed two times, one simulcast low, the other simulcast large. 
I get an updating event so the event comes async.

[...]

PR #3027, opened 2022-07-14
An optional id in the subscribe request to match a subscription with newly created elements in the stream list

JanFellner:
Followup for this one: https://github.com/meetecho/janus-gateway/issues/3024
Comments welcome
(changed from private janus fork to estos fork to better follow the changes on our side)

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -1175,6 +1175,7 @@ room-<unique room ID>: {
 		{
 			"feed" : <unique ID of publisher owning the new stream to subscribe to>,
 			"mid" : "<unique mid of the publisher stream to subscribe to; optional>"
+			"crossrefid" : "<id to map this subscription with entries in streams list; optional>"
 			// Optionally, simulcast or SVC targets (defaults if missing)
 		},
 		// Other new streams to subscribe to
@@ -1278,6 +1279,7 @@ room-<unique room ID>: {
 		{
 			"feed" : <unique ID of publisher owning the new stream to subscribe to>,
 			"mid" : "<unique mid of the publisher stream to subscribe to; optional>"
+			"crossrefid" : "<id to map this subscription with entries in streams list; optional>"
 			// Optionally, simulcast or SVC targets (defaults if missing)
 		},
 		// Other new streams to subscribe to
@@ -1837,7 +1839,7 @@ static struct janus_json_parameter subscriber_parameters[] = {
 };
 static struct janus_json_parameter subscriber_stream_parameters[] = {
 	{"mid", JANUS_JSON_STRING, 0},
-	{"opaque_id", JANUS_JSON_STRING, 0},
+	{"crossrefid", JANUS_JSON_STRING, 0},
 	/* For VP8 (or H.264) simulcast */
 	{"substream", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"temporal", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
@@ -2206,7 +2208,7 @@ typedef struct janus_videoroom_subscriber_stream {
 	GSList *publisher_streams;						/* Complete list of publisher streams (e.g., when this is data) */
 	int mindex;				/* The media index of this stream (may not be the same as the publisher stream) */
 	char *mid;				/* The mid of this stream (may not be the same as the publisher stream) */
-	char* opaqueid;			/* An id provided while subscribing to uniquely identify the subscription in the updated events */
+	char *crossrefid;		/* An id provided while subscribing to uniquely identify the subscription in the list of subscriptions */
 	gboolean send;			/* Whether this stream media must be sent to this subscriber */
 	/* The following properties are copied from the source, in case this stream becomes inactive */
 	janus_videoroom_media type;			/* Type of this stream (audio, video or data) */
@@ -2273,7 +2275,7 @@ static void janus_videoroom_subscriber_stream_free(const janus_refcount *s_ref) 
 	/* This subscriber stream can be destroyed, free all the resources */
 		/* TODO Anything else we should free? */
 	g_free(s->mid);
-	g_free(s->opaqueid);
+	g_free(s->crossrefid);
 	g_free(s->h264_profile);
 	g_free(s->vp9_profile);
 	g_free(s);
@@ -2994,7 +2996,7 @@ static void janus_videoroom_create_dummy_publisher(janus_videoroom *room, GHashT
 /* Helpers for subscription streams */
 static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(janus_videoroom_subscriber *subscriber,
 		janus_videoroom_publisher_stream *ps,
-		const char* opaqueid,
+		const char *crossrefid,
 		gboolean legacy, gboolean do_audio, gboolean do_video, gboolean do_data) {
 	/* If this is a legacy subscription ("feed"), use the deprecated properties */
 	if(legacy && ((ps->type == JANUS_VIDEOROOM_MEDIA_AUDIO && !do_audio) ||
@@ -3007,7 +3009,7 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(
 	/* Allocate a new subscriber stream instance */
 	janus_videoroom_subscriber_stream *stream = g_malloc0(sizeof(janus_videoroom_subscriber_stream));
 	stream->mindex = g_list_length(subscriber->streams);
-	stream->opaqueid = g_strdup(opaqueid);
+	stream->crossrefid = g_strdup(crossrefid);
 	stream->subscriber = subscriber;
 	stream->publisher_streams = g_slist_append(stream->publisher_streams, ps);
 	/* Copy properties from the source */
@@ -3057,7 +3059,7 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(
 
 static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add_or_replace(janus_videoroom_subscriber *subscriber,
 		janus_videoroom_publisher_stream *ps,
-		const char* opaqueid) {
+		const char *crossrefid) {
 	if(subscriber == NULL || ps == NULL)
 		return NULL;
 	/* First of all, let's check if there's an m-line we can reuse */
@@ -3118,11 +3120,13 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add_
 		temp = temp->next;
 	}
 	if(found)  {
-		stream->opaqueid = g_strdup(opaqueid);
+		if(stream->crossrefid)
+			g_free(stream->crossrefid);
+		stream->crossrefid = g_strdup(crossrefid);
 		return stream;
 	}
 	/* We couldn't find any, add a new one */
-	return janus_videoroom_subscriber_stream_add(subscriber, ps, opaqueid, FALSE, FALSE, FALSE, FALSE);
+	return janus_videoroom_subscriber_stream_add(subscriber, ps, crossrefid, FALSE, FALSE, FALSE, FALSE);
 }
 
 static void janus_videoroom_subscriber_stream_remove(janus_videoroom_subscriber_stream *s,
@@ -3176,8 +3180,8 @@ static json_t *janus_videoroom_subscriber_streams_summary(janus_videoroom_subscr
 		json_object_set_new(m, "active", (ps || stream->type == JANUS_VIDEOROOM_MEDIA_DATA) ? json_true() : json_false());
 		json_object_set_new(m, "mindex", json_integer(stream->mindex));
 		json_object_set_new(m, "mid", json_string(stream->mid));
-		if(stream->opaqueid)
-			json_object_set_new(m, "opaque_id", json_string(stream->opaqueid));
+		if(stream->crossrefid)
+			json_object_set_new(m, "crossrefid", json_string(stream->crossrefid));
 		json_object_set_new(m, "ready", g_atomic_int_get(&stream->ready) ? json_true() : json_false());
 		json_object_set_new(m, "send", stream->send ? json_true() : json_false());
 		if(ps && stream->type == JANUS_VIDEOROOM_MEDIA_DATA) {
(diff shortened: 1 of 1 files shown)


src/plugins/janus_videoroom.c
```diff
@@ -2205,6 +2206,7 @@ typedef struct janus_videoroom_subscriber_stream {
 	GSList *publisher_streams;						/* Complete list of publisher streams (e.g., when this is data) */
 	int mindex;				/* The media index of this stream (may not be the same as the publisher stream) */
 	char *mid;				/* The mid of this stream (may not be the same as the publisher stream) */
+	char* opaqueid;			/* An id provided while subscribing to uniquely identify the subscription in the updated events */
```
lminiero, line 2209: Code style: * next to the pointer, not the type.

src/plugins/janus_videoroom.c
```diff
@@ -2991,6 +2994,7 @@ static void janus_videoroom_create_dummy_publisher(janus_videoroom *room, GHashT
 /* Helpers for subscription streams */
 static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(janus_videoroom_subscriber *subscriber,
 		janus_videoroom_publisher_stream *ps,
+		const char* opaqueid,
```
lminiero, line 2997: See above. Besides, per code style, this should be `opaque_id', not `opaqueid`.

src/plugins/janus_videoroom.c
```diff
@@ -3003,6 +3007,7 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(
 	/* Allocate a new subscriber stream instance */
 	janus_videoroom_subscriber_stream *stream = g_malloc0(sizeof(janus_videoroom_subscriber_stream));
 	stream->mindex = g_list_length(subscriber->streams);
+	stream->opaqueid = g_strdup(opaqueid);
```
lminiero, line 3010: This assumes `opaqueid` is always non-NULL, which is wrong. You should only duplicate the string if it exists.

src/plugins/janus_videoroom.c
```diff
@@ -3051,7 +3056,8 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(
 }
 
 static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add_or_replace(janus_videoroom_subscriber *subscriber,
-		janus_videoroom_publisher_stream *ps) {
+		janus_videoroom_publisher_stream *ps,
+		const char* opaqueid) {
```
lminiero, line 3060: See above,

src/plugins/janus_videoroom.c
```diff
@@ -3111,10 +3117,12 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add_
 		}
 		temp = temp->next;
 	}
-	if(found)
+	if(found)  {
+		stream->opaqueid = g_strdup(opaqueid);
```
lminiero, line 3121: See above. Besides, you're introducing leaks if `stream->opaqueid` exists already, as you're replacing the pointer and the old one will never be freed.

src/plugins/janus_videoroom.c
```diff
@@ -8458,6 +8468,7 @@ static void *janus_videoroom_handler(void *data) {
 					}
 					janus_mutex_lock(&publisher->streams_mutex);
 					const char *mid = json_string_value(json_object_get(s, "mid"));
+					const char *opaqueid = json_string_value(json_object_get(s, "opaque_id"));
```
lminiero, line 8471: New attributes that are added must be added to the related `_parameters` validation too.

src/plugins/janus_videoroom.c
```diff
@@ -9459,6 +9470,7 @@ static void *janus_videoroom_handler(void *data) {
 						}
 						/* Are we subscribing to this publisher as a whole or only to a single stream? */
 						const char *mid = json_string_value(json_object_get(s, "mid"));
+						const char *opaqueid = json_string_value(json_object_get(s, "opaque_id"));
```
lminiero, line 9473: Same here.

src/plugins/janus_videoroom.c
```diff
@@ -3168,6 +3176,8 @@ static json_t *janus_videoroom_subscriber_streams_summary(janus_videoroom_subscr
 		json_object_set_new(m, "active", (ps || stream->type == JANUS_VIDEOROOM_MEDIA_DATA) ? json_true() : json_false());
 		json_object_set_new(m, "mindex", json_integer(stream->mindex));
 		json_object_set_new(m, "mid", json_string(stream->mid));
+		if(stream->opaqueid)
+			json_object_set_new(m, "opaque_id", json_string(stream->opaqueid));
```
lminiero, line 3180: I'm not sure this should always be here. `janus_videoroom_subscriber_streams_summary` is used for different targets, not just notifying updates back to the subscriber, and this opaque identifier feels like something that only belongs to that subscriber's events.

lminiero (commented):
I had a quick glance. First of all, I'm not in love with the `opaque_id` terminology, mostly because it exists already in the core and will cause confusion when used in a plugin as well. Maybe it can just be something as simple as `description`, which is similar to what we have for publisher streams: since a `description` can be anything, it can be used for display purposes if you want, or for correlation purposes if you want. This is especially relevant considering you call it an ID, but looking at the code it can change any time you update a subscription, meaning it's probably not an ID but more something like metadata.

That said, I understand the point, but I'm still not 100% sure it would help you solve the problem, especially if this opaque property is changed across requests for the same m-line. Since as we discussed you may not receive updates in response to each subscribe/unsubscribe/update you send, but may receive less events that "group" the effect of multiple requests: if different opaque values were passed for the same m-line across different requests in sequence, you'll only get the last one in the event, so you may be waiting on an opaque you'll never receive.

I added a few additional notes on things that need to be fixed in the code too.

JanFellner:
Hi Lorenzo,

thanks for the review, before stepping into the code let´s shortly discuss the general topics above the code :).

> I had a quick glance. First of all, I'm not in love with the `opaque_id` terminology, mostly because it exists already in the core and will cause confusion when used in a plugin as well. Maybe it can just be something as simple as `description`, which is similar to what we have for publisher streams: since a `description` can be anything, it can be used for display purposes if you want, or for correlation purposes if you want. This is especially relevant considering you call it an ID, but looking at the code it can change any time you update a subscription, meaning it's probably not an ID but more something like metadata.

I followed the opaque_id as it did express exactly the wanted purpose to me (https://github.com/meetecho/janus-gateway/pull/748). I am fine with any name. Reusing the same is fine to me. I would otherwise call it cross ref id (as also used in csta context when monitoring something).
Just tell me if we should reuse description which is already existing (max also cause confusion as the description is then two times in the event, feed_description and description) or if we should have an own name for that purpose. The suggestion from my side would then be **crossrefid**.
If it is currently possible to change that attribute in an update request this is a non wanted behaviour. I see it really as correlation id between the subscription request and the first event that hands back the ids on the subscriber peerconnection but also in future events to be able to simplier align the janus event payload with own lookup maps.

> That said, I understand the point, but I'm still not 100% sure it would help you solve the problem, especially if this opaque property is changed across requests for the same m-line. Since as we discussed you may not receive updates in response to each subscribe/unsubscribe/update you send, but may receive less events that "group" the effect of multiple requests: if different opaque values were passed for the same m-line across different requests in sequence, you'll only get the last one in the event, so you may be waiting on an opaque you'll never receive.
- It should not be possible to change it. If thats currently the case i need to check where on an update this is overwritten.
- No receiving upates to all unique operations i call is fine to me, the logic does not care.
- The property 100% matches the need to be able to identify the subscribe operation with following events, no matter if with the next or any consequent following.

> I added a few additional notes on things that need to be fixed in the code too.
I would go over the code once we have the basic constraints fixed (naming etc.)

Just to have it mentioned:
g_strdup returns null if the argument is null. You want to have that checked with an if (argument) g_strup i can do that, technically it´s not needed but ensures that the variable it points to is properly nulled.

src/plugins/janus_videoroom.c
JanFellner: fixed

src/plugins/janus_videoroom.c
JanFellner: I see the place actually exactly here. This is the element in the streams list i cannot associated if i don´t have the cross reference id with the subscribe event. Thus this is the idea of the PR to have the correlation id in the streams list as provided in the subscribe request.

src/plugins/janus_videoroom.c
JanFellner: Am i missing something? It is added in the subscriber_stream_parameters object...

src/plugins/janus_videoroom.c
JanFellner: Same here.
Am i missing something? It is added in the subscriber_stream_parameters object...

JanFellner:
Changed to ready for review, running now "produtive" on our dev environment

src/plugins/janus_videoroom.c
```diff
@@ -2991,6 +2996,7 @@ static void janus_videoroom_create_dummy_publisher(janus_videoroom *room, GHashT
 /* Helpers for subscription streams */
 static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(janus_videoroom_subscriber *subscriber,
 		janus_videoroom_publisher_stream *ps,
+		const char *crossrefid,
```
lminiero, line 2999: No need for a new line, just move after ps.

lminiero (commented):
Thanks for the update. I'll be busy at the IETF meeting until the end of the month, so I don't think I'll have time to test this before I come back: I added just a couple of notes from a quick glance at the code. I'll let @atoppi do a deeper review.

src/plugins/janus_videoroom.c
```diff
@@ -3051,7 +3058,8 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add(
 }
 
 static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add_or_replace(janus_videoroom_subscriber *subscriber,
-		janus_videoroom_publisher_stream *ps) {
+		janus_videoroom_publisher_stream *ps,
+		const char *crossrefid) {
```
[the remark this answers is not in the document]
JanFellner: fixed

src/plugins/janus_videoroom.c
```diff
@@ -3111,10 +3119,14 @@ static janus_videoroom_subscriber_stream *janus_videoroom_subscriber_stream_add_
 		}
 		temp = temp->next;
 	}
-	if(found)
+	if(found)  {
+		if(stream->crossrefid)
+			g_free(stream->crossrefid);
```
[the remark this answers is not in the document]
JanFellner: fixed

src/plugins/janus_videoroom.c
JanFellner: fixed

atoppi:
Changes look good to me.
Tested with a modified version of `mvideoroomtest.js` and it seems to work as expected with the `update` API.

@JanFellner According to your use cases, I'm wondering if it would make any sense to add this `crossrefid` to the `streams` list when joining a new subscriber.

```
{
  "janus": "message",
  "body": {
    "request": "join",
    "room": 1234,
    "ptype": "subscriber",
    "streams": [
      {
        "feed": 4446040227822893,
        "mid": "0"
      },
      {
        "feed": 4446040227822893,
        "mid": "1"
      }
    ],
    "private_id": 2847518190
  },
  "transaction": "AvkXpeLdGoLZ",
  "session_id": 431131763148161,
  "handle_id": 7854428491773330
}
```

JanFellner:
@atoppi maybe i am missing the core question but the crossrefid is added from the client calling janus in exactly that request and returend in any stream related event from janus.

This is how such a request and response looks like.

<pre>
{
  "_": "2022-07-25T11:00:17.345Z - ws - out",
  "janus": "message",
  "transaction": "S_message_7_6",
  "handle_id": 6602915452921274,
  "session_id": 7047460590128940,
  "timeout": 500,
  "body": {
    "request": "join",
    "ptype": "subscriber",
    "room": "R_55698ee42ed94463ae6b0ae9ace87fa7",
    "streams": [
      {
        "feed": "56174d8c-d6e2-48ae-ad6b-7da87e973362",
        "mid": "0",
        "crossrefid": "S_82575842511f4d39ac2b88b5722b0d79"
      },
      {
        "feed": "56174d8c-d6e2-48ae-ad6b-7da87e973362",
        "mid": "1",
        "crossrefid": "S_82575842511f4d39ac2b88b5722b0d79",
        "substream": 2,
        "temporal": 1
      }
    ]
  },
  "opaque_id": "P_bea32134fe0c44c3bee95cfced8c133c|publisher"
}
</pre>

<pre>
{
  "_": "2022-07-25T11:00:17.388Z - ws - in",
  "janus": "event",
  "session_id": 7047460590128940,
  "transaction": "S_message_7_6",
  "sender": 6602915452921274,
  "plugindata": {
    "plugin": "janus.plugin.videoroom",
    "data": {
      "videoroom": "attached",
      "room": "R_55698ee42ed94463ae6b0ae9ace87fa7",
      "streams": [
        {
          "type": "audio",
          "active": true,
          "mindex": 0,
          "mid": "0",
          "crossrefid": "S_82575842511f4d39ac2b88b5722b0d79",
          "ready": false,
          "send": true,
          "feed_id": "56174d8c-d6e2-48ae-ad6b-7da87e973362",
          "feed_display": "P_8f248c81a2174b10a2153eed2a483152",
          "feed_mid": "0",
          "feed_description": "Video 2 - left human main",
          "codec": "opus"
        },
        {
          "type": "video",
          "active": true,
          "mindex": 1,
          "mid": "1",
          "crossrefid": "S_82575842511f4d39ac2b88b5722b0d79",
          "ready": false,
          "send": true,
          "feed_id": "56174d8c-d6e2-48ae-ad6b-7da87e973362",
          "feed_display": "P_8f248c81a2174b10a2153eed2a483152",
          "feed_mid": "1",
          "feed_description": "Video 2 - left human main",
          "codec": "vp8",
          "simulcast": {
            "substream": -1,
            "substream-target": 2,
            "temporal-layer": -1,
            "temporal-layer-target": 1
          }
        }
      ]
    }
  },
  "jsep": {
    "type": "offer",
    "sdp": "..."
  }
}
</pre>

lminiero:
Yes, it's a per-subscriber property, not per-publisher from what I understood.

JanFellner:
In my understanding / terminology a per subscription id. 
Cause i als a subscriber join one time but hold 1:n subscriptions on different feeds in the room.
To differ these feeds, especially in case where i hold multiple subscriptions on the same feed, i need this id.

atoppi:
> @atoppi maybe i am missing the core question but the crossrefid is added from the client calling janus in exactly that request and returend in any stream related event from janus.

You're right, I forgot to customize the JS client in one point, sorry for the noise.
Anyway the documentation [is still missing ](https://github.com/ESTOS/janus-gateway/blob/feature/3024-add-opaque-id-to-subscriptions/src/plugins/janus_videoroom.c#L1051)the `crossrefid` parameter in the `join` as a subscriber section.

JanFellner:
Fixed :)
https://github.com/ESTOS/janus-gateway/blob/feature/3024-add-opaque-id-to-subscriptions/src/plugins/janus_videoroom.c#L1055

JanFellner:
So *whom* clicks merge now 😆

lminiero:
Ack, thanks for the patch and for your patience, merging now :+1:

PR #3035, opened 2022-08-09
Unix socket TURN REST API

mgetka:
This PR allows to specify unix domain socket of TURN REST API service.

Current TURN REST API client implementation gives little flexibility. This may be overcome by implementing a middleware server standing between the janus service and the remote server actually generating the TURN server credentials. It may be especially useful if the remote server requires more sophisticated request authentication than just static API keys.

In such scenario, in most cases, it will be expected to restrict middleware service access only for janus. This can, for eg. be achieved by binding on localhost and establishing API key allowing for interactions between janus and the service. Other option is to bind on unix domain socket and depend on fs ownership and permissions model for authentication & authorization.

In this PR we are making second option feasible.


januscla:
Thanks for your contribution, @mgetka! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

src/turnrest.c
```diff
@@ -94,6 +98,12 @@ void janus_turnrest_set_backend(const char *server, const char *key, const char
 				api_http_get = FALSE;
 			}
 		}
+		if(unixpath != NULL) {
+			if(access(unixpath, R_OK | W_OK) != 0) {
+				JANUS_LOG(LOG_WARN, "TURN REST API unix socket '%s' was not accessible by the service during startup\n", unixpath);
```
mgetka, line 103: I account for the situation that the socket is not yet ready at janus startup, eg. because related service is still initializing. The socket must be accessible only at the time of the request, and then, appropriate error will be generated by curl.

lminiero:
I'm on holiday so I won't be able to review until I'm back, which should happen in the last week of August.

src/turnrest.c
mgetka: While getting acquainted with janues in more detail, I came across a comment on `ignore_unreachable_ice_server` configuration variable.

> In case you want to allow Janus to start even if the configured STUN or TURN
> server is unreachable, you can set 'ignore_unreachable_ice_server' to true.
> WARNING: We do not recommend to ignore reachability problems, particularly
> if you run Janus in the cloud. Before enabling this flag, make sure your
> system is correctly configured and Janus starts after the network layer of
> your machine is ready. Note that Linux distributions offer such directives.
> You could use the following directive in systemd: 'After=network-online.target'

This gives me insight, that maybe one of the janus project assumption is that, by default, the process will fail if it recognizes that its runtime environment is not yet ready or malfunctioning. If it is so, I will rewrite this aspect.

Similarly to cited above note on ignore_unreachable_ice_server, adding a note on 
`Requires`/`Wants`/`After` systemd unit option may be considered.

lminiero:
Apologies for the late response.

To be honest, I don't see this as something we can merge: while the existing TURN REST API can be considered a "standard" (there's an old draft for it, and to my knowledge it's adopted by some TURN servers too), using it on other protocols, including Unix Sockets, isn't. As such, this feels like a bespoke modification you're making because you actually want to use it that way, but probably not something that can be used by anyone else. I'd rather only add things that benefit most people, and possibly using a standard solution.

lminiero:
I wouldn't say it's not valuable (all contributions are, and are always welcome!) but I do think it's too niche and would add another surface to maintain for little additional benefit. As such, for now I'll close it.

Thanks anyway for your contribution!

PR #3036, opened 2022-08-10
Update CHANGELOG

rprata:
We need to adjusting `CHANGELOG` according legacy and new version of Janus Gateway. It helps developer to understand the difference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 
-## [v1.0.4] - 2022-08-01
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @rprata! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
The changelog for the different versions is not the same, so your suggestion cannot be accepted, sorry. If you need the changelog for legacy, you need to check the `CHANGELOG.md` file in the `0.x` branch.

PR #3047, opened 2022-08-29
Set CURLOPT_ERRORBUFFER in Streaming plugin for more verbose RTSP debugging (fixes #3043)

### LINKED ISSUE #3043: [1.x] CURLOPT_ERRORBUFFER not being used in libcurl requests
CURLOPT_ERRORBUFFER isn't being allocated or set, so when CURL errors happen, we miss out on some possible further information. This could be added to both 1.x and 0.x. See https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html or more info.

For context, we're getting 
`[ERR] [plugins/janus_streaming.c:janus_streaming_rtsp_connect_to_server:5324] Couldn't send DESCRIBE request: Unsupported protocol`
when trying to connect to a camera, and have yet to figure out the root cause. The error buffer might have a bit more info.

lminiero:
As tht eitle says, this PR sets [CURLOPT_ERRORBUFFER](https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html) when setting up libcurl for an RTSP session, so that we can print more info in case something goes wrong. Example from a quick test I did locally trying to create an RTSP mountpoint for an unreachable camera:

    [ERR] [plugins/janus_streaming.c:janus_streaming_rtsp_connect_to_server:7476] Couldn't send DESCRIBE request: Couldn't connect to server (Failed to connect to 192.168.1.6 port 88 after 3059 ms: No route to host)

### DIFF
--- a/src/plugins/janus_streaming.c
+++ b/src/plugins/janus_streaming.c
@@ -1303,6 +1303,7 @@ typedef struct janus_streaming_rtp_source {
 #ifdef HAVE_LIBCURL
 	gboolean rtsp;
 	CURL *curl;
+	char *curl_errbuf;
 	janus_streaming_buffer *curldata;
 	char *rtsp_url;
 	char *rtsp_username, *rtsp_password;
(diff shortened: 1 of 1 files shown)


lminiero:
Merging.

PR #3051, opened 2022-09-02
Fixed not muting if there is more than one audio stream

rjnpnfigueiredo:
Whenever there is more than one audio track due to changing audio devices during a call, such as switching to a bluetooth device, muting would no longer work because it always assumed that the first track is the only one.

Switch to cycle through all tracks and set the mute status on all.

Version 0.x also affected

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -2874,7 +2874,7 @@ function Janus(gatewayCallbacks) {
 				}
 				transceiver.sender.track.enabled = mute ? false : true;
 			} else {
-				for(const videostream of config.myStream.getVideoTracks()){
+				for(const videostream of config.myStream.getVideoTracks()) {
 					videostream.enabled = !mute
 				}
 			}
@@ -2897,7 +2897,7 @@ function Janus(gatewayCallbacks) {
 				}
 				transceiver.sender.track.enabled = mute ? false : true;
 			} else {
-				for(const audiostream of config.myStream.getAudioTracks()){
+				for(const audiostream of config.myStream.getAudioTracks()) {
 					audiostream.enabled = !mute
 				}
 			}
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @rjnpnfigueiredo! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

html/janus.js
```diff
@@ -2874,7 +2874,9 @@ function Janus(gatewayCallbacks) {
 				}
 				transceiver.sender.track.enabled = mute ? false : true;
 			} else {
-				config.myStream.getVideoTracks()[0].enabled = mute ? false : true;
+				for(const videostream of config.myStream.getVideoTracks()){
```
lminiero, line 2877: As a code style nit, please add a space before the curly bracket.

lminiero (commented):
Thanks for spotting this, it's probably a leftover from `0.x` that we forgot to update. I don't think it's an issue in `0.x` instead, since that one is limited to 1 audio + 1 video: if you have two audio m-lines, the first one is the only one that's taken into account anyway, and all other ones are ignored and discarded.

html/janus.js
```diff
@@ -2895,7 +2897,9 @@ function Janus(gatewayCallbacks) {
 				}
 				transceiver.sender.track.enabled = mute ? false : true;
 			} else {
-				config.myStream.getAudioTracks()[0].enabled = mute ? false : true;
+				for(const audiostream of config.myStream.getAudioTracks()){
```
lminiero, line 2900: See above.

lminiero:
@rjnpnfigueiredo can you address the comments so that we can merge? Thanks!

rjnpnfigueiredo:
@lminiero Sorry, been on holiday, I think the latest commit should improve code style to match.

lminiero:
@rjnpnfigueiredo before we can merge, can you sign the CLA? The link is in the first comment. Thanks!

rjnpnfigueiredo:
@lminiero Forgot about that, done!

lminiero:
Thanks, merging!

PR #3056, opened 2022-09-06
Don't call .stop() in onremotetrack when track is notified as removed by janus.js (fixes #3055)

### LINKED ISSUE #3055: Videroom multistream bug after unpublish/publlish
const char *janus_build_git_sha = "bc8cc9c3f64f07fb73da21d034b6748a6b8165c4";
const char *janus_build_git_time = "Mon Feb 14 12:34:55 UTC 2022";
int janus_version = 1000;
const char *janus_version_string = "1.0.0";
const char *libnice_version_string = "0.1.18.1";

v1.0.4

Yes

**Was this working before?**
No idea

No

**Additional context**
Videroom multiroom.

Publisher A (video and audio)
Publisher B (video and audio)

Step 1: A and B join the room -> Both streams are visibile
Step 2: B Leaves the room/unpublish -> A can no longer see B (as expected)
Step 3: B joins the room again and can see its own feed and A. A can not see B (empty blank or black stream). onremotetrack gets fired for the new tracks of B on A. Stream gets added, but stays blank. Both these tracks have the same Id as previous tracks. Both tracks have readyState: "ended". This is problematic.

Even if B closes its browser between step 2 and 3 it fails. Even after cleaning all the cache. The only way to get it working again is to let A unpublish as well, before both join again.

Both my own implementation and the online videoroom multistream example have this issue.

Janus version 1.0.4

**When it works**
The odd thing is that it works in the following situation:

1. B unpublish
2. A receives multiple onremotetrack(track,mid,false)
3. Normally the code would stop the specific tracks associated with the specific stream. If we disable/remove `mst.stop()` in `onremotetrack `in `subscribeTo` everthing works. Removing the MediaStream from the remoteTrack array is not an issue.
4. B rejoins

Issue occures in Edge,Chrome,Firefox, Samsung Internet (Android)

I noticed that Janus is reusing track-ids. Especially after having closed browser as publisher, and having rejoined the room. One would expect fresh track-ids.

**Possible solution**
1. Do not reuse trackIds in the same videoroom. Gives issues with the readyState of the track.
2. remove mst.stop in onremotetrack (not ideal).

lminiero:
This is a fix for the issue reported in #3055, where the symptom was the multistream VideoRoom demo breaking when someone left a room and then joined again (or someone else joined): for existing subscribers, m-lines would be reused, and media would be broken. This was caused by us closing the remote stream in the `onremotetrack` callback for a removed track: since the new track (new user) actually remains the same after the renegotiation, this ends up breaking the rendering.

Initially we thought it might be caused by us re-using track IDs in SDPs from Janus, but even after making sure we generate fresh ones when needed, the problem remained the same, and we could see the `ontrack` callback of the PeerConnection would still report the previous track ID as being added: not sure why, but maybe because as far as SDP is concerned the fact we changed the ID of the track didn't change anything, since the m-line still "lives", and so the browser assumed the previous one was still in use and ignored the new one.

As such, we decided to go for a different fix, and just removed all calls to `.stop()` for the remote track when one is removed. We gave this some thought and realized this could have caused issues also when a track was only temporarily paused, e.g., freezed because of losses: in fact, the `onremotetrack(false)` event is fired in response to an `onmuted` event on the track, not `onended`, which is by definition not conclusive. Besides, the `.stop()` calls we had were probably a leftover when we still cloned tracks before using them in demos, which is something we stopped doing in #3009.

Not sure if this can cause leaks in demos, now, but it definitely fixes the problem. Feedback would be welcome before we merge, of course. Also pinging @fippo who may point out gross mistakes we're doing :hand_over_mouth:

### DIFF
--- a/html/audiobridgetest.js
+++ b/html/audiobridgetest.js
@@ -304,16 +304,6 @@ $(document).ready(function() {
 										return;
 									if(!on) {
 										// Track removed, get rid of the stream and the rendering
-										if(remoteStream) {
-											try {
-												var tracks = remoteStream.getTracks();
-												for(var i in tracks) {
-													var mst = tracks[i];
-													if(mst)
-														mst.stop();
-												}
-											} catch(e) {}
-										}
 										remoteStream = null;
 										$('#roomaudio').remove();
 										return;
--- a/html/canvas.js
+++ b/html/canvas.js
@@ -173,17 +173,6 @@ $(document).ready(function() {
 									Janus.debug("Remote track (mid=" + mid + ") " + (on ? "added" : "removed") + ":", track);
 									if(!on) {
 										// Track removed, get rid of the stream and the rendering
-										var stream = remoteTracks[mid];
-										if(stream) {
-											try {
-												var tracks = stream.getTracks();
-												for(var i in tracks) {
-													var mst = tracks[i];
-													if(mst)
-														mst.stop();
-												}
-											} catch(e) {}
-										}
 										$('#peervideo' + mid).remove();
 										if(track.kind === "video") {
 											remoteVideos--;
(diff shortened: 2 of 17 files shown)


lminiero:
Merging.

fippo:
You are doing less gross mistakes now which is always the way to go ;-)

PR #3057, opened 2022-09-07
Fix libnice build command

kenju:
as described in the README.md of libnice,
the build command using meson/ninja has been changed

https://gitlab.freedesktop.org/libnice/libnice

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,9 @@ On Ubuntu or Debian, it would require something like this:
 While `libnice` is typically available in most distros as a package, the version available out of the box in Ubuntu is known to cause problems. As such, we always recommend manually compiling and installing the master version of libnice.
 To build libnice, you need Python 3, Meson and Ninja:
 
-	git clone https://gitlab.freedesktop.org/libnice/libnice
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @kenju! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

atoppi:
Hi @kenju ,
I have tested the current instructions with the latest libnice `master` and it works as expected, nothing changed in the build process as far I can tell.

I'm afraid we can not accept this change since your patch:

- renames the libnice build directory from `build` to `buildir` (pointless change)
- removes the recommended libnice version (`0.1.19`), switching to `master`
- removes the destination prefix for the built library (`/usr`)

lminiero:
> removes the recommended libnice version (`0.1.19`), switching to `master`

Actually this is incorrect, this patch added checking out 0.1.19 (which is indeed suggested). But on the rest I agree, I don't see any relevant change, since we were already using meson and ninja in the README.

atoppi:
> removes the recommended libnice version (0.1.19), switching to master

Sorry for the mistake, of course you're doing the opposite! Still the whole patch is not acceptable because the build process is working as it is.

PR #3059, opened 2022-09-08
Convert strlen calls to strnlen

tmatth:
- For `strlen(x) > c` comparisons, use `strnlen(x, c + 1) > c` (so O(1) instead of O(n))
- For `strlen(x) < c` comparisons, use `strnlen(x, c) < c`, so (so O(1) instead of O(n))

I didn't change `strlen(x) > 0` comparisons (or any other comparison to 0) since compilers already optimize away the `strlen` call by transforming those to e.g., `x[0] != '\0'`, see https://godbolt.org/z/KGW359ca1

N.B. Some of these calls are still followed by `strlen` calls, but this is always in the error and/or logging case (not the most likely path).

Inspired by https://code.videolan.org/videolan/vlc/-/merge_requests/2477/diffs

### DIFF
--- a/src/events/janus_wsevh.c
+++ b/src/events/janus_wsevh.c
@@ -376,7 +376,7 @@ int janus_wsevh_init(const char *config_path) {
 		goto error;
 	}
 	path[0] = '/';
-	if(strlen(p) > 1)
+	if(strnlen(p, 1 + 1) > 1)
 		g_strlcpy(path + 1, p, sizeof(path)-2);
 	/* Before connecting, let's check if the server expects a subprotocol */
 	item = janus_config_get(config, config_general, janus_config_type_item, "subprotocol");
--- a/src/plugins/janus_streaming.c
+++ b/src/plugins/janus_streaming.c
@@ -7663,7 +7663,7 @@ static int janus_streaming_rtsp_connect_to_server(janus_streaming_mountpoint *mp
 					index++;
 					continue;
 				}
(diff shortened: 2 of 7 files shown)


atoppi:
Always cool stuff from @tmatth :sunglasses:
Kudos for the online service compiling the code in real time with different compilers and flags.
That's brilliant.

I can see the enhancement here, but being pedantic

> For strlen(x) > c comparisons, use strnlen(x, c + 1) > c (so O(1) instead of O(n))

AFAIU that will not be O(1), but O(c+1).
It would theoretically be O(1) if C had a string representation like e.g. Golang, with a "length" attribute.

atoppi:
mh looks like [strnlen has been excluded from C23 standard](https://stackoverflow.com/questions/72392985/why-is-strnlen-not-considered-for-inclusion-in-c23), I'm wondering if that might be a problem in future

tmatth:
> mh looks like [strnlen has been excluded from C23 standard](https://stackoverflow.com/questions/72392985/why-is-strnlen-not-considered-for-inclusion-in-c23), I'm wondering if that might be a problem in future

That's an interesting find. I think it's ok since it's already not part of the standard, but won't be dropped from POSIX. If it does at some point become a problem it'd be trivial to add a fallback using `memchr` (but I wouldn't use `memchr` directly since it's a bit less readable).

lminiero:
Thanks for the discussion and the contribution, merging then!

Issue #3031, opened 2022-07-26
[0.x] SIP events mixed in event handling


oriol-c:
**What version of Janus is this happening on?**
0.11.7

**Have you tested a more recent version of Janus too?**
No, code in relation to janus_sip_sofia_logger has not been changed.

**Additional context**
When sending about 20cps via SIP, some of the sip-out events sent using the events handlers contain mixed SIP messages (https://pastebin.com/RYx4aW7e) . As you can see, the body starts with an ACK message but is then mixed with the next INVITE message, it actually goes one line (sofia log line) of the ACK and one line of the INVITE.
I could see that the sip plugin redirects the sofia log to this function janus_sip_sofia_logger, but I'm not sure if something can be done in janus to prevent the logs coming from sofia from being mixed.

lminiero:
No, to my knowledge that's the only way we can intercept outgoing and incoming messages using the NUA stack. I don't remember mixed messages ever being an issue, if not initially before we added locking to the logger function. Not sure if anything changed in Sofia SIP that can cause the problem now, but to be honest I wouldn't know how to fix that, since we're working with raw buffers to reconstruct messages.

lminiero:
Thinking about it, the logger is shared among all sessions, but we have global properties (e.g., `sofia_log` and `call_id`, but `skip`, `started`, `append` too) that we sometimes access without a lock, and the function may be called by different threads.

Since you're using 0.x, can you check if the diff below helps? It uses `sessions_mutex` as a way to protect the whole function, since we were using it before already to look up call-ids. Notice I haven't tested this, so I don't know if it has other issues, but I'm mostly interested in checking if it helps with the mixup of lines.

<details>
<summary>diff</summary>

```
diff --git a/plugins/janus_sip.c b/plugins/janus_sip.c
index bbc31d14..bafa8191 100644
--- a/plugins/janus_sip.c
+++ b/plugins/janus_sip.c
@@ -1672,12 +1672,14 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
 #pragma GCC diagnostic pop
 #endif
 #pragma GCC diagnostic pop
+	janus_mutex_lock(&sessions_mutex);
 	if(skip) {
 		/* This is a message we're not interested in: just check when it ends */
 		if(line[3] == '-') {
 			skip = FALSE;
 			append = FALSE;
 		}
+		janus_mutex_unlock(&sessions_mutex);
 		return;
 	}
 	if(append) {
[...]

PR #3060, opened 2022-09-08
increased sip buffer size

kenangenjac:
In the last few days we have been getting a lot of spam logs from Janus on our WebRTC platform.
The log messages looks like this: **Truncation occurred, 2132 >= 2048**, and are usually caused by destination buffer being close to the size of the `sofia_log` array max size , which is set to 2048 inside `janus_sip.c` file.

The truncation occurs by invoking `janus_strlcat` function on the 1744th and 1757th line inside `janus_sip_sofia_logger` function, located in `janus_sip.c`. When the `janus_strlcat` is invoked, destination buffer, source buffer and destination size in bytes are concatenated and truncation occurs.

In this PR I made a change to the `sofia_log` array, making it 3kB, instead of the original 2kB, as I saw that the return value of the `janus_strlcat`, which causes the overflow and truncation, is never more than 2.3kB.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1655,7 +1655,7 @@ static void janus_sip_parse_custom_contact_params(json_t *root, char *custom_par
 /* Sofia SIP logger function: when the Event Handlers mechanism is enabled,
  * we use this to intercept SIP messages sent by the stack (received
  * messages are more easily recoverable in janus_sip_sofia_callback) */
-char sofia_log[2048];
+char sofia_log[3072];
 char call_id[255];
 gboolean skip = FALSE, started = FALSE, append = FALSE;
 static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {


januscla:
Thanks for your contribution, @kenangenjac! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
That variable is only needed when using event handlers with the SIP plugin: is this what you're doing? If so, have you checked if you're also impacted by #3031, and if the diff attached in one of the comment helps?

kenangenjac:
We aren't using the event handlers, but we fear we might miss something important due to this message spamming our logs, which is why we proposed this buffer increase.

lminiero:
You're not missing anything, if you're not using event handlers. That said, the patch may still be useful for those that do.

kenangenjac:
Yes, I think it would possibly be useful for others as it could be a general issue. With that in mind, if the PR is not going to be accepted and is not for merging, could You please close it.

lminiero:
No need, I'll merge, and backport to `0.x`. Thanks!

kenangenjac:
Great! Thank You, glad I contributed!

PR #3063, opened 2022-09-12
Use different callback functions to get SIP log messages for Sofia-SIP 1.12 and Sofia-SIP 1.13. 

oriol-c:
Use different callback functions to get SIP log messages for Sofia-SIP 1.12 and Sofia-SIP 1.13. 

In Sofia-SIP 1.12 the callback was called for each line in the SIP message. In Sofia-SIP 1.13 this behavior was changed to send the entire SIP message in a single callback.

This could also be implemented using macros in the Makefile and setting the Sofia SIP version in the compile runtime, but that would remove the flexibility of the compiled Janus to run simultaneously against both Sofia SIP versions.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -867,6 +867,7 @@ static char *sips_certs_dir = NULL;
 static int sip_timer_t1x64 = JANUS_DEFAULT_SIP_TIMER_T1X64;
 
 static gboolean query_contact_header = FALSE;
+static gboolean message_log_single_callback = FALSE;
 
 static GThread *handler_thread;
 static void *janus_sip_handler(void *data);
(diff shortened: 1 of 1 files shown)


src/plugins/janus_sip.c
```diff
@@ -1771,6 +1772,127 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
 
+/* Sofia SIP logger function: when the Event Handlers mechanism is enabled,
+ * we use this to intercept SIP messages sent by the stack (received
+ * messages are more easily recoverable in janus_sip_sofia_callback) */
+static void janus_sip_sofia_logger_single(void *stream, char const *fmt, va_list ap) {
+	if(!fmt)
+		return;
+	char line[255];
+	char message[16000];
+	char sofia_log_buf[16000];
```
lminiero, line 1783: These buffers are way too large.

src/plugins/janus_sip.c
```diff
@@ -1771,6 +1772,127 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+			cr = strchr(line, '\r');
+			if(cr != NULL)
+				*cr = '\0';
+			if(*line == '\0') {
+				if(cr != NULL)
+					*cr = '\r';
+				index++;
+				continue;
+			}
+
```
lminiero, line 1812: Empty line is unneeded here.

src/plugins/janus_sip.c
```diff
@@ -1771,6 +1772,127 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+			/* First line of the log indicates if we send or receive a message. */
+			/* We are only interested in the sent ones */
```
lminiero, line 1814: "send or receive" should be "sent or received I guess?
Just a nit per code style: use a single comment block, not two, e.g.

	/* First line of the log indicates if we sent or received a message:
	 * we are only interested in the sent ones */

src/plugins/janus_sip.c
```diff
@@ -1771,6 +1772,127 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+				if(strstr(line, "send ") == line) {
+					int length = atoi(line+5);
+					JANUS_LOG(LOG_HUGE, "Intercepting message (%d bytes)\n", length);
+					index++;
+					continue;
+				}
+				g_strfreev(parts);
+				return ;
+			}
+
```
lminiero, line 1825: Unneeded empty line.

src/plugins/janus_sip.c
```diff
@@ -1771,6 +1772,127 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+			if(strlen(line) == 1) {
```
lminiero, line 1826: As explained in #3059, this should use `strnlen` instead of `strlen`, to optimize performance. Since you're interested in checking if the size is exactly 1, I guess the right approach is passing at least 2, e.g.:

    if(strnlen(line, 1 + 1) == 1) {

(but @tmatth may correct me if I'm wrong).

src/plugins/janus_sip.c
```diff
@@ -1771,6 +1772,127 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+				janus_strlcat(sofia_log_buf, "\r\n", sizeof(sofia_log_buf));
+			} else if(strstr(line, "-") == line ) {
+				index++;
+				continue;
+			} else if(strstr(line, "OPTIONS") == line) {
+				index++;
+				continue;
+			} else {
+				/* Is this a Call-ID header? Keep note of it */
+				if(strstr(line, "Call-ID") == line) {
```
lminiero, line 1837: Per code style, since the action is simple there's no need for curly brackets.

src/plugins/janus_sip.c
```diff
@@ -1771,6 +1772,127 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+					g_snprintf(call_id_hdr, sizeof(call_id_hdr), "%s", line+9);
```
lminiero, line 1838: Why is this 9? Are you assuming `Call-ID: `? The `strstr` call above is looking for `Call-ID`, not `Call-ID: `, which can cause problems if the header is formatted differently, or if `Call-ID` is not followed by a semicolon.

src/plugins/janus_sip.c
```diff
@@ -1771,6 +1772,127 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+		}
+		g_strfreev(parts);
+	} else {
+		return;
+	}
+
+	/* Look for the session this message belongs to */
+	janus_sip_session *session = NULL;
+	janus_mutex_lock(&sessions_mutex);
+	if(strlen(call_id_hdr))
```
lminiero, line 1856: This should be changed to

    if(strlen(call_id_hdr) > 0)

otherwise I'm not sure if the compiler optimizations discussed in #3059 will kick in.

lminiero (commented):
Thanks! I've added a few notes inline, mostly editorial. I'm wondering if the two functions can be merged in one that behaves differently depending on the version, but that's something that can be done later on if needed. Is this approach working fine for you, without the mixed lines you were experiencing at higher CPS?

src/plugins/janus_sip.c
tmatth: From a cursory check of a few compilers (e.g., https://godbolt.org/z/enodKdhMc) `if(strlen(call_id_hdr))` and `if(strlen(call_id_hdr) > 0)` will compile down to the same check of `call_id_hdr[0] != 0`

So you can use whichever you find more readable.

lminiero: Thanks for checking!

src/plugins/janus_sip.c
oriol-c: I set those values to match the maximum buffer that Sofia SIP defines in https://github.com/freeswitch/sofia-sip/blob/master/libsofia-sip-ua/tport/tport_logging.c#L843

src/plugins/janus_sip.c
oriol-c: I will remove it

src/plugins/janus_sip.c
oriol-c: Will fix both.

src/plugins/janus_sip.c
oriol-c: I will remove it

src/plugins/janus_sip.c
oriol-c: This part already exists in the janus_sip_sofia_logger callback function (the one for line-by-line logs). I can fix it in both functions if you want.

src/plugins/janus_sip.c
oriol-c: This also existed in janus_sip_sofia_logger. I suppose it's safe to assume that "Call-ID: " will always be sent with this format,  as it is Sofia-SIP library the one that generates those messages and we only process the messages sent from Sofia SIP stack.

src/plugins/janus_sip.c
lminiero: I think it's still too large of a buffer, especially to create any time the function is called. The SIP SDPs in Janus are never very large, so I'd put a cap on it.

src/plugins/janus_sip.c
lminiero: Ouch, I ended up reviewing my own ugly code :hand_over_mouth:

src/plugins/janus_sip.c
lminiero: Makes sense!

lminiero:
Adding the multistream label as the patch is for master. I'll take care of backporting to `0.x` once this is merged.

lminiero:
@oriol-c any update on the fixes? As soon as they happen, I'll make some tests and in case merge. Thanks!

lminiero:
@oriol-c please let us know if you've abandoned plans to update this PR. In case, we'll just close it. Thanks.

PR #3068, opened 2022-09-15
Cache videoroom publisher keyframes if enabled

chriswiggins:
This is a draft PR (for feedback) on shamelessly copying the keyframe caching code from the streaming plugin, into the videoroom plugin. The `cache_publisher_keyframe` config option on a room enables this, and keeps the latest keyframe packets from a publisher.

When a subscriber sends a PLI, if we have the keyframe cache enabled, janus sends that cache data instead of forwarding the PLI to the publisher.

In my (simple so far) testing, this works really well. We were facing problems where a subscriber was joining a room with a publisher that couldn't perform a PLI response, and they would just sit there until the next keyframe rolled in. Think live-streaming sources where you're forwarding pre-existing h264 into janus, so your publisher is unable to generate a new PLI. The video for the subscriber does jump a wee bit, but getting slightly jumbled video for a few seconds is better than none at all.

I've opened this PR as a draft to gather initial feedback, and potential things that should / could be improved on.

Let me know what you think 😁

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -120,6 +120,10 @@ room-<unique room ID>: {
 				optionally with a fmtp attribute to match (codec/fmtp properties).
 				If not provided, all codecs enabled in the room are offered, with no fmtp.
 				Notice that the fmtp is parsed, and only a few codecs are supported.
+	cache_publisher_keyframes = true|false (whether to cache keyframes for new subscribers
+				so that they can be sent when a PLI from a subscriber is requested. Note that this
+				stops the PLI from a subscriber being sent to a publisher, as the cached keyframe is
+				sent instead; default=false)
 }
 \endverbatim
  *
@@ -1627,7 +1631,8 @@ static struct janus_json_parameter create_parameters[] = {
 	{"notify_joining", JANUS_JSON_BOOL, 0},
 	{"require_e2ee", JANUS_JSON_BOOL, 0},
 	{"dummy_publisher", JANUS_JSON_BOOL, 0},
-	{"dummy_streams", JANUS_JSON_ARRAY, 0}
+	{"dummy_streams", JANUS_JSON_ARRAY, 0},
+	{"cache_publisher_keyframe", JANUS_JSON_BOOL, 0}
 };
 static struct janus_json_parameter edit_parameters[] = {
 	{"secret", JSON_STRING, 0},
@@ -2035,6 +2040,7 @@ typedef struct janus_videoroom {
 	gboolean check_allowed;		/* Whether to check tokens when participants join (see below) */
 	GHashTable *allowed;		/* Map of participants (as tokens) allowed to join */
 	gboolean notify_joining;	/* Whether an event is sent to notify all participants if a new participant joins the room */
+	gboolean cache_publisher_keyframe; /* Whether to cache keyframes of the publishers */
 	janus_mutex mutex;			/* Mutex to lock this room instance */
 	janus_refcount ref;			/* Reference counter for this room */
 } janus_videoroom;
@@ -2144,6 +2150,16 @@ static GMainLoop *rtcpfwd_loop = NULL;
 static GThread *rtcpfwd_thread = NULL;
 static void *janus_videoroom_rtp_forwarder_rtcp_thread(void *data);
 
+typedef struct janus_videoroom_publisher_keyframe {
+	gboolean enabled;
+	/* If enabled, we store the packets of the last keyframe, to immediately send them for new viewers */
+	GList *latest_keyframe;
+	/* This is where we store packets while we're still collecting the whole keyframe */
+	GList *temp_keyframe;
+	guint32 temp_ts;
+	janus_mutex mutex;
+} janus_videoroom_publisher_keyframe;
+
 typedef struct janus_videoroom_publisher {
 	janus_videoroom_session *session;
 	janus_videoroom *room;	/* Room */
@@ -2246,6 +2262,7 @@ typedef struct janus_videoroom_publisher_stream {
 	janus_mutex subscribers_mutex;
 	volatile gint destroyed;
 	janus_refcount ref;
+	janus_videoroom_publisher_keyframe keyframe;
 } janus_videoroom_publisher_stream;
 /* Helper to add a new RTP forwarder for a specific stream sent by publisher */
 static janus_videoroom_rtp_forwarder *janus_videoroom_rtp_forwarder_add_helper(janus_videoroom_publisher *p,
@@ -2427,6 +2444,13 @@ static void janus_videoroom_publisher_stream_dereference_void(void *ps) {
 	janus_videoroom_publisher_stream_unref((janus_videoroom_publisher_stream *)ps);
 }
 
+static void janus_videoroom_rtp_relay_packet_free(janus_videoroom_rtp_relay_packet *pkt) {
+	if(pkt == NULL)
+		return;
+	g_free(pkt->data);
+	g_free(pkt);
+}
+
 static void janus_videoroom_publisher_stream_free(const janus_refcount *ps_ref) {
 	janus_videoroom_publisher_stream *ps = janus_refcount_containerof(ps_ref, janus_videoroom_publisher_stream, ref);
 	/* This publisher stream can be destroyed, free all the resources */
@@ -2444,6 +2468,13 @@ static void janus_videoroom_publisher_stream_free(const janus_refcount *ps_ref) 
 	janus_mutex_destroy(&ps->subscribers_mutex);
 	janus_mutex_destroy(&ps->rid_mutex);
 	janus_rtp_simulcasting_cleanup(NULL, NULL, ps->rid, NULL);
+
+	janus_mutex_lock(&ps->keyframe.mutex);
+	if(ps->keyframe.latest_keyframe != NULL)
+		g_list_free_full(ps->keyframe.latest_keyframe, (GDestroyNotify)janus_videoroom_rtp_relay_packet_free);
+	ps->keyframe.latest_keyframe = NULL;
+	janus_mutex_unlock(&ps->keyframe.mutex);
+
 	g_free(ps);
 }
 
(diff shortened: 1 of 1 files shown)


chriswiggins:
Some things to think about:

* Work out if the publisher can actually support PLI's for this stream, and if so, don't cache (unless that's overridden), or alternatively, still send the PLI regardless?
* Can we actually cache **all** of the packets since the last keyframe, and send that whole lot out on first connection? Would give a 'fast-start' type approach
* I think I'm missing returning this option being set in some places in the API, so that will need to be fixed
* Should we send the keyframe cache on first connection, rather than when the subscriber requests a PLI?
* Should this also be added into places that request PLIs (like recording start etc)

lminiero:
Thanks for this contribution!

To be honest, I'm not sure how helpful that feature is in general. At the time we added it to the Streaming plugin exactly because of the reasons you listed (getting something on the screen soon for streaming sources that didn't know PLI), but it never worked as expected: while it can get you an image soon, it's always a frozen image, since the keyframe you send to a new subscriber may be from 5 seconds ago, and all real-time frames that are sent after that are completely unrelated to that, which confuse the decoder that just gives up until a real keyframe arrives later on, causing a frozen image for several seconds. Best case scenario, you get some ghosting artifacts rather than a complete freeze. Whether a frozen image for a few seconds is better than no image at all is debatable: in our experience it was never useful, and could actually be harmful, as seeing a frozen image people may think something was wrong and refresh the screen. I'm not sure if anyone else is using that feature in the Streaming plugin, and in fact I sometimes thought of just getting rid of it.

Not sure if your experience has been better with it, but I doubt how the browsers handle that has changed much. Coming to your points:

> * Work out if the publisher can actually support PLI's for this stream, and if so, don't cache (unless that's overridden), or alternatively, still send the PLI regardless?

I think that'd be more trouble than it's worth. IMHO, the application owner knows which source will be "real" and which will be a Streaming-like source, so whether to enable such a feature or not can be left to the API, without self-detection mechanism that may be too complex and still fail.

> * Can we actually cache **all** of the packets since the last keyframe, and send that whole lot out on first connection? Would give a 'fast-start' type approach

Most definitely not... by default, browsers send a keyframe every 3000 frames, which is 100 seconds at 30fps. That's A LOT of packets. Besides, you may lose a keyframe, which means you may go even beyond that with buffering; or a keyframe may never arrive at all. I don't want to turn Janus into a memory hog. That said, it would be useless anyway, as even when the interval is shorter you can't send 10s worth of packets to someone, the network usage would spike incredibly high.

> * Should we send the keyframe cache on first connection, rather than when the subscriber requests a PLI?

If you send it any time they request a PLI, I'm pretty sure the experience would be crappy, exactly because of the reasons explained above: the new frames would clash with the old keyframe, and everything would just freeze.

> * Should this also be added into places that request PLIs (like recording start etc)

I don't think that's a good idea, because of the above.

In a nutshell, while I appreciate the effort, I'm not entirely sold on whether this is a good idea or not. Hopefully users will start playing with it to share their opinion and feedback so that we have a clearer idea, since mine is mostly based on those failed experiments some time ago.

atoppi:
Thanks for the effort @chriswiggins.
I agree with Lorenzo's review, keyframe caching may sound a good idea but in practice (according to our experience) the UX has slight to none benefits.

As of your open questions, I'm strongly against the caching of all packets and the subsequent forwarding of all the content upon a user connection. On top of the issues mentioned by Lorenzo, I am sure that RTP stacks will be highly confused by the tremendous spike, leading to issues in operating the jitter buffer and possibly in lip-syncing.

In case of streaming a pre-recorded content, I guess the best approach would be re-encoding the media by increasing the frequency of KFs, ideally close to a live-streaming interval (e.g. 1 each 3-5 secs).

Nonetheless I'm not against merging the feature in a opt-in manner for a user, similarly to what we do in the streaming plugin.
Maybe some tweaking in the encoded media might lead to an enhanced UX, so imho it's better to give the user the chance to try by himself.

lminiero:
> I am sure that RTP stacks will be highly confused by the tremendous spike, leading to issues in operating the jitter buffer and possibly in lip-syncing

Good point, which makes me think of something else. One of the reasons why it's working poorly in the Streaming plugin may be that, when we send the keyframe at startup (in `setup_media`, which your patch doesn't seem to be doing though), we're actually sending old packets, with old timestamps and sequence numbers; when we start sending regular packets after that, the browser can see a change in timestamps and missing sequence numbers, which may cause the stream to break until the next keyframe. I guess a potential way to address that is to artificially change the SSRC in the RTP header context after sending the stored keyframe, which would force the plugin to align the incoming headers to match what was sent shortly before. But I'm just thinking out loud, and I'm not sure it would indeed make a difference: I'll make a couple of tests later to see if it does.

chriswiggins:
> 
> To be honest, I'm not sure how helpful that feature is in general. At the time we added it to the Streaming plugin exactly because of the reasons you listed (getting something on the screen soon for streaming sources that didn't know PLI), but it never worked as expected: while it can get you an image soon, it's always a frozen image, since the keyframe you send to a new subscriber may be from 5 seconds ago, and all real-time frames that are sent after that are completely unrelated to that, which confuse the decoder that just gives up until a real keyframe arrives later on, causing a frozen image for several seconds. Best case scenario, you get some ghosting artifacts rather than a complete freeze. Whether a frozen image for a few seconds is better than no image at all is debatable: in our experience it was never useful, and could actually be harmful, as seeing a frozen image people may think something was wrong and refresh the screen. I'm not sure if anyone else is using that feature in the Streaming plugin, and in fact I sometimes thought of just getting rid of it.

This is the thing - in my testing with Chrome, the keyframe kicks off the decoding process. Yes you get slightly jumbled video, but for us, having something (even if slightly crappy) for a few seconds is better than sitting at a black screen with nothing at all. I'd have to do testing on other decoders to see how they fair up to know for sure if this is the same on other decoders. The SSRC being set to 1 seems to signal to the decoder to start, and the awesome-ness inside janus automatically handles the fact that it appears as out of order to the normal stream

> 
> Not sure if your experience has been better with it, but I doubt how the browsers handle that has changed much. Coming to your points:
> 
> > * Work out if the publisher can actually support PLI's for this stream, and if so, don't cache (unless that's overridden), or alternatively, still send the PLI regardless?
> 
> I think that'd be more trouble than it's worth. IMHO, the application owner knows which source will be "real" and which will be a Streaming-like source, so whether to enable such a feature or not can be left to the API, without self-detection mechanism that may be too complex and still fail.
> 

Fair enough - as I was writing the API into the videoroom-level config, I was thinking it could potentially also be a config option on the producer / producer stream as well, to make it more selective, so I might try that.

> > * Can we actually cache **all** of the packets since the last keyframe, and send that whole lot out on first connection? Would give a 'fast-start' type approach
> 
> Most definitely not... by default, browsers send a keyframe every 3000 frames, which is 100 seconds at 30fps. That's A LOT of packets. Besides, you may lose a keyframe, which means you may go even beyond that with buffering; or a keyframe may never arrive at all. I don't want to turn Janus into a memory hog. That said, it would be useless anyway, as even when the interval is shorter you can't send 10s worth of packets to someone, the network usage would spike incredibly high.
> 

Good point. I did try this in a bit of a hack and you're right - it was terrible! Your response about browsers sending keyframes at that sort of interval is a good discussion point. We don't have browser producers - they're GStreamer producers, and the sources behind them would never send a keyframe that occasionally. The most I'd ever expect is 5-10 seconds (and that's not that usual). Most sources send keyframes between 1-5 seconds. You might ask why we're not using the streaming plugin, and the reasons for that is that it's actually far easier to establish a WebRTC connection 😄 

> > * Should we send the keyframe cache on first connection, rather than when the subscriber requests a PLI?
> 
> If you send it any time they request a PLI, I'm pretty sure the experience would be crappy, exactly because of the reasons explained above: the new frames would clash with the old keyframe, and everything would just freeze.
> 

Again, I'm not experiencing the freezing you talk about. Yes the experience is potentially crappy, but again, for us, that's better than missed video. I might try sending it on first connection and see how that differs to sending it when a PLI comes through, as that might make the decoder even happier with the weird ordering of packets

> > * Should this also be added into places that request PLIs (like recording start etc)
> 
> I don't think that's a good idea, because of the above.
> 
> In a nutshell, while I appreciate the effort, I'm not entirely sold on whether this is a good idea or not. Hopefully users will start playing with it to share their opinion and feedback so that we have a clearer idea, since mine is mostly based on those failed experiments some time ago.

Firstly, can I say this is one of the nicest sets of constructive criticism feedback I've ever seen on a PR! This feature is definitely for more niche use-cases. I agree wholeheartedly it doesn't suit browser-based producers, or producers that **can** send PLIs. It's for those cases where you have a source that sends keyframes relatively frequently, but not often enough that it can take a bit too long to see video.

Do you guys have a workflow whereby you check for memory leaks? I'd love to try this in production, but while I know enough C to be dangerous, the last thing I'd want is to introduce a memory leak in our janus instances handling ~1Gbps of traffic 🤣

chriswiggins:
Sorry @atoppi I missed your reply when replying to @lminiero!

> In case of streaming a pre-recorded content, I guess the best approach would be re-encoding the media by increasing the frequency of KFs, ideally close to a live-streaming interval (e.g. 1 each 3-5 secs).
> 

Thats what we want to avoid. If we're getting good h264 video from the source, re-encoding it is just unnecessary compute power - trust me, we used to do this for HLS (so that we could have a keyframe every second) to reduce the latency introduced, and it was terrible 😆 



> Good point, which makes me think of something else. One of the reasons why it's working poorly in the Streaming plugin may be that, when we send the keyframe at startup (in `setup_media`, which your patch doesn't seem to be doing though), we're actually sending old packets, with old timestamps and sequence numbers; when we start sending regular packets after that, the browser can see a change in timestamps and missing sequence numbers, which may cause the stream to break until the next keyframe. I guess a potential way to address that is to artificially change the SSRC in the RTP header context after sending the stored keyframe, which would force the plugin to align the incoming headers to match what was sent shortly before. But I'm just thinking out loud, and I'm not sure it would indeed make a difference: I'll make a couple of tests later to see if it does.

If I can help with the tests, let me know! In my limited testing if I didn't set the SSRC to 1, then the decoder ignored the packet (or it was jumbled by janus) because it was so far out from the actual SSRC of existing packets. Is the fact I'm sending this when a PLI is requested possibly aiding in the decoding process, rather than on initial connection?

fippo:
This can't work (or is not a good idea for most use-cases)... 🐻 with me.

Video frames are encoded with references to previous frames. In WebRTC that typically means
KF(1) <-> DELTA(2) <- DELTA(3) <- DELTA(4) <- DELTA(5) <- ...
Frame 5 can only properly be decoded if you have decoded all the frames referenced by it (which is 4, 4 can only be decoded if you decode 3, ...)

What you are proposing effectively means that the decoder gets presented with
KF(1) <- DELTA(5)
but since H264 (unlike VP8) lacks actual picture ids, neither the frame buffer nor the decoder know that they are trying to decode frame 5 which references 4 but you are giving it the keyframe (which it needs).

In VPx this leads to garbled output (unless there are picture ids in which case you will see a freeze and a PLI). Like this
image

In H264 this kinda works. Kinda... the result is smearing which largely depends on how much movement there was in the frames that aren't decoded. If you have a lot of static parts of the image those will be ok.
That may work sometimes. But there is no way to measure how well even.

lminiero:
Thanks for the analysis @fippo! This seems to confirm that the feature would be problematic to have in the plugin, and only marginally helpful (if not possibly confusing to those not aware of the limitations and impact). @chriswiggins I'm afraid I'm more and more inclined not to consider this for merging, especially considering no one else vouched support for this either.

lminiero:
Closing because of the considerations made above. Thanks for the contribution anyway @chriswiggins! :heart:

PR #3069, opened 2022-09-16
replace sha1 to sha256 at self signed dtls cert, more verbose error logging

tgabi333:
When Janus start it looks for an existing cert, if there is no, it will generate a self signed one.

On newer operating systems (eg RHEL9) OpenSSL SECLEVEL 2 is the default, which does not allow creation of SHA1 signatures. 

SHA1 is replaced with SHA256.


lminiero:
Thanks, merging!

PR #3074, opened 2022-09-21
Add descriptive errors on user event exceptions

davel:


lminiero:
Thanks, but not sure why those are needed: aren't the line numbers the error log shows enough to allow you to figure out where the issue happened?

davel:
> Thanks, but not sure why those are needed: aren't the line numbers the error log shows enough to allow you to figure out where the issue happened?

This is helpful to us as in our application we report errors seen in the wild back to the server. We thus don't have access to the full backtrace like you would in the browser debugger.

html/janus.js
```diff
@@ -1915,7 +1915,7 @@ function Janus(gatewayCallbacks) {
 			try {
 				pluginHandle.onremotetrack(event.track, mid, true);
 			} catch(e) {
-				Janus.error(e);
+				Janus.error("Error calling track event handler", e);
```
lminiero, line 1918: I'd say `onremotetrack` instead of `track event handler`, since we have two different handlers for tracks, depending on whether they're local or remote. Same for the lines you changed below.

html/janus.js
```diff
@@ -2615,7 +2615,7 @@ function Janus(gatewayCallbacks) {
 						config.myStream.removeTrack(rt);
 						pluginHandle.onlocaltrack(rt, false);
 					} catch(e) {
-						Janus.error(e);
+						Janus.error("Error on removing local track", e);
```
lminiero, line 2618: You have the same error message being printed for two different things. This one is a track being removed as a consequence of a rinegotiation the application is asking for, and so is a manual removal; the one below is a track being removed because `onended` was called on the track, and so often a renegotiation originated by the remote peer. If you're interested in unique log lines to sort out the cause of a problem, I'd write them differently.

lminiero (commented):
Added a couple of notes inline.

lminiero:
@davel did you have a chance to review the notes I added? Thanks!

davel:
> @davel did you have a chance to review the notes I added? Thanks!

Yes, they make sense to me. Sorry, I'll update soon.

lminiero:
Thanks @davel! Looks good, merging then :v:

PR #3075, opened 2022-09-22
Add versioning to .so files

lminiero:
See discussion on the [group](https://groups.google.com/g/meetecho-janus/c/ESCW9mQJpYc) for the rationale. Shouldn't impact new or existing deployments, but adding it as a PR to have a reference in the changelog when it's merged.


lminiero:
Merging.

PR #3087, opened 2022-10-14
Enforce media direction policies when SIP call is on hold

lminiero:
This is a PR inspired by [this post](https://groups.google.com/g/meetecho-janus/c/c8Rs5RVM6ec) on the group, where setting a call on hold doesn't seem to be working as expected. In this specific case, it looks like Janus is correctly negotiating sendonly when putting a call on hold, but the call is then handled via a B2BUA, that sets sendrecv on the other end: since we don't have checks on whether we should actually send a packet, or accept an incoming one, when the call is on hold, this results in media still flowing in both directions when it shouldn't.

This patch adds such checks, and adds an enforced check on receiving packets as well when not on hold, which seemed to be missing. I didn't test this yet, so you may want to check whether or not this introduces any regression. While the patch is for `0.x` (that's what the OP is using), I'll port it to `master` as well in case it's considered safe to merge.


lminiero:
Merging.

PR #3088, opened 2022-10-18
audiobridge: memory leak fixed

RSATom:
in participant sender thread.

The problem is [it has more conditions](https://github.com/meetecho/janus-gateway/blob/1870a5a5e32ab5fde07793ea58069cc29ea9b78d/src/plugins/janus_audiobridge.c#L8565) than just check `mixedpkt` for `NULL`...


lminiero:
Good catch! Merging :+1: 
I'll make sure the same fix is applied to `0.x` too.

PR #3093, opened 2022-10-25
Added code to send PLI to SIP peer when recording (fixes #3092)

### LINKED ISSUE #3092: [1.1.0]  Janus sip plugin recorded peer video is black, because of unimplemented keyframe request
1.1.0

Yes, Master

**Was this working before?**
No

No

**Additional context**
I see that in the source code of sip plugin, when recording is enabled, we don't request PLI. So when I try to play, recorded video is black.   Blame: [See source code here](https://github.com/meetecho/janus-gateway/blob/5d55eeb2e17c99a258e8448b815b179729524ddf/src/plugins/janus_sip.c#L4556)

I just want to ask if there is an easy way to get peer session reference and send PLI as we do it [here](https://github.com/meetecho/janus-gateway/blob/5d55eeb2e17c99a258e8448b815b179729524ddf/src/plugins/janus_sip.c#L4623):

lminiero:
As the title says, this patch tries to ensure we do send a PLI to the SIP peer as well when recording, a problem highlighted in #3092. More specifically, the patch first of all checks if PLI support was negotiated in the SDP by the peer, and only in that case it sends a PLI when needed (as the peer may not be a browser, and so may not know anything about what a PLI is). I guess in the future we may want to add a way to send a FIR too (for interaction with legacy devices), but it's not what this PR is about.

I have not tested yet due to lack of time, so please test and provide feedback, so that this can be merged soon. If it works as expected, I'll backport this to the `0.x` branch as well.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1035,6 +1035,7 @@ typedef struct janus_sip_media {
 	srtp_policy_t video_remote_policy, video_local_policy;
 	char *video_srtp_local_profile, *video_srtp_local_crypto;
 	gboolean video_send, video_recv;
+	gboolean video_pli_supported;
 	janus_sdp_mdirection hold_video_dir, pre_hold_video_dir;
 	janus_rtp_switching_context acontext, vcontext;
 	int pipefd[2];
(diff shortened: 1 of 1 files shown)


lminiero:
Merging.

PR #3094, opened 2022-10-27
Send PLI in sip plugin Fixes #3092 enhances #3093

### LINKED ISSUE #3092: [1.1.0]  Janus sip plugin recorded peer video is black, because of unimplemented keyframe request

tvildo:
Hello, I added additional option based on [pull request ](https://github.com/meetecho/janus-gateway/pull/3093)to force send PLI if detection fails .

**Why that is required?** 
I use asterisk which has only Janus clients(web browsers), problem is that asterisk by default won't generate SDP with r`tcp-fb`.  There is a option in asterisk [see link](https://github.com/asterisk/asterisk/blob/005b6c8d50441ef6e3bed70c71807a5a8861bb45/res/res_pjsip_sdp_rtp.c#L1329) which allows to enable `rtcp-fb` from `pjsip.conf` config:

```
[111@test.com]
auth=111@test.com
aors=111@test.com
type=endpoint
context=test
disallow=all
allow=alaw,ulaw,vp8,vp9,h264
webrtc=yes <--- This option
```
But when I enable that option in asterisk call from Janus fails, after accepting it is hanged instantly. That option will also enable additional things like[ see link](https://github.com/asterisk/asterisk/blob/005b6c8d50441ef6e3bed70c71807a5a8861bb45/res/res_pjsip/pjsip_config.xml#L1378) so maybe there is some incompatibility between Janus <---> Asterisk after that is enabled in asterisk.

So my solution was to bypass automatic detection and allow to force send PLI when recording is started

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -826,6 +826,7 @@ static struct janus_json_parameter recording_parameters[] = {
 	{"video", JANUS_JSON_BOOL, 0},
 	{"peer_audio", JANUS_JSON_BOOL, 0},
 	{"peer_video", JANUS_JSON_BOOL, 0},
+	{"send_peer_pli", JANUS_JSON_BOOL, 0},
 	{"filename", JSON_STRING, 0}
 };
 static struct janus_json_parameter dtmf_info_parameters[] = {
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -531,6 +531,7 @@
 	"video" : <true|false; whether or not our video should be recorded>,
 	"peer_audio" : <true|false; whether or not our peer's audio should be recorded>,
 	"peer_video" : <true|false; whether or not our peer's video should be recorded>,
+	"send_peer_pli" : <true|false; whether or not send pli to request keyframe from peer>,
 	"filename" : "<base path/filename to use for all the recordings>"
 }
 \endverbatim
@@ -4501,7 +4502,7 @@ static void *janus_sip_handler(void *data) {
 				goto error;
 			}
 			gboolean record_audio = FALSE, record_video = FALSE,	/* No media is recorded by default */
-				record_peer_audio = FALSE, record_peer_video = FALSE;
+				record_peer_audio = FALSE, record_peer_video = FALSE, send_peer_pli = FALSE;
 			json_t *audio = json_object_get(root, "audio");
 			record_audio = audio ? json_is_true(audio) : FALSE;
 			json_t *video = json_object_get(root, "video");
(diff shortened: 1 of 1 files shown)


lminiero:
> But when I enable that option in asterisk call from Janus fails, after accepting it is hanged instantly

I think that's expected. Setting webrtc=yes probably means Asterisk expects to use ICE+DTLS on that call, which would not happen with the SIP plugin (the SIP plugin acts as a "regular" softphone, no WebRTC capabilities), and so should indeed not be enabled.

That said, you say this extends #3093, but then based this PR on master. This is causing many more commits to be displayed, and since you've integrated other commits that came after that, this is hard to follow. Please rebase your branch on top of that PR branch (which hasn't been merged yet) and update the base in this PR accordingly (I tried to do it now but, again, it's displaying unrelated commits as if they were yours).

lminiero:
I just merged the PR, so rebasing should be easier.

lminiero:
@tvildo any update on the rebase?

tvildo:
@lminiero  Hi, I did rebase.

lminiero:
Mh doesn't look like you did a rebase, as I still see my own new commits in there. Maybe it's easier to apply your patch in a new PR?

tvildo:
@lminiero  Please check, I think I did it right

src/plugins/janus_sip.c
```diff
@@ -4510,6 +4511,8 @@ static void *janus_sip_handler(void *data) {
 			record_peer_audio = peer_audio ? json_is_true(peer_audio) : FALSE;
 			json_t *peer_video = json_object_get(root, "peer_video");
 			record_peer_video = peer_video ? json_is_true(peer_video) : FALSE;
+			json_t *peer_pli = json_object_get(root, "send_peer_pli");
```
lminiero, line 4514: Since you added a new argument, this needs to be added to the automated validation ("is this a boolean?"), otherwise people may send a different type causing issues. You can check where "peer_video" and "peer_audio" are validated and add your new property there too: this is currently happening in `recording_parameters` in the code.

lminiero (commented):
Thanks! I do see a proper rebase now. I've added a single note inline.

tvildo:
@lminiero  Can you take a look again ?

src/plugins/janus_sip.c
```diff
@@ -531,6 +531,7 @@
 	"video" : <true|false; whether or not our video should be recorded>,
 	"peer_audio" : <true|false; whether or not our peer's audio should be recorded>,
 	"peer_video" : <true|false; whether or not our peer's video should be recorded>,
+	"send_peer_pli" : <true|false; whether or not send pli to request keyframe from peer>,
```
lminiero, line 534: s/send pli/send a PLI

lminiero (commented):
Apologies for the delay, @tvildo: the validation looks fine, thanks! I added a very minor note that's simply editorial (we prefer referring to RTCP feedback by using their upper case names) for the documentation, but otherwise it's fine to me, so once you change that bit it can be merged for me :+1:

PR #3096, opened 2022-10-31
janus_streaming: remove always false branch

tmatth:
No functional change, this is just for clarity.

Since every branch of the previous `if (session->mountpoint != NULL)` conditional terminates in either `goto error` or `goto done`, there is no case where `session->mountpoint` will be non-null when arriving at this check.


tmatth:
For context we are trying to debug what seems to be a deadlock in the streaming plugin API that at some point in the sequence logs an `Already watching a stream...` error that I was going to add more detail to in order to distinguish if the error was being `session->mountpoint` was NULL vs. the session was found in the list `mp->viewers` when I realized that it could only ever be the latter case.

lminiero:
Yep, good catch: initially I was wondering why you though the `session->mountpoint` check was unneeded, and then realized it was in "watch" and after checking if this was for a renegotiation.

lminiero:
Merging :+1:

PR #3098, opened 2022-11-02
janus_streaming: remove always false branch

tmatth:
No functional change, this is just for clarity.

Since every branch of the previous `if (session->mountpoint != NULL)` conditional terminates in either `goto error` or `goto done`, there is no case where `session->mountpoint` will be non-null when arriving at this check.

(cherry picked from commit de77c51eed223efda588e38376a75b72ae8d65fd)

This is a backport of #3096


lminiero:
Thanks, merging :+1:

Issue #3105, opened 2022-11-17
[0.x] janus.plugin.streaming: Already watching a stream


tmatth:
We believe we're seeing some spurious "Already watching a stream..." errors in the `janus.plugin.streaming` plugin in production, but AFAICT for this to be true 2 conditions must be met:

1. there's a dangling viewer in the `mp->viewers` list (or a race condition on removing/adding a viewer to this list)
2. since the `viewers` list is checked using `g_list_find`, it's possible that a new `viewer` is being assigned the same address (i.e. by malloc) as a previously allocated viewer that never got removed from the list, and so it's being rejected there.

For the second condition, I was thinking it might be more robust to use `g_list_find_custom` instead of just `g_list_find` and check some other unique-ish identifier like `session->sdp_sessid` or even `session->handle` (although this would potentially have the same drawback).

So instead of:

```
// Assume mp->viewers consisting of one viewer 0x1234 
// Assume mp->viewers gets freed but is not removed from the list for some reason
// Assume new session is allocated also with address 0x1234
g_list_find(mp->viewers, session);  // oh no, collision! 
```
we do
```
g_list_find_by_custom(mp->viewers, session, viewer_comparator)
```
where viewer_comparator checks some other field for safety.

**What version of Janus is this happening on?**
0.x (commit 34e3f4abc87ac546d22aeea3dc1bb1af0959cfe7)

**Have you tested a more recent version of Janus too?**
No.

**Was this working before?**
No.

**Is there a gdb or libasan trace of the issue?**
Not Yet

**Additional context**
Since we're at a loss for how to reproduce this, we may just start by submitting PRs to get more context/confirm our suspicion. We've been looking over all the viewer list access synchronization to see if there's an issue there but so far it's not obvious.

Just to clarify I doubt that this is specific to 0.x but we haven't been able to try 1.x in the same context so we can't be sure.

lminiero:
I'm not sure I'm fond of the `g_list_find_by_custom`, which seems more work and slower, as it requires dereferencing a pointer in the traversal of the list where right now it's a simple comparison. I'd rather try and figure out if there's indeed circumstances where a pointer may not be removed from the list when it should. If it's a matter of re-used pointers, as you said, it could be a race condition of some sort, which should then be fixed properly, where I see `g_list_find_by_custom` more as a temporary patch.

tmatth:
[...]

PR #3106, opened 2022-11-17
janus_streaming: enrich already watching error

tmatth:
This includes the address (the session pointer) that we found in the mp->viewers list, as well as the id of the mountpoint in question.

Refs #3105


lminiero:
Ack, merging! Please let us know if you find more info on the root issue.

lminiero:
PS: was this meant for `0.x` rather than master, considering #3105 referenced that other branch?

tmatth:
> PS: was this meant for `0.x` rather than master, considering #3105 referenced that other branch?

I try to always patch against master first than backport to 0.x (unless the code is only in one and not the other). I will open a PR for that now :)

PR #3110, opened 2022-11-28
Capture potential null access bugs in janus.js

veeting:
Hi Lorenzo,

We have recently seen some exceptions in our logs that indicate that config.pc was null when trying to call getTransceivers(). We don't know why it is happening, from the logs it seams to only happen on Android devices so far.

This PR adds three null checks to mitigate this potential issues.

Kind regards,

Fabian

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1941,8 +1941,9 @@ function Janus(gatewayCallbacks) {
 					trackMutedTimeoutId = setTimeout(function() {
 						Janus.log('Removing remote track');
 						// Notify the application the track is gone
-						let transceiver = config.pc.getTransceivers().find(
-							t => t.receiver.track === ev.target);
+						let transceivers = config.pc ? config.pc.getTransceivers() : null;
+						let transceiver = transceivers ? transceivers.find(
+							t => t.receiver.track === ev.target) : null;
 						let mid = transceiver ? transceiver.mid : ev.target.id;
 						try {
 							pluginHandle.onremotetrack(ev.target, mid, false);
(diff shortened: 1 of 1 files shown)


html/janus.js
```diff
@@ -1941,8 +1941,9 @@ function Janus(gatewayCallbacks) {
 					trackMutedTimeoutId = setTimeout(function() {
 						Janus.log('Removing remote track');
 						// Notify the application the track is gone
-						let transceiver = config.pc.getTransceivers().find(
-							t => t.receiver.track === ev.target);
+						let transceivers = config.pc ? config.pc.getTransceivers() : null;
+						let transceiver = transceivers ? transceivers.find(
+							t => t.receiver.track === ev.target) : null;
```
atoppi, line 1946: since `pc.getTransceivers()` will always return an array, everything could be simplified in a form more similar to the original code

```js
let transceiver = config.pc ? config.pc.getTransceivers().find(t => t.receiver.track === ev.target) : null;
```

veeting:
Thanks, @atoppi 

Your short hand writing is nicer, however, I tried to be consistent with similar checks which already exist in master, for instance:

https://github.com/meetecho/janus-gateway/blob/master/html/janus.js#L1928

Would you prefer me to

1. leave the code in this PR as is, to match the existing code, or
2. just rewrite this PR, or
3. rewrite all occurrences in the existing code to match your suggestion?

I am happy with all, however, I would prefer 1 or 3, to keep the code consistent.

atoppi:
I see.
Okay, let's keep the code consistent: option no. 1

lminiero:
Merging then, thanks! I'll check if the same fix needs to be packported to `0.x`.

PR #3113, opened 2022-11-29
Update to webrtc adapter 8.2.0

tmatth:


lminiero:
I guess we should find a way to make the script line something that just needs to be fixed once so that it works for all pages :grin:  Merging, thanks!

PR #3116, opened 2022-12-02
Add support for L16 codec (uncompressed audio)

lminiero:
One of the hidden "secrets" in Chrome is that you can tell it, by munging the SDP, to send RTP packets with uncompressed audio data, instead of encoding them. This means that, instead of using Opus, G.711 or whatever else, the samples are sent exactly as they're captured. This audio format is called [L16](https://datatracker.ietf.org/doc/html/rfc1890#section-4.4.8), and uses two bytes per sample.

In general, you can enable it in the SDP by using the `L16/<samplerate>/<channels>` rtpmap in the SDP, e.g.:

    a=rtpmap:111 L16/16000

Chrome supports a few variants of those (not sure how many, I haven't tried them all), so to start using it with Janus I added support for two different versions: `L16/16000` (uncompressed wideband) and `L16/48000` (uncompressed fullband), in both cases mono only and not stereo (there will be time to expand on this).

In case you're wondering why support this at all (as uncompressed audio is very inefficient to send around), the main reason is facilitating the experimentation with WASM audio codecs. You may have heard about Lyra, a new audio codec with very low bandwidth: a WASM version of it has already surfaced, and so it would be cool to see how it works in a WebRTC context, even if it's not available as a codec in browsers themselves. A way to do that is basically use a combination of WASM codecs, L16 samples, and Insertable Frames: you basically ask the browser for uncompressed audio samples, and then you encode/decode them yourself, using Insertable Streams to "replace" the uncompressed audio data in the original RTP packets with the bytes that have been encoded with WASM. In order to test all this with Janus, you need support for L16 negotiation, which is what we're adding here.

I tested this briefly with plain L16, and it seems to be working as expected. The patch also supports postprocessing L16 recordings (which of course only works if the RTP packets actually contain L16 data, and not Insertable Streams replacements). In case you do plan to experiment with WASM codecs and Janus as well, please do let me know, so that we can share some results.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -558,6 +558,8 @@ janus_pp_rec_SOURCES = \
 	postprocessing/pp-g711.h \
 	postprocessing/pp-g722.c \
 	postprocessing/pp-g722.h \
+	postprocessing/pp-l16.c \
+	postprocessing/pp-l16.h \
 	postprocessing/pp-h264.c \
 	postprocessing/pp-h264.h \
 	postprocessing/pp-av1.c \
--- a/src/postprocessing/janus-pp-rec.c
+++ b/src/postprocessing/janus-pp-rec.c
@@ -129,6 +129,7 @@ Usage: janus-pp-rec [OPTIONS] source.mjr
 #include "pp-opus.h"
 #include "pp-g711.h"
 #include "pp-g722.h"
+#include "pp-l16.h"
 #include "pp-srt.h"
 #include "pp-binary.h"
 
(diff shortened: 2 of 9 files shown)


fippo:
can you please file an issue in https://github.com/w3c/webrtc-nv-use-cases requesting access to raw audio data?
The not-really-announced support for L16 has blown up earlier this year after a certain vendor did very much non-spec things with it. I think L16 is quite interesting as a standardized format for raw data.

lminiero:
Merging.

baudneo:
Very nice, a project wz_mini_hacks, adds RTSP server to wyze cameras. It sends v2/v3 audio as `L16/44100`

```a=rtpmap:97 L16/44100````

 At the moment, it causes the stream not to play in chrome until the audio stream is removed. Looking forward to support for more sampling rates.

Thank you for your work!

PR #3121, opened 2022-12-13
Fix overwriting of 7-bit PictureID when doing VP8 simulcast

lminiero:
We noticed a problem in our simulcast management when using non-browser sources, specifically an FFmpeg source for RTP packets. When doing VP8 simulcast, the server needs to ensure monotonically increasing PictureID values, which is what we do in Janus. Older versions of FFmpeg (before this was [patched](https://github.com/FFmpeg/FFmpeg/commit/8ae15b565533944d042d3caf25f7262e002e8953)), though, used 7-bit PictureID values in the VP8 payload header, and we found out our code didn't handle that well, since we were always assuming 15-bit values when overwriting it: this resulted in the payload getting one byte "eaten" away, and the VP8 payload to be corrupted as well.

This patch tries to address that, and to ensure we don't overflow either 7-bit or 15-bit PictureID values when an overwrite process is taking place. If you're using simulcast, please do test this out to ensure we're not introducing any regression, and that it still works as expected for you. Feedback on whether you notice something wrong in the code is welcome too, of course!

Edit: side note, as explained [here](https://groups.google.com/g/discuss-webrtc/c/333-L02vuWA), browsers don't really support 7-bit properly anyway (there's a [bug](https://bugs.chromium.org/p/webrtc/issues/detail?id=7675) open), so our fix for 7-bit PictureID improves things but doesn't fix them when 7-bit values are used (video still freezes occasionally). That said, I still think it's important we fix the behaviour in Janus, in case it's fixed in browsers too eventually.

### DIFF
--- a/fuzzers/rtp_fuzzer.c
+++ b/fuzzers/rtp_fuzzer.c
@@ -92,12 +92,13 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 	janus_h264_is_keyframe(payload, plen);
 
 	/* VP8 targets */
+	gboolean m = FALSE;
 	uint16_t picid = 0;
 	uint8_t tlzi = 0, tid = 0, ybit = 0, keyidx = 0;
 	janus_vp8_simulcast_context vp8_context;
 	memset(&vp8_context, 0, sizeof(janus_vp8_simulcast_context));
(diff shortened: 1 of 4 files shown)


lminiero:
Merging.

PR #3122, opened 2022-12-13
Return a proper error when an incoming request can't be parsed (fixes #3120)

### LINKED ISSUE #3120: [0.x] Two Janus requests in one websocket message results in 'Missing mandatory element (transaction)'
0.11.2

No, though I've inspected the code and, if I'm correct in my current assessment, the issue will happen in 'master' as well.

**Was this working before?**
I'm not sure, but I suspect not.

Unfortunatley not gdb or libasan. I can share wireshark logs though if useful.

**Additional context**
We were seeing this error in the title from Janus intermittently. Digging into wireshark logs I see this occurs when Janus would be processing a single websocket message with two janus request message objects in it.

Looking at [janus_websockets.c](https://github.com/meetecho/janus-gateway/blob/f5208073e702606ab338a31fa3dfbf38651d8e6c/src/transports/janus_websockets.c#L1314) it doesn't look like we support this properly with the current handling of `json_loads`. 

Following this code through to [janus.c](https://github.com/meetecho/janus-gateway/blob/f5208073e702606ab338a31fa3dfbf38651d8e6c/src/janus.c#L1044) it looks like nothing detects the failure to parse from `json_loads` until we check for the required request parameters and find that the object doesn't have a 'transaction' element (in fact its a null object, or so I believe).

I have work in progress (just getting setup to test now) to correctly handle multiple json objects in one websocket message, and to detect and report any json parsing errors directly. Since I haven't validated this through tests, I haven't made that PR but am happy to do so with current state if it helps the discussion.

lminiero:
As the title says: you can find more context in the #3120 discussion.
Planning to merge quite soon, unless there's objections. When that happens, I'll backport the fix to `0.x` too.

### DIFF
--- a/src/janus.c
+++ b/src/janus.c
@@ -1021,6 +1021,12 @@ int janus_process_incoming_request(janus_request *request) {
 		JANUS_LOG(LOG_ERR, "Missing request or payload to process, giving up...\n");
 		return ret;
 	}
+	janus_session *session = NULL;
+	janus_ice_handle *handle = NULL;
+	if(request->message == NULL) {
+		ret = janus_process_error_string(request, 0, NULL, JANUS_ERROR_INVALID_JSON, (char *)"Invalid Janus API request");
+		goto jsondone;
+	}
 	int error_code = 0;
 	char error_cause[100];
 	json_t *root = request->message;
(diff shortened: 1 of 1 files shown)


jwittner (changes requested):
I have made a [draft PR ](https://github.com/meetecho/janus-gateway/pull/3123) which works to support multiple requests as well as improves the error reporting.

jwittner:
I published the above PR after getting good results in testing today.

lminiero:
Closing as superseded by #3123.

PR #3123, opened 2022-12-13
Support multiple requests in a single websocket message

### LINKED ISSUE #3120: [0.x] Two Janus requests in one websocket message results in 'Missing mandatory element (transaction)'

jwittner:
* Parses all json objects as individual requests
* Reports request parsing error details

Resolves #3120

### DIFF
--- a/src/janus.c
+++ b/src/janus.c
@@ -3383,7 +3383,7 @@ void janus_transport_incoming_request(janus_transport *plugin, janus_transport_s
 	if(message == NULL) {
 		/* Report details about the parsing error */
 		janus_process_error(request, /*session_id=*/0, /*transaction=*/NULL, JANUS_ERROR_INVALID_JSON,
-			"Failed loading json request - %s at %s,(%d:%d)", error->text, error->source, error->line, error->column);
+			"Failed loading json request - %s from %s(%d:%d)", error->text, error->source, error->line, error->column);
 		janus_request_destroy(request);
 	}
 	else {
--- a/src/janus.c
+++ b/src/janus.c
@@ -3382,8 +3382,8 @@ void janus_transport_incoming_request(janus_transport *plugin, janus_transport_s
 	janus_request *request = janus_request_new(plugin, transport, request_id, admin, message);
 	if(message == NULL) {
 		/* Report details about the parsing error */
-		janus_process_error(request, /*session_id=*/0, /*transaction=*/NULL, JANUS_ERROR_INVALID_JSON,
-			"Failed loading json request - %s from %s(%d:%d)", error->text, error->source, error->line, error->column);
+		janus_process_error(request, 0, NULL, JANUS_ERROR_INVALID_JSON,
+			"Failed loading JSON request - %s from %s(%d:%d)", error->text, error->source, error->line, error->column);
 		janus_request_destroy(request);
 	}
 	else {
--- a/src/janus.c
+++ b/src/janus.c
@@ -3380,8 +3380,16 @@ void janus_transport_incoming_request(janus_transport *plugin, janus_transport_s
 	JANUS_LOG(LOG_VERB, "Got %s API request from %s (%p)\n", admin ? "an admin" : "a Janus", plugin->get_package(), transport);
 	/* Create a janus_request instance to handle the request */
 	janus_request *request = janus_request_new(plugin, transport, request_id, admin, message);
-	/* Enqueue the request, the thread will pick it up */
-	g_async_queue_push(requests, request);
+	if(message == NULL) {
+		/* Report details about the parsing error */
+		janus_process_error(request, /*session_id=*/0, /*transaction=*/NULL, JANUS_ERROR_INVALID_JSON,
+			"Failed loading json request - %s at %s,(%d:%d)", error->text, error->source, error->line, error->column);
+		janus_request_destroy(request);
+	}
+	else {
+		/* Enqueue the request, the thread will pick it up */
+		g_async_queue_push(requests, request);
+	}
 }
 
 void janus_transport_gone(janus_transport *plugin, janus_transport_session *transport) {
--- a/src/janus.c
+++ b/src/janus.c
@@ -899,13 +899,18 @@ static void janus_request_free(const janus_refcount *request_ref) {
 		janus_refcount_decrease(&request->instance->ref);
 	request->instance = NULL;
 	request->request_id = NULL;
-	if(request->message)
+	if(request->message) {
 		json_decref(request->message);
-	request->message = NULL;
+		request->message = NULL;
+	}
+	if(request->error) {
+		g_free(request->error)
+		request->error = NULL;
+	}
 	g_free(request);
 }
 
-janus_request *janus_request_new(janus_transport *transport, janus_transport_session *instance, void *request_id, gboolean admin, json_t *message) {
+janus_request *janus_request_new(janus_transport *transport, janus_transport_session *instance, void *request_id, gboolean admin, json_t *message, json_error_t *error) {
 	janus_request *request = g_malloc(sizeof(janus_request));
 	request->transport = transport;
 	request->instance = instance;
@@ -913,6 +918,13 @@ janus_request *janus_request_new(janus_transport *transport, janus_transport_ses
 	request->request_id = request_id;
 	request->admin = admin;
 	request->message = message;
+	if(error) {
+		request->error = (json_error_t*)g_malloc(sizeof(json_error_t))
+		*request->error = *error
+	}
+	else {
+		request->error = NULL;
+	}
 	g_atomic_int_set(&request->destroyed, 0);
 	janus_refcount_init(&request->ref, janus_request_free);
 	return request;
@@ -1021,9 +1033,18 @@ int janus_process_incoming_request(janus_request *request) {
 		JANUS_LOG(LOG_ERR, "Missing request or payload to process, giving up...\n");
 		return ret;
 	}
-	int error_code = 0;
-	char error_cause[100];
 	json_t *root = request->message;
+	if(root == NULL) {
+		json_error_t *error = request->error;
+		if(error != NULL) {
+			ret = janus_process_error(request, 0, NULL, JANUS_ERROR_INVALID_JSON,
+				"Invalid Janus API request - %s(%d,%d): %s", error->source, error->line, error->column, error->text);
+		}
+		else{
+			ret = janus_process_error_string(request, 0, NULL, JANUS_ERROR_INVALID_JSON, (char *)"Invalid Janus API request");
+		}
+		return ret;
+	}
 	/* Ok, let's start with the ids */
 	guint64 session_id = 0, handle_id = 0;
 	json_t *s = json_object_get(root, "session_id");
@@ -1040,6 +1061,8 @@ int janus_process_incoming_request(janus_request *request) {
 	janus_session *session = NULL;
 	janus_ice_handle *handle = NULL;
 
+	int error_code = 0;
+	char error_cause[100];
 	/* Get transaction and message request */
 	JANUS_VALIDATE_JSON_OBJECT(root, incoming_request_parameters,
 		error_code, error_cause, FALSE,
@@ -1102,7 +1125,7 @@ int janus_process_incoming_request(janus_request *request) {
 		/* We increase the counter as this request is using the session */
 		janus_refcount_increase(&session->ref);
 		/* Take note of the request source that originated this session (HTTP, WebSockets, RabbitMQ?) */
-		session->source = janus_request_new(request->transport, request->instance, NULL, FALSE, NULL);
+		session->source = janus_request_new(request->transport, request->instance, NULL, FALSE, NULL, NULL);
 		/* Notify the source that a new session has been created */
 		request->transport->session_created(request->instance, session->session_id);
 		/* Notify event handlers */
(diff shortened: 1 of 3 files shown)


januscla:
Thanks for your contribution, @jwittner! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

src/janus.c
```diff
@@ -3380,8 +3380,16 @@ void janus_transport_incoming_request(janus_transport *plugin, janus_transport_s
 	JANUS_LOG(LOG_VERB, "Got %s API request from %s (%p)\n", admin ? "an admin" : "a Janus", plugin->get_package(), transport);
 	/* Create a janus_request instance to handle the request */
 	janus_request *request = janus_request_new(plugin, transport, request_id, admin, message);
-	/* Enqueue the request, the thread will pick it up */
-	g_async_queue_push(requests, request);
+	if(message == NULL) {
+		/* Report details about the parsing error */
+		janus_process_error(request, /*session_id=*/0, /*transaction=*/NULL, JANUS_ERROR_INVALID_JSON,
```
jwittner, line 3385: Please do let me know if there are any issues with reporting results from the websocket messaging thread for this error case. We could cache the error information instead, if necessary.

jwittner:
I was able to get echo.py to mimic my doubled up messages and test these changes. My message with two requests (full transactions each) both succeeded. I tested the error path and see good logs about the json formatting reported.

src/janus.c
lminiero: I'd rather put the check (and response) where I added it in my PR. You're only taking into account the WebSocket transport since that's what you use, but we have many different transports, each with their own threading model, and more may be added in the future. Adding it where I added it would ensure consistency with how the core usually takes care of responses.

As a side note, please don't add those inline comments for session ID and transaction.

src/transports/janus_websockets.c
```diff
@@ -1287,35 +1287,49 @@ static int janus_websockets_common_callback(
...
 				JANUS_LOG(LOG_HUGE, "[%s-%p] Waiting for more fragments\n", log_prefix, wsi);
 				return 0;
 			}
-			JANUS_LOG(LOG_HUGE, "[%s-%p] Done, parsing message: %zu bytes\n", log_prefix, wsi, strlen(ws_client->incoming));
+			JANUS_LOG(LOG_HUGE, "[%s-%p] Done, parsing message: %zu bytes\n", log_prefix, wsi, incoming_length);
 			/* If we got here, the message is complete: parse the JSON payload */
-			json_error_t error;
-			json_t *root = json_loads(ws_client->incoming, 0, &error);
+			const char* request_curr = ws_client->incoming;
+			const char* request_end = ws_client->incoming + incoming_length;
```
lminiero, line 1317: Code style nit: our pointers have the `*` next to the variable, not the type.

src/transports/janus_websockets.c
```diff
@@ -1287,35 +1287,49 @@ static int janus_websockets_common_callback(
...
+			do { /* load all json objects from the message */
```
lminiero, line 1318: Code style nit: please put the comment above the `do`, and capitilize the first letter. We also use a capitlized version of JSON pretty much everywhere (comments, messages, documentation), so please ensure that's done in your patch too.

src/transports/janus_websockets.c
```diff
@@ -1287,35 +1287,49 @@ static int janus_websockets_common_callback(
...
+				json_error_t error;
+				json_t *root = json_loads(request_curr, JSON_DISABLE_EOF_CHECK, &error);
```
lminiero, line 1320: This is a cool flag I was unaware of, but I'm wondering if it may cause issues in some cases, e.g., with malformed messages (think of one with two closing brackets, for instance). This may cause the decoder to accept part of the input as valid when it really isn't, where right now we have some validation on incoming messages. I also see this is only available starting from version 2.1 of Jansson, so in case this is accepted it means we'll need to explicitly add this requirement to `configure.ac` as well, or it will break on instances using older versions of the library (I'm not sure how common those are).

lminiero (commented):

src/transports/janus_websockets.c
jwittner: Updated!

src/transports/janus_websockets.c
jwittner: Updated!

src/janus.c
jwittner: Moved the handling into the request management thread.

src/transports/janus_websockets.c
jwittner: Looks like configure.ac already specifies 2.5 for Jansson - so no work on that front yeah?

jwittner: I see what you mean. I've pushed updated code that parses all the messages before processing any and fails in entirety if json is not well formed until EOF. I'm gonna pull it into my fork for testing, so I expect there might be another update to this.

jwittner: With the udpates, with two closing brackes I now get this error:

```
Returning Janus API error 454 (Invalid Janus API request - <string>(1,1): '[' or '{' expected near '}')
```

and the message before the extra closing is ignored.

jwittner: Testing from my fork this is working well - I'm seeing multiple requests handled and errors for all if there's any parsing problems in the set.

jwittner:
All the feedback should be addressed now. Please re-review @lminiero when you can - thanks!

src/janus.c
```diff
@@ -899,20 +899,32 @@ static void janus_request_free(const janus_refcount *request_ref) {
...
 	janus_request *request = g_malloc(sizeof(janus_request));
 	request->transport = transport;
 	request->instance = instance;
 	janus_refcount_increase(&instance->ref);
 	request->request_id = request_id;
 	request->admin = admin;
 	request->message = message;
+	if(error) {
+		request->error = (json_error_t*)g_malloc(sizeof(json_error_t));
+		*request->error = *error;
```
lminiero, line 923: I don't understand this. You're first allocating memory for `request->error`, but then just have the pointer point to an existing memory area?

src/janus.c
```diff
@@ -899,20 +899,32 @@ static void janus_request_free(const janus_refcount *request_ref) {
...
+	}
+	else {
```
lminiero, line 925: Code style nit: else should be on same line as closing bracket.

src/janus.c
```diff
@@ -1021,9 +1033,18 @@ int janus_process_incoming_request(janus_request *request) {
...
-	int error_code = 0;
-	char error_cause[100];
 	json_t *root = request->message;
+	if(root == NULL) {
+		json_error_t *error = request->error;
+		if(error != NULL) {
+			ret = janus_process_error(request, 0, NULL, JANUS_ERROR_INVALID_JSON,
+				"Invalid Janus API request - %s(%d,%d): %s", error->source, error->line, error->column, error->text);
+		}
+		else{
```
lminiero, line 1043: Same here.

src/transports/janus_websockets.c
```diff
@@ -1287,35 +1287,77 @@ static int janus_websockets_common_callback(
...
+						json_t **msg_end = message_buffer + message_buffer_count;
+						while(msg != msg_end) {
+							/* Notify the core, no error since we know there weren't any */
+							gateway->incoming_request(&janus_websockets_transport, ws_client->ts, NULL, admin, *msg++, NULL);
+						}
+						/* Notify the core, no error since we know there weren't any */
+						gateway->incoming_request(&janus_websockets_transport, ws_client->ts, NULL, admin, message, NULL);
+						break;
+					}
+					else {
```
lminiero, line 1340: Same here.

src/transports/janus_websockets.c
```diff
@@ -1287,35 +1287,77 @@ static int janus_websockets_common_callback(
...
+						/* Buffer the message */						
+						message_buffer = (json_t**)g_realloc(message_buffer, sizeof(json_t*) * (message_buffer_count + 1));
+						message_buffer[message_buffer_count++] = message;
+					}
+				}
+				else {
```
lminiero, line 1346: Same here.

lminiero (commented):
Added a few more notes :+1:

src/janus.c
tmatth: I think the intent is to initialize the struct pointed to by `request->error` using the contents of `error` (e.g., *dst = *src)...I have some vague memory that this isn't safe on all platforms (maybe depending on padding/alignment?) but I can't find a definitive reference to support that:

Maybe not the most rigorous test but in this example `memcpy` and `dereferencing` lead to the same asm: 

https://godbolt.org/z/5x69qsqaG

src/janus.c
jwittner: Sorry for missing that styling, thanks for the patience and diligence. =) Honestly the style in this repo is growing on me. =)

src/janus.c
jwittner: > I think the intent is to initialize the struct pointed to by `request->error` using the contents of `error` (e.g., *dst = *src).

Exactly, I'm copying the struct from the incoming error struct (on the stack in this case) to the instance store here (on the heap as it may or may not be set). 

I'd be really surprised if there's a platform problem with assigning one struct to another. I verified that the error struct uses fixed sized arrays so the copy should be, as you point out, equivalent to a memcpy. =)

Open to any changes desired here of course.

jwittner: @lminiero - maybe you missed the deference of `request->error` on that line?

lminiero: I had seen it, but not knowing what `json_error_t` is I just assumed it was a string, rather than a struct. Maybe an explicit memcpy is a safer approach, even though I think we use the struct copy assignment ourselves around Janus (e.g., for RTP extensions) so if that's broken somewhere, we're in trouble anyway :hand_over_mouth:

tmatth: Yeah I think you can disregard what I said, structure copy by assignment is fine (if not better since the compiler can make smarter optimizations re: padding) in C90 (which Janus already requires IIRC), I think my memory of having to use `memcpy` was for some niche case.

jwittner:
Alright, all feedback should be addressed. @lminiero / @tmatth  - let me know if there's anything else needed to get this merged. And thanks for all the help!

src/janus.c
```diff
@@ -899,20 +899,31 @@ static void janus_request_free(const janus_refcount *request_ref) {
...
-	if(request->message)
+	if(request->message) {
 		json_decref(request->message);
-	request->message = NULL;
+		request->message = NULL;
+	}
+	if(request->error) {
+		g_free(request->error);
+		request->error = NULL;
+	}
```
tmatth, line 909: minor nit: you don't need to check if `request->error` is NULL or not, `g_free(NULL)` is totally safe (like `free(NULL)`).

jwittner: I was leaning toward the `error` being `NULL` most of the time so avoiding the call and the assignment in that case, but having a double test (mine and the one in `g_free`) when there is an error. If I drop my check then I get a test (in `g_free`) and an assignment all the time.

Happy to change though if we likely the cleanliness of free/assign all the time. =)

jwittner:
Can either of you approve the workflow so I can feel certain about the build/tests?

src/janus.c
tmatth: I was going to add that you don't need to set `request->error` to `NULL` since `request` is freed at the end of this function, but then I realized the existing code is also setting members to NULL (redundantly AFAICT) but dropping those is outside of the scope of this PR, so I can't fault you for following the existing convention.

In any case, I haven't tested but LGTM on reading!

jwittner:
Alright, all comments resolved and tests are passing. Let me know if there's anything else.

Also, once merged, should I move to 0.x or will y'all be doing that?

lminiero:
> Also, once merged, should I move to 0.x or will y'all be doing that?

Looks like the patch should be easy enough to apply cleanly to `0.x` too, so I can do it myself after this is merged.

jwittner:
Great! Anything holding this back? Not sure what the flow is for the project on merging PRs. =)

lminiero:
I'll need time to review and test, so please bear with me 🤭 I'm also busy on a few other activities so I may need a couple of days.

jwittner:
> I'll need time to review and test, so please bear with me 🤭 I'm also busy on a few other activities so I may need a couple of days.

No problem at all! Thanks again for all the help. If there are any tests I can write to help out, do let me know!

lminiero:
Looks good to me! I'll merge and backport it to `0.x` now :v: 
Thanks for your contribution and your patience!

PR #3135, opened 2022-12-30
Cleanup some g_free usage

tmatth:
Namely:
- simplify `if (x) g_free(x)` to just `g_free(x)` (this was the main motivation for this PR).
- use `char *` instead of `const char *` for values that will be freed to avoid having to cast away constness
- remove some instances of `if (x) { g_free(x); x = NULL }` where `x` goes out of scope.


tmatth:
> (this was the main motivation for this PR).

The underlying motivation being about making the code more compact/readable, in terms of operations I expect the difference to be trivial best.

lminiero:
Makes sense, thanks: merging!

PR #3136, opened 2023-01-02
Don't allow mid changes for m-line during renegotiations

lminiero:
We noticed that there's some stacks (possibly GStreamer?) that sometimes end up changing the mid of an m-line during a renegotiation. This sounded "wrong" to me, and asking around it indeed looks like JSEP forbids this, and browsers never do it too. Considering this can cause issues (race conditions come to mind), I decided to disallow this in Janus too: this patch implements this behaviour by simply showing a warning on the console when it happens, and keeping the previous mid in place. I don't expect this to cause problems, but in case you're talking to Janus from non-libwebrtc stacks you may want to let us know if you notice any regression and broken behaviours before I merge.


lminiero:
Merging.

PR #3137, opened 2023-01-02
[0.x] Parse media SSRC in RTCP reports when remote SSRCs are unknown

atoppi:
While making some tests between pion and Janus we found out that in some circumstances Janus could ignore incoming RTCP due to unrecognized SSRCs.

That happens when Janus receives RTCP **Receiver Reports** in connections where sources have unknown remote SSRCs (e.g. _recvonly_ connections). Pion picks a random SSRC as _sender_ SSRC and puts the _remote media_ SSRC in the _media_ header of the packet.

When Janus receives a RTCP packets from sources with unknown remote SSRCs, it parses the SSRC through `janus_rtcp_get_sender_ssrc`, that basically reads the _sender_ SSRC (in this case a random one, hence dropping the packet).

In this PR we try to add a fallback mechanism to read also from _media_ SSRC through `janus_rtcp_get_receiver_ssrc` in case the SSRC read from `janus_rtcp_get_sender_ssrc` is not recognized.

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -3067,23 +3067,45 @@ static void janus_ice_cb_nice_recv(NiceAgent *agent, guint stream_id, guint comp
 					} else {
 						/* Check the remote SSRC, compare it to what we have: in case
 							* we're simulcasting, let's compare to the other SSRCs too */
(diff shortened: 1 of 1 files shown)


lminiero:
Merging :+1:

PR #3143, opened 2023-01-05
Close mountpoint sockets when leaving relay thread (fixes #3111)

### LINKED ISSUE #3111: [1.x]  Streaming Plugin Mountpoint Fast Teardown/Setup Triggers "address already in use" Error
Hello everyone,

we experienced this on Janus 1.0.4 on Debian 11.5 aarch64, Linux 5.10.0 -- but I don't see why it shouln't be the same for 1.1.0 and current HEAD, since relevant code parts haven't changed I think. Older versions are (I guess) the same, but we aren't using the streaming plugin that long so we can't say.

We recognized a problem when in fast succession tearing down and creating streaming plugin mountpoints (specifying the same audioport) again. The mountpoint wouldn't be created and the following error(s) logged:

```
[ERR] [plugins/janus_streaming.c:janus_streaming_create_fd:6665] [(null)] Bind failed for audio (port 10358)... 98 (Address already in use)
[ERR] [plugins/janus_streaming.c:janus_streaming_create_rtp_source_stream:6855] [(null)] Can't bind to port 10358...
[ERR] [plugins/janus_streaming.c:janus_streaming_process_synchronous_request:3451] Can't add 'rtp' stream '(null)', error creating data source stream...
```

The attached patch fixes this problem for us and I think there should be no situations in which using it would have negative consequences. (I tested it on 1.0.4 and 1.1.0)
If you see fit, maybe you can incorporate the patch.

All the best, Chris

[janus-streaming-sock_reuse.patch.txt](https://github.com/meetecho/janus-gateway/files/10112372/janus-streaming-sock_reuse.patch.txt)

lminiero:
This PR closes the sockets used by a mountpoint when leaving the related thread: since a "destroy" request joins the thread, this ensures ports are free once a response is sent. Before, sockets would only be closed in the mountpoint free callback, so when reference counters gave the green light, which could take longer and cause the issue described in #3111.

### DIFF
--- a/src/plugins/janus_streaming.c
+++ b/src/plugins/janus_streaming.c
@@ -9316,6 +9316,25 @@ static void *janus_streaming_relay_thread(void *data) {
 		}
 	}
 
+	/* Close the ports we bound to */
+	temp = source->media;
+	while(temp) {
+		janus_streaming_rtp_source_stream *stream = (janus_streaming_rtp_source_stream *)temp->data;
+		if(stream->fd[0] > -1)
+			close(stream->fd[0]);
+		stream->fd[0] = -1;
+		if(stream->fd[1] > -1)
+			close(stream->fd[1]);
+		stream->fd[1] = -1;
+		if(stream->fd[2] > -1)
+			close(stream->fd[2]);
+		stream->fd[2] = -1;
+		if(stream->rtcp_fd > -1)
+			close(stream->rtcp_fd);
+		stream->rtcp_fd = -1;
+		temp = temp->next;
+	}
+
 	/* Notify users this mountpoint is done */
 	janus_mutex_lock(&mountpoint->mutex);
 	GList *viewer = g_list_first(mountpoint->viewers);


src/plugins/janus_streaming.c
```diff
@@ -9316,6 +9316,25 @@ static void *janus_streaming_relay_thread(void *data) {
...
+		janus_streaming_rtp_source_stream *stream = (janus_streaming_rtp_source_stream *)temp->data;
+		if(stream->fd[0] > -1)
+			close(stream->fd[0]);
+		stream->fd[0] = -1;
+		if(stream->fd[1] > -1)
+			close(stream->fd[1]);
+		stream->fd[1] = -1;
+		if(stream->fd[2] > -1)
+			close(stream->fd[2]);
+		stream->fd[2] = -1;
```
atoppi, line 9331: This looks a bit redundant, why not use a simple loop?
```c
	for(int i=0; i<3; i++) {
  		if(stream->fd[i] >= 0) {
			close(stream->fd[i]);
  			stream->fd[i] = -1;
		}
	}
```

lminiero: Since i was small enough and known, it felt easier to do that. Besides, I just copied what we had in the _free callback already.

lminiero:
Merging.

PR #3145, opened 2023-01-10
janus_streaming: unlock before decref of mounpoint

tmatth:
The mutex should not be held while decrementing the reference count. In the unlikely event that the refcount were 0, janus_streaming_mountpoint_free would deadlock trying to acquire this same mutex.

This is again unlikely since the reference count is incremented right before and so presumably is never 0 here.


tmatth:
This addresses the issue discussed here https://github.com/meetecho/janus-gateway/issues/3108#issuecomment-1377419887

lminiero:
Ack, thanks, merging!

PR #3146, opened 2023-01-11
Convert label_length to host byte order before use

uxmaster:
Just adding missing `ntohs`.

BTW. We got plenty of `channel %d is in state 1 instead of CLOSED`. State 1 is `DATA_CHANNEL_CONNECTING` and it seems everything works despite the message. Maybe it should be treated as a warning instead of an error.


lminiero:
Thanks, good catch on the missing `ntohs`! I'll apply the same fix on master.

PR #3150, opened 2023-01-13
Add reason of track being added/removed in onremotetrack in janus.js

marekpiechut:
Added a metadata object to janus.js `onremotetrack` callbacks with reason why track is being added/removed. Thanks to that users can decide if they want to handle the call or ignore it.

It can be used to fix blinking screen-sharing sessions as discussed here: https://groups.google.com/g/meetecho-janus/c/bZaQA4MqJcs
(user is being notified about track removal, when it's only a `mute` call)

After the change `onremotetrack` call will lok like this:
```
 onremotetrack: function(track, mid, added, metadata) {
            // A remote track (working PeerConnection!) with a specific mid has just been added or removed
            // You can query metadata to get some more information on why track was added or removed
            // metadata fields:
            //   - reason: 'created' | 'ended' | 'mute' | 'unmute'
        },
```

Change should be backwards compatible, as it's only adding a new parameter to callbacks that will be ignored by current client code.

I have tested demo pages with the change and it looks like everything is fine.
Also updated documentation.

### DIFF
--- a/html/audiobridgetest.js
+++ b/html/audiobridgetest.js
@@ -298,8 +298,12 @@ $(document).ready(function() {
 									$('#room').removeClass('hide').show();
 									$('#participant').removeClass('hide').html(myusername).show();
 								},
-								onremotetrack: function(track, mid, on) {
-									Janus.debug("Remote track (mid=" + mid + ") " + (on ? "added" : "removed") + ":", track);
+								onremotetrack: function(track, mid, on, metadata) {
+									Janus.debug(
+										"Remote track (mid=" + mid + ") " +
+										(on ? "added" : "removed") +
+										(metadata ? " (" + metadata.reason + ") ": "") + ":", track
+									);
 									if(remoteStream || track.kind !== "audio")
 										return;
 									if(!on) {
--- a/html/audiobridgetest.js
+++ b/html/audiobridgetest.js
@@ -302,7 +302,7 @@ $(document).ready(function() {
 									Janus.debug(
 										"Remote track (mid=" + mid + ") " +
 										(on ? "added" : "removed") +
-										(metadata ? " (" + metadata.reason + ") ": "") + ":", track
+										(metadata ? " (" + metadata.reason + ") " : "") + ":", track
 									);
 									if(remoteStream || track.kind !== "audio")
 										return;
--- a/html/canvas.js
+++ b/html/canvas.js
@@ -169,8 +169,12 @@ $(document).ready(function() {
 										});
 									}
 								},
-								onremotetrack: function(track, mid, on) {
-									Janus.debug("Remote track (mid=" + mid + ") " + (on ? "added" : "removed") + ":", track);
+								onremotetrack: function(track, mid, on, metadata) {
+									Janus.debug(
+										"Remote track (mid=" + mid + ") " +
+										(on ? "added" : "removed") +
+										(metadata? " (" + metadata.reason + ") ": "") + ":", track
+									);
 									if(!on) {
 										// Track removed, get rid of the stream and the rendering
 										$('#peervideo' + mid).remove();
--- a/html/canvas.js
+++ b/html/canvas.js
@@ -173,7 +173,7 @@ $(document).ready(function() {
 									Janus.debug(
 										"Remote track (mid=" + mid + ") " +
 										(on ? "added" : "removed") +
-										(metadata? " (" + metadata.reason + ") ": "") + ":", track
+										(metadata? " (" + metadata.reason + ") " : "") + ":", track
 									);
 									if(!on) {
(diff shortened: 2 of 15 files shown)


januscla:
Thanks for your contribution, @marekpiechut! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks! Not sure I like `reason` as a name for the property (maybe `status`?) but that's semantics. In principle I think the feature makes sense.

marekpiechut:
Thanks for the review @lminiero. On Monday I'll go through the demos and update signature.
Sorry that it takes so long, but it's been a busy week.

lminiero:
> Sorry that it takes so long, but it's been a busy week.

No problem, take your time!

marekpiechut:
I've added metadata to all demos `onremotetrack` callbacks. Most of them just display it in debug (had no good idea to actually showcase it).

For screen-sharing demo I've added code to ignore mute/unmute events with description why it might be useful (just like in my issue with blinking screen-sharing streams).

I've also found that in `vp9svctest.js` was using undeclared `doSimulcast` variable. This was crashing the demo. I've removed it and just disabled simulcast for this demo.

Let me know if something is still missing.

html/audiobridgetest.js
```diff
@@ -298,8 +298,12 @@ $(document).ready(function() {
 									$('#room').removeClass('hide').show();
 									$('#participant').removeClass('hide').html(myusername).show();
 								},
-								onremotetrack: function(track, mid, on) {
-									Janus.debug("Remote track (mid=" + mid + ") " + (on ? "added" : "removed") + ":", track);
+								onremotetrack: function(track, mid, on, metadata) {
+									Janus.debug(
+										"Remote track (mid=" + mid + ") " +
+										(on ? "added" : "removed") +
+										(metadata ? " (" + metadata.reason + ") ": "") + ":", track
```
lminiero, line 305: Just a code style nit: please put an extra space before the `:` that belongs to the ternary operator, as it's not clear otherwise. The same nit applies to a few other instances I see you've added.

html/canvas.js
```diff
@@ -169,8 +169,12 @@ $(document).ready(function() {
 										});
 									}
 								},
-								onremotetrack: function(track, mid, on) {
-									Janus.debug("Remote track (mid=" + mid + ") " + (on ? "added" : "removed") + ":", track);
+								onremotetrack: function(track, mid, on, metadata) {
+									Janus.debug(
+										"Remote track (mid=" + mid + ") " +
+										(on ? "added" : "removed") +
+										(metadata? " (" + metadata.reason + ") ": "") + ":", track
```
lminiero, line 176: Same here but for the question mark, that should be separated from `metadata` by an extra space.

html/screensharingtest.js
```diff
@@ -511,8 +511,23 @@ function newRemoteFeed(id, display) {
...
+					(metadata? " (" + metadata.reason + ") ": "") + ":", track
+				);
+				// Screen sharing tracks are sometimes muted/unmuted by browser
+				// when data is not flowing fast enough, this can make streams blink,
+				// we can ignore these
+				if(
+					track.kind === "video"
+					&& metadata
+					&& (metadata.reason === "mute" || metadata.reason === "unmute")
+				) {
```
lminiero, line 527: Code style: please put the conditions on the same line as the `if`. In case it ends up being too long, we do a carriage return with two tabs, so that it's "more" indented than the action.

html/screensharingtest.js
```diff
@@ -511,8 +511,23 @@ function newRemoteFeed(id, display) {
...
-			onremotetrack: function(track, mid, on) {
-				Janus.debug("Remote track (mid=" + mid + ") " + (on ? "added" : "removed") + ":", track);
+			onremotetrack: function(track, mid, on, metadata) {
+				Janus.debug(
+					"Remote track (mid=" + mid + ") " +
+					(on ? "added" : "removed") +
+					(metadata? " (" + metadata.reason + ") ": "") + ":", track
+				);
+				// Screen sharing tracks are sometimes muted/unmuted by browser
+				// when data is not flowing fast enough, this can make streams blink,
```
lminiero, line 521: I'd use a semicolon rather than a comma after `enough`, to split the sentence in preamble and explaination.

lminiero (commented):
Thanks! I added a few small notes.

marekpiechut:
@lminiero I think I have fixed all the code style issues. I got too used to auto-formatters :)

lminiero:
Thanks! Abroad for FOSDEM but I'll have a look as soon as I get back on Monday :muscle:

lminiero:
Thanks, merging!

PR #3157, opened 2023-01-31
Handling an abnormally close of the websocket in janus.js

simi1401:
Added functionality to try the next server if the websocket connection is unexpectedly terminated. Reconnection will only be done when the websocket is closed with closecode 1006. This status code is returned, for example, when the network connection, a reverse proxy or the Janus gateway itself fails. This behaviour was tested with Chrome, Edge and Firefox.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1024,12 +1024,30 @@ function Janus(gatewayCallbacks) {
 					handleEvent(JSON.parse(event.data));
 				},
 
-				'close': function() {
+				'close': function(event) {
+					Janus.debug(event)
 					if(!server || !connected) {
 						return;
 					}
 					connected = false;
-					// FIXME What if this is called when the page is closed?
+					// 1006 is a status code to indicate that the connection was closed abnormally,
+					// so we have to try the other servers
+					if (event.code === 1006) {
+						if (Janus.isArray(servers) && !callbacks["reconnect"]) {
+							serversIndex++;
+							if (serversIndex === servers.length) {
+								// We tried all the servers the user gave us, and they all failed
+								callbacks.error("Error connecting to any of the provided Janus servers: Is the server down?");
+								return;
+							}
+							// Let's try the next server
+							server = null;
+							setTimeout(function () {
+								createSession(callbacks);
+							}, 200);
+							return;
+						}
+					}
 					gatewayCallbacks.error("Lost connection to the server (is it down?)");
 				}
 			};
@@ -1067,7 +1085,7 @@ function Janus(gatewayCallbacks) {
 				if(Janus.isArray(servers) && !callbacks["reconnect"]) {
 					serversIndex++;
 					if(serversIndex === servers.length) {
-						// We tried all the servers the user gave us and they all failed
+						// We tried all the servers the user gave us, and they all failed
 						callbacks.error("Error connecting to any of the provided Janus servers: Is the server down?");
 						return;
 					}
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @simi1401! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Mh I'm not sure this works. I can see the advantage of doing that when you're first trying to connect, but what if you lose connectivity while you've been in a session for a while? In that case the application may want to have a say in what to do, e.g., reconnect to the same server and reclaim the session. With this patch you automatically reconnect somewhere else with this patch, which I'm not convinced about.

tmatth:
> Mh I'm not sure this works. I can see the advantage of doing that when you're first trying to connect, but what if you lose connectivity while you've been in a session for a while? In that case the application may want to have a say in what to do, e.g., reconnect to the same server and reclaim the session. With this patch you automatically reconnect somewhere else from scratch, which I'm not convinced about.

Maybe this behaviour should be behind a default-false option (like `autoTryOtherServers` but catchier)

simi1401:
A flag is probably useful, but I think the flag should also be used with the http API. In the error method of the httpAPICall method, the next server will also be attempted.

What do you think?

lminiero:
I don't have a strong opinion, especially considering I personally never used the feature, so I guess this is mostly up to who instead does. My only note was about that possibly unexpected behaviour depending on the state, but for what I know it could be what who uses the feature would indeed expect.

html/janus.js
```diff
@@ -613,6 +613,10 @@ function Janus(gatewayCallbacks) {
 	this.destroyOnUnload = true;
 	if(typeof gatewayCallbacks.destroyOnUnload !== 'undefined' && gatewayCallbacks.destroyOnUnload !== null)
 		this.destroyOnUnload = (gatewayCallbacks.destroyOnUnload === true);
+	// Whether we should try the other servers when the websocket closed abnormally or the HttpApiCall fails
+	let autoTryOtherServers  = true;
+	if(typeof gatewayCallbacks.autoTryOtherServers  !== 'undefined' && gatewayCallbacks.autoTryOtherServers  !== null)
+		autoTryOtherServers  = (gatewayCallbacks.autoTryOtherServers  === true);
```
lminiero, line 619: Small nit: I see some double spaces in the lines above, where just one would suffice.

html/janus.js
```diff
@@ -1064,10 +1086,10 @@ function Janus(gatewayCallbacks) {
 			},
 			error: function(textStatus, errorThrown) {
 				Janus.error(textStatus + ":", errorThrown);	// FIXME
-				if(Janus.isArray(servers) && !callbacks["reconnect"]) {
+				if(autoTryOtherServers && Janus.isArray(servers) && !callbacks["reconnect"]) {
```
lminiero, line 1089: I think this is missing some condition here: my point in my previous comment was that we do want an automated fallback to other servers if we fail to connect entirely (that's the purpose of servers), but that should not happen if we managed to connect to a server, did something, and then lost connection. It looks like this new `autoTryOtherServers` property is conditionally changing the behaviour of the first case, and not dealing with the second one.

In both cases, it may be enough to check what the value of `connected` is, since we set it to `true` the moment we successfully connected to a server: if it's `true`, don't fallback (or maybe let's envisage a new property to allow a fallback for that specific case), while if it's false, do fallback (or see what `autoTryOtherServers` says).

lminiero (commented):
I added a couple of notes, since I don't think this new patch captures my original comment. Thanks!

lminiero:
@simi1401 is there any update on this? Thanks!

PR #3159, opened 2023-02-01
Use getaddrinfo instead of gethostbyname (fixes #3156)

### LINKED ISSUE #3156: [1.x] Replace gethostbyname with getaddrinfo
Commit 963c4b6, current head of repository

No - issue affects future releases

**Was this working before?**
Not applicable

Not applicable

**Additional context**
When building for inclusion in Fedora, warning to replace `gethostbyname` with `getaddrinfo` is raised.  This will allow use of IPv6 addresses. `gethostbyname` appears at:
- https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_nosip.c#L2310
- https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_nosip.c#L2321
-  https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_sip.c#L6878
- https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_sip.c#L6889
- https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_audiobridge.c#L6570

`getaddrinfo` is used at:
- https://github.com/meetecho/janus-gateway/blob/master/src/turnrest.c#L268
- https://github.com/meetecho/janus-gateway/blob/master/src/events/janus_gelfevh.c#L162
- https://github.com/meetecho/janus-gateway/blob/master/src/ice.c#L1183
- https://github.com/meetecho/janus-gateway/blob/master/src/ice.c#L1241
- https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_audiobridge.c#L2257
- https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_audiobridge.c#L4708

Can make a pull request if of interest.

lminiero:
As the title says, this patch gets rid of the few places where we were still using `gethostbyname` (AudioBridge, SIP and NoSIP plugins) in order to use `getaddrinfo` instead, which we were already using in other places. This addresses issue #3156 which is where this was pointed out.

That said, this patch may have some implications that is worth investigating. In fact, the main point of `getaddrinfo` is that it makes it easier to use IPv6 as well, rather than stick to IPv4 which is what those plugins were doing, but this means we now use IPv6 sockets whether it will actually be v6 or not (so with `IPV6_V6ONLY` set to `0`, via `setsockopt`). This is something we did with forwarders too, but in the past we figured out this could actually be problematic, e.g., on MacOS or systems that have IPv6 entirely disabled. As such, I'll have to probably add some tweaks to the current patch so that we create the socket with a specific family, maybe depending on the `local_ip` the plugins are using (still not sure).

Apart from this, simple functional tests (using IPv4 addresses) seemed to still work fine in both SIP and NoSIP plugins (I haven't tested plain RTP participation in the AudioBridge yet with this changes). If you use any of those plugins for RTP usage, you're strongly encouraged to test this PR, so that you're not caught by surprise if any regression we didn't anticipate shows up in your deployments.

### DIFF
--- a/src/ip-utils.c
+++ b/src/ip-utils.c
@@ -321,3 +321,49 @@ char *janus_network_detect_local_ip_as_string(janus_network_query_options addr_t
 		return NULL;
 	return g_strdup(janus_network_address_string_from_buffer(&buf));
 }
+
+int janus_network_resolve_address(const char *host, struct sockaddr_storage *address) {
+	if(!host || !address)
+		return -EINVAL;
+	/* Check whether we need to resolve the address*/
+	gboolean resolved = FALSE;
+	if(strstr(host, ":")) {
+		struct sockaddr_in6 *addr6 = (struct sockaddr_in6 *)address;
+		addr6->sin6_family = AF_INET6;
+		if(inet_pton(AF_INET6, host, &addr6->sin6_addr) == 1) {
+			/* Numeric IPv6 address */
+			resolved = TRUE;
+		}
+	} else {
+		struct sockaddr_in *addr = (struct sockaddr_in *)address;
+		addr->sin_family = AF_INET;
+		if(inet_pton(AF_INET, host, &addr->sin_addr) == 1) {
+			/* Numeric IPv4 address */
+			resolved = TRUE;
+		}
+	}
+	if(!resolved) {
+		/* Perform a getaddrinfo on the address */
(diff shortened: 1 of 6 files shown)


lminiero:
Since we received no feedback, I'll merge as it is. In case it turns out this causes problems in some environments, we'll address them later on.

PR #3160, opened 2023-02-08
Optionally support X-Forwarded-For in both HTTP and WebSocket transports (fixes #3158)

### LINKED ISSUE #3158: proxy protocol support on transport
some transport allow setting acl to limit access to resource (admin or api)
In environment where janus transport is behind proxy this acl is useless because janus take source address from only tcp socket and not http header

In a nutshell (for http and websocket transport) it should recover the source ip from [http header (if set)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)  and not only by data in [sockaddr](https://github.com/meetecho/janus-gateway/blob/11247a642ab1cb5d93b36e7e0d73d8837865997a/src/transports/janus_http.c#L1330)

In more general environment janus can support [proxy protocol](https://www.haproxy.com/blog/use-the-proxy-protocol-to-preserve-a-clients-ip-address/)

`What is the [Proxy Protocol](https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt)? It is a network protocol for preserving a client’s IP address when the client’s TCP connection passes through a proxy. Without such a mechanism, proxies lose this information because they act as a surrogate for the client, relaying messages to the server, but replacing the client’s IP address with their own. This distorts the logs of upstream servers because the logs incorrectly indicate that all traffic originated at the proxy.
[...]
The Proxy Protocol works by adding a header that contains the client’s IP address to the beginning of a TCP connection`

lminiero:
See #3158 for rationale. This patch allows both HTTP and WebSocket transports to optionally inspect the `X-Forwarded-For` header in requests to check whether or not a request is allowed by the transport's ACL rules. The support is optional and disabled by default to avoid abuses, since spoofing the header is trivial: as such, it should only be configured if a proxy is indeed in front of Janus, it's the only one able to talk to Janus, and there's reasons to filter clients contacting the proxy.

I haven't checked if the APIs I used require specific versions of MHD or libwebsockets. I'll wait for our Github Actions to see if the ones we use normally do support them. Feedback welcome, as usual.

### DIFF
--- a/conf/janus.transport.http.jcfg.sample
+++ b/conf/janus.transport.http.jcfg.sample
@@ -20,6 +20,8 @@ general: {
 	#secure_interface = "eth0"		# Whether we should bind this server to a specific interface only
 	#secure_ip = "192.168.0.1"		# Whether we should bind this server to a specific IP address (v4 or v6) only
 	#acl = "127.,192.168.0."		# Only allow requests coming from this comma separated list of addresses
+	#acl_forwarded = true			# Whether we should check the X-Forwarded-For header too for the ACL
+									# (default=false, since without a proxy in the middle this could be abused)
 	#mhd_connection_limit = 1020		# Open connections limit in libmicrohttpd (default=1020)
 	#mhd_debug = false					# Ask libmicrohttpd to write warning and error messages to stderr (default=false)
 }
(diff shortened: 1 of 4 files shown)


lminiero:
Adding the `multistream` label as that's what this PR is from, but I'll backport it to `0.x` once merged.

lminiero:
Merging.

PR #3162, opened 2023-02-09
Allow Lua and Duktape plugins to relay extensions when relaying RTP packets

lminiero:
Since we merged #1884, the Lua and Duktape plugins had basically no RTP extensions support, except for those originated by the core (e.g., mid). In fact, since #1884 required the plugins to explicitly use the `.extensions` property to carry on specific extensions, and Lua/Duktape didn't use it (simply resetting the struct), this resulted in extension properties passed by media senders to be stripped by the core when relaying packets to receivers. This patch addresses this shortcoming, and ensures we now copy the struct along properly. I've tested this briefly with the `audio-level` extension in both plugins, and it looks like it's working as expected.

Planning to merge soon, so if you have feedback please do say so.


lminiero:
Merging.

Issue #3154, opened 2023-01-24
[1.x] segmentation faults on videoroom hangup


spscream:
**What version of Janus is this happening on?**
1.1.0, 1.1.1

**Have you tested a more recent version of Janus too?**
problem occured first on 1.1.0 and we upgraded to 1.1.1 and it steal appears.

**Was this working before?**
this is occuring only on one of our customers servers and not occuring on others with the same version(1.1.0 and 1.1.1).

**Is there a gdb or libasan trace of the issue?**
https://gist.github.com/spscream/6ee8265dbc7d0c8f7fa24c2c9c0f8eef

**Additional context**
We use videoroom plugin mainly and this segmentation fault appers only on one customer. 
This customer also having some random issues with turn/stun servers and connectivity between janus and webrtc clients,
 so this could not be directly related to the crash, but maybe this leads to this segmentation fault.
 
 maybe this duplicates https://github.com/meetecho/janus-gateway/issues/3124

lminiero:
The current version of Janus master is 1.1.3, please try with that one too.

spscream:
tha same happened on master

atoppi:
@spscream thanks for checking, would you please share an ASan stack trace from the current master?

lminiero:
@spscream if you also have ways that we can use to consistently replicate it ourselves that would help too!

spscream:
sure, here is https://gist.github.com/spscream/b7ebe454303c3b044bc24f83cd8aea0e

spscream:
I don't know hot to make repetition of this, it happens only on one environment

spscream:
maybe this related to some buggy participant which calls unpublish twice in parallel, we are trying to figure it out

lminiero:
Since I suspect it's an issue with reference counting, where something is freed sooner than it should, one thing that might help is if you uncommented [this line](https://github.com/meetecho/janus-gateway/blob/master/src/refcount.h#L63) and recompiled Janus. It will make the logs explode in size, since it will be tracking all refs and unrefs, but should the problem occur again it would give us much more info to track where and why the affected resources was freed before its time. You may only want to do that in your own local deployment, maybe, rather than the one in production: your choice! But that would definitely help, especially if you're able to replicate it eventually.

spscream:
Hi!
we got some crashes and logs with refcount debug here:
[...]

Issue #3124, opened 2022-12-13
[1.x] Strange segmentation fault on hangup media


petarminchev:
Have upgraded to Multistream Janus 1.1.1 a week ago. Maybe it is just a coincidence and the issue was already in previous versions. But today have received one segmentation fault, which cannot fully understand -> libasan: https://pastebin.com/Bjj1XK4r

Line 3391 seems to be the **json_object_set_new** and error happened inside **json_string**:
			if(ps->description)
				json_object_set_new(m, "feed_description", json_string(ps->description));
				
We do not set any description for feeds, so it is strange it goes inside the **if** clause in the first place.				

What is your opinion on this? Is it possible Janus sets description erroneously to some internal value or frees the field twice?

Thank you in advance for the help!

lminiero:
The snippet you shared is not the libasan dump, but a gdb stacktrace. Do you have the actual libasan output instead? That should tell you why it crashed in there, e.g., if it was because the property had been freed already (and in case where).

lminiero:
Any update on this?

lminiero:
@petarminchev we have reasons to suspect #3154 is a duplicate of the issue you originally reported: we managed to gather more information from the logs they shared, and so the PR above attempts to provide a fix to that. Whether you can reproduce the problem or not (apparently it didn't happen to you anymore), I still encourage you to test that branch, since we want to ensure it fixes the problem and at the same time doesn't introduce any regression.

lminiero:
Closing as I haven't seen further reports here, and the other referenced issue is still open.

PR #3167, opened 2023-02-17
Fix for VideoRoom race condition (see #3124 and 3154)

lminiero:
This is a first attempt to fix a nasty (and rare) race condition that can occur in the VideoRoom, and that has been discovered thanks to logs provided in #3154: there's very good chances that #3124 is a duplicate of the same issue, which is why I'm grouping them together here.

The issue that happens is basically the following:

1. a new handle is attached to the VideoRoom;
2. a "join" as a subscriber is sent on that handle, which is queued for processing;
3. in the meanwhile, the handle is detached, so `destroy_session` and `hangup_media` are called, which do nothing since at that point the handle is still neither a publisher nor a subscriber;
4. in parallel, the "join" request is processed and goes through, adding the subscriber as a recipient for the specified publisher(s), but the session is now freed;
5. some time later, the publisher goes away, and when accessing the session for that subscriber, there's a crash since session has been freed already.

The problem is clearly the race condition between the "join" processing and the detach, which causes a `janus_videoroom_subscriber` instance to keep on living, even though its session has gone, and refering a session pointer that's now garbage. It's not something that can be easily fixed with just an additional reference, since that may solve the crash, but would leave a leak: in fact, the problem is that the subscriber instance is still in the publisher's list of recipients, and that subscriber can only be removed by a `destroy_session` and/or `hangup_media`, but those happened already and are not coming back (the session is over).

As such, the fix this patch attempts is using `session_mutex` on a "join" (for both publishers and subscribers) as a way to ensure the race condition doesn't happen. In fact, `session_mutex` is what we use to protect the sessions hashtable, which means it's used when a session is marked as destroyed: we already had checks in "join" for whether the session was destroyed or not, which could be bypassed in case of race conditions like the one above, and now should instead do their job properly.

Marking this PR as a draft now, so that those who reported the problem can test it, and ensure there's no regression. From a quick check I don't think this new mutex usage should cause any instance of inverse lock order problems, but some more testing should help verify no deadlock will occur now. The PR is also a draft just in case it turns out a different mutex may be a better option (e.g., `session->mutex`). As usual, feedback is more than welcome: I hope that, considering many people use the VideoRoom, we'll get more feedback than usual.

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -9152,6 +9152,7 @@ static void *janus_videoroom_handler(void *data) {
 			}
 			janus_refcount_increase(&videoroom->ref);
 			janus_mutex_unlock(&rooms_mutex);
+			janus_mutex_lock(&sessions_mutex);
 			janus_mutex_lock(&videoroom->mutex);
 			json_t *ptype = json_object_get(root, "ptype");
 			const char *ptype_text = json_string_value(ptype);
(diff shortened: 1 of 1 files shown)


lminiero:
While there still are some race conditions to address, they seem to be happening in different places for now. As such, I'll start merging these fixes as they help greatly already. We'll address the other problems #3154 spotted in a new PR.

PR #3168, opened 2023-02-17
Add option to manually insert SPS/PPS RTP packets for H.264 mountpoints (see #3149)

lminiero:
Multistream version of the feature added to `0.x` in #3149, so see the description there for context.
Not tested yet, will do that next week.


lminiero:
Had a chance to test different configurations, and it seems to be working as expected, so I'll remove the draft status. Will merge soon, so please provide feedback before that happens, if needed.

lminiero:
Merging.

PR #3170, opened 2023-02-20
Lint whole JS codebase and force tab indentation

atoppi:
As the title explains, Just fixed all the eslint errors/warnings and enforced the tab indent.

### DIFF
--- a/html/screensharingtest.js
+++ b/html/screensharingtest.js
@@ -298,7 +298,8 @@ $(document).ready(function() {
 										});
 									}
 								},
-								onremotetrack: function() {
+								// eslint-disable-next-line no-unused-vars
+								onremotetrack: function(track, mid, on) {
 									// The publisher stream is sendonly, we don't expect anything here
 								},
 								oncleanup: function() {
@@ -513,7 +514,8 @@ function newRemoteFeed(id, display) {
 						});
 				}
 			},
-			onlocaltrack: function() {
+			// eslint-disable-next-line no-unused-vars
+			onlocaltrack: function(track, on) {
 				// The subscriber stream is recvonly, we don't expect anything here
 			},
 			onremotetrack: function(track, mid, on, metadata) {
--- a/html/screensharingtest.js
+++ b/html/screensharingtest.js
@@ -1,3 +1,6 @@
+/* global iceServers:readable, Janus:readable, server:readable */
+/* global bootbox:readable, Spinner:readable */
+
 // We import the settings.js file to know which address we should contact
 // to talk to Janus, and optionally which STUN/TURN servers should be
 // used as well. Specifically, that file defines the "server" and
@@ -22,13 +25,13 @@ var spinner = null;
 
 // Just an helper to generate random usernames
 function randomString(len, charSet) {
-    charSet = charSet || 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
-    var randomString = '';
-    for (var i = 0; i < len; i++) {
-    	var randomPoz = Math.floor(Math.random() * charSet.length);
-    	randomString += charSet.substring(randomPoz,randomPoz+1);
-    }
-    return randomString;
+	charSet = charSet || 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+	var randomString = '';
+	for (var i = 0; i < len; i++) {
+		var randomPoz = Math.floor(Math.random() * charSet.length);
+		randomString += charSet.substring(randomPoz,randomPoz+1);
+	}
+	return randomString;
 }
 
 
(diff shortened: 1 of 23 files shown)


atoppi:
The rationale for this effort is:
1) Potentially automate checks in github-actions for all the js files (currently CI is linting only `janus.js`)
2) Clear out the output for IDEs with native eslint support, in order not to lose any warning/error from the linter

html/screensharingtest.js
```diff
@@ -22,13 +24,13 @@ var spinner = null;
 
 // Just an helper to generate random usernames
 function randomString(len, charSet) {
```
lminiero, line 26: I think we can get rid of this function entirely. since we have `Janus.randomString` already (which incidentally is used just a few lines above). Possibly unrelated to the other changes, but since we're cleaning up...

html/streamingtest.js
```diff
@@ -489,15 +491,6 @@ function addPanel(panelId, mid, desc) {
...
 
-// Helper to escape XML tags
-function escapeXmlTags(value) {
-	if(value) {
-		var escapedValue = value.replace(new RegExp('<', 'g'), '&lt');
-		escapedValue = escapedValue.replace(new RegExp('>', 'g'), '&gt');
-		return escapedValue;
-	}
-}
-
```
lminiero, line 500: Why has this function been removed? I remember it was added in #2817 to avoid issues of user provided descriptions (e.g., names, titles, etc.) that could end up being interpreted wrongly or maliciously when used in web pages. Looking at the original PR, we definitely used it in the Streaming demo, so if we're not anymore I want to check why (maybe we're using `.text()` instead of `.html()`?).

html/textroomtest.js
```diff
@@ -94,7 +96,7 @@ $(document).ready(function() {
 											});
 									}
 								},
-								ondataopen: function(data) {
+								ondataopen: function() {
```
lminiero, line 99: `ondataopen` is not a function without arguments:

	// Notify the open data channel
	pluginHandle.ondataopen(label, protocol);

The same comment applies to the other demos too, where I see the same change as been made. Maybe we were using a variable name that didn't make it clear, and ignoring the protocol part which was probably added later, but I think it's a mistake to not list them as attributes here, whether they're used or not.

html/textroomtest.js
```diff
@@ -153,16 +155,12 @@ $(document).ready(function() {
 										$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;
 									} else if(what === "leave") {
 										// Somebody left
-										var username = json["username"];
-										var when = new Date();
```
lminiero, line 157: Why has `username` been removed? This is a remote user that's gone, and we need that ID to remove the element as the following lines show.

html/textroomtest.js
```diff
@@ -153,16 +155,12 @@ $(document).ready(function() {
...
 										$('#rp' + username).remove();
 										$('#chatroom').append('<p style="color: green;">[' + getDateString() + '] <i>' + participants[username] + ' left</i></p>');
 										$('#chatroom').get(0).scrollTop = $('#chatroom').get(0).scrollHeight;
 										delete participants[username];
 									} else if(what === "kicked") {
 										// Somebody was kicked
-										var username = json["username"];
-										var when = new Date();
```
lminiero, line 165: See above.

html/textroomtest.js
```diff
@@ -362,13 +361,13 @@ function getDateString(jsonDate) {
 
 // Just an helper to generate random usernames
 function randomString(len, charSet) {
```
lminiero, line 363: Same as in screensharing: this function can be removed.

lminiero (commented):
Had a first quick look at the PR. I can't say I've looked in detail at every single change (I may need to do a second review later on), but these are some things I spotted in this round.

html/screensharingtest.js
atoppi: Ok, I am removing it in the next commit.

atoppi: oh it looks like it is being used in the script
https://github.com/meetecho/janus-gateway/blob/961b5764b34b5f84c56825f8acf05890d4551aa4/html/screensharingtest.js#L380
https://github.com/meetecho/janus-gateway/blob/961b5764b34b5f84c56825f8acf05890d4551aa4/html/screensharingtest.js#L419

html/streamingtest.js
atoppi: because it's being defined twice (https://github.com/meetecho/janus-gateway/blob/961b5764b34b5f84c56825f8acf05890d4551aa4/html/streamingtest.js#L466 and https://github.com/meetecho/janus-gateway/blob/961b5764b34b5f84c56825f8acf05890d4551aa4/html/streamingtest.js#L493)

html/textroomtest.js
atoppi: Removed since it was an unused param.
I'll add `label` and `protocol` params and the eslint directive to ignore the unused vars.

html/screensharingtest.js
lminiero: Yes but we can just use Janus.randomString() where the other function is used, they're exactly the same function.

html/textroomtest.js
atoppi: Yes my fault, there was a `var username` a couple of lines before.
Fixing it.

lminiero:
Merging.

PR #3172, opened 2023-02-28
Remove VoiceMail plugin

lminiero:
I decided to get rid of the VoiceMail plugin. It's a plugin I created at the very beginning as a way to test if a plugin could get Opus packets from a WebRTC participant, and then package them in an Opus file to distribute later, but it never went beyond that. As a matter of fact, I don't think anyone uses it, especially because to properly test it you have to do some weird configuration of the folder where you save to, plus the folder the file is visible from in a web server, and as a consequence even the online demo never really worked as it should. This makes it a pretty useless plugin from a functionality perspective: I don't think anyone's going to miss it, since if you need to record a message, the Record&Play plugin is vastly superior, and if you need RTP-to-Opus, `janus-pp-rec` does it much better too.

This is a PR to remove it from `1.x`, but I'll do the same in `0.x` once this is merged.


lminiero:
@renich pinging you just as an FYI, as it might mean one less package to provide once your process is ready.

lminiero:
Merging.

PR #3173, opened 2023-02-28
Added support for getDisplayMedia audio capture

rjnpnfigueiredo:
This is meant to address the issue of getDisplayMedia requests not asking for audio capture when capturing the screen.

The example of screensharingtest.html actually captures microphone audio and not the desktop audio which can cause unintended effects.

This is a possible fix that adds a new track type of "desktopAudio" which can be used in the track list:

> tracks: [
> 	{ type: 'desktopAudio', capture: true, recv: false },
> 	{ type: 'screen', capture: true, recv: false }
> ]

This allows for specific request of audio capture while screen capturing, avoiding confusing with microphone audio.

Since the audio and video capture request must be made with a single request and single constraint, the captured stream is buffered to be split into audio and video tracks as per regular behavior.

Audio capture in getDisplayMedia is still dependent on:
- Browser
- Operating System
- User permission

### DIFF
--- a/html/demos/janus.js
+++ b/html/demos/janus.js
@@ -2338,11 +2338,12 @@ var Janus = (function (factory) {
 			let groups = {};
 			for(let track of tracks) {
 				delete track.gumGroup;
-				if(!track.type || !['audio', 'video'].includes(track.type))
+				if(!track.type || !['audio', 'video', 'screen'].includes(track.type))
 					continue;
 				if(!track.capture || track.capture instanceof MediaStreamTrack)
 					continue;
-				let group = track.group ? track.group : 'default';
+				// Avoid assigning regular default group if it is a screen capture
+				let group = track.group ? track.group : (track.type === 'screen' ? 'screendefault' : 'default');
 				if(!groups[group])
 					groups[group] = {};
 				if(groups[group][track.type])
@@ -2353,11 +2354,13 @@ var Janus = (function (factory) {
 			let keys = Object.keys(groups);
 			for(let key of keys) {
 				let group = groups[key];
-				if(!group.audio || !group.video) {
+				if(!group.audio || (!group.video && !group.screen)) {
 					if(group.audio)
 						delete group.audio.gumGroup;
 					if(group.video)
 						delete group.video.gumGroup;
+					if(group.screen)
+						delete group.screen.gumGroup;
 					delete groups[key];
 				}
 			}
@@ -2474,7 +2477,7 @@ var Janus = (function (factory) {
 				}
 				if(track.capture) {
 					if(track.gumGroup && groups[track.gumGroup] && groups[track.gumGroup].stream) {
-						// We did a getUserMedia before already
+						// We did a getUserMedia/getDisplayMedia before already
 						let stream = groups[track.gumGroup].stream;
 						nt = (track.type === 'audio' ? stream.getAudioTracks()[0] : stream.getVideoTracks()[0]);
 						delete groups[track.gumGroup].stream;
@@ -2489,26 +2492,35 @@ var Janus = (function (factory) {
 							pluginHandle.consentDialog(true);
 						}
 						let constraints = Janus.trackConstraints(track), stream = null;
-						if(track.type === 'audio' || track.type === 'video') {
-							// Use getUserMedia: check if we need to group audio and video together
+						if(track.type === 'audio' || track.type === 'video' || track.type === 'screen') {
+							// Set if getUserMedia should be used or if it should be getDisplayMedia
+							let captureUserMedia = (track.type !== 'screen');
+							// Check if we need to group audio and video together
 							if(track.gumGroup) {
 								let otherType = (track.type === 'audio' ? 'video' : 'audio');
 								if(groups[track.gumGroup] && groups[track.gumGroup][otherType]) {
 									let otherTrack = groups[track.gumGroup][otherType];
+									if(!otherTrack && otherType === 'video') {
+										// If the other track is not video, then it is screen and should be a getDisplayMedia capture
+										otherType = 'screen'
+										otherTrack = groups[track.gumGroup]['screen'];
+										captureUserMedia = false;
+									}
 									let otherConstraints = Janus.trackConstraints(otherTrack);
 									constraints[otherType] = otherConstraints[otherType];
 								}
 							}
-							stream = await navigator.mediaDevices.getUserMedia(constraints);
+							if(captureUserMedia) {
+								stream = await navigator.mediaDevices.getUserMedia(constraints);
+							} else {
+								stream = await navigator.mediaDevices.getDisplayMedia(constraints);
+							}
 							if(track.gumGroup && constraints.audio && constraints.video) {
 								// We just performed a grouped getUserMedia, keep track of the
 								// stream so that we can immediately assign the track later
 								groups[track.gumGroup].stream = stream;
 								delete track.gumGroup;
 							}
-						} else {
-							// Use getDisplayMedia
-							stream = await navigator.mediaDevices.getDisplayMedia(constraints);
 						}
 						nt = (track.type === 'audio' ? stream.getAudioTracks()[0] : stream.getVideoTracks()[0]);
 					}


html/janus.js
```diff
@@ -2403,9 +2406,32 @@ function Janus(gatewayCallbacks) {
...
+						if(track.type === 'screen') {
+
+							//double check if audio capture is required (messy way)
+							for(const othertrack of tracks) {
+								if(othertrack.type === 'desktopAudio' && othertrack.capture) {
+									constraints.audio = othertrack.capture;
+								}
+							}
+							stream = await navigator.mediaDevices.getDisplayMedia(constraints);
+
```
lminiero, line 2419: See above on empty lines.

html/janus.js
```diff
@@ -2403,9 +2406,32 @@ function Janus(gatewayCallbacks) {
...
+							desktopAudio = stream;
+						}
+						if(track.type === 'desktopAudio') {
+							stream = desktopAudio;
+						}
+					}
+					if(track.type === 'desktopAudio' && (desktopAudio == null || desktopAudio.getAudioTracks().length == 0) ) {
+						continue; //ignore no audio in screen capture as it is set at the moment of screen capture
+					}
+
```
lminiero, line 2429: See above on empty lines.

html/janus.js
```diff
@@ -2403,9 +2406,32 @@ function Janus(gatewayCallbacks) {
...
+					nt = (track.type === 'audio' || track.type === 'desktopAudio' ? stream.getAudioTracks()[0] : stream.getVideoTracks()[0]);
+
```
lminiero, line 2431: See above on empty lines.

html/janus.js
```diff
@@ -2403,9 +2406,32 @@ function Janus(gatewayCallbacks) {
...
+					if(track.type === 'desktopAudio') {
+						track.type = 'audio'; //overrulling from here
```
lminiero, line 2433: See above on comment style.

html/screensharingtest.js
```diff
@@ -147,7 +147,7 @@ $(document).ready(function() {
 														{
 															// We want sendonly audio and screensharing
 															tracks: [
-																{ type: 'audio', capture: true, recv: false },
+																{ type: 'desktopAudio', capture: true, recv: false },
```
lminiero, line 150: I don't think this is a good idea, in the demo, or at the very least it should be optional and/or configurable. At least some time ago, desktop audio was only supported on Windows, and not other systems, which means you thought you asked for something and got something else entirely. This is why we fell back to "regular" audio in this demo, as it was meant as a way to "talk on top of your screen".

html/janus.js
```diff
@@ -2403,9 +2406,32 @@ function Janus(gatewayCallbacks) {
...
 					} else {
 						// Use getDisplayMedia
-						stream = await navigator.mediaDevices.getDisplayMedia(constraints);
+						// Figure out if it is a video or audio capture request
+						if(track.type === 'screen') {
+
+							//double check if audio capture is required (messy way)
+							for(const othertrack of tracks) {
+								if(othertrack.type === 'desktopAudio' && othertrack.capture) {
+									constraints.audio = othertrack.capture;
```
lminiero, line 2415: For some reason this comment I added disappeared, so I'll add a new single comment again. I don't like this approach, as it breaks the way `tracks` is supposed to work. You're not meant to simply pick any other random audio track from that list and use it here: we explicitly conceived `tracks` to be a collection of independent tracks, that you can either ask separately, or in some cases group together, e.g., in a single `getUserMedia` (which makes sense for instance when you're just asking for mic and webcam). Asking things together is done via the concept of `group`, which is explained in the original PR #3003. What you'e doing breaks that assumption, and will cause undefined or unexpected behaviours when one uses `tracks` to capture many different tracks at the same time.

html/screensharingtest.js
rjnpnfigueiredo: Currently Chrome/Edge on windows can share audio in all modes (tab, window, screen) as long as the user ticks the permission box in the share screen prompt (tick for audio only appears if you request audio in the getDisplayMedia)

For other systems, Chrome/Edge can still share audio, but only of tabs.

html/janus.js
rjnpnfigueiredo: Hello,

Thanks for the feedback on this, I tried to stick to the code style conventions better this time.
I've reworked the code to make 'screen' work along side 'video' and 'audio' in the groups:

- The 'screen' track type has its own default group name to avoid collisions with expected behaviors of default audio/video group
-  A choice is made to acquire stream from `getDisplayMedia` rather than `getUserMedia` if the track is of the 'screen' type or the track belongs to a group where 'screen' is the type of the other track.
- Current demo for screenshare works the same as before (reverted changes)
- Actually sharing screen with system audio now requires that the 'screen' track is in a group with an 'audio' track

I'm still evaluating combinations of tracks to ensure proper behavior, but I wanted to check if this would an acceptable direction for a solution.

html/janus.js
```diff
@@ -2269,11 +2269,12 @@ function Janus(gatewayCallbacks) {
...
 		for(let track of tracks) {
 			delete track.gumGroup;
-			if(!track.type || !['audio', 'video'].includes(track.type))
+			if(!track.type || !['audio', 'video', 'screen'].includes(track.type))
 				continue;
 			if(!track.capture || track.capture instanceof MediaStreamTrack)
 				continue;
-			let group = track.group ? track.group : 'default';
+			// Avoid assigning regular default group if it is a screen capture
+			let group = track.group ? track.group : track.type=='screen'?'sharescreendefault':'default';
```
lminiero, line 2277: Please put the second block of the ternary operator in brackets to avoid ambiguity, and space things accordingly to make it more readable, e.g.:

    (track.type === 'screen' ? 'sharescreendefault' : 'default')

As a side note, I'm not fond `sharescreendefault` as a name, too long and overly verbose: maybe just `screendefault`? I was going to suggest updating the documentation in `mainpage.dox` accordingly, but I just noticed we don't have any text about `group` in there at all where we talk of `tracks`, so I can add something myself after this is merged.

html/janus.js
```diff
@@ -2384,26 +2387,35 @@ function Janus(gatewayCallbacks) {
 						pluginHandle.consentDialog(true);
 					}
 					let constraints = Janus.trackConstraints(track), stream = null;
-					if(track.type === 'audio' || track.type === 'video') {
+					if(track.type === 'audio' || track.type === 'video' || track.type === 'screen') {
+						// Set if User Media should be used of if it should be Display Media
+						let captureUserMedia = track.type !== 'screen';
```
lminiero, line 2392: Please use brackets to avoid ambiguity and improve readability. As a side note, please write `getUserMedia` and `getDisplayMedia` in the comment above that, not "User Media" and "Display Media".

html/janus.js
```diff
@@ -2384,26 +2387,35 @@ function Janus(gatewayCallbacks) {
...
 						// Use getUserMedia: check if we need to group audio and video together
 						if(track.gumGroup) {
 							let otherType = (track.type === 'audio' ? 'video' : 'audio');
 							if(groups[track.gumGroup] && groups[track.gumGroup][otherType]) {
 								let otherTrack = groups[track.gumGroup][otherType];
+								if(!otherTrack && otherType === 'video') {
+									// If there other track is not video, then it is screen and should be a Display Media capture
```
lminiero, line 2399: Same here: `getDisplayMedia`

lminiero (commented):
Added a couple of extra notes.

html/janus.js
rjnpnfigueiredo: Tweaked code to better fit style as per comments.
Chained ternaries are not really readable but I don't think spinning it off into a function would fit into the current code style.
I'd say `screendefault` sounds good, lets keep that.

lminiero:
@rjnpnfigueiredo apologies for the lack of feedback: I'm currently abroad for a conference, so I won't be able to look at this for a little while longer. Pinging @atoppi so that he can check them too.

html/janus.js
```diff
@@ -2384,26 +2387,35 @@ function Janus(gatewayCallbacks) {
 						pluginHandle.consentDialog(true);
 					}
 					let constraints = Janus.trackConstraints(track), stream = null;
-					if(track.type === 'audio' || track.type === 'video') {
-						// Use getUserMedia: check if we need to group audio and video together
+					if(track.type === 'audio' || (track.type === 'video' || track.type === 'screen') ) {
```
atoppi, line 2390: unneeded extra parenthesis and space at the end of condition

html/janus.js
```diff
@@ -2384,26 +2387,35 @@ function Janus(gatewayCallbacks) {
...
+						// Set if getUserMedia should be used or if it should be getDisplayMedia
+						let captureUserMedia = track.type !== 'screen';
```
lminiero, line 2392: Just for the sake of readability, and to be in line with the rest of the code, please wrap the check above in brackets so that it's less ambiguous.

html/janus.js
```diff
@@ -2384,26 +2387,35 @@ function Janus(gatewayCallbacks) {
...
+						// Check if we need to group audio and video together
 						if(track.gumGroup) {
 							let otherType = (track.type === 'audio' ? 'video' : 'audio');
 							if(groups[track.gumGroup] && groups[track.gumGroup][otherType]) {
 								let otherTrack = groups[track.gumGroup][otherType];
+								if(!otherTrack && otherType === 'video') {
+									// If there other track is not video, then it is screen and should be a getDisplayMedia capture
```
lminiero, line 2399: I'm trying to wrap up my mind around this check and the action it implements as a consequence, but I'm struggling to understand the logic behind it. Could you elaborate on why this should work that way? The check and the comment seem at odds.

lminiero (commented):
Apologies for this late review, I've been abroad for quite some time and spent last week catching up with other activities. I added some notes inline: please also address @atoppi's comment, which seems to point out a syntax error (double closing bracket where only one was opened).

html/janus.js
rjnpnfigueiredo: Sorry for my own late reply, been otherwise occupied.

Logic is a bit strange, yes. This is meant to address the issue of asking for audio before screen in the group.
If the audio track is handled first, we need to know what the other track type is. If it fails to get the other track and we had the type at video (from the other type ternary above) then it means the other type is actually screen and we should treat this as a display media request.

rjnpnfigueiredo:
Hello, rebasing was proving a bit tricky so I just redid the code changes based on the latest master.

Issue #3169, opened 2023-02-19
[1.x] VP9 SVC room not working.


faihan6:
**What version of Janus is this happening on?**
1.1.3

**Have you tested a more recent version of Janus too?**
No.

**Was this working before?**
I am not sure.

**Is there a gdb or libasan trace of the issue?**
No.

**Additional context**
We have a demo VP9 SVC room to test chrome's Spatial SVC capabilities. Chrome has now made it possible that we can enable Spatial Scaling using the 'scalabilityMode' property via addTransceiver. The respective Chrome beta version is 111.0.5563.19.

I tried accessing the VP9 SVC demo room with two participants from the above mentioned browser and SVC was not working. The SVC selection controls above the feeds were not present.

I tried the same with Stable Chrome using the flag `--force-fieldtrials=WebRTC-SupportVP9SVC/EnabledByFlag_2SL3TL/` and I got same results.

Upon checking the webrtc-intenals, I could see that the `scalabilityMode` parameter was not present in the `outbound-rtp` stat. I checked the JS code and looks like we are missing to add the SVC parameter in the track object.

Commit 961b5764b34b5f84c56825f8acf05890d4551aa4 - File vp9svctest.js, Line 437
`tracks.push({ type: 'video', capture: true, recv: false, simulcast: false, svc: 'S3T3'});`

Making this change enabled VP9 SVC from the browser. I could see the `scalabilityMode` parameter now and the framesSent/s parameter has also doubled indicating that SVC is indeed working on the browser's end. However, the SVC controls on the video feeds were not present yet.

I checked the Source code of Janus Server and looks like the `janus_videoroom_publisher_stream` object of the appropriate mid will have a svc property and it has to be set to 1 for SVC to work. Unfortunately it has not been set.

I manually set the value to 1 by adding `if(!ps->svc) ps->svc = 1;` (inserted at line 8462. Filename: janus_videoroom.c. Same commit as mentioned above) and added logs to the `janus_vp9_parse_svc` function. It worked like a charm! I could see Two (not three unfortunately) spatial layers and Three temporal layers in the logs and SVC controls had now appeared in the browser. Clicking on those controls also changed the incoming resolution and framerate as expected.

I hope that's enough information to debug this issue. Thank you.

lminiero:
[...]

PR #3174, opened 2023-03-01
SVC related tweaks

lminiero:
This PR makes quite a few SVC related changes, since SVC was so far only limited to a custom VideoRoom demo that needed a specific room configuration too. Following the discussion in #3169, I decided to turn SVC more into a "first class citizen" instead, in order to allow it to be used pretty much as we do with simulcast today.

The first thing this PR does is add new helper functions for SVC processing, which work pretty much like the simulcast ones do: this way, we can use them independently in differenty plugins, rather than have a custom hardcoded behaviour as we did before. At the moment, it assumes VP9-SVC (since AV1 SVC is pretty much unusable right now), but we could extend it later on. These helper functions have then been integrated in both the EchoTest and VideoRoom plugins.

Another thing the PR does, now, is assume that the `jsep` property you send to a plugin can be extended with an array called `svc`, formatted like this:

    {
        "type": "offer",    // same as before
        "sdp": "v=0...",    // same as before
        "svc": [
            { "mindex": 1, "mid": "1", "svc": "L3T3" }
        ]
    }

This is needed because when you use SVC, the SDP doesn't reflect that, so if we want Janus (and the plugins) to know you're sending SVC, we need to say it out of band. Using custom messages in plugins was a possibility, but that would have forced each plugin to implement its own way, which I didn't like: better to have a core way to signal this, especially considering the `jsep` object does get to plugins, and so they can all process it the same way. This new `svc` array is quite basic, since it can contain a list of objects that say which stream being published is SVC, and treat it accordingly: at the time of writing, if you use `janus.js`, this property is set automatically (we iterate on transceivers before sending the SDP to Janus), but if you use your own stack or SDK you'll have to add it manually yourself instead.

To test all this in the EchoTest, just add a query string argument like `?vcodec=vp9&svc=L3T3`: this will tell `janus.js` to use SVC for the video stream, enrich the `jsep` object accordingly, and allow you to change spatial/temporal layers exactly as the simulcast flavour of the demo currently does. Notice that this only works if you use the "regular" EchoTest demo: the Lua and Duktape versions don't support SVC yet.

For what concerns the VideoRoom, I got rid of the `video_svc` property you can set in rooms, which means now regular, simulcast and SVC streams can all coexist in the same room, without any prerequisite other than "the related codec must be supported in the room". As a consequence, I also got rid of the `vp9svctest.html/.js` demo, since it's useless now: you can publish SVC in the regular VideoRoom demos instead, using the same query string argument shown for the EchoTest (make sure `vp9` is a supported video codec in your room, of course).

An example of simulcast and SVC publishers coexisting in the same room can be seen in the screenshot below:

Screenshot_2023-03-01_18-08-46

As you can see, both the publishers we supported too gave us options (we have buttons to select things), but one is using simulcast, and the other is using SVC, as the labels next to their current video resolution tells us.

Please do test this, especially if you're interested in testing SVC support, as I'd like to merge soon. Notice that, due to the amount of changes, this will remain a multistream only PR, as I don't plan to backport the same changes to `0.x` too.


lminiero:
Merging.

PR #3179, opened 2023-03-13
Don't create IPv6 sockets if IPv6 is completely disabled (see #3159)

lminiero:
As explained in #3159, the change we made there to switch to `getaddrinfo` caused the involved plugins to break on machines where IPv6 is entirely disabled (and not just not present). See [this post on Discourse](https://janus.discourse.group/t/error-creating-audio-sockets/65) for context. This PR applies the same fix we added in #2916 for forwarders to the sockets we changed in SIP/NoSIP/AudioBridge.

Please test and provide feedback, as I have no way to test this myself (I don't have IPv6 on my machine but it's not disabled).


lminiero:
Merging.

PR #3184, opened 2023-03-17
Add newline for SIP headers that are overflown in length

zayim:
Before this PR, if some header (key+value) is overflown (longer than 256 chars), it would result in it being truncated, but no newline was added, so it would result in next header actually being written on same line as overflown one, hence it was not interpreted as separate header on SIP level.

This PR makes sure that newline is added always, even if header (key+value) is longer than 256 chars.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1619,9 +1619,10 @@ static void janus_sip_parse_custom_headers(json_t *root, char *custom_headers, s
 					continue;
 				}
 				char h[255];
-				g_snprintf(h, 255, "%s: %s\r\n", key, json_string_value(value));
+				g_snprintf(h, 255, "%s: %s", key, json_string_value(value));
 				JANUS_LOG(LOG_VERB, "Adding custom header, %s\n", h);
 				janus_strlcat(custom_headers, h, size);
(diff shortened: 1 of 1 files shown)


lminiero:
I'm not sure I understand why this should work. If the original string fills the string already, trying to append `\r\n` with `janus_strlcat` after that would still result in truncation and those two characters not being added. Am I missing something?

zayim:
that might be true if we overflow entire `custom_headers` string, I was solving the case when we overflow one single header that is being written in `h` string.

I can add handling of the case when entire `custom_headers` string is overflown, make sure that we always have newline at the end of it

lminiero:
I think that might help, thanks!

zayim:
When you have time, check this again, it is now guaranteed that `custom_headers` will also end with `\r\n`, even if it is overflown.

lminiero:
Hi @zayim, apologies for this very late reply, but I've been abroad the past few weeks and have just come back to work. I'll review the changes later today and let you know, thanks!

lminiero:
Sorry again for the late review, was sidetracked by other activities... merging, thanks! I'll backport this to `0.x` too, since it's an easy patch to apply.

Issue #3196, opened 2023-04-01
[1.x] any reason for switch to not have substream param?


brave44:
In videoroom plugin, switch command doesn't have substream param, as subscribe or configure command have (https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_videoroom.c#L9428). Simulcast just resets on switching (https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_videoroom.c#L11237 
).

It forces us to do configure to needed substream every time we doing switch, not ideal, especially if room have many streams.

Is it some specific reason to such implementation is switch, or it's just not implemented? Can we do PR with this logic?

lminiero:
No specific reason: I think we recently fixed the simulcast configuration in multistream VideoRoom, and probably just forgot to do the same for switch too. A pull request would indeed be welcome: please notice I'll be abroad in vacation the whole week, so I may be very slow to review until I'm back.

lminiero:
Merged.

PR #3197, opened 2023-04-02
Add substream param to switch in VideoRoom

brave44:
Ability to specify substream while switching in VideoRoom.

https://github.com/meetecho/janus-gateway/issues/3196

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -1878,7 +1878,10 @@ static struct janus_json_parameter switch_parameters[] = {
 static struct janus_json_parameter switch_update_parameters[] = {
 	//~ {"feed", JANUS_JSON_INTEGER, JANUS_JSON_PARAM_REQUIRED | JANUS_JSON_PARAM_POSITIVE},
 	{"mid", JANUS_JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
-	{"sub_mid", JANUS_JSON_STRING, JANUS_JSON_PARAM_REQUIRED}
+	{"sub_mid", JANUS_JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
+	/* For VP8 (or H.264) simulcast */
+	{"substream", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
+	{"temporal", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE}
 };
 static struct janus_json_parameter publish_remotely_parameters[] = {
 	{"secret", JSON_STRING, 0},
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @brave44! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

brave44:
> Apologies for this late response, I have been abroad for a few weeks.
> 
> I think this is very incomplete at the moment. First of all, you're accessing a new property but not validating it, which is something we do automatically with `JANUS_VALIDATE_JSON_OBJECT`: check `switch_update_parameters` which is where the stream properties we support are defined (and which looks like where you're adding the new property). Besides, for some reason you only added a way to specify substreams but not the simulcast temporal layer: any functionality extension that only adds one of them and not both is by definition incomplete.

Thanks for reply, I'll correct the request according to your comments.

Right now we found out that `simulcasting_context_reset` in switch causes video freeze right after switch. Picture freezes and we monitor freezeCount and totalFreezesDuration in webRTC stats.

By removing 

```
janus_rtp_simulcasting_context_reset(&stream->sim_context);
janus_vp8_simulcast_context_reset(&stream->vp8_context);
janus_rtp_svc_context_reset(&stream->svc_context);
```

from switch (along with setting substream param) we have no freezes at all. Plus, if only set substream but not remove _context_reset, I see bitrate spike right after switch. (Spike to default bitrate witch is max bitrate). I think it's understandable and that what context resetting causes - it resets to defaults.

Maybe we don't need context resetting in switch at all? Simulcast context already been set, we working with existed stream, and just need new data in this stream, with ability to set desired substream. Why reset needed? What you think?

lminiero:
> Maybe we don't need context resetting in switch at all? Simulcast context already been set, we working with existed stream, and just need new data in this stream, with ability to set desired substream. Why reset needed? What you think?

The reset is needed because the video publisher is not the same. In fact, we reset and then copy the `rid_ext_id` property, which may be different (or entirely unavailable), plus the new substream/temporal targets. This way it's as if we're starting fresh with a new stream (which indeed we are).

Are you sure the problem is not in `janus_vp8_simulcast_context_reset(&stream->vp8_context)` instead? That one may indeed have to be left alone, since just as `janus_rtp_switching_context` it's more specific to the subscriber stream itself: more precisely, it dictates how to rewrite PictureIDs, when doing simulcast, so resetting that may cause the wrong one to be written in RTP packets, possibly causing freezes.

brave44:
> Are you sure the problem is not in `janus_vp8_simulcast_context_reset(&stream->vp8_context)` instead? That one may indeed have to be left alone, since just as `janus_rtp_switching_context` it's more specific to the subscriber stream itself: more precisely, it dictates how to rewrite PictureIDs, when doing simulcast, so resetting that may cause the wrong one to be written in RTP packets, possibly causing freezes.

Thanks, will try and let you know the results.

brave44:
@lminiero I think you was right, janus_vp8_simulcast_context_reset was causing freezes, everything working good without it. Plus, I added validations and temporal layer param. Hope you'll take a look.

lminiero:
Thanks for the patch! I'll have a look later this afternoon and merge if it works for me too :+1:

lminiero:
Looks good, merging :v:

lminiero:
I guess I'll have to add a similar approach for SVC as well, which at the moment isn't covered either.

brave44:
Thanks for your great work, man!

Issue #3201, opened 2023-04-20
[1.x] Cannot change video devices in android after initial set up


rjnpnfigueiredo:
Issue can be quickly verified by using the Janus device selection demo from the main website
1.x - https://janus.conf.meetecho.com/devicetest.html
0.x - https://janus-legacy.conf.meetecho.com/devicetest.html

Steps to reproduce:

1. Start demo
2. Select any video device
3. Press "Change Devices"
4. Wait for video to start
5. Select a different video device
6. Press "Change Devices"
7. Issue should happen

**What version of Janus is this happening on?**
This is happening to 1.x but 0.x versions seem to work fine

**Have you tested a more recent version of Janus too?**
Tested most recent version

**Was this working before?**
0.x is working, 1.x is not

**Is there a gdb or libasan trace of the issue?**
From android chrome console logs:

```
janus.js:2273 DOMException: Could not start video source
replaceTracks @ janus.js:2273
await in replaceTracks (async)
replaceTracks @ janus.js:1359
restartCapture @ devicetest.js:124
(anonymous) @ devicetest.js:88
dispatch @ jquery.min.js:3
v.handle @ jquery.min.js:3
```

lminiero:
`Could not start video source` means the WebRTC stack could not access the webcam, so it doesn't look like a video device changev ssue to me, but a problem with the device itself.

rjnpnfigueiredo:
> `Could not start video source` means the WebRTC stack could not access the webcam, so it doesn't look like a video device change issue to me, but a problem with the device itself.

I initially though the same, but given that the same example using the 0.x version works fine on the exact same device under the same circumstances, there may be some issue on how it is acquiring access to a different video device after initial capture.

lminiero:
I've noticed that the demo behaves weirdly on desktop too, since both webcams remain active in the local box. I suspect that's what may be causing the problem on Android too (maybe you can't capture both webcams at the same time?). I'll check what causes the demo to behave like that and come up with a fix, in case.

lminiero:
@rjnpnfigueiredo please refer to the PR above.

rjnpnfigueiredo:
@lminiero Looks all good on my end!

lminiero:
Ack, I'll merge then, thanks!

PR #3203, opened 2023-04-24
Remove old tracks before adding/replacing new ones in janus.js

lminiero:
When we implemented #3003 in `janus.js`, we made it so when adding/replacing new tracks, we'd first capture the new track and only after that we'd remove the old one: actually, when doing `replace: true`, the previous track would not be removed at all. Apparently this caused problems on some devices, e.g., as reported in #3201.

This patch tries to address that, by removing a track before it's updated. We had a FIXME in there since this may cause problems, as with this change there could be some time in between a track is removed and then replaced when nothing is added. As such, I'm providing this as a PR: please test and ensure it doesn't cause regressions or problems, as I plan to merge soon if it fixes the problem.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -2390,11 +2390,46 @@ function Janus(gatewayCallbacks) {
 			}
 			// Capture the new track, if we need to
 			let nt = null, trackId = null;
-			if(track.remove) {
+			if(track.remove || track.replace) {
 				Janus.log('Removing track from PeerConnection', track);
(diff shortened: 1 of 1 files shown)


rjnpnfigueiredo:
With this code change Android has the correct behavior and has not introduced any issues on other platforms as far as I can test.

lminiero:
Ack, merging then :+1:

PR #3208, opened 2023-04-27
avoid parsing whitespace as invalid JSON

htrendev:
A single WebSocket message may contain multiple requests. The code parses one request, then continues looking for more requests in the same WebSocket message. This works fine as long as the **last** request has no trailing whitespace. However, if there is any trailing whitespace after the last request an attempt to parse it as JSON results in an error.

This PR simply skips any trailing whitespace after each request

### DIFF
--- a/src/transports/janus_websockets.c
+++ b/src/transports/janus_websockets.c
@@ -1347,6 +1347,9 @@ static int janus_websockets_common_callback(
 					incoming_curr += error.position;
 					JANUS_LOG(LOG_HUGE, "[%s-%p] Parsed JSON message - consumed %zu/%zu bytes\n",
 						log_prefix, wsi, (size_t)(incoming_curr - ws_client->incoming), incoming_length);
+					/* Trailing whitespace after the last message results in invalid JSON error */
+					while (incoming_curr < incoming_end && isspace(*incoming_curr))
+						incoming_curr++;
 					if(incoming_curr == incoming_end) {
 						/* Process messages in order */
 						json_t **msg = message_buffer;


januscla:
Thanks for your contribution, @htrendev! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

htrendev:
CLA signed

src/transports/janus_websockets.c
```diff
@@ -1347,6 +1347,9 @@ static int janus_websockets_common_callback(
 					incoming_curr += error.position;
 					JANUS_LOG(LOG_HUGE, "[%s-%p] Parsed JSON message - consumed %zu/%zu bytes\n",
 						log_prefix, wsi, (size_t)(incoming_curr - ws_client->incoming), incoming_length);
+					/* Trailing whitespace after the last message results in invalid JSON error */
+					while isspace(*incoming_curr)
+						incoming_curr++;
```
lminiero, line 1352: This sounds like a dangerous check to me, though, since it's increasing the pointer and dereferencing it without ever checking if it got to the end and beyond (i.e., after `incoming_end`). This will crash if `incoming_curr` leaves the boundaries of the allocated space.

lminiero (commented):
Thanks for the patch, Hristo! I've added a note inline.

src/transports/janus_websockets.c
htrendev: I considered this for a moment, but couldn't see how this could overflow. Still, I agree that it's better to be safe than sorry, so I pushed an update. Thanks for the feedback!

lminiero: Thanks! I'll make a couple of quick ckecks later (or tomorrow morning) and in case I can't spot any issue, I'll merge both PRs (thanks for providing a backported version too) :+1:

PR #3211, opened 2023-04-29
Javascript tweaks

i8-pi:
A few improvements in the Javascript that should not impact functionality

The first commit
- Adds the html plugin to eslint config to enable linting Javascript in html script tags
- Adds a command in package.json to allow running `npm run lint` from the command line, which checks both js and html files
- Updates github ci to run the same command

The second commit swaps plain objects with JS Maps where the objects are used as arbitrary key-value stores. Plain objects come with some predefined keys
``` js
const obj = {};
typeof obj.toString === 'function' // true
```
which can be a problem if the keys come from user input, while Maps are specifically designed to be arbitrary key-value stores
The code changes are mostly mechanical
``` js
obj[key]
// becomes
map.get(key)

obj[key] = value
// becomes
map.set(key, value)

delete obj[key]
// becomes
map.delete(key)

for (const key in obj)
  if (obj.hasOwnProperty(key))
    body
// becomes
for (const key of map.keys())
  body
```


lminiero:
THanks! I'll let @atoppi comment on this, since most of the JS stuff I'm clueless about.

package.json
```diff
@@ -18,6 +18,10 @@
     "webrtc-adapter": "8.2.2"
   },
   "devDependencies": {
-    "eslint": "8.35.0"
+    "eslint": "8.35.0",
+    "eslint-plugin-html": "^7.1.0"
+  },
+  "scripts": {
+    "lint": "eslint html --ext .js,.html"
```
atoppi, line 25: What is the scope of `eslint` here?
Are we sure that the script is going to use the binary installed by npm (e.g. the one under `./node_modules`) and not a global one already installed in the environment?

package.json
```diff
@@ -18,6 +18,10 @@
     "webrtc-adapter": "8.2.2"
   },
   "devDependencies": {
-    "eslint": "8.35.0"
+    "eslint": "8.35.0",
+    "eslint-plugin-html": "^7.1.0"
```
atoppi, line 22: Could you please:
- upgrade eslint to the latest version `8.39.0`
- use a fixed version of the `eslint-plugin-html`

html/janus.js
```diff
@@ -2339,9 +2339,9 @@ function Janus(gatewayCallbacks) {
 				};
 				continue;
 			}
-			if((typeof track.add === 'undefined' && track.add === null) &&
-					(typeof track.remove === 'undefined' && track.remove === null) &&
-					(typeof track.replace === 'undefined' && track.replace === null)) {
+			if(track.add == null &&
+					track.remove == null &&
+					track.replace == null) {
```
atoppi, line 2344: indeed the previous check was wrong but we must address both `undefined` and `null` cases, so it should be:

```js
			if((typeof track.add === 'undefined' || track.add === null) &&
					(typeof track.remove === 'undefined' || track.remove === null) &&
					(typeof track.replace === 'undefined' || track.replace === null)) {
```

atoppi (changes requested):
Hello @i8-pi and thanks for the effort!

Since we are migrating core objects to maps, I'd like to switch[ Janus.sessions](https://github.com/meetecho/janus-gateway/blob/013155d6e806277f8eee6272388dc95ab5294ef8/html/janus.js#L28) too.
Could you please add it to the PR?

Also please address the notes I've left.

package.json
i8-pi: The locally installed version will be used (first) when using `npm run`

package.json
i8-pi: Fixing the package versions is typically done by committing the package-lock.json file, which also specifies the versions of any indirect dependencies. If you want fixed package versions I recommend doing that

html/janus.js
i8-pi: `x == null` is actually equivalent to `x === undefined || x === null`, but yes, it is not obvious to anyone who hasn't memorized the truth table. I'll update this

i8-pi:
I will update Janus.sessions to use Maps as well. This will be externally visible though, unlike the other two changes, so maybe we should mentions that in the changelog

package.json
atoppi: I agree a package lock is a more convenient way of fixing packets, but that can be done in a separate effort.
Since eslint is already fixed to a specific version, I'd just want to align the linter and the plugin versions.

atoppi:
> I will update Janus.sessions to use Maps as well. This will be externally visible though, unlike the other two changes, so maybe we should mentions that in the changelog

Good point.
That might break things for people using that specific object. Anyway mixing things up (e.g. using objects here, maps there etc.) for similar structs is not a good design imho.

Pinging @lminiero to help taking a decision, since he is the main maintainer/user/hacker of the library.

Side note: in case we decide to proceed with sessions Map, we should also update the types.

i8-pi:
I've made the requested changes. The Janus.sessions change is there if you want it. The typescript files don't mention Janus.sessions and no changes should be required for that

atoppi (approved):

PR #3214, opened 2023-05-03
Use speexdsp's jitter buffer in the AudioBridge

lminiero:
The audio buffer in the AudioBridge has historically been a bit of its Achilles heel: while it worked fine in most situations, it definitely struggled more with problematic connections (e.g., high jitter or packet loss), which often resulted in broken or spotty audio for specific participants that could only be solved by having them reconnect (until the problem happened again).

This PR tries to fix this by using a proper jitter buffer, specifically the one made available as part of the Speex DSP library. In our tests in different circumstances where we could replicate problems before it seems to perform much better, so we thought we'd make this available as a PR so that you can test it before we merge. As a result of this change, we refactored a bit of related code too, which means that for instance the `prebuffer` property you used before is gone from configuration and API: in fact, that was only meaningful to our own old buffer, and doesn't make much sense using this new jitter buffer instead.

If you use the AudioBridge for anything in your application, you most definitely want to test this. Notice that this will be a breaking change for many: in fact, the fact we now rely on Speex DSP for its jitter buffer also means that the AudioBridge will not compile unless the library is available when rebuilding Janus. In most deployments currently relying on the AudioBridge this may not be the case, meaning they'd get stuck with an older version of the plugin. As such, make sure you install the library before compiling this PR, or when updating Janus after it's merged.

Also notice this PR is for the `0.x` branch: when it will come the time to merge this, I'll port all the same changes to `master` as well.

Looking forward to your feedback!

### DIFF
--- a/plugins/janus_audiobridge.c
+++ b/plugins/janus_audiobridge.c
@@ -38,7 +38,6 @@ room-<unique room ID>: {
 	audiolevel_event = true|false (whether to emit event to other users or not, default=false)
 	audio_active_packets = 100 (number of packets with audio level, default=100, 2 seconds)
 	audio_level_average = 25 (average value of audio level, 127=muted, 0='too loud', default=25)
-	default_prebuffering = number of packets to buffer before decoding each participant (default=DEFAULT_PREBUFFERING)
 	default_expectedloss = percent of packets we expect participants may miss, to help with FEC (default=0, max=20; automatically used for forwarders too)
 	default_bitrate = default bitrate in bps to use for the all participants (default=0, which means libopus decides; automatically used for forwarders too)
 	record = true|false (whether this room should be recorded, default=false)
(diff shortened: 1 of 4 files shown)


plugins/janus_audiobridge.c
```diff
@@ -8495,6 +8355,7 @@ static void *janus_audiobridge_participant_thread(void *data) {
 			g_free(mixedpkt->data);
 			g_free(mixedpkt);
 		}
+		g_usleep(2500);
```
tmatth, line 8358: Was this added to avoid busy waiting?

tmatth (commented):
Just some questions but LGTM.

plugins/janus_audiobridge.c
lminiero: Yep: before, `janus_audiobridge_participant_thread` was only used for outgoing packets, and so we had a `g_async_queue_timeout_pop` that would wait 100ms at max. Since now there's two different queues to probe, we peek both and return right away if they're empty (to check if we need to need anything with the other). This way we give the thread some idle time.

plugins/janus_audiobridge.c
```diff
@@ -8430,11 +8166,135 @@ static void *janus_audiobridge_participant_thread(void *data) {
...
+					janus_mutex_unlock(&participant->qmutex);
+					first = FALSE;
+					locked = FALSE;
+					rtp = (janus_rtp_header *)bpkt->buffer;
+					/* If this is Opus, check if there's a packet gap we should fix with FEC */
+					use_fec = FALSE;
+					if(!first && participant->codec == JANUS_AUDIOCODEC_OPUS && participant->fec) {
+						if(ntohs(rtp->seq_number) != participant->expected_seq) {
+							/* Lost a packet here? Use FEC to recover */
+							use_fec = TRUE;
```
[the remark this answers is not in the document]
lminiero: It's supposed to be implicit, in the sense that this check is done on an out of order packet, so the packet that follows a loss (we detect a gap). In that case, even if there was a loss higher than 1, we only inject a single FEC decoded packet anyway, since previous losses if any are ignored.

lminiero:
> Just some questions but LGTM.

Thanks for the review!

lminiero:
Merging.

PR #3219, opened 2023-05-11
Add an extra reference to session for VideoRoom participants

lminiero:
This is a PR to address a different crash caused by a race condition, as reported in #3154.


tmatth:
@lminiero does this need a backport?

lminiero:
I'm not sure: this patch, and #3167 before it, was quite focused on how the multistream version of the VideoRoom plugin works. I'm not aware of similar issues in `0.x`. That said, this is generic enough that it might make sense.

PR #3221, opened 2023-05-12
New regular event in handlers for core status

lminiero:
This PR adds a new core event that is sent via event handlers, to provide, on a regular basis, short info on the status of the Janus server. At the moment, this includes:

* the number of active sessions
* the number of active handles (across all sessions)
* the number of active PeerConnections (across all sessions)
* the interval at which event handlers are currently configured to send media reports for each PC

While I may add more stuff in the future, this is basically it for now, as it should provide some very basic health state information that can be consumed without having to poll the Admin API or track all events that flow through.

This is an example of how the event looks like:

    {
        "emitter": "MyJanusInstance",
        "type": 256,
        "subtype": 1,
        "timestamp": 1683906438598123,
        "event": {
            "status": "update",
            "info": {
                "sessions": 0,
                "handles": 0,
                "peerconnections": 0,
                "stats-period": 1
            }
        }
    }

In this PR, I hardcoded the frequency of this event to 15 seconds. Not sure if it makes sense to make this configurable too, but in case it should be easy to do. Feedback welcome!

### DIFF
--- a/src/ice.c
+++ b/src/ice.c
@@ -396,6 +396,12 @@ gboolean janus_ice_event_get_combine_media_stats(void) {
 	return janus_ice_event_combine_media_stats;
 }
 
+/* Number of active PeerConnection (for stats) */
+static volatile gint pc_num = 0;
+int janus_ice_get_peerconnection_num(void) {
+	return g_atomic_int_get(&pc_num);
+}
+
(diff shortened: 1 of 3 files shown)


lminiero:
Merging.

PR #3227, opened 2023-05-29
pp-av1: use obuparse

tmatth:
WIP

Currently a discrepancy after ever "Detected new resolution", size is 1 less and header type is 2 instead of 6:

image

Upstream is here: https://github.com/dwbuiten/obuparse
License is permissve (ISC)

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -572,6 +572,8 @@ janus_pp_rec_SOURCES = \
 	postprocessing/pp-webm.c \
 	postprocessing/pp-webm.h \
 	postprocessing/janus-pp-rec.c \
+	postprocessing/obuparse.c \
+	postprocessing/obuparse.h \
 	log.c \
 	utils.c \
 	version.c \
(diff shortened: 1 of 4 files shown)


lminiero:
That looks like a lot of additional code, and I'm not sure if there's a reason for it: after all, the post-processor code mostly only cares about OBU when figuring out the resolution, and only in part for depacketization. For that, isn't our simpler code enough? Besides, to be honest I'm not sure I'm comfortable embedding a bunch of external code I'd then have to keep up to date with: I went through that with Duktape in the past and it was a mess, and in fact we changed that to an external dependency to be installed separately.

tmatth:
> That looks like a lot of additional code, and I'm not sure if there's a reason for it: after all, the post-processor code mostly only cares about OBU when figuring out the resolution, and only in part for depacketization. For that, isn't our simpler code enough? Besides, to be honest I'm not sure I'm comfortable embedding a bunch of external code I'd then have to keep up to date with: I went through that with Duktape in the past and it was a mess, and in fact we changed that to an external dependency to be installed separately.

@lminiero I marked this as draft as I'm not sure if it's worth merging it yet. I also want to validate the existing parsing.

lminiero:
Ack, I'll wait for more feedback then, thanks for the clarification!

tmatth:
I decided to drop this since the OBUParse use case isn't really the same fit as RTP parsing

PR #3228, opened 2023-05-31
Fix rtcp out quality stats

atoppi:
We got notified about some spurious values of out link quality metrics.

This PR aims at fixing that unwanted behavior, specifically by:
- parsing the reported total lost as a signed integer since it could be a negative value
```
cumulative number of packets lost: 24 bits
      The total number of RTP data packets from source SSRC_n that have
      been lost since the beginning of reception.  This number is
      defined to be the number of packets expected less the number of
      packets actually received, where the number of packets received
      includes any which are late or duplicates.  Thus, packets that
      arrive late are not counted as lost, and the loss may be negative
      if there are duplicates.
```
- improving the float math when evaluating the metrics
- as a side effect, dealing internally with inbound losses by using signed integer, but still skipping negative cumulative losses in the RTCP reports sent to RTCP peers
```
	uint32_t expected_interval = ctx->expected - ctx->expected_prior;
	uint32_t received_interval = ctx->received - ctx->received_prior;
	int32_t lost_interval = 0;
	if (expected_interval > received_interval) {
		lost_interval = expected_interval - received_interval;
	}
	ctx->lost += lost_interval;
```

### DIFF
--- a/src/ice.c
+++ b/src/ice.c
@@ -4216,7 +4216,8 @@ static gboolean janus_ice_outgoing_rtcp_handle(gpointer user_data) {
 				.video = (medium->type == JANUS_MEDIA_VIDEO), .buffer = rtcpbuf, .length = srlen+sdeslen };
 			janus_ice_relay_rtcp_internal(handle, medium, &rtcp, FALSE);
 			/* Check if we detected too many losses, and send a slowlink event in case */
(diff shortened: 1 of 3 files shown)


lminiero:
Merging :+1:  I'll apply the same fixes to `0.x` too.

PR #3231, opened 2023-06-06
T.140 support in SIP plugin, and WebRTC gateway (again! see #1898)

lminiero:
This is an attempt to revamp the effort originally started in #1898, and at the time described in [this blog post](https://www.meetecho.com/blog/realtime-text-sip-and-webrtc/), but this time based on the current master (which means we leverage the existing RED support rather than add it to the plugin as we did before). You can refer to that PR and blog post for more context. This will also be discussed at Kamailio World 2023 in a presentation, so I'll share info on that too once that's done.


Odinvt:
Based on the RNNoise PR (#3185) theme of adding opt-in AI processing to the AudioBridge plugin. We have a feature down the roadmap of adding AI transcription using OpenAI's whisper C++ port https://github.com/ggerganov/whisper.cpp which i think would be a great addition for people that bridge between the SIP & AudioBridge plugins along with T.140 support. I will be giving it a try if you guys think it's a good idea & if you would indulge a man's very poor C skills ^^'

lminiero:
That's something I'd expect to happen orthogonally to Janus and not in Janus itself, though. I already wrote of how [AudioBridge and SIP can work together](https://www.meetecho.com/blog/audiobridge-sip/), for instance, and transcriptions themselves are things we've been working on too, but using RTP forwarders to do them externally.

edwtie:
we have now implement this features at our new project teletyptel

edwtie:
added https://xmpp.org/extensions/inbox/jingle-rtt-sync.html

PR #3233, opened 2023-06-12
Use speexdsp's jitter buffer in the AudioBridge (1.x)

lminiero:
Same as #3214, but for master. Again, this introduces a new mandatory dependency for the AudioBridge, so make sure you test it so that you're prepared for the change, as soon as we merge this.


lminiero:
Merging.

PR #3234, opened 2023-06-13
Add support for ICE consent freshness (requires libnice >= 0.1.19)

lminiero:
As the subject says, this adds support for ICE consent freshness, which was recently (since 0.1.19) added to libnice. While we already had configurable support for using connectivity checks as keep-alives (`ice_keepalive_conncheck: true` in `janus.jcfg`), we didn't have support for consent freshness yet. The documentation for the libnice feature can be found [here](https://libnice.freedesktop.org/libnice/NiceAgent.html#NiceAgent--consent-freshness).

This PR adds configurable support for consent freshness too, now, via a new configuration property called `ice_consent_freshness`. Notice that, as explained in the documentation, enabling consent freshness automatically enables `keepalive-conncheck` too.

As part of this patch, we added a new `ice-failed` event that the Janus API can originate. If, as a consequence of enabling these features, Janus detects the PC going to a `failed` state after it was previously connected/ready (meaning the PC may be gone), an `ice-failed` event will be sent to the user to whom the handle belongs. The rationale for that is that we wait for a few seconds before closing the PC for good, which should give the user some time to possibly try and recover, e.g., via an ICE restart (in case the `failed` detection wasn't a truly fatal one).

While we plan to merge soon, we encourage you to test this a bit and provide feedback, most importantly to ensure enabling this feature doesn't introduce potential issues (e.g., too aggressively closing things that really shouldn't be closed, maybe because of occasional packet loss affecting connectivity checks rather than a PC that has truly gone away).

As usual, I'll backport this to `0.x` too once this is deemed ready to be merged.

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -270,8 +270,11 @@ media: {
 # it should work in ICE-Lite mode (by default it doesn't). If libnice is
 # at least 0.1.15, you can choose which ICE nomination mode to use: valid
 # values are "regular" and "aggressive" (the default depends on the libnice
-# version itself; if we can set it, we set aggressive nomination). You can
-# also configure whether to use connectivity checks as keep-alives, which
+# version itself; if we can set it, we set aggressive nomination). If
+# libnice is at least 0.1.19, you can enable consent freshness checks for
+# PeerConnections as well: this will issue regular checks to check whether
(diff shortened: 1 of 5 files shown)


lminiero:
Merging.

lminiero:
Backported to `0.x` too.

PR #3239, opened 2023-06-29
Batch configure support for streaming plugin

petarminchev:
Supporting a streams array param to configure multiple mids in one batch request.

### DIFF
--- a/src/plugins/janus_streaming.c
+++ b/src/plugins/janus_streaming.c
@@ -1180,6 +1180,7 @@ static struct janus_json_parameter svc_parameters[] = {
 };
 static struct janus_json_parameter configure_parameters[] = {
 	{"mid", JANUS_JSON_STRING, 0},
+	{"streams", JANUS_JSON_ARRAY, 0},
 	{"send", JANUS_JSON_BOOL, 0},
 	/* For VP8 (or H.264) simulcast */
 	{"substream", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @petarminchev! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero (commented):
It seems ok, but I haven't tested it yet: I only added a couple of editorial comments. What's definitely missing, though, is documentation: the `streams` array for "configure" is not documented in the doxygen part of the code, which we did in #2986.

I just noticed though that the "configure" request in the Streaming plugin seems to only do something on RTP (and RTSP) mountpoints, though, which means that the file-based mountpoints were never impacted. This is a note to self, since it means I'll probably have to add something for them too (at the very least to tweak the `send` property, for instance).

lminiero:
@petarminchev thanks! I'll make a couple of tests later today, and I've launched the CI checks on the branch too.
I noticed you haven't signed the CLA though? Please do that at your earliest convenience, since without that step I can't merge this.

lminiero:
It must be signed by you: you're the author of the patch.

lminiero:
Sorry for the lack of feedback! I plan to make a couple of tests later today and merge in case it's fine :+1:

lminiero:
Made a few tests with a multistream mountpoint with multiple simulcast video streams, and it does seem to work as expected :+1:  I vedified that a "legacy" configure request with no mid seems to be enforced on all streams, which means it is backwards compatible too. Merging then, thanks again!

PR #3241, opened 2023-07-12
janus.d.ts: fix typo and update PluginCreateAnswerParam to use tracks

veeting:
Hello!

This is a very simple PR affecting only the janus.d.ts file.

1/ We found a typo where **TrackOptions** should be **TrackOption**
2/ We also have updated the PluginCreateAnswerParam type to have the newer **tracks** parameter and marked the **media** as deprecated. (same as OfferParam)

Best regards, Aymeric

### DIFF
--- a/npm/janus.d.ts
+++ b/npm/janus.d.ts
@@ -167,7 +167,7 @@ declare namespace JanusJS {
 	}
 
 	interface OfferParams {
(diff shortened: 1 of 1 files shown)


atoppi:
lgtm :+1:

lminiero:
Ack, merging then!

PR #3243, opened 2023-07-13
Fix simulated leave message for longer string ID rooms

adnanel:
Currently, on abrupt disconnections, participants get stuck in text rooms and aren't able to re-join immediately.

This makes auto reconnect mechanisms needlessly complicated.

In this PR I added automatic leaving of all rooms when the session is being destroyed.

The bug can be reproduced in the official janus textroom demo:
- Join textroom via phone
- Join textroom via PC
- Disconnect internet on phone
- On PC you can still see this participant for

### DIFF
--- a/src/plugins/janus_textroom.c
+++ b/src/plugins/janus_textroom.c
@@ -3027,7 +3027,7 @@ static void janus_textroom_hangup_media_internal(janus_plugin_session *handle) {
 	}
 	janus_mutex_unlock(&session->mutex);
 	JANUS_LOG(LOG_VERB, "Leaving %d rooms\n", g_list_length(list));
-	char request[100];
+	char request[200];
 	GList *first = list;
 	while(list) {
 		char *room_id_str = (char *)list->data;


lminiero:
Thanks for the patch, but I think you're doing those things in the wrong place. You've added some logic to the `destroy_session` callack, but it's actually the `hangup_media_internal` one where those things should be done, since once the PeerConnection is gone (whether the session exists or not) then all relationships to room resources are gone. As a matter of fact, we have code for that already:

https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_textroom.c#L3009-L3043

which is supposed to do exactly that, that is leaving rooms you're in and destroying the participants resources. If you think that's not working as expected for some reason, please check what needs to be fixed there instead.

lminiero:
I just tried to replicate what you did, and it seems to work just fine in my case. Obviously the user only leaves once the session timeout kicks in, so ~60s later by default, but the same thing would happen with your patch too, since you're intercepting `destroy_session` which is triggered by a handle detach, so only once the session timeout gets rid of all related handles.

lminiero:
Then yes, the problem may be in the code I linked above, which is where the logic for "let's remove the user" is, so that's where I'd dig: I couldn't replicate the problem, as the session timeout cleaned the user correctly for me. It may have something to do with the "fake" message we inject to simulate a "leave" message from the user: my guess is that you do end up calling that code, and the user is removed, but the event is not propagated as it should because of something that goes wrong with that "fake" message.

adnanel:
Hi @lminiero 

I had time to work on this again, turned out your leave command failed for us due to being an invalid JSON, which in turn is invalid due to our (string) room ID being too long (it's an UUID with a suffix).

I changed the PR to a much simpler version which only expands the request string buffer to 200 characters. Will test today and confirm whether this fixes everything.

lminiero:
@adnanel apologies for the very late response, I've been abroad for the IETF meeting and I'm catching up with all the pending activities. So it was just a matter of the buffer for the "fake" message being too short? If that's the case, it's indeed a very easy fix and I'll merge!

Issue #3242, opened 2023-07-13
[1.x] Videoroom use-after-free when unsubscribing and closing a room


thatsmydoing:
**What version of Janus is this happening on?**
`6d2dd1b44d16d51c086f58f5ffc59c87a5e9b0c7`

**Have you tested a more recent version of Janus too?**
Yes. `fe993140577ff91ec626eb7b5f03a2898809674f` 

**Was this working before?**
I'm not sure, it also crashes on 1.0.4. In practice, it doesn't crash that often. Maybe once a day in our usage.

**Is there a gdb or libasan trace of the issue?**
Reproduced locally against `fe993140577ff91ec626eb7b5f03a2898809674f`, https://gist.github.com/thatsmydoing/c8e82a531f3431cb8f066e4b54cb6dec

**Additional context**
I can reproduce it pretty reliably with https://github.com/thatsmydoing/janus-gateway-2775/

lminiero:
Ah, it looks like a cleanup code stepping on its own feet: we're looping on `stream->publisher_streams`, which is a `GList`, but while iterating there (`list = list->next`), we call `janus_videoroom_subscriber_stream_remove` which updates that same list. As such, the `list` pointer we're using may not be a valid pointer anymore, because destroyed in the process, hence the crash. I'll try to remember why exactly the code was written like that, and to come up with a fix.

lminiero:
If only I could replicate the problem, though... I'm at attempt 900 with your tool and still nothing :rofl:

thatsmydoing:
Oh no. On my machine I get the error on the 1st attempt. Let me check again.

lminiero:
I stopped the tool at about 2000 attempts, as I don't think it's going to die on me at this point :slightly_smiling_face:

thatsmydoing:
I'm not sure what to say, I tried in a different environment and it's still consistently crashing on the first attempt. If you've cloned the repository before, I've pushed some changes which you may not have pulled? It also only reliably crashes with asan, otherwise it just keeps going.

Otherwise, it might be the library versions being used? I'll see if I can make docker reproduction case.

lminiero:
No, there's no further commits if I do a `git pull`, and my local Janus instance is indeed always build with libasan, since I use that to track issues I stumble upon while developing.

thatsmydoing:
[...]

PR #3247, opened 2023-08-01
Fix rare race condition in VideoRoom (see #3242)

lminiero:
As the title says, this is an attempt to fix some race conditions emerged in #3242. Please test.


lminiero:
Merging.

PR #3250, opened 2023-08-04
Extend sessions_mutex scope to avoid rogue inserts in mp viewers list (see #3246)

lminiero:
Port to `1.x` of #3246, see discussion there for context.


atoppi:
lgtm :+1:

lminiero:
Merging.

PR #3255, opened 2023-08-07
Update janus.d.ts: add missing RemoteTrackMetadata parameter and definition

veeting:
PR #3150 added a new parameter "metadata" to the onremotetrack callback function of janus.js. 

This PR adds the new parameter to the Typescript definition in janus.d.ts. The parameter is defined as an optional parameter to not break existing projects.


lminiero:
Thanks, merging!

Issue #3249, opened 2023-08-03
Janus crash with VideoRoom streams summary [1.x]


zevarito:
**What version of Janus is this happening on?**
It does happen in latest [e1c7704](https://github.com/meetecho/janus-gateway/commit/e1c7704cfdef76cd5b177ba7fbb92c041bddd40e) but it is happening since a few months now.

**Have you tested a more recent version of Janus too?**
Yes

**Was this working before?**
This same exception and a similar one (same function) I think is around since May at least.

**Is there a gdb or libasan trace of the issue?**
* [gdb stack trace of core.hloop 2.1 ](https://gist.github.com/zevarito/e26ab9c2a20d3e86ecaaa5924a5b6d1c)
* [Inspected Subscriber param](https://gist.github.com/zevarito/e26ab9c2a20d3e86ecaaa5924a5b6d1c#file-gistfile1-txt-L59)
* [plugins/janus_videoroom.c:3117](https://github.com/meetecho/janus-gateway/blob/e1c7704cfdef76cd5b177ba7fbb92c041bddd40e/src/plugins/janus_videoroom.c#L3117)
```
#1  0x00007f253827a5b8 in janus_videoroom_subscriber_streams_summary (subscriber=subscriber@entry=0x7f25086416e0, legacy=legacy@entry=0, event=event@entry=0x0)
    at plugins/janus_videoroom.c:3117
        stream = 0x7f250800e050
        ps = <optimized out>
        m = 0x7f2528591a70
        media = 0x7f2528478c10
        temp = 0x7f252803eaa0
        __FUNCTION__ = "janus_videoroom_subscriber_streams_summary"
```

**Additional context**
I've also seen a related exception that I don't have the backtrace at hand which fails in the same function but a little bit earlier and the core dump generated says `janus` instead of `hloop`.
My guess is that it is some sort of race condition when that list is generated and a publisher leaves the room at the same time.
It crash approximately after 2 weeks on every server, but I have the sense that latest build might also crash early, updated recently will let you know.

Let me know if you need me to provide any extra info and thank you very much for looking into this issue!

lminiero:
Please test with #3247 too, since it addresses a couple of race conditions in the VideoRoom.

zevarito:
Thanks @lminiero will do.

zevarito:
I know you guys are on vacation right now, I am just updating the issue to keep you in the loop when you got back.

[...]

PR #3259, opened 2023-08-10
videoroom: check participant before dereference

tmatth:
Refs #3249


lminiero:
Thanks! Looks good to me, so I'll merge.

Issue #3248, opened 2023-08-02
[1.x] SIP plugin binds to ANY network interface


pawnnail:
Have started looking at the multistream implementation of SIP gateway. Have noticed that the SIP plugin now unconditionally binds to any network interface for media. The singelstream version binds to the address specified by _local_media_ip_ (otherwise, to a single local address).

It may be a security risk since the media sockets unsuspectingly receive traffic from any network interface unrelated to the _local_media_ip_ advertised by SDP. Additionally, the procedure requires a target UDP port to be available on all interfaces that may be resource wasteful if successful or more prone to the bind failure otherwise.

The change was introduced by  [this commit](https://github.com/meetecho/janus-gateway/commit/acb137827a454f52b11a78fa7f3c16691aeecca0). I read through the comments but couldn't see a compelling reason for the change of behaviour. It may well be I am missing something.

Prior to the commit we had what the singlestream version still has:
`inet_pton(AF_INET, (local_media_ip ? local_media_ip : local_ip), &audio_rtp_address.sin_addr.s_addr);`

This has changed to 
`audio_rtp_address.sin6_addr = in6addr_any;`
and INADDR_ANY for ipv4.

Could we please review the change again and consider an appropriate correction unless there is a good reason for not to. 
Thanks

lminiero:
The SIP plugin only supported IPv4 before, now it supports both IPv4 and IPv6, and we use an IPv6 socket to bind to both if that's needed. `local_media_ip` could be either, and so would need additional and custom code to only bind to that, a bit like the Streaming plugin does. I'm not against that, of course, but to be honest I don't have time to work on that now (vacations are at the door). If you're willing to write a pull request, I'll be happy to review it though.

pawnnail:
Will do, thanks.

lminiero:
Merged the PR.

PR #3263, opened 2023-08-22
Allow SIP plugin to bind media to a specific address (see #3248)

pawnnail:
The multi-stream version of the SIP plugin no longer binds RTP/RTCP sockets to a specific address just like the single-stream version used to. Instead, the multi-stream SIP plugin binds to "any" address for IPv4 and IPv6 (if supported). This highlights a potential security risk if deployed on a multihomed host in a hostile environment. Perhaps, is also wasteful as UDP ports are allocated on multiple addresses and a change of behaviour when migrating from a single-stream to multi. The change was added by PR #3159.

I would like the SIP plugin to use a specific address, IPv4 or IPV6, for media (RTP/RTCP) while preserving the option of binding to any interface (PR #3159). The binding address changes driven by the value of configuration parameter _local_media_ip_ are summarized below.

|  _local_media_ip_ | single-stream  | multi-stream |  multi-stream proposed |
|-------------------|----------------|--------------|------------------------|
| \<empty string\>    | not supported      | not supported | any (IPv4 and IPv6)|
| \<unspecified\>     | _local_ip_ (IPv4)     | any (IPv4 and IPv6)  | any (IPv4 and IPv6) |
| "0.0.0.0"    | any (IPv4)      | any (IPv4 and IPv6) | any (IPv4 and IPv6)|
| "::"    | not supported      | any (IPv4 and IPv6) | any (IPv4 and IPv6)|
| IPv4, e.g. "192.168.171.8"    | 192.168.171.8      | any (IPv4 and IPv6)  | 192.168.171.8|
| IPv4-mapped IPv6, e.g. "::ffff:192.168.171.8"    | not supported| any (IPv4 and IPv6) | 192.168.171.8 |
| IPVv6, e.g. "2406:da1c:aa9:9401:2825:674b:3e05:fca2"| not supported  | any (IPv4 and IPv6)  | 2406:da1c:aa9:9401:2825:674b:3e05:fca2|

So there is an impact on PR #3159: if _local_media_ip_ happens to be specified then RTP/RTCP sockets will bind to that address only rather than to all. There is nothing I can do about it other than to warn. I hope the proposal is flexible enough to address most desired configuration scenarios going forward.

Have tested the above scenarios on Linux and Windows WSL.

Have made _sdp_ip_ used for the SDP connection address header (c=) to fallback to a valid value of either _local_media_ip_ or _local_ip_.

Also fixed an issue where _bind()_ intermittently fails due to _audio_rtp_address_ and _audio_rtcp_address_ not initialized to 0.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -854,6 +854,7 @@ static gboolean ipv6_disabled = FALSE;
 static janus_callbacks *gateway = NULL;
 
 static char *local_ip = NULL, *sdp_ip = NULL, *local_media_ip = NULL;
+static janus_network_address janus_network_local_media_ip = { 0 };
 static int keepalive_interval = 120;
 static gboolean behind_nat = FALSE;
 static char *user_agent;
@@ -1882,11 +1883,11 @@ int janus_sip_init(janus_callbacks *callback, const char *config_path) {
 		}
 
 		item = janus_config_get(config, config_general, janus_config_type_item, "local_media_ip");
-		if(item && item->value)
+		if(item && item->value && strlen(item->value) > 0)
 			local_media_ip = g_strdup(item->value);
 
 		item = janus_config_get(config, config_general, janus_config_type_item, "sdp_ip");
-		if(item && item->value) {
+		if(item && item->value && strlen(item->value) > 0) {
 			sdp_ip = g_strdup(item->value);
 			JANUS_LOG(LOG_VERB, "IP to advertise in SDP: %s\n", sdp_ip);
 		}
@@ -2003,6 +2004,27 @@ int janus_sip_init(janus_callbacks *callback, const char *config_path) {
 	}
 	JANUS_LOG(LOG_VERB, "Local IP set to %s\n", local_ip);
 
+	/*
+		Since we might have to derive SDP connection address from local_media_ip make sure it has a meaningful value
+		for the prupose of using it in the SDP c= header.
+	*/
+	janus_network_address_nullify(&janus_network_local_media_ip);
+	if(local_media_ip) {
+		if(janus_network_string_to_address(janus_network_query_options_any_ip, local_media_ip, &janus_network_local_media_ip) != 0) {
+			JANUS_LOG(LOG_ERR, "Invalid local media IP address [%s]...\n", local_media_ip);
+			return -1;
+		}
+		if((janus_network_local_media_ip.family == AF_INET && janus_network_local_media_ip.ipv4.s_addr == INADDR_ANY) ||
+				(janus_network_local_media_ip.family == AF_INET6 && IN6_IS_ADDR_UNSPECIFIED(&janus_network_local_media_ip.ipv6))) {
+			janus_network_address_nullify(&janus_network_local_media_ip);
+		}
+	}
+	JANUS_LOG(LOG_VERB, "Binding media address set to [%s]...\n", janus_network_address_is_null(&janus_network_local_media_ip) ? "any" : local_media_ip);
+	if(!sdp_ip) {
+		sdp_ip = janus_network_address_is_null(&janus_network_local_media_ip) ? local_ip : local_media_ip;
+		JANUS_LOG(LOG_VERB, "IP to advertise in SDP: %s\n", sdp_ip);
+	}
+
 	/* Setup sofia */
 	su_init();
 	if(notify_events && callback->events_is_enabled()) {
@@ -2022,19 +2044,24 @@ int janus_sip_init(janus_callbacks *callback, const char *config_path) {
 	/* This is the callback we'll need to invoke to contact the Janus core */
 	gateway = callback;
 
-	/* Finally, let's check if IPv6 is disabled, as we may need to know for RTP/RTCP sockets */
-	int fd = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
-	if(fd <= 0) {
-		ipv6_disabled = TRUE;
-	} else {
-		int v6only = 0;
-		if(setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, &v6only, sizeof(v6only)) != 0)
+	if(janus_network_address_is_null(&janus_network_local_media_ip)) {
+		/* Finally, let's check if IPv6 is disabled, as we may need to know for RTP/RTCP sockets */
+		int fd = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
+		if(fd <= 0) {
 			ipv6_disabled = TRUE;
-	}
-	if(fd > 0)
-		close(fd);
-	if(ipv6_disabled) {
-		JANUS_LOG(LOG_WARN, "IPv6 disabled, will only use IPv4 for RTP/RTCP sockets (SIP)\n");
+		} else {
+			int v6only = 0;
+			if(setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, &v6only, sizeof(v6only)) != 0)
+				ipv6_disabled = TRUE;
+		}
+		if(fd > 0)
+			close(fd);
+		if(ipv6_disabled) {
+			JANUS_LOG(LOG_WARN, "IPv6 disabled, will only use IPv4 for RTP/RTCP sockets (SIP)\n");
+		}
+	} else if(janus_network_local_media_ip.family == AF_INET) {
+		/* Disable if we have a specified IPv4 address for RTP/RTCP sockets */
+		ipv6_disabled = TRUE;
 	}
 
 	g_atomic_int_set(&initialized, 1);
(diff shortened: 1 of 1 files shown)


src/plugins/janus_sip.c
```diff
@@ -2003,6 +2004,27 @@ int janus_sip_init(janus_callbacks *callback, const char *config_path) {
 	}
 	JANUS_LOG(LOG_VERB, "Local IP set to %s\n", local_ip);
 
+	/*
+		Since we might have to derive SDP connection address from local_media_ip make sure it has a meaningful value
+		for the prupose of using it in the SDP c= header.
+	*/
```
lminiero, line 2010: Please don't use this format for comments. Align to what we use in other parts of the code, so something the lines of

    /* Since ...
     * for the ... */

As a side notem there's a typo in "purpose".

src/plugins/janus_sip.c
```diff
@@ -6625,14 +6652,27 @@ static int janus_sip_allocate_local_ports(janus_sip_session *session, gboolean u
 			session->media.pipefd[1] = -1;
 		}
 	}
+	/* Determine the address family to use */
+	if(ipv6_disabled &&
+		!janus_network_address_is_null(&janus_network_local_media_ip) &&
+		janus_network_local_media_ip.family == AF_INET6) {
```
lminiero, line 6658: Code style: you need another tab of indent for the two additional checks above, otherwise it ends up lined with the action below.

src/plugins/janus_sip.c
```diff
@@ -6625,14 +6652,27 @@ static int janus_sip_allocate_local_ports(janus_sip_session *session, gboolean u
...
+		JANUS_LOG(LOG_ERR, "Error setting address for media sockets, IPv6 is disabled and local media address is IPv6...\n");
```
lminiero, line 6659: I don't like this action failing at runtime. It looks like a check that can be done when thew plugin is loaded, that is after the `janus_network_local_media_ip` property is filled in and we know if `ipv6_disabled` is `FALSE`. Better to have the plugin not load at all, if we already know it will never allocate sockets for media.

src/plugins/janus_sip.c
```diff
@@ -6625,14 +6652,27 @@ static int janus_sip_allocate_local_ports(janus_sip_session *session, gboolean u
...
+		JANUS_LOG(LOG_VERB, "Allocating audio ports using address [%s]\n",
+			janus_network_address_is_null(&janus_network_local_media_ip) ? "any" : local_media_ip);
 		struct sockaddr_storage audio_rtp_address, audio_rtcp_address;
 		while(session->media.local_audio_rtp_port == 0 || session->media.local_audio_rtcp_port == 0) {
 			if(attempts == 0)	/* Too many failures */
 				return -1;
+			memset(&audio_rtp_address, 0, sizeof(audio_rtp_address));
+			memset(&audio_rtcp_address, 0, sizeof(audio_rtcp_address));
 			if(session->media.audio_rtp_fd == -1) {
 				session->media.audio_rtp_fd = socket(!ipv6_disabled ? AF_INET6 : AF_INET, SOCK_DGRAM, 0);
```
lminiero, line 6677: Did you switch to `use_ipv6_address_family` here? (and when creating the RTCP socket too, a few lines below). There's other checks that now use `use_ipv6_address_family` in this code block that should probably move to the other boolean, since it's that one that now dictates the addrlen.

src/plugins/janus_sip.c
```diff
@@ -6708,17 +6749,20 @@ static int janus_sip_allocate_local_ports(janus_sip_session *session, gboolean u
...
+		JANUS_LOG(LOG_VERB, "Allocating video ports using address [%s]\n",
+			janus_network_address_is_null(&janus_network_local_media_ip  ) ?"any" : local_media_ip);
 		struct sockaddr_storage video_rtp_address, video_rtcp_address;
 		while(session->media.local_video_rtp_port == 0 || session->media.local_video_rtcp_port == 0) {
 			if(attempts == 0)	/* Too many failures */
 				return -1;
+			memset(&video_rtp_address, 0, sizeof(video_rtp_address));
+			memset(&video_rtcp_address, 0, sizeof(video_rtcp_address));
 			if(session->media.video_rtp_fd == -1) {
 				session->media.video_rtp_fd = socket(!ipv6_disabled ? AF_INET6 : AF_INET, SOCK_DGRAM, 0);
```
lminiero, line 6767: Same here (including RTCP and related checks).

lminiero (commented):
Added some notes inline.

pawnnail:
Thanks for the review, Lorenzo. Will get it sorted within the next few days.

src/plugins/janus_sip.c
pawnnail: The point taken.

src/plugins/janus_sip.c
pawnnail: Point taken. The condition is removed as redundant.

src/plugins/janus_sip.c
pawnnail: Makes sense. Moved the failure to the config loading time.

src/plugins/janus_sip.c
pawnnail: Essentially `!ipv6_disabled` and `use_ipv6_address_family` have the same effect, replaced one with the other for clarity.

src/plugins/janus_sip.c
pawnnail: Also done.

pawnnail:
There are few more things I would like to do to tidy up port binding but not in the same PR. Arbitrary `int attempts = 100` tends to fail often on a busy system, especially in a limited port range especially considering RTPC ports need to be adjacent to RTP port. And then it only makes sense to retry  `bind()` on transient errors like hitting a busy port. But one thing at a time.

Retested again. Haven't done the rebase yet, will do once all the changes are approved.
Thanks.

src/plugins/janus_sip.c
lminiero: The format of the comment is still incorrect.

pawnnail: Sorry about that. Fixed.

lminiero:
Apologies for the lack of feedback, I've been pretty busy lately and this is an effort I need to study well, especially considering that, if it works, I'll need to port it to the NoSIP plugin as well, besides backporting it to `0.x` for both plugins. I'll try to reserve some time later this week to have a deeper look and make some tests. Scream at me if it doesn't happen :laughing:

pawnnail:
No problems and thanks. I am not in a hurry. Let me know if I can help.

lminiero:
I'm adapting the patch to 0.x, and then I'll port the changes to NoSIP too. If all goes well, I'll try to merge later today. Apologies again for the delay!

src/plugins/janus_sip.c
```diff
@@ -2003,6 +2004,25 @@ int janus_sip_init(janus_callbacks *callback, const char *config_path) {
...
+			return -1;
+		}
+		if((janus_network_local_media_ip.family == AF_INET && janus_network_local_media_ip.ipv4.s_addr == INADDR_ANY) ||
+				(janus_network_local_media_ip.family == AF_INET6 && IN6_IS_ADDR_UNSPECIFIED(&janus_network_local_media_ip.ipv6))) {
+			janus_network_address_nullify(&janus_network_local_media_ip);
+		}
+	}
+	JANUS_LOG(LOG_VERB, "Binding media address set to [%s]...\n", janus_network_address_is_null(&janus_network_local_media_ip) ? "any" : local_media_ip);
+	if(!sdp_ip) {
+		sdp_ip = janus_network_address_is_null(&janus_network_local_media_ip) ? local_ip : local_media_ip;
```
lminiero, line 2022: This must be a `g_strdup`, or the plugin will crash at shutdown. In fact, you're simply assinging to `sdp_ip` the value of either `local_ip` or `local_media_ip`, but at the end we do a `g_free` for all of them:

	g_free(local_ip);
	g_free(local_media_ip);
	g_free(sdp_ip);

which means that this will cause a double free on the same pointer.

src/plugins/janus_sip.c
```diff
@@ -6708,21 +6745,24 @@ static int janus_sip_allocate_local_ports(janus_sip_session *session, gboolean u
...
-			JANUS_LOG(LOG_VERB, "Audio RTCP listener bound to %s:%d(%d)\n", (local_media_ip ? local_media_ip : local_ip), rtcp_port, session->media.audio_rtcp_fd);
+			JANUS_LOG(LOG_VERB, "Audio RTCP listener bound to [%s]:%d(%d)\n", janus_network_address_is_null(&janus_network_local_media_ip) ?"any":local_media_ip, rtcp_port, session->media.audio_rtcp_fd);
 			session->media.local_audio_rtp_port = rtp_port;
 			session->media.local_audio_rtcp_port = rtcp_port;
 		}
 	}
 	if(session->media.has_video) {
-		JANUS_LOG(LOG_VERB, "Allocating video ports:\n");
+		JANUS_LOG(LOG_VERB, "Allocating video ports using address [%s]\n",
+			janus_network_address_is_null(&janus_network_local_media_ip  ) ?"any" : local_media_ip);
```
lminiero, line 6755: There are some weird spaces in here. Per the code style, it should be

    janus_network_address_is_null(&janus_network_local_media_ip) ? "any" : local_media_ip);

lminiero (commented):
@pawnnail while updating other branches and plugins, I stumbled upon these two things. One is a bug and needs to be fixed (it will cause a crash), the other is simply a code style nit. Thanks!

lminiero:
I'll merge and fix the remaining nits myself, since this has been open quite some time. Thanks again for the contribution and the patience!

pawnnail:
Thanks!

src/plugins/janus_sip.c
pawnnail: Good point, thank you for catching it.

PR #3266, opened 2023-08-28
Update PluginDataParam type definitions to include text label and pro…

veeting:
We have identified that PluginDataParam has a few missing type definitions in janus.d.ts

- text
- label as optional
- protocol as optional

NOTE: 

janus.js is either using **text** or **data** which is confusing. The html demos are only referring to **text**.
Thus, we have decided to not add **data** and add **text** as optional to avoid breaking any other app.

### DIFF
--- a/npm/janus.d.ts
+++ b/npm/janus.d.ts
@@ -317,6 +317,9 @@ declare namespace JanusJS {
 	}
 
 	type PluginDataParam = {
+		text?: string;
+		label?: string;
+		protocol?: string;
 		success?: (data: unknown) => void;
 		error?: (error: string) => void;
 	}


lminiero:
> janus.js is either using **text** or **data** which is confusing. The html demos are only referring to **text**.
> Thus, we have decided to not add **data** and add **text** as optional to avoid breaking any other app.

I'd rather both are kept in there, with the idea only `data` will remain someday. `text` is legacy from when we _only_ supported text data, which is not the case anymore: we support binary data too. The fact that demos only refer to that is simply a conseguence of those legacy assumptions.

lminiero:
Thanks for the fix, merging!

PR #3267, opened 2023-08-28
Update janus.js to allow a flexible usage of insertable streams and e2ee

veeting:
This PR will allow a more flexible usage of insertable streams and e2ee.

When using **externalEncryption** in offer or answer janus.js will activate "encodedInsertableStreams" and will use **e2ee: true** parameter, even if encodedInsertableStreams is not supported and even if "transforms" callbacks are not provided.

The change also allows developers to handle the encryption of the streams (either with createEncodedStreams or RTCRtpScriptTransform) outside of janus.js.

### DIFF
--- a/html/janus.js
+++ b/html/janus.js
@@ -1889,6 +1889,12 @@ function Janus(gatewayCallbacks) {
 				}
 			}
 		}
+
+		if (callbacks.externalEncryption) {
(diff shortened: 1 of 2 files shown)


lminiero:
Thanks, merging then!

PR #3274, opened 2023-10-05
Fix pp-rec post-processing for h265

heyfluke:
All complete nalu except FU/AP should be convert back from rtp-packed format to original ones which are separated by delimiters, though I only encountered the type==1 (HEVC_NAL_TRAIL_R) so only added this type.

### DIFF
--- a/src/postprocessing/pp-h265.c
+++ b/src/postprocessing/pp-h265.c
@@ -513,7 +513,7 @@ int janus_pp_h265_process(FILE *file, janus_pp_frame_packet *list, int *working)
 			memcpy(&unit, buffer, sizeof(uint16_t));
 			unit = ntohs(unit);
 			uint8_t type = (unit & 0x7E00) >> 9;
-			if(type == 32 || type == 33 || type == 34) {
+			if(type == 32 || type == 33 || type == 34 || type == 1) {
 				if(type == 32 || type == 33) {
 					keyFrame = 1;
 					if(!keyframe_found) {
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @heyfluke! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

src/postprocessing/pp-h265.c
```diff
@@ -582,6 +582,8 @@ int janus_pp_h265_process(FILE *file, janus_pp_frame_packet *list, int *working)
 				/* Skip the FU header */
 				buffer += 3;
 				len -= 3;
+			} else {
+				JANUS_LOG(LOG_HUGE, "unhandled nalu type: %d\n", type);
```
lminiero, line 586: I don't think this else is needed here, there's not anything we do with it anyway.

lminiero (commented):
Sorry about the late feedback. As I said in response to another post, I personally never use H.265 so I have no way to test this. From what I can see, this only ensures we add a NAL delimiter for that type 1, since we still won't use it for anything specifically (we call it a keyframe only for 32/33). This is probably to make the decoder happy, which explains why it wouldn't play before? Not sure if this means we'll have to add the NAL delimiter for all typs except 48, so more simply add an `else` (where you now added the log line) that only implements the "Add the NAL delimiter" action. This would ensure we add the NAL delimiter for all types, but I have no idea if this can cause issues.

Anyway, if you can confirm this simple fix does it for you and doesn't break other H.265 recordings, I think it's safe to merge.

src/postprocessing/pp-h265.c
lminiero: I'll merge and remove this myself. Thanks!

PR #3279, opened 2023-10-15
Update versions of external css and js resources

bkmgit:


lminiero:
Sorry for the late feedback. This PR is not only updating the versions, it's changing the source of those JS files as well, which I personally don't like and don't want to go through. Any reason why you did that?

bkmgit:
The second source also publishes hashes. But can calculate these using
https://www.srihash.org/
Described at:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
If updating version is ok, can change source back to what it was.

lminiero:
> If updating version is ok, can change source back to what it was.

For the time being I'd prefer that, yes. The change of source is something that can be discussed in a separate PR.

That said, I'm not sure how safe updating jQuery is, especially when building the documentation, as I don't know if there's things Doxygen relies upon that may be affected by breaking changes.

You're also updating the version of Bootstrap to 5.x, which is wrong. Bootstrap 5 is completely incompatible with the templates we use, which uses Bootstrap 3. A few times I've tried working on upgrading our demos to Bootstrap 4 or 5, and always failed miserably, because there's too much to change and I always got stuck in some places.

Please make sure to only submit commits that upgrade versions that you've explicitly tested yourself, for all demos.

lminiero:
@bkmgit any update on this? Thanks!

bkmgit:
Thanks for your comments. Sorry for the delay.  Will check with a deployment.

bkmgit:
Closing in favor of https://github.com/meetecho/janus-gateway/pull/3295

PR #3280, opened 2023-10-16
SipPlugin read rfc2833 dtmf and convert to sip info event

ywmoyue:
Implement the features proposed by jkmchinese from https://janus.discourse.group/t/audiobridge-with-sip-participants/131/7

this PR get rfc2833 data from rtp stream and assemble a sip-info event push to user

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -868,13 +868,13 @@ static int dscp_video_rtp = 0;
 static char *sips_certs_dir = NULL;
 #define JANUS_DEFAULT_SIP_TIMER_T1X64 32000
 static int sip_timer_t1x64 = JANUS_DEFAULT_SIP_TIMER_T1X64;
+static uint16_t dtmf_keys[] = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '*', '#', 'A', 'B', 'C', 'D'};
 
 static gboolean query_contact_header = FALSE;
 
 static GThread *handler_thread;
 static void *janus_sip_handler(void *data);
 static void janus_sip_hangup_media_internal(janus_plugin_session *handle);
-
 
 typedef struct janus_sip_message {
 	janus_plugin_session *handle;
@@ -1085,7 +1085,7 @@ typedef struct janus_sip_session {
 	GList *incoming_header_prefixes;
 	GList *active_calls;
 	janus_refcount ref;
-	janus_sip_dtmf *latest_dtmf;
+	janus_sip_dtmf latest_dtmf;
 } janus_sip_session;
 
 static GHashTable *sessions;
@@ -7404,39 +7404,32 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 /* Check peer RTP has RFC2833 and push event */
 static void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer, int len) {
 	janus_rtp_header *rtp_header = (janus_rtp_header *)buffer;
-	if (rtp_header->type != JANUS_RTP_RFC2833_DEFAULT_PAYLOAD_TYPE)
+	if(rtp_header->type != JANUS_RTP_RFC2833_DEFAULT_PAYLOAD_TYPE)
 		return;
 	int plen = 0;
 	char *payload_buffer = janus_rtp_payload(buffer, len, &plen);
-	int rfc2833_payload_size = sizeof(janus_rtp_rfc2833_payload);
-	if(plen < rfc2833_payload_size)
+	if(plen < sizeof(janus_rtp_rfc2833_payload))
 		return;
 	janus_rtp_rfc2833_payload *rfc2833_payload = (janus_rtp_rfc2833_payload *)payload_buffer;
 	uint16_t duration = ntohs(rfc2833_payload->duration);
-	if (rfc2833_payload->end == 0)
+	if(rfc2833_payload->end == 0)
 		return;
 
 	/* Set up last dtmf to avoid duplication */
-	if(session->latest_dtmf != NULL && session->latest_dtmf->dtmf_event_id == rfc2833_payload->event && session->latest_dtmf->timestamp == rtp_header->timestamp)
+	if(session->latest_dtmf.dtmf_event_id == rfc2833_payload->event && session->latest_dtmf.timestamp == rtp_header->timestamp)
 		return;
-	if(session->latest_dtmf != NULL){
-		g_free(session->latest_dtmf);
-		session->latest_dtmf = NULL;
-	}
-	session->latest_dtmf = g_malloc0(sizeof(janus_sip_dtmf));
-	session->latest_dtmf->dtmf_event_id = rfc2833_payload->event;
-	session->latest_dtmf->timestamp = rtp_header->timestamp;
+	session->latest_dtmf.dtmf_event_id = rfc2833_payload->event;
+	session->latest_dtmf.timestamp = rtp_header->timestamp;
 
 	/* Parse dtmf key */ 
 	uint16_t dtmf_key;
 	uint16_t dtmf_keys[] = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '*', '#', 'A', 'B', 'C', 'D'};
-	if (rfc2833_payload->event < 0 || rfc2833_payload->event > 15) {
-  		dtmf_key = '?';
-	} else {
-  		dtmf_key = dtmf_keys[rfc2833_payload->event];
-	}
+	if(rfc2833_payload->event < 0 || rfc2833_payload->event > 15)
+		return;
+  	dtmf_key = dtmf_keys[rfc2833_payload->event];
 	char dtmf_key_str[2];
-	g_snprintf(dtmf_key_str, sizeof(dtmf_key_str), "%c", dtmf_key);
+	dtmf_key_str[0] = dtmf_key;
+	dtmf_key_str[1] = '\0';
 
 	/* Notify the application */
 	json_t *info = json_object();
@@ -7446,7 +7439,7 @@ static void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer, in
 	json_object_set_new(result, "sender", json_string(session->callee));
 	json_object_set_new(result, "signal", json_string(dtmf_key_str));
 	json_object_set_new(result, "duration", json_integer(duration));
-	if (session->callid)
+	if(session->callid)
 		json_object_set_new(info, "call_id", json_string(session->callid));
 	json_object_set_new(info, "result", result);
 	int ret = gateway->push_event(session->handle, &janus_sip_plugin, session->transaction, info, NULL);
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -7408,7 +7408,7 @@ static void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer, in
 		return;
 	int plen = 0;
 	char *payload_buffer = janus_rtp_payload(buffer, len, &plen);
-	if(plen < sizeof(janus_rtp_rfc2833_payload))
+	if(plen < 0 || (size_t)plen < sizeof(janus_rtp_rfc2833_payload))
 		return;
 	janus_rtp_rfc2833_payload *rfc2833_payload = (janus_rtp_rfc2833_payload *)payload_buffer;
 	uint16_t duration = ntohs(rfc2833_payload->duration);
@@ -7423,7 +7423,7 @@ static void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer, in
 
 	/* Parse dtmf key */ 
 	uint16_t dtmf_key;
-	if(rfc2833_payload->event < 0 || rfc2833_payload->event > 15)
+	if(rfc2833_payload->event > 15)
 		return;
   	dtmf_key = dtmf_keys[rfc2833_payload->event];
 	char dtmf_key_str[2];
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -875,6 +875,9 @@ static GThread *handler_thread;
 static void *janus_sip_handler(void *data);
 static void janus_sip_hangup_media_internal(janus_plugin_session *handle);
 
+#define JANUS_RTP_HEADER_LENGTH 12
+#define JANUS_RTP_RFC2833_DEFAULT_PAYLOAD_TYPE 101
+
 typedef struct janus_sip_message {
 	janus_plugin_session *handle;
 	char *transaction;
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @ywmoyue! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

ywmoyue:
> Thanks for your contribution, @ywmoyue! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

I have signed

src/plugins/janus_sip.c
```diff
@@ -875,6 +875,9 @@ static GThread *handler_thread;
 static void *janus_sip_handler(void *data);
 static void janus_sip_hangup_media_internal(janus_plugin_session *handle);
 
+#define JANUS_RTP_HEADER_LENGTH 12
```
lminiero, line 878: This is not needed, we have a define for that already in `rtp.h`.

src/plugins/janus_sip.c
```diff
@@ -1047,6 +1050,27 @@ typedef struct janus_sip_media {
...
+	uint8_t volume : 6;
+	uint16_t duration;
+#elif __BYTE_ORDER == __LITTLE_ENDIAN
+	uint8_t event;
+	uint8_t volume : 6;
+	uint8_t reserved : 1;
+	uint8_t end : 1;
+	uint16_t duration;
+#endif
+} janus_rtp_rfc2833_payload;
```
lminiero, line 1067: I think this struct belongs to `rtp.h`, where we define other RTP and RTCP related properties. This will allow other plugins interested in using that payload to refer to an existing structure.

As a side note, for code style don't put any spaces in the size specifier (so it should be, e.g., `volume:6` and not `volume : 6`).

src/plugins/janus_sip.c
```diff
@@ -1512,7 +1538,7 @@ char *janus_sip_sdp_manipulate(janus_sip_session *session, janus_sdp *sdp, gbool
 static int janus_sip_allocate_local_ports(janus_sip_session *session, gboolean update);
 static void *janus_sip_relay_thread(void *data);
 static void janus_sip_media_cleanup(janus_sip_session *session);
-
+void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer);
```
lminiero, line 1541: This should be `static`.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7419,56 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 	return NULL;
 }
 
+/* check peer rtp has rfc2833 and push event */
```
lminiero, line 7422: Code style: please start comments with a capitalized letter. Besides, both "rtp" and rfc" should be all caps. I see other comments that should be fixed the same way.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7419,56 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer) {
+	janus_rtp_header *rtp_header = (janus_rtp_header *)buffer;
+	if (rtp_header->type != JANUS_RTP_RFC2833_DEFAULT_PAYLOAD_TYPE)
+		return;
+	char *payload_buffer = buffer + JANUS_RTP_HEADER_LENGTH + rtp_header->csrccount * sizeof(uint32_t);
```
lminiero, line 7427: This is wrong, and doesn't skip extensions for instance, if present. There's a `janus_rtp_payload` function in `rtp.h` that does exactly all that for you. Please refer to how it's used in this and other plugins to see how to use it.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7419,56 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+	janus_rtp_rfc2833_payload *rfc2833_payload = (janus_rtp_rfc2833_payload *)payload_buffer;
```
lminiero, line 7428: You're not checking if the payload is large enough to allow for the casting. If it isn't, it will crash Janus.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7419,56 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+	/* set up cache to avoid duplication */
+	if(session->dtmf_cache != NULL && session->dtmf_cache->dtmf_event_id == rfc2833_payload->event && session->dtmf_cache->timestamp == rtp_header->timestamp)
+		return;
+	if(session->dtmf_cache != NULL){
+		g_free(session->dtmf_cache);
+		session->dtmf_cache = NULL;
+	}
+	session->dtmf_cache = g_malloc0(sizeof(janus_sip_dtmf_cache));
+	session->dtmf_cache->dtmf_event_id = rfc2833_payload->event;
+	session->dtmf_cache->timestamp = rtp_header->timestamp;
```
lminiero, line 7442: I don't understand why it's called a cache, and why it's allocated to be stored, especially since you're firing an event right away, which means you won't actually need or read it later on.  I see you mention avoiding duplication, but in that case you don't need an allocation: just add the struct as it is there, and update it's fields. Per code style, it's more apt to call it "latest_dtmf", since cache is misleading as a name.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7419,56 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+	json_t *info = json_object();
+	json_object_set_new(info, "sip", json_string("event"));
+	json_t *result = json_object();
+	json_object_set_new(result, "event", json_string("info"));
+	json_object_set_new(result, "sender", json_string(session->callee));
+	json_object_set_new(result, "type", json_string("application/dtmf-relay"));
+	json_object_set_new(result, "content", json_string(payload));
+	if (session->callid)
+		json_object_set_new(info, "call_id", json_string(session->callid));
+	json_object_set_new(info, "result", result);
```
lminiero, line 7465: I don't like this simulates an info, since this is not a SIP INFO, and may confuse applications into thinking it's supported when it isn't. Better to create a new event (nothing bad about it), e.g., called `dtmf`, where you specify those fields (duration, and key) as separate properties that can be read and parsed much more easily than having to process a crafted string.

lminiero (commented):
Apologies for the late feedback. I've added some notes inline.

ywmoyue:
@lminiero Thank you very much for your review. I will revise and self-test and submit the code again next week.

src/plugins/janus_sip.c
ywmoyue: I removed this line in my last commit.

src/plugins/janus_sip.c
ywmoyue: I moved this struct definition to rtp.h in my last commit.

src/plugins/janus_sip.c
ywmoyue: I fixed it.

src/plugins/janus_sip.c
ywmoyue: I fixed it.

src/plugins/janus_sip.c
ywmoyue: In my last commit, I used the janus_rtp_payload function refer to how used in other plugins, but I don't know if it is correct, works fine in my case. please review it

src/plugins/janus_sip.c
ywmoyue: I'm sorry for my poor English. I may not understand what you mean. In my last commit, I changed the variable name and the parameter name in session to ”latest_dtmf“.

ywmoyue:
@lminiero I have accepted your comments and made the changes, can you please review this PR again?

src/plugins/janus_sip.c
```diff
@@ -875,6 +875,7 @@ static GThread *handler_thread;
 static void *janus_sip_handler(void *data);
 static void janus_sip_hangup_media_internal(janus_plugin_session *handle);
 
+
```
lminiero, line 878: There's an extra empty line here that's unneeded.

src/plugins/janus_sip.c
```diff
@@ -1079,7 +1085,9 @@ typedef struct janus_sip_session {
 	GList *incoming_header_prefixes;
 	GList *active_calls;
 	janus_refcount ref;
+	janus_sip_dtmf *latest_dtmf;
```
lminiero, line 1088: This is still a pointer, instead of just an instance of the struct.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7401,60 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
 	return NULL;
 }
 
+/* Check peer RTP has RFC2833 and push event */
+static void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer, int len) {
+	janus_rtp_header *rtp_header = (janus_rtp_header *)buffer;
+	if (rtp_header->type != JANUS_RTP_RFC2833_DEFAULT_PAYLOAD_TYPE)
```
lminiero, line 7407: Code style nit: please don't add any space between the `if` and the `(`. There's a few other occurrences of this in your patch.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7401,60 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+		return;
+
+	/* Set up last dtmf to avoid duplication */
+	if(session->latest_dtmf != NULL && session->latest_dtmf->dtmf_event_id == rfc2833_payload->event && session->latest_dtmf->timestamp == rtp_header->timestamp)
+		return;
+	if(session->latest_dtmf != NULL){
+		g_free(session->latest_dtmf);
+		session->latest_dtmf = NULL;
+	}
+	session->latest_dtmf = g_malloc0(sizeof(janus_sip_dtmf));
```
lminiero, line 7426: See note above: not sure why this needs to be allocated, rather than just using an instance of the struct as part of the session.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7401,60 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+	session->latest_dtmf->dtmf_event_id = rfc2833_payload->event;
+	session->latest_dtmf->timestamp = rtp_header->timestamp;
+
+	/* Parse dtmf key */ 
+	uint16_t dtmf_key;
+	uint16_t dtmf_keys[] = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '*', '#', 'A', 'B', 'C', 'D'};
```
lminiero, line 7432: I guess this could be a static array defined as part of the plugin globals: no need to redefine it any time we invoke this function, those values never change.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7401,60 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+	if (rfc2833_payload->event < 0 || rfc2833_payload->event > 15) {
+  		dtmf_key = '?';
```
lminiero, line 7434: Can't we just ignore the event if the value is invalid? Not sure what the point of relaying a `?` would be.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7401,60 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+	} else {
+  		dtmf_key = dtmf_keys[rfc2833_payload->event];
+	}
+	char dtmf_key_str[2];
+	g_snprintf(dtmf_key_str, sizeof(dtmf_key_str), "%c", dtmf_key);
```
lminiero, line 7439: A `g_snprintf` is probably unneeded, since we're working with chars. A simpler approach may be something like:

    dtmf_key_str[0] = dtmf_key;
    dtmf_key_str[1] = '\0';

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7401,60 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
 
+/* Check peer RTP has RFC2833 and push event */
+static void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer, int len) {
+	janus_rtp_header *rtp_header = (janus_rtp_header *)buffer;
+	if (rtp_header->type != JANUS_RTP_RFC2833_DEFAULT_PAYLOAD_TYPE)
+		return;
+	int plen = 0;
+	char *payload_buffer = janus_rtp_payload(buffer, len, &plen);
+	int rfc2833_payload_size = sizeof(janus_rtp_rfc2833_payload);
+	if(plen < rfc2833_payload_size)
```
lminiero, line 7412: Why not simply do this?

    if(plen < sizeof(janus_rtp_rfc2833_payload))

I don't think there's any need to assign the value to a new `rfc2833_payload_size` variable, since we don't use it anywhere else.

lminiero (commented):
Added a few more comments inline.

ywmoyue:
@lminiero Thanks for the review again.  I have accepted your comments and made the changes.

lminiero:
Thanks! I'll see if I can test this somehow.

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7401,52 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
 }
 
+/* Check peer RTP has RFC2833 and push event */
+static void janus_sip_check_rfc2833(janus_sip_session *session, char *buffer, int len) {
+	janus_rtp_header *rtp_header = (janus_rtp_header *)buffer;
+	if(rtp_header->type != JANUS_RTP_RFC2833_DEFAULT_PAYLOAD_TYPE)
+		return;
+	int plen = 0;
+	char *payload_buffer = janus_rtp_payload(buffer, len, &plen);
+	if(plen < sizeof(janus_rtp_rfc2833_payload))
```
lminiero, line 7411: The compiler compains about the different signedness, here, since plen is a signed integer, while sizeof returns an unsigned (`size_t`). I guess a possible fix could be something like this

	if(plen < 0 || (size_t)plen < sizeof(janus_rtp_rfc2833_payload))

src/plugins/janus_sip.c
```diff
@@ -7392,6 +7401,52 @@ gpointer janus_sip_sofia_thread(gpointer user_data) {
...
+
+	/* Set up last dtmf to avoid duplication */
+	if(session->latest_dtmf.dtmf_event_id == rfc2833_payload->event && session->latest_dtmf.timestamp == rtp_header->timestamp)
+		return;
+	session->latest_dtmf.dtmf_event_id = rfc2833_payload->event;
+	session->latest_dtmf.timestamp = rtp_header->timestamp;
+
+	/* Parse dtmf key */ 
+	uint16_t dtmf_key;
+	if(rfc2833_payload->event < 0 || rfc2833_payload->event > 15)
```
lminiero, line 7426: There's another warning here, cause by the fact that, since event is unsigned, it can never be < 0, so that part of the check should be removed.

lminiero (commented):
Just tested this and it seems to be working nicely, thanks! I'm just reporting a couple of warnings my compiler gave: as soon as you can address those, this is good to merge for me.

ywmoyue:
I fixed the two errors, can you please check again

lminiero:
Thanks, merging then! :+1:

ycherniavskyi:
@ywmoyue, I'm curious about the choice to set a hardcoded payload type of `101` for the `telephone-event` RTP payload format. According to [RFC2833](https://datatracker.ietf.org/doc/html/rfc2833), this type is typically assigned dynamically. Could you elucidate the reasons for opting for a hardcoded value instead of retrieving the actual payload type dynamically from the SDP?

My concern stems from potential compatibility issues, such as with SIP calls initiated by the Janus SIP plugin. It appears that the [WebRTC library](https://webrtc.googlesource.com/src/+/refs/heads/main/media/engine/payload_type_mapper.cc#81) used by browsers uses different (and currently hardcoded) payload types for `telephone-event`, which could lead to functional mismatches

ywmoyue:
Yes, you are right to concern 

The reason why I use 101 as the telephone-event payload is that the terminals I tested here use 101 as the telephone-event payload by default, except for webrtc, including yealink phone, linphone, real android phone with VOLTE, and others soft phone

The correct way is indeed to get the telephone-event payload value from sdp, I will see if I can submit a PR to fix this later

PR #3282, opened 2023-10-24
Support larger values for SDP attributes

petarminchev:
Currently the limit is 512 bytes for the SDP attribute value, which limits to maximum 75 video + audio mids (150 mids).
Increased limit to 2048 bytes, which should support up to around 267 video + audio mids (534 mids)

### DIFF
--- a/src/sdp-utils.c
+++ b/src/sdp-utils.c
@@ -1119,9 +1119,9 @@ char *janus_sdp_write(janus_sdp *imported) {
 	if(!imported)
 		return NULL;
 	janus_refcount_increase(&imported->ref);
-	char *sdp = g_malloc(4096), mline[8192], buffer[2048];
+	char *sdp = g_malloc(2560), mline[8192], buffer[2048];
 	*sdp = '\0';
-	size_t sdplen = 4096, mlen = sizeof(mline), offset = 0, moffset = 0;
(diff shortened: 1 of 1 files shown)


src/sdp-utils.c
```diff
@@ -1119,7 +1119,7 @@ char *janus_sdp_write(janus_sdp *imported) {
 	if(!imported)
 		return NULL;
 	janus_refcount_increase(&imported->ref);
-	char *sdp = g_malloc(1024), mline[8192], buffer[512];
+	char *sdp = g_malloc(1024), mline[8192], buffer[2048];
```
lminiero, line 1122: If the buffer is this large, now, then I guess the `sdp` allocation we start from should be larger too? We only reallocate it when appending m-lines, which means it should be large enough to handle global attributes before that.

lminiero (commented):
Adding a note inline.

src/sdp-utils.c
lminiero: Mh I think it might be overkill, the vast majority of SDPs are never that large: I think we started from a lower value on purpose. I'll give this some thought and get back to you later on.

lminiero: To be more conservative, maybe let's keep the same diff that was before, so 2048+512=2560? It's still large, but not THAT large.

lminiero:
Thanks! I think this is good to merge :ok:

Issue #2973, opened 2022-05-10
[1.x] missing SIP-Contact-header on SUBSCRIBE-messages


doerner:
**What version of Janus is this happening on?**
Janus 1.0.1-r0 on alpinelinux

**Have you tested a more recent version of Janus too?**
no

**Was this working before?**
no

**Is there a gdb or libasan trace of the issue?**
no

**Additional context**
SIP-SUBSCRUBE messages (using the sip-plugin) do not contain the mandatory Contact-header.

lminiero:
Sounds like something that should be fixed in Sofia SIP, if needed, not Janus, as we don't control the SIP logic. Please try a more recent version of Sofia SIP, if you have 1.2.11 installed, and reconfigure/recompile Janus.

lminiero:
@doerner did you have a chance to try a more recent version of Sofia SIP as suggested?

doerner:
The alpine Linux distribution uses by standard a more recent Sofia SIP package (1.13.7). This is as far as I know the current version.

lminiero:
Then I'm afraid there's nothing we can do about that. Setting the Contact header is up to the stack when you use NUA.

lminiero:
This should now be fixed, closing.

PR #3286, opened 2023-10-29
SIP plugin: add missing SIP Contact header in re-INVITE request

ycherniavskyi:
Upon transitioning to a newer version of Sofia SIP (version >= 1.13, inferred from the `query_contact_header` variable initialization comment), we began encountering errors from certain soft-switches, such as "400 Syntax Check For Dialog Failed: ('header', 'contact')" when executing the "hold" request.

During my investigation, I came across #2708 and #2973 which address this type of issue, albeit not comprehensively.

The first commit in this MR rectifies the "hold" request issue. Following that, the second commit applies the same technique to all `nua_invite` function calls proactively and also carries out minor refactoring to ensure code consistency.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -3396,7 +3396,7 @@ static void *janus_sip_handler(void *data) {
 				SIPTAG_TO_STR(to),
 				SIPTAG_EVENT_STR(event_type),
 				SIPTAG_CALL_ID_STR(callid),
-				TAG_IF(contact_header != NULL, SIPTAG_CONTACT_STR(contact_header),
+				TAG_IF(contact_header != NULL, SIPTAG_CONTACT_STR(contact_header)),
 				SIPTAG_ACCEPT_STR(accept),
 				SIPTAG_EXPIRES_STR(ttl_text),
 				NUTAG_PROXY(session->helper && session->master ?
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -4442,9 +4442,12 @@ static void *janus_sip_handler(void *data) {
 				char custom_headers[2048];
 				janus_sip_parse_custom_headers(root, (char *)&custom_headers, sizeof(custom_headers));
 
+				/* Check if we need to manually add the Contact header */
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks! I'm not entirely sure about the changes. What I noticed is that we seem to in some cases (correctly) refer to the different purpose of handles (master vs helper), while in others we just assume the handle owns the stack without taking into account it might be a helper (and so not own a SIP stack instance of its own). I guess you're not using helpers and so never encountered a problem, but it's something that would probably need to be taken into account as part of this PR (especially considering you're adding this to more requests).

ycherniavskyi:
Understood.

I propose we divide this PR into two distinct segments:
1. Inclusion of the `Contact` header: According to "Table 2: Summary of header fields" in [RFC3261](https://datatracker.ietf.org/doc/html/rfc3261) (page 161), the `Contact` header is a required header for each `nua_invite`. If this interpretation aligns with your understanding, we can proceed to the next point.
2. The logic for constructing the `contact_header` variable differs between `nua_subscribe` and `nua_invite`: The current PR adopts the existing approach from `nua_invite` for consistency. However, if the consensus is to utilize the method of creating `contact_header` akin to what's used for `nua_subscribe`, then I will introduce a helper function to harmonize its construction across all usage places.

ycherniavskyi:
@lminiero, could you kindly advise on any further evidence or explanations required to facilitate the acceptance of this PR?

lminiero:
I've been busy at the IETF meeting, and I'm only now on my way back home.
I'll have a look on Monday.

lminiero:
Looks good to me! Have you tested this with and without helpers?

evilmuscleman-super:
111

lminiero:
@ycherniavskyi I did a couple of tests with master and helper sessions, using INVITE and hold/unhold, and it seems to be working fine. If you could do some more tests with the other requests you updated, that might help, since my SIP setup is quite limited. Once that's done, I think it will be good to merge for me :+1:

ycherniavskyi:
@lminiero, apologies for the delayed response. I wanted to conduct thorough testing in my environment before commenting. As of today, I've completed my tests and am ready to report that everything works as expected (in my environment) 😊.

PR #3299, opened 2023-11-29
Change SipPlugin's dtmf pt hardcoded to get from sdp

ywmoyue:
Fix bug raised by @ycherniavskyi from https://github.com/meetecho/janus-gateway/pull/3280#issuecomment-1830743601

https://github.com/meetecho/janus-gateway/pull/3280  set a hardcoded payload type of 101 for the telephone-event RTP payload format.

This PR Change dtmf payload hardcoded to get the telephone-event payload value from sdp

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1505,6 +1505,7 @@ static void janus_sip_media_reset(janus_sip_session *session) {
 	session->media.pre_hold_video_dir = JANUS_SDP_DEFAULT;
 	session->media.video_orientation_extension_id = -1;
 	session->media.audio_level_extension_id = -1;
+	session->media.dtmf_pt = -1;
 	janus_rtp_switching_context_reset(&session->media.acontext);
 	janus_rtp_switching_context_reset(&session->media.vcontext);
 }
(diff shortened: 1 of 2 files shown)


src/plugins/janus_sip.c
```diff
@@ -1047,6 +1047,7 @@ typedef struct janus_sip_media {
 	gboolean updated;
 	int video_orientation_extension_id;
 	int audio_level_extension_id;
+	int dtmf_pt;
```
lminiero, line 1050: Any field you add to `janus_sip_media` must be reset along the others in `janus_nosip_media_reset`, which your patch is currently not doing. Please add a reset to `-1` in there.

src/plugins/janus_sip.c
```diff
@@ -6383,6 +6384,7 @@ void janus_sip_sdp_process(janus_sip_session *session, janus_sdp *sdp, gboolean
 				session->media.has_audio = TRUE;
 				session->media.remote_audio_rtp_port = m->port;
 				session->media.remote_audio_rtcp_port = m->port+1;	/* FIXME We're assuming RTCP is on the next port */
+				session->media.dtmf_pt = janus_sdp_get_codec_pt(sdp, 0, "dtmf");
```
lminiero, line 6387: You're assuming `0` will always be the m-index of audio, which is wrong: there's no guarantee the first m-line will be for audio, or that there will be audio at all. The proper way to use `janus_sdp_get_codec_pt` in those cases is by passing `-1`: since `dtmf` is associated to audio, it will make the SDP utils find the first audio m-line in the SDP and look there.

lminiero (commented):
Thanks for the fix! I've added a couple of notes inline.

As a side note, while doing some unrelated tests today, I noticed that your DTMF detection only seems to be working when using "master" sessions, and not when using "helper" sessions instead (that is, the additional call lines you can associate to the main registration). I didn't have time to investigate yet, but I just thought I'd raise it to your attention, in case you have cycles to debug that too.

ywmoyue:
First of all, thank you for notes, and secondly, thank's for the reminder about the 'helper' session. I haven't used 'helper' session before, I will use and test this dtmf detection later.

For the first note，did you mean the `janus_sip_media_reset` function? I have just modified it. I am not sure whether I need to add the `dtmf_pt` field to `janus_no_sip_media` struct.

lminiero:
> did you mean the `janus_sip_media_reset` function?

Yes, apologies for the typo!

lminiero:
> I haven't used 'helper' session before, I will use and test this dtmf detection later.

It may also be a bug in the demos, as that's what I was using to test a call between two users via the SIP plugin and injecting DTMF digits. I'll make some more tests on that later.

lminiero:
Looks good to me, merging then!

PR #3298, opened 2023-11-28
Upgrade to font-awesome 6.4.2

bkmgit:
Can be previewed at https://janus.kichakatokizito.solutions The documentation folder has hard coded links to https://janus.conf.meetecho.com/ have not changed these.

PR #3300, opened 2023-11-29
Update demos and docs to Bootstrap 5.x

lminiero:
This is a proper refactoring of all the demos and the generated documentation so that it uses more up-to-date stuff, specifically Bootstrap 5.x, which #3295 and #3296 did incorrectly. The patch also updates font-awesome to 6.x, so it already includes (and actually expands, since there was stuff that was incorrect there too) #3298. Notice I will only do this for master: all the 0.x demo/docs will remain as they are.

This required a huge effort, since I first needed to migrate to Bootstrap 4.x (which took quite a while), and from there to 5.x (less problematic but still not a walk in the park). I tried testing all demos in multiple scenarios, but obviously I couldn't test _everything_, so please give them a go yourself too just to ensure I didn't break anything in the process. From a functional perspective, everything should be there.

Before merging, I'll try to also try and take advantage of a few things that Bootstrap has, but that we're relying on external libraries for, like spinners and toasts. Ideally that should help us getting closer to get rid of jQuery, even though that will take longer, and will done (in case) in a separate PR, since we depend on it for some stuff (BlockUI, all selectors, all html appends/prepends and maybe a few other things).

Feedback welcome!

### DIFF
--- a/docs/doxy-boot.js
+++ b/docs/doxy-boot.js
@@ -1,8 +1,8 @@
 $(document).ready(function() {
 
-	$("div.headertitle").addClass("page-header");
+	$("div.headertitle").addClass("pb-2 mt-4 mb-2 border-bottom");
 	$("div.title").addClass("h1");
-	
+
 	$('li > a[href="index.html"] > span').before("<i class='fa fa-cog'></i> ");
 	// $('li > a[href="index.html"] > span').text("CoActionOS");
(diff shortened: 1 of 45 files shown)


lminiero:
Pinging @bkmgit since he was the author of the original efforts.

bkmgit:
Thanks for the updates.

PR #3302, opened 2023-12-06
Move all demo pages to a demos/ subfolder

lminiero:
After updating demos and docs to use Bootstrap 5.x in #3300, I reorganized the `html` folder a bit, by moving all demos to a `demos` subfolder: the purpose was to cleanup the web folder a bit, like when we moved all the code to `src`.

As part of this restructuring, I also renamed most of the demo pages too: many of them still had a `test` in it that was unneeded, so I got rid of it. Besides, I added a link to the related plugin's documentation on top of every demo: this will hopefully make people more aware of which plugin is powering each demo, and convince them to checkout the related docs (which not everyone does, unfortunately) if they want to learn more.


lminiero:
That wouldn't work, because it would put all html files nt in `demos` to be put in the `janus` parent folder, which would be wrong. But I agree the double `demos` isn't a good look: maybe we can do `$(datadir)/janus/html`, which would make more sense since `html` is how it's called in the repo too (which would make it `$(datadir)/janus/html/demos` for demos, as expected).

bkmgit:
That seems ok, thanks.

PR #3305, opened 2023-12-07
DRAFT: Load javascript last

bkmgit:
The aim is to give a better user experience on slow connections, in particular mobile. Some reference material:
- https://developers.google.com/speed/docs/insights/mobile?csw=1#PutStylesBeforeScripts
- https://stackoverflow.com/questions/1860482/when-do-you-choose-to-load-your-javascript-at-the-bottom-of-the-page-instead-of
- https://webmasters.stackexchange.com/questions/883/does-putting-javascript-at-the-bottom-defeat-the-purpose-of-document-ready

If this would be considered, other options are:
- to only have the jquery, title bar loader and footer loader scripts at the top, and everything else at the bottom.
- to add attributes such as defer, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

### DIFF
--- a/html/citeus.html
+++ b/html/citeus.html
@@ -5,17 +5,6 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
 <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
 <title>Janus WebRTC Server (multistream): Cite us!</title>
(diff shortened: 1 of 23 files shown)


lminiero:
Please rebase it on #3302, which moved demos to a different folder.

Issue #3297, opened 2023-11-27
[1.x] crack sounds on audiobridge, then publisher has drops on output stream


spscream:
**What version of Janus is this happening on?**
1.2.1 and master

**Have you tested a more recent version of Janus too?**
tes, tested on master and on rnnoise branch

**Was this working before?**
We are always has this issue, I reported to discourse group before.

**Is there a gdb or libasan trace of the issue?**
no

**Additional context**
I tested drops using clumsy utility. listeners start to hear the crack sounds starting from 5% of drops.  I also tried to set expectedloss values for ab, but it doesn't help. On vr we don't hear the same cracks.

spscream:
I tried to add log then fec is used(https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_audiobridge.c#L8271), but it doesn't trigger any logs(only single occurence of it then participant starts stream)

spscream:
> I tried to add log then fec is used(https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_audiobridge.c#L8271), but it doesn't trigger any logs(only single occurence of it then participant starts stream)

looks like mine logging issue, fec occurence is logged out after publisher has been destroyed

lminiero:
I think you may be a bit confused as to how FEC works. There's FEC on the way in, and there's FEC on the way out.

Our code to process FEC on the way _in_ (the line you reference in your comment) depends on whether or not the _browser_ is actually adding any FEC to the packets; if they aren't, there won't be any FEC to take advantage of when decoding the media. Browsers won't send any FEC unless they receive reports via RTCP that a lot of packets are being lost (can't remember what the threshold is).

FEC on the way _out_ is whether or not the _AudioBridge_ adds FEC to outgoing packets, so when encoding Opus. That's what the expectedloss property is for: you're telling Janus you expect X% of packets to be lost in delivery, and so telling the plugin to add FEC when encoding media. It has nothing to do with FEC on the way in.

lminiero:
That it doesn't appear in the VideoRoom is irrelevant: the VR only relays media, which means it's the browser that terminates audio doing its magic there. The AudioBridge has to terminate stuff. Again, if the browser is not sending FEC, then we can't use it. Try checking the webrtc-internals to see if FEC is actually being added to packets, and how much.

spscream:
[...]

PR #3308, opened 2023-12-13
Add insert of concealed packets

spscream:
this changes were created to address #3297 for our environment. 
I'm not sure if it is fully correct, but it mostly fixes our problems.

### DIFF
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -1699,6 +1699,7 @@ typedef struct janus_audiobridge_participant {
 	uint16_t expected_seq;		/* Expected sequence number */
 	uint16_t probation; 		/* Used to determine new ssrc validity */
 	uint32_t last_timestamp;	/* Last in seq timestamp */
+	uint16_t last_seq;	/* Last sequence number */
 	gboolean reset;				/* Whether or not the Opus context must be reset, without re-joining the room */
 	GThread *thread;			/* Encoding thread for this participant */
 	gboolean mjr_active;		/* Whether this participant has to be recorded to an mjr file or not */
@@ -2135,7 +2136,7 @@ static int janus_audiobridge_resample(int16_t *input, int input_num, int input_r
 #define JITTER_BUFFER_MIN_PACKETS 2
 #define JITTER_BUFFER_MAX_PACKETS 40
 #define JITTER_BUFFER_CHECK_USECS 1*G_USEC_PER_SEC
-#define QUEUE_IN_MAX_PACKETS 4
+#define QUEUE_IN_MAX_PACKETS 20
 
 
 /* Error codes */
@@ -6504,6 +6505,7 @@ static void *janus_audiobridge_handler(void *data) {
 				participant->expected_seq = 0;
 				participant->probation = 0;
 				participant->last_timestamp = 0;
+				participant->last_seq = 0;
 				janus_mutex_init(&participant->qmutex);
 				participant->arc = NULL;
 				janus_audiobridge_plainrtp_media_cleanup(&participant->plainrtp_media);
@@ -8613,6 +8615,7 @@ static void *janus_audiobridge_participant_thread(void *data) {
 	gint64 now = janus_get_monotonic_time(), before = now;
 	gboolean first = TRUE, use_fec = FALSE;
 	int ret = 0;
+	int lost_packets = 0;
 
 	/* Start working: check both the incoming queue (to decode and queue) and the outgoing one (to encode and send) */
 	while(!g_atomic_int_get(&stopping) && g_atomic_int_get(&session->destroyed) == 0) {
@@ -8659,8 +8662,27 @@ static void *janus_audiobridge_participant_thread(void *data) {
 					/* If this is Opus, check if there's a packet gap we should fix with FEC */
 					use_fec = FALSE;
 					if(!first && participant->codec == JANUS_AUDIOCODEC_OPUS && participant->fec) {
-						if(ntohs(rtp->seq_number) == (participant->expected_seq + 1)) {
-							/* Lost a packet here? Use FEC to recover */
+						if(ntohs(rtp->seq_number) != (participant->expected_seq)) {
+							lost_packets = ntohs(rtp->seq_number) - participant->expected_seq;
+							JANUS_LOG(LOG_ERR, "[%d] got rtp seq, expected_seq: [%d], diff: [%d]\n", ntohs(rtp->seq_number), participant->expected_seq, lost_packets);
+
+							janus_mutex_lock(&participant->qmutex);
+
+							for(int i=1; i < lost_packets; i++) {
+								pkt = g_malloc(sizeof(janus_audiobridge_rtp_relay_packet));
+								pkt->data = g_malloc0(OPUS_SAMPLES * (participant->stereo ? 2 : 1) * sizeof(opus_int16));
+								pkt->ssrc = 0;
+								pkt->timestamp = participant->last_timestamp + 960 * i;
+								pkt->seq_number = participant->last_seq + 1 * i;
+								/* This is a redundant packet, so we can't parse any extension info */
+								pkt->silence = FALSE;
+								pkt->length = opus_decode(participant->decoder, NULL, 0, (opus_int16 *)pkt->data, OPUS_SAMPLES, 0);
+								JANUS_LOG(LOG_ERR, "[%d] packet concealed, length: [%d], timestamp: [%d]\n", pkt->seq_number, pkt->length, pkt->timestamp);
+								participant->inbuf = g_list_append(participant->inbuf, pkt);
+							}
+
+							janus_mutex_unlock(&participant->qmutex);
+
 							use_fec = TRUE;
 						}
 					}
@@ -8668,16 +8690,18 @@ static void *janus_audiobridge_participant_thread(void *data) {
 					if(use_fec) {
 						/* There was a gap, try to get decode from redundant info first */
 						pkt = g_malloc(sizeof(janus_audiobridge_rtp_relay_packet));
-						pkt->data = g_malloc0(BUFFER_SAMPLES*sizeof(opus_int16));
+						pkt->data = g_malloc0(OPUS_SAMPLES * (participant->stereo ? 2 : 1) * sizeof(opus_int16));
 						pkt->ssrc = 0;
-						pkt->timestamp = participant->last_timestamp + 960;	/* FIXME */
-						pkt->seq_number = participant->expected_seq;		/* FIXME */
+						pkt->timestamp = ntohl(rtp->timestamp) - 960;	/* FIXME */
+						pkt->seq_number = ntohs(rtp->seq_number) - 1;		/* FIXME */
 						/* This is a redundant packet, so we can't parse any extension info */
 						pkt->silence = FALSE;
-						/* Decode the lost packet using fec=1 */
-						int32_t output_samples;
-						opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
-						pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, output_samples, 1);
+
+						pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, OPUS_SAMPLES, 1);
+
+						JANUS_LOG(LOG_ERR, "[%d] packet fec decoded [%d] pkt->length, timestamp: [%d]\n",
+								pkt->seq_number, pkt->length, pkt->timestamp);
+
 						/* Queue the decoded redundant packet for the mixer */
 						janus_mutex_lock(&participant->qmutex);
 						participant->inbuf = g_list_append(participant->inbuf, pkt);
@@ -8720,6 +8744,7 @@ static void *janus_audiobridge_participant_thread(void *data) {
 					/* Get rid of the buffered packet */
 					janus_audiobridge_buffer_packet_destroy(bpkt);
 					/* Update the details */
+					participant->last_seq = pkt->seq_number;
 					participant->last_timestamp = pkt->timestamp;
 					participant->expected_seq = pkt->seq_number + 1;
 					g_atomic_int_set(&participant->decoding, 0);
@@ -8738,7 +8763,7 @@ static void *janus_audiobridge_participant_thread(void *data) {
 					locked = TRUE;
 					/* Do not let queue-in grow too much */
 					guint count = g_list_length(participant->inbuf);
-					if(count > QUEUE_IN_MAX_PACKETS) {
+					if((int) count > (QUEUE_IN_MAX_PACKETS + lost_packets)) {
 						JANUS_LOG(LOG_WARN, "Participant queue-in contains too many packets, clearing now (count=%u)\n", count);
 						janus_audiobridge_participant_clear_inbuf(participant);
 					}
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -8658,6 +8658,7 @@ static void *janus_audiobridge_participant_thread(void *data) {
 						janus_audiobridge_buffer_packet_destroy(bpkt);
 						break;
 					}
+
 					rtp = (janus_rtp_header *)bpkt->buffer;
 					/* If this is Opus, check if there's a packet gap we should fix with FEC */
 					use_fec = FALSE;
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @spscream! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

src/plugins/janus_audiobridge.c
```diff
@@ -8659,25 +8662,46 @@ static void *janus_audiobridge_participant_thread(void *data) {
...
 					if(!first && participant->codec == JANUS_AUDIOCODEC_OPUS && participant->fec) {
-						if(ntohs(rtp->seq_number) == (participant->expected_seq + 1)) {
-							/* Lost a packet here? Use FEC to recover */
+						if(ntohs(rtp->seq_number) != (participant->expected_seq)) {
+							lost_packets = ntohs(rtp->seq_number) - participant->expected_seq;
+							JANUS_LOG(LOG_ERR, "[%d] got rtp seq, expected_seq: [%d], diff: [%d]\n", ntohs(rtp->seq_number), participant->expected_seq, lost_packets);
+
+							janus_mutex_lock(&participant->qmutex);
+
+							for(int i=1; i < lost_packets; i++) {
```
spscream, line 8671: I think to get rid of cycle and produce plc data which  `frame_size` will be `lost_packets * OPUS_SAMPLES` and shouldn't exceed `BUFFER_SAMPLES` size.
In our production I've seen losses of 20+ packets in a row, so plc for 20 packet won't fit BUFFER_SAMPLES and I'm not sure how to behave in this case.

src/plugins/janus_audiobridge.c
```diff
@@ -1699,6 +1699,7 @@ typedef struct janus_audiobridge_participant {
 	uint16_t expected_seq;		/* Expected sequence number */
 	uint16_t probation; 		/* Used to determine new ssrc validity */
 	uint32_t last_timestamp;	/* Last in seq timestamp */
+	uint16_t last_seq;	/* Last sequence number */
```
lminiero, line 1702: Code style: indentation of comment looks broken here.

src/plugins/janus_audiobridge.c
```diff
@@ -2135,7 +2136,7 @@ static int janus_audiobridge_resample(int16_t *input, int input_num, int input_r
 #define JITTER_BUFFER_MIN_PACKETS 2
 #define JITTER_BUFFER_MAX_PACKETS 40
 #define JITTER_BUFFER_CHECK_USECS 1*G_USEC_PER_SEC
-#define QUEUE_IN_MAX_PACKETS 4
+#define QUEUE_IN_MAX_PACKETS 20
```
lminiero, line 2139: I personally don't like this: it will cause a permanent delay increase for audio of this participant after a burst of lost packets caused this to grow because of PLC, ven way after the problem was solved. I'd argue that a burst of lost packets should indeed cause artifacts (like a speed up), and that if not 4, then a way smaller number should be more than enough to absorb shorter bursts.

src/plugins/janus_audiobridge.c
lminiero: It's actually the whole thing that perplexes me. The way it's done right now, you're basically breaking FEC, as it would never be done. You're preserving the `use_fec = true` bit, but it will be useless, since later on we'll try to get redundant information from a packet that doesn't have any (the PLC data you've added). I think PLC and FEC should be handled separately: if the gap is 1, then do what was done before; if it's higher than 1, do PLC (new code, without use_fec=true). Or do you have a reason for implementing it this way?

As a side not, the LOG_ERR there is probably overly verbose, and would clog the logs any time there's bursts of lost packets, which may happen often. A LOG_VERB or even LOG_HUGE would probably be a better choice (but that's something that can be discussed, it may still be useful to have info on the logs when this happens).

src/plugins/janus_audiobridge.c
```diff
@@ -8655,29 +8658,55 @@ static void *janus_audiobridge_participant_thread(void *data) {
...
+
+							int32_t output_samples;
+							opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+
+							for(int i=1; i < lost_packets; i++) {
+								pkt = g_malloc(sizeof(janus_audiobridge_rtp_relay_packet));
+								pkt->data = g_malloc0(OPUS_SAMPLES * (participant->stereo ? 2 : 1) * sizeof(opus_int16));
+								pkt->ssrc = 0;
+								pkt->timestamp = participant->last_timestamp + 960 * i;
+								pkt->seq_number = participant->last_seq + 1 * i;
```
lminiero, line 8678: I'm not 100% sure the RTP info here is correct, since `last_timestamp` and `last_seq` may come from out of order packets. But it's probably not that relevant, since it has a limited use anyway in the code.

src/plugins/janus_audiobridge.c
```diff
@@ -8655,29 +8658,55 @@ static void *janus_audiobridge_participant_thread(void *data) {
...
+						pkt->data = g_malloc0(OPUS_SAMPLES * (participant->stereo ? 2 : 1) * sizeof(opus_int16));
 						pkt->ssrc = 0;
-						pkt->timestamp = participant->last_timestamp + 960;	/* FIXME */
-						pkt->seq_number = participant->expected_seq;		/* FIXME */
+						pkt->timestamp = ntohl(rtp->timestamp) - 960;	/* FIXME */
+						pkt->seq_number = ntohs(rtp->seq_number) - 1;		/* FIXME */
 						/* This is a redundant packet, so we can't parse any extension info */
 						pkt->silence = FALSE;
-						/* Decode the lost packet using fec=1 */
+
```
lminiero, line 8701: I'm not sure why you completely changed the way we handle FEC? The existing FEC management should remain the same, if not broken. See comments above.

src/plugins/janus_audiobridge.c
```diff
@@ -8655,29 +8658,55 @@ static void *janus_audiobridge_participant_thread(void *data) {
...
 						int32_t output_samples;
 						opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+
 						pkt->length = opus_decode(participant->decoder, payload, plen, (opus_int16 *)pkt->data, output_samples, 1);
+
+						JANUS_LOG(LOG_ERR, "[%d] packet fec decoded [%d] pkt->length, timestamp: [%d]\n",
+								pkt->seq_number, pkt->length, pkt->timestamp);
```
lminiero, line 8708: See above on LOG_ERR usage.

src/plugins/janus_audiobridge.c
```diff
@@ -8738,7 +8768,7 @@ static void *janus_audiobridge_participant_thread(void *data) {
 					locked = TRUE;
 					/* Do not let queue-in grow too much */
 					guint count = g_list_length(participant->inbuf);
-					if(count > QUEUE_IN_MAX_PACKETS) {
+					if((int) count > (QUEUE_IN_MAX_PACKETS + lost_packets)) {
```
lminiero, line 8771: Once you increase QUEUE_IN_MAX_PACKETS, then there's no reason to make it even less strict like you're doing now IMHO. It will make the internal mixer queue for the participant grow beyond control and cause huge and uncoverable latencies.

src/plugins/janus_audiobridge.c
spscream: This code generates plc for gap sequence, which can't be recovered using fec. Last gap member is generated using fec of arrived rtp packet, if it's possible(opus_decode will return plc in case fec data isn't available), for example:
- gap of 1 will be covered using fec, since for cycle wont match `i < lost_packets` condition
- gap of 2 will generate 1 plc and 1 will be recovered from fec information from rtp payload

I'm affraid of the following problem now - frame size can be differ from packet to packet and decoder should adapt to changes of frame size accordingly since encoder may change frame_size. Looks like chrome encoder changes frame size in case of losses - opus_packet_get_nb_frames returns 10ms frame size periodically on bursts of losses for rtp packet after gap. Chromium realisation of fec decode uses opus_packet_get_nb_frames to determine fec frame_size https://chromium.googlesource.com/external/webrtc/+/refs/heads/main/modules/audio_coding/codecs/opus/opus_interface.cc#681).

I left LOG_ERR for debug purpose, i'm going to change it to LOG_VERB or LOG_HUGE for production.

src/plugins/janus_audiobridge.c
spscream: I thought jitter buffer is reordering out of order packets and duplicates? At least I've expiremented with out of order and duplicates on output and everything still worked fine

src/plugins/janus_audiobridge.c
spscream: FEC management which were implemented earlier didn't work in case of gap > 1, I didn't change too much, just adapted it to work using latest arrived rtp packet info instead of expected value based on last decoded packet.

src/plugins/janus_audiobridge.c
spscream: I tried to set it to default 4 and got pops and cracks. On screen is 20-30% of drops set on output channel.
image

With 20 QUEUE_IN_MAX_PACKETS I don't have such pops/cracks and I don't get over the limit.

lminiero:
@spscream have you tried what happens if, instead of ingesting `lost_packets` concealed packets in that for loop, you add only one? My guess is that the end result should be the same, since it will try to recreate stuff from the last thing it received, and should have the added benefit of not inflating the queue as your patch would do now.

atoppi:
Thanks @spscream for the effort!

My objection to the implementation is related to the timing of PLC  packets generation.

FEC needs packet N+1 to decode packet N (so it will naturally introduce/leverage a playout delay), meanwhile PLC is only calculated from already decoded samples. According to this assumption, the generation of a PLC packet should be triggered in a condition where the bridge encoder "needs something from a participant but there's nothing available", and not after the arrival of a random packet not adjacent to the previous one.

Reusing FEC approach will inevitably lead to issues like unwanted delay, discarding of good packets, uncontrollable queue growth etc.

I still have to think  about a better approach in detail, however one idea is to trigger a PLC whenever a participant's queue-in is empty (e.g. detected participant underflow). That means that a loss burst has exhausted both the jitter buffer and the participant's queue (decoded samples), hence the mixer has no to other chance to introduce concealment for the participant.

spscream:
> @spscream have you tried what happens if, instead of ingesting `lost_packets` concealed packets in that `for` loop, you add only one? My guess is that the end result should be the same, since it will try to recreate stuff from the last thing it received, and should have the added benefit of not inflating the queue as your patch would do now.

yes, I tried it and sound cracks

spscream:
> > @spscream have you tried what happens if, instead of ingesting `lost_packets` concealed packets in that `for` loop, you add only one? My guess is that the end result should be the same, since it will try to recreate stuff from the last thing it received, and should have the added benefit of not inflating the queue as your patch would do now.
> 
> yes, I tried it and sound cracks

I can insert n packets of maximum size instead of number of loss packets, but plc should cover time period of lost bursts accordingly to docs. `frame_size` parameter of opus_decode has the following description: 

Number of samples per channel of available space in pcm. If this is less than the maximum packet duration (120ms; 5760 for 48kHz), this function will not be capable of decoding some packets. In the case of PLC (data==NULL) or FEC (decode_fec=1), then frame_size needs to be exactly the duration of audio that is missing, otherwise the decoder will not be in the optimal state to decode the next incoming packet. For the PLC and FEC cases, frame_size must be a multiple of 2.5 ms.

spscream:
> I still have to think about a better approach in detail, however one idea is to trigger a PLC whenever a participant's queue-in is empty (e.g. detected participant underflow). That means that a loss burst has exhausted both the jitter buffer and the participant's queue (decoded samples), hence the mixer has no to other chance but introduce concealment for the participant.

This buffer underrun can explain why we still have this cracks/pops for our production users. But I afraid I can't rewrite it to such aproach myself or at least not this year.

lminiero:
I agree with @atoppi that it's the road we should explore, even though it does come with its own challenges: for one, the mixer is the one who'd read from that queue, and it uses a different thread, meaning I'd prefer it not to be the one that should perform that PLC decode. As such, we'll have to better think of how to do that in the participant thread instead. This is something we can definitely discuss together after the vacations, no need to hurry and no need for you to actually implement it yourself either, if we manage to replicate the problem and come up with a potential fix.

spscream:
@atoppi  `jitter_buffer_get` has JITTER_BUFFER_MISSING and JITTER_BUFFER_INSERTION ret codes, so we can use it as a signal to insert plc, you can take it to considerations.

By the way I also discovered possible incorrect usage of JITTER_BUFFER_SET_MARGIN in current implementation - its value should be the same unit as the `frame_size` (e.g. JITTER_BUFFER_MIN_PACKETS * frame_size), since its used as timestamp under the hood, not as count of packets.

lminiero:
> its value should be the same unit as the `frame_size`

I don't think so. Besides, we're [changing](#3312) what actually ends in the buffer: not the payload anymore, but a pointer to a duplicate of the `janus_plugin_rtp` we got from the core.

spscream:
> > its value should be the same unit as the `frame_size`
> 
> I don't think so. Besides, we're [changing](#3312) what actually ends in the buffer: not the payload anymore, but a pointer to a duplicate of the `janus_plugin_rtp` we got from the core.

I get the following situation with buffering if `JITTER_BUFFER_SET_MARGIN` is 2, the jitter buffer doesn't mantain its minimal buffer size:
```
[WARN] [18844549281349] jitter buffer avail: 0
[WARN] [18844549281349] jitter_buffer_get result 1
[WARN] [18844549281771] jitter buffer avail: 0
[WARN] [18844549281771] jitter_buffer_get result 1
[WARN] [18844549281829] jitter buffer avail: 2
[WARN] [18844549281829] jitter_buffer_get result 0
[WARN] [18844549300041] jitter buffer avail: 0
[WARN] [18844549300041] jitter_buffer_get result 1
[WARN] [18844549300112] jitter buffer avail: 2
[WARN] [18844549300112] jitter_buffer_get result 0
[WARN] [18844549302419] jitter buffer avail: 0
[WARN] [18844549302419] jitter_buffer_get result 1
[WARN] [18844549321048] jitter buffer avail: 2
[WARN] [18844549321048] jitter_buffer_get result 0
[WARN] [18844549321144] jitter buffer avail: 0
[WARN] [18844549321144] jitter_buffer_get result 1
[WARN] [18844549323703] jitter buffer avail: 0
[WARN] [18844549323703] jitter_buffer_get result 1
```
if I multiply `JITTER_BUFFER_MIN_PACKETS` by `frame_size` for `JITTER_BUFFER_SET_MARGIN`:
```
[WARN] [18844964419785] jitter_buffer_get result 0
[WARN] [18844964438318] jitter buffer avail: 3
[WARN] [18844964438318] jitter_buffer_get result 0
[WARN] [18844964459607] jitter buffer avail: 3
[WARN] [18844964459607] jitter_buffer_get result 0
[WARN] [18844964478355] jitter buffer avail: 3
[WARN] [18844964478355] jitter_buffer_get result 0
[WARN] [18844964499529] jitter buffer avail: 3
[WARN] [18844964499529] jitter_buffer_get result 0
[WARN] [18844964518172] jitter buffer avail: 3
[WARN] [18844964518172] jitter_buffer_get result 0
[WARN] [18844964539348] jitter buffer avail: 3
[WARN] [18844964539348] jitter_buffer_get result 0
[WARN] [18844964557793] jitter buffer avail: 3
[WARN] [18844964557793] jitter_buffer_get result 0
[WARN] [18844964578942] jitter buffer avail: 3
[WARN] [18844964578942] jitter_buffer_get result 0
[WARN] [18844964600042] jitter buffer avail: 3
[WARN] [18844964600042] jitter_buffer_get result 0
[WARN] [18844964618636] jitter buffer avail: 3
[WARN] [18844964618636] jitter_buffer_get result 0
[WARN] [18844964639613] jitter buffer avail: 3
[WARN] [18844964639613] jitter_buffer_get result 0
```

atoppi:
> By the way I also discovered possible incorrect usage of JITTER_BUFFER_SET_MARGIN in current implementation - its value should be the same unit as the frame_size (e.g. JITTER_BUFFER_MIN_PACKETS * frame_size), since its used as timestamp under the hood, not as count of packets.

@spscream yeah you're probably right, the comments in the speexdsp code are probably outdated/misleading.
Internally the macro sets a `buffer_margin` variable that is used as a RTP timestamp.

spscream:
I've tried another approach here and it works in our case even with huge loses: https://github.com/meetecho/janus-gateway/compare/master...spscream:janus-gateway:fix/am/audiobridge-cracks-fix-rnnoise#diff-476cfd26c102e4b3311b3b8e8f848bfeba283c1712a5f3f117254e0593a986f4R8959

@atoppi could you please look on it? If it is ok, i could refactor it and make another pr for you. Idea is the following:
- if i got ret == JITTER_BUFFER_MISSING || ret == JITTER_BUFFER_INSERTION from jitter_buffer_get I insert concealed packets
- if loss gap is huge I'm stopping to insert plc

I also tried to insert fec, but it seams it also leads to cracks

atoppi:
@spscream the diff is huge and includes unrelated changes, can you please provide a more readable patch?

spscream:
@atoppi hi! any thoughts?

spscream:
If it is ok, I could make a new pr with my latest changes if it will be more convenient

atoppi:
@spscream If you're able to propose a cleaner version of the patch if would greatly help thanks.
I'll try to review this in the next few days.

PR #3312, opened 2023-12-18
Perform audio-level detection in AudioBridge participant thread

### LINKED ISSUE #3304: [1.x] AudioBridge: after network interrupt there's no `stopped-talking` event until timeout
v1.2.1

No. (Which more recent version? :-)

**Was this working before?**
No

**Additional context**
When you start talking, you receive a `talking` event along these lines:

```
{
   "janus": "event",
   "session_id": 7115030027245058,
   "sender": 3472346807042796,
   "plugindata": {
      "plugin": "janus.plugin.audiobridge",
      "data": {
         "audiobridge": "talking",
         "room": "563861637229053",
         "id": "my-network-will-be-interrupted"
      }
   }
}
```

When you stop talking, you'll receive a `stopped-talking` event in turn. This all works very well. But if you cut the connection, there will never be a `stopped-talking` event.

In essence this seems to be because all relevant logic takes place in `janus_audiobridge_incoming_rtp()`, but in this scenario there is no incoming RTP packet.

https://github.com/meetecho/janus-gateway/blame/6f6ac5eaf679bc2dc4d5de39051c22c233c38a74/src/plugins/janus_audiobridge.c#L5853

Ideally it would send a `stopped-talking` event long before the eventual timeout, but it's not immediately clear to me if that's relatively simple (no RTP packets for a couple of seconds = stopped talking) or very hard to achieve within the current architecture.

lminiero:
This is an attempt to fix #3304, which causes talk-events to stop flowing if RTP stops coming in. The way this patch tries to solve it is by moving the talk-detection stuff to the participant thread, taking advantage of the new jitter buffer refactoring: this allows us to still advance the talk-detection timers even when there's no audio coming in. Notice that this won't work in case the participant is suspended, since in that case we don't perform the pull every 20ms we otherwise do.

I only tested if audio still works, since this required a refactoring of what goes through the jitter buffer (which was needed to pass the parsed audio levels along too). I haven't tested yet if this does indeed solve the issue, so please test and provide feedback.

This is for `master`, I'll backport to `0.x` too after merging if this works.

### DIFF
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -1729,29 +1729,27 @@ typedef struct janus_audiobridge_rtp_relay_packet {
 /* Buffered audio/video packet */
 typedef struct janus_audiobridge_buffer_packet {
 	/* Pointer to the packet data, if RTP */
-	char *buffer;
-	/* Size of the packet */
-	int len;
-	/* Whether the packet contains silence, according to the RTP extension */
-	gboolean silence;
+	janus_plugin_rtp *rtp;
 	/* Monotonic insert time */
 	int64_t inserted;
 } janus_audiobridge_buffer_packet;
-static janus_audiobridge_buffer_packet *janus_audiobridge_buffer_packet_create(char *buffer, int len, gboolean silence) {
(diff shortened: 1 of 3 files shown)


atoppi:
lgtm :+1:

lminiero:
Merging.

PR #3316, opened 2023-12-28
Fixed broken faststart when postprocessing AV1 recordings (fixes #3315)

corthiclem:
This fixes #3315. It replicates what has been done in adea15f but for AV1 post-processing.

PR #3317, opened 2023-12-29
Fixed broken faststart when postprocessing AV1 recordings

### LINKED ISSUE #3315: [1.x] janus-pp-rec fails with faststart option when running on AV1 recordings
v1.2.1

No, v1.2.1 is the latest release as of today.

**Was this working before?**
Never tried before

No

**Additional context**
I am trying to use the `faststart` option of `janus-pp-rec` with **AV1** recordings to create `.mp4` files but it appears to be broken somehow. I get the following error when running `janus-pp-rec -t input.mjr output.mp4`:
```
Unable to re-open  output file for shifting data
```

It seems to be related to #3037.

I'll try to create a pull request to fix the issue.

corthiclem:
This fixes #3315. It replicates what has been done in adea15f24b04301c6a925800c923fa2835fda7e6 but for AV1 post-processing. (And re-opens #3316 without typo in the branch name)

### DIFF
--- a/src/postprocessing/pp-av1.c
+++ b/src/postprocessing/pp-av1.c
@@ -55,6 +55,10 @@ int janus_pp_av1_create(char *destination, char *metadata, gboolean faststart, c
 		return -1;
 	}
 
+	char filename[1024];
+	snprintf(filename, sizeof(filename), "%s", destination);
+	fctx->url = g_strdup(filename);
+
 	vStream = janus_pp_new_video_avstream(fctx, AV_CODEC_ID_AV1, max_width, max_height);
 	if(vStream == NULL) {
 		JANUS_LOG(LOG_ERR, "Error adding stream\n");
--- a/src/postprocessing/pp-av1.c
+++ b/src/postprocessing/pp-av1.c
@@ -528,7 +528,8 @@ void janus_pp_av1_close(void) {
 	if(fctx != NULL) {
 		av_write_trailer(fctx);
 		avio_close(fctx->pb);
-		avformat_free_context(fctx);
 		g_free(fctx->url);
+		fctx->url = NULL;
+		avformat_free_context(fctx);
 	}
 }
--- a/src/postprocessing/pp-av1.c
+++ b/src/postprocessing/pp-av1.c
@@ -529,5 +529,6 @@ void janus_pp_av1_close(void) {
 		av_write_trailer(fctx);
 		avio_close(fctx->pb);
 		avformat_free_context(fctx);
+		g_free(fctx->url);
 	}
 }
--- a/src/postprocessing/pp-av1.c
+++ b/src/postprocessing/pp-av1.c
@@ -55,9 +55,7 @@ int janus_pp_av1_create(char *destination, char *metadata, gboolean faststart, c
 		return -1;
 	}
 
-	char filename[1024];
-	snprintf(filename, sizeof(filename), "%s", destination);
-	fctx->url = g_strdup(filename);
+	fctx->url = g_strdup(destination);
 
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @corthiclem! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

src/postprocessing/pp-av1.c
```diff
@@ -55,6 +55,10 @@ int janus_pp_av1_create(char *destination, char *metadata, gboolean faststart, c
 		return -1;
 	}
 
+	char filename[1024];
+	snprintf(filename, sizeof(filename), "%s", destination);
+	fctx->url = g_strdup(filename);
```
lminiero, line 60: I don't understand why you're going through a `sprintf` here. Can't you just `g_strdup(destination)`?

corthiclem: You're right, thanks! Fixed.

lminiero: Don't forget the `g_free(fctx->url);` that the commit you referenced includes, or you'll be introducing a leak. I see that patch also did a few other things for each codec, like setting `CODEC_FLAG_GLOBAL_HEADER`: are those other changes unneeded for AV1 processing?

corthiclem: If I am not wrong, the `avformat_free_context` function is used in `pp_av1.c` when closing the file, which already frees the memory allocated (according to [ffmpeg documentation](https://ffmpeg.org/doxygen/trunk/group__lavf__core.html#gac2990b13b68e831a408fce8e1d0d6445)), so there is no need for `g_free(fctx->url)`.

I think the patch just removed unused comments and that `CODEC_FLAG_GLOBAL_HEADER` flag was already there for `H264` and `H265` before the patch I mentionned. I can try to add the global header flag but I am not familiar enough with `libav` to know what will be the impact on the resulting video. I think it's not used either for WebM postprocessing.

corthiclem: When I add `g_free` before `avformat_free_context`, I do get a `free(): double free detected in tcache 2` but it seems to be working fine if I add it after. I updated the pull request. 

What I could do is use `av_strdup` instead of `g_strdup` I guess.

lminiero: That's because you ignored part of my previous message :slightly_smiling_face: 

> You have to free it yourself with g_free and then set the pointer to NULL, which is what the other patch did.

You didn't set the pointer to `NULL`, so `avformat_free_context` tries to free something that doesn't exist anymore.

corthiclem: Oh sorry, Monday mornings ... 😅 
I can move it up if you prefer.

Concerning the global flag headers, I suggest doing it in another pull request as it requires a bit more work (creating a codec context and switching to a codec parameter which is not yet done for in `pp-av1`)

lminiero: Sure, no problem! I asked about the global flag header only because it was in the other patch, and so I wondered if it could be needed for AV1 too: it may very well be that it isn't, I rarely work with any AV1 recording at all so I don't have experience with them.

lminiero:
Thanks for the fixes! If you can confirm it works for your use case, now, I'll merge :+1:

src/postprocessing/pp-av1.c
corthiclem: I'm new to AV1 recording, so I don't know if the global flag header helps but I'll take a deeper look at this as soon as I have time.

lminiero:
Ack, merging then! I'll backport to `0.x` too. Thanks again!

PR #3323, opened 2024-01-22
feat: An attempt to forward a handle's media stream with multiple threads for videoroom.

wswaaa123:
Hi, this PR tries to add multithreaded packet forwarding support for videoroom.

I found that when a meeting is distributed by multiple people, if a handle forwards multiple media streams in the same thread, the "Discarding too old outgoing packet" situation will occur, then the whole server stucks. I wonder if this problem can be tackled with multiple threads forwarding a single handle's media stream on sufficient server resources. So this PR is an attempt to forward a handle's media stream with multiple threads.

### DIFF
--- a/src/ice.c
+++ b/src/ice.c
@@ -227,6 +227,17 @@ static void *janus_ice_static_event_loop_thread(void *data) {
 	janus_refcount_decrease(&loop->ref);
 	return NULL;
 }
+static void janus_relay_helper_destroy(janus_relay_helper* helper) {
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @wswaaa123! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
@wswaaa123 have you checked #3067? That's exactly what that patch does. Please test and provide feedback to that.

lminiero:
Closing as a duplicate of an already existing PR. If you believe your patch provides anything that one doesn't already, please provide more info.

PR #3325, opened 2024-01-27
FIx type declaration file for `getBitrate`.

shoka6:
This PR modifies the type definition of the 'getBitrate' function, referencing the following documentation.

> getBitrate(mid): gets a verbose description of the currently received video stream bitrate (optional mid to specify the stream, first video stream if missing);

https://janus.conf.meetecho.com/docs/JS.html


lminiero:
Thanks for the fix, merging!

PR #3330, opened 2024-02-15
Remove .nfo file provisioning in auviousrecordplay Plugin

anger232:
Removed all .nfo file provisioning in auviousrecordplay plugin. Recording metadata in auvious is handled by recorder service so we do not need this logic. Also with many recordings over time this would lead to unnecessary memory usage and could also lead to erroneous behaviour, which could occur if the janus container was restarted, since it would attempt to load all previous .nfo files and associated recording files.


lminiero:
Created a PR against the wrong repo, huh? 🤭

anger232:
Yes sincere apologies!

Issue #3332, opened 2024-02-23
[1.x] Sip session get stuck


Jonbeckas:
**What version of Janus is this happening on?**
1.2.2; b98e3bb91bd728ce21f6fd56519a303f2775f755

**Have you tested a more recent version of Janus too?**
Yes, on the master branch

**Was this working before?**
Not sure, behaviour of the telephone system we use changed (3cx)

**Additional context**
If a session has been hangup, the `JANUS_ICE_WEBRTC_ALERT` flag will be set in `janus_ice_webrtc_hangup` and removed in the next call in `janus_ice_setup_local` which is called by `janus_plugin_handle_sdp`.
For a denied incoming sip call with no sdp body, the `JANUS_ICE_WEBRTC_ALERT` will not be removed and during the following hangup the 
`janus_ice_webrtc_hangup` will be abortet before the plugin is notified and the establishing attribute will be set to 0, so the session will get stuck and denies all incoming calls.

lminiero:
If you can provide replication steps (including maybe a sipp script), I'l ltry to have a look at what the issue might be. As I wrote in reply to the PR, I don't think the patch is a proper fix, since it would introduce different problems, and so I'd like to investigate a different solution.

Jonbeckas:
The replication steps are:
1. Get INVITE without sdp offer
2. Decline it with Janus
3. Get another INVITE without an offer
4. Decline it with Janus
5. Janus Sip Plugin will keep the session as establishing an deny further calls

I am not really familliar with sipp, but i try to add a sipp script later.

lminiero:
If these invites are offerless, then I don't think the core or alert states have anything to do with it: there wouldn't be any SDP to trigger a new PeerConnection establishment. It's much more likely an inconsistent state within the SIP plugin itself. I'll try to replicate and let you know.

lminiero:
I think I have a better understanding now, and why you were trying to tinker with the alert flag. It's true, as I said, that there's no PeerConnection establishment involved, but that's apparently the very root of the issue, rather than the reason why it shouldn't happen.

[...]

PR #3333, opened 2024-02-23
Fix stuck sip sessions

Jonbeckas:
Structural not compleatly clean, but it should fix issue #3332

### DIFF
--- a/src/ice.c
+++ b/src/ice.c
@@ -3571,6 +3571,14 @@ void janus_ice_setup_remote_candidates(janus_ice_handle *handle, guint stream_id
 	pc->process_started = TRUE;
 }
 
+void janus_ice_reset_webrtc_alert(janus_ice_handle *handle) {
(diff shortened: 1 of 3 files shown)


lminiero:
I haven't investigated the issue yet, but I don't think this is a proper fix. Resetting a PeerConnection state in the Janus core is not simply a matter of clearing a flag: there are steps that must be traversed. If you're not getting the PC closed, then maybe the plugin is not closing the PC, so it's the SIP plugin that should be updated, not the core.

Jonbeckas:
I don't think it is possible to fix this only in the sip_plugin, because the hangup event is not fired, because the session still has the flag `JANUS_ICE_HANDLE_WEBRTC_ALERT` when no sdp was send yet.
It's possible to fix it just in the sip_plugin if we set `session->establishing` to 0 on decline, but it would be a duplication with the later called hangup function.

PR #3339, opened 2024-03-05
Don't call close_pc in SIP plugin if there was no SDP (fixes #3332)

### LINKED ISSUE #3332: [1.x] Sip session get stuck

lminiero:
You can refer to the comments in #3332 for a more comprehensive discussion around the need for this fix.
I'll backport to `0.x` as well, once this is merged.

### DIFF
--- a/src/ice.c
+++ b/src/ice.c
@@ -1411,6 +1411,7 @@ janus_ice_handle *janus_ice_handle_create(void *core_session, const char *opaque
 	handle->queued_candidates = g_async_queue_new();
 	handle->queued_packets = g_async_queue_new();
 	janus_mutex_init(&handle->mutex);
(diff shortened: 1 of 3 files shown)


lminiero:
Merging.

PR #3340, opened 2024-03-07
Fix getBitrate()

avenal:
Since "mediaType" property seems deprecated as it was renamed to "kind" (https://developer.mozilla.org/en-US/docs/Web/API/RTCInboundRtpStreamStats), i would like to introduce small change in janus.js getBitrate function, so that it will return the correct values when called on iPhones and iPads

### DIFF
--- a/html/demos/janus.js
+++ b/html/demos/janus.js
@@ -3045,7 +3045,7 @@ function Janus(gatewayCallbacks) {
 									return;
 								let inStats = false;
 								// Check if these are statistics on incoming media
-								if((res.mediaType === "video" || res.id.toLowerCase().indexOf("video") > -1) &&
+								if((res.kind === "video" || res.id.toLowerCase().indexOf("video") > -1) &&
 										res.type === "inbound-rtp" && res.id.indexOf("rtcp") < 0) {
 									// New stats
 									inStats = true;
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @avenal! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

html/demos/janus.js
```diff
@@ -3045,7 +3045,7 @@ function Janus(gatewayCallbacks) {
 									return;
 								let inStats = false;
 								// Check if these are statistics on incoming media
-								if((res.mediaType === "video" || res.id.toLowerCase().indexOf("video") > -1) &&
+								if((res.kind === "video" || res.id.toLowerCase().indexOf("video") > -1) &&
```
lminiero, line 3048: Wouldn't it better to keep both? Like

    res.mediaType === "video" || res.kind === "video" || ...

I just want to avoid scenarios where there's still older browsers that don't support the new syntax.

avenal: Seems reasonable, I hope it won't be redundant as it was the only "mediaType" property used in the file.

lminiero:
Merging, thanks!

PR #3342, opened 2024-03-11
[0.x] janus.js: "connectionState" callback added

RSATom:
it can be useful since latest Chrome versions
doesn't assign `failed` state to ICE connection state.


lminiero:
Apologies for the delay: looks good to me! I'll merge both PRs now :+1:

PR #3343, opened 2024-03-11
janus.js: "connectionState" callback added

RSATom:
it can be useful since latest Chrome versions
doesn't assign `failed` state to ICE connection state.


lminiero:
Just an editorial note: maybe `pcState` would be a shorter and "better" name? We already shorten PeerConnection to pc in other places (e.g., pcStuff), and the other state related callbacks have a short name too. Or maybe let's just call it `connectionState`, since that's what the actual callback is?

RSATom:
I've thought about using `connectionState` but worried it can confusing for users since they can thing it's related to http/WebSockets connection. But if you think it's ok - I'll change it.

And I don't like too significant abbreviation (I'm about `pcState`) but anyway it's up to you what exactly to choose.

RSATom:
@lminiero FYI `peerConnectionState` -> `connectionState`

PR #3347, opened 2024-03-20
Avoid applying an offset to a null pointer

atoppi:
This PR addresses an undefined behavior (sum of `int` and `NULL` pointer) discovered by UBSan in the websockets transport.
Note: the sanitizer issues a runtime exception only while executing a janus binary **built with clang.**


lminiero:
Merging.

PR #3348, opened 2024-03-26
Ship speexdsp's jitter buffer as part of local AudioBridge dependencies

lminiero:
Some time ago, in #3214 we modified the AudioBridge plugin to use the jitter buffer provided by [SpeexDSP](https://gitlab.xiph.org/xiph/speexdsp).

This was quite successful and greatly improved the audio experience, but we soon found out that the library comes with a nasty memory leak under some circumstances, specifically when dealing with very late packets. Some projects addressed this by forking the library and applying a fix themselves (e.g., like [here](https://github.com/xpilot-project/speexdsp/commit/36e68e3e33f72c328da8097a6c55662e6068b221)), while we initially chose to fix this externally instead, by mimicking the same check SpeexDSP performs internally that leads to the leak (see a15e3fdfc6049927de9748d88e7fac3b19a3455d and ad326e0cd3f1debc725a3156cf20bcb20fda2323). We found out that under very specific circumstances that's not enough, though, and so we decided to just include the relevant jitter buffer files as part of the repo, specifically as part of local AudioBridge dependencies.

This allowed us to

1. get rid of the `speexdsp` dependency (we only needed the jitter buffer, not everything else);
2. fix the memory leak directly in the library.

We'll probably take advantage of this to also fix a few other slightly problematic behaviours, but my main priority at the moment is fixing the leak in a consistent and reliable manner.

Not sure if this will cause problems to existing deployments, e.g., due to symbol clashes if you don't do a `make clean`. Please make sure you test this, to ensure it works as expected and that it doesn't introduce regressions (which I doubt it will, but that's the point of testing). Should this all work nicely, I'll backport to `0.x` after merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,6 @@ A couple of plugins depend on a few more libraries:
 
 * [Sofia-SIP](https://github.com/freeswitch/sofia-sip) (only needed for the SIP plugin)
 * [libopus](https://opus-codec.org/) (only needed for the AudioBridge plugin)
-* [speexdsp](https://www.speex.org/) (only needed for AudioBridge plugin)
 * [libogg](https://xiph.org/ogg/) (needed for the VoiceMail plugin and/or post-processor, and optionally AudioBridge and Streaming plugins)
 * [libcurl](https://curl.haxx.se/libcurl/) (only needed if you are interested in RTSP support in the Streaming plugin or in the sample Event Handler plugin)
 * [Lua](https://www.lua.org/download.html) (only needed for the Lua plugin)
@@ -52,7 +51,7 @@ All of those libraries are usually available on most of the most common distribu
     yum install libmicrohttpd-devel jansson-devel \
        openssl-devel libsrtp-devel sofia-sip-devel glib2-devel \
        opus-devel libogg-devel libcurl-devel pkgconfig \
-       speexdsp-devel libconfig-devel libtool autoconf automake
+       libconfig-devel libtool autoconf automake
 
 Notice that you may have to `yum install epel-release` as well if you're attempting an installation on a CentOS machine instead.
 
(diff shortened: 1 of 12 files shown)


tmatth:
FWIW you could probably drop a lot/all of the `os_support`/`arch` macro/wrapper stuff given that:
* the jitterbuffer isn't doing DSP per se so you don't have any float/fixed-point equivalents being called AFAIK
* Janus isn't targetting the same range (or age) of platforms as speexdsp was
* Janus doesn't support custom memory allocators (so you could either use the standard `memmove` et al or glib wrappers)

lminiero:
@tmatth thanks for the suggestions! Actually, since in the #3185 PR we do use the speexdsp resampler, we may indeed some of that stuff, which is why I kept it in for now. Besides, I'd rather not touch the code too much, since I'm not very familiar with it and don't want to risk breaking stuff. The plan for now is, in following efforts:

1. merge the PR as it is, to address the main problem
2. adapt #3185 so that we bring in `resampler.c` too
3. fix a few things @atoppi wanted to fix (mainly some internal queue sizes, IIRC)
4. possibly strip the code as you suggested

For 4. later on, help from you would indeed be invaluable, considering your expertise with the library!

atoppi:
@tmatth maybe consider adding the fix related to memory leaks in upstream too.

tmatth:
> @tmatth maybe consider adding the fix related to memory leaks in upstream too.

I'd be happy to look at an MR to gitlab.xiph.org/speexdsp or if someone has a reproducer for the leak I can work on one myself.

lminiero:
> I'd be happy to look at an MR to https://gitlab.xiph.org/xiph/speexdsp or if someone has a reproducer for the leak I can work on one myself.

The patch would be as simple as the tiny change we made on our own copy. I'll see if creating a simple reproducer is easy, since the leak can be triggered by adding a very late packet, which I guess means passing a big timestamp gap.

lminiero:
Merging, I'll backport to `0.x` too shortly.

PR #3349, opened 2024-04-02
Add packet concealment for opus decoder in audiobridge

spscream:
I made clear changes for adding of plc to audiobridge, based on https://github.com/meetecho/janus-gateway/pull/3308 and to address issue with crack noises https://github.com/meetecho/janus-gateway/issues/3297


src/plugins/janus_audiobridge.c
```diff
@@ -1698,9 +1698,9 @@ typedef struct janus_audiobridge_participant {
 	OpusDecoder *decoder;		/* Opus decoder instance */
 	gboolean fec;				/* Opus FEC status */
 	int expected_loss;			/* Percentage of expected loss, to configure libopus FEC behaviour (default=0, no FEC even if negotiated) */
-	uint16_t expected_seq;		/* Expected sequence number */
 	uint16_t probation; 		/* Used to determine new ssrc validity */
 	uint32_t last_timestamp;	/* Last in seq timestamp */
+	uint16_t last_seq; /* Last sequence number */
```
atoppi, line 1703: fix comment alignment

atoppi (changes requested):
Thanks @spscream for preparing a clean PR for this feature. I've left some notes.

The main point of the discussion here is that we're basically ditching the FEC in favor of the PLC. There a couple of reasons for doing this:
1) Packet concealment is something browsers already do (and it definitely helps when there is loss), however audiobridge is terminating the RTP, so having a mechanism like this will greatly help to match audio quality experienced in sfu/p2p.
2) opus FEC is currently limited to a single packet, so it's totally useless in case of bursts of losses. Since bursts are probably the more frequent and disruptive case, PLC might have an advantage over FEC.

src/plugins/janus_audiobridge.c
```diff
@@ -8572,9 +8574,53 @@ static void *janus_audiobridge_participant_thread(void *data) {
 			if(participant->jitter) {
 				ret = jitter_buffer_get(participant->jitter, &jbp, participant->codec == JANUS_AUDIOCODEC_OPUS ? 960 : 160, NULL);
 				jitter_buffer_tick(participant->jitter);
-				if(ret != JITTER_BUFFER_OK) {
-					/* No packet in the jitter buffer? Move on the talking detection, if needed */
+
+				if(!first && participant->codec == JANUS_AUDIOCODEC_OPUS && (ret == JITTER_BUFFER_MISSING || ret == JITTER_BUFFER_INSERTION) && lost_packets_gap <= JITTER_BUFFER_MAX_GAP_SIZE) {
```
atoppi, line 8578: The previous code invoked `janus_audiobridge_participant_istalking` regardless of any other condition than `ret != JITTER_BUFFER_OK`, please restore this beahvior.

lminiero:
> we're basically ditching the FEC in favor of the PLC

I'd be against that. While it may be true it's useless for bursts > 1, it's still much better than concealment when loss is indeed 1.

src/plugins/janus_audiobridge.c
```diff
@@ -8547,8 +8548,9 @@ static void *janus_audiobridge_participant_thread(void *data) {
 	janus_audiobridge_rtp_relay_packet *mixedpkt = NULL;
 	janus_rtp_header *rtp = NULL;
 	gint64 now = janus_get_monotonic_time(), before = now;
-	gboolean first = TRUE, use_fec = FALSE;
+	gboolean first = TRUE;
 	int ret = 0;
+	int lost_packets_gap = 0;
```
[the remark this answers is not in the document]
spscream: I missed the logic for setting this to zero, you are right, thanks. I'll fix it.
Main idea of this - then webrtc client stops sending rtp media(e.g. in our case we are doing this then mic is muted on client), we still adding plc decodes and to avoid generation of it I added this condition. I could remove this logic, since this is not an expensive operation.

spscream:
> > we're basically ditching the FEC in favor of the PLC
> 
> I'd be against that. While it may be true it's useless for bursts > 1, it's still much better than concealment when loss is indeed 1.

I can add 1 fec packet recovery in case of first MISSING or INSERTION error, but It doesn't make sence too much. In our production environment fec recovery still leads to click sounds for some reason and I removed it completely. 
We use audiobridge with plc patch in production installs for our customers since last year and we had no issues with clicks at all since then.

src/plugins/janus_audiobridge.c
```diff
@@ -8572,9 +8574,53 @@ static void *janus_audiobridge_participant_thread(void *data) {
...
+						break;
+					}
+
+					int32_t output_samples;
+					opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+
+					pkt = g_malloc(sizeof(janus_audiobridge_rtp_relay_packet));
+					pkt->data = g_malloc0(OPUS_SAMPLES * (participant->stereo ? 2 : 1) * sizeof(opus_int16));
+					pkt->ssrc = 0;
+					pkt->timestamp = participant->last_timestamp + 960;
```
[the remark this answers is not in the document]
spscream: can you please explain this? isn't timestamp incemented by 960 in case of normal decode?

src/plugins/janus_audiobridge.c
```diff
@@ -8572,9 +8574,53 @@ static void *janus_audiobridge_participant_thread(void *data) {
...
+						/* This means we're cleaning up, so don't try to decode */
+						janus_audiobridge_buffer_packet_destroy(bpkt);
+						break;
+					}
+
+					int32_t output_samples;
+					opus_decoder_ctl(participant->decoder, OPUS_GET_LAST_PACKET_DURATION(&output_samples));
+
+					pkt = g_malloc(sizeof(janus_audiobridge_rtp_relay_packet));
+					pkt->data = g_malloc0(OPUS_SAMPLES * (participant->stereo ? 2 : 1) * sizeof(opus_int16));
```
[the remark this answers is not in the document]
spscream: fixed

src/plugins/janus_audiobridge.c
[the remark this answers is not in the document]
spscream: fixed

src/plugins/janus_audiobridge.c
spscream: fixed

src/plugins/janus_audiobridge.c
spscream: this is left on line 8577, jitter_buffer_get doesn't return anything other than JITTER_BUFFER_MISSING || JITTER_bUFFER_INSERTION || JITTER_BUFFER_OK

src/plugins/janus_audiobridge.c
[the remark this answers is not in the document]
atoppi: Yes it's increased by 960, that matches the `OPUS_SAMPLES` macro. This is just for uniformity with another ts increase in plugin code.
In general `OPUS_SAMPES` is the number of opus samples in a 20ms packet with 48Khz sampling. That matches the RTP clock.

src/plugins/janus_audiobridge.c
atoppi: It's not the same.
The old code does this:
```
// every 20 ms loop
ret = jitter_buffer_get(participant->jitter, &jbp, participant->codec == JANUS_AUDIOCODEC_OPUS ? 960 : 160, NULL);
jitter_buffer_tick(participant->jitter);
if(ret != JITTER_BUFFER_OK) {
	janus_audiobridge_participant_istalking(session, participant, NULL, NULL);
} else {
	//parse g711 or decode opus (with fec)
	janus_audiobridge_participant_istalking
}
```

Your patch does not cover some paths, for example a g711 packet goes always in the decode path -> WRONG, should just do `janus_audiobridge_participant_istalking`.

IMHO this should be written as:

```
if(ret != JITTER_BUFFER_OK) {
	if(!first && participant->codec == JANUS_AUDIOCODEC_OPUS && lost_packets_gap <= JITTER_BUFFER_MAX_GAP_SIZE) {
		// Opus PLC stuff
                ...
	}
	else {
		/* No packet in the jitter buffer? Move on the talking detection, if needed */
		janus_audiobridge_participant_istalking(session, participant, NULL, NULL);
	}
} else {
	/* Decode the audio packet */
        ....
}
```

src/plugins/janus_audiobridge.c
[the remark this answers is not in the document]
atoppi: > Main idea of this - then webrtc client stops sending rtp media(e.g. in our case we are doing this then mic is muted on client), we still adding plc decodes and to avoid generation of it I added this condition. I could remove this logic, since this is not an expensive operation.
I see, that's a smart thing to do indeed.
In that case I'm not sure 20 packets are enough to consider incoming data as "silenced" (they are just about 400ms of audio, a network drop like that could not be so uncommon).
I'd consider increasing this number and test if generated concealed packets do not contain glitches of some kind that disrupts the audio or trigger the talking detection.

atoppi:
I'm in general in favor of this feature since its current absence is one evidence that make p2p calls slightly better than audiobridge.
Nonetheless I agree with @lminiero with the FEC aspect. We should use FEC data when available and then switch to PLC for the rest.

>  In our production environment fec recovery still leads to click sounds for some reason and I removed it completely.

@spscream
That might be due to several reasons:
- Was the sender really attaching fec data to the packets? Some browsers only do this when loss is being noticed
- Was the loss spotty or in bursts? As we already said, FEC will do nothing against bursts
- I'd not exclude bugs in the FEC decoding code, we already patched it a couple of times

spscream:
> I'm in general in favor of this feature since its current absence is one evidence that make p2p calls slightly better than audiobridge. Nonetheless I agree with @lminiero with the FEC aspect. We should use FEC data when available and then switch to PLC for the rest.
> 
> > In our production environment fec recovery still leads to click sounds for some reason and I removed it completely.
> 
> @spscream That might be due to several reasons:
> 
> * Was the sender really attaching fec data to the packets? Some browsers only do this when loss is being noticed
> * Was the loss spotty or in bursts? As we already said, FEC will do nothing against bursts
> * I'd not exclude bugs in the FEC decoding code, we already patched it a couple of times

I'm sure fec were attached to packets, we use only latest chrome and electron clients. Losses were in bursts and spotty.

I'm not sure how to implement fec now, since if first loss occured we generate plc now and output for encoder is going monotonically. If we delay generation of plc until second MISSING or INSERTION error, output of encoders will contain click sound, since input  isn't monotonically anymore and empty input on encode will cause click sound.

Is it encoder input has some sort of queue? If so I can try to postpone plc generation on first error until normal packet, if second MISSING or INSERTION error received, I will generate 2 plcs and don't use fec on subsequent JITTER_BUFFER_OK received.

src/plugins/janus_audiobridge.c
spscream: ok, changed

atoppi (approved):

atoppi:
Code looks good now.
Maybe we should add a configuration for a room: either use PLC or FEC. I agree that mixing the two things is not an easy task.
Also @spscream mentioning a buffer for th eencoder lead me to think about the effectiveness of FEC on "almost" empty `queue-in` buffers.
What happens if we mess a packet, mixer mixes-in an empty spot, then both new packet and previous forward corrected packet are put into the queue-in mixer queue. I guess some audio disruption is the effect (the robotic/metallic voice mentioned here?).

lminiero:
@spscream FYI, we just merged a couple of important fixes in the AudioBridge, plus the (optional) support for RNNoise that we had open for quite some time. I noticed this unfortunately caused some conflicts in your PR, could you rebase on the latest changes when you can? In the next few days we plan to make some tests of your effort too :v:

spscream:
> @spscream FYI, we just merged a couple of important fixes in the AudioBridge, plus the (optional) support for RNNoise that we had open for quite some time. I noticed this unfortunately caused some conflicts in your PR, could you rebase on the latest changes when you can? In the next few days we plan to make some tests of your effort too ✌️

rebased

src/plugins/janus_audiobridge.c
```diff
@@ -8777,9 +8779,61 @@ static void *janus_audiobridge_participant_thread(void *data) {
...
+						pkt = g_malloc(sizeof(janus_audiobridge_rtp_relay_packet));
+						pkt->data = g_malloc0(BUFFER_SAMPLES * sizeof(opus_int16));
+						pkt->ssrc = 0;
+						pkt->timestamp = participant->last_timestamp + OPUS_SAMPLES;
+						pkt->seq_number = participant->last_seq + 1;
+						/* This is a redundant packet, so we can't parse any extension info */
+						pkt->silence = FALSE;
+						janus_audiobridge_participant_istalking(session, participant, NULL, NULL);
+						pkt->length = opus_decode(participant->decoder, NULL, 0, (opus_int16 *)pkt->data, output_samples, 0);
+#ifdef HAVE_RNNOISE
```
spscream, line 8808: do we need to run this on plc packets?

lminiero: I think so, because PLC packets are generated from some sort of interpolation from the encoded audio, I believe, that we may have cleaned out of noise in previous iterations after decoding it. Not denoising the PLC audio could result in discontinuities in the audio.

spscream:
@lminiero any updates on this?

atoppi:
@spscream we are adding #3368 to the audiobridge in order to cleanup the buffers in muted/unmuted scenarios.
However a patch to this PR is still needed in order to check for `participant->muted` otherwise a muted participant will trigger PLC until the max gap size is reached.

src/plugins/janus_audiobridge.c
```diff
@@ -8777,9 +8779,61 @@ static void *janus_audiobridge_participant_thread(void *data) {
 					jitter_ticks = 0;
 				}
 				jitter_buffer_tick(participant->jitter);
+
 				if(ret != JITTER_BUFFER_OK) {
-					/* No packet in the jitter buffer? Move on the talking detection, if needed */
-					janus_audiobridge_participant_istalking(session, participant, NULL, NULL);
+					if(!first && participant->codec == JANUS_AUDIOCODEC_OPUS && lost_packets_gap <= JITTER_BUFFER_MAX_GAP_SIZE) {
```
atoppi, line 8784: Please add a check here for `participant->muted`, otherwise when a participant is muted every buffer tick will generate a concealed packet.

atoppi (changes requested):

src/plugins/janus_audiobridge.c
spscream: ok, added

atoppi (approved):

atoppi:
I'm adding a "please test" tag since I think this PR is ready to be tested.
I'd like to get some feedback from folks using audiobridge and interested in enhancing the perceived audio quality.

We've tested it a bit and did not find such a remarkable difference in respect to the FEC. Nevertheless I personally think this is a better approach overall, primarily because the current opus FEC is limited to a single packet.

spscream:
@lminiero done, rebased

lminiero:
Thanks! Merging then :+1:

PR #3361, opened 2024-04-24
Add new participant mutex to VideoRoom

### LINKED ISSUE #3360: [1.x] Crash / segfault on 1.2.3
Janus version: 1203 (1.2.3) Janus commit: #3355 

No, after the previous issue we had and the fix we applied we ran it to see if it's stable.

**Was this working before?**
Since recompiling with the proposed patch following the issue #3354, it just crashed again with the out put in the gist below.

https://gist.github.com/tespio/4ac1473acb77cc7a0c8cda3ed9834a5d

**Additional context**
(https://github.com/meetecho/janus-gateway/issues/3354#issuecomment-2074432879)

lminiero:
Attempt to fix #3360, by using a new mutex aimed at better protecting the `room` property the `janus_videoroom_publisher` property has.


lminiero:
Merging.

Issue #3364, opened 2024-04-29
[1.x] videoroom: support of string_ids/string_ids_user


IgorKhomenko:
This is more a feature request rather than an issue.

There is a case when integrate Janus with 3rd party system where room id is a string, but user id is a number.

With current version of Janus, it's only possible to have ether string or number for both.

It would be very useful to have a way to configure it separately, e.g. to set `string_ids = TRUE;` and to have a separate variable for users: `string_ids_user = FALSE;`
https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_videoroom.c#L1951

I already have a custom branch with this change and wondering is it something you are interested in to have, so I could do a PR.

Thank you in advance

amnonbb:
In our case we need change for users string but leave rooms with number. Is it possible in your branch?

IgorKhomenko:
@amnonbb here is a PR https://github.com/meetecho/janus-gateway/pull/3369

lminiero:
Closing as not planned: if numbers are needed for rooms and strings for IDs, it's easier to just use strings for everything and stringify room numbers in the API, rather than introducing different variations/combinations of ID types (which may have repercussions on other plugins too since we try to keep some form of consistency across them).

PR #3369, opened 2024-05-06
video room plugin: introduced string_ids_user

IgorKhomenko:
Issue reference https://github.com/meetecho/janus-gateway/issues/3364#issuecomment-2088184440

### DIFF
--- a/conf/janus.plugin.videoroom.jcfg.sample
+++ b/conf/janus.plugin.videoroom.jcfg.sample
@@ -65,8 +65,9 @@ general: {
 									# handlers (default=true)
 
 	# By default, integers are used as a unique ID for both rooms and participants.
-	# In case you want to use strings instead (e.g., a UUID), set string_ids to true.
+	# In case you want to use strings instead (e.g., a UUID), set string_ids/string_ids_user to true.
 	#string_ids = true
+  #string_ids_user = true
 }
 
 room-1234: {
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @IgorKhomenko! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
I see many problems in this patch:

1. there's no place where you initialize `string_ids_user` from the configuration file, which is what we do for `string_ids`; this means people need to set the static value to `TRUE` and recompile if they want to use this, which makes little sense;
2. this is not backwards compatible, since if people set `string_ids` to `TRUE`, the user counterpart defaults to `FALSE` meaning user IDs will _not_ be strings, breaking people's expectations (and possibly applications);
3. it makes little sense to only offer this flexibility for user IDs; as someone else mentioned in the issue, someone may want to do it the other way around (strings for rooms, numbers for users).

As such, this patch needs a lot of fixing IMHO before we can review/test it. While it's ok to use the dedicated booleans for the checks, these should all be configurable via the configuration file, and when the new properties are not provided, the old `string_ids` should act as a fallback.

IgorKhomenko:
@lminiero

1 - this is addressed and pushed
2 - what you say make total sense. Probably we can check if `string_ids_user ` is intentionally presented in config (either set to true or false) and if not - then use `string_ids ` as fallback. 
3 - did not get the point. With this patch there is now a way to separately configure it for rooms and users, e.g. it can be any combination, e.g. `true/true` or `true/false` or `false/true` or `false/false`

lminiero:
@IgorKhomenko did you have a chance to revisit the PR with the changes I suggested?

lminiero:
Hi @IgorKhomenko, did you close the PR to create a different one, or are you not interested in adding the feature anymore? I was waiting for feedback but didn't get any. Thanks!

PR #3374, opened 2024-05-11
Add note about freeing transaction in handle_message

vincentfretin:
Add note about freeing transaction after calling `push_event` in plugin interface documentation.

The only place this was mentioned seems to be in this comment in the `janus.c`
https://github.com/meetecho/janus-gateway/blob/9be6a74452e27c5ae295f39dd1e2d710c25e44df/src/janus.c#L1759
and in plugin implementation like
https://github.com/meetecho/janus-gateway/blob/9be6a74452e27c5ae295f39dd1e2d710c25e44df/src/plugins/janus_textroom.c#L1261-L1265

We had a memory leak for this in the rust sfu plugin I fixed in
https://github.com/networked-aframe/janus-plugin-sfu/pull/11/commits/4226d0091b386576d221988c76e11c0d0ef7d215
that only showed in ASan output when specifying loop_events option.
The original creator of the plugin probably overlooked this because this was not documented in `plugin.h`.

### DIFF
--- a/src/plugins/plugin.h
+++ b/src/plugins/plugin.h
@@ -283,10 +283,10 @@ struct janus_plugin {
 	 * @param[out] error An integer that may contain information about any error */
 	void (* const create_session)(janus_plugin_session *handle, int *error);
 	/*! \brief Method to handle an incoming message/request from a peer
-	 * @note The Janus core increases the references to both the \c message and \c jsep
-	 * json_t objects. This means that you'll have to decrease your own
+	 * @note The Janus core leaves ownership of both the \c message and \c jsep
+	 * json_t objects to plugins. This means that you'll have to decrease your own
 	 * reference yourself with a \c json_decref when you're done with them.
-	 * You'll also have to free \c transaction with \c g_free.
+	 * You'll also have to free \c transaction with \c g_free
 	 * @param[in] handle The plugin/gateway session used for this peer
 	 * @param[in] transaction The transaction identifier for this message/request
 	 * @param[in] message The json_t object containing the message/request JSON
@@ -360,7 +360,7 @@ struct janus_callbacks {
 	/*! \brief Callback to push events/messages to a peer
 	 * @note The Janus core increases the references to both the \c message and \c jsep
 	 * json_t objects. This means that you'll have to decrease your own
-	 * reference yourself with a \c json_decref after calling \c push_event.
+	 * reference yourself with a \c json_decref after calling \c push_event
 	 * @param[in] handle The plugin/gateway session used for this peer
 	 * @param[in] plugin The plugin instance that is sending the message/event
 	 * @param[in] transaction The transaction identifier this message refers to
--- a/src/plugins/plugin.h
+++ b/src/plugins/plugin.h
@@ -283,6 +283,10 @@ struct janus_plugin {
 	 * @param[out] error An integer that may contain information about any error */
 	void (* const create_session)(janus_plugin_session *handle, int *error);
 	/*! \brief Method to handle an incoming message/request from a peer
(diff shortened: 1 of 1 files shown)


lminiero:
> Add note about freeing transaction after calling `push_event` in plugin interface documentation.

This is incorrect. You only need to free `transaction` if you allocated it yourself, since `push_event` will treat it as a const. None of the existing plugins free it for instance, since we pass an existing transaction from a transaction, which typically comes from an incoming request. The transaction you must free is the one you get from the core via `handle_message` (which all core plugins do, and maybe you're not doing).

vincentfretin:
You're right, it's the one from handle_message I'm freeing now. I put the comment in the wrong place then, I'll update the PR.

vincentfretin:
I moved the comment at the right place.

src/plugins/plugin.h
```diff
@@ -283,6 +283,10 @@ struct janus_plugin {
 	 * @param[out] error An integer that may contain information about any error */
 	void (* const create_session)(janus_plugin_session *handle, int *error);
 	/*! \brief Method to handle an incoming message/request from a peer
+	 * @note The Janus core increases the references to both the message and jsep
```
lminiero, line 286: Please add a `\c` before both `message` and `jsep`.

src/plugins/plugin.h
```diff
@@ -283,6 +283,10 @@ struct janus_plugin {
...
+	 * json_t objects. This means that you'll have to decrease your own
+	 * reference yourself with a \c json_decref when you're done with it.
```
lminiero, line 288: "with them" (since there's two).

src/plugins/plugin.h
```diff
@@ -283,6 +283,10 @@ struct janus_plugin {
...
+	 * You'll also have to free transaction with \c g_free.
```
lminiero, line 289: This is missing the `\c` before `transaction` I mentioned in my previous review.

lminiero (commented):
Thanks, added some notes.

vincentfretin:
I made the changes. I also added the `\c` for the note in push_event where I mainly copied the text from.

src/plugins/plugin.h
```diff
@@ -283,6 +283,10 @@ struct janus_plugin {
 	 * @param[out] error An integer that may contain information about any error */
 	void (* const create_session)(janus_plugin_session *handle, int *error);
 	/*! \brief Method to handle an incoming message/request from a peer
+	 * @note The Janus core increases the references to both the \c message and \c jsep
```
lminiero, line 286: I think that, since in this case this is a method that goes from core to plugin, it's more correct to say something like "The Janus core leaves ownership of both ...... objects to plugins. Ths means..". The "increases" part makes more sense for `push_event` since in that case the core does indeed increase the references to those objects, since plugins may decrease their own.

src/plugins/plugin.h
```diff
@@ -283,6 +283,10 @@ struct janus_plugin {
...
+	 * json_t objects. This means that you'll have to decrease your own
+	 * reference yourself with a \c json_decref when you're done with them.
+	 * You'll also have to free \c transaction with \c g_free.
```
lminiero, line 289: Better remove the dot at the end, after `g_free`, since otherwise it will appear as part of the `code` block. Besides, we don't have it at the end of lines in most of the other doxygen comments.

lminiero (commented):
Thanks! I added just a couple of minor notes after re-reading the changes, but after those I think this is good to be merged.

vincentfretin:
That's valid feedback. I did the changes.

lminiero:
Thanks! Merging :+1:

PR #3377, opened 2024-05-16
Added UMD

redexp:
No braking changes, but now this file can be treat as module for commonjs or require.js or have regular global Janus var

I used this [UMD](https://github.com/umdjs/umd) template https://github.com/umdjs/umd/blob/master/templates/returnExports.js

### DIFF
--- a/html/demos/janus.js
+++ b/html/demos/janus.js
@@ -24,7 +24,7 @@
 	OTHER DEALINGS IN THE SOFTWARE.
  */
 
-/*eslint indent: ["error", "tab", { "ignoredNodes": ["CallExpression[callee.type=FunctionExpression][arguments.length=2] > FunctionExpression.arguments > BlockStatement"] }]*/
+/*eslint indent: ["warn", "tab", { "ignoredNodes": ["CallExpression[callee.type=FunctionExpression][arguments.length=2] > FunctionExpression.arguments > BlockStatement"] }]*/
 (function (root, factory) {
 	/* global define, module */
 	if (typeof define === 'function' && define.amd) {
--- a/html/demos/janus.js
+++ b/html/demos/janus.js
@@ -401,27 +401,27 @@ Janus.init = function(options) {
 		} else if(Array.isArray(options.debug)) {
 			for(let d of options.debug) {
 				switch(d) {
-					case "trace":
-						Janus.trace = console.trace.bind(console);
-						break;
-					case "debug":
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @redexp! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

redexp:
Another pros - you can define as many private var as you need. Like that var `defaultExtension` that I assumed should be private, but right now it's in global scope and could be overwritten in some other file.

redexp:
linter errors on `typeof define` I think if I change it to `typeof root.define` then it should pass linter

atoppi (commented):
Please address the linting issues.

I see nothing wrong about the PR.
If @lminiero is fine with moving the `Janus` definition from the top level to an immediately invoked function then we're good to go.

redexp:
@atoppi 
```
    28:38  error    'define' is not defined                       no-undef
    29:3   error    'define' is not defined                       no-undef
    30:43  error    'module' is not defined                       no-undef
    31:3   error    'module' is not defined                       no-undef
```

atoppi:
@redexp yeah, please fix those and also the issues about indentation.
If you wanna test locally before triggering our github workflow, just use `eslint janus.js` with our eslint configuration file (`.eslintrc.js`).

redexp:
@atoppi I'v added rule to ignore one indent in main part code. It's ok practice with UMD, eslint itself [shows example](https://eslint.org/docs/latest/rules/indent#ignorednodes) how to do it

atoppi:
Merging, I'll figure out how to better fix the eslint stuff by myself.

atoppi:
@redexp add to do a bunch of modifications due to the broken rollup script after your merge.
Can you test if this still works for you after the commit above?

atoppi:
Unfortunately [we have a rollup script ](https://github.com/meetecho/janus-gateway/blob/493a97bd0a5253d11f4a3633be93df80b801833c/npm/rollup.config.mjs)that bundles `janus.js` in an es module.
It basically takes [this template](https://github.com/meetecho/janus-gateway/blob/493a97bd0a5253d11f4a3633be93df80b801833c/npm/module.js), replaces `@JANUS_CODE@` with the content of `janus.js` and exports a `Janus` object, hence we need that object to be explicitly declared by `janus.js`.

PR #3378, opened 2024-05-16
Pull out all helper methods

redexp:
There three helper methods which was in `init` just because of using `Janus.error` or `Janus.warn` but we can pull out those log methods too with default `noop` values. 

Also added returns of cached errors so user can do extra checks like 
```js
const err = Janus.attachMediaStream(el, stream);

if (err) {
  // ...
}
```

### DIFF
--- a/html/demos/janus.js
+++ b/html/demos/janus.js
@@ -338,6 +338,14 @@ Janus.trackConstraints = function(track) {
 
 Janus.noop = function() {};
 
+// Console logging (all debugging disabled by default)
+Janus.trace = Janus.noop;
+Janus.debug = Janus.noop;
+Janus.vdebug = Janus.noop;
+Janus.log = Janus.noop;
+Janus.warn = Janus.noop;
+Janus.error = Janus.noop;
+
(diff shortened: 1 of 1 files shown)


lminiero:
I guess it does make sense, it's static enough stuff that it can stay outside. I guess it's not a problem that `isGetUserMediaAvailable` is defined after `listDevices` which uses it, right? I'm not so sure about the return value, instead: this seems like it could be considered an inconsistency, if some methods to it and other don't (and I'd rather not modify all of them).

redexp:
about `isGetUserMediaAvailable` - yes, it's not a problem.

about returning value - I'll remove it

redexp:
@lminiero done

lminiero:
Please rebase your PR, as there's been a considerable refactoring of `janus.js` in #3377 and 493a97bd0a5253d11f4a3633be93df80b801833c

lminiero:
@redexp did you have a chance to rebase as mentioned? The PR as it is right now has a conflict.

PR #3384, opened 2024-05-23
Bring back switches "-V" and "--version".

### LINKED ISSUE #3383: [1.x] Commandline option "--version" no longer exists
```
Janus version: 1203 (1.2.3)
Janus commit: bc044995a94c623b747bb35ecb9f059190903803
```

no

**Was this working before?**
This is working with the 0.x branch.

**Additional context**
Looks like this got removed by the migration away from gengetopt which added a handler for `--version` automatically.

We used this on our build systems as a simple check to see if all dependencies are available, so it would be great if this could be re-added. I'll check if I find some time the next days to prepare a PR myself.

fancycode:
Fixes #3383

### DIFF
--- a/src/janus.c
+++ b/src/janus.c
@@ -4421,6 +4421,11 @@ gint main(int argc, char *argv[]) {
 	if(!janus_options_parse(&options, argc, argv))
 		exit(1);
 
+	if(options.print_version) {
+		janus_options_destroy();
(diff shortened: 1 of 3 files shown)


atoppi (approved):
LGTM

lminiero:
Thanks, merging!

PR #3386, opened 2024-05-31
SIP plugin: add logger callback function for processing SIP message traces in a single log callback for Sofia >= 1.13

### LINKED ISSUE #3365: [1.x] Increased event handlers SIP logging buffer size one more time or make it configurable
v1.2.2

-

**Was this working before?**
No.

-

**Additional context**
After enabling the Event Handlers mechanism, I started receiving numerous error messages indicating log truncation. Here's a sample of the log errors:
```
...
2024-05-01T20:31:12.176557169Z Creating new handle in session 1508735635354590: 7576670150389404; 0x7fa3592ec7f0 0x7fa3592d6490                                              
2024-05-01T20:31:12.280057472Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072                                                                           
2024-05-01T20:31:12.373323775Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072                                                                           
2024-05-01T20:31:12.373323775Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072                                                                           
2024-05-01T20:31:12.434606151Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072                                                                           
2024-05-01T20:31:12.434606151Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072                                                                           
2024-05-01T20:31:12.434606151Z Registering as a helper                                                                                                                       
2024-05-01T20:31:12.473137075Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072                                                                           
2024-05-01T20:31:12.583697369Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072                                                                           
2024-05-01T20:31:12.651736579Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072                                                                           
2024-05-01T20:31:12.651736579Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072
2024-05-01T20:31:12.651736579Z [ERR] [utils.c:janus_strlcat:278] Truncation occurred, 3322 >= 3072  
...
```
It appears that the current buffer size of 3072 bytes is insufficient under certain environmental conditions, leading to frequent log truncation errors. A similar issue already was solved within #3060.

Would it be reasonable to make the buffer size for [`sofia_log`](https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_sip.c#L1688) configurable to accommodate different operational needs?
Alternatively, could the buffer size be increased once more, perhaps to 4096 bytes, to prevent truncation?

ycherniavskyi:
Fixes #3365.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1829,18 +1829,16 @@ static void janus_sip_sofia_logger_siptrace_single_log_callback(void *stream, ch
 		if(strstr(buffer, "OPTIONS") == buffer)
 			return;
 		/* Search for the Call-ID header and extract its value if it exists */
-		{
-			char *from = strstr(buffer, "Call-ID: ");
-			if(from) {
-				char *start = from + 9;
-				if(start) {
-					char *end = strchr(start, '\n');
-					if(end) {
-						size_t count = end - start;
-						count = count >= sizeof(call_id) ? sizeof(call_id) - 1 : count;
-						memcpy(call_id, start, count);
-						call_id[count] = '\0';
-					}
+		char *from = strstr(buffer, "Call-ID: ");
+		if(from) {
+			char *start = from + 9;
+			if(start) {
+				char *end = strchr(start, '\n');
+				if(end) {
+					size_t count = end - start;
+					count = count >= sizeof(call_id) ? sizeof(call_id) - 1 : count;
+					memcpy(call_id, start, count);
+					call_id[count] = '\0';
 				}
 			}
 		}
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1863,7 +1863,6 @@ static void janus_sip_sofia_logger_siptrace_single_log_callback(void *stream, ch
 							count = count >= sizeof(call_id) ? sizeof(call_id) - 1 : count;
 							memcpy(call_id, start, count);
 							call_id[count] = '\0';
-
 							session = g_hash_table_lookup(identities, call_id);
 						}
 					}
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1688,7 +1688,7 @@ static void janus_sip_parse_custom_contact_params(json_t *root, char *custom_par
 char sofia_log[3072];
 char call_id[255];
 gboolean skip = FALSE, started = FALSE, append = FALSE;
-static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
+static void janus_sip_sofia_logger_siptrace_multi_log_callback(void *stream, char const *fmt, va_list ap) {
 	if(!fmt)
 		return;
 	char line[255];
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for reviving this! I'll review in the next few days and get back to you :+1: 
Have you tested this with both 1.12 and 1.13?

ycherniavskyi:
I've specifically tested on version 1.13. And in current implementation logic for version 1.12 was not changed at all.

src/plugins/janus_sip.c
```diff
@@ -1801,6 +1801,95 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+		if(!buffer) {
+			JANUS_LOG(LOG_WARN, "No newline after the separator in this message, dropping it...\n");
+			return;
+		}
+		buffer++;
+		/* If this is an OPTIONS, we don't care: drop it */
+		if(strstr(buffer, "OPTIONS") == buffer)
+			return;
+		/* Search for the Call-ID header and extract its value if it exists */
+		{
```
lminiero, line 1832: What's this bracket doing here? There's no condition. Please get rid of it and properly indent the following code.

src/plugins/janus_sip.c
```diff
@@ -1801,6 +1801,95 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
 	}
 }
 
+static void janus_sip_sofia_logger_siptrace_single_log_callback(void *stream, char const *fmt, va_list ap) {
+	/* Since the fmt format string in the current Sofia SIP tport_log_msg function implementation is just "%s\n",
+	 * it's more efficient to directly work with the siptrace buffer. */
+	char *buffer = va_arg(ap, char *);
+
+	/* Check if this is a message we need */
+	if(buffer[0] == 's' && buffer[1] == 'e' && buffer[2] == 'n' && buffer[3] == 'd' && buffer[4] == ' ') {
```
lminiero, line 1810: You're accessing 5 bytes of a buffer that hasn't been validated. I understand that this comes from Sofia, and so it's likely good, but still I'd rather see a `strlen` done on `buffer` before, so that we can check if there are indeed enough bytes for your check.

src/plugins/janus_sip.c
```diff
@@ -1801,6 +1801,95 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+		/* Skip the separator line */
+		buffer = strchr(buffer, '\n');
+		if(!buffer) {
+			JANUS_LOG(LOG_WARN, "No newline after the separator in this message, dropping it...\n");
+			return;
+		}
+		buffer++;
+		/* If this is an OPTIONS, we don't care: drop it */
+		if(strstr(buffer, "OPTIONS") == buffer)
+			return;
```
lminiero, line 1830: Why? This callback is meant to pass _ALL_ SIP messages to event handlers, as long as they belong to a session. This OPTIONS may be part of a dialog.

src/plugins/janus_sip.c
```diff
@@ -1801,6 +1801,95 @@ static void janus_sip_sofia_logger(void *stream, char const *fmt, va_list ap) {
...
+					char *start = strstr(from + 4, "<");
+					if(start) {
+						start++;
+						char *end = strstr(from, ">");
+						if(end) {
+							size_t count = end - start;
+							count = count >= sizeof(call_id) ? sizeof(call_id) - 1 : count;
+							memcpy(call_id, start, count);
+							call_id[count] = '\0';
+
```
lminiero, line 1868: Unneeded empty line here.

lminiero (commented):
Thanks, I've added some notes inline. Revisiting the logger, I'm starting to wonder whether it's safe to use the static `call_id` as we do here: are we sure the logger is always invoked by the same thread? Have you checked this as part of your tests?

src/plugins/janus_sip.c
ycherniavskyi: This logic was adapted from the original `janus_sip_sofia_logger` handler function [here](https://github.com/meetecho/janus-gateway/blob/ea2799743d49938077c8267d85af1cf79777d3c5/src/plugins/janus_sip.c#L1778-L1781). I included it here to maintain consistency with the original implementation. If you approve of having different logic in logger handlers, I can remove it.

src/plugins/janus_sip.c
ycherniavskyi: I used this bracket to group the `Call-ID` finding logic and limit the scope of the `from` local variable.
I will remove it as per your request.

src/plugins/janus_sip.c
ycherniavskyi: Sure, I'll remove this empty line.
I had originally used it to visually separate these two code blocks.

src/plugins/janus_sip.c
ycherniavskyi: Understood. I will add a `strlen` check on `buffer`.

ycherniavskyi: Alternatively, to maintain consistency with other parts of the code, we can use the `strstr` function instead of following the logic of the original logger handler here.

src/plugins/janus_sip.c
ycherniavskyi: @lminiero waiting for your confirmation to remove this condition to drop OPTIONS messages.
However, it’s worth noting that in my cases, the PBX-es sends numerous automated without any meaning OPTIONS messages (roughly every 2-5 seconds) within each SIP session, likely to maintain NAT ports. This was probable the reason they were originally dropped in the logger handler.

ycherniavskyi: I completely agree with you. Therefore, I will leave the current logic unchanged for now.

lminiero:
@ycherniavskyi were you waiting for more feedback from me, or have all notes been addressed in the commits that followed?

ycherniavskyi:
@lminiero I've addressed all your comments except for the one about dropping OPTIONS, which we decided to leave (assuming I understood you correctly). Therefore, from my perspective, this PR is ready to merge! 😊

lminiero:
Thanks, merging then! I'll also backport this to `0.x` :+1:

PR #3391, opened 2024-06-09
AudioBridge-List_Announcements

keremcadirci:
### Requirement:
As of today, we need the `file_id` of a playback audio file (announcement) to stop playing.
A random admin may not know the `file_id` to stop it

### Proposol
Add a list of `announcements` into `listparticipants` api
in that case the response of the `listparticipants` api would be:  

```
{
	"audiobridge" : "participants",
	"participants" : [		// Array of participant objects
		{	// Participant #1
			"id" : <unique numeric ID of the participant>,
			// ....
		},
		// Other participants
	],
	"announcements" : [		// Array of announcement objects
		{	// Announcement #1
			"file_id" : "<unique string ID of the announcement>",
			"filename": "<path to the Opus file to play>",
			"playing" : <true|false, whether or not the file is playing>,
			"loop": <true|false, depending on whether or not the file is playing in a loop forever>
		}
		// Other announcements
	]
}
```


lminiero:
While this is an interesting requirement, I wouldn't add this to the list of participants. It's true that we highjacked the participants structure for announcements, but they are _not_ really participants. I'd much rather see a new "listannouncements" request or something like this, pretty much as we do for forwarders. This would allow us to differentiate announcements, and also protect them separately: to do a `listforwarders`, for instance, you need to know the room secret if configured, while this is not true for participants, and I think announcements are something only admins with access to the secret should have access to.

keremcadirci:
@lminiero  `listannouncements` api added
- Request:
```
{
	"request" : "listannouncements",
	"room" : <unique numeric ID of the room>
}
```
- Response:
```
{
	"audiobridge" : "announcements",
	"announcements" : [		// Array of announcement objects
		{	// Announcement #1
			"file_id" : "<unique string ID of the announcement>",
			"filename": "<path to the Opus file to play>",
			"playing" : <true|false, whether or not the file is playing>,
			"loop": <true|false, depending on whether or not the file is playing in a loop forever>
		}
		// Other announcements
	]
}
```

lminiero (commented):
Thanks! Added some notes inline.

lminiero:
Thanks for the fixes, merging!

PR #3392, opened 2024-06-09
AudioBridge-stop_all_files

keremcadirci:
### Requirement:
One or more announcement from file can be played by an admin in the audiobridge room
An admin without knowing `file_id`s may want to stop all announcements.
Currently we need to know all `file_id`s and we need call many `stop_file` api to stop all annoucements.

### Proposol
A new `stop_all_files` api that will stop all playing announcements.
A `announcement-stopped` event will be fired for all stopped announcements
`stop_all_files` api will return stopped `file_id` list

- API Request:
```
{
	"request" : "stop_all_files",
	"room" : <unique numeric ID of the room where the playback is taking place>,
	"secret" : "<room password, if configured>"
}
```

- API Responce:  
```
{
	"audiobridge" : "success",
	"room" : <unique numeric ID, same as request>,
	"file_id_list" : [
		// Array of file_Id: "<unique string ID of the now interrupted announcement>""
		]
}
```


src/plugins/janus_audiobridge.c
```diff
@@ -774,6 +775,29 @@ room-<unique room ID>: {
...
+ *
+ * A successful request will result in a \c success response:
+ *
+\verbatim
+{
+	"audiobridge" : "success",
+	"room" : <unique numeric ID, same as request>,
+	"file_id_list" : [
+		// Array of file_Id: "<unique string ID of the now interrupted announcement>""
+		]
```
lminiero, line 799: Indentation seems broken here (there's one extra tab).

src/plugins/janus_audiobridge.c
```diff
@@ -5559,6 +5583,131 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+		if(audiobridge->destroyed) {
+			janus_mutex_unlock(&audiobridge->mutex);
+			janus_mutex_unlock(&rooms_mutex);
+			JANUS_LOG(LOG_ERR, "No such room (%s)\n", room_id_str);
+			error_code = JANUS_AUDIOBRIDGE_ERROR_NO_SUCH_ROOM;
+			g_snprintf(error_cause, 512, "No such room (%s)", room_id_str);
+			goto prepare_response;
+		}
+
+		/* Get list of started announcements and send a stop announcement notification*/
```
lminiero, line 5657: There's a missing space at the end of the sentence (the asterisk of the comment is right next to the last word).

src/plugins/janus_audiobridge.c
```diff
@@ -5559,6 +5583,131 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+					json_object_set_new(info, "file_id", json_string(p->annc->id));
+					gateway->notify_event(&janus_audiobridge_plugin, NULL, info);
+				}
+				items_to_remove = g_list_prepend(items_to_remove, p->annc->id);
+			}
+			if(p)
+				janus_refcount_decrease(&p->ref);
+		}
+
+		/* Get rid of the announcements */
```
lminiero, line 5691: I don't understand why this is a separate block. You were already looping on all announcements before, so that's where you can populate the array too.

src/plugins/janus_audiobridge.c
```diff
@@ -5559,6 +5583,131 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+		json_t *listAnncRemoved = json_array();
```
lminiero, line 5692: Code style: we don't use camel case for variables.

src/plugins/janus_audiobridge.c
```diff
@@ -5559,6 +5583,131 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+		for (GList *l = items_to_remove; l!= NULL; l = l->next) {
```
lminiero, line 5693: Code style: don't define variables in the `for` itself. Besides, please don't add an extra space between directive and bracket (it's `for(`, not `for (`).

src/plugins/janus_audiobridge.c
```diff
@@ -5559,6 +5583,131 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+			json_t *file_id = json_string(l->data);
+		    if( g_hash_table_remove(audiobridge->anncs, l->data))
+			{
```
lminiero, line 5696: Code style: same for the `if`, there's an extra space that shouldn't be there before `g_hash_table_remove`.
The bracket of the check should be on the same line as the check, not below as you did here (`if(..) {`)

lminiero (commented):

src/plugins/janus_audiobridge.c
keremcadirci: I don't like either but since we are removing an element (`g_hash_table_remove`) inside audiobridge->anncs inside the loop, `while` loop above crashes with `g_hash_table_iter_next: assertion 'ri->version' == ri->hash_table->version....` error.
Sorry, I'm not expert on C and hash_table, I'm open other suggestions
I can copy `audiobridge->anncs`  into another hashtable changing the ref and while looping the new hashtable

lminiero:
Did you mean to base this upon your "listannouncements" branch? I see changes coming from that PR too, in this branch.

keremcadirci:
A cleaner version of this PR is created : #3403
Please continue this PR on the new PR #3403

PR #3403, opened 2024-07-08
AudioBridge stop_all_files api

keremcadirci:
The continuation of the PR #3392
This is a cleaner PR with all #3392 commits are merged into a single commit.
All Conflicts are solved

### DIFF
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -78,8 +78,8 @@ room-<unique room ID>: {
  * \c create, \c edit, \c destroy, \c exists, \c allowed, \c kick, \c list,
  * \c mute, \c unmute, \c mute_room, \c unmute_room, \c listparticipants,
  * \c listannouncements, \c resetdecoder, \c rtp_forward, \c stop_rtp_forward,
- * \c list_forwarders, \c play_file, \c is_playing and \c stop_file are
- * synchronous requests, which means you'll get a response directly within
+ * \c list_forwarders, \c play_file, \c is_playing, \c stop_file and \c stop_all_files
+ * are synchronous requests, which means you'll get a response directly within
  * the context of the transaction. \c create allows you to create a new audio
  * conference bridge dynamically, as an alternative to using the configuration file;
  * \c edit allows you to dynamically edit some room properties (e.g., the PIN);
@@ -100,7 +100,8 @@ room-<unique room ID>: {
  * \c list_forwarders request; finally, \c play_file allows you to
  * reproduce an audio .opus file in a mix (e.g., to play an announcement
  * or some background music), \c is_playing checks if a specific file is
- * still playing, while \c stop_file will stop such a playback instead.
+ * still playing, while \c stop_file will stop such a playback instead and
+ * \c stop_all_files will stop all announcements.
  *
  * The \c join , \c configure , \c changeroom and \c leave requests
  * instead are all asynchronous, which means you'll get a notification
@@ -720,7 +721,8 @@ room-<unique room ID>: {
  * A similar event is also sent whenever the playback stops, whether it's
  * because the file ended and \c loop was \c FALSE (which will automatically
  * clear the resources) or because a \c stop_file request asked for the
- * playback to be interrupted:
+ * playback to be interrupted or \c stop_all_files request asked for all
+ * playbacks to be interrupted:
  *
 \verbatim
 {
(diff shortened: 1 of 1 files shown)


src/plugins/janus_audiobridge.c
```diff
@@ -5663,6 +5688,131 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+			JANUS_LOG(LOG_ERR, "No such room (%s)\n", room_id_str);
+			error_code = JANUS_AUDIOBRIDGE_ERROR_NO_SUCH_ROOM;
+			g_snprintf(error_cause, 512, "No such room (%s)", room_id_str);
+			goto prepare_response;
+		}
+
+		/* Get list of started announcements and send a stop announcement notification */
+		GList *items_to_remove = NULL;
+		GHashTableIter iterAnnc;
+		gpointer valueAnnc;
```
lminiero, line 5765: Code style: we don't use camel style for variable names. Please rename those variables (e.g., `iter_annc` and `value_annc`, or even just `iter` and `value` as we do pretty much everywhere else).

src/plugins/janus_audiobridge.c
```diff
@@ -5663,6 +5688,131 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+			if(p)
+				janus_refcount_increase(&p->ref);
+			if(started) {
+				JANUS_LOG(LOG_INFO, "[%s] Announcement stopped (%s)\n", audiobridge->room_id_str, p->annc->id);
+				json_t *event = json_object();
+				json_object_set_new(event, "audiobridge", json_string("announcement-stopped"));
+				json_object_set_new(event, "room",
+				string_ids ? json_string(audiobridge->room_id_str) : json_integer(audiobridge->room_id));
+				json_object_set_new(event, "file_id", json_string(p->annc->id));
+				janus_audiobridge_notify_participants(p, event, TRUE);
```
lminiero, line 5779: Why are you passing `TRUE` here? That's `notify_source_participant`, which should be `FALSE` in this case because the `janus_audiobridge_participant` instance you're passing to the function is an announcement, and so not a real participant with a Janus API session we can notify about stuff.

src/plugins/janus_audiobridge.c
```diff
@@ -5663,6 +5688,131 @@ static json_t *janus_audiobridge_process_synchronous_request(janus_audiobridge_s
...
+				/* Also notify event handlers */
+				if(notify_events && gateway->events_is_enabled()) {
+					json_t *info = json_object();
+					json_object_set_new(info, "event", json_string("announcement-stopped"));
+					json_object_set_new(info, "room",
+						string_ids ? json_string(audiobridge->room_id_str) : json_integer(audiobridge->room_id));
+					json_object_set_new(info, "file_id", json_string(p->annc->id));
+					gateway->notify_event(&janus_audiobridge_plugin, NULL, info);
+				}
+				items_to_remove = g_list_prepend(items_to_remove, p->annc->id);
```
lminiero, line 5790: Not sure if we discussed this in the original PR, but I don't think there's any need for the separate list to track instances to remove. You can use `g_hash_table_iter_remove` within a `while(g_hash_table_iter_next(..))` loop. We do it in a few other places already (e.g., `janus.c`) if you want to see how that works.

lminiero (commented):
Apologies for the late reply, I'm catching up with pending activities after the summer holidays. I've added a couple of notes inline.

src/plugins/janus_audiobridge.c
keremcadirci: Hi @lminiero 

- `stop_file` audiobridge sends `announcement-stopped` event. 
https://github.com/keremcadirci/janus-gateway/blob/65a23f56da9788bd4d2a733f5f1f613ff0ff489d/src/plugins/janus_audiobridge.c#L5668

- `play_file` results with` announcement-started`

How should we continue? Should we change `stop_file` and play_file to `FALSE` too

My own preference is to leave `TRUE` since in we want to notify participants about these events

src/plugins/janus_audiobridge.c
keremcadirci: [bb753c5](https://github.com/meetecho/janus-gateway/pull/3403/commits/bb753c55cb45bc7670a6ef73aacc595bfefb1c7a)

src/plugins/janus_audiobridge.c
keremcadirci: [bb753c5](https://github.com/meetecho/janus-gateway/pull/3403/commits/bb753c55cb45bc7670a6ef73aacc595bfefb1c7a)

keremcadirci:
@lminiero 
The 2 conversation above are resolved in the fallowing commit:
[bb753c5](https://github.com/meetecho/janus-gateway/pull/3403/commits/bb753c55cb45bc7670a6ef73aacc595bfefb1c7a)

Waiting your comments on the unresolved conversation

I have tested the commit with multiple simultaneous `play_file` and later `stop_all`, it seems OK

src/plugins/janus_audiobridge.c
lminiero: Apologies for the awfully late review, I've only now started to look back at file playout after adding support for it in Janode.

> My own preference is to leave `TRUE` since in we want to notify participants about these events

That's not what `TRUE` means there. That flag is to notify the participant pass as the first argument _too_, besides everyone else in the room.

Since you're using `janus_audiobridge_notify_participants` to notify everyone in the room, and that as the source participant you're passing an announcement instance (as `announcements` is what you're iterating on), then the last argument passed to that function must be `FALSE`, as otherwise the `janus_audiobridge_notify_participants` function wil try to notify that participant as well, which being a fake participant (an announcement) makes no sense.

But I see now that we're indeed doing the same thing in `stop_file` (which I guess is where you took it from?), so it was my mistake in the first place. I guess we can keep it as it is for now then (we probably have checks in place not to send any event for fake participants anyway), and I can fix that later on in both instances.

PR #3407, opened 2024-07-15
AudioBridge destroyed doc updated

keremcadirci:
AudioBridge `destroy` message response `destroyed` documentation updated


lminiero:
Apologies for the late reply, we were away due to the holidays. Thanks for the fix, merging!

PR #3412, opened 2024-07-25
Fix handling of "data" stream parameters in the streaming plugin

### LINKED ISSUE #3411: [1.x] databuffermsg/buffermsg not working for streaming plugin
1.2.3

No, currently this is the latest version.

**Was this working before?**
I do not know

No

**Additional context**
Last message is never sent, even when enabling buffermsg on data stream.

I believe this bug is has two different causes:

https://github.com/meetecho/janus-gateway/blob/0d1da91daebb1afcc500b4fff63bef8fb846dfbf/src/plugins/janus_streaming.c#L2182-L2187

I think buffermsg container should be *m* instead of *cat*. This is not a big issue because there is a workaround: to set `buffermsg = true` in the root of the mountpoint in the configuration file (*dataype* is also affected by the same problem, and it should be *type* instead) 

https://github.com/meetecho/janus-gateway/blob/0d1da91daebb1afcc500b4fff63bef8fb846dfbf/src/plugins/janus_streaming.c#L9949-L9961

*pkt* is created and filled with the contents of the last relay packet, but then it is not used for anything else (causing, I believe, a memory leak) I suppose the idea was to copy the data in last_msg (after freeing the previous last_msg). Finally, I think that *packet* in the lines 9955 to 9958 should be *pkt*

atoppi:
Attempt to fix #3411

### DIFF
--- a/src/plugins/janus_streaming.c
+++ b/src/plugins/janus_streaming.c
@@ -1118,6 +1118,7 @@ static struct janus_json_parameter rtp_media_parameters[] = {
 	{"port3", JANUS_JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"svc", JANUS_JSON_BOOL, 0},
 	/* Data only */
+	{"datatype", JANUS_JSON_STRING, 0},
 	{"buffermsg", JANUS_JSON_BOOL, 0},
 };
 static struct janus_json_parameter rtp_audio_parameters[] = {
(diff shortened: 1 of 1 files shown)


atoppi:
Converted the PR into draft since the changes have become more error prone and I'd like to have a review first.

In the current status the sending of the buffered message happens in the `data_ready` callback. In the callback we iterate the session, looking for streams that have the `buffermsg` flag set to true.
I'm not sure if that callback might be issued multiple times, in particular when multistream and renegotiation come into play.
Also the condition to drop packets in `relay_rtp_packet` has been changed to let early data packets come through.

atoppi:
Updated the patch in order to send a buffered message only the first time `data_ready` is being called.
I'm still not sure of potential side-effects / bad behaviors in case of renegotiation or restart.

lminiero:
> I'm still not sure of potential side-effects / bad behaviors in case of renegotiation or restart.

The `data_ready` callback is sent any time the buffers in the SCTP stack empty, and so it's safe to send again. That's why we also use it as a trigger for when we can start sending the first time. This has nothing to do with renegotiations or restart, since it has to do with the SCTP stack, and using the `dataready` volatile int should be safe since it's only reset when `hangup_media` hits.

atoppi:
I have reverted a condition that was meaningless.
Merging since it seems to work.

PR #3414, opened 2024-08-05
Limit sdp lines (Credit to OSS-Fuzz)

atoppi:
This PR is a workaround for an issue discovered by OSS-Fuzz when trying to parse a SDP with a large number of lines.
Long story short, the `g_strsplit` function is not efficient when dealing with string with a lot of delimiter (`\n`) occurrences.

As a consequence we limit the number of SDP lines to `10000` in order to avoid the timeout on OSS-Fuzz and work around the issue in Janus.
Of course this is not the optimal solution since we do not want to hardcode a number of lines. In the long term we plan to rewrite the splitting algorithm to avoid such inefficiencies.


atoppi:
See #3424 for a proper fix.

PR #3424, opened 2024-09-09
Parse SDP with strtok_r

atoppi:
See #3414 for some more background.

This patch replaces `g_strplit` with `strtok_r` when tokenizing the SDP in `janus_sdp_parse`.


lminiero:
Merging.

PR #3426, opened 2024-09-16
Add display property on leave or unpublish event

amnonbb:
As title say. If it's not doing problem somewhere it's help to get display property in event as it's done in other places.


lminiero:
Good catch: is that what you meant to do @amnonbb?

amnonbb:
> Good catch: is that what you meant to do @amnonbb?

Yes. it is. Almost in all places we got display property except  kicked / unpublished / leaving

amnonbb:
> Just to double check since it is not being mentioned: this is for the `kicked` / `unpublished` / `leaving` event emitted **to the event handlers**, so unrelated to the Janus API event.

Sorry, now i understood what you ask. Event handlers is where i used it and keep fork only for this change. But generally may be right to add it in Janus API event as well.

lminiero:
@amnonbb looks good to me! Could you also update the documentation of the impacted events, though, to clarify there could be a `display` property too now?

oksy31 (approved):

lminiero:
Thanks, merging! I'll consider if porting the same patch to 0.x as well.

PR #3443, opened 2024-10-04
Change the way call-IDs are tracked in the SIP plugin (fixes #3404)

### LINKED ISSUE #3404: [1.x] SIP plugin: possibility of inconsistent session mapping in 'callids' hash table
1.2.4 (195bac04d431301f6efb84156a5eaf8a4f06d8ea)

Yes. I am testing with latest `master` branch.

**Was this working before?**
No.

-

**Additional context**
In case of call between two SIP handlers registered within one Janus (so when outgoing and incoming legs of one call going via the same Janus) the `callids` hash table will hold reference to latter initiated SIP call session of such call ([incoming](https://github.com/meetecho/janus-gateway/blob/c161ca023fe9be31016407172f52145cddc6f94b/src/plugins/janus_sip.c#L5401) one, because it is follow [outgoing](https://github.com/meetecho/janus-gateway/blob/c161ca023fe9be31016407172f52145cddc6f94b/src/plugins/janus_sip.c#L3911) one). This occurs because `callids` uses the `callid` as a key, which remains the same for both legs of the call.

This issue leads to incorrect functionality during attended transfers of such call, specifically in [composing the `Refer-To` header](https://github.com/meetecho/janus-gateway/blob/c161ca023fe9be31016407172f52145cddc6f94b/src/plugins/janus_sip.c#L4464) (incorrect order of `from-tag` and `to-tag`), as it uses the incoming leg instead of the outgoing one. (And yes if SIP handlers registered within different Janus all transfers functionality works as expected.)

Upon reviewing the usage of the `callids` hash table, it appears that the Sofia-SIP logger functionality also behaves incorrectly in scenarios where both legs of the same call traverse a single Janus instance, because all SIP out events will be associate with an incoming leg of such call.

lminiero:
This PR is an attempt to address the problem @ycherniavskyi reported in issue #3404. We have to thank Yurii for this fix as well, as his company, @WebTrit, sponsored the effort to investigate the issue and come up with a potential solution. Thanks, guys!

Coming to the core of the issue, Yurii identified an inconsistency in how Call-IDs were used as ways to address sessions in the SIP plugin, that would occur specifically when caller and callee were handled by the same Janus instance, and so the map would be overwritten. This could cause problems in two specific cases:

1. When using event handlers to notify incoming/outgoing SIP messages to an external backend (e.g., Homer) and mapping them to the plugin session (and so Janus handle) they belonged to.
2. When doing attended SIP transfers to fill in the `Replaces` header properly.

This PR introduces a new kind of mapping where we don't map the call-ID to a single session, but instead map it to a new structure that can keep track of both a caller and a callee, by keeping track of caller tags too. That structure is only updated with info on sessions handled by the Janus instance, and allows us to track both parties in a call from the same call-ID. This allowed us to solve the two problems above in two separate ways:

1. For event handlers, we find the mapping from call-ID, and then check From/To headers to find the right session looking at tags in the request.
2. For attended transfers, we limit the scope of lookups to sessions handled by the session implementing the transfer: in fact, transfers in Janus rely on the concept of master/helper sessions to implement multiple lines, and limiting the scope to those lines allows us to avoid ambiguities.

As usual, while this PR is currently for `master`, I'll backport it to `0.x` once it's merged. I tested the PR briefly and it seems to work, but of course my ability to test SIP deployments (and those use cases in particular) is limited. I'm looking forward to feedback, to ensure that first of all this fixes the two problems we mentioned, but also that we're not introducing regressions with these changes. If you use SIP extensively in your deployments, please make sure to test this and give us all the feedback we need to make sure it's merged as soon as possible.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1063,7 +1063,7 @@ typedef struct janus_sip_session {
 	janus_sip_media media;
 	char *transaction;
 	char *callee;
-	char *callid;
+	char *callid, *local_tag;
 	guint32 refer_id;			/* In case we were asked to transfer, keep track of the ID */
 	janus_sdp *sdp;				/* The SDP this user sent */
 	janus_recorder *arc;		/* The Janus recorder instance for this user's audio, if enabled */
@@ -1089,6 +1089,11 @@ typedef struct janus_sip_session {
 	janus_refcount ref;
 	janus_sip_dtmf latest_dtmf;
 } janus_sip_session;
+
+typedef struct janus_sip_call {
+	janus_sip_session *caller;
+	janus_sip_session *callee;
+} janus_sip_call;
 
 static GHashTable *sessions;
 static GHashTable *identities;
@@ -1121,8 +1126,17 @@ static void janus_sip_session_destroy(janus_sip_session *session) {
 	if(session && g_atomic_int_compare_and_exchange(&session->destroyed, 0, 1)) {
 		if(session->master == NULL && session->account.identity)
 			g_hash_table_remove(identities, session->account.identity);
-		if(session->callid)
-			g_hash_table_remove(callids, session->callid);
+		if(session->callid) {
+			janus_sip_call *call = g_hash_table_lookup(callids, session->callid);
+			if(call) {
+				if(call->caller == session)
+					call->caller = NULL;
+				else if(call->callee == session)
+					call->callee = NULL;
+				if(call->caller == NULL && call->callee == NULL)
+					g_hash_table_remove(callids, session->callid);
+			}
+		}
 		janus_refcount_decrease(&session->ref);
 	}
 }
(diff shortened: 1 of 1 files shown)


lminiero:
Pinging @lmangani as well, since before this fix SIP messages sent to Homer may have been misattributed by the core.

lminiero:
@ycherniavskyi any feedback on this?

ycherniavskyi:
@lminiero yes, we just finished our testing, and unfortunately we get Janus crash in some cases.

lminiero:
If there's crashes, I'll need a gdb stacktrace or, even better, a libasan dump, to see _where_ it crashes. Just the SIP messages (which is what I see in that gist) won't help unfortunately.

lminiero:
@ycherniavskyi while we wait for those crash logs I mentioned, the only thing I noticed in your log is that there seem to be many SIP BYEs that are never answered, I guess because, as you pointed out, Janus seems to be crashing when it gets the first one. That said, we don't do anything in `nua_i_bye` except save the reason for the BYE, and that wasn't changed in this PR. I personally can't replicate this crash, so I'll need debugging info from you to look into it.

lminiero:
Merging :v:

PR #3446, opened 2024-10-07
Guard all accesses to publisher->streams with publisher->streams_mutex

atoppi:
This PR addresses some race conditions we found on the videoroom plugin when accessing the `publisher->streams` list.


lminiero:
Merging :+1:

PR #3359, opened 2024-04-23
videoroom: Release some references created by remote publishers

fancycode:
~~Fixed some cases where the `rtp_forwarders` map was created without the destructor function. This resulted in the forwarders not being released when the publisher got destroyed.~~

PR #3449, opened 2024-10-09
Add SRTP for remote publishers

spscream:
hi, I got working SRTP encryption for remote publishers, tested and worked in our environment

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -1608,6 +1608,8 @@ room-<unique room ID>: {
 	"mcast" : "<multicast group port for receiving RTP packets, if any>",
 	"iface" : "<network interface or IP address to bind to, if any (binds to all otherwise)>",
 	"port" : <local port for receiving all RTP packets; 0 will bind to a random one (default)>,
+	"srtp_suite" : <length of authentication tag (32 or 80); optional>,
+	"srtp_crypto" : "<key to use as crypto (base64 encoded key as in SDES); optional>",
 	"streams" : [
 		{
 			"type" : "<type of published stream #1 (audio|video|data)">,
@@ -1657,6 +1659,8 @@ room-<unique room ID>: {
 	"secret" : "<password required to edit the room, mandatory if configured in the room>",
 	"display" : "<new display name for the remote publisher; optional>",
 	"metadata" : <new valid json object of metadata; optional>,
+	"srtp_suite" : <length of authentication tag (32 or 80); optional>,
+	"srtp_crypto" : "<key to use as crypto (base64 encoded key as in SDES); optional>",
 	"streams" : [
 		{
 			// Same syntax as add_remote_publisher: only needs to
@@ -1713,7 +1717,9 @@ room-<unique room ID>: {
 	"host" : "<host address to forward the RTP and data packets to>",
 	"host_family" : "<ipv4|ipv6, if we need to resolve the host address to an IP; by default, whatever we get>",
 	"port" : <port to forward the packets to>,
-	"rtcp_port" : <port to contact to receive RTCP feedback from the recipient; optional, and only for RTP streams, not data>
+	"rtcp_port" : <port to contact to receive RTCP feedback from the recipient; optional, and only for RTP streams, not data>,
+	"srtp_suite" : <length of authentication tag (32 or 80); optional>,
+	"srtp_crypto" : "<key to use as crypto (base64 encoded key as in SDES); optional>"
 }
 \endverbatim
  *
@@ -2194,7 +2200,9 @@ static struct janus_json_parameter publish_remotely_parameters[] = {
 	{"host", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"host_family", JSON_STRING, 0},
 	{"port", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE | JANUS_JSON_PARAM_REQUIRED},
-	{"rtcp_port", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE}
+	{"rtcp_port", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
+	{"srtp_suite", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
+	{"srtp_crypto", JSON_STRING, 0}
 };
 static struct janus_json_parameter unpublish_remotely_parameters[] = {
 	{"secret", JSON_STRING, 0},
@@ -2207,13 +2215,17 @@ static struct janus_json_parameter remote_publisher_parameters[] = {
 	{"iface", JANUS_JSON_STRING, 0},
 	{"port", JANUS_JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
 	{"streams", JANUS_JSON_ARRAY, JANUS_JSON_PARAM_REQUIRED},
-	{"metadata", JSON_OBJECT, 0}
+	{"metadata", JSON_OBJECT, 0},
+	{"srtp_suite", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
+	{"srtp_crypto", JSON_STRING, 0}
 };
 static struct janus_json_parameter remote_publisher_update_parameters[] = {
 	{"secret", JSON_STRING, 0},
 	{"display", JANUS_JSON_STRING, 0},
 	{"metadata", JSON_OBJECT, 0},
-	{"streams", JANUS_JSON_ARRAY, JANUS_JSON_PARAM_REQUIRED}
+	{"streams", JANUS_JSON_ARRAY, JANUS_JSON_PARAM_REQUIRED},
+	{"srtp_suite", JSON_INTEGER, JANUS_JSON_PARAM_POSITIVE},
+	{"srtp_crypto", JSON_STRING, 0}
 };
 static struct janus_json_parameter remote_publisher_stream_parameters[] = {
 	{"mid", JANUS_JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
@@ -2498,6 +2510,12 @@ typedef struct janus_videoroom_publisher_stream {
 	volatile gint need_pli;		/* Whether we need to send a PLI later */
 	volatile gint sending_pli;	/* Whether we're currently sending a PLI */
 	gint64 pli_latest;			/* Time of latest sent PLI (to avoid flooding) */
+	/* Only needed for SRTP support for remote publisher */
+	gboolean is_srtp;
+	int srtp_suite;
+	char *srtp_crypto;
+	srtp_t srtp_ctx;
+	srtp_policy_t srtp_policy;
 	/* Subscriptions to this publisher stream (who's receiving it)  */
 	GSList *subscribers;
 	janus_mutex subscribers_mutex;
@@ -2622,11 +2640,15 @@ typedef struct janus_videoroom_remote_recipient {
 	uint16_t port;			/* Port this publisher is being relayed to */
 	uint16_t rtcp_port;		/* RTCP port this publisher is going to latch to */
 	gboolean rtcp_added;	/* Whether we created an RTCP socket for this remotization */
+	/* Only needed for SRTP support for remote publisher */
+	int srtp_suite;
+	char *srtp_crypto;
 } janus_videoroom_remote_recipient;
 static void janus_videoroom_remote_recipient_free(janus_videoroom_remote_recipient *r) {
 	if(r) {
 		g_free(r->remote_id);
 		g_free(r->host);
+		g_free(r->srtp_crypto);
 		g_free(r);
 	}
 }
@@ -2710,6 +2732,11 @@ static void janus_videoroom_publisher_stream_free(const janus_refcount *ps_ref) 
 	janus_mutex_destroy(&ps->subscribers_mutex);
 	janus_mutex_destroy(&ps->rid_mutex);
 	janus_rtp_simulcasting_cleanup(NULL, NULL, ps->rid, NULL);
+	if(ps->is_srtp) {
+		g_free(ps->srtp_crypto);
+		srtp_dealloc(ps->srtp_ctx);
+		g_free(ps->srtp_policy.key);
+	}
 	g_free(ps);
 }
 
@@ -7120,6 +7147,22 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 			if(error_code != 0)
 				goto prepare_response;
 		}
+		/* We may need to SRTP-encrypt this stream */
+		int srtp_suite = 0;
+		const char *srtp_crypto = NULL;
+		json_t *s_suite = json_object_get(root, "srtp_suite");
+		json_t *s_crypto = json_object_get(root, "srtp_crypto");
+		if(s_suite && s_crypto) {
+			srtp_suite = json_integer_value(s_suite);
+			if(srtp_suite != 32 && srtp_suite != 80) {
+				JANUS_LOG(LOG_ERR, "Invalid SRTP suite (%d)\n", srtp_suite);
+				error_code = JANUS_VIDEOROOM_ERROR_INVALID_ELEMENT;
+				g_snprintf(error_cause, 512, "Invalid SRTP suite (%d)", srtp_suite);
+				goto prepare_response;
+			}
+			srtp_crypto = json_string_value(s_crypto);
+			JANUS_LOG(LOG_VERB, "SRTP setting s_suite (%d) and s_crypto (%s) on publish_remotely\n", srtp_suite, srtp_crypto);
+		}
 		const char *remote_id = json_string_value(json_object_get(root, "remote_id"));
 		json_t *pub_id = json_object_get(root, "publisher_id");
 		json_t *json_host = json_object_get(root, "host");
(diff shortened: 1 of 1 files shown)


lminiero:
@spscream thanks! So you got it working with the same SRTP crypto context for all streams, correct?
I'll review and get back to you in the next few days.

spscream:
no, for each sender on source side rtp forwarder helper is created, so each helper have its own srtp context, and on receiving side we have srtp context per publisher stream

src/plugins/janus_videoroom.c
```diff
@@ -7080,6 +7107,25 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 			if(error_code != 0)
 				goto prepare_response;
 		}
+
```
lminiero, line 7110: Unneeded empty line.

src/plugins/janus_videoroom.c
```diff
@@ -7080,6 +7107,25 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+		if(s_suite && s_crypto) {
+			srtp_suite = json_integer_value(s_suite);
+			if(srtp_suite != 32 && srtp_suite != 80) {
+				JANUS_LOG(LOG_ERR, "Invalid SRTP suite (%d)\n", srtp_suite);
+				error_code = JANUS_VIDEOROOM_ERROR_INVALID_ELEMENT;
+				g_snprintf(error_cause, 512, "Invalid SRTP suite (%d)", srtp_suite);
+				goto prepare_response;
+			}
+			srtp_crypto = json_string_value(s_crypto);
+
```
lminiero, line 7125: Unneeded empty line.

src/plugins/janus_videoroom.c
```diff
@@ -7080,6 +7107,25 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+			JANUS_LOG(LOG_VERB, "SRTP setting s_suite (%d) and s_crypto (%s) on publish_remotely \n", srtp_suite, srtp_crypto);
```
lminiero, line 7126: Unneeded space before the `\n`.

src/plugins/janus_videoroom.c
```diff
@@ -7080,6 +7107,25 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+		}
+
```
lminiero, line 7128: Unneeded empty line.

src/plugins/janus_videoroom.c
```diff
@@ -7276,6 +7322,8 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 		recipient->port = port;
 		recipient->rtcp_port = rtcp_port;
 		recipient->rtcp_added = rtcp_added;
+		recipient->srtp_suite = srtp_suite;
+		recipient->srtp_crypto = g_strdup(srtp_crypto);
```
lminiero, line 7326: I can't remember if glib complains if you pass `NULL` to `g_strdup`. At any rate, for a matter of consistency with the rest of the code, it might be better to use a ternary here, e.g.

    recipient->srtp_crypto = srtp_crypto ? g_strdup(srtp_crypto) : NULL;

src/plugins/janus_videoroom.c
```diff
@@ -7551,6 +7599,23 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 				}
 			}
 		}
+
```
lminiero, line 7602: Unneeded empty line.

src/plugins/janus_videoroom.c
```diff
@@ -7551,6 +7599,23 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+		if(s_suite && s_crypto) {
+			srtp_suite = json_integer_value(s_suite);
+			if(srtp_suite != 32 && srtp_suite != 80) {
+				JANUS_LOG(LOG_ERR, "Invalid SRTP suite (%d)\n", srtp_suite);
+				error_code = JANUS_VIDEOROOM_ERROR_INVALID_ELEMENT;
+				g_snprintf(error_cause, 512, "Invalid SRTP suite (%d)", srtp_suite);
+				goto prepare_response;
+			}
+			srtp_crypto = json_string_value(s_crypto);
+			JANUS_LOG(LOG_VERB, "SRTP setting s_suite (%d) and s_crypto (%s) on add_remote_publisher \n", srtp_suite, srtp_crypto);
```
lminiero, line 7617: Unneeded space before `\n`.

src/plugins/janus_videoroom.c
```diff
@@ -7729,6 +7794,45 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 			gboolean disabled = json_is_true(json_object_get(s, "disabled"));
 			/* Create a publisher stream */
 			ps = g_malloc0(sizeof(janus_videoroom_publisher_stream));
+			if(mtype == JANUS_VIDEOROOM_MEDIA_AUDIO || mtype == JANUS_VIDEOROOM_MEDIA_VIDEO) {
+				/* First of all, let's check if we need to setup an SRTP for remote publisher */
+				JANUS_LOG(LOG_HUGE, "trying to enable SRTP crypto (%s) suite (%d) for stream. \n", srtp_crypto, srtp_suite);
```
lminiero, line 7799: I don't think the text in this log line is correct. You don't even know if SRTP is enabled, when you get here. I'd remove this log line entirely. But, in general:

s/trying/Trying

Unneeded dot and space before `\n`

src/plugins/janus_videoroom.c
```diff
@@ -7729,6 +7794,45 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+				if(srtp_suite > 0 && srtp_crypto != NULL) {
+					JANUS_LOG(LOG_VERB, "enabling SRTP crypto (%s) for stream. \n", srtp_crypto);
```
lminiero, line 7801: s/enabling/Enabling

Unneeded dot and space before `\n`

src/plugins/janus_videoroom.c
```diff
@@ -7729,6 +7794,45 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+					gsize len = 0;
+					guchar *srtp_crypto_decoded = g_base64_decode(srtp_crypto, &len);
+					if(len < SRTP_MASTER_LENGTH) {
+						/* Something went wrong */
+						g_free(srtp_crypto_decoded);
+						JANUS_LOG(LOG_ERR, "Invalid SRTP crypto (%s), disabling stream.\n", srtp_crypto);
```
lminiero, line 7807: Unneeded dot before `\n`

src/plugins/janus_videoroom.c
```diff
@@ -7729,6 +7794,45 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+						disabled = TRUE;
+					} else {
+						/* Set SRTP policy */
+						srtp_policy_t *policy = &ps->srtp_policy;
+						srtp_crypto_policy_set_rtp_default(&(policy->rtp));
```
lminiero, line 7812: There's no need for parenthesis here, `&policy->rtp` is fine too (the same applies to the lines below).

src/plugins/janus_videoroom.c
```diff
@@ -7729,6 +7794,45 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+						if(srtp_suite == 32) {
+							srtp_crypto_policy_set_aes_cm_128_hmac_sha1_32(&(policy->rtp));
+						} else if(srtp_suite == 80) {
+							srtp_crypto_policy_set_aes_cm_128_hmac_sha1_80(&(policy->rtp));
+						}
+						policy->ssrc.type = ssrc_any_inbound;
+						policy->key = srtp_crypto_decoded;
+						policy->next = NULL;
+
```
lminiero, line 7821: Unneeded empty line.

src/plugins/janus_videoroom.c
```diff
@@ -7927,6 +8031,22 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+		if(s_suite && s_crypto) {
+			srtp_suite = json_integer_value(s_suite);
+			if(srtp_suite != 32 && srtp_suite != 80) {
+				JANUS_LOG(LOG_ERR, "Invalid SRTP suite (%d)\n", srtp_suite);
+				error_code = JANUS_VIDEOROOM_ERROR_INVALID_ELEMENT;
+				g_snprintf(error_cause, 512, "Invalid SRTP suite (%d)", srtp_suite);
+				goto prepare_response;
+			}
+			srtp_crypto = json_string_value(s_crypto);
+			JANUS_LOG(LOG_VERB, "SRTP setting s_suite (%d) and s_crypto (%s) on add_remote_publisher \n", srtp_suite, srtp_crypto);
```
lminiero, line 8048: Unneeded space before `\n`

src/plugins/janus_videoroom.c
```diff
@@ -8007,6 +8127,47 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
 			gboolean disabled = json_is_true(json_object_get(s, "disabled"));
 			/* Create a publisher stream */
 			ps = g_malloc0(sizeof(janus_videoroom_publisher_stream));
+			if(mtype == JANUS_VIDEOROOM_MEDIA_AUDIO || mtype == JANUS_VIDEOROOM_MEDIA_VIDEO) {
+				/* First of all, let's check if we need to setup an SRTP for remote publisher */
+				JANUS_LOG(LOG_HUGE, "trying to enable SRTP crypto (%s) suite (%d) for stream. \n", srtp_crypto, srtp_suite);
```
lminiero, line 8132: See note on the same message above.

src/plugins/janus_videoroom.c
```diff
@@ -8007,6 +8127,47 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+				if(srtp_suite > 0 && srtp_crypto != NULL) {
+					JANUS_LOG(LOG_VERB, "enabling SRTP crypto (%s) for stream. \n", srtp_crypto);
```
lminiero, line 8134: s/enabling/Enabling

Unneeded dot and space before `\n`

src/plugins/janus_videoroom.c
```diff
@@ -8007,6 +8127,47 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+					gsize len = 0;
+					guchar *srtp_crypto_decoded = g_base64_decode(srtp_crypto, &len);
+					if(len < SRTP_MASTER_LENGTH) {
+						/* Something went wrong */
+						g_free(srtp_crypto_decoded);
+						JANUS_LOG(LOG_ERR, "Invalid SRTP crypto (%s), disabling stream.\n", srtp_crypto);
```
lminiero, line 8140: Unneeded dot before `\n`

src/plugins/janus_videoroom.c
```diff
@@ -8007,6 +8127,47 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+						disabled = TRUE;
+					} else {
+						/* Set SRTP policy */
+						srtp_policy_t *policy = &ps->srtp_policy;
+						srtp_crypto_policy_set_rtp_default(&(policy->rtp));
```
lminiero, line 8145: See note above on parenthesis.

src/plugins/janus_videoroom.c
```diff
@@ -8007,6 +8127,47 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+						if(srtp_suite == 32) {
+							srtp_crypto_policy_set_aes_cm_128_hmac_sha1_32(&(policy->rtp));
+						} else if(srtp_suite == 80) {
+							srtp_crypto_policy_set_aes_cm_128_hmac_sha1_80(&(policy->rtp));
+						}
+						policy->ssrc.type = ssrc_any_inbound;
+						policy->key = srtp_crypto_decoded;
+						policy->next = NULL;
+
```
lminiero, line 8154: Unneeded empty line.

src/plugins/janus_videoroom.c
```diff
@@ -8007,6 +8127,47 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+						/* Create SRTP context */
+						srtp_err_status_t res = srtp_create(&ps->srtp_ctx, policy);
+						if(res == srtp_err_status_ok) {
+							ps->is_srtp = TRUE;
+							ps->srtp_suite = srtp_suite;
+							ps->srtp_crypto = g_strdup(srtp_crypto);
+						} else {
+							/* Something went wrong... */
+							JANUS_LOG(LOG_ERR, "Error creating SRTP context: %d (%s), disabling stream.\n", res, janus_srtp_error_str(res));
```
lminiero, line 8163: Unneeded dot before `\n`

src/plugins/janus_videoroom.c
```diff
@@ -8007,6 +8127,47 @@ static json_t *janus_videoroom_process_synchronous_request(janus_videoroom_sessi
...
+							disabled = TRUE;
+						}
+					}
+				} else {
+					JANUS_LOG(LOG_ERR, "SRTP crypto (%d) (%s) not enabled for stream. \n", srtp_suite, srtp_crypto);
```
lminiero, line 8168: Unneeded dot and space before `\n`

src/plugins/janus_videoroom.c
```diff
@@ -13643,6 +13804,21 @@ static void *janus_videoroom_remote_publisher_thread(void *user_data) {
 					janus_videoroom_incoming_data_internal(publisher->session, publisher, &data);
 					continue;
 				}
+
```
lminiero, line 13807: Unneeded empty line.

src/plugins/janus_videoroom.c
```diff
@@ -13643,6 +13804,21 @@ static void *janus_videoroom_remote_publisher_thread(void *user_data) {
...
+				/* Is this SRTP? */
+				if(ps->is_srtp) {
+					int buflen = bytes;
+					srtp_err_status_t res = srtp_unprotect(ps->srtp_ctx, buffer, &buflen);
+					if(res != srtp_err_status_ok) {
+						janus_mutex_unlock(&publisher->streams_mutex);
+						guint32 timestamp = ntohl(rtp->timestamp);
+						guint16 seq = ntohs(rtp->seq_number);
+						JANUS_LOG(LOG_ERR, "[%s] Publisher stream (#%d) SRTP unprotect error: %s (len=%d-->%d, ts=%"SCNu32", seq=%"SCNu16")\n",
+							publisher->user_id_str, ps->mindex, janus_srtp_error_str(res), bytes, buflen, timestamp, seq);
```
lminiero, line 13817: This will flood your logs in case you start getting traffic that's inconsistent with your SRTP context, since it will print a log line for every incoming RTP packet, and there may be a lot of them. In the core we try to limit that by showing an error summary every couple of seconds or so to limit the damage. An alternative might be having some sort of toggle: the first time you get an SRTP error you print the error, and then you don't anymore until you get a proper error (SRTP decoding succeeded) that resets the toggle (but this could still cause log flooding if success/error alternate often).

lminiero (commented):
I just did a first review of the code: it looks mostly fine to me, and in fact you'll see that most comments are editorial and nitpicking. I did add a note on SRTP error logging that could be problematic the way it's done now.

I also have a doubt on just setting a stream to `disabled = TRUE` in case SRTP creation failed, considering there may be other things that re-enable that same stream, e.g., by subsequent API calls that set `disabled` back to `TRUE` on a stream that has a broken SRTP context at that point.  Not sure if that could cause the code to crash, but it's worth investigating what could happen in that case.

src/plugins/janus_videoroom.c
spscream: Do you have any example from core, where you summarize logs? I copied this part from streaming plugin.

spscream: I found it in ice.c, I'll make the same - write to debug log on every packet and write  summary periodically.

lminiero: If it's there too then nevermind, I can look into this later on and fix it on both. I just noticed that a few of the nits I mentioned were actually in that code too (e.g., the `&policy->rtp` no parenthesis thing).

lminiero: No need, don't worry: I'll take care of that in both plugins myself so that it's done in a consistent way (I'll have to check if the same happens in SIP and NoSIP plugins too).

spscream: okay, I removed parethesis anyway, pushed fixed code, without log fix

lminiero: Ack, thanks! I think the only thing left to look at is the `disabled` part I mentioned in my message then.

spscream: I think about  disabled, maybe we should add new flag for stream? like srtp_init_failed or similar and set it to true and add required checks in logic.

spscream: btw it was is_rtp=FALSE by default, I set it to be more explicit.

lminiero:
Just as a side note, please notice that this will be merged only after #3359 is merged too (also considering I asked @fancycode to do rebases more than once already :sweat_smile: ). You may want to have a look at/review that PR as well, since it's aimed at addressing some memory leaks that apparently remote publishers cause (and since you've started using them, you'll probably be interested in that).

spscream:
so what is the plan? I could rebase myself, this is not the problem

lminiero:
I'm out of office today, there's a few PRs (including this one) that I plan to review one last time and then merge early next week.

spscream:
sure, updated

PR #3452, opened 2024-10-16
rtcp: add receiver_ssrc parsing for PSFB packets

tmatth:
This fixes handling of PLIs received from Pion streaming plugin clients.

### DIFF
--- a/rtcp.c
+++ b/rtcp.c
@@ -143,6 +143,11 @@ guint32 janus_rtcp_get_receiver_ssrc(char *packet, int len) {
 				janus_rtcp_fb *rtcpfb = (janus_rtcp_fb *)rtcp;
 				return ntohl(rtcpfb->media);
 			}
+			case RTCP_PSFB: {
+				/* PSFB, Payload-specific FB message (rfc4585) */
+				janus_rtcp_fb *rtcpfb = (janus_rtcp_fb *)rtcp;
+				return ntohl(rtcpfb->media);
+			}
 			default:
 				break;
 		}


tmatth:
I will add some more specifics but the idea here is to address that when adding:

```diff
			go func() {
				ticker := time.NewTicker(time.Second * 3)
				for range ticker.C {
					rtcpSendErr := peerConnection.WriteRTCP([]rtcp.Packet{&rtcp.PictureLossIndication{MediaSSRC: uint32(track.SSRC())}})
					if rtcpSendErr != nil {
						log.Println(rtcpSendErr)
					} else {
						log.Printf("Sent PLI for SSRC %v", track.SSRC())
					}
				}
			}()
```

to a program like https://github.com/pion/example-webrtc-applications/blob/master/janus-gateway/streaming/main.go so that it sends PLIs every 3 seconds, these are disappearing in Janus (specifically, here https://github.com/meetecho/janus-gateway/blob/f902f69d5114453b08ccd828667af3e4ddb74870/ice.c#L3207). 

With the proposed changes, the keyframe request gets recognized by the streaming plugin.

This PR is against 0.x since I've only tested against that. If this approach makes sense I can forward port the change to master.

rtcp.c
```diff
@@ -143,6 +143,12 @@ guint32 janus_rtcp_get_receiver_ssrc(char *packet, int len) {
 				janus_rtcp_fb *rtcpfb = (janus_rtcp_fb *)rtcp;
 				return ntohl(rtcpfb->media);
 			}
+			case RTCP_PSFB: {
+				/* PSFB, Payload-specific FB message (rfc4585) */
+				janus_rtcp_fb *rtcpfb = (janus_rtcp_fb *)rtcp;
+				/* Fallback to media SSRC if sender SSRC is not set */
+				return ntohl(rtcpfb->ssrc) || ntohl(rtcpfb->media);
```
lminiero, line 150: I don't think this is correct: `rtcpfb->ssrc` is the SSRC of the packet sender (https://datatracker.ietf.org/doc/html/rfc4585#section-6.1), so IMO it's wrong to return that in a method that looks for the receiver SSRC. We do return `rtcpfb->ssrc` for RTCP_PSFB already, but in `janus_rtcp_get_sender_ssrc`. As such, I'd only leave `rtcpfb->media` in there, as we do for RTCP_RTPFB just a few lines above which uses the same struct.

If that works, it probably does make sense for 1.x as well since they share most of this RTCP code.

lminiero:
Thanks, merging then!

atoppi:
@tmatth I've restricted the PSFB parsing to just PLIs and added the length check for FCI, can you check that everything is still working as intended?

tmatth:
> @tmatth I've restricted the PSFB parsing to just PLIs and added the length check for FCI, can you check that everything is still working as intended?

Oh thanks I was wondering about that, yes that's still working as expected.

PR #3454, opened 2024-10-17
rtcp: add receiver_ssrc parsing for PSFB packets

tmatth:
This fixes handling of PLIs received from Pion streaming plugin clients.


tmatth:
This is the 1.x equivalent of #3452

lminiero:
Merging :v:

Issue #3464, opened 2024-10-24
[1.x] Streaming plugin can leak RTP port


woutd:
**What version of Janus is this happening on?**
1.2.4

**Have you tested a more recent version of Janus too?**
No

**Was this working before?**
Probably not

**Is there a gdb or libasan trace of the issue?**
No

**Additional context**
When the first media port for a create request is free but the second port is in use Janus returns an error and does not create the streaming mountpoint. But it keeps listening on the first port. It is now unable to free this port.

To reproduce:

Create first mountpoint and listen to 24210 and 24211:
```
{'janus': 'message', 'body': {'request': 'create', 'type': 'rtp', 'id': '12345', 'is_private': True, 'media': [{'mid': 'video', 'type': 'video', 'iface': '127.0.0.1', 'port': 24210, 'pt': 96, 'codec': 'h264', 'fmtp': 'profile-level-id=42e01f;packetization-mode=1'}, {'mid': 'audio', 'type': 'audio', 'iface': '127.0.0.1', 'port': 24211, 'pt': 111, 'codec': 'opus'}], 'secret': 'XXX', 'pin': 'foobar', 'permanent': True, 'admin_key': 'YYY'}, 'transaction': 'foobar'}
```

Janus returns:
```
{'janus': 'success', 'session_id': 8590268524389867, 'transaction': 'foobar', 'sender': 3926924562162867, 'plugindata': {'plugin': 'janus.plugin.streaming', 'data': {'streaming': 'created', 'created': 'mp-12345', 'permanent': True, 'stream': {'id': '12345', 'type': 'live', 'description': 'mp-12345', 'is_private': True, 'host': '127.0.0.1', 'ports': [{'type': 'video', 'mid': 'video', 'port': 24210}, {'type': 'audio', 'mid': 'audio', 'port': 24211}]}}}}
```

Janus is listening to 24210 and 24211:
```
udp        0      0 127.0.0.1:24210        0.0.0.0:*                           3541152/janus       
udp        0      0 127.0.0.1:24211        0.0.0.0:*                           3541152/janus 
```

Try to create second mountpoint to listen on 24212 and 24210:
```
[...]

PR #3465, opened 2024-10-24
Cleanup allocated mp streams when janus_streaming_create_rtp_source_stream fails (see #3464)

atoppi:
This is a tentative fix for #3464 .
The patch is not complete since we identified other potential leaks in the streaming plugin when a mountpoint creation fails. Starting to track the efforts here.


lminiero:
Merging.

PR #3467, opened 2024-10-28
Add metadata for videoroom publishers

spscream:
I we want to introduce ability to set metadata for publishers on join. 
I added metadata field with json_t and save info to it on join and send it in "joined"/"published" events.

We use it on our branch now to build custom business logic for videoroom, for example we set in metadata now:
- server_id from which user has connected(we build remote publishers logic for our backends around this field)
- users ip/network, we use it to build acl for publishers, we have "contours" feature for our publishers, screen sharing of user can be viewed only from users from the same network/contour/group/role as publisher


lminiero:
It does make sense, yes, but as it is it looks like it's an immutable object. I'd add an option to modify the metadata object in a configure request (and notify the change to other attendees), just as you can edit the display name dynamically.

src/plugins/janus_videoroom.c
```diff
@@ -4424,6 +4447,8 @@ static void janus_videoroom_leave_or_unpublish(janus_videoroom_publisher *partic
 		json_object_set_new(info, "id", string_ids ? json_string(participant->user_id_str) : json_integer(participant->user_id));
 		if(participant->display)
 			json_object_set_new(info, "display", json_string(participant->display));
+		if(participant->metadata)
+			json_object_set_new(info, "metadata", json_deep_copy(participant->metadata));
```
amnonbb, line 4451: I think here we need to notify metadata janus api as well

spscream: sorry, I don't get it. Please explain what you mean?

spscream: ah, I see. thanks. Added

lminiero:
Thanks for the changes, looks good to me! Once you fix the conflicts, I think this is ready to be merged.

spscream:
@lminiero rebased

lminiero:
Thanks! The only thing I'm wondering about, now, is that if you use a configure to update both `display` and `metadata` (and maybe something else) you get different notifications for eadh of them, instead of a single event that notifies a change on everything. But that's something I can look into later on, especially considering it may have been an issue before already (e.g., with `display` and other properties).

lminiero:
Merging :+1:

Issue #3463, opened 2024-10-24
[1.x] stuck and crash on version from master


spscream:
**What version of Janus is this happening on?**
branch based on master from b092e3d342c997e45dd3918952ea7d638b921e1e commit

**Have you tested a more recent version of Janus too?**
yes, branch based on master from 504daf5aef333d6f37e41c30b00be24cfb6c83bf

**Was this working before?**
yes

**Is there a gdb or libasan trace of the issue?**
no

**Additional context**
we got janus stuck in production on recent commits from master, it stops to accept creation of sessions, logs are having "DestroyingWebSocket client" message repeatedly - it accepts ws connections but doesn't create new sessions:
```
2024-10-24T07:46:05.181877549Z [WSS-0x7f9e608f5420] Destroying WebSocket client
2024-10-24T07:46:05.195416485Z [WSS-0x7f9e608f5980] Destroying WebSocket client
2024-10-24T07:46:05.220954130Z [WSS-0x7f9e608f5ea0] Destroying WebSocket client
2024-10-24T07:46:05.224866921Z [WSS-0x7f9e608f6400] Destroying WebSocket client
2024-10-24T07:46:05.237925976Z [WSS-0x7f9e608f6960] Destroying WebSocket client
2024-10-24T07:46:05.241920620Z [WSS-0x7f9e57f8e010] Destroying WebSocket client
2024-10-24T07:46:05.245381267Z [WSS-0x7f9e57f8e550] Destroying WebSocket client
2024-10-24T07:46:05.254307072Z [WSS-0x7f9e57f8f520] Destroying WebSocket client
2024-10-24T07:46:05.258169941Z [WSS-0x7f9e57f8f000] Destroying WebSocket client

```
before stuck logs looks normally:
```
2024-10-24T07:02:18.806658918Z [WSS-0x7f9e608f0030] Destroying WebSocket client
2024-10-24T07:02:18.806665306Z Destroying session 5866593402157672; 0x7f9e5ad2eb50
2024-10-24T07:02:18.806668003Z Detaching handle from JANUS VideoRoom plugin; 0x7f9dff8ff790 0x7f9e24a842e0 0x7f9dff8ff790 0x7f9e4f1254e0
2024-10-24T07:02:18.806670957Z Detaching handle from JANUS AudioBridge plugin; 0x7f9e1cc50920 0x7f9e4cb184d0 0x7f9e1cc50920 0x7f9e0610ecf0
[...]

PR #3468, opened 2024-10-28
Fix possible deadlock on udp socket opening error on rtp_forward

spscream:
this possibly fixes https://github.com/meetecho/janus-gateway/issues/3463 introduced by https://github.com/meetecho/janus-gateway/commit/d014cff9cf1d4cac75e2e2862f695c2d2096aeda


lminiero:
Ah good catch, thanks for spotting this! Merging right away then :+1:

PR #3469, opened 2024-10-29
Fix deadlock on janus_videoroom_rtp_forwarder_rtcp_receive

spscream:
another one possible deadlock found


src/plugins/janus_videoroom.c
```diff
@@ -13270,6 +13270,7 @@ static void janus_videoroom_rtp_forwarder_rtcp_receive(janus_rtp_forwarder *rf,
 				janus_mutex_lock(&p->streams_mutex);
 				janus_mutex_lock(&p->rtp_forwarders_mutex);
 				if(g_hash_table_size(p->rtp_forwarders) == 0) {
+					janus_mutex_unlock(&p->streams_mutex);
```
lminiero, line 13273: This should be moved down one line, because `streams_mutex` was obtained before `rtp_forwarders_mutex` and so the release order should be reversed.

spscream: sure, updated

lminiero:
Merging :+1:

PR #3471, opened 2024-10-30
[0.x] Optimize NACKs handling

atoppi:
This PR includes a set of optimizations for NACKs handling.
We identified some inefficient code paths that could potentially cause some performance issues when the server receives a high number of NACK requests.
Once completed this, a similar effort will be done on 1.x too.

**TODO**
- [x]  replace g_list with g_queue to only use O(1) operations and avoid full list  scanning
- [x]  use a pre-allocated queue per component
<del> reduce memory allocations when preparing packets to retransmit </del>
- [x] add a backoff algo when retransmitting the same packet multiple times

### DIFF
--- a/ice.c
+++ b/ice.c
@@ -1887,9 +1887,8 @@ static void janus_ice_component_free(const janus_refcount *component_ref) {
 		g_queue_free(component->video_retransmit_buffer);
 		g_hash_table_destroy(component->video_retransmit_seqs);
 	}
-	if(component->nacks_queue != NULL) {
+	if(component->nacks_queue != NULL)
 		g_queue_free(component->nacks_queue);
-	}
 	if(component->candidates != NULL) {
 		GSList *i = NULL, *candidates = component->candidates;
 		for(i = candidates; i; i = i->next) {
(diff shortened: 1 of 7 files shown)


atoppi:
This is now ready to be tested.

ice.c
```diff
@@ -5246,16 +5245,23 @@ static gboolean janus_ice_outgoing_traffic_handle(janus_ice_handle *handle, janu
 						}
 						if(p == NULL) {
 							/* If we're not doing RFC4588, we're saving the SRTP packet as it is */
-							p = g_malloc(sizeof(janus_rtp_packet));
+							p = g_malloc(sizeof(janus_ice_queued_packet));
 							p->data = g_malloc(protected);
 							memcpy(p->data, pkt->data, protected);
 							p->length = protected;
-							janus_plugin_rtp_extensions_reset(&p->extensions);
```
atoppi, line 5253: @lminiero not sure this is right

plugins/plugin.h
```diff
@@ -171,7 +171,7 @@ janus_plugin *create(void) {
  * Janus instance or it will crash.
  *
  */
-#define JANUS_PLUGIN_API_VERSION	19
+#define JANUS_PLUGIN_API_VERSION	18
```
tmatth, line 174: Shouldn't this be increasing?

lminiero:
Merging.

PR #3474, opened 2024-11-04
videoroom: always take room mutex before streams mutex (see #3463)

atoppi:
This patch tries to fix the mutex order reversal discussed in #3463.
We tried to enforce the order `videoroom->mutex` ---> `publisher/subscriber->streams_mutex`.

This is highly experimental and poorly tested, pinging @spscream since we know they stress a lot those critical paths  subject to deadlocking.


spscream:
we tested it since yesterday, no hangs yet. Deploying to more servers for test

brave44:
@atoppi we tested it for two days on production and have no problems, so I think this time fix is on point, thanks! Waiting for merging now so we can use master again.

lminiero:

PR #3475, opened 2024-11-06
Fix references counting for remote publisher and streams (see #3359)

atoppi:
This PR tries to solve the issues described in #3359.
It's a bit more convoluted than the original patch since it addresses also some error cases.


fancycode:
LGTM as commented in https://github.com/meetecho/janus-gateway/pull/3359#issuecomment-2461608953

fancycode (approved):

lminiero:
Merging.

PR #3480, opened 2024-11-27
Sip module - get sip password from external url

dvillaume:
Hello,

We wanted to avoid end user to see SIP password in flows.

The idea here would be to retrieve a web_session_id from webrtc client and to get the associated sip_password from http request (accessible only from Janus).

hope this could be helpfull for other people and avoid SIP credentials leak.

Tell me if there is anything i can enhance as my C skills are pretty low.

### DIFF
--- a/conf/janus.plugin.sip.jcfg.sample
+++ b/conf/janus.plugin.sip.jcfg.sample
@@ -52,4 +52,9 @@ general: {
 	# engine (default 32000 milliseconds)
 	sip_timer_t1x64 = 32000
 
+	# Set password retrieval on private url from web_session_id (Avoiding to expose reusable sip credentials)
+	# if web_session_id is defined, the sip password will be retrieved at $sip_passwd_url/$web_session_id
+	get_sip_passwd_from_url = false
+	#sip_passwd_url = "https://mydomain.com"
+
 }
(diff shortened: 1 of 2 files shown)


januscla:
Thanks for your contribution, @dvillaume! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero (commented):
Hi David, thanks for your contribution! I've added some generic notes inline.

On a general note, I'm not 100% sure whether this could be a good idea or not. This seems to assume that a backend component has access to the credentials of users, which sounds like it's "bad"? When would such a setup be considered acceptable? It also seems to auto-default to using this for plain text secrets, when a different approach might be better (but I'm not a SIP expert, so I'm not familiar about what alternatives could exist here). There doesn't seem to be any way to provide some form of authentication, since all it relies on is a path to use for lookups: I guess that in controlled environments this is probably ok, but shouldn't this account for the possibility that it might not be the case? I don't know if there are standard ways of retrieving credentials, e.g., for provisioning purposes, that may make sense here.

I'll leave it to other users of the SIP plugin to discuss whether or not such a feature could be helpful to them as well.

dvillaume:
Thanks for your answers @lminiero.
i figured a better way to achieve the no leak password goal (cf https://github.com/meetecho/janus-gateway/pull/3481).
I close this PR.
Thanks

PR #3481, opened 2024-11-27
Add AES-256 encryption support for SIP secret management

dvillaume:
Hello,

in order to avoid SIP password leaks to final users, i implemented a way to use the AES encrypted value of the SIP password with a per server key based.

Tell me if this could be usefull for the community.


januscla:
Thanks for your contribution, @dvillaume! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

PR #3487, opened 2024-12-09
Fix memory leak (free the descriptor)

m08pvv:
Noticed a memory leak in record.c, here is a small patch.
If someone needs libasan trace log with specific lines that pointed to this descriptor, I can post it here or paste it somewhere (it's a nice 8 lines stacktrace from which you only need 3.5)

### DIFF
--- a/src/record.c
+++ b/src/record.c
@@ -74,6 +74,8 @@ static void janus_recorder_free(const janus_refcount *recorder_ref) {
 	recorder->codec = NULL;
 	g_free(recorder->fmtp);
 	recorder->fmtp = NULL;
+	g_free(recorder->description);
+	recorder->description = NULL;
 	if(recorder->extensions != NULL)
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @m08pvv! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

ugenef:
@lminiero 
Hey, 
we're facing the same issue. Looks like the PR is going to fix that.

lminiero:
> If someone needs libasan trace log with specific lines that pointed to this descriptor, I can post it here or paste it somewhere

No need, it's a pretty obvious leak, thanks for spotting it! I think it went unnoticed so far because it's basically never used: only the VideoRoom adds a description, but only if a publisher stream has a description, which I think our stock demos never add, and probably not many people using the plugin do either. I'll merge and backport to 0.x.

PR #3488, opened 2024-12-10
janus-pp-rec: improve DTX detection

atoppi:
This patch improves opus DTX detection in janus-pp-rec by considering as DTX those packets with a plen greater than 20 ms.
We are aware that WebRTC/opus supports plen up to 60ms, but historically janus-pp-rec has only been capable of processing 20ms packets so we're not introducing any regression from that perspective.

Since pp-rec restamping (`-r`) artificially introduces larger timestamp gaps, we mark restamped packets in the first loop of janus-pp-rec in order not to confuse them with DTX packets.


lminiero:
Merging.

PR #3498, opened 2025-01-02
Move exit handlers setting at the beginning of main

atoppi:
As the title says this PR moves the exit routines setting at the beginning of the main process.

This change should fix the missing log prints when the process exits early due to using `--version` param or encountering an error while parsing the configuration.

### DIFF
--- a/src/janus.c
+++ b/src/janus.c
@@ -4407,6 +4407,11 @@ gint main(int argc, char *argv[]) {
 
 	janus_mark_started();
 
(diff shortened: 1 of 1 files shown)


tmatth:
@atoppi can this be backported?

atoppi:
Done in https://github.com/meetecho/janus-gateway/commit/1b6c0005c0c7f422b85237f7f45c8fafc6fa688c.
The backport commit is slightly different since 0.x is still using gengetopt.

PR #3499, opened 2025-01-03
Fix typo in Streaming plugin

peter15914:
Value ```host``` was written to json-object ```ml``` instead of ```info```.


januscla:
Thanks for your contribution, @peter15914! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Good catch! It was probably a copy/paste error or a leftover from the 0.x compatibility mode (where we do indeed put all that info in `ml` instead). Merging, thanks!

PR #3504, opened 2025-01-10
Experimental support for the video-layers-allocation extension

lminiero:
This PR adds experimental support for the `rtp-hdrext/video-layers-allocation00` RTP extension:
https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/video-layers-allocation00/

As explained in the documentation above, this extension, when negotiated, provides info associated with a simulcast/SVC topology, including the number of available spatial and temporal layers, info on the target bitrates, and the associated resolutions. To keep things simple, when we negotiate this extension we only extract the number of spatial and temporal layers, and then make it available to plugins as part of the `extensions` structure.

For a POC integration, I added support to this extension to the EchoTest plugin, which always tries to negotiate it. When it finds VLA info associated to an incoming packet, it enriches the existing simulcast/SVC events with that info too (`total_spatial_layers`, `total_temporal_layers`), that the EchoTest demo then uses to dynamically show/hide the associated buttons in the demo page.

Should this prove to work nicely, we can add the same functionality to the VideoRoom plugin as well, where I can definitely see this information being useful to interested subscribers (dynamically or as part of the info the plugin advertises about publishers). In the future, we can consider parsing and advertising bitrates/resolutions too, but that would be considerably more work: just notifying about existing layers is already a useful piece of information with very little overhead.


lminiero:
Merging as it is, for now. Adding support to the VideoRoom for this extension can be done in a later effort.

PR #3507, opened 2025-01-17
Fix type error

### LINKED ISSUE #3506: [1.3.0]Parameter 'error' and 'sourceId' implicitly has an 'any' type.
I am using the npm package janus-gateway with version ^1.3.0. The issue occurs in the type definitions at line 39 of janus.d.ts, where parameters in GetScreenCallback lack explicit types

I am using the npm package version ^1.3.0, which is the latest version as per today

**Was this working before?**
I do not have information on whether this issue existed in earlier versions of the npm package.

This issue does not produce a runtime error or crash but rather a TypeScript compilation error due to implicit any types. A trace or debug tool is not applicable here as it’s purely a type definition issue.

**Additional context**

itsmeabhishek03:
Fixes #3506

This PR resolves TypeScript compilation errors caused by missing explicit types in the GetScreenCallback type definition in janus.d.ts.

Issue Details:
Problem: The GetScreenCallback type was defined as:

type GetScreenCallback = (error?, sourceId?) => void;
This caused TypeScript errors (TS7006) when compiling in strict mode, as the parameters error and sourceId implicitly had the any type.

Solution: The parameters have been updated with explicit types to ensure compatibility with strict TypeScript settings:
type GetScreenCallback = (error:string, sourceId:string) => void

### DIFF
--- a/npm/janus.d.ts
+++ b/npm/janus.d.ts
@@ -36,7 +36,7 @@ declare namespace JanusJS {
 		init: () => void;
 	}
 
-	type GetScreenCallback = (error?, sourceId?) => void
+	type GetScreenCallback = (error:string, sourceId:string) => void
 
 	type HttpApiCallOption = {
 		async: boolean,


januscla:
Thanks for your contribution, @itsmeabhishek03! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks for spotting this and for the fix, @itsmeabhishek03, merging!

PR #3508, opened 2025-01-20
Cleanup remote publishers on videoroom destroy

spscream:
As I described here: https://janus.discourse.group/t/add-remote-publisher-and-room-destroy/1452/1 janus continues listening ports if remove_remote_publisher wasn't called before room destroy. 

We tested it on our production servers and seems like all is good - ports are freeing and memory isn't leeking.


lminiero:
Excellent, thanks for testing!

PR #3509, opened 2025-01-21
Fix recording for remote publishers

spscream:
as explained in https://janus.discourse.group/t/remote-publishers-and-recording/1455 on add_remote_publisher/update_remote_publisher remote publisher isn't recorded if record was started before adding remote publisher


lminiero:
Apologies, this must have slipped through my notifications. Thanks for the fix, merging!

Issue #3510, opened 2025-01-28
[1.x] Deadlock in janus_videoroom


Dev2Trailblazer:
**What version of Janus is this happening on?**
v1.3.0 (dfd86e34112f512b3a15ba6b7b52a3aa01ea347a)

**Have you tested a more recent version of Janus too?**
No.

**Was this working before?**
Yes, before we upgrade from 0.x.

**Is there a gdb or libasan trace of the issue?**
Compilation flags: `-O3 -g3 -ggdb3 -fno-omit-frame-pointer`

https://gist.github.com/Dev2Trailblazer/5dc846843713364a9a2abb41eba11f8b

**Additional context**

New issue from #3478.

The janus-gateway process seems to deadlock and stops responding to HTTP requests. This is quite hard to reproduce, even in production, but when it does so we are able to connect with GDB and debug it. From GDB we were able to capture the stack traces I've posted above.

From @atoppi's analysis we know:

> This thread is holding the `participant->streams_mutex`, while waiting for `handle->mutex`:
> 
> ```
> Thread 6 (LWP 4193):
> #0  0x00007abe10e1a88d in syscall () from target:/lib/x86_64-linux-gnu/libc.so.6
> #1  0x00007abe115a0abc in ?? () from target:/lib/x86_64-linux-gnu/libglib-2.0.so.0
> #2  0x000059efdb13f589 in janus_ice_relay_rtcp (packet=0x7abe0dfcc1a0, handle=0x7abdf4020960) at ice.c:5100
> #3  janus_ice_relay_rtcp (handle=0x7abdf4020960, packet=0x7abe0dfcc1a0) at ice.c:5096
> #4  0x000059efdb13f998 in janus_ice_send_pli_stream (handle=0x7abdf4020960, mindex=1) at ice.c:5160
> #5  0x00007abe0dfe348e in janus_videoroom_reqpli (ps=0x7abdfd0f3610, reason=0x7abe0e041f8d "Keyframe request") at /usr/local/src/janus-gateway/src/janus_videoroom.c:2982
> #6  0x00007abe0e03da4e in janus_videoroom_handler (data=<optimized out>) at /usr/local/src/janus-gateway/src/janus_videoroom.c:10914
> ```
> As a consequence two threads are blocked on the `participant->streams_mutex`:
> ```
> Thread 11 (LWP 200239):
> #0  0x00007abe10e1a88d in syscall () from target:/lib/x86_64-linux-gnu/libc.so.6
[...]

PR #3513, opened 2025-02-07
Fix deadlock on handle lock in janus.c

spscream:
this possibly fixes https://github.com/meetecho/janus-gateway/issues/3510


lminiero:
I think that's a good catch, thanks :+1:  Merging!

PR #3514, opened 2025-02-12
New experimental multistream support in SIP and NoSIP plugins

lminiero:
As the title says, this PR tries to add multistream support to both the SIP and NoSIP plugins. In fact, while the Janus core has supported multistream for quite some time already, only a few plugins were updated to take advantage of that, and that didn't include neither the SIP nor the NoSIP plugin. As a consequence, both plugins were still limited to 1 audio/1 video calls. Considering the context (traditional VoIP), that hasn't been much of an issue, as the overwhelming majority of SIP calls are made of a single audio stream anyway, with the occasional single video stream to come with that. That said, it still was an artificial limitation that should be removed, as there may be  use cases where being able to use more than one audio and/or video stream wuld be helpful. This also paves the way for a less hacky integration of non-audio-video data, e.g., data channels for Real-Time text as done in #3231.

This patch adds multistream support to both the SIP and NoSIP plugins, meaning that both of them can now serve and route multiple m-lines. At the moment, there's a hardcoded limit of 10 m-lines you can have per call: this may or may not be a good/smart idea, but it's done like that as the state for multiple m-lines is kept in a static array. We can always increase the hardcoded limit before merging, or figure out a way to make that dynamic (which would complicate things a bit, though, maybe needlessly so).

I've marked it as a draft pull request as, while this should be functionally working (at least from my limited testing), there's more checks and tests that should be done before this can be merged. There are a few key changes in a few areas, for instance (an important one being us disabling the SOA in Sofia, meaning we control the negotiation in its entirety), and so I want to be 100% sure there aren't any regressions in things that work correctly now. If you use either plugin a lot in production, please do take the time to perform some testing, and provide us with feedback on the results. The sooner we know if everything works or if there are things to change/fix, the sooner we can eventually merge this upstream.

As a side note, this is probably obvious due to the nature of the change, but this will be a 1.x only update. The 0.x version of Janus doesn't support multistream at all, and so both plugins there will not be updated accordingly, meaning there won't be any backport.


lminiero:
The latest commit pulls in the new RTP forwarders, and makes them compatible with the multistream nature of both plugins via an (optional) `mindex` property you can pass for each stream: if the `mindex` is omitted, the RTP forwarding applies to the first media stream of the "right" kind (audio, video), which means the "legacy" API will still work if all you do is a call with a single audio and/or video stream. This hasn't been tested yet, though, so I'll have to check if this actually works, and this doesn't introduce leaks, regressions or crashes. I plan to do that in a few days.

PR #3517, opened 2025-02-13
New programmatic API to ask for keyframes in SIP and NoSIP plugins

lminiero:
This PR adds a new request to both the SIP and NoSIP plugins to programmatically schedule a keyframe request to the WebRTC user, to the SIP peer, or both. This request is particularly useful when the SIP peer doesn't support RTCP PLI, and so may use other mechanisms (e.g., via signalling) to ask for a keyframe to get video working. By using this request, the WebRTC user can ask Janus to originate a PLI programmatically.

The syntax is basic, since in bth plugins the request looks like this:

```
{
    "request": "keyframe",
    "user": <true|false>,
    "peer": <true|false>
}
```

When `user` is `true`, Janus will send an RTCP PLI back to the WebRTC user that sent the request. When `peer` is `true`, the SIP plugin will send an RTCP PLI to the SIP peer (unless PLI was not negotiated on the SIP side, in which case nothing will happen).

I'm planning to merge soon, so please test to make sure this doesn't break anything for you (it shouldn't) or if this doesn't work as expected. This is a patch for `0.x`, but as soon as it's merged I'll push the same fix to `1.x` as well.

### DIFF
--- a/plugins/janus_nosip.c
+++ b/plugins/janus_nosip.c
@@ -120,11 +120,11 @@
  *
  * An \c hangingup event will be sent back, as this is an asynchronous request.
  *
- * Finally, just as in the SIP and SIPre plugins, the multimedia session
+ * Finally, just as in the SIP plugin, the multimedia session
  * can be recorded. Considering the NoSIP plugin also assumes two peers
  * are in a call with each other (although it makes no assumptions on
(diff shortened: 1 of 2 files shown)


lminiero:
Merging.

PR #3518, opened 2025-02-20
Add support for private recordings in Record&Play

lminiero:
As the title says, this patch adds a new property to recordings created using the Record&Play plugin. Following the convention we have in Streaming, AudioBridge and VideoRoom, a recording marked as private is not returned in the list of recordings when a client asks for them. This can be useful when showcasing the plugin in public contexts, like our public demos. Private recordings can still be retrieved if the proper admin key is passed (exactly as in the other mentioned plugins).

The patch slightly changes one of the events (`done`), as it's used as a trigger from a conformation prompt displayed when a recording is done. Now the demo always asks the user if they want to replay the recording they just made: this is especially needed when making private recordings, since in that case those wouldn't appear in the list, and the user wouldn't be able to replay them from there.

Planning to merge soon, and backport to `0.x` after that.

### DIFF
--- a/conf/janus.plugin.recordplay.jcfg.sample.in
+++ b/conf/janus.plugin.recordplay.jcfg.sample.in
@@ -1,7 +1,14 @@
 # path = where to place recordings in the file system
+# admin_key = plugin management secret; if set, private recordings can
+#             be listed only if this key is provided in the request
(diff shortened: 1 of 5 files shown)


lminiero:
Merging.

PR #3519, opened 2025-02-21
Check if IPv6 is disabled when loading the Streaming plugin (fixes #3470)

### LINKED ISSUE #3470: [1.x] can't create socket for streaming endpoint if ipv6 is disabled on host
master

no

**Was this working before?**
I don't have information

no

**Additional context**
We have server with disabled ipv6 and if we create stream with following parameters:
```
    request_body = %{
      "request" => "create",
      "type" => "rtp",
      "id" => stream_id,
      "name" => description,
      "description" => description,
      "media" => [%{
        "type" => "audio",
        "mid" => "1",
        "port" => 0,
        "pt" => 111,
        "codec" => "opus"
      }]
    }
```
we got an error:
```
[ERR] [plugins/janus_streaming.c:janus_streaming_create_fd:7292] [(null)] Cannot create socket for audio... 97 (Address family not supported by protocol)
[ERR] [plugins/janus_streaming.c:janus_streaming_create_rtp_source_stream:7503] [(null)] Can't bind to port 0...
[ERR] [plugins/janus_streaming.c:janus_streaming_process_synchronous_request:3575] Can't add 'rtp' stream '(null)', error creating data source stream...
```

if I specify `iface` parameter for `create` all is working good. I think its due this ternary here(https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_streaming.c#L7289):
```
fd = socket(family == AF_INET ? AF_INET : AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
```

In case iface isn't set when family is 0, but AF_INET is 2.

lminiero:
This patch adds a check on whether or not IPv6 is entirely disabled on the machine, as other plugins (e.g., VideoRoom, AudioBridge, SIP and others) already did. See #3470 for more context.

### DIFF
--- a/src/plugins/janus_streaming.c
+++ b/src/plugins/janus_streaming.c
@@ -1239,6 +1239,7 @@ static janus_mutex config_mutex = JANUS_MUTEX_INITIALIZER;
 static volatile gint initialized = 0, stopping = 0;
 static gboolean notify_events = TRUE;
 static gboolean string_ids = FALSE;
+static gboolean ipv6_disabled = FALSE;
 static janus_callbacks *gateway = NULL;
 static GThread *handler_thread;
 static void *janus_streaming_handler(void *data);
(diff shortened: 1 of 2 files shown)


lminiero:
Merging.

PR #3520, opened 2025-02-25
Upgrade python websockets version to fix the Vulnerability CWE 208

ursravis:
There is vulnerability found in python websocket package 
https://cwe.mitre.org/data/definitions/208.html
The issue is fixed here https://security.archlinux.org/CVE-2021-33880

### DIFF
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -1,2 +1,2 @@
-websockets==8.1
+websockets==9.1
 aiortc==0.9.26
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -1,2 +1,2 @@
-websockets==9.1
+websockets==10.0
 aiortc==0.9.26


januscla:
Thanks for your contribution, @ursravis! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
Thanks for spotting this @ursravis! To be honest, we haven't touched that test in a very long time, so I don't even know if it still works as expected, but it's definitely a good idea to make sure dependencies are updated nevertheless. Out of curiosity, why did you put 10 as the required version? The link you shared seems to say it was fixed in 9.1-1.

ursravis:
Hi @lminiero 
Thanks for checking this PR.
Yes the issue is fixed in 9.1 itself but our internal Vulnerability Monitoring tool suggested to use 10.0 and above.

atoppi:
Thanks @ursravis for pointing this out.
The commit https://github.com/meetecho/janus-gateway/commit/2445b674712cb0093d67d15b5fc10d66e93136d8 supersedes this patch, bumping both `aiortc` and `websockets` to the latest  available versions.
Closing the PR.

PR #3525, opened 2025-03-03
#3523: Using relative path for #include

### LINKED ISSUE #3523: [1.x] Please install .pc file for easy discoverability
The ustream project that uses janus fails to find the janus headers:
In file included from src/acap.c:43:
In file included from src/logging.h:23:
/usr/local/include/janus/plugins/plugin.h:162:10: fatal error: 'refcount.h' file not found
  162 | #include "refcount.h"
      |          ^~~~~~~~~~~~
1 error generated.

This is because janus doesn't provide any way to discover it or to determine what CFLAGS does it need.

The acrual reason is that this CFLAGS argument is missing: -I$(PREFIX)/include/janson.

CFLAGS and LDFLAGS should be supplied by the .pc file so that ```pkg-config --cflags --libs janus``` would print all necessary flags.

mdevaev:
Following #3523

### DIFF
--- a/src/plugins/plugin.h
+++ b/src/plugins/plugin.h
@@ -159,7 +159,7 @@ janus_plugin *create(void) {
 
 #include <glib.h>
 
-#include "refcount.h"
+#include "../refcount.h"
 
 
 /*! \brief Version of the API, to match the one plugins were compiled against
--- a/src/transports/transport.h
+++ b/src/transports/transport.h
@@ -93,7 +93,7 @@ janus_transport *create(void) {
 #include <glib.h>
 #include <jansson.h>
 
-#include "refcount.h"
+#include "../refcount.h"
 
 
 /*! \brief Version of the API, to match the one transport plugins were compiled against */


lminiero:
I'm not sure that's correct, and it's very likely to break the compilation of existing external stuff out there. Our Makefile installs headers to something like `/<path>/include/janus`, so the `refcount.h` file will be found in whatever path is passed as one to find stuff in, rather than relatively to the `plugin.h` folder. As such, for third party plugins (like yours, I assume), you should do different includes there, not modify this one.

In the Janus NDI plugin, for instance, we define CFLAGS [like this](https://github.com/meetecho/janus-ndi/blob/main/Makefile#L11):

```
CFLAGS += -I$(JANUSP)/include/janus $(shell pkg-config --cflags glib-2.0 jansson opus libcurl) -D_GNU_SOURCE -DHAVE_SRTP_2
```

and then the plugin includes the `plugin.h` file [like this](https://github.com/meetecho/janus-ndi/blob/main/src/janus_ndi.c#L6):

```
#include "plugins/plugin.h"
```

and it works.

lminiero:
That said, I tried your patch with the NDI plugin and it still works, so I may be wrong about the potential impact.

mdevaev:
It shouldn't break anything because you are giving a clear indication to the processor about the location of the header relative to the position of the current file. 

I don't really like using `-I$(JANUSP)/include/janus` because it robs the code of the convenient and understandable prefix `janus/`, showing where everything comes from.

lminiero:
> I don't really like using `-I$(JANUSP)/include/janus`

I get that, but that's what many are using, so I can't just ignore that, and if it could break things for people out there.

lminiero:
Makes sense. Merging then, thanks!

mdevaev:
Thank you!

PR #3528, opened 2025-03-14
rtcp: fix FIR parsing

gdesmott:
FIR RTCP packets sent by GStreamer were not handled by Janus but PLI ones are.

According to https://www.rfc-editor.org/rfc/rfc5104#section-4.3.1 FIR are PSFB packets with FMT=4 so I'm now parsing them as we do with PLI.

### DIFF
--- a/src/rtcp.c
+++ b/src/rtcp.c
@@ -1154,6 +1154,12 @@ gboolean janus_rtcp_has_fir(char *packet, int len) {
 		switch(rtcp->type) {
 			case RTCP_FIR:
 				return TRUE;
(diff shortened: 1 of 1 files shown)


lminiero:
Adding a new toggle is fine, but I wouldn't remove the check we had, since that was definitely working with other implementations we tested FIR with back then (which is arguable ages ago, before we definitely moved to PLI).

atoppi:
LGTM :+1:

lminiero:
Thanks, merging then! I'll backport that to `0.x` as well if it turns out it's needed.

PR #3534, opened 2025-04-02
Allow plain RTP participants in AudioBridge to use generate_offer too (replaces #3380)

lminiero:
As the title says, this PR extends the existing `generate_offer` functionality in the AudioBridge, which was previously limited to WebRTC users, to plain RTP participants. This basically allows you to ask the plugin to first provide their own RTP contact details, and then provide your own. This enables a couple of interesting scenarios:

1. allowing you to have the AudioBridge "dial-out" (e.g., when using this with SIP)
2. allowing for cascading of AudioBridge instances via plain RTP, as now one can be the "offerer" and the other the "answerer".

I already played a bit with the SIP scenario, by extending the [demo I wrote at the time](https://www.meetecho.com/blog/audiobridge-sip/) to now have the AudioBridge "call" a SIP endpoint and add them to a room, and it seems to be working as expected. I plan to play a bit with cascading in the next few days, instead. I may provide some more details in a new blog post, if I think there's anything interesting to tell.

As the title says, this replaces #3380, which was a PR contributed some months ago to basically try and do the same thing, even though from a different perspective. I personally felt the PR to be a bit too cumbersome, so I tried to see if I could come up with a simpler patch to try and reuse the existing pattern if possible. @keremcadirci kudos nevertheless to you for the contribution! I'm definitely interested to hear if this addresses the requirement you had at the time.

As usual, feedback welcome if you play with this. The plan is to backport this to `0.x` when it's merged.

### DIFF
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -945,6 +945,9 @@ room-<unique room ID>: {
  * Notice that, after a plain RTP session has been established, the
  * AudioBridge plugin will only start sending media via RTP after it
  * has received at least a valid RTP packet from the remote endpoint.
+ * If you're interested in supporting scenarios where the AudioBridge
+ * "dials out" (e.g., for outgoing INVITES to SIP endpoints) check the
+ * \ref aboffer section.
  *
  * At this point, whether the participant will be interacting via WebRTC
  * or plain RTP, the media-related settings of the participant can be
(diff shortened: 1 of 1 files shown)


lminiero:
For those interested in more details on this (and cascaded mixing) I [blogged](https://www.meetecho.com/blog/cascaded-mixing/) about it.

lminiero:
Merging.

Issue #3537, opened 2025-04-14
[1.x] videoroom: dummy publishers leaking


m08pvv:
**What version of Janus is this happening on?**
505d97d7fee6bc6370f28bf5e51d491cc5716b27

**Have you tested a more recent version of Janus too?**
it's latest master so far

**Was this working before?**
no idea

**Is there a gdb or libasan trace of the issue?**
some refcount logs: https://gist.github.com/m08pvv/46cc52ec5b7d6134c7b4a3de7d874b66
(full log is from a branch where I tried to add a decrease in janus_videoroom_publisher_stream_free)

**Additional context**
we use dummy publishers and recordings
tried to add a decrease in janus_videoroom_publisher_stream_free, but it seems that it wasn't called (at least it's not in the log which contains all output)

m08pvv:
So, found in `janus_videoroom_publisher_stream_destroy` a comment `/* TODO Should unref the publisher instance? */` and in `janus_videoroom_create_dummy_publisher` there is a code:
`ps->publisher = publisher;
janus_refcount_increase(&publisher->ref);	/* Add a reference to the publisher */`
So it seems that not only publisher stream waiting for publisher to be destroyed (refcount 0), but also publisher waits for publisher streams to be destroyed. Sounds like a circular dependency.

m08pvv:
@atoppi, yes, it seems that it's not leaking anymore.

PR #3540, opened 2025-04-21
Remove refcount cyclic dependency publisher <=> stream

m08pvv:
As mentioned in  https://github.com/meetecho/janus-gateway/issues/3537
It seems that dummy publisher counts references to streams and streams count references to dummy publisher (a circular dependency).
Tried 505d97d7fee6bc6370f28bf5e51d491cc5716b27 under valgrind - got messages about dummy publisher-related leaks.
With this patch (everything else was exactly the same) valgrind was silent!
It seems that it fixes this leak, but I'm not quite sure what was the purpose of that circular dependency (maybe it's something important that I'm missing while looking just at dummy publishers).

### DIFF
--- a/src/plugins/janus_videoroom.c
+++ b/src/plugins/janus_videoroom.c
@@ -3166,7 +3166,6 @@ static void janus_videoroom_create_dummy_publisher(janus_videoroom *room, GHashT
 		g_snprintf(mid, sizeof(mid), "%d", mindex);
 		ps->mid = g_strdup(mid);
 		ps->publisher = publisher;
-		janus_refcount_increase(&publisher->ref);	/* Add a reference to the publisher */
 		ps->active = TRUE;
 		ps->acodec = room->acodec[i];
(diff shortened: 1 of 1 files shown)


lminiero:
I was working on a patch before the Easter holiday break that addressed that too. Long story short, just removing the references isn't enough, as there's some more subtle consequences of that. I'll post my PR once I get back to work and I have a chance to give it another look.

lminiero:
Please see the PR above, which includes the additional changes I mentioned. Notice that it does _not_ remove references from remote publishers, which is what you did in your patch: those references are needed, as remote publishers go through a proper cleanup just as regular publishers do. Dummy publishers don't, which is why cleanup is different for them, and why those additional references were not being removed.

m08pvv:
Closed because it was just a dummy approach to fix the result, not the root cause.
The proper fix is in https://github.com/meetecho/janus-gateway/pull/3541

PR #3541, opened 2025-04-22
Fix leaks caused by dummy publishers (fixers #3537)

lminiero:
This is an attempt to fix the leaks in the VideoRoom plugin when dummy publishers are used in a room. It fixes the issue opened in #3537 and is a replacement for the partial (and possibly broken) PR contributed in #3540.


lminiero:
Merging :v:

PR #3546, opened 2025-05-13
Remove "nice_debug" option and "set_libnice_debug" API from Janus

atoppi:
To enable libnice logging:
- Set `NICE_DEBUG` env var to `all` or one of these values https://gitlab.freedesktop.org/libnice/libnice/-/blob/ae3eb16fd7d1237353bf64e899c612b8a63bca8a/agent/debug.c#L59
- Set `G_MESSAGES_DEBUG` env var to `all` or one of these values https://gitlab.freedesktop.org/libnice/libnice/-/blob/ae3eb16fd7d1237353bf64e899c612b8a63bca8a/agent/debug.c#L67

E.g.
```
NICE_DEBUG=nice G_MESSAGES_DEBUG=libnice /path/to/janus/bin
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,8 +285,6 @@ or on the command line:
                                   (experimental)  (default=off)
 	-O, --ipv6-link-local         Whether IPv6 link-local candidates should be
                                   gathered as well  (default=off)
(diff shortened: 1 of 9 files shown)


lminiero:
LGTM :+1:

PR #3550, opened 2025-05-19
Listen for an optional signal handler for log rotation

atoppi:
See #3548 for context.

Add a CLI parameter "--log-rotate-sig" / "-R" and a jcfg option "log_rotate_sig" to setup a signal handler used for log rotation.
The only supported values are "SIGUSR1" and "SIGHUP" for now.
The handler will reload (flush, close, open) the log file when receiving the signal.
By default the option is unset, which disables the signal handler for log rotation.


lminiero:
Merging :v:

PR #3551, opened 2025-06-05
test_aiortc: only start Janus if its not running

sudipm-mukherjee:
On Debian and Ubuntu autopkgtests, the built package will be installed before running the test script. And Janus from the installed package is already running. Starting another Janus and trying to kill it causes the tests to fail.

Check if Janus already running before launching a new Jaanus.

### DIFF
--- a/test/test_aiortc.sh
+++ b/test/test_aiortc.sh
@@ -11,9 +11,14 @@ JANUS_SRC="$( dirname $SCRIPTPATH )"
 TEST=${1-"$SCRIPTPATH/echo.py"}
 URL=${2-"ws://localhost:8188/"}
 
-echo "Starting Janus binary from $JANUS_SRC ..."
-$JANUS_SRC/src/janus >/dev/null 2>&1 &
-JANUS_PID=$!
+JANUS_RUNNING=$(pgrep -xc janus || true)
+if [ $JANUS_RUNNING -gt 0 ]; then
+    echo "Not starting Janus, already running ..."
+else
+    echo "Starting Janus binary from $JANUS_SRC ..."
+    $JANUS_SRC/src/janus >/dev/null 2>&1 &
+    JANUS_PID=$!
+fi
 
 echo "Waiting for some seconds before launching the test ..."
 sleep 5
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @sudipm-mukherjee! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

test/test_aiortc.sh
```diff
@@ -11,9 +11,14 @@ JANUS_SRC="$( dirname $SCRIPTPATH )"
 TEST=${1-"$SCRIPTPATH/echo.py"}
 URL=${2-"ws://localhost:8188/"}
 
-echo "Starting Janus binary from $JANUS_SRC ..."
-$JANUS_SRC/src/janus >/dev/null 2>&1 &
-JANUS_PID=$!
+JANUS_RUNNING=$(ps ax | grep "bin\/janus" | grep -v grep | wc -l)
```
atoppi, line 14: This seems convoluted and overly complicated.

I'd suggest using instead:
```bash
JANUS_RUNNING=$(pgrep -x janus)
JANUS_RUNNING=${JANUS_RUNNING:-0}
```

atoppi (commented):
Thanks for the patch.
I only have one comment, suggesting a more compact and idiomatic way to detect if the process is running.

test/test_aiortc.sh
atoppi: What if you replace
```bash
JANUS_RUNNING=$(pgrep -x janus)
JANUS_RUNNING=${JANUS_RUNNING:-0}
```
with:
```bash
JANUS_RUNNING=$(pgrep -xc janus || true)
```

sudipm-mukherjee: thanks, that has worked in both Debian and Ubuntu.

atoppi:
@sudipm-mukherjee please sign the CLA

sudipm-mukherjee:
> @sudipm-mukherjee please sign the CLA

Tried, but that is asking for informations like my address, my contact numbers, my `employer's policy on contributing to open source projects`.
I think its best to close this PR.

lminiero:
You don't need to provide detailed info. The address can be the company official address, for instance, same for contact number. The policy for contributing is mostly informational, so you can put what you want.

PR #3556, opened 2025-06-28
manual ringing

adnanel:
Currently, SIP plugin sends ringing automatically when an incoming call is received. This doesn't allow us to have a real ringing indicator if our WebRTC side has a ringing notifier. 

For example, we have certain setups where we asynchronously detect that the call cannot be handled on WebRTC and want to reject the incoming call altogether, and as such there was never any actual ringing.

SIP side, however, thinks the call was ringing and then rejected.

I'd like to have a way for janus not to send this event. Instead, when we receive confirmation that the call is actually ringing, we would then ask janus to actually emit the ringing event and continue normally.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -115,6 +115,7 @@
 	"force_tcp" : <true|false; if true, forces TCP for the SIP messaging; optional>,
 	"sips" : <true|false; if true, configures a SIPS URI too when registering; optional>,
 	"rfc2543_cancel" : <true|false; if true, configures sip client to CANCEL pending INVITEs without having received a provisional response first; optional>,
+	"automatic_ringing" : <true|false; if false, don't generate ringing automatically as soon as an INVITE; optional>,
 	"username" : "<SIP URI to register; mandatory>",
 	"secret" : "<password to use to register; optional>",
 	"ha1_secret" : "<prehashed password to use to register; optional>",
@@ -808,6 +809,7 @@ static struct janus_json_parameter register_parameters[] = {
 	{"force_tcp", JANUS_JSON_BOOL, 0},
 	{"sips", JANUS_JSON_BOOL, 0},
 	{"rfc2543_cancel", JANUS_JSON_BOOL, 0},
+	{"automatic_ringing", JANUS_JSON_BOOL, 0},
 	{"username", JSON_STRING, JANUS_JSON_PARAM_REQUIRED},
 	{"secret", JSON_STRING, 0},
 	{"ha1_secret", JSON_STRING, 0},
@@ -1047,6 +1049,7 @@ typedef struct janus_sip_account {
 	gboolean force_tcp;
 	gboolean sips;
 	gboolean rfc2543_cancel;
+	gboolean automatic_ringing;
 	char *username;
 	char *display_name;		/* Used for outgoing calls in the From header */
 	char *authuser;			/**< username to use for authentication */
@@ -2323,6 +2326,7 @@ void janus_sip_create_session(janus_plugin_session *handle, int *error) {
 	session->account.force_tcp = FALSE;
 	session->account.sips = FALSE;
 	session->account.rfc2543_cancel = FALSE;
+	session->account.automatic_ringing = TRUE;
 	session->account.username = NULL;
 	session->account.display_name = NULL;
 	session->account.user_agent = NULL;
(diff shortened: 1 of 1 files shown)


lminiero:
I do agree that applications should have control on whether a ringing should be sent back automatically or not, since Janus is not the end user. Of course whatever change would need to be optional and defaulting to disabled for backwards compatibility. I'll review your PR in the next few days and get back to you. Thanks!

src/plugins/janus_sip.c
```diff
@@ -4947,6 +4958,16 @@ static void *janus_sip_handler(void *data) {
 			/* Notify the operation */
 			result = json_object();
 			json_object_set_new(result, "event", json_string("infosent"));
+		} else if(!strcasecmp(request_text, "send_ringing")) {
+			if(session->status != janus_sip_call_status_invited
+				&& session->status != janus_sip_call_status_progress) {
```
lminiero, line 4963: Code style nit: the `&&` must be on the previous line, and this second line must have a second tab of indentation.

src/plugins/janus_sip.c
```diff
@@ -4947,6 +4958,16 @@ static void *janus_sip_handler(void *data) {
...
+				JANUS_LOG(LOG_ERR, "Wrong state (not invited or progress? status=%s)\n", janus_sip_call_status_string(session->status));
+				g_snprintf(error_cause, 512, "Wrong state (not in a call?)");
+				goto error;
+			}
+			if (session->stack->s_nh_i) {
```
lminiero, line 4968: Code style nit: no space between if and bracket. Besides, since the action is simple, there's no need for curly brackets.

src/plugins/janus_sip.c
```diff
@@ -5644,8 +5665,10 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			su_free(session->stack->s_home, callee_text);
 			g_free(referred_by);
 			if(!reinvite) {
-				/* Send a Ringing back */
-				nua_respond(nh, 180, sip_status_phrase(180), TAG_END());
+				if (session->account.automatic_ringing) {
```
lminiero, line 5668: Code style: space after if.

lminiero (commented):
I looked at the patch and, apart from a few editorial notes, it seems fine to me. The only doubt I have is related to the `send_ringing` request: I was wondering it it may make sense to have a generic method for sending generic error codes back, but thinking about it it would probably be a mess, as it would complicate the checks on when you can send stuff, and we have dedicated messages for rejecting calls for instance.

adnanel:
I apologize for the codestyle mistakes. Should be fine now.

lminiero:
> I apologize for the codestyle mistakes. Should be fine now.

No need to apologize at all, I myself forget about whatever style has been used throughout the code sometimes :rofl:

src/plugins/janus_sip.c
```diff
@@ -4947,6 +4958,14 @@ static void *janus_sip_handler(void *data) {
...
 			result = json_object();
 			json_object_set_new(result, "event", json_string("infosent"));
+		} else if(!strcasecmp(request_text, "send_ringing")) {
+			if(session->status != janus_sip_call_status_invited && session->status != janus_sip_call_status_progress) {
+				JANUS_LOG(LOG_ERR, "Wrong state (not invited or progress? status=%s)\n", janus_sip_call_status_string(session->status));
+				g_snprintf(error_cause, 512, "Wrong state (not in a call?)");
+				goto error;
+			}
+			if(session->stack->s_nh_i)
+				nua_respond(session->stack->s_nh_i, 180, sip_status_phrase(180), TAG_END());
```
lminiero, line 4968: I have one small doubt: shouldn't this part also check whether automatic ringing is enabled, and/or whether or not we already sent a 180 back? Not sure what happens in the SIP world if you send a 180 twice or more.

adnanel: Yes, I had the same concern about state corruption if you do this multiple times. Frankly I'm no SIP expert but to my knowledge (based on quick google searching RFCs), it's specified that multiple progress messages may be sent freely:
> A UAS MAY
   send as many provisional responses as it likes.  Each of these MUST
   indicate the same dialog ID.  However, these will not be delivered
   reliably.

https://www.rfc-editor.org/rfc/rfc3261.html#section-13.3.1.1

For the first question - the same rationale as above, having automatic ringing doesn't mean someone wouldn't want to send multiple progresses later on - for what purpose I do not know, but I figured I Janus shouldn't be too opinionated in this regard.

lminiero: Makes sense, thanks for the feedback! I think this is good to merge, so I'll go ahead, and see if this is easy to backport to `0.x` too. Thanks again for your contribution!

As a side and unrelated note, not sure if you had the chance to look at #3514: as a regular user of the SIP plugin, I'd be interested in feedback on that part (even though it may not be needed in your scenarios).

PR #3557, opened 2025-06-30
Fix some crashes when remote publishers are used

atoppi:
This PR fixes some issues encountered when dealing with remote publishers:

- More robust checks before requesting a PLI to a publisher session (we got a crash due to an invalid session when a remotized publisher received a RTCP PLI through a forwarder)
- Safe removal of subscriber streams when a remote publisher has gone (there was an unprotected access to a stream list)


lminiero:
Merging.

PR #3564, opened 2025-07-22
Refactor keyframe buffering in Streaming plugin to store following deltas too

lminiero:
Hi all,

this pull request is a sponsored development effort funded by our friends at [Byborg](https://www.byborgenterprises.com/), so thank you, guys!

To provide some context on the changes this PR introduces, it's a good idea to first make a step backward, and describe what was there originally, what is was trying to solve and why that wasn't enough.

## Motivation

The main problem is that, when using the Streaming plugin, new viewers may join a mountpoint at any time. With video having keyframes at more or less specific intervals, this means that new viewers may jump in a few seconds after a keyframe was received, and so wouldn't be able to decode/display any video until the subsequent keyframe is received, which could take a while depending on the keyframe frequency of the video. The usual way subscribers handle that with WebRTC is by sending a PLI message, to ask the source for a new keyframe right away, but that only works if the source can actually handle it: for Streaming plugin subscribers, this definitely works if the mountpoint source is, e.g., a VideoRoom publisher using RTP forwarders with RTCP support, but that would not work if the source is completely external (e.g., a generic external script feeding RTP packets to the mountpoint, or an RTSP server), as either there's no way to relay the PLI to that generic source, or they wouldn't support it anyway.

## Existing (faulty) solution

A few years ago we added a tentative way to address that, where we basically would optionally store in the Streaming plugin the latest keyframe we received from the source, which means storing in memory all the RTP packets composing the latest keyframe: then, for new viewers, we'd send those RTP packets first, and then start serving the "fresh" packets from the source. This helped, but introduced other problems:

1. Since we were only storing a single frame (the latest keyframe), a new viewer would indeed be able to decode that, but then that decoded image would often be displayed as frozen until the next keyframe, since fresh deltas would not apply to that, and the decoder would not know how to deal with them.
2. When not frozen, the video would actually show a lot of artifacts, for the same reason: fresh deltas would refer to previous deltas the viewer never received, rather than the base keyframe, causing the decoder to get confused and as such generate ghosting effects and artifacts.

While some people are ok with that effect while waiting for a keyframe, there are others that aren't, mostly because it suggests something's wrong with the stream when there really isn't anything wrong: it's just a matter of waiting for a proper sync point. We personally preferred never to use that feature, although available in Janus, as I'm of those that think that a frozen image for a few seconds does indeed give the wrong impression: better not to show anything at all (or a spinning wheel), rather than to show something that suggests technical issues.

## New approach

Which now brings us to this PR. Our friends at [Byborg](https://www.byborgenterprises.com/) contacted us to try and come up with an alternative solution to just storing the keyframe, and the idea was to not only store the keyframe itself, but also following deltas, and then shoot all those RTP packets to new viewers before fresh ones, in order to give recipients all the context they need to kickstart a new Streaming plugin session without waiting and without freezes/artifacts. This sounded like a doable idea, even though with a couple of important drawbacks to take into account:

1. Sending the latest keyframe + deltas to a new viewer as soon as they connect may cause a huge burst of data to be sent: this may cause trouble, depending on how many packets need to be sent and the available bandwidth the recipient has.
2. Sending those RTP packets without rewriting timestamps would cause an unrecoverable delay in the video displayed by the recipient, as the starting reference time would be the one from the keyframe itself, which may have originally been received many seconds ago.

We addressed those drawbacks by adding some options to configure the behaviour:

* As far as timestamps are concerned, the idea was to basically rewrite them so that all stored packets would be sent as if they came right one after the other, so with a small (1) gap between them; this is supposed to trick the recipient into thinking that, for a little while, framerate was really high, and so cause a fast decode through keyframe and deltas to catchup with fresh frames ASAP (which is what video players internally do as well, when seeking).
* To limit the impact of bursts, we added a couple of options to configure this new behaviour, where when enabling the feature you can choose how many seconds of data you want to store at max (starting from the keyframe) and/or enforce a cap on the overall bytes you want to store (again starting from the keyframe). This effectively moves the problem to who creates the mountpoint, but since they know the application space this would be used in, it at least gives them control on how much "bursting" they may considered acceptable.

## Enabling the feature

Enabling the feature is relatively easy, since it only requires you to set a new global feature when creating an RTP or RTSP mountpoint. Specifically, when you create a new mountpoint (statically, in the config file, or dynamically, via API) there are two new mountpoint-level properties you can configure:

* `bufferkf_ms`
* `bufferkf_bytes`

The Streaming plugin configuration file has a new extensive documentation section about them, so I encourage you to refer to that. In a nutshell, as the names suggest, `bufferkf_ms` imposes a temporal limit (only store RTP packets up to X milliseconds), while `bufferkf_bytes` imposes a size limit (only store RTP packets as long as the sum of their size does not exceed X bytes). Both are disabled by default (value 0). The two properties are not mutually exclusive: if both are set, the first one that is triggered causes following RTP packets not to be stored. Limits are reset any time a new keyframe is received, since the previous list is pruned and we start from scratch again: the RTP packets belonging to the keyframe itself are not impacted, which means that if a keyframe exceeds the value set in `bufferkf_bytes`, those RTP packets will be saved nevertheless, but the following deltas won't (as they would cross the already exceeded configured limit).

Since they're mountpoint level, this means they automatically apply to all video streams configured in that specific mountpoint: each video stream would have their own, separate, keyframe+deltas buffer. This also makes it work for RTSP streams, where we don't know the video profile in advance.

## Backwards compatibilty

The refactoring of the feature was conceived to still allow people to use the previous "only save the keyframe" behaviour, if they liked that. While the old `videobufferkf` is deprecated and ignored (a warning is displayed if you try to use it), you can still get the same behaviour if the set `bufferkf_ms` or `bufferkf_bytes` to `1`: both will only give the code enough "room" to only save the keyframe itself (which is always excluded by those limits), while dropping the following deltas.

## Testing the feature

I tested the feature briefly, and it seems to be working as expected in both Firefox and Chrome. That said, more extensive testing should be performed, especially to analyse the impact of bursting in networks other than the local setup I was using. If you are interested in the future, please do test this PR and let us know, taking into account that, due to the drawbacks we mentioned, this isn't supposed to be a magical solution that will work in all situations (especially in case keyframe intervals are very long).

Please notice that, to facilitate debugging, at the moment the code prints quite verbose logging for all packets as they're stored, in order to have an explicit indication of what an incoming packet is (keyframe vs delta) and how it's handled (e.g., whether it's being stored, or ignored because of exceeded limits). That logging will of course disappear when I merge this PR, and be moved to much lower debug levels than `LOG_INFO` and `LOG_WARN`.

Also notice that the code does make some assumptions, in particular on the order of incoming packets. It currently expects incoming RTP packets to be ordered, as implementing reordering would have made the code way more complex than that: this may have an impact on how we store packets, since we currently store different frames with different SSRCs to trigger the automatic "timestamp+1" mechanism to simulate the fast decode functionality, and out of order packets may incorrectly be marked as frames of their own. Implementing reordering, if needed, can be delegated to other efforts in the future: the idea is that, when WebRTC publishers are not involved (for which as mentioned we already have RTCP support in RTP forwarders todeal with keyframes automatically), generic sources are supposed to be close enough to Janus instances that this shouldn't be a big deal.

## Feedback welcome

I've talked enough: now go and test this :grimacing:

### DIFF
--- a/conf/janus.plugin.streaming.jcfg.sample.in
+++ b/conf/janus.plugin.streaming.jcfg.sample.in
@@ -1,3 +1,14 @@
+# You can configure static mountpoints that should be made available
+# when Janus starts in this configuration file. The syntax and the
+# available properties are listed below. Notice that, for the sake of
+# simplicity, only the global and legacy properties are listed in the
+# following text: you can use the new stream-based syntax as well (see
+# the 'multistream-test' example below for reference), but in that case
+# all properties whose names start with audio/video/data should be
+# renamed to have that prefix removed. Please refer to the Streaming
+# plugin documentation for more details, as the static configuration
+# in that case mirrors the dynamic API syntax.
+#
 # stream-name: {
 # type = rtp|live|ondemand|rtsp
 #        rtp = stream originated by an external tool (e.g., gstreamer or
@@ -36,8 +47,6 @@
 # videoiface = network interface or IP address to bind to, if any (binds to all otherwise)
 # videopt = <video RTP payload type> (e.g., 100)
 # videocodec = name of the video codec (e.g., vp8)
-# videobufferkf = true|false (whether the plugin should store the latest
-#		keyframe and send it immediately for new viewers, EXPERIMENTAL)
 # videosimulcast = true|false (do|don't enable video simulcasting)
 # videoport2 = second local port for receiving video frames (only for rtp, and simulcasting)
 # videoport3 = third local port for receiving video frames (only for rtp, and simulcasting)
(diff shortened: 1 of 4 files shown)


lminiero:
Merging this PR, thanks again to [Byborg](https://www.byborgenterprises.com/) for making this possible!
In a couple of weeks I'll see about porting this to 0.x as well, unless this requires too much effort.

PR #3565, opened 2025-07-26
Fix memory leak in janus_sip.c

oleksandr-mihal-zultys:
Fixed memory leak in janus_sip_get_incoming_headers().
json_object_set() does not require a new memory allocation for the key.


lminiero:
That's a good catch, but I'm wondering how it may have gone unnoticed for ~6 years, since this was first added in #1873. It may be because I never use the feature myself, and so my libasan setup never warned me about the leak: I guess it's such a tiny leak it never caused any issue for people who do use it. I'll make a couple of tests to replicate and get back to you.

lminiero:
Just tested and can confirm the leak, so I'll merge both PRs, thanks!

oleksandr-mihal-zultys:
Thank you for comments!

Issue #3573, opened 2025-08-27
[1.x] `TrackOption.capture` type does not support `MediaStreamTrack`


manupik:
The `TrackOption` type should allow the `capture` property to accept a `MediaStreamTrack`, in addition to the existing options.

**Current definition**
```ts
type TrackOption = {
	// ...
	capture: boolean | MediaTrackConstraints;
	// ...
}
```

**Suggested update**
```ts
type TrackOption = {
	// ...
	capture: boolean | MediaTrackConstraints | MediaStreamTrack;
	// ...
}
```

**Rationale**
This is consistent with both the implementation and the documentation:

- **Code reference:**
https://github.com/meetecho/janus-gateway/blob/90677e2b69ebe8997bcfa90f21ada069e62e7fdc/html/demos/janus.js#L2484

- **Documentation excerpt:**

> `capture`: in case something must be captured (e.g., a microphone for "audio" or a "webcam" for video), the `capture` property can be used to dictate what and how; passing `true` asks for the default device, but `getUserMedia` (for audio/video) or `getDisplayMedia` (for screen sharing) constraints can be passed as well as objects, and in case those will be used instead; passing a `MediaStreamTrack` instance will tell the library not to capture anything, but use the provided track as a source instead.
> https://janus.conf.meetecho.com/docs/JS.html

lminiero:
That makes sense, thanks for spotting this. Could you create a PR with your proposed change? I'm out of office for a few days but I can review it next week.

manupik:
Sure, it’s up already. 👍🏼

PR #3574, opened 2025-08-27
Fix `TrackOption.capture` type to allow `MediaStreamTrack`

manupik:
Updates the `TrackOption` type definition so that the `capture` property accepts `MediaStreamTrack` in addition to `boolean` and `MediaTrackConstraints`. This aligns the types with both the documentation and the actual implementation.

Closes [#3573](https://github.com/meetecho/janus-gateway/issues/3573)


lminiero:
Thanks, looks good to me! Merging :pray:

PR #3578, opened 2025-09-12
Add mechanism for notifying about RTSP disconnections/reconnections (fixes #3577)

### LINKED ISSUE #3577: [1.x] Publish event when streaming source is not available
**Description**
As an administrator, I would like to be notified when an RTSP source used by the streaming plugin becomes unavailable, so that this error can be forwarded to the application user.

**Current Behavior**
The streaming plugin has been successfully configured and publishes an RTSP stream. However, when the RTSP source goes down, only log messages are generated, for example:
```
[ERR] [plugins/janus_streaming.c:janus_streaming_rtsp_connect_to_server:8164] Couldn't send DESCRIBE request: Couldn't connect to server (Failed to connect to 192.168.178.46 port 8554 after 0 ms: Couldn't connect to server)
[WARN] [99] Reconnection of the RTSP stream failed, trying again in a few seconds...
```

Currently, these events are not published to the event handler or exposed to the application user.

**Proposed Improvement**
Enable the streaming plugin to publish events when an RTSP source is unavailable. This would allow administrators or applications to handle such errors proactively and inform users accordingly.

**Reference**
https://janus.discourse.group/t/handle-unavailable-stream/1713

**Benefit** 
- Allows applications to provide timely feedback to users
- Reduces downtime caused by unnoticed stream failures

Cheers,
Robert

lminiero:
See #3577 for context. By default, this only notifies the event via event handlers, when enabled: to also notify subscribers, you need to set the `rtsp_notify_changes` property to `true` when creating the new mountpoint (statically or dynamically).

Please test and provide feedback.

### DIFF
--- a/conf/janus.plugin.streaming.jcfg.sample.in
+++ b/conf/janus.plugin.streaming.jcfg.sample.in
@@ -148,6 +148,8 @@
 # 		formula: timeout = min(session_timeout, rtsp_session_timeout / 2). (default=0s)
 # rtsp_timeout = communication timeout (CURLOPT_TIMEOUT) for cURL call gathering the RTSP information (default=10s)
 # rtsp_conn_timeout = connection timeout for cURL (CURLOPT_CONNECTTIMEOUT) call gathering the RTSP information (default=5s)
+# rtsp_notify_changes = if set to true, will send an event to connected users when the RTSP session
+#		gets disconnected, and when it's reconnected (default=false)
 #
 # Notice that, for 'rtsp' mountpoints, normally the plugin uses the exact
 # SDP codec and fmtp attributes the remote camera or RTSP server sent.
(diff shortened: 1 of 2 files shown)


yoolandes:
@lminiero 
I could successfully test your changes.
- [x] Notifications via event handler
- [x] Messages via subscribers

### Event handler
```
{
  emitter: 'MyJanusInstance',
  type: 64,
  timestamp: 1758054384739269,
  event: {
    plugin: 'janus.plugin.streaming',
    data: { event: 'rtsp-reconnected', id: 99 }
  }
}

{
  emitter: 'MyJanusInstance',
  type: 64,
  timestamp: 1758054312821737,
  event: {
    plugin: 'janus.plugin.streaming',
    data: { event: 'rtsp-disconnected', id: 99 }
  }
}
```

### Message via subscriber
```
{
    "event": "rtsp-reconnected",
    "id": 99
}
```

Thank you very much for your PR!

lminiero:
rtsp_notify_changes is a per-mountpoint property, not a global one.

lminiero:
@yoolandes I pulled the crash fix in this branch as well, so please provide me some feedback on this too when you can.

yoolandes:
@lminiero
I’ve tested the code changes, and everything works well on my end with no issues.

lminiero:
Ack, thanks :+1: 
I'll merge, then, and backport to `0.x` if it isn't too hard.

Issue #3582, opened 2025-09-24
[1.x] Audiobridge global mjr flag ignored for resume mechanic


ettubrute47:
**What version of Janus is this happening on?**
1.3.2 a44ebb646a26439fadcbffb86425b5f3d3da2995

**Have you tested a more recent version of Janus too?**
No, exists in most recent version

**Was this working before?**
No

**Is there a gdb or libasan trace of the issue?**
Not that kind of bug

**Additional context**
If you set mjrs = true in room definition, or use enable_mjrs to turn on individual mjr recordings, it sets audiobridge->mjrs = true. However if you suspend an individual (with stop_record = true), and then resume them via resume api, it doesn't check this flag, it only resumes recording if you specifically tell it in that api call. Not really a bug since documentation is accurate, it makes clear resume record is false by default. 

I think it should default to the global state of the room instead. I believe all that is involved is modifying resume check to look at audiobridge->mjrs too, e.g., `if(audiobridge->mjrs || record && json_is_true(record))`, and then update documentation to say default is to room mjrs state. This could've been purposeful decision, but in cursory look didn't find any discussions on this.

lminiero:
> I believe all that is involved is modifying resume check to look at audiobridge->mjrs too, e.g., `if(audiobridge->mjrs || record && json_is_true(record))`, and then update documentation to say default is to room mjrs state

While I understand your point, this would be a breaking change, and could cause a behaviour people using the feature in the wild right now wouldn't expect. In general, when patching features for things that aren't really bugs, we try to avoid changing the behaviour.

My feeling is that this is better handled by explaining it clearly in the docs that resuming recording must be explicit for automatically recorded rooms. As an alternative, we could add a new flag for that, e.g., something like "mjrs_autoresume" that, if set to `true` for rooms with `mjrs=true`, does what you're asking here (default being `false` to ensure nothing changes for people that don't touch anything).

ettubrute47:
Yea oops I'm being dumb definitely would have ended up with people all of a sudden getting recordings when they weren't expecting it if we did what I suggested...

Yea I'm following you, I agree with the alternative, if there is anything to do about it to begin with (I know extending API shouldn't be taken lightly either). I'd like to try to do a pull request implementing the mjrs_autoresume flag as a small exercise for me though, if that's cool, no pressure in taking it.

lminiero:
[...]

PR #3584, opened 2025-09-25
Add mjrs_autoresume flag to audiobridge plugin (#3582)

ettubrute47:
This is meant to allow automatically restart recording on resume from suspend without having to specify in the resume API call record: true if mjrs is enabled for the room. To avoid breaking change we add a flag mjrs_autoresume, default to false. See discussion in #3582 for further details.

I have tested compilation and acceptance of the flag in parsing the config, but have not tested the actual logic via suspend/resume yet. I will add to this when I do, but publishing this now for code visibility.


lminiero:
Apologies for the late feedback, and thanks for the PR! From a glance it looks ok, and the items I suggested in the issue seem all handled correctly. Looking forward to your testing in that regard!

lminiero:
@ettubrute47 any update on testing?

lminiero:
@ettubrute47 ping :grin:

PR #3586, opened 2025-10-04
Fix post-processor compile errors with FFmpeg 8

i8-pi:
FFmpeg 8.0 removed the deprecated `avcodec_close` function. This change fixes the compile errors caused by that

`avcodec_free_context` was added in 2014 (https://git.ffmpeg.org/gitweb/ffmpeg.git/commitdiff/68c05185e229b093bf5c2236c2883cbc296ea938?hp=ff17d8b56ec87fe1516ddd49b0bdea81f22904af). Instead of keeping the old code around and wrapping it in `#if LIBAVCODEC_VER_AT_LEAST(55, 63)`, I think we can assume people are using a new enough version of ffmpeg

### DIFF
--- a/src/postprocessing/pp-g722.c
+++ b/src/postprocessing/pp-g722.c
@@ -239,9 +239,7 @@ int janus_pp_g722_process(FILE *file, janus_pp_frame_packet *list, int *working)
 
 void janus_pp_g722_close(void) {
 	/* Close decoder */
-	avcodec_close(dec_ctx);
-	av_free(dec_ctx);
-	dec_ctx = NULL;
+	avcodec_free_context(&dec_ctx);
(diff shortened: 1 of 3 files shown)


lminiero:
> I think we can assume people are using a new enough version of ffmpeg

I'm afraid we can't always assume that: not all dstros jump on the latest lbraries available, and there's actually some that are quite conservatives in the updates they make. This is why I try to keep backwards compatibility when I can. 

That said, I guess that the method you mentioned has been around long enough that it should be ok to start using that one instead. I'll make some tests locally and in case I don't see issues I'll merge, thanks!

lminiero:
Merging, thanks!

PR #3593, opened 2025-10-31
AudioBridge audio limiter

m08pvv:
Worked a bit with AudioBridge and it was painful to hear all crackling noises when multiple streams summed up, so I decided to implement an audio limiter. I took AGC2 from Google's WebRTC (BSD 3-clause license that allows modifications and usage as long as license file stays in repo) and translated it to C (didn't write in C since school, so it might be not clean enough), adapted it for AudioBridge and also added SSE 4.2 and AVX2 versions (with proper initialization from available at runtime).

Also added a parameter to AudioBridge: use_limiter (boolean, default=false), which enables limiter. If limiter is disabled - only clamping (saturation to int16 boundaries) applied.

Here are some results of many participants speaking (5 participants speaking and the record is from incoming ab track):
image

Top - current Janus release with AudioBridge
Bottom - version with limiter and use_limiter = true
Full-spectrum lines - artifacts (crackling noise and clipping)

As you can see, current Janus AB implementation (just sum up all tracks) results in artifacts (values outside int16 range) whereas limiter reduces volume dynamically to normal level.

How it works:

- On ab initialization we set references to the most efficient (from available) implementation (scalar, sse 4.2, avx2)
- For each segment of audio:
- - If we mix 2 or more tracks (recording 2+ or talking 3+) - calculate scaling factors
- - If we record 2+ tracks mixed - scale and clamp, otherwise just clamp
- - For each participant:
- - - If there are 2 or more tracks besides this participant - scale and clamp, otherwise just clamp

Usage:
By default limiter is disabled and only clamping is applied.
Set `use_limiter` property in `create` audiobridge request to `true` to enable limiter.

Any suggestions are welcome!


lminiero:
@m08pvv thanks for the patch! This is an interesting effort, as I had started looking into some options a few months ago, including an attempt to dynamically load LV2 plugins as part of the AudioBridge, to try and address that too.

I'm currently abroad for a conference, so I won't be able to look at this before I get back (in a week or so). Just a couple of notes from a quick glance:

* I see you've copied code from libwebrtc. Please make sure to check the licensing implications of this, since it's not something we're linking to, but code we're embedding directly in the Janus code base.
* I don't know what the limiter code does, but I assume it's mostly a gain reduction on contributions (all of them? some?) to ensure the mix doesn't exceed the maximum sample values. If the number of active contributions plays a role in this, we should check what happens when most of those are actually unmuted but sending silence (and so would not contribute to saturation even if unmuted), and what happens when users mute/unmute changing the number of active users in the mix (is the volume of contributions in the mix changing drastically when it happens?).

m08pvv:
@lminiero thanks for a quick response! Have a good time at conference!
I'll check twice what exactly that BSD-3-clause license requires.
Speaking about limiter: coefficients are calculated once for the whole mix and then applied to different tracks (which ab sends to each participant). Yes, I can optimize it by calculating one resulting track for all silent participants (I was actually already thinking about this optimization while working on limiter and probably will implement it soon)!

m08pvv:
Speaking about license, it requires:

1.  To include the license text in sources that were adapted from WebRTC (limiter.c and limiter.h)
2.  To include the license text in documentation that comes with binaries
3.  Not to use name of Google or names of contributors from WebRTC project to endorse or promote Janus

1 - I added text in these files
2 - I added text in a section about limiter in AudioBridge plugin documentation
3 - I guess it's ok not to use name of Google in advertisements or so

m08pvv:
And about calculations - I pushed an update that calculates only once the resulting piece of audio for all silent participants (and for the record if any) because they are the same

whatisor:
It does not fix static noise for me( when someone is too loud).
After merged, I enabled in room config:
image

whatisor:
I used to use 1 simple threshold limiter for short like this when I mux multiple source:
// Tune merged audio at 0.9 MAX/MIN
#define SHRT_MAX_TUNED 29491
#define SHRT_MIN_TUNED -29491

Do you think that should be added into limiter?

m08pvv:
@whatisor could you provide a sample of audio that causes issues when mixing?
Note that currently limiter is applied only to a mix of 2 or more tracks. If you have a single track that causes issues then it seems that you're sending a broken audio to AudioBridge.

whatisor:
When you open 5 clients for audiobridge demo(better to try with 5 real persons), try to talk louder at the same time. It mostly would happen

m08pvv:
@whatisor, I'll check demo (maybe I missed some configuration and it simply doesn't turn limiter on).
On the picture at the top of this thread you can see exactly 5 loud persons (actually one person, me, but 5 tabs of an app opened). I create audiobridge via `create` request and it works for me, but the purpose of this PR is to make it work for everyone, so I'll check what could go wrong with demo.

m08pvv:
@whatisor I took a default `janus.plugin.audiobridge.jcfg.sample`, added `use_limiter = true` and launched the audiobridge demo.
Everything works fine. Could you provide settings of your audiobridge room that you used in demo? It would be also great if you switch log level to 5 and attach Janus' logs.

src/plugins/janus_audiobridge.c
```diff
@@ -8515,6 +8580,21 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
 
+	/* For audio limiter we split frame in subframes */
+	int samples_in_sub_frame = samples / K_SUB_FRAMES_IN_FRAME;
+	float *per_sample_scaling_factors = NULL;
+	float *envelope = NULL;
+	float *scaling_factors = NULL;
+	float filter_state_level = K_INITIAL_FILTER_STATE_LEVEL;
+	float last_scaling_factor = 1.f;
+	/* In case audio limiter enabled, we need some buffers for it */
+	if (audiobridge->use_limiter) {
```
lminiero: Code style: no space between `if` and bracket (I see other lines with the same nit to fix, please do that there too).

src/plugins/janus_audiobridge.c
```diff
@@ -8515,6 +8580,21 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+		JANUS_LOG(LOG_INFO, "Audio limiter enabled\n");
```
lminiero: This line should mention the room for which the limiter was enabled.

src/plugins/janus_audiobridge.c
```diff
@@ -8515,6 +8580,21 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+		per_sample_scaling_factors = g_malloc0((audiobridge->spatial_audio ? OPUS_SAMPLES*2 : OPUS_SAMPLES) * sizeof(float));
+		envelope = g_malloc0(K_SUB_FRAMES_IN_FRAME * sizeof(float));
+		scaling_factors = g_malloc0((K_SUB_FRAMES_IN_FRAME + 1) * sizeof(float));
```
lminiero: Are all these just allocated and set to 0? Doesn't it matter what value they have, considering they're passed to `compute_scaling_factors`?

src/plugins/janus_audiobridge.c
```diff
@@ -8830,15 +8919,37 @@ static void *janus_audiobridge_mixer_thread(void *data) {
 			for(index=0; index<groups_num; index++) {
 				for(i=0; i<samples; i++)
 					buffer[i] += *(groupBuffers + index*samples + i);
+				mix_count++;
+			}
+		}
+		/* If we use limiter we should initialize it if we have more than 2 tracks mixed or we have more than 1 track mixed and recording */
+		if (audiobridge->use_limiter && (mix_count > 2 || (audiobridge->recording && mix_count > 1)))
+			compute_scaling_factors(buffer, envelope, scaling_factors, per_sample_scaling_factors, 
+						  samples_in_sub_frame, &filter_state_level, &last_scaling_factor);
```
lminiero: Code style: indentation is broken here.

src/plugins/janus_audiobridge.c
```diff
@@ -8878,42 +8989,63 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+				if(audiobridge->use_limiter && (mix_count - (curBuffer ? 1 : 0)) > 1) {
+					/* If we use limiter and sumBuffer contains mix of more than 1 track, apply it */
+					scale_buffer(sumBuffer, samples, per_sample_scaling_factors, outBuffer);
+				} else {
+					/* Otherwise just clamp values that are outside int16 boundaries */
+					clamp_buffer(sumBuffer, samples, outBuffer);
+				}
+				mixBuffer = outBuffer;
+			} else if (!has_silent) {
+				JANUS_LOG(LOG_WARN, "[%s] No packet for participant (%s), but we have no silent participants either\n", audiobridge->room_id_str, p->user_id_str);
```
lminiero: Is this just a temporary debug line? This looks like something that could spam the logs.

lminiero (commented):
I did a first initial review. I'm not sure I 100% understand the logic, so I added a few comments inline.

It's also not clear to me if the limiter expects to be working with a specific sampling rate: your code doesn't seem to think it does, but I have some doubts. For RNNoise, for instance, we always upsample to 48000 if the room isn't 48000 before denoising, and then downsample to the room sampling rate after denoising, because RNNoise does expect a 48000 sampling rate: it will do _something_ for lower sampling rates, but that something will be broken (e.g., excessive denoising) because it will use more samples than it should due to the lower sampling rate. I'm wondering if the limiter works the same way, because if it does it means the same kind of upsampling would need to happen there too.

whatisor:
> @whatisor I took a default `janus.plugin.audiobridge.jcfg.sample`, added `use_limiter = true` and launched the audiobridge demo. Everything works fine. Could you provide settings of your audiobridge room that you used in demo? It would be also great if you switch log level to 5 and attach Janus' logs.

@m08pvv Thank you for your quick response. I confirm it works with denoise : false. I got issue because of **denoise: true** feature(maybe you wanna look at that in the future)
I am looking forward this PR released.

src/plugins/audiobridge-deps/limiter/limiter_scale_test.c
```diff
@@ -0,0 +1,118 @@
+#include <opus/opus.h>
```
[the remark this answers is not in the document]
m08pvv: deleted

src/plugins/janus_audiobridge.c
```diff
@@ -1240,6 +1288,7 @@ room-<unique room ID>: {
 #include "../utils.h"
 #include "../ip-utils.h"
 
+#include "audiobridge-deps/limiter/limiter.h"
```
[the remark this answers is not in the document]
m08pvv: moved next to other deps and added comment

src/plugins/janus_audiobridge.c
m08pvv: commented out (can remove, but it might be needed later for debugging)

m08pvv:
@whatisor thanks for the information! I'll definitely take a look because I'll also use `denoise` `true` in some cases.

lminiero:
I'm trying to better understand the scope for this effort, and after giving it some thought, I think that this patch may be trying to do the right thing but in the wrong place.

If what we want is to prevent saturation on the mix itself, then the patch in #3601 is probably enough, but what we want a limiter to do is "limit" the contribution of each individual participant, especially noisy ones, which probably means the limiter code should be applied to the audio we decode from participants (so _before_ the mix) rather than on the full mix itself. This would be in line with how the denoiser works, in fact: for denoising too, it's individual participants we denoise _before_ mixing, rather than denoising a mix after the fact, so that we end up with a mix that is as clean as possible. By limiting individual participants, I assume we'd have more or less normalized contrbutions, and in case their sum goes too loud, #3601 kicks in to stay within the -0.9/0.9 limit. This would also allow us to selectively choose which participants should be limited, if any, with a room default that can be used in case no property is explicitly set for participants (again, which is how denoising works too).

Does this make sense? What do you think @m08pvv @whatisor ?

m08pvv:
@lminiero in current master the resulting mix is a sum, so if we have 5 speakers (A, B, C, D, E) then:
recording is A+B+C+D+E
participant A receives B+C+D+E
...
participant E receives A+B+C+D

We assume that all tracks (A, B, C, D, E) do not exceed 100% volume (otherwise the audio is already broken).
The sum of 4 tracks will be in range [0, 400%] of volume and sum of 5 tracks will be in [0, 500%], so we need to reduce resulting tracks at least in 4 and 5 times respectively.
The main idea was that we can calculate one set of scale factors (for the whole mix, in this case it's A+B+C+D+E) and then apply it to each resulting track (A+B+C+D, ..., B+C+D+E), so the audiobridge would not introduce the x4-x5 gain if all participants speaking.

Normalizing every track individually would be N times more CPU-intensive and usually the volume of individual speakers adjusted by volume_gain (at least I thought so).

So, if calculating N set of factors and applying them to each track individually is ok (shouldn't take much CPU, but it's a hot path in audiobridge) - then I can do so, but the sum of tracks should be normalized anyway because otherwise even if we have 5 participants that are in range [0, 50%] - resulting tracks would be [0, 200%] and [0, 250%] for participants and recording respectively.

lminiero:
Yeah I get all that, but my point is that if we have 5 participants whose sum gets to 500%, we already have that other PR that brings the mix down to 90% (instead of cutting to 100% as we did before), since we're using 32-bit samples for mixing so we're not actually losing anything. That should already be equivalent to a gain reduction in each sample. I'm personally more interested in limiting individual participants that may be too loud, as what's most boresome in my experience is not the actual sum of things, but the relative dominance of some speakers over others due to their potentially different individual gains.

lminiero:
> Normalizing every track individually would be N times more CPU-intensive and usually the volume of individual speakers adjusted by volume_gain (at least I thought so).

It's true that we have the volume gain for that, but that's manual. I was thinking of something automated, which is what I assumed a limiter could bring to the table. That would pretty much be the same as audio compression (which is what I assume any limiter implements, more or less).

m08pvv:
1. In #3601 the sum is not scaled to 0.9 it's just clamped at that level, so if we have e.g. a sine wave - we will get almost a square wave (meander), which is not what we want I guess. So then it needs to also scale it somehow before or after sum.

lminiero:
> even if we have 5 participants that are in range [0, 50%] - resulting tracks would be [0, 200%] and [0, 250%] for participants and recording respectively.

I think this is where we may have a source of confusion. It's true that when you mix many contributions you can exceed the 100%, but that's exactly why we use an `int32_t` for the mixing buffer: that is to allow a sum even if it goes beyond the limits, without losing anything. When we then snap those samples back to `int16_t`, the more basic code in that other PR brings it back to max 90% no matter what it was, which is more or less equivalent to a gain reduction. I guess that a better way to do that would be to figure out much the loudest sample exceeds the mix, and apply the same division to all samples in those 20ms, but it's short enough snippets that I don't think it would matter much.

Anyway, I'll need to find some simple ways to test both patches, possibly using some automated participants in the AudioBridge that provide some talking. Once I do that I'll have a clearer picture of the whole thing.

lminiero:
> 1. In [Fix overflow mixing #3601](https://github.com/meetecho/janus-gateway/pull/3601) the sum is not scaled to 0.9 it's just clamped at that level, so if we have e.g. a sine wave - we will get almost a square wave (meander), which is not what we want I guess. So then it needs to also scale it somehow before or after sum.

Yep, makes sense, which is what I mentioned in a subsequent comment too. I'll need to make some tests.

m08pvv:
Just for visual representation of clipping:
image
If we just limit min and max values without actual scaling, we get clipping artifacts.

src/plugins/janus_audiobridge.c
m08pvv: fixed

m08pvv:
@whatisor what rnnoise version do you use? I built from f6662ad41f5bf7bf244967a04e95334c81e5af4c with 55+MB rnnoise_data.tar.gz and I have zero clipping even using default demo's 16kHz audiobridge room.

lminiero:
I still need to prepare a test with a few automated AudioBridge participants for testing purposes, but I was thinking that maybe we could achieve the same kind of result in a "simpler" way, and without the additional dependency:

1. we mix as we do know (sum to an int32 array)
2. once we have a sum of all participants, find the "loudest" sample, and see what negative gain needs to be applied to stay within the -0.9/0.9 range (if any: we may be in the range already)
3. apply that gain to the mix and save it to a different array, that we'll use for all passive participants (non-talkers, recording, forwarders)
4. for participants that are talking, as we do now, remove their contribution and do the same thing (find the loudest sample, find the gain, apply the gain to the mix for that participant)

This way we don't need to worry about how many people are talking, what's silence or not, since we always get a mix that is loud enough for everyone involved, whether they're active or not. Most importantly, we don't needlessly bring the whole mix down any time more than one person talks, since (if I understood our patch correctly) as your patch reduces the volume of active participants depending on how many there are, with 4 people you decrease each of their volumes to ~25%, which will get the mix for non-active participants to ~100%, but active speakers will get a much quieter mix (75%, since their contribution will not be a part of that). Besides, silence detection as it's done is tricky, as simple background noise will not be detected as silence but will be low enough not to contribute anything meaningful and still bring the whole volume down (people talking will suddenly be quieter if someone's unmuted but not really saying anything).

Anyway, this was just me thinking of a possible alternative solution to a problem that does exist, but that's not to say that it will be THE solution. I'll need to setup some testbed to make some tests (with different fake people speaking, as me alone wouldn't cut it), and once I have it I can prepare a tentative patch that implements my idea and I can do some comparisons.

lminiero:
@m08pvv @whatisor I prepared a first version of my proposed idea in the PR above. I did a few tests using a bunch of fake participants sending pre-recorded talking audio (plus my ugly and noisy voice) and it seems to be doing _something_: if I look at the recording, it seems to be doing its job, but of course I may be wrong. Considering both of you have made some tests with multiple participants, could you let me know if this more or less does what you need it to? Thanks!

m08pvv:
This (adapted from WebRTC) implementation is an AGC (Adaptive Gain Control), so it uses piecewise linear interpolation with pre-calculated gain curves, providing precise control over different signal levels and it includes attack/decay filtering that prevents abrupt gain changes, avoiding audible artifacts. Simple peak scaling can introduce pumping or breathing effects.
This limiter calculates per-sample scaling factors for smooth transitions, especially important during attack phases.
Yes it uses sophisticated envelope detection with lookahead to anticipate signal peaks, but it allows for more graceful gain reduction compared to reactive peak detection. It also divides audio into sub-frames for more granular control, enabling better transient preservation and reduced distortion.
The gain curves are specifically designed to minimize harmonic distortion while maintaining signal integrity, based on extensive psychoacoustic research (done by WebRTC team).

And yes, this code we already run in production and it sounds great in calls with 15 people.

whatisor:
> ce we're using 32-bit samples for

I am using https://github.com/xiph/rnnoise/releases/tag/v0.2

m08pvv:
@whatisor, I tried 0.2 (with ~21 MB model) and it sounds terrible, try latest master of rnnoise (it might fade-in beginnings of phrases but at least it doesn't produce noise)

m08pvv:
@lminiero, did you have a chance to look at the code?

lminiero:
Not yet, sorry... if not in the next few days, I'll definitely make this a priority as soon as I get back to the office after the holidays.

m08pvv:
@lminiero, any luck to look at this PR?

lminiero:
Ouch, I haven't checked it yet, apologies! I'm currently busy updating the MoQ library (new version of the draft is out), I'll 100% have a deep dive right after FOSDEM next week. If I don't feel free to shout at me :sweat_smile:

src/plugins/audiobridge-deps/limiter/limiter.h
```diff
@@ -0,0 +1,99 @@
+/* The code in this file contains code adapted from WebRTC project. */
```
lminiero: Same here.

src/plugins/janus_audiobridge.c
```diff
@@ -8830,15 +8874,37 @@ static void *janus_audiobridge_mixer_thread(void *data) {
...
+				samples_in_sub_frame, &filter_state_level, &last_scaling_factor);
+		
+		/* If we have silent participants, calculate once the full mix */
+		if(has_silent) {
+			if(audiobridge->use_limiter && mix_count > 1) {
+				/* If we use limiter and we mixed more than 1 track, apply it */
+				scale_buffer(buffer, samples, per_sample_scaling_factors, fullmix_buffer);
+			} else {
+				/* Otherwise just clamp values that are outside int16 boundaries */
+				clamp_buffer(buffer, samples, fullmix_buffer);
```
lminiero: Why is clamping done only when there's silent participants? Isn't clamping what we always do now?

lminiero (commented):
Apologies for the delay, post-FOSDEM flu knocked me out for a few days. I've done a first quick review of the code, but there's a few things I still don't understand. More details inline.

m08pvv:
Is there anything else I can do to help to complete this PR?

lminiero:
> Is there anything else I can do to help to complete this PR?

No need at the moment, thanks! The ball is in my court, so it's on me to find time to review this. I'll probably only be able to do that towards the end of the month, as we have the IETF meeting next week.

Apologies if this is taking a while, but this is a big and heavy change on the AudioBridge fundamentals. Considering we use the AB heavily in production ourselves, I can't risk merging something I'm not very familiar with, or I'm not on board 100%. I'll need time to properly and thoroughly review everything, as there were things I had doubts about.

brave44:
@m08pvv @lminiero FYI we tried this in our production environment and got very good results. 1) users with very sensitive mics and loud voices now sounds much better; 2) if they speak at the same time it sounds better too.

Overall I think this can be very good PR for audio bridge, especially for rooms where everybody can speak, not just one speaker.

m08pvv:
So, @lminiero how it's going?

lminiero:
It's Easter holidays here, I'll bring it to the rest of the team when we get back. Study of the code apart, we'll need to set it up internally for some specific testing, since as I mentioned it's a very fundamental part of our products, so I won't merge it until I'm 100% sure it won't cause issues. I'm half thinking of putting it on the demo server, and set up one room with and one without: if we set it up so that the default room (1234) uses it, we'll have data in case people using it crash it for any reason.

m08pvv:
Waiting for test results and hope they'll be ok.

lminiero:
@m08pvv the current plan is:

1. wait for #3606 to be aligned to the latest changes (whch will happen soon, I hope), stress test that, and if everything works merge (hopefully by the end of the week)
2. evaluate the changes in #3635 (low impact) and merge relatively soon (hopefully by the end of the week too)
3. once both are merged, ask you to re-align/re-base to the latest master (not only for the changes in the AudioBridge, but also because we recently merged some security fixes that I want in, before making other tests)
4. use your branch as the version we deploy on the official demo test server, so that it's what people will use for testing (one more reason why it would be important for the code to be aligned to the latest changes)

This will allow me to 1. evaluate the general stability of the code, since random people from random networks doing random things will use it, and 2. by adding different rooms (some with limiting enabled, others not) me and my team can make some more tests related to the audio quality.

Does that make sense for you as a way forward?

m08pvv:
@lminiero, good plan. Looking forward for upcoming merges!

lminiero:
@m08pvv FYI I've merged both the PRs I mentioned. When you have some time, could you align to the latest changes there? Hopefully it won't be too complicated, since I see some conflicts there. Thanks!

m08pvv:
@lminiero, will update my branch today hopefully

m08pvv:
Merged, will check for any issues after merge (just in case) a bit later

lminiero:
@m08pvv thanks! I'm out for a conference for a few days, but as soon as I'm back I'll update the demo server as we discussed (which whould also give you time to double check everything's still ok).

m08pvv:
@lminiero, I'll check today that merge didn't break anything (it was a pretty simple merge, but just in case)

m08pvv:
So, tested with 5 participants, all works fine

m08pvv:
@lminiero  I'm just curious how it's going? Is it already on your demo server? Hope everything is ok (keeping my fingers crossed🤞)

lminiero:
My apologies, it's not on the demo server yet... It's been a brutal couple of weeks, as I was to two separate conferences the past two weeks. I'll switch the demo server base to that in the next couple of days :pray:

lminiero:
I've just updated the demo server to use your branch. There are two rooms configured:

* `1234`, the default, has `use_limiter` set to `true`, but also `denoise` (as before we had used it to test the RNnoise integration). This should help verify if the two features together don't break each other.
* `4321` is a room with everything disabled instead.

The second room can be accessed passing the `?room=XYZ` query string argument demos support:

https://janus.conf.meetecho.com/demos/audiobridge?room=4321

Both rooms are configured to use 16000 as a sampling rate, as it's probably the most used. We collect core dumps when the server crashes, so that should help identify potential causes of issues.

A couple of notes from while I was upgrading. There's a warning when building:

```
plugins/janus_audiobridge.c: In function 'janus_audiobridge_mixer_thread':
plugins/janus_audiobridge.c:8654:18: warning: variable 'has_silent' set but not used [-Wunused-but-set-variable]
 8654 |         gboolean has_silent = FALSE;
      |                  ^~~~~~~~~~
```

Besides, I see the "list" request hasn't been updated to return info on whether the limiter is enabled for specific rooms. That may be helpful, as right now for instance I have no idea if how I configured the rooms actually did the trick or not. I see there are other features we may need to add to "list" (like denoise and others) so that's on me too.

You also didn't update the sample configuration file, but only the description in the Audiobridge doxygen section. The configuration sample does need to list use_limiter as an option, as in the docs, as most people only read what's written there and not the documentation (sad but true).

m08pvv:
Will check and fix all points mentioned. Not sure when (this/next week I hope)

lminiero:
@m08pvv sorry for the radio silence! I finally found some time to properly test your patch, with the help of a colleague. We focused on our own conversation and many other fake participants chiming in at different times. I have to say the results were much better than I expected: as I mentioned in previous comments, I was afraid that just unmuting could cause drastic lowerings of the volume, but we didn't experience any of that, and the audio, while chaotic (but that was normal due to the fake participant's audio that was out of context), was always quite good. The end result was even more impressive when we did the same test on the "regular" AudioBridge, where clipping and artifacts were indeed brutal. I want to make some more tests involving my other colleagues (for a "proper" test with "proper" audio), but I think it's safe to say that I definitely see this as a valid improvement, and so there's a practical plan for merging sooner rather than later!

One area where we think the current patch may be improved would be addressing overflow mixing in general, as in louder scenarios we were still getting very close to the maximum levels. I think that simply integrating the overflow mixing done in #3601 may be enough, were the cap is not INT16_MAX but about 0.9 of that. That would provide some nice limiting to stay away from saturation territory.

While we plan for more internal testing, could you address this and the other notes I mentioned in previous comments? I think what's left to be done is mainly:

1. overflow mixing limit
2. naming conventions (e.g., a `janus_audiobridge_` prefix for all new functions)
3. info on the feature in requests that return info on rooms and participants (e.g., list, or query_session for the Admin API; not sure if it makes sense for instance to return info on each partipants's state when pinging there)

More may come to mind but I think this is mostly what's missing right now before we get to a mergeable state.

Looking forward to your thoughts, and thanks again for your work (and most importantly patience :laughing: )

m08pvv:
@lminiero, thanks for all the tests and a review. I'm back from vacation and hope will find some time sooner rather than later  :wink: to fix all mentioned points.

m08pvv:
@lminiero, I pushed some of the changes (added janus_audiobridge_  prefix to functions, added info to 'list' command and it's documentation, added use_limiter option to the audiobridge sample configuration)

m08pvv:
And regarding saturation - can you attach some audio samples that cause issues? Because a simple cut-off for a complex audio signal might sound much worse than an audio that almost reaches maximum 🤷‍♂️

lminiero:
> And regarding saturation - can you attach some audio samples that cause issues? Because a simple cut-off for a complex audio signal might sound much worse than an audio that almost reaches maximum 🤷‍♂️

There's no actual issue: the whole idea was to avoid that the mix ever gets to samples with the loudest value possible, hence limiting the max value we accept to ~0.9 the actual max value (INT16_MAX). This is what some audio solutions do in order to avoid problems with playback. So I'm not proposing a hard cut-off, but a smaller max that your code should take into account when limiting. I guess it would be hard to enforce the same when the limiter is not active, though, i.e., when there's just a couple of people in the room.

m08pvv:
Basically the limiter itself limits the mix and it almost never gets to such high values (except some cases when volume changes in some non-linear way that it's hard for the algorithm to smoothly catch). If you have an example of such audio I can take a look what we can do to avoid it. Maybe we can just add some 0.9 gain if we encounter such case, but anyway have to see an example.

m08pvv:
@lminiero, let's sync what's left before the merge?

lminiero:
The current plan is to merge after the summer break: I'll go to the IETF meeting in a few days, and when I'll be back I'll only have a few days before a bit of a holiday break. When back, unless the testing on the public demo says otherwise, I think we can merge.

m08pvv:
@lminiero, have a nice holiday break! Looking forward for the merge!

m08pvv:
@lminiero, how it's going?

lminiero:
I'll probably merge next week. There's some PRs I'm looking at that I'd like to merge before tagging the next version, then I'll do a version bump and merge yours.

m08pvv:
Hope everything is ok 😅

lminiero:
It is, just a little more patience :grin:  In a matter of days I'll tag the new version.

lminiero:
Finally merging :sweat_smile:  Thanks for the support and patience! :pray:

PR #3597, opened 2025-11-05
Add trailing LF to debug traces

oleksandr-mihal-zultys:


lminiero:
Good catches, thanks! Merging :pray:

PR #3601, opened 2025-11-06
Fix overflow mixing

whatisor:
Mixing without overflow checking will cause static noise.

### DIFF
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -1276,13 +1276,13 @@ void janus_audiobridge_destroy_session(janus_plugin_session *handle, int *error)
 json_t *janus_audiobridge_query_session(janus_plugin_session *handle);
 
 /* Audio mixing with overflow protection - tuned to 0.9 of MAX/MIN to prevent clipping */
-#define SHRT_MAX_TUNED 29491
-#define SHRT_MIN_TUNED -29491
+#define JANUS_AUDIOBRIDGE_SHRT_MAX_TUNED 29491
+#define JANUS_AUDIOBRIDGE_SHRT_MIN_TUNED -29491
 
 /* Optimized inline function to clamp 16-bit audio samples */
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @whatisor! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
I think this may more easily be done in a single place, which is where we define the buffer we'll actually send, rather than in all places where we do a sum:

https://github.com/meetecho/janus-gateway/blob/master/src/plugins/janus_audiobridge.c#L8908-L8910

We use an int32 on purpose to do mixing, to allow us to go beyond, and do N-1 mixing effectively. My guess is that N-1 could be broken if the source is touched and the mix clipped.

tmatth:
FWIW opus comes with a nice implementation to this end but it's for floating point audio:
https://gitlab.xiph.org/xiph/opus/-/blob/main/include/opus.h?ref_type=heads#L800

lminiero:
> FWIW opus comes with a nice implementation to this end but it's for floating point audio:
> https://gitlab.xiph.org/xiph/opus/-/blob/main/include/opus.h?ref_type=heads#L800

Turning the samples to float and calling `opus_encode_float` instead might be an option, but that would be a transformation anyway, so it might be easier to do it ourselves with what we have.

whatisor:
@lminiero  You are correct, that is much more efficient because you already use int buffer for sum.
I think it works in my test.

lminiero (commented):
Added a few notes inline, thanks!

lminiero:
I think this looks good, in general, but is incomplete. At the moment this only smoothens the mix meant to be sent back to participants, but ignores the places where we do the same sum for the recording (L8839 in current master) and for RTP forwarders (L8969, L8994, L8999), meaning the audio could still be saturated there.

lminiero:
Closing as we merged #3593: please do let us know if you think something like this is still needed.

Issue #3562, opened 2025-07-15
[1.x] SUBSCRIBE lifecycle broken


digiboridev:
**What version of Janus is this happening on?**
1.3.2

**Have you tested a more recent version of Janus too?**
Yes

**Was this working before?**
No, tried 1.3.1 and 1.2.4

**Is there a gdb or libasan trace of the issue?**
[gist log](https://gist.github.com/digiboridev/8339ade67c597f355a1f4c3b2c899aef).

**Additional context**
I'm facing an issue with maintaining subscription. Creating a subscription works fine, but when expiration timer is near to end the Janus start doing some undocumented crap. Looks like Janus tries to renew subscription itself but fails 10 times and then it tries to cancel it using expiration 0 and also fails 10 times. Its produce a lot of problems: 
- from the SIP server perspective it blocks the real subscription termination event and answers for it and further NOTIFY events with with 481 code, so its confuses sip server.
- from the client perspective it sends fake "subscription_failed" events two times on the each try to renew subscription, this causes a race condition with real subscription termination, so i cant truly track subscription end time to start new one, only way is to wait about 2 minutes and then create a new unique subscription, but this is not a good solution.

Expected behavior: Fix autorenewal logic and add documentation for it, so we can understand how it works and what to expect from it. Or provide the way to disable it, so we can handle subscription termination by ourselves using raw NOTIFY event content.

lminiero:
> doing some undocumented crap

Please moderate your language. I will not tolerate this kind of language on here.

That said, we leverage Sofia SIP for all aspects related to SIP communication. As such, it could be an issue in the underlying library itself, or something that we're doing wrong using the APIs, but it's hard to say just from your description. Try sharing some pcap dumps of the problem as it occurs, so that I check how we handle specific incoming messages. Enabling the Sofia SIP low level debugging may also help check if it's something happening there.

digiboridev:
> Please moderate your language. I will not tolerate this kind of language on here.
> 
> That said, we leverage Sofia SIP for all aspects related to SIP communication. As such, it could be an issue in the underlying library itself, or something that we're doing wrong using the APIs, but it's hard to say just from your description. Try sharing some pcap dumps of the problem as it occurs, so that I check how we handle specific incoming messages. Enabling the Sofia SIP low level debugging may also help check if it's something happening there.

Thanks for quick answer, and sorry for my language. Will try to collect that information.

digiboridev:
[...]

PR #3602, opened 2025-11-07
fix: sip subscription renew to wrong host

digiboridev:
This is the fix for the issue https://github.com/meetecho/janus-gateway/issues/3562
Problem happened if the SIP server stays behind proxy and Janus sends subscription renew request to wrong destination(ignores proxy from dialog), this happened because NUTAG_PROXY(NULL) have such side-effect.

Here is a pcap dumps before and after [dumps.zip](https://github.com/user-attachments/files/23415176/dumps.zip)

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -3567,7 +3567,7 @@ static void *janus_sip_handler(void *data) {
 			char *contact_header = janus_sip_session_contact_header_retrieve(session);
 			/* Retrieve the outbound proxy */
 			char *proxy = session->helper && session->master ?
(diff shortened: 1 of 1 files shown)


januscla:
Thanks for your contribution, @digiboridev! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero (commented):
Thanks, I see you've pushed many pull requests which is really appreciated! I'm currently abroad for a conference, so I can't review in detail. I'll just provide a few notes, and do a deeper review when I come back home next week.

lminiero:
Thanks, merging!

PR #3610, opened 2025-11-14
Smoothen AudioBridge mix instead of clamping/truncating (see #3593 and #3601)

lminiero:
Tentative patch to try and more properly limit the mix the AudioBridge prepares, in presence of loud(er) contributions. This is an attempt to try and do something similar as #3593 (but without the external libwebrtc dependency), also taking into account the cap proposed in #3601 (in order never to clip). See https://github.com/meetecho/janus-gateway/pull/3593#issuecomment-3532008445 for details.

I tested this briefly and it seems to be doing its job, but it will need to be properly tested in all the different potential AudioBridge configurations, most importantly with stereo and/or grouping too. Recording seems fine, forwarders will need to be evaluated.

### DIFF
--- a/src/plugins/janus_audiobridge.c
+++ b/src/plugins/janus_audiobridge.c
@@ -1254,6 +1254,10 @@ room-<unique room ID>: {
 
 #define JANUS_AUDIOBRIDGE_MAX_GROUPS		5
 
+/* Audio mixing with overflow protection - tuned to 0.9 of MAX/MIN to prevent clipping */
+#define JANUS_AUDIOBRIDGE_SHRT_MAX_TUNED	29491
+#define JANUS_AUDIOBRIDGE_SHRT_MIN_TUNED	-29491
+
 /* Plugin methods */
 janus_plugin *create(void);
 int janus_audiobridge_init(janus_callbacks *callback, const char *config_path);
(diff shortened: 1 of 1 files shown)


m08pvv:
In this approach there is no smooth transition between audio segments. Each segment has it's own gain which knows nothing about other gains, so transition between segments would be very rough. In AGC previous segments affect scale factors and transition is clean and smooth.

m08pvv:
Moreover, there is only one gain value for the whole segment, so if we have a single spike inside normal segment - entire segment will be scaled.

lminiero:
Yes, it is a more simplified approach, but it's also hundreds of less lines of code and no dependency :sweat_smile:  Anything that gets merged I have to mantain, which means I need to know what the code does, which is harder to do with too many changes I have little control on. Did you have a chance to test this patch with multiple voices, or was this an analysis just by looking at the code?

m08pvv:
I didn't test this PR, just looked at the code, but I have some experience with simple approaches to audio processing (long story short, they work only in a very limited cases, but not in real life).
I can check (with my ears) this PR when have some free time.

m08pvv:
So, tried commit ab93ea0e43381b5d885f65eef44b6b253f219526 with same audio that I used in my tests (4 tabs with same audio input and 1 tab in another browser that listens)
image
We can clearly see (and hear) a lot of artifacts (clicking, crackling, etc.)

lminiero:
@m08pvv ack, thanks for testing, for its simpler approach it was worth a try :sweat_smile: 
I'm a bit busy on other activities at the moment, but I'll get back to your PR hopefully next week :pray:

src/plugins/janus_audiobridge.c
```diff
@@ -8451,6 +8455,30 @@ static void janus_audiobridge_update_wav_header(janus_audiobridge_room *audiobri
 	}
 }
 
+/* Helper to limit/clamp mixed audio */
+static inline void janus_audiobridge_limit_mix(opus_int32 *sumBuffer, opus_int16 *outBuffer, int samples) {
```
whatisor: These are overkill and not reliable. 1 random false noise could make whole mixing result worse.

PR #3611, opened 2025-11-14
Update link to multistream VideoRoom demo

fippo:
which was pointing to a renamed file


fippo:
broken since #3302 -- 1.5 years and nobody noticed or told ya :-|

lminiero:
Thanks, I missed that too! :sweat_smile:

PR #3613, opened 2025-11-26
Prevent 'use-after-free' in session helpers

oleksandr-mihal-zultys:


lminiero:
Thanks, merging both PRs! :pray:

Issue #3627, opened 2026-03-03
[1.x] crash on videoroom


spscream:
**What version of Janus is this happening on?**
we got this on fork from master(so lines of gdb dump isn't match current master).

**Have you tested a more recent version of Janus too?**
checked on fork from latest master

**Was this working before?**
no info

**Is there a gdb or libasan trace of the issue?**
(gdb) bt
#0  0x00007fbe88e17485 in janus_videoroom_hangup_media_internal (session_data=session_data@entry=0x7fbe878735c0) at plugins/janus_videoroom.c:9717
#1  0x00007fbe88e18e98 in janus_videoroom_hangup_media (handle=<optimized out>) at plugins/janus_videoroom.c:9674
#2  0x0000564d912498d0 in janus_ice_outgoing_traffic_handle (handle=0x7fbe831cdb20, pkt=0x564d912c7840 <janus_ice_hangup_peerconnection>) at ice.c:4631
#3  0x0000564d9124cd41 in janus_ice_outgoing_traffic_dispatch (source=0x7fbe85cef490, callback=<optimized out>, user_data=<optimized out>) at ice.c:528
#4  0x00007fbe8a33f255 in ?? () from /usr/lib/libglib-2.0.so.0
#5  0x00007fbe8a342387 in ?? () from /usr/lib/libglib-2.0.so.0
#6  0x00007fbe8a342c77 in g_main_loop_run () from /usr/lib/libglib-2.0.so.0
#7  0x0000564d9123e2c8 in janus_ice_handle_thread (data=0x7fbe831cdb20) at ice.c:1354
#8  0x00007fbe8a3725f0 in ?? () from /usr/lib/libglib-2.0.so.0
#9  0x00007fbe8a72e34f in start (p=0x7fbe682228e8) at src/thread/pthread_create.c:207
#10 0x00007fbe8a730965 in __clone () at src/thread/x86_64/clone.s:22

some investigation shows what ss->subscriber contains random data in memory, so it used after free:
```
(gdb) x/4s ss->subscriber
0x7fbe877587f0: "{\n   \"janus\": \"event\",\n   \"session_id\": 4487988242871326,\n   \"sender\": 3600588038572687,\n   \"plugindata\": {\n    ! \"plugin\": \""
0x7fbe8775886e: ""
0x7fbe8775886f: ""
0x7fbe87758870: ""
``` 

**Additional context**
[...]

PR #3628, opened 2026-03-03
Fix videoroom crash on janus_videoroom_hangup_media_internal

spscream:
possible fix of #3627


lminiero:
From a quick glance, the use of a mutex in `janus_videoroom_subscriber_free` is a big no. That's the callback that's invoked when the reference count gets to zero, and it's unpredictable when it will be called, especially in terms of which mutex is locked when it happens. There's a high risk of encountering random inverse order of mutex locking because of this.

spscream:
ok, I see.
Another idea is to make increment of subscriber->ref on adding to publisher streams? So subscriber won't be freed early and crash won't occure.

PR #3633, opened 2026-04-10
Change default JSON serialization to compact wherever applicable

lminiero:
As the title says, this PR changes the default serialization we do of JSON objects to strings to compact, where before it was a pretty JSON with an indentation of 3. The discussion in https://github.com/OpenSIPS/opensips/pull/3840 made me realize it made little sense for the default being a pretty indentation, considering that it's almost always meant to just be for exchanging API messages, rather than something that needs to be consumed by a human (but even in that case, prettifying it after received is always an option). Compact messages also means less data on the wire, so this should be considered an optimization.

This shouldn't impact any existing application, so I plan to merge soon. If for any reason you need indented JSON, remember it's always possible to change the default in the associated configuration file.

### DIFF
--- a/conf/janus.eventhandler.mqttevh.jcfg.sample
+++ b/conf/janus.eventhandler.mqttevh.jcfg.sample
@@ -12,7 +12,7 @@ general: {
 							# in. Valid values are none, sessions, handles, jsep, webrtc,
 							# media, plugins, transports, core, external and all. By
 							# default we subscribe to everything (all)
-	json = "indented"		# Whether the JSON messages should be indented (default),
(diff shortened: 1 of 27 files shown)


JanFellner:
Can't check the code from here (in this hell crowd of ppl in Garda land :))

For developing purposes it's always handy to have it readable in the logs. --So asking if I would make sense to have it configurable.-- did not read to the end like always :)) so forget about it. 

Defaulting to not pretty printed. On our side we do development builds using readable data and release builds compact.

lminiero:
> did not read to the end like always :)) so forget about it

No worries, enjoy Gardaland! :smile:

PR #3634, opened 2026-04-12
Memory leaks in the MQTT transport

amnonbb:
1. Receive Path Properties Leak: Cleans up the `janus_mqtt_transaction_states` hash table on terminal responses (success/error/ack) so MQTTProperties structs are properly freed.
 2. Send Path Properties Leak: Replaces `MQTTProperties_copy` with a shallow copy to prevent double-allocation (Paho handles the deep copy internally).
 3. String Duplication Leaks: Removes `g_strdup` and `g_strndup` on transaction and correlationData strings during send, as Paho duplicates them internally.
 4. Paho Queue Bug: Changes `ret = FALSE` to `ret = TRUE` in the message_arrived callback so Paho doesn't infinitely queue unacknowledged messages in RAM.
 5. Malformed JSON Leak: Adds `json_decref(root)` if the transaction string is missing to prevent leaking the JSON object.

### DIFF
--- a/src/transports/janus_mqtt.c
+++ b/src/transports/janus_mqtt.c
@@ -885,7 +885,7 @@ int janus_mqtt_send_message(janus_transport_session *transport, void *request_id
 
 			g_rw_lock_reader_unlock(&janus_mqtt_transaction_states_lock);
 
-			// If this is a terminal response (success/error/ack), we can clean up the transaction state
+			/* If this is a terminal response (success/error/ack), we can clean up the transaction state */
 			const char *janus_type = json_string_value(json_object_get(message, "janus"));
 			if(janus_type && (!strcmp(janus_type, "success") || !strcmp(janus_type, "error") || !strcmp(janus_type, "ack"))) {
 				g_rw_lock_writer_lock(&janus_mqtt_transaction_states_lock);
@@ -1154,9 +1154,7 @@ int janus_mqtt_client_message_arrived(void *context, char *topicName, int topicL
 			const gchar *transaction = json_string_value(json_object_get(root, "transaction"));
 			if(transaction == NULL) {
 				JANUS_LOG(LOG_WARN, "`transaction` is missing or not a string\n");
-				if(root) {
-					json_decref(root);
-				}
+				if(root) json_decref(root);
 				goto done;
 			}
 
@@ -1540,9 +1538,7 @@ int janus_mqtt_client_publish_message5(janus_mqtt_context *ctx, char *payload, g
 	msg.payloadlen = strlen(payload);
 	msg.qos = ctx->publish.qos;
 	msg.retained = FALSE;
-	if(properties != NULL) {
-		msg.properties = *properties;
-	}
+	if(properties != NULL) msg.properties = *properties;
 
 	char *topic;
 	if(custom_topic) {
--- a/src/transports/janus_mqtt.c
+++ b/src/transports/janus_mqtt.c
@@ -870,7 +870,7 @@ int janus_mqtt_send_message(janus_transport_session *transport, void *request_id
 	if(ctx->connect.mqtt_version == MQTTVERSION_5) {
 		char *response_topic = NULL;
 		MQTTProperties properties = MQTTProperties_initializer;
-		char *transaction = g_strdup(json_string_value(json_object_get(message, "transaction")));
+		char *transaction = json_string_value(json_object_get(message, "transaction"));
 		janus_mqtt_transaction_state *state = NULL;
 
 		if(transaction != NULL) {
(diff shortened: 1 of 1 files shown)


src/transports/janus_mqtt.c
```diff
@@ -884,6 +884,14 @@ int janus_mqtt_send_message(janus_transport_session *transport, void *request_id
 			}
 
 			g_rw_lock_reader_unlock(&janus_mqtt_transaction_states_lock);
+
+			// If this is a terminal response (success/error/ack), we can clean up the transaction state
```
lminiero: Code style: comments should be in `/* .. */` form.

src/transports/janus_mqtt.c
```diff
@@ -945,7 +953,7 @@ void janus_mqtt_proxy_properties(janus_mqtt_transaction_state *state, GArray *us
 				response_prop.identifier = MQTTPROPERTY_CODE_USER_PROPERTY;
 				response_prop.value.data.data = key;
 				response_prop.value.data.len = key_len;
-				response_prop.value.value.data = g_strndup(request_prop.value.value.data, request_prop.value.value.len);
+				response_prop.value.value.data = request_prop.value.value.data;
```
lminiero, line 956: If as you say Paho copies the string internally, then we're covered. I'm only wondering if this may have been a recent change, and if so if we need a version fence of some sort: it may have been like that for a specific reason, originally, like the request being asynchronous and needing to be sure the string was available when needed.

src/transports/janus_mqtt.c
```diff
@@ -1143,9 +1151,12 @@ int janus_mqtt_client_message_arrived(void *context, char *topicName, int topicL
 #ifdef MQTTVERSION_5
 		if(ctx->connect.mqtt_version == MQTTVERSION_5 && !admin) {
 			/* Save MQTT 5 properties copy to the state */
-			const gchar *transaction = g_strdup(json_string_value(json_object_get(root, "transaction")));
+			const gchar *transaction = json_string_value(json_object_get(root, "transaction"));
 			if(transaction == NULL) {
 				JANUS_LOG(LOG_WARN, "`transaction` is missing or not a string\n");
+				if(root) {
+					json_decref(root);
```
lminiero: Code style: since the action is simple, you don't need to wrap it in curly brackets.

src/transports/janus_mqtt.c
```diff
@@ -1529,7 +1540,9 @@ int janus_mqtt_client_publish_message5(janus_mqtt_context *ctx, char *payload, g
 	msg.payloadlen = strlen(payload);
 	msg.qos = ctx->publish.qos;
 	msg.retained = FALSE;
-	msg.properties = MQTTProperties_copy(properties);
+	if(properties != NULL) {
+		msg.properties = *properties;
```
lminiero: Same note with respect to the simple action.

lminiero (commented):
Thanks for the patch! The rationale for the changes make sense, but I added a few notes inline. Pinging @manifest too, who authored the transport plugin at the time, just to make sure those properties were not used like that for a specific reason, and if the library may indeed have changed in the meanwhile. We should also check if the MQTT event handler needs the same changes too, in case.

lminiero:
Pinging @feymartynov too as from the history I see the v5 changes were authored by him.

src/transports/janus_mqtt.c
amnonbb: The MQTTProperties struct and MQTT v5 support were first introduced in Paho MQTT C version 1.3.0 (2018)
Since the very first implementation in v1.3.0, the MQTTProperties_copy function has always called MQTTProperties_add under the hood.
MQTTProperties_add from version 1.3.0 explicitly deep-copies strings using a datadup helper function

```c 
\\ From paho.mqtt.c v1.3.0 src/MQTTProperties.c
static char* datadup(const MQTTLenString* str)
{
    char* temp = malloc(str->len);
    memcpy(temp, str->data, str->len);
    return temp;
}
// Inside MQTTProperties_add():
case MQTTPROPERTY_TYPE_UTF_8_ENCODED_STRING:
case MQTTPROPERTY_TYPE_UTF_8_STRING_PAIR:
    len = 2 + prop->value.data.len;
    props->array[props->count-1].value.data.data = datadup(&prop->value.data);
    if (type == MQTTPROPERTY_TYPE_UTF_8_STRING_PAIR)
    {
        len += 2 + prop->value.value.len;
        props->array[props->count-1].value.value.data = datadup(&prop->value.value);
    }

lminiero: Thanks for the clarification! Then, if you can confirm it works fine for you with these changes, this is good to merge for me :+1:

manifest: Sorry for the long responding. The current design is just a result of how it was first developed, rather than a deliberate choice.

lminiero: Thanks for the clarification @manifest!

manifest: By the way, feymartynov has worked with us. There wouldn’t be any objection on our side about the change. Unfortunately we also wouldn’t be able to assist with testing the feature since we stopped using the plugin and MQTT in general a few years ago.

amnonbb:
image (1)

str2 with fix, str3,4 without

lminiero:
Thanks @amnonbb, looks good to me, merging :pray:

Issue #3641, opened 2026-04-30
[1.x] Janus handles incorrectly SSRC from RTCP Receiver Report from Chrome >= 147.0.7700.0


RSATom:
Actually I'm not sure 100% is it a bug on Janus side or on Chrome side. But let me provide some details first.

I found this issue when was trying to understand  why [PiKVM](https://pikvm.org/) has issues with mic on the latest Chrome versions. PiKVM internally uses Janus to stream video/audio from PC (attached with HDMI cable) to remote browser, and also to receive audio from mic on remote browser side.

PiKVM can use following audio/video configurations:
1. Janus sends only video
2. Janus send video/audio
3. Janus send video/audio and receives audio

The problem occurs only in the latter case. After [one of the latest patches](https://webrtc.googlesource.com/src/+/d7b28ebba1bb063e00d203fe831ec435af223813) Chrome can send RR with sender SSRC of audio track, but source SSRC of video track. And this [breaks medium search logic in Janus ](https://github.com/meetecho/janus-gateway/blob/94408ffccbc1a7c39dac82d55e5dd475c807e770/src/ice.c#L3104-L3123), i.e RR for video track is marked as for audio for plugin rtcp handler (janus_plugin_rtcp.video = false). Also it happens for other RTCP packet types.

So the main question is, is it Janus or Chrome behaves wrong?

lminiero:
To be honest I'm not sure, but those SSRCs belong to different m-lines, and so different "contexts", which intuitively makes me think it may be Chrome who's wrong, but I'm not sure what the RFC says on the matter. I suspect it may have to do with the fact that audio is bidirectional, while video is monodirectional (unless the two audio streams are actually two separate m-lines? in which case they'd be monodirectional too). For monodirectional streams you only have source or sender, and IIRC in that case the one that is missing is set to 0 normally, or to whatever unique value the endpoint has assigned to the non-existing direction (which is what we do for RTCP messages we send ourselves).

In the specific example, you say:

> Chrome can send RR with sender SSRC of audio track, but source SSRC of video track

which means the confusion comes from the fact we check the sender first, which the [RFC](https://www.rfc-editor.org/rfc/rfc3550#section-6.4.2) describes as the packet sender. If we don't recognize that (e.g., SSRC=0 or unknown) we look at the source. I guess one way to try and fix that might be to revert the order, and check for source first and sender next, but who knows if that would introduce a similar issue for, e.g., SR?

[...]

PR #3642, opened 2026-04-30
Reverse the order of SSRC checks when receiving RTCP (see #3641)

lminiero:
See discussion in #3641. To test thoroughly before merging, to make sure this addresses the issue without causing regressions.


RSATom:
@lminiero did quick test - now it works as expected. Thanks!

mdevaev:
@lminiero I just want to let you know: it's been working great for us for three weeks now, and clients haven't reported any problems.

lminiero:
@mdevaev ack, thanks! I'll merge, then, and in the next few weeks we can figure out if we need better RTCP parsing procedures.

PR #3645, opened 2026-05-16
[0.x] streaming: fix extension check

tmatth:
This was breaking compilation for builds with ogg disabled.

Regression from 0b6ddadb9c4f86e07dc6b3e663d76f6f06d6383b

`master` is not impacted but v0.16.1 is.


lminiero:
Argh, ugly copy/paste error, thanks for spotting and fixing it!

PR #3646, opened 2026-05-29
core: add disable_info_api option to restrict janus=info endpoint

EricLi-400:
The janus=info request is handled before auth checks, allowing any caller to enumerate loaded plugins and server details without credentials. Add a disable_info_api config option (default: false) for operators who want to restrict this endpoint, e.g., when Janus is behind a reverse proxy and client auto-discovery is not needed.

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -156,6 +156,15 @@ general: {
 									# application in the related Janus API responses
 									# or events; in case you need them to be in the
 									# Janus API too, set this property to 'true'.
+	#disable_info_api = true		# By default, the "info" endpoint is unauthenticated,
+									# allowing any caller to discover loaded plugins and
+									# server details. Setting this to 'true' restricts
+									# the endpoint so that a valid api_secret or token
+									# is required to access it, e.g., when Janus runs
+									# behind a reverse proxy and you want to reduce
+									# server fingerprinting. Note that this option has
+									# no effect unless api_secret or token_auth is
+									# also configured.
 	#hide_dependencies = true		# By default, a call to the "info" endpoint of
 									# either the Janus or Admin API now also returns
 									# the versions of the main dependencies (e.g.,
--- a/src/janus.c
+++ b/src/janus.c
@@ -310,6 +310,10 @@ static uint candidates_timeout = DEFAULT_CANDIDATES_TIMEOUT;
 /* By default we list dependencies details, but some may prefer not to */
 static gboolean hide_dependencies = FALSE;
 
+/* By default the "info" endpoint is unauthenticated, but operators who want
+ * to reduce server fingerprinting can require auth to access it */
+static gboolean disable_info_api = FALSE;
+
 /* By default we do not exit if a shared library cannot be loaded or is missing an expected symbol */
 static gboolean exit_on_dl_error = FALSE;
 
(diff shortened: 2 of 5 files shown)


januscla:
Thanks for your contribution, @EricLi-400! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

EricLi-400:
> Thanks for your contribution, @EricLi-400! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

I have signed the CLA.

EricLi-400:
We deployed a Janus server in our production environment for our QA and test team. Within 6 days, it had created 1.2M sessions with only 19 actual plugin join messages, and the log file had grown to over 1.4 GB at the default log level (INFO).

At that point, neither token_auth nor api_secret had been configured. Eventually, Janus crashed with:

(janus:268886): GLib-ERROR **: 12:10:47.991: Creating pipes for GWakeup: Too many open files

The most likely scenario: external bots discovered the server via the unauthenticated janus=info endpoint, enumerated the loaded plugin names, then ran automated create+attach requests in a loop — likely a WebRTC scanner or generic API fuzzer that found Janus on port 8088 and replayed valid session setup sequences with no real intent to use media. The resulting file descriptor exhaustion (~5 FDs per session) eventually hit the OS limit and brought down the process.

This PR adds a disable_info_api option to address exactly this scenario. Operators who don't need client auto-discovery — for example, when Janus runs behind a reverse proxy with a known configuration — can now block the endpoint entirely, reducing the server's exposure to automated probing.

lminiero:
That looks like a useful setting to have, thanks! I assume this only disables "info" on the Janus API and not the Admin API, correct? There are legitimate reasons to have a running "info" endpoint available even when you're not trying to discover anything, but you're, e.g., simply trying to validate whether some settings were correctly enabled/disabled. As such, having a protected way of getting access to the same info would be beneficial.

EricLi-400:
sorry for the late reply.
Correct — the guard is only in janus_process_incoming_request(), so the Admin API's info endpoint is unaffected (and already sits behind admin_secret).

Good point on protected access. I've updated the implementation: instead of hard-blocking, when disable_info_api = true the handler now calls janus_request_check_secret() inline. Authenticated requests (valid api_secret or token) are served normally; unauthenticated ones get a 403. This way operators can still validate their configuration while keeping the endpoint invisible to unauthenticated probing.

EricLi-400:
Also noted that if disable_info_api=true is set but neither api_secret nor token_auth is configured, janus_request_check_secret() would pass all requests through (nothing to check). A startup LOG_WARN is now emitted in that case so the operator is alerted that the option has no effect without auth configured.

EricLi-400:
One more question — now that the behavior is auth-required rather than a hard-block, do you think disable_info_api is still a good name for this option? It could be argued that restrict_info_api is more accurate since the endpoint isn't actually disabled, just protected. Happy to rename it if you prefer, or keep disable_info_api if that fits the existing naming conventions better.

lminiero:
@EricLi-400 apologies for this super-late feedback, I've been busy on other things and I've just come back from a conference. I think the patch looks good now, but I agree that `restrict_info_api` is better as a name. As soon as you make the change I'll merge and backport to `0.x` :pray:

lminiero:
@EricLi-400 any chance you can rename the property and push the changes? As soon as you do, it's good to merge for me. Thanks!

PR #3649, opened 2026-06-22
Add screenshot support (new plugin + Streaming plugin request)

mshvaiko:
Introduce a way to grab single still images (PNG/JPEG/WebP) out of H.264 video streams, in two complementary forms:

- A new standalone Screenshot plugin (janus.plugin.screenshot) that can capture either from a WebRTC PeerConnection negotiated directly with it, or from plain H.264 RTP forwarded to one of its configured listeners. Each listener is a numeric-id "stream" bound to its own videoport, and the "take" request selects the source by id (recordplay-style).

- A "screenshot" request in the Streaming plugin that taps the keyframe it already buffers for new viewers, so a mountpoint can be screenshotted by id with no extra ingest or RTP forwarding, mirroring how "recording" works. Guarded by JANUS_STREAMING_SCREENSHOT and the screenshot plugin's FFmpeg dependency, so it degrades gracefully when built without it.

Both paths share the same H.264 keyframe reassembly and libavcodec/ libswscale decode+encode logic. Includes Doxygen docs for both plugins and the build wiring (configure.ac, Makefile.am) for the FFmpeg dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ cov-int
 .vscode
 node_modules
 .venv
+
+# Local development run directory (configs, recordings, screenshots)
+dev-run/
--- a/conf/janus.plugin.screenshot.jcfg.sample
+++ b/conf/janus.plugin.screenshot.jcfg.sample
@@ -1,0 +1,24 @@
+general: {
+	events = true
+	path = "./screenshots"
+}
+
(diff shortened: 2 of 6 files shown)


januscla:
Thanks for your contribution, @mshvaiko! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

PR #3651, opened 2026-06-26
Add optional per-handle ICE port range (`min_port` / `max_port` on `attach`)

Odinvt:
## Summary

Two optional `attach` parameters, `min_port` and `max_port`, that confine a handle's ICE host-candidate UDP ports to the given range. When absent, behavior is byte-identical to today and the handle uses the global static config `rtp_port_range`. Core-only, plugin-agnostic, opt-in, and in the spirit of the existing optional per-handle attributes `opaque_id` and
`loop_index`.

```
attach { "plugin": "...", "opaque_id": "...", "loop_index": N,
         "min_port": 20000, "max_port": 29999 }   // min_port/max_port optional
```

## Motivation

The global `rtp_port_range` sets the total UDP surface Janus may use, but gives no control over which connection lands where. Every PeerConnection draws from one pool with no per-connection input, yet the range a connection should use is often a function of what it is (its media, tenant, room, or priority). That is runtime knowledge the application has at
`attach`, not config-time knowledge in `janus.jcfg`.

A per-handle optional override makes port placement a per-connection decision, with no restart and for any plugin, since it is just a property of the handle's ICE agent.

Our motivating case is QoS by port. On a managed network, media is classified by UDP port at the network edge to apply DiffServ/DSCP, because endpoint-set DSCP is untrusted and bleached across administrative boundaries.

This needs media on deterministic, classifiable port bands, which a per-handle range provides and a single global pool cannot.

The same primitive fits other needs, such as isolating each tenant's or room's media in its own band, segmenting firewalls and ACLs into narrow predictable sub-ranges, or steering a subset of connections onto canary or A/B network paths.

I may well have missed something here. Feedback and corrections are very welcome, Thank you.

### DIFF
--- a/src/janus.c
+++ b/src/janus.c
@@ -311,6 +311,9 @@ static uint candidates_timeout = DEFAULT_CANDIDATES_TIMEOUT;
 
 /* By default we list dependencies details, but some may prefer not to */
 static gboolean hide_dependencies = FALSE;
+/* Cached config: the configured RTP/RTCP port range, and whether handles may request a sub-range on attach */
+static uint16_t rtp_port_range_min = 0, rtp_port_range_max = 0;
+static gboolean rtp_port_range_perhandle_override = FALSE;
 
 /* By default we do not exit if a shared library cannot be loaded or is missing an expected symbol */
 static gboolean exit_on_dl_error = FALSE;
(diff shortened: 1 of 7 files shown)


januscla:
Thanks for your contribution, @Odinvt! Please make sure you sign our [CLA](https://cla.conf.meetecho.com/janus), as it's a required step before we can merge this.

lminiero:
While this does look like an interesting feature, I think it's risky as it is, as it's basically a way to override the main range an administrator may have configured, possibly with good reason. I think the user-provided min/max should be confined to the main range (you shouldn't be able to go beyond that), and there should also be a configuration property to specify whether it should be possible to override it in the first place (default being false, to be backwards compatible with how Janus works right now).

Odinvt:
Thanks, that makes sense. Updated:

- Added a media option `rtp_port_range_perhandle_override`, default `false`, so behaviour is unchanged unless an admin opts in.
- When enabled, the requested `min_port`/`max_port` is confined to the configured `rtp_port_range`. A range outside it (or one sent while the option is disabled) is ignored and the global range is used instead. I chose to ignore & warn rather than return an error to avoid complicating the change.

I'm not sure this is quite the approach you had in mind, happy to adjust.

src/janus.c
```diff
@@ -309,6 +311,9 @@ static uint candidates_timeout = DEFAULT_CANDIDATES_TIMEOUT;
 
 /* By default we list dependencies details, but some may prefer not to */
 static gboolean hide_dependencies = FALSE;
+/* Cached config: the configured RTP/RTCP port range, and whether handles may request a sub-range on attach */
+static uint16_t rtp_port_range_min = 0, rtp_port_range_max = 0;
```
lminiero: I don't think you need to cache them here. They're saved in `ice.c` as static values already, so it's probably enough to simply add some getters for them and expose them in `ice.h`, as we do for other properties stored there that `janus.c` needs to access (e.g., for Admin API info purposes). If they're not set they're both 0.

One thing I noticed, though, is that if the range is invalid, we're not resetting the value back to 0, but only showing a warning. That's something we probably need to fix.

Odinvt:
Thanks, both addressed : 

- The range is now read through exposed getters on the existing `ice.c` statics instead of being cached in `janus.c`
- An invalid per-handle range now resets the handle values back to 0 before it falls back to the global range

Happy to adjust anything else.

lminiero:
Thanks! I'll do a deeper review and a couple of tests in the next few days.

lminiero:
Just tested and it seems to be working nicely, thanks! I'll merge and backport to `0.x` :pray:

PR #3652, opened 2026-06-27
rtcp: validate SR/RR report blocks against packet length

tmatth:
Use the current RTCP packet's declared length when validating SR/RR report block counts, instead of the remaining compound-buffer length. This prevents malformed compound RTCP packets from letting rb[1] access cross the current packet boundary.

### DIFF
--- a/src/rtcp.c
+++ b/src/rtcp.c
@@ -454,12 +454,21 @@ gboolean janus_rtcp_check_len(janus_rtcp_header *rtcp, int len) {
 }
 
 gboolean janus_rtcp_check_sr(janus_rtcp_header *rtcp, int len) {
+	int packet_len = 4*(int)ntohs(rtcp->length) + 4;
 	if (len < (int)sizeof(janus_rtcp_header) + (int)sizeof(uint32_t) + (int)sizeof(sender_info)) {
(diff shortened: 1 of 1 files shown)


tmatth:
This would need to be backported to 0.x. I don't think it's a security concern since the behavior is bounded to the next packet, more of a parsing correctness thing.

lminiero:
Thanks! Did you notice some issues due to the existing code?

tmatth:
> Thanks! Did you notice some issues due to the existing code?

No it just came out of static analysis, in practice it would be tricky to do anything with this (malformed rtcp that at worst borks the next packet)

lminiero:
Makes sense! Not sure if we should also check packet_len vs len, though? If packet_len is the "law", now, we want to make sure that it's not more than we actually received, in order to avoid overflows.

tmatth:
> Makes sense! Not sure if we should also check packet_len vs len, though? If packet_len is the "law", now, we want to make sure that it's not more than we actually received, in order to avoid overflows.

Yeah at current call sites `janus_rtcp_check_len` already guarantees `packet_len <= len`, but since these are public functions I'll add the check so they don't depend on that calling convention.

lminiero:
Thanks, merging then! :pray:

PR #3655, opened 2026-07-07
Added option to disable manual injection of SPS/PPS for RTSP servers that put them in the SDP (see #3168)

lminiero:
In #3168 we added a way for the Janus Streaming plugin to manually inject SPS/PPS NALs in the RTP stream, in case RTSP servers advertised them in the SDP via fmtp attributes. This helped address scenarios where the RTSP server would _only_ put SPS/PPS in a `sprop-parameter-sets`: WebRTC endoints never look at that, and expect them to be part of the RTP stream, so this manual injection ensured video could be decoded.

This seems to not always work as expected with RTSP cameras that do advertise SPS/PPS via the SDP, but then _also_ send SPS/PPS in the RTP stream. The end result is that, on the WebRTC side, SPS/PPS are duplicated, since there is the one sent by the camera itself, and then the one we add ourselves, which apparently confuses browsers and causes broken video.

This PR is a first attempt to fix this. It adds a new configuration property you can add when creating an RTSP mountpoint (statically or dynamically) called `rtsp_fmtp_sps`: if the value of that property is `"always"` (default), then it behaves as it does now, and so if it finds SPS/PPS data in the SDP, it will inject it manually in RTP; if the value is `"never"`, instead, we ignore fmtp data sent in the RTSP SDP.

The property is a string and not a boolean as I want to keep doors open for more flexibility. In the future, for instance, it might be helpful to have something like an `"auto"` setting, where the plugin stores the SPS/PPS found in the fmtp, but only uses it if it never sees an SPS on the wire. This would allow the code to handle more dynamically scenarios where you don't know in advance what a camera will do. That said, this automatic mode would require more code than I wanted to add now: the existing setting should already help address the main issue in scenarios where the behaviour of the cameras is known.

I haven't tested this yet, so feedback is welcome. In case it works fine and with no regressions, I plan to merge soon and backport to `0.x`.

### DIFF
--- a/conf/janus.plugin.streaming.jcfg.sample.in
+++ b/conf/janus.plugin.streaming.jcfg.sample.in
@@ -150,6 +150,9 @@
 # rtsp_conn_timeout = connection timeout for cURL (CURLOPT_CONNECTTIMEOUT) call gathering the RTSP information (default=5s)
 # rtsp_notify_changes = if set to true, will send an event to connected users when the RTSP session
 #		gets disconnected, and when it's reconnected (default=false)
+# rtsp_fmtp_sps = what the plugin should do it if finds SPS/PPS info in an SDP fmtp attribute sent by the RTSP server;
+#		the default value is "always", to manually inject them in RTP traffic, as that's what WebRTC endpoints need;
+#		to disable this behaviour (e.g., if the camera already sends SPS/PPS in-stream via RTP too) use "never"
 #
 # Notice that, for 'rtsp' mountpoints, normally the plugin uses the exact
 # SDP codec and fmtp attributes the remote camera or RTSP server sent.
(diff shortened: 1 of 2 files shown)


lminiero:
Merging.

groupboard:
Note that this rtsp_fmtp_sps change breaks the build if HAVE_LIBCURL isn't defined.

plugins/janus_streaming.c: In function ‘janus_streaming_parse_sprop’:
plugins/janus_streaming.c:1707:48: error: ‘janus_streaming_rtp_source’ {aka ‘struct janus_streaming_rtp_source’} has no member named ‘rtsp_fmtp_sps’
 1707 |     if(!source->rtsp || (source->rtsp && source->rtsp_fmtp_sps)) {
 
 Looks like it's just that one line that needs fixed, by adding an ifdef around the ||:
 
```
    if(!source->rtsp
#ifdef HAVE_LIBCURL
                                                || (source->rtsp && source->rtsp_fmtp_sps)
#endif
                                                ) {
```
                                    
 Compiles fine after making that change.

lminiero:
@groupboard mh that's weird, we do have CI recipes on Github without libcurl too, and they never complained about that. Is it on master or `0.x`?

lminiero:
Ok, verified the issue was only in `0.x` (which is why the CI didn't flag it, it only targets `master`), should be fixed now. Thanks for the heads up!

groupboard:
Yeah, 0.x. Thanks.

Issue #3648, opened 2026-06-11
[1.x] Enhancement: Support modern hashing algorithms (SHA-256) for authentication


fggreeff:
**What version of Janus is this happening on?**
V1.4.1 (latest version)

**Have you tested a more recent version of Janus too?**
N/A - This is an architectural enhancement request regarding the cryptographic algorithms supported in the authentication layer, rather than a regression or a bug

**Was this working before?**
N/A - SHA-1 has been the default implementation for this authentication method

**Is there a gdb or libasan trace of the issue?**
N/A - No crash or memory issue is occurring

**Additional context**
We would like to propose an architectural enhancement to Janus's authentication mechanisms to support modern cryptographic standards and phase out the reliance on SHA-1. 

While SHA-1 has long been deprecated for digital signatures due to collision vulnerabilities, security standards are continuing to tighten across the industry:
1. **Modern Baseline:** Secure production environments widely mandate a baseline of SHA-2 (such as SHA-256) or SHA-3 (256-bits or higher).
2. **Post-Quantum Readiness:** The open-source ecosystem is actively shifting toward post-quantum cryptography (PQC), with a growing focus on algorithms like ML-DSA (FIPS 204) for long-term signatures.

Currently, this specific authentication pipeline relies on SHA-1. Moving forward, we recommend updating the code to:
* Support at least SHA-256 as a secure, modern fallback or baseline.
* If backward compatibility with older deployments is required, introduce a configuration option/flag in the configuration file to allow administrators to enforce a minimum secure hash algorithm (e.g., rejecting SHA-1).

Moving away from SHA-1 ensures Janus remains secure by default, aligns with modern internet security best practices, and prevents integration blockers in hardened environments where legacy SHA-1 libraries are entirely disabled at the OS level

lminiero:
I thought I had answered this already, but apparently I didn't...

Not sure what this issue is about. We don't use SHA-1 for anything. There's ancient references to SHA-1 in the SDP code, that's only there for backwards compatibility, and has probably not been used in 15 years.

This feels like some text some AI spewed out without even looking at the code. Please provide more info and/or context or I'll just close this.

mcadecio:
I believe [fggreeff](https://github.com/fggreeff) is referring to the Signed-Token based authentication.

In the auth.c, the methods janus_auth_check_signature and janus_auth_check_signature_contains contains a snippet of code that generates a HMAC signature to verify the provided signature matches the signature in the token.

```
[...]

PR #3658, opened 2026-07-13
Adding support for HMAC-SHA256 for signed_tokens (see #3648)

mcadecio:
* Users can configure which hash function is used for signed_tokens signature verification by setting `token_auth_hash` to `sha1` or `sha256`
* If token_auth_hash is not in the config, sha1 will be used to preserve backwards compatibility

### DIFF
--- a/conf/janus.jcfg.sample.in
+++ b/conf/janus.jcfg.sample.in
@@ -63,10 +63,12 @@ general: {
 									# a valid token in all requests. Useful if
 									# you want to authenticate requests from web
 									# users.
-	#token_auth_secret = "janus"	# Use HMAC-SHA1 signed tokens (with token_auth). Note that
+	#token_auth_secret = "janus"	# Use HMAC signed tokens (with token_auth). Note that
 									# without this, the Admin API MUST
 									# be enabled, as tokens are added and removed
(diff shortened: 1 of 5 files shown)


lminiero:
Thanks! From a quick glance it looks ok, even though I think the `exit(1)` is a bit abrupt: it may need a `janus_options_destroy();` before that, as we do in a few other places.

I'm also wondering if it may make sense to provide an example in `mainpage.dox` for sha256 too, but if changing the hash function in the existing function is all that's needed to get it to work, then probably not (I assume you made some tests about that?)

mcadecio:
Thank you for the swift review! You are right, I have added the `janus_options_destroy()` before exit.
Yes, I did think about updating the JavaScript example, but the change itself is so minuscule I didn't feel compelled to update the example. All that would change is:

`crypto.createHmac('sha1', secret);`
to
`crypto.createHmac('sha256', secret);`

Regarding tests, I tested Janus is able to verify sha1 and sha256 signatures by generating a token using the javascript example code.

lminiero:
Makes sense, thanks! I'll merge, then, and backport to `0.x` :pray: 
Thanks again for contributing this!

PR #3660, opened 2026-07-27
Add support for outgoing late offers (offerless INVITE) to the SIP plugin

lminiero:
While we supported late offers (offerless INVITEs) in the SIP plugin, we only allowed them on the way in, that is when handling incoming calls from the SIP side and targetet to the WebRTC user. This pull request now adds support to the other direction as well, which means the WebRTC user can now originate an offerless INVITE via the SIP plugin.

To use the feature, the way you use the plugin changes slightly, as you need to:

1. not provide a JSEP SDP, but add a `late_offer: true` property to the `call` request;
2. be prepared, in the `accepted` event, to handle an SDP _offer_, rather than an _answer_ (as would normally be the case)
3. when a JSEP SDP answer has been prepared, send it back using the new `late_ack` request.

I tested this using SIPp as a backend and it seems to be working as expected, but of course more feedback would help, so in case you're interested in the feature, please do give it a try and let us know. Notice that, while this is based on the `master` branch, I'll backport it to `0.x` as well when merged.

### DIFF
--- a/html/demos/sip.js
+++ b/html/demos/sip.js
@@ -311,36 +311,7 @@ $(document).ready(function() {
 											Janus.log(result["username"] + " accepted the call!", jsep);
 											// Call can start, now: handle the remote answer
 											if(jsep) {
-												let doAudio = (jsep.sdp.indexOf("m=audio ") > -1);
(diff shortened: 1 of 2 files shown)


lminiero:
Merging.

Issue #3663, opened 2026-09-01
[1.x] double free in audiobridge participant thread


spscream:
**What version of Janus is this happening on?**
1.4.2, our branch based on 4602fcc5315b77dce2a5d8363a4286ac672183b1 commit. we have some changes in audiobridge, but this part of participant thread and audiobridge-deps/jitter.c are identical to master.

**Have you tested a more recent version of Janus too?**
no, but code in master is the same

**Was this working before?**
no info

**Is there a gdb or libasan trace of the issue?**

```
(gdb) bt
  #0  get_meta (p=p@entry=0x75d591c7e750 "\220\322M\251\325u") at src/malloc/mallocng/meta.h:141
  #1  0x000075d5b85aab1f in __libc_free (p=0x75d591c7e750) at src/malloc/mallocng/free.c:105
  #2  0x000075d5b6bfc85b in janus_audiobridge_buffer_packet_destroy (pkt=<optimized out>) at plugins/janus_audiobridge.c:1841
  #3  janus_audiobridge_participant_thread (data=0x75d59d9d3d40) at plugins/janus_audiobridge.c:9352
  #4  0x000075d5b81f5426 in ?? () from /usr/lib/libglib-2.0.so.0
  #5  0x000075d5b85cb972 in start (p=0x75d5927a6888) at src/thread/pthread_create.c:207
  #6  0x000075d5b85cd2b4 in __clone () at src/thread/x86_64/clone.s:22
  Backtrace stopped: frame did not save the PC
```

we are on musl, so free() crashes on broken metadata of already freed chunk.

gdb shows line 9352, but compiler merged two identical blocks into one tail (the second one jumps into it), and by data from core the real path is the plc branch, in master it is here:
https://github.com/meetecho/janus-gateway/blob/4602fcc5315b77dce2a5d8363a4286ac672183b1/src/plugins/janus_audiobridge.c#L9233

state in core:

```
jbp = {data = 0x75d591c7e750, len = 0, timestamp = 0, span = 0, sequence = 0, user_data = 0}
jitter->reset_state = 1, all jitter->packets[i].data == NULL, jitter->lost_count = 0
lost_packets_gap = 14
participant->decoder = 0x0, participant->encoder = 0x0, participant->room = 0x0,
participant->muted = 1, participant->display = 0x0, participant->active = 0
```

[...]

PR #3664, opened 2026-09-01
Fix double free of buffered packet in audiobridge participant thread

spscream:
possible fix of #3663


lminiero:
I think this was a good catch, especially the extra `janus_audiobridge_buffer_packet_destroy` call where it wasn't supposed to be. Looks good to merge for me, I'll backport to `0.x` too. Thanks!

PR #3665, opened 2026-09-01
Assign `handle->static_event_loop` when a handle is placed by loop index (fixes #3661)

### LINKED ISSUE #3661: [1.4.1] allow_loop_indication: handles placed by loop index are never removed from the loop's accounting
> **Provenance:** this bug was found by [Claude Code](https://claude.com/claude-code) running Claude Opus 5, during a source review of `allow_loop_indication`. The analysis below is its work; I have checked the code references and I am filing it under my own name, so treat me as responsible for it. Flagging the origin because it is a static-analysis finding with no runtime reproduction behind it, and you should weigh it accordingly.

v1.4.1, commit `85304a7a4bdb2719583ad53ae04130a8273dba6a`.

Yes — the code is unchanged on master as of `4602fcc5315b77dce2a5d8363a4286ac672183b1`, and the line references below are to master. To be clear about what kind of report this is: it comes from reading the source, not from a reproduction. We run static event loops with `allow_loop_indication` left off, and this came out of reviewing the option before enabling it — so we have not exercised the manual path in production, and on the strength of this we decided not to.

**Was this working before?**

Not bisected. The manual-index branch does not appear ever to have assigned `handle->static_event_loop`, so this looks like it dates to the introduction of `allow_loop_indication` rather than being a regression.

No. The symptom is silent accounting drift plus a refcount that is never released, not a crash, so there is nothing for a trace to catch. The argument is entirely in the source quoted below.

**Additional context**

In `janus_ice_handle_attach_plugin` (`src/ice.c`), with `static_event_loops > 0`, a handle can get its loop two ways.

The automatic path sets the back-pointer (`src/ice.c:1475-1480`):

```c
janus_refcount_increase(&loop->ref);
loop->handles++;
handle->mainctx = loop->mainctx;
handle->mainloop = loop->mainloop;
handle->static_event_loop = loop;
```

The manual path — taken when `allow_loop_indication` is on and the API supplies a valid `loop_index` — does everything except that (`src/ice.c:1450-1454`):

```c
janus_refcount_increase(&loop->ref);
automatic_selection = FALSE;
handle->mainctx = loop->mainctx;
handle->mainloop = loop->mainloop;
loop->handles++;
/* handle->static_event_loop is never assigned */
```

`janus_ice_handle_destroy` gates the cleanup on exactly that pointer (`src/ice.c:1533-1538`):

```c
if(handle->static_event_loop != NULL) {
    janus_ice_static_event_loop *loop = (janus_ice_static_event_loop *)handle->static_event_loop;
    loop->handles--;
    janus_refcount_decrease(&loop->ref);
    JANUS_LOG(LOG_VERB, "[%"SCNu64"] Manually removed handle from loop #%d\n", handle->handle_id, loop->id);
}
```

So for every handle placed by loop index, two things never happen:

- **`loop->handles` is incremented and never decremented.** This is not confined to the manual path: that counter is what the automatic branch scans to pick the least-loaded loop, so the skew degrades placement for handles that never touched the manual API. It is also a `uint16_t`, so it wraps at 65536. `loops_info` reports the inflated values as though they were real, which makes the drift hard to notice.
- **The loop's refcount is increased and never released,** so the loop object cannot be freed.

Both are permanent for the lifetime of the process.

Incidentally, the log line inside that `if` reads "Manually removed handle from loop #%d" — the one case it can never describe, since a manually-added handle is exactly what fails the condition. That reads like the assignment was intended to be there.

Assigning `handle->static_event_loop = loop;` in the manual branch, next to the existing `loop->handles++`, would make the two branches symmetric and looks sufficient. Note that the extra `janus_refcount_increase(&handle->ref)` taken earlier for the static-loop case is *not* affected — it is released in `janus_ice_outgoing_traffic_finalize` under a global `static_event_loops > 0` check (`src/ice.c:542-545`), so it balances either way.

Happy to open a PR if that is the fix you would want.

artmatsak:
Fixes #3661.

**Provenance:** as with the issue, the analysis behind this patch is [Claude Code](https://claude.com/claude-code)'s work; I have checked the code references and I am submitting it under my own name, so treat me as responsible for it. It is a static-analysis fix — see the verification note at the bottom for exactly what was and was not exercised.

With `static_event_loops > 0`, `janus_ice_handle_attach_plugin` can place a handle on a loop two ways. The automatic branch sets the back-pointer; the manual branch — taken when `allow_loop_indication` is on and the API supplies a valid `loop_index` — did everything except that. Since `janus_ice_handle_destroy` gates its cleanup on exactly that pointer, every manually-placed handle left `loop->handles` incremented forever and never released the loop's refcount.

The skew is not confined to the manual path: `loop->handles` is what the automatic branch scans to pick the least-loaded loop, so handles that never touched the manual API get placed against inflated counts. It is a `uint16_t`, so it wraps at 65536, and `loops_info` reports the inflated values as though they were real.

This patch assigns `handle->static_event_loop = loop;` in the manual branch and reorders the two adjacent lines so both branches read identically. The extra `janus_refcount_increase(&handle->ref)` taken earlier for the static-loop case is unaffected — it is released in `janus_ice_outgoing_traffic_finalize` under a global `static_event_loops > 0` check, so it balanced either way.

It also drops one word from the log line in `janus_ice_handle_destroy`, which read "Manually removed handle from loop #%d" — the one case it could never describe, since before this patch a manually-added handle was exactly what failed the condition. It now covers both kinds of placement, so "Removed handle from loop #%d" is what it should say. The attach-side "Manually added" / "Automatically added" messages are correct as they are and are left alone.

If you would rather remove the possibility of the two branches drifting again, the shared work (`janus_refcount_increase`, `loop->handles++`, `mainctx`, `mainloop`, `static_event_loop`) could be hoisted below the `if`/`else` with each branch only selecting the loop. I kept the diff minimal instead, but happy to rework it that way if you prefer.

**Verification:** the argument above is from reading the source. We run static event loops with `allow_loop_indication` off, so I have not exercised the manual path in production, and there is no runtime reproduction behind this. Compilation is covered by CI.

### DIFF
--- a/src/ice.c
+++ b/src/ice.c
@@ -1449,9 +1449,10 @@ gint janus_ice_handle_attach_plugin(void *core_session, janus_ice_handle *handle
 			} else {
 				janus_refcount_increase(&loop->ref);
 				automatic_selection = FALSE;
+				loop->handles++;
 				handle->mainctx = loop->mainctx;
 				handle->mainloop = loop->mainloop;
-				loop->handles++;
+				handle->static_event_loop = loop;
 				JANUS_LOG(LOG_VERB, "[%"SCNu64"] Manually added handle to loop #%d\n", handle->handle_id, loop->id);
 			}
 		}
@@ -1534,7 +1535,7 @@ gint janus_ice_handle_destroy(void *core_session, janus_ice_handle *handle) {
 		janus_ice_static_event_loop *loop = (janus_ice_static_event_loop *)handle->static_event_loop;
 		loop->handles--;
 		janus_refcount_decrease(&loop->ref);
-		JANUS_LOG(LOG_VERB, "[%"SCNu64"] Manually removed handle from loop #%d\n", handle->handle_id, loop->id);
+		JANUS_LOG(LOG_VERB, "[%"SCNu64"] Removed handle from loop #%d\n", handle->handle_id, loop->id);
 	}
 	janus_mutex_unlock(&event_loops_mutex);
 	janus_plugin *plugin_t = (janus_plugin *)handle->app;


lminiero:
Thanks, merging! I'll backport to `0.x` too, if needed.

PR #3669, opened 2026-09-04
SIP plugin: release the nua handles and references the application owns

alvm:
Eight fixes for one recurring mistake: sofia's nua layer hands the application
an object whose release is the application's duty, and `janus_sip` never
performs it. None of it is a leak in the LeakSanitizer sense — every byte stays
*reachable*, handles in nua's handle list, records in a GHashTable, ETags in a
handle's `su_home` — so ASan, LSan, valgrind and heaptrack all report zero. What
finds this class is reference accounting (bpftrace uprobes on the refcount API)
and black-box A/B measurement.

| commit | what it costs unfixed |
|---|---|
| Destroy the nua handle after auto-answered OPTIONS, MESSAGE and INFO | **2.3 kB per request on musl, 2.9 kB on glibc**, retained for the life of the session — an instance answering pings for a few dozen registered bindings loses tens of MB per hour, one answering none loses nothing |
| Destroy the handle and return the call reference on INVITE rejects | ~1.4 kB per rejected INVITE, plus a session pinned for good |
| Release the saved REFER event and orphaned handle on transfer teardown | two sofia objects per transfer |
| Sweep all of a closing session's transfers from the global table | the entry, its session reference — **and a reproducible abort at plugin unload** |
| Do not record a transfer for a session that is already being torn down | the same abort, reached by a race |
| Do not leak a json_string reference per forwarded incoming header | ~100 B per forwarded header, on every event that forwards them |
| Destroy the call handle when the call terminates | **one handle per call, unbounded** |
| Release the subscription handle once its dialog usage is gone | 1.5–3 kB per subscribe/unsubscribe cycle |

Each commit message carries the mechanism, the sofia contract it follows and how
it was measured, so they are worth reading individually rather than as a block.

**Three notes a reviewer will want up front.**

The guard in the first commit is subtle and is shared by five of the eight
through one helper. sofia's documentation suggests "destroy if the handle is not
bound" (`hmagic == NULL`), which is almost right: the master session adopts the
incoming-call handle *without* `nua_handle_bind`, so `s_nh_i` is unbound as
well and must be spared, or an in-dialog INFO or MESSAGE during an incoming call
would destroy a live call.

The teardown race fix relies on an ordering that is worth confirming: the sweep
and the `destroyed` flag are set inside the same `sessions_mutex` section, since
`sessions` is created with `janus_sip_session_destroy` as its value destructor
and the `g_hash_table_remove(sessions, handle)` at the end of that region is
what sets the flag. Sweep first, flag second, one lock — so checking the flag
under the same mutex leaves no window.

The call handle can only be destroyed in `nua_i_terminated`, not in
`i_state(terminated)`: `nua_handle_destroy()` invalidates the handle at once and
nua then drops every event still queued for it, including the one that returns
the active-call reference the session teardown waits for.

**Evidence, and where it comes from.** These fixes have run in production for
weeks — but on a **downstream fork** of this plugin, not on master, and that
deployment had other leaks running at the same time, including ones in
presence/PUBLISH code that does not exist upstream. The end-to-end numbers
below are therefore not attributable to this series alone and should not be
read as master's rate:

- One instance had the image rolled back by an unrelated deployment, giving a
  before/after on identical traffic: **+0.30 MB/h with these fixes over 217
  hours and no restart, against +48 to +53 MB/h and an OOM kill roughly every
  60 hours without them.**
- A second instance, on SIP over TCP, went from +6.0 MB/h to +0.27 MB/h at an
  unchanged subscriber count, and its CPU fell from 582 m to 11 m — on a
  leaking instance much of the CPU goes on allocation and page faults inside
  the oversized process rather than on load.

**What does carry over, and why.** The retention is sofia's, and the OPTIONS
handler this series touches is byte-identical between that fork and master —
verified by diff, not assumed. Per-path measurements that are attributable:
898 handles created and 0 destroyed over 180 s on an affected instance, and one
handle retained per call, which the stand shows as a population of 4 for 4
accounts after ten calls with the fix against fourteen without it.

**Checked against master itself.** The twelve commits build on `4602fcc5` on a
clean Debian Trixie against **unpatched sofia v1.13.18** with zero compiler
warnings, and on that build: ten calls with real ICE/DTLS/SRTP leave RSS flat
across two minutes of quiet, forty cycles of hang-up-mid-ring and every reject
code produce no abort, and the handle census reads 4 for 4 accounts.

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1578,6 +1578,12 @@ static void janus_sip_message_free(janus_sip_message *msg) {
 	g_free(msg);
 }
 
+/* Matcher for sweeping a closing session's transfers out of the global table */
+static gboolean janus_sip_transfer_owned_by_session(gpointer key, gpointer value, gpointer user_data) {
+	janus_sip_transfer *t = (janus_sip_transfer *)value;
+	return t != NULL && t->session == (janus_sip_session *)user_data;
+}
+
 static void janus_sip_transfer_destroy(janus_sip_transfer *t) {
 	if(t == NULL)
 		return;
@@ -2776,11 +2782,12 @@ void janus_sip_destroy_session(janus_plugin_session *handle, int *error) {
 			janus_mutex_unlock(&master->mutex);
 		}
 	}
-	/* If this session was involved in a transfer, get rid of the reference */
-	if(session->refer_id) {
-		g_hash_table_remove(transfers, GUINT_TO_POINTER(session->refer_id));
-		session->refer_id = 0;
-	}
+	/* Every entry holds a reference to the session, and session->refer_id is
+	 * only the last transfer the application acted upon, so sweep them all.
+	 * sessions_mutex is held here. */
+	if(transfers != NULL)
+		g_hash_table_foreach_remove(transfers, janus_sip_transfer_owned_by_session, session);
+	session->refer_id = 0;
 	/* Shutdown the NUA */
 	if(session->stack) {
 		janus_mutex_lock(&session->stack->smutex);
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -6572,6 +6572,9 @@ void janus_sip_sofia_callback(nua_event_t event, int status, char const *phrase,
 			if(sip == NULL || sip->sip_refer_to == NULL) {
 				JANUS_LOG(LOG_ERR, "Missing Refer-To header\n");
 				nua_respond(nh, 400, sip_status_phrase(400), TAG_END());
+				/* REFER is an application method, so the stack will not destroy
+				 * the per-request handle after the response we just sent */
+				janus_sip_destroy_unbound_handle(session, nh, hmagic);
 				break;
 			}
 			/* Access the headers we need */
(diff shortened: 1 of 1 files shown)


lminiero:
Thanks for the PR. To be honest the OOM every 60 hours seems a bit surprising, considering the SIP plugin is quite popular and a lot of people use it in production without experiencing this (at least to my knowledge). That said, it makes sense to check if this is actually an issue or not. These are a lot of changes, so I'll have a deeper look next week.

alvm:
Thanks for looking, and the surprise is fair — I'd have the same reaction. Let me
give you the numbers in a form you can check, because the rate is not a property
of the plugin so much as of one specific traffic pattern, and that explains why
it is not something most deployments would ever see.

**Scale first, since it is easy to misread.** The OOM kills were a container
hitting a **cgroup limit of 3 GiB**, not a machine running out of memory. At
~48 MB/h that limit is reached in about 64 hours, which is where "every two to
three days" comes from. On a host without a limit the same instance would simply
keep growing; the 60-hour figure is the limit divided by the rate, nothing more.

**The rate follows from one multiplication.** Retention is one nua handle per
auto-answered out-of-dialog request, and its cost is sofia's, not the plugin's:
about 2.3 kB on musl and about 2.9 kB on glibc, handle plus `soa_clone()` plus
the `nua_dialog_store_peer_info()` copies. In the affected deployment the PBX
pings **every registered binding with an OPTIONS on a 15 s cadence**. One
instance carried 89 bindings:

    89 bindings / 15 s        = 5.9 requests/s
    5.9 req/s x 2.3 kB        = 13.6 kB/s = ~48 MB/h
    3 GiB limit / 48 MB/h     = ~64 h to the OOM kill

Measured independently on that instance with uprobes on the stock build:
**898 `nua_stack_incoming_handle` = 898 `nh_create_handle` = 898 `soa_clone`
against 0 `nua_handle_destroy` over 180 s** — 5.0 handles/s retained, which is
the ping rate.

**Why a popular plugin would not be full of these reports.** The defect needs
incoming *out-of-dialog* OPTIONS, MESSAGE or INFO, arriving continuously, at a
session that lives for days. Most deployments have none of that: SIP keepalives
usually run the other way (the client pings the registrar), and a plugin session
that ends after a call takes its handles with it when the session is destroyed.
The same fleet makes the point better than I can argue it — one instance there
runs **2 574 threads and ~45 000 REGISTER/h and does not leak at all**, because
its PBX sends it no OPTIONS. Two other quiet instances on the same unfixed build:
flat for weeks. The discriminator was never load; it was whether anything pings
inward.

**Provenance, stated plainly.** Those absolute numbers come from a deployment
running a downstream fork of this plugin, and that deployment had other leaks at
the same time, some in presence/PUBLISH code that does not exist upstream. What
transfers to master is the mechanism and the per-request cost, and I checked that
rather than assuming it: the `nua_i_options` handler is **byte-identical** between
`4602fcc5` and the fork, and the handle is created by nua
(`nua_stack_incoming_handle`, nua_server.c) and left undestroyed because
`sip_response_terminates_dialog()` returns 0 for any 2xx, so
`nua_base_server_report()` returns before its `nh_destroy()` branch. Nothing in
that chain is fork-specific.

**Reproducing it without our stand**, on plain master, in a couple of minutes:
register one account against any PBX or a `sipp` UAS, then send it OPTIONS
out-of-dialog at a steady rate — 20/s is plenty — and watch either RSS or, more
directly, the handle count. `sipp -sn uas` on one side and a loop of
`sipp -sn options` at the plugin's SIP port on the other reproduces it; with
`--debug-level=7` you can also see the stack auto-answering 200 with no
`nua_handle_destroy` following. Happy to send the exact scripts we used, or a
short capture, if that is easier than assembling it.

**On the size of the PR.** It is eight commits because it is eight distinct
lifetime bugs found in the same subsystem, and each commit stands alone with its
own mechanism and measurement — the first one (auto-answered OPTIONS/MESSAGE/INFO)
is the one that accounts for the growth above, and it is four lines plus a helper.
If it would be easier to review, I am glad to split it: that first commit as its
own PR, and the remaining seven — call-handle, transfer teardown, the two
transfer-table fixes, the JSON reference, the subscription handle — in a second.
Say the word and I will reopen it that way; no rush at all, next week is fine.

lminiero:
I wonder when I'll get to talk to the real Oleksii, instead of his LLM :stuck_out_tongue_winking_eye: 
Thanks for the clarification, anyway.

alvm:
> I wonder when I'll get to talk to the real Oleksii, instead of his LLM 😜 Thanks for the clarification, anyway.

Here I am, carefully managing every step of my AI. :)

I spent days chasing memory growth on a handful of production instances, and the first commit in this PR is what fixed almost all of it — I could have stopped right there.

By that point, though, I had the probes, the metrics and a way to A/B my own patches against live traffic. And, being armed with AI ;D, another few days of it seemed cheap. So I kept pulling the thread and collected the rest of the lifetime and memory-management issues I could find in the same subsystem. Each of them is much smaller than the first, but it felt like a waste to leave all that instrumentation idle with the job half finished.

alvm:
Hi @lminiero — a gentle ping on this one.
#3670 and #3671 are merged now (thanks!), so this handle-lifetime PR is the last of the set.

lminiero:
There are a ton of changes in this, and I don't have time to look into this at the moment, sorry, especially considering I'll be abroad for a conference soon. While I plan to review this, I don't expect it will happen before a couple of weeks.

PR #3670, opened 2026-09-04
WebSockets event handler: bound the outgoing queue, and keep the cap armed

alvm:
Two fixes for the same failure: the outgoing event queue growing without limit
while the handler believes it is connected.

**`events_cap_on_reconnect` is consulted only in the `reconnect` state.** A
backend that completes the WebSocket handshake and then stops reading never
puts the handler into that state — lws cannot drain the socket, nothing pops
the queue, and events accumulate until lws itself gives up on the connection
minutes later. Measured with `events = "all"`, grouping and broadcast at only
~20 events/s: RSS grew 25.4 kB/s (about 91 MB/h) for three and a half minutes,
then the stalled connection was dropped and 3 002 queued events were discarded
in one go. The cycle repeated with the same amplitude and RSS never came back.
One caveat on those bytes, since the run was done on a downstream build: it
serialised events with `JSON_INDENT(3)`, while master has defaulted to
`JSON_COMPACT` since #3633. The event count is unaffected — the queue is
unbounded either way — but on master the same stall holds fewer bytes for the
same number of events.

The first commit adds `events_max_queue`, a ceiling that applies whatever the
connection state is; when both caps are set the smaller wins, and it defaults
to 0, which keeps today's behaviour for anyone who does not set it.

**The cap is also disarmed while a connect attempt is in flight.**
`janus_wsevh_connect_attempt()` clears `reconnect` as soon as
`lws_client_connect_via_info()` returns a wsi, but that only *starts* an
asynchronous connect: with `connect_timeout_secs = 5` the failure arrives up to
five seconds later, and for that whole window the cap does not apply. A backend
down long enough for the retry delay to reach its 8 s maximum leaves the queue
uncapped for roughly a third of the outage. `LWS_CALLBACK_CLIENT_ESTABLISHED`
already clears the flag where the connection really exists, so on lws >= 3.2 the
premature clear can simply go; the pre-3.2 service loop keys off the same flag
to decide whether to reconnect, so it keeps the old behaviour. The version guard
is the one already used at the top of this file.

The new setting is documented in `janus.eventhandler.wsevh.jcfg.sample` and can
be changed at runtime through the existing `tweak` request.

### DIFF
--- a/conf/janus.eventhandler.wsevh.jcfg.sample
+++ b/conf/janus.eventhandler.wsevh.jcfg.sample
@@ -45,6 +45,16 @@ general: {
 						# events when a reconnection actually ends up taking place.
 	# events_cap_on_reconnect = 10
 
+						# 'events_cap_on_reconnect' only applies while we are actually
+						# reconnecting. A backend that completes the WebSocket handshake
+						# and then stops reading never puts us in that state, so nothing
+						# pops the queue and it grows until libwebsockets itself gives up
+						# on the connection minutes later. 'events_max_queue' is an
+						# absolute ceiling that applies whatever the connection state is:
+						# when both are set the smaller one wins. As above, 0 disables it
+						# (the default), so the minimum useful value is 1.
+	# events_max_queue = 20000
+
 						# In case you need to debug connection issues, you can configure
 						# the libwebsockets debugging level as a comma separated list of things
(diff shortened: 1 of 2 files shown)


lminiero:
Looks good to me, thanks, merging!

PR #3671, opened 2026-09-04
WebSockets event handler: report the outgoing queue depth over the Admin API

alvm:
**Depends on #3670 — it reports `events_max_queue`, which that PR introduces.
Please merge this one second, or say the word and I will rebase it to stand
alone.**

There is no in-band way to see how far behind the event handler is.
`janus_wsevh_handle_request()` implements only `tweak` and answers with a bare
result or error, and the Admin API's `query_eventhandler` forwards to it, so the
queue depth is invisible. The only signal today is the "N events lost in between
reconnections" log line, emitted after the connection has already been dropped —
minutes too late to be a diagnostic.

That matters because a stalled backend and a leak in the server produce
numerically similar growth slopes, and an operator needs a cheap way to tell
them apart. Send-Q on the handler's socket is the out-of-band check; this is the
in-band one.

    {"janus":"query_eventhandler","handler":"janus.eventhandler.wsevh",
     "request":{"request":"info"},"transaction":"x","admin_secret":"..."}

The reply carries the queue depth, whether the handler considers itself
connected, the events dropped since the last reconnection, and the three
settings `tweak` can change. Note that on lws < 3.2 `connected` reads true while
a connect attempt is still in flight, since the flag is cleared when the attempt
starts.

This is a feature rather than a fix, which is why it is not folded into the
queue PR.

### DIFF
--- a/conf/janus.eventhandler.wsevh.jcfg.sample
+++ b/conf/janus.eventhandler.wsevh.jcfg.sample
@@ -45,6 +45,16 @@ general: {
 						# events when a reconnection actually ends up taking place.
 	# events_cap_on_reconnect = 10
 
+						# 'events_cap_on_reconnect' only applies while we are actually
+						# reconnecting. A backend that completes the WebSocket handshake
+						# and then stops reading never puts us in that state, so nothing
+						# pops the queue and it grows until libwebsockets itself gives up
+						# on the connection minutes later. 'events_max_queue' is an
+						# absolute ceiling that applies whatever the connection state is:
+						# when both are set the smaller one wins. As above, 0 disables it
+						# (the default), so the minimum useful value is 1.
+	# events_max_queue = 20000
+
 						# In case you need to debug connection issues, you can configure
 						# the libwebsockets debugging level as a comma separated list of things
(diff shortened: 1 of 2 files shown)


lminiero:
You say this depends on #3670, but it looks like it includes it in its entirety. I guess that's because the base for this PR is set to master, rather than your other PR. Please do that instead, so that we can check only the changes of this PR, and not both.

alvm:
You're right that it includes #3670 in its entirety, but I can't point the base at that branch: it lives in my fork, and when a PR comes from a fork GitHub only offers branches of the base repository as the target. The base here can only be one of meetecho's own branches, so `wsevh-bound-queue` isn't selectable.

Two ways out, whichever suits you better. If #3670 goes in first, this PR will show only its own commit — the diff is computed from the merge base, so the two queue commits drop out of it as soon as they're in master. Otherwise I can make this one standalone: the only thing it borrows from #3670 is a single field in the `info` reply (`events_max_queue`), and dropping that leaves one commit against master. Just say which you'd prefer.

lminiero:
Let's leave it like that, no worries.

PR #3672, opened 2026-09-09
Replace sleep with gcond

adnanel:
We are currently seeing dropped/delayed calls because SIP registration takes long. We are trying to host thousands of calls per janus instance. 

On janus, request messages are processed sequentially on a single thread, and if the action being processed is register, there is a sleep 100ms that blocks this thread. If this happens often (which it does, in our testing), it causes dramatic slowdowns that accumulate as messages queue up.

A first band aid fix of replacing 100ms with 1ms fixed it dramatically, but i went for a cleaner approach if im already touching this.

Instead of sleeping, im adding a gcond and simply waiting for it to be fulfilled with the same 2 second timeout.

Of course this isnt ideal either, the real proper fix would be to perform this async, but that would be quite a refactor :)

### DIFF
--- a/src/plugins/janus_sip.c
+++ b/src/plugins/janus_sip.c
@@ -1364,8 +1364,8 @@ typedef struct janus_sip_session {
 	gboolean helper;		/* Whether this session is a helper or not */
 	GList *helpers;			/* The helper sessions, if this is the "master" */
 	janus_mutex mutex;
-	GMutex stack_mutex;			/* Protects the Sofia stack readiness signaling below */
-	GCond stack_cond;			/* Signaled by the Sofia thread once the NUA stack (s_nua) is ready */
+	janus_mutex stack_mutex;		/* Protects the Sofia stack readiness signaling below */
+	janus_condition stack_cond;		/* Signaled by the Sofia thread once the NUA stack (s_nua) is ready */
 	char *hangup_reason_header;
 	char *hangup_reason_header_protocol;
 	char *hangup_reason_header_cause;
@@ -1461,8 +1461,8 @@ static void janus_sip_session_free(const janus_refcount *session_ref) {
 		g_free(session->stack);
 		session->stack = NULL;
 	}
-	g_cond_clear(&session->stack_cond);
-	g_mutex_clear(&session->stack_mutex);
+	janus_condition_destroy(&session->stack_cond);
+	janus_mutex_destroy(&session->stack_mutex);
 	if(session->account.proxy) {
 		g_free(session->account.proxy);
 		session->account.proxy = NULL;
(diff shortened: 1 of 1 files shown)


lminiero:
The problem is not SIP registrations, but the time to create the SIP stack in its dedicated thread. I guess one way to get around that would be to pre-create it when the SIP handle is created, so in the "create_session" callback, but at that point we still don't know the registration/contact info, so it's a no-go. Another possible alternative could be using a single SIP stack for all sessions (which was part of my effort in the "SIP trunking" branch), but that would require quite some work. As such, I think your condition based mechanism is probably the way to go.

adnanel:
I tested this with a few test runs. My stress test is fragile but i managed to measure 10 registrations per second on master, and 89.3 registrations per second on this branch.

I'd consider this ready for review.

adnanel:
Ok i spent the evening writing a better stress test to ensure this is now fast enough for our load. I can't share the code as it wraps around our API and only indirectly causes registrations on janus.

We consider a registration successful if we receive registered in 2 seconds after sending register.


Before / master:
```
20 reg/sec -> 100% success rate, throughput is 9.8 reg/sec
50 reg/sec -> 51% success rate, throughput is 8.3 reg/sec
100 reg/sec -> 8% success rate, throughput is 2.8 reg/sec
200 reg/sec -> 0% success rate
```

After my fix:
```
20 reg/sec -> 100% success rate, throughput is 10.4 reg/sec
50 reg/sec -> 100% success rate, throughput is 46.9 reg/sec
100 reg/sec -> 84% success rate, throughput is 32.3 reg/sec
200 reg/sec -> 81% success rate, throughput is 47.0 reg/sec
```

Keep in mind that this is a single dev instance of janus with very little resources, but the comparison is what matters. 

It's not ideal, i'd rather measure time between register and registering, as that's essentially the time between enqueuing and dequeuing the message, but it is what it is.

lminiero:
It's still 10x better, so that's definitely something! I'd only check if using conditions can cause issues if, e.g., the handle is destroyed while the thread is being started. In theory it shouldn't considering the reference counters, but it may be worth making sure. In case that ends up working fine, no objection from me to merge (I'd take care of backporting it to `0.x`, which should be trivial enough).

adnanel:
Yes looking at the code the only place where the mutexes are destroyed is janus_sip_session_free, which hits only when refcount hits zero, and the handler holds the ref the entire time. This is the main reason why I avoided refactoring the flow to be async, as I didn't want to leave the existing controller scope / object ownership.

Another "proof" is that we've been running this on a few prod machines for the past ~24 hours and I haven't been woken up yet, with couple hundred registrations per second at peak currently. We're hoping to get 10x traffic on these instances in the coming days, so if you know of a better way to test this let me know, as far as my know-how goes this seems fine.

lminiero:
I think testing the way you are is great! As it gives us feedback from practical and realistic scenarios. In case you don't notice anything broken by Monday, let me know and I'll merge :pray:

adnanel:
So my snapshot has been running on 2 prod machines since monday morning / sunday night, and we processed almost 2.5 million calls via these janus instances. Here is the graph of registrations per second per janus instance:

The hw config of these instance is pretty modest, 8GB RAM and 8 vcpus.

_Ignore the dip to zero in the first day, metric scraping failed for a moment. Also note that the bars are stacking, i.e. the total is ~45 reg/sec max_

image

No crashes yet, so I'd consider this good enough :D

lminiero:
Thanks for the tests and feedback! Ok to merge for me :pray:

